PR #184, opened 2023-07-17
Fix `class_hash` hack in Declare cheatcode

abulenok:


starknet-foundry/crates/forge/src/cheatcodes_hint_processor.rs
```diff
@@ -476,6 +471,17 @@ fn declare(
...
+fn get_class_hash(casm_contract: &str) -> ClassHash {
+    let compiled_class = serde_json::from_str::<CompiledClass>(casm_contract)
+        .expect("Failed to deserialize casm contract into CompiledClass");
+    let class_hash = compiled_class
+        .class_hash()
+        .expect("Failed to compute class_hash for CompiledClass");
+    let class_hash = StarkFelt::new(class_hash.to_bytes_be())
+        .unwrap_or_else(|_| panic!("Failed to create StarkFelt from class_hash: {class_hash:?}"));
+    ClassHash(class_hash)
+}
```
cptartur, line 483: I think this should return `Result<ClassHash>` so we can handle it like other errors. Not by unwrapping.

cptartur: Errors should be bubbled `?`

starknet-foundry/crates/forge/src/cheatcodes_hint_processor.rs
```diff
@@ -731,4 +739,45 @@ mod test {
...
+                "target/dev/simple_package_HelloStarknet.casm.json",
+            ),
+        ];
+
+        for (expected_class_hash, casm_contract_path) in cases {
+            let casm_contract_path = temp_dir_path.join(casm_contract_path);
+            let casm_contract_path = casm_contract_path.as_path();
+
+            let casm_contract_definition = std::fs::read_to_string(casm_contract_path)
+                .unwrap_or_else(|_| panic!("Failed to read file: {casm_contract_path:?}"));
```
cptartur, line 775: Can just use `.unwrap()` here.

cptartur (changes requested):

cptartur (approved):

Issue #68, opened 2023-07-06
Export cheatcodes as scarb package


cptartur:
https://github.com/software-mansion/protostar/issues/1997

Currently corelib is configured by setting a `CARGO_MANIFEST_DIR` environment variable. We should consider a better way of detecting the corelib.

Note: as of creation of this issue, corelib provided by scarb cannot be used. Protostar uses a modified corelib.

piotmag769:
Blocked until Cairo 2.1.0 is released

piotmag769:
https://github.com/foundry-rs/starknet-foundry/pull/212#discussion_r1269072821

cptartur:
Can resolving #331 unblock this?

piotmag769:
@cptartur it should

PR #212, opened 2023-07-18
Remove cairo fork

piotmag769:


starknet-foundry/crates/forge/src/scarb.rs
```diff
@@ -140,7 +140,14 @@ pub fn dependencies_for_package(
...
 
-    Ok((base_path, dependencies, target_name))
+    let corelib = compilation_unit
+        .components
+        .iter()
+        .find(|du| du.source_path.to_string().contains("core/src"))
+        .context("corelib could not be found")?;
+    let corelib_path = Utf8PathBuf::from(corelib.source_root());
+
+    Ok((base_path, corelib_path, dependencies, target_name))
```
piotmag769, line 150: This function returns a corelib, but for now the corelib bundled with scarb is not compatible with the compiler version we use (see comments in `Cargo.toml`). It is prepared to be used when 2.1.0 is release - we can then bump Cairo to 2.1.0 and use the corelib provided by appropriate Scarb version

starknet-foundry/crates/forge/Cargo.toml
```diff
@@ -15,15 +15,15 @@ tempfile.workspace = true
...
-cairo-lang-filesystem = { path = "../../../cairo/crates/cairo-lang-filesystem", version = "2.0.1" }
+cairo-lang-test-runner = { git = "https://github.com/starkware-libs/cairo", rev = "f9e7b41" }
+cairo-lang-runner = { git = "https://github.com/starkware-libs/cairo", rev = "f9e7b41" }
+cairo-lang-sierra-to-casm = { git = "https://github.com/starkware-libs/cairo", rev = "f9e7b41" }
+cairo-lang-sierra = { git = "https://github.com/starkware-libs/cairo", rev = "f9e7b41" }
+cairo-lang-utils = { git = "https://github.com/starkware-libs/cairo", rev = "f9e7b41" }
+cairo-lang-casm = { git = "https://github.com/starkware-libs/cairo", rev = "f9e7b41" }
+cairo-lang-starknet = { git = "https://github.com/starkware-libs/cairo", rev = "f9e7b41" }
+cairo-lang-compiler = { git = "https://github.com/starkware-libs/cairo", rev = "f9e7b41" }
+cairo-lang-filesystem = { git = "https://github.com/starkware-libs/cairo", rev = "f9e7b41" }
```
piotmag769, line 26: It has to be this way for now, since 2.0.2 doesn't include changes that made some much needed functions public. We cannot use 2.1.0-rc0 since !(2.1.0-rc0 >= 2.0.0) and blockifier uses 2.0.0.

This commit points to the main branch merged with 2.0.2

starknet-foundry/crates/forge/src/main.rs
```diff
@@ -71,7 +71,8 @@ fn main_execution() -> Result<()> {
 
     for package in &scarb_metadata.workspace.members {
         let forge_config = forge::scarb::config_from_scarb_for_package(&scarb_metadata, package)?;
-        let (base_path, dependencies, target_name) =
+
+        let (base_path, _corelib_path, dependencies, target_name) =
```
piotmag769, line 75: `_corelib_path` is unused for now, see the comment in `starknet-foundry/crates/forge/src/scarb.rs`

piotmag769 (commented):
The corelib is copied from the appropriate revision of Cairo repository, it is a temporary solution - see comments below

CONTRIBUTING.md
```diff
@@ -20,9 +20,6 @@ To run tests scripts, you have to install:
 - [asdf](https://asdf-vm.com/guide/getting-started.html)
 - [starknet-devnet](https://0xspaceshard.github.io/starknet-devnet/docs/intro)
 
-> ⚠️ Make sure you run `./scripts/prepare_for_tests.sh` after setting up the development environment, otherwise the
-> tests will fail.
-
```
cptartur, line 25: Is this no longer necessary? Did it only setup fork?

corelib/Scarb.toml
cptartur, line 1: Won't the fact that we are including a whole corelib cause compatibility problems with the compiler? I though we were only supposed to include the cheatcodes part and use the rest of the corelib from cairo compiler. This doesn't seem good.

starknet-foundry/crates/forge/Cargo.toml
cptartur: If we merge with this, and issue to use deps directly from crates.io should be created.

starknet-foundry/crates/forge/src/scarb.rs
```diff
@@ -140,7 +140,14 @@ pub fn dependencies_for_package(
 
     let target_name = compilation_unit.target.name.clone();
 
-    Ok((base_path, dependencies, target_name))
+    let corelib = compilation_unit
+        .components
+        .iter()
+        .find(|du| du.source_path.to_string().contains("core/src"))
```
cptartur, line 146: Just to make sure, isn't there a more direct way of finding corelib through scarb?

starknet-foundry/crates/forge/src/running.rs
```diff
@@ -69,7 +69,6 @@ pub(crate) fn run_from_test_case(
         runner: Some(runner),
         starknet_state: StarknetState::default(),
         string_to_hint,
-        blockifier_state: None,
```
cptartur, line 72: 👀

CONTRIBUTING.md
piotmag769: I checked and `test_cast.sh` runs `prepare_for_tests.sh` so I guess it's no longer needed

starknet-foundry/crates/forge/src/scarb.rs
piotmag769: Not sure, but I think this way is the best - we will have appropriate version of corelib every time + Marek told me that metadata call can sometimes change the content of the corelib (resolving etc.)

starknet-foundry/crates/forge/src/running.rs
piotmag769: Yea, I don't know why it was there in the first place XD

corelib/Scarb.toml
piotmag769: Yup we are supposed to do that in the future when we can use corelib provided by Scarb. We cannot really extract cheatcodes to package since Scarb will check its code during the build. For now, the semantic analysis will fail, since Cairo 2.0.2 does not yet include e.g. the `cheatcode` function.

Arcticae (commented):
It's certainly nice that it's removing the fork but i think the thing we wanted to achieve was extracting the cheats as a scarb library, so i'm confused why this order of action (i thought fork removal should be an effect, not the mean to an end).

Also this PR makes it impossible to tell which version of corelib we are running ATM

CONTRIBUTING.md
cptartur: I think we could still leave it here? In case someone wants to run tests with `cargo test`?

cptartur:
> It's certainly nice that it's removing the fork but i think the thing we wanted to achieve was extracting the cheats as a scarb library, so i'm confused why this order of action (i thought fork removal should be an effect, not the mean to an end).
> 
> Also this PR makes it impossible to tell which version of corelib we are running ATM

I wonder about this too, perhaps it would make sense to wait until cairo 2.1.0 so we can achieve both at the same time?

corelib/src/forge_print.cairo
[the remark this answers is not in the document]
piotmag769: https://github.com/foundry-rs/starknet-foundry/pull/212#discussion_r1269161755

corelib/src/cheatcodes.cairo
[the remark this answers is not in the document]
piotmag769: https://github.com/foundry-rs/starknet-foundry/pull/212#discussion_r1269161755

corelib/Scarb.toml
piotmag769: Yup, exactly

CONTRIBUTING.md
piotmag769: Sure, will revert it

starknet-foundry/crates/forge/Cargo.toml
piotmag769: I think it is kinda obvious, but will link this comment in the issue

piotmag769: https://github.com/foundry-rs/starknet-foundry/issues/68#issuecomment-1643615388

piotmag769:
I wouldn't wait for 2.1.0, I think we can merge it and get rid of the fork for now, then export cheatcodes as package when 2.1.0 is released

cptartur (approved):

PR #232, opened 2023-07-19
Split forge tests, so they can be filtered out

THenry14:
## Introduced changes

- split e2e and integration tests, so they can now be invoked separately:
```
cargo test integration --manifest-path ./starknet-foundry/Cargo.toml -p forge
cargo test e2e --manifest-path ./starknet-foundry/Cargo.toml -p forge
```
of course running all tests is still possible using the same command as prviously

- Run e2e tests on separate runner, to decrease overall forge tests execution time

## Breaking changes

None

## Checklist

- [X] Linked relevant issue
- [X] Added relevant tests


.github/workflows/ci.yml
```diff
@@ -89,7 +110,7 @@ jobs:
   clippy:
     runs-on: ubuntu-latest
     env:
-      # Make sure CI fails on all warnings, including Clippy lints.
+      # Make sure CI fails on all warnings - including Clippy lints.
```
THenry14, line 113: this was just to trigger workflow with generated cach, don't mind this

Arcticae: You can always re-run manually BTW, and it should have the cache

Arcticae (approved):
Great idea! :)
This is kind of temporary measure and we perhaps consider more automatic solution, or a pattern of splitting those (cairo-lang has high granularity crates and the tests are blazingly fast :P)

THenry14:
> This is kind of temporary measure and we perhaps consider more automatic solution, or a pattern of splitting those (cairo-lang has high granularity crates and the tests are blazingly fast :P)

100%, this is just temp solution :+1:

.github/workflows/ci.yml
```diff
@@ -26,7 +26,28 @@ jobs:
       - uses: software-mansion/setup-scarb@v1.0.0
         with:
           scarb-version: ${{ env.SCARB_VERSION }}
-      - run: cargo test --manifest-path ./starknet-foundry/Cargo.toml -p forge
+      - run: cargo test --lib --manifest-path ./starknet-foundry/Cargo.toml -p forge
```
cptartur, line 29: Shouldn't this command be removed?

cptartur (commented):
If it works, I think we can merge it. 

Though I'm unsure about one thing: Each file in `test` directory is treated as separate create, I wonder if grouping them like that changes this behavior.

.github/workflows/ci.yml
```diff
@@ -8,8 +8,8 @@ on:
       - master
 
 jobs:
-  test-forge:
-    name: Test Forge
+  test-forge-unit-and-integration:
```
karol-bisztyga, line 11: hmm, why unit and integration tests are together? couldn't we separate them into two jobs?

karol-bisztyga (approved):

karol-bisztyga (changes requested):
I think we should add information to readme

.github/workflows/ci.yml
piotmag769: Units tests are really fast so I don't think we benefit a lot from splitting them

.github/workflows/ci.yml
piotmag769: Isn't it to run unit tests?

piotmag769 (approved):

.github/workflows/ci.yml
THenry14: Exactly, --lib just runs unit tests

THenry14:
> I think we should add information to readme

What info exactly do you have in mind? Things changed for CI, nothing changed from user PoV (the tests still can be run the same as previously)

THenry14:
> Though I'm unsure about one thing: Each file in `test` directory is treated as separate create, I wonder if grouping them like that changes this behavior.

not sure about this either - basically I just renamed main to mod, so I think we should be good? but perhaps some better eperts can have a say

cptartur:
> > Though I'm unsure about one thing: Each file in `test` directory is treated as separate create, I wonder if grouping them like that changes this behavior.
> 
> not sure about this either - basically I just renamed main to mod, so I think we should be good? but perhaps some better eperts can have a say

cc @mkaput

mkaput:
I'm not up-to-date with the discussion but from my experiments in Scarb, the best gains I got by a) splitting integration tests into as many test binaries as possible b) extracting shared tests code to a separate private crate c) only running PR tests on Linux and d) using [nextest and its partitioning feature](https://nexte.st/book/partitioning.html) to let it offload running tests on multiple runners.

THenry14:
> I'm not up-to-date with the discussion but from my experiments in Scarb, the best gains I got by a) splitting integration tests into as many test binaries as possible b) extracting shared tests code to a separate private crate c) only running PR tests on Linux and d) using [nextest and its partitioning feature](https://nexte.st/book/partitioning.html) to let it offload running tests on multiple runners.

Cool, I think this all should be doable here. For now though, we were mainly wondering about:



 > Though I'm unsure about one thing: Each file in test directory is treated as separate create, I wonder if grouping them like that changes this behavior.

> not sure about this either - basically I just renamed main to mod, so I think we should be good? but perhaps some better eperts can have a say

any thoughts on that? @mkaput 🙏

mkaput:
Yeah, from what I see here you have coalesced all tests into a single crate there, because there is only one crate root now (a test named `main` defined by `main.rs` file)

karol-bisztyga:
> > I think we should add information to readme
> 
> What info exactly do you have in mind? Things changed for CI, nothing changed from user PoV (the tests still can be run the same as previously)

I was thinking we could add information on how to run integration tests and how to run e2e tests separately, not just altogether using `cargo test`.

karol-bisztyga (approved):

THenry14:
> > > I think we should add information to readme
> > 
> > 
> > What info exactly do you have in mind? Things changed for CI, nothing changed from user PoV (the tests still can be run the same as previously)
> 
> I was thinking we could add information on how to run integration tests and how to run e2e tests separately, not just altogether using `cargo test`.

Sure, makes sense. Added :+1:.

PR #238, opened 2023-07-20
[CLI] `account` command

### LINKED ISSUE #7: Add deploy account command
https://github.com/software-mansion/protostar/issues/2057

war-in:
Closes #7 

## Introduced changes

- added `account` command:
   - `create` subcommand
   - `deploy` subcommand
- added relevant tests

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


starknet-foundry/crates/cast/src/starknet_commands/account/create.rs
```diff
@@ -0,0 +1,187 @@
...
+    network: &str,
+    maybe_salt: Option<FieldElement>,
+    add_profile: bool,
+) -> Result<Vec<(&'static str, String)>> {
+    let private_key = SigningKey::from_random();
+    let public_key = private_key.verifying_key();
+    let salt = match maybe_salt {
+        Some(salt) => salt,
+        None => FieldElement::from(OsRng.next_u64()),
+    };
```
karol-bisztyga, line 53: https://doc.rust-lang.org/std/option/enum.Option.html#method.unwrap_or

starknet-foundry/crates/cast/src/starknet_commands/account/create.rs
```diff
@@ -0,0 +1,187 @@
...
+            signer,
+            provider,
+        )
+        .await?;
+        let deployment = factory.deploy(salt);
+
+        deployment.estimate_fee().await?.overall_fee
+    };
+
+    if !accounts_file_path.exists() {
```
karol-bisztyga, line 76: What is the point of this? Maybe we should break immediately? From what I understand this is going to break anyway since there will be no accounts and a desired account will not be found, right?

starknet-foundry/crates/cast/src/starknet_commands/account/create.rs
```diff
@@ -0,0 +1,187 @@
...
+        }
+        Err(error) => {
+            print_formatted(vec![("error", error.to_string())], int_format, json, true)?;
+        }
+    };
+
+    Ok(())
+}
+
+pub fn add_created_profile_to_configuration(name: String, network: String, url: String) -> String {
```
karol-bisztyga, line 139:
```suggestion
pub fn add_created_profile_to_configuration(name: String, network: String, url: String) -> Result<String> {
```
why do you return string, not result?

karol-bisztyga (changes requested):

starknet-foundry/crates/cast/src/starknet_commands/account/create.rs
war-in: As you can see we check if accounts file exists and if not we just create it

starknet-foundry/crates/cast/src/main.rs
```diff
@@ -1,6 +1,8 @@
 use crate::helpers::scarb_utils::parse_scarb_config;
+use crate::starknet_commands::account::create::print_account_create_result;
+use crate::starknet_commands::account::Account;
```
THenry14, line 3:
```suggestion
use crate::starknet_commands::account::{
    create::print_account_create_result,
    Account
};
```

nitpick

starknet-foundry/crates/cast/src/main.rs
```diff
@@ -79,7 +84,8 @@ async fn main() -> Result<()> {
 
     let accounts_file_path =
         Utf8PathBuf::from(shellexpand::tilde(&cli.accounts_file_path).to_string());
-    if !&accounts_file_path.exists() {
+    if let Commands::Account(_) = &cli.command {
+    } else if !&accounts_file_path.exists() {
         bail! {"Accounts file {} does not exist! Make sure to supply correct path to accounts file.", cli.accounts_file_path}
     }
```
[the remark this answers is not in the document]
THenry14: I'd just create this file here, if it does not exist.
Or even better - it would be cool to create the file if the flag is of default value, and error out if someone passes custom path (not sure if it is possible though)

starknet-foundry/crates/cast/src/starknet_commands/account/create.rs
```diff
@@ -0,0 +1,187 @@
...
+use starknet::providers::JsonRpcClient;
+use starknet::signers::{LocalWallet, SigningKey};
+use std::fs::OpenOptions;
+use std::io::Write;
+use toml::Value;
+
+#[derive(Args, Debug)]
+#[command(about = "Create an account with all important secrets")]
+pub struct Create {
+    /// Account name under which secrets are going to be saved
```
THenry14, line 24: Technically, not only secrets. Maybe something like "... account information is going to be saved" ? or something similar?

starknet-foundry/crates/cast/src/starknet_commands/account/create.rs
```diff
@@ -0,0 +1,187 @@
...
+    pub name: String,
+
+    /// Salt for the address
+    #[clap(short, long)]
+    pub salt: Option<FieldElement>,
+
+    // If passed, a profile with corresponding data will be created in Scarb.toml
+    #[clap(short, long)]
+    pub add_profile: bool,
+    // TODO: think about supporting different account providers
```
THenry14, line 35: if this is not going to be added in this PR, let's create an issue for that (we can link it here)

starknet-foundry/crates/cast/src/starknet_commands/account/create.rs
```diff
@@ -0,0 +1,187 @@
...
+        Err(error) => {
+            print_formatted(vec![("error", error.to_string())], int_format, json, true)?;
+        }
+    };
+
+    Ok(())
+}
+
+pub fn add_created_profile_to_configuration(name: String, network: String, url: String) -> String {
+    let manifest_path = find_manifest_path(None).expect("Failed to obtain Scarb.toml file path");
```
THenry14, line 145: should we allow passing `path-to-scarb-toml` here?

starknet-foundry/crates/cast/src/starknet_commands/account/create.rs
```diff
@@ -0,0 +1,187 @@
...
+        return format!("Failed to add {name} profile to the Scarb.toml. Profile already exists");
+    }
+
+    let toml_string = {
+        let mut tool_sncast = toml::value::Table::new();
+        let mut new_profile = toml::value::Table::new();
+
+        new_profile.insert("network".to_string(), Value::String(network));
+        new_profile.insert("url".to_string(), Value::String(url));
+        new_profile.insert("account".to_string(), Value::String(name.clone()));
```
THenry14, line 172: Would it be possible to use `CastConfig` here?

starknet-foundry/crates/cast/src/starknet_commands/account/deploy.rs
```diff
@@ -0,0 +1,18 @@
+use camino::Utf8PathBuf;
+use clap::Args;
+use starknet::core::types::FieldElement;
+
+#[derive(Args, Debug)]
+#[command(about = "Deploy an account to the Starknet")]
+pub struct Deploy {
+    /// Path to the file where the account secrets are stored
```
THenry14: what is this file? is it something different than accounts-file ?

starknet-foundry/crates/cast/src/main.rs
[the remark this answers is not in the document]
war-in: I'm not sure if we should create this file here. It would apply to every command and if someone is using cast for calls only it is unwanted behaviour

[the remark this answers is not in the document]
THenry14: Yeah, I thought so as well. Maybe wqe could approach it somewhat different, like adjusting `declare` to instruct to create/add account to file, if the file does not exist? I wonder if, in that scenario, we should count on user to make it work, or try and write some kind of helper for this. What are your thoughts, gents?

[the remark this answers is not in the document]
THenry14: btw we could also add this info, if file exists but account does not exist in this file, probably

starknet-foundry/crates/cast/src/starknet_commands/account/create.rs
war-in: Maybe, but I do not see benefits. We would have to create CastConfig instance and then insert each value to the `new_profile`, right?

THenry14: It probably depends on `toml` package - perhaps it can do something like `CastConfig.to_string()`? I don't know, that was my question :D

starknet-foundry/crates/cast/src/starknet_commands/account/deploy.rs
THenry14: sooo, which one is it? :D

starknet-foundry/crates/cast/src/starknet_commands/account/create.rs
karol-bisztyga: Yes but what is the point of creating this file? It's not intuitive, because a user gets something like this: `account xxx does not exist` with no clue it comes from a certain file, right? We could print an error like `accounts.json does not exist, create one and put accounts there ...`.

starknet-foundry/crates/cast/src/starknet_commands/account/deploy.rs
war-in: This is accounts.json 😅

THenry14: You mean `~/.starknet_accounts/starknet_open_zeppelin_accounts.json` (by default) ? than why not use the --accounts-file flag?

war-in: Right, I've already changed it in this next PR 😅

starknet-foundry/crates/cast/src/main.rs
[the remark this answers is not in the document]
karol-bisztyga: Ok I understand now what's happening here and I think that the logic for the account file is split (here we check for existence but only if the command's not `account` and in `starknet_commands::account::create::create` we do the rest of the stuff). I do not think this is good.

I think you do not have to declare `accounts_file_path` here. You can just define a method that lazily does it or create a lazy static for this. What do you think?

starknet-foundry/crates/cast/src/starknet_commands/account/create.rs
war-in: Ok, this error is fine but how is it connected with this PR? This file is created because someone is creating an account (so its secrets have to be saved somewhere)

karol-bisztyga (approved):

starknet-foundry/crates/cast/src/starknet_commands/account/create.rs
war-in: Unfortunately it does not work as expected :/
image

war-in: https://github.com/orgs/foundry-rs/projects/3/views/2?filterQuery=milestone:Cast&pane=issue&itemId=34135288

starknet-foundry/crates/cast/tests/e2e/account/create.rs
```diff
@@ -0,0 +1,176 @@
...
+    args.append(&mut vec![
+        "account", "create", "--name", "user1", "--salt", "0x1",
+    ]);
+
+    let snapbox = runner(&args);
+
+    snapbox.assert().stderr_matches(indoc! {r#"
+        error: Account with provided name already exists in this network
+    "#});
+}
```
THenry14, line 176: Let's maybe also add test when file already exists? I think in all happy cases you create a new file right?

starknet-foundry/crates/cast/src/main.rs
```diff
@@ -215,5 +223,26 @@ async fn main() -> Result<()> {
...
+                    create.salt,
+                    create.add_profile,
+                )
+                .await;
+
+                print_account_create_result(result, cli.int_format, cli.json)?;
+                Ok(())
+            }
+            account::Commands::Deploy(_deploy) => {
+                unimplemented!()
```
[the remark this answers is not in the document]
war-in: I've disabled auto merge and created a draft for this command

THenry14 (commented):
This looks great, only thing I am missing are docs :+1:

THenry14 (approved):

PR #251, opened 2023-07-24
[CLI] Fix resolving scarb.toml

war-in:
Closes #

## Introduced changes

- `manifest_path` variable was ignored

## Checklist

- [ ] Linked relevant issue
- [ ] Added relevant tests

### DIFF
--- a/starknet-foundry/crates/cast/src/helpers/scarb_utils.rs
+++ b/starknet-foundry/crates/cast/src/helpers/scarb_utils.rs
@@ -29,11 +29,12 @@ fn get_property(
             .ok_or(anyhow!(
                 "Profile or property not found in Scarb.toml: {p}, {property}"
             )),
-        None => profiled
-            .and_then(|t| t.get(property))
-            .and_then(Value::as_str)
-            .map(String::from)
-            .ok_or(anyhow!("Property not found in tool: {property}")),
+        None => match profiled.and_then(|t| t.get(property)) {
+            Some(property) => Ok(String::from(
+                property.as_str().expect("Couldn't cast property to &str"),
+            )),
+            None => Ok(String::new()),
+        },
     }
 }
 
@@ -132,10 +133,11 @@ mod tests {
             &None,
             &Some(Utf8PathBuf::from("tests/data/files/noconfig_Scarb.toml")),
         )
-        .unwrap_err();
-        assert!(config
-            .to_string()
-            .contains("Property not found in tool: url"));
+        .unwrap();
+
+        assert!(config.rpc_url.is_empty());
+        assert!(config.network.is_empty());
+        assert!(config.account.is_empty());
     }
 
     #[test]
@@ -158,9 +160,8 @@ mod tests {
             &None,
             &Some(Utf8PathBuf::from("tests/data/files/somemissing_Scarb.toml")),
         )
-        .unwrap_err();
-        assert!(config
-            .to_string()
-            .contains("Property not found in tool: account"));
+        .unwrap();
+
+        assert!(config.account.is_empty());
     }
 }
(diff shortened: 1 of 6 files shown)


THenry14 (approved):

starknet-foundry/crates/cast/src/helpers/scarb_utils.rs
```diff
@@ -31,30 +31,34 @@ fn get_property(
...
+    path: &Option<Utf8PathBuf>,
 ) -> Result<CastConfig> {
-    let manifest_path = find_manifest_path(path.as_ref().map(camino::Utf8PathBuf::as_path))
-        .expect("Failed to obtain Scarb.toml file path");
-
-    if let Some(manifest_path) = path {
-        if !manifest_path.exists() {
-            bail! {"{manifest_path} file does not exist!"};
-        };
+    let found_path = find_manifest_path(path.as_ref().map(Utf8PathBuf::as_path));
```
karol-bisztyga, line 44: are you sure this works when `path` is `None`?

starknet-foundry/crates/cast/src/helpers/scarb_utils.rs
```diff
@@ -31,30 +31,34 @@ fn get_property(
...
+            .and_then(Value::as_str)
             .map(String::from)
             .ok_or(anyhow!("Property not found in tool: {property}")),
     }
 }
 
 pub fn parse_scarb_config(
     profile: &Option<String>,
-    path: Option<Utf8PathBuf>,
+    path: &Option<Utf8PathBuf>,
```
karol-bisztyga, line 42: what is the reason you use a ref here?

starknet-foundry/crates/cast/src/helpers/scarb_utils.rs
karol-bisztyga, line 44: I don't get why we need `path.as_ref()` when `path` is already a ref.

karol-bisztyga (changes requested):

starknet-foundry/crates/cast/src/helpers/scarb_utils.rs
piotmag769, line 42: I would rename it to `cli_manifest_path` or `cli_path`, the `path` name doesn't really tell much

piotmag769 (approved):
The fix itself is fine, I would add a couple of tests though

starknet-foundry/crates/cast/src/helpers/scarb_utils.rs
piotmag769: Probably clippy

karol-bisztyga: right about the value being consumed

starknet-foundry/crates/cast/src/helpers/scarb_utils.rs
piotmag769: It should, Scarb searches for the manifest file up the working tree

war-in: I'm not an expert, but here is an error
image

piotmag769: It's because
`path` - `&Option<Utf8PathBuf>`
`path.as_ref()` - `Option<&Utf8PathBuf>`

starknet-foundry/crates/cast/src/helpers/scarb_utils.rs
war-in: Hmm, I would say that path inside the `parse_scarb_config` function tells enough 😅  but can change it if you want

karol-bisztyga (approved):

PR #252, opened 2023-07-24
Add library call syscall test

piotmag769:


starknet-foundry/crates/forge/tests/integration/syscalls.rs
```diff
@@ -0,0 +1,115 @@
...
+                    #[storage]
+                    struct Storage {}
+                
+                    #[external(v0)]
+                    fn call_libfunc_syscall(
+                        self: @ContractState, class_hash: felt252, function_selector: felt252, calldata: Span<felt252>
+                    ) -> Span<felt252> {
+                        library_call_syscall(
+                            class_hash_try_from_felt252(class_hash).unwrap(), function_selector, calldata
+                        )
```
cptartur, line 79: Nit, but this conversion could be done at test level and here we could just accept `ClassHash`

cptartur (changes requested):
I think we should test that the delegate call actually behaves like a delegate call: 

1. deploy `Executor` at some address
2. `add_two` should store something in `Executor` contract storage
3. call `add_two` through `Caller` delegate call
4. call `Executor` directly through its dispatcher, verify that the value wasn't actually set

Perhaps even more steps can be added to this for more robust verification, but that's my current idea

starknet-foundry/crates/forge/tests/integration/syscalls.rs
```diff
@@ -0,0 +1,143 @@
...
+            let prepared = PreparedContract {
+                class_hash: class_hash, constructor_calldata: @ArrayTrait::new()
+            };
+            let contract_address = deploy(prepared).unwrap();
+            
+            contract_address.try_into().unwrap()
+        }
+        
+        #[test]
+        fn test_increase_balance() {
```
cptartur, line 48:
```suggestion
        fn test_library_call() {
```

cptartur (approved):
👌

Radinyn (approved):

PR #280, opened 2023-07-26
Declare signature fixed

### LINKED ISSUE #60: Make `declare` not return `Result`
https://github.com/software-mansion/protostar/issues/1909

Some cheaatcodes, especially related to mocking technically cannot fail in recoverable manner. We should verify if these should still return a `Result` with an error code instead of just `()` (or some other type).

_Originally posted by @MaksymilianDemitraszek in https://github.com/software-mansion/protostar/pull/1891#discussion_r1192134979_

MaksymilianDemitraszek:
Closes #60 

## Introduced changes

- Updated declare signature as it should never fail

## Breaking changes

- declare signature does not return `Result` anymore, `felt252` instead

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests

### DIFF
--- a/corelib/src/cheatcodes.cairo
+++ b/corelib/src/cheatcodes.cairo
@@ -27,17 +27,13 @@ impl RevertedTransactionImpl of RevertedTransactionTrait {
     }
 }
 
-fn declare(contract: felt252) -> Result::<felt252, felt252> {
+fn declare(contract: felt252) -> felt252 {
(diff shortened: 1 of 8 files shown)


Radinyn (approved):

piotmag769:
If we merge it before the release we should also update our template

Arcticae (approved):

PR #286, opened 2023-07-26
Prepare 0.2.0 release

cptartur:
Closes #

## Introduced changes

-
-

## Checklist

- [ ] Linked relevant issue
- [ ] Added relevant tests

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -1,6 +1,6 @@
 # Instruction For Creating New Starknet Forge Releases
 
-1. Bump Starknet Forge versions in the top-level `Cargo.toml` file
+1. Bump Starknet Foundry version in the top-level `Cargo.toml` file
 2. Regenerate locks using `cargo generate-lockfile`
 3. Merge introduced changes
 4. Create a new tag in repository with format `vMAJOR.MINOR.PATCH`. This will trigger the release workflow
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -1,0 +1,29 @@
+# Instruction For Creating New Starknet Forge Releases
+
+1. Bump Starknet Forge versions in
+    1. Top-level `Cargo.toml`
+    2. `crates/forge/Cargo.toml`
+    3. `crates/cast/Cargo.toml`
+    4. Any other applicable user-facing tools
+2. Regenerate locks using `cargo generatae-lockfile`
+3. Merge introduced changes
+4. Create a new tag in repository with format `vMAJOR.MINOR.PATCH`
+5. Wait for release workflows to pass. A new draft release will be created on GitHub.
+6. Update the release contents using template below and publish it
+
+## Release Template
+
+Optional release description
+
+## What's Changed
+
+- list
+- of
+- features
+
+## Compatible Scarb Versions
+
+| Tool      | Scarb Version |
+|-----------|---------------|
+| `snforge` | `X.Y.Z`       |
+| `sncast`  | `X.Y.Z`       |
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -7,7 +7,7 @@
     4. Any other applicable user-facing tools
 2. Regenerate locks using `cargo generate-lockfile`
 3. Merge introduced changes
-4. Create a new tag in repository with format `vMAJOR.MINOR.PATCH`
+4. Create a new tag in repository with format `vMAJOR.MINOR.PATCH`. This will trigger the release workflow
 5. Wait for release workflows to pass. A new draft release will be created on GitHub.
 6. Update the release contents using template below and publish it
 
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -5,7 +5,7 @@
     2. `crates/forge/Cargo.toml`
     3. `crates/cast/Cargo.toml`
     4. Any other applicable user-facing tools
-2. Regenerate locks using `cargo generatae-lockfile`
+2. Regenerate locks using `cargo generate-lockfile`
 3. Merge introduced changes
 4. Create a new tag in repository with format `vMAJOR.MINOR.PATCH`
 5. Wait for release workflows to pass. A new draft release will be created on GitHub.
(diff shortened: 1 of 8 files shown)


RELEASING.md
```diff
@@ -0,0 +1,29 @@
+# Instruction For Creating New Starknet Forge Releases
+
+1. Bump Starknet Forge versions in
+    1. Top-level `Cargo.toml`
+    2. `crates/forge/Cargo.toml`
+    3. `crates/cast/Cargo.toml`
+    4. Any other applicable user-facing tools
+2. Regenerate locks using `cargo generatae-lockfile`
```
MaksymilianDemitraszek, line 8: typo

RELEASING.md
```diff
@@ -0,0 +1,29 @@
...
+3. Merge introduced changes
+4. Create a new tag in repository with format `vMAJOR.MINOR.PATCH`
+5. Wait for release workflows to pass. A new draft release will be created on GitHub.
+6. Update the release contents using template below and publish it
+
+## Release Template
```
MaksymilianDemitraszek, line 14:
```suggestion
# <Version>
```

docs/src/testing/testing.md
```diff
@@ -36,6 +36,11 @@ Tests: 1 passed, 0 failed, 0 skipped
 Forge considers all functions in your project marked with `#[test]` attribute as tests.
 Test functions cannot return any values and cannot take any arguments.
 
+Starknet Forge will collect tests only from these places:
+
+- any files reachable from the package root (declared as `mod` in `lib.cairo` or its children)
+- files in directories outside the main package
```
MaksymilianDemitraszek, line 42:
```suggestion
- files in all directories outside of the main package
```
all or tests dir only?

starknet-foundry/crates/forge/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "forge"
-version = "0.1.0"
+version = "0.2.0"
```
MaksymilianDemitraszek, line 3: Imo we should do a patch instead 0.1.1

MaksymilianDemitraszek (changes requested):

RELEASING.md
```diff
@@ -0,0 +1,29 @@
...
+
+1. Bump Starknet Forge versions in
+    1. Top-level `Cargo.toml`
+    2. `crates/forge/Cargo.toml`
+    3. `crates/cast/Cargo.toml`
+    4. Any other applicable user-facing tools
+2. Regenerate locks using `cargo generatae-lockfile`
+3. Merge introduced changes
+4. Create a new tag in repository with format `vMAJOR.MINOR.PATCH`
+5. Wait for release workflows to pass. A new draft release will be created on GitHub.
```
war-in, line 11: Will release workflows start automatically? And when?

docs/src/getting-started/first-steps.md
```diff
@@ -6,7 +6,7 @@ We demonstrate how to create a new project, compile, and test it.
 To start a new project with Starknet Foundry, clone the template repository
 
 ```shell
-$ git clone https://github.com/foundry-rs/starknet_forge_template.git
+git clone https://github.com/foundry-rs/starknet_forge_template.git
```
THenry14, line 9: This is a nitpick, but I believe in most cases we have `$` to indicate, that this should be run as normal user (in contrary with `#` for superuser). Perhaps we should be consistent, and either have `$` for every shell example, or for none?

starknet-foundry/crates/forge/Cargo.toml
cptartur: We are changing the behavior of test collecting, it's not just a fix

RELEASING.md
cptartur: I can change it, but version is a different input box on the release page, so it's not very useful anyways.

docs/src/testing/testing.md
cptartur: It's all files outside of the package actually. In directories or not.

RELEASING.md
cptartur: It triggers on new tag creation

docs/src/getting-started/first-steps.md
piotmag769: I think it is left this way so it can be copy-pasted to the terminal without removing the dollar sign manually. It makes sense for command snippets that are supposed to be copied 1:1 imo. In other cases - the dollar sign makes it more aesthetic, but that's just my opinion

cptartur: We use `$` in most cases and I agree it makes sense, it's just impractical to have it here because when copying the snippet to clipboard, it also includes the `$` sign.

THenry14: sure, I've no strong opinion about that :+1:

starknet-foundry/crates/cast/Cargo.toml
```diff
@@ -1,8 +1,8 @@
 [package]
 name = "cast"
 description = "Cast - a Starknet Foundry CLI client"
-version = "0.1.0"
-edition = "2021"
+version.workspace = true
```
cptartur, line 4: I've inherited the workspace settings for `cast` and `forge`, not sure if it makes sense to do the same for cheatable starknet and test collector. It kind of make sense to have separate versioning in them.

RELEASING.md
```diff
@@ -0,0 +1,29 @@
+# Instruction For Creating New Starknet Forge Releases
+
+1. Bump Starknet Forge versions in
+    1. Top-level `Cargo.toml`
+    2. `crates/forge/Cargo.toml`
+    3. `crates/cast/Cargo.toml`
+    4. Any other applicable user-facing tools
```
piotmag769, line 7: I think it is not true with `version.workspace=true` anymore

RELEASING.md
```diff
@@ -0,0 +1,25 @@
+# Instruction For Creating New Starknet Forge Releases
+
+1. Bump Starknet Forge versions in the top-level `Cargo.toml` file
```
piotmag769, line 3:
```suggestion
1. Bump Starknet Foundry version in the top-level `Cargo.toml` file
```

MaksymilianDemitraszek (approved):

PR #288, opened 2023-07-26
[CLI] Account docs

### LINKED ISSUE #285: Add `account` command docs
Lets add the docs once `account create` and `account deploy` are merged

war-in:
Closes #285 

## Introduced changes

- account create docs
- account deploy docs
- account deployment flow example

## Checklist

- [ ] Linked relevant issue
- [ ] Added relevant tests

### DIFF
--- a/docs/src/appendix/cast/account/create.md
+++ b/docs/src/appendix/cast/account/create.md
@@ -1,0 +1,22 @@
+# `create`
+Create all data required for account deployment.
+
+## Required common arguments - passed by value or specified in Scarb.toml
+
+* [`network`](../common.md#--network--n-network_name)
+* [`url`](../common.md#--url--u-rpc_url)
+
+## `--name, -n <ACCOUNT_NAME>`
+Required.
+
+Account name under which account information is going to be saved.
+
+## `--salt, -s <SALT>`
+Optional.
+
+Salt for the account address. If omitted random one will be used.
+
+## `--add-profile, -a`
+Optional.
+
+If passed, a profile with corresponding data will be created in Scarb.toml
--- a/docs/src/appendix/cast/account/deploy.md
+++ b/docs/src/appendix/cast/account/deploy.md
@@ -1,0 +1,17 @@
+# `deploy`
+Deploy previously created account to Starknet.
+
+## Required common arguments - passed by value or specified in Scarb.toml
+
+* [`network`](../common.md#--network--n-network_name)
+* [`url`](../common.md#--url--u-rpc_url)
+
+## `--name, -n <ACCOUNT_NAME>`
+Required.
+
+Name of the (previously created) account to be deployed.
+
+## `--max-fee, -m <MAX_FEE>`
+Required.
+
+Max fee for deploy account transaction.
--- a/docs/src/starknet/account.md
+++ b/docs/src/starknet/account.md
@@ -1,0 +1,49 @@
+# Creating And Deploying Accounts
+
+Starknet Foundry cast supports entire account management flow with the `sncast account create` and `sncast account deploy` commands.
+
+Do the following to start interacting with the Starknet:
+
+- create account with the `sncast account create` command
+
+    ```shell
+    $ sncast \
+      --url http://127.0.0.1:5050
+      --network testnet \
+      account create \
+      --name some-name
+      
+    Account successfully created. Prefund generated address with at least 432300000000 tokens. It is good to send more in the case of higher demand, max_fee * 2 = 864600000000
+    command: Create account
+    max_fee: 0x64a7168300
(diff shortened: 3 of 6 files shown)


docs/src/appendix/cast/account/create.md
```diff
@@ -0,0 +1,24 @@
+# `create`
+Create all data required for account deployment.
```
THenry14, line 2: This is fine actually, but I wonder if it wouldn't be better if we phrase it like:

`Prepare all prerequisites for account deployment.`

what do you think?

docs/src/appendix/cast/account/create.md
```diff
@@ -0,0 +1,24 @@
...
+Account information will be saved inside `~/.starknet_accounts/starknet_open_zeppelin_accounts.json` or in the file
+specified with `--accounts-file` argument.
```
THenry14, line 4:
```suggestion
Account information will be saved to the file specified by `--accounts-file` argument, by default `~/.starknet_accounts/starknet_open_zeppelin_accounts.json`.
```

or something similar?

docs/src/appendix/cast/account/create.md
```diff
@@ -0,0 +1,24 @@
...
+
+## Required common arguments - passed by value or specified in Scarb.toml
```
THenry14, line 6:
```suggestion
## Required common arguments - passed by CLI or specified in Scarb.toml
```

docs/src/appendix/cast/account/create.md
```diff
@@ -0,0 +1,24 @@
...
+
+## `--name, -n <ACCOUNT_NAME>`
+Required.
+
+Account name under which account information is going to be saved.
+
+## `--salt, -s <SALT>`
+Optional.
+
+Salt for the account address. If omitted random one will be used.
```
THenry14, line 19:
```suggestion
Salt for the account address. If omitted random one will be generated.
```

docs/src/appendix/cast/account/create.md
```diff
@@ -0,0 +1,24 @@
...
+
+## `--add-profile, -a`
+Optional.
+
+If passed, a profile with corresponding data will be created in Scarb.toml
```
THenry14, line 24:
```suggestion
If passed, a profile with corresponding data will be added in Scarb.toml
```

THenry14: btw, will that include rpc url? Should it?

docs/src/appendix/cast/account/deploy.md
```diff
@@ -0,0 +1,17 @@
+# `deploy`
+Deploy previously created account to Starknet.
+
+## Required common arguments - passed by value or specified in Scarb.toml
```
THenry14, line 4:
```suggestion
## Required common arguments - passed by CLI or specified in Scarb.toml
```

docs/src/starknet/account.md
```diff
@@ -0,0 +1,49 @@
+# Creating And Deploying Accounts
+
+Starknet Foundry cast supports entire account management flow with the `sncast account create` and `sncast account deploy` commands.
+
+Do the following to start interacting with the Starknet:
```
THenry14, line 5: I think it could use a rephrase. Maybe a sentence about that basically all interaction with starknet needs an account at some point, and that one could start off with an account creation? I also think it may be a good idea to say a few words about what is the difference between creating and deploying accounts.

docs/src/starknet/account.md
```diff
@@ -0,0 +1,49 @@
...
+    ```
+
+    You can also pass common `--accounts-file` argument with a path to (existing or not existing) file where you want to save account info.
+    
+    For a detailed CLI description, see [account create command reference](../appendix/cast/account/create.md).
+
+
+- prefund generated address with tokens
+  
+    You can do it both by sending tokens from another starknet account or by bridging them with StarkGate.
```
THenry14, line 29: No-one decided, so let's - I think this might not be a bad idea to link to StarkGate. :+1:

docs/src/appendix/cast/account/create.md
war-in: Yes it would and I think it also should

THenry14: I see. technically, the same account would be available through any rpc url as long as it's deployed right? But I guess it maybe makes sense to add this url as well, since users probab;y won't switch between urls that much

THenry14 (approved):

docs/src/appendix/cast/account/create.md
war-in: Yup, you're right

Radinyn (approved):

PR #296, opened 2023-07-26
[CLI] `mulicall run` as single tx

### LINKED ISSUE #121: Make multicall a single transaction
As discussed on cast channel on slack, multicall can be done as a single tx. We should do that, after initial multicall support is merged.

war-in:
Closes #121 

## Introduced changes

- multicall is the`real` multicall now
- few functions were extracted to improve codebase

## Checklist

- [x] Linked relevant issue
- [ ] Added relevant tests


starknet-foundry/crates/cast/tests/e2e/multicall/run.rs
```diff
@@ -109,10 +108,7 @@ async fn test_deploy_success_invoke_fails() {
     let bdg = snapbox.assert();
     let out = bdg.get_output();
 
-    let stdout_str =
-        std::str::from_utf8(&out.stdout).expect("failed to convert command output to string");
     let stderr_str =
-        std::str::from_utf8(&out.stderr).expect("failed to convert command output to string");
-    assert!(stdout_str.contains("command: Deploy"));
+        dbg!(std::str::from_utf8(&out.stderr).expect("failed to convert command output to string"));
```
THenry14, line 112: `dbg` ? 👀

THenry14 (commented):
Looks very good 🔥 
We might want to adjust docs as well (changed interface, include info that multicall is in one tx)

docs/src/starknet/multicall.md
```diff
@@ -26,21 +29,21 @@ unique = false
...
-contract_address = "0x38b7b9507ccf73d79cb42c2cc4e58cf3af1248f342112879bfdf5aa4f606cc9"
+contract_address = "map_contract"
 function = "put"
 inputs = ["0x123", "234"]
 ```
 
 After running `sncast multicall run --path file.toml`, a declared contract will be first deployed, and then its function `put` will be invoked.
 
+> 📝 **Note**
+> You can add property `id` to the deploy call specification and then use it as a `contract_address` when declaring an invoke call 🔥
```
THenry14, line 40: Maybe something like "In the example above, we have used `id` property in `deploy` call (...)` ?

starknet-foundry/crates/cast/src/starknet_commands/multicall/run.rs
```diff
@@ -104,5 +138,5 @@ pub async fn run(
         }
     }
 
-    Ok(())
+    execute_calls(account, parsed_calls, max_fee).await
```
THenry14, line 141: Is it possible to await when calling `run`? I think it would make more sense

war-in: Do you mean returning future here?

THenry14: yup. if you think it would make sense.

war-in: In the next PR I've unified all commands to return `Result<Vec<FieldElement>>` and I would like to keep it that way

THenry14 (approved):

Radinyn (approved):

PR #300, opened 2023-07-26
Add invoke/call tests

piotmag769:


starknet-foundry/crates/forge/tests/data/test_files/test_proxy_storage.cairo
```diff
@@ -0,0 +1,59 @@
...
+    let caller_address = deploy_contract('Caller');
+    let executor_address = deploy_contract('Executor');
+
+    let caller_dispatcher = ICallerSafeDispatcher { contract_address: caller_address };
+    let executor_dispatcher = IExecutorSafeDispatcher { contract_address: executor_address };
+
+    let ns = NestedStruct { d: 6 };
+    let cs = CustomStruct { a: 2, b: 3, c: ns };
+
+    let result = caller_dispatcher.call_executor(executor_address, cs).unwrap();
```
Arcticae, line 52: No point of calling safe dispatcher if you're unwrapping afterwards

starknet-foundry/crates/forge/tests/integration/dispatchers.rs
Arcticae, line 1: It seems a bit weird to me that the test is containing the contract while not containing the actual test code. The more intuitive would be to do it the other way around, so that the contracts can be reused between tests.

Arcticae (approved):

starknet-foundry/crates/forge/tests/integration/dispatchers.rs
piotmag769: I don't think those contracts are particularly useful tbh, that's why I did it this way. On the other hand, the test code is quite long, so it makes sense to put it in a separate file

starknet-foundry/crates/forge/tests/integration/common/runner.rs
```diff
@@ -80,11 +80,18 @@ impl<'a> TestCase {
...
+            test_code_or_path.to_string()
+        };
+
         let dir = TempDir::new()?;
         let test_file = dir.child(Self::TEST_PATH);
         test_file.touch()?;
-        test_file.write_str(test_code)?;
+        test_file.write_str(&test_code)?;
 
         dir.child("src/lib.cairo").touch().unwrap();
```
cptartur, line 96: I'd rather we have one consistent approach for all integration tests, otherwise it will get hard to maintain. Lets keep the previous way of having test code in the tests themselves. Otherwise all test files will just be `test_case!(path_x)` and will soon have great amount of files in data that are not obviously connected with anything.

cptartur (changes requested):

starknet-foundry/crates/forge/tests/integration/common/runner.rs
piotmag769: Sure, if you insist

cptartur (approved):

PR #306, opened 2023-07-27
[CLI] Print command name when error

### LINKED ISSUE #270: Print command name when error occurs
Currently, we show that an error occurred but not which command did that

Now:
```
error: xxx
```

We should also print the command name
Then:
```
command: declare
error: xxx
```

war-in:
Closes #270 

## Introduced changes

- commands return their name when an error occurs
- refactored printing output

## Checklist

- [ ] Linked relevant issue
- [ ] Added relevant tests


RELEASING.md
```diff
@@ -0,0 +1,25 @@
+# Instruction For Creating New Starknet Forge Releases
```
THenry14, line 1: perhaps some rebase is needed? :D

PR #311, opened 2023-07-27
[CLI] Print command name on error

### LINKED ISSUE #270: Print command name when error occurs

war-in:
Closes #270

## Introduced changes

- commands return their name when an error occurs
- refactored printing output

## Checklist

- [ ] Linked relevant issue
- [ ] Added relevant tests

### DIFF
--- a/starknet-foundry/crates/cast/src/main.rs
+++ b/starknet-foundry/crates/cast/src/main.rs
@@ -1,5 +1,5 @@
 use crate::helpers::scarb_utils::parse_scarb_config;
-use crate::starknet_commands::account::{create::print_account_create_result, Account};
+use crate::starknet_commands::account::Account;
 use crate::starknet_commands::{
     account, call::Call, declare::Declare, deploy::Deploy, invoke::Invoke, multicall::Multicall,
 };
@@ -100,14 +100,14 @@ async fn main() -> Result<()> {
             account_file_exists(&accounts_file_path)?;
             let mut account = get_account(&account, &accounts_file_path, &provider, &network)?;
 
-            let result = starknet_commands::declare::declare(
+            let mut result = starknet_commands::declare::declare(
                 &declare.contract,
                 declare.max_fee,
                 &mut account,
             )
             .await;
 
-            print_command_result("declare", result, cli.int_format, cli.json)?;
+            print_command_result("declare", &mut result, cli.int_format, cli.json)?;
 
             Ok(())
         }
@@ -115,7 +115,7 @@ async fn main() -> Result<()> {
             account_file_exists(&accounts_file_path)?;
             let account = get_account(&account, &accounts_file_path, &provider, &network)?;
 
-            let result = starknet_commands::deploy::deploy(
+            let mut result = starknet_commands::deploy::deploy(
                 deploy.class_hash,
                 deploy.constructor_calldata,
                 deploy.salt,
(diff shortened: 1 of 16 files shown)


starknet-foundry/crates/cast/src/main.rs
```diff
@@ -204,31 +184,27 @@ async fn main() -> Result<()> {
                         new.output_path.clone(),
                         new.overwrite.unwrap_or(false),
                     )?;
-                    starknet_commands::multicall::new::print_new_result(result.as_str());
+                    println!("{result}");
```
THenry14, line 187: should this be here?

starknet-foundry/crates/cast/tests/helpers/fixtures.rs
```diff
@@ -135,7 +135,7 @@ pub async fn get_transaction_receipt(tx_hash: FieldElement) -> TransactionReceip
             "id": 0,
         }
     );
-    let resp: Value = serde_json::from_str(
+    let resp: Value = dbg!(serde_json::from_str(
```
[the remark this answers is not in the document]
war-in: ...

starknet-foundry/crates/cast/src/main.rs
war-in: Yes, definitely. The previous function did the same. I think there is no reason to keep such a function

THenry14: Aaah, yes I understand now. Good. :+1:

starknet-foundry/crates/cast/src/starknet_commands/invoke.rs
```diff
@@ -73,7 +50,7 @@ pub async fn execute_calls(
     account: &SingleOwnerAccount<&JsonRpcClient<HttpTransport>, LocalWallet>,
     calls: Vec<Call>,
     max_fee: Option<FieldElement>,
-) -> Result<FieldElement> {
+) -> Result<Vec<(&'static str, String)>> {
```
Arcticae, line 53: Won't returning strings everywhere make it harder to reuse this later in deployment scripts?

piotmag769 (approved):
Looks good but did not check the business logic

starknet-foundry/crates/cast/src/starknet_commands/invoke.rs
war-in: Unfortunately, I don't know :/ But we are returning name and value, so we can make FieldElements from Strings then. If there are problems with those returning values we will think of another solution.

war-in: Changed to the structs 🔥

starknet-foundry/crates/cast/src/starknet_commands/account/create.rs
```diff
@@ -123,7 +124,15 @@ pub async fn create(
...
     }
 
-    Ok(output)
+    Ok(AccountCreateResponse {
+        address,
+        max_fee: FieldElement::from(max_fee),
+        add_profile: if add_profile {
+            "Profile successfully added to Scarb.toml".to_string()
+        } else {
+            "Flag was not set. No profile added to Scarb.toml".to_string()
```
THenry14: Should we specify what flag it is?

starknet-foundry/crates/cast/src/starknet_commands/response_structs.rs
```diff
@@ -0,0 +1,31 @@
+use serde::Serialize;
```
THenry14: Let's maybe move this file to helpers?

Issue #244, opened 2023-07-21
Use `path_to_scarb_toml` when running `scarb build` in declare


war-in:
Currently, we are using the `path_to_scarb_toml` argument to find profile settings. We should also use it inside `sncast declare` command.

PR #340, opened 2023-07-31
Szymczyk/remove scarb lib

THenry14:
Closes #https://github.com/foundry-rs/starknet-foundry/issues/244

## Introduced changes

- Remove scarb lib
- Use found manifest path in declare

## Breaking changes

None

## Checklist

- [X] Linked relevant issue
- [X] Added relevant tests


crates/cast/src/helpers/scarb_utils.rs
```diff
@@ -38,11 +39,31 @@ pub fn get_property(
...
+        || {
+            let output = Command::new("scarb")
+                .arg("manifest-path")
+                .stdout(Stdio::piped())
+                .output()
+                .expect("Failed to execute scarb manifest-path command");
+            let output_str = String::from_utf8(output.stdout)
+                .expect("Invalid output of scarb manifest-path command");
+            let path = Utf8PathBuf::from_str(output_str.trim())
+                .expect("Scarb manifest-path returned invalid path");
```
war-in, line 55: We would like to improve the errors and I wonder if we should panic here or print something like

```
command: ...
error: Failed to execute scarb manifest-path command
```

crates/cast/tests/e2e/declare.rs
```diff
@@ -118,10 +118,13 @@ fn scarb_build_fails(contract_path: &str, accounts_file_path: &str) {
...
         .args(args);
+    let assert = snapbox.assert().success();
+    let stderr_output = String::from_utf8(assert.get_output().stderr.clone()).unwrap();
 
-    snapbox.assert().success().stderr_matches(indoc! {r#"
-        error: scarb build returned non-zero exit code: 1
-    "#});
+    assert!(
+        stderr_output.contains("error: Scarb build returned non-zero exit code: 1"),
+        "Expected error message not found in stderr: {stderr_output}",
```
war-in, line 126: Should we start adding messages to all the `assert` statements?

crates/cast/src/helpers/scarb_utils.rs
war-in: You can improve added `expects` here and we should also create an issue for the rest

crates/cast/tests/e2e/declare.rs
THenry14: I found it very handy for this test, as by default failed test result did not provide much information:

```
---- e2e::declare::scarb_build_fails::when_cairo_2_and_scarb_toml_does_not_exist stdout ----
thread 'e2e::declare::scarb_build_fails::when_cairo_2_and_scarb_toml_does_not_exist' panicked at 'assertion failed: stderr_output.contains(\"error: Scarb build returned non-zero exit codee: 1\")', crates/cast/tests/e2e/declare.rs:124:5
```

Sometimes for asserts there is expected vs returned comparison, so in such cases it is not necessary. But, in scenarios like this one I'd say yes - we should add more context to errors.

crates/cast/src/helpers/scarb_utils.rs
THenry14: Changed in [commit](https://github.com/foundry-rs/starknet-foundry/pull/340/commits/9a6634311907e0812c7910151045d3c5b3ca4fa1). Created an issue https://github.com/foundry-rs/starknet-foundry/issues/352

crates/cast/tests/e2e/declare.rs
war-in: Could you please create an issue then? 🙏 

Comparison between expected and actual output is provided if we use `snapbox.assert().stderr_matches`

THenry14: Sure, https://github.com/foundry-rs/starknet-foundry/issues/353 :+1:

war-in (approved):

crates/cast/src/helpers/scarb_utils.rs
```diff
@@ -38,11 +39,34 @@ pub fn get_property(
...
+
+    Ok(path)
+}
+
 pub fn parse_scarb_config(
     profile: &Option<String>,
     path: &Option<Utf8PathBuf>,
 ) -> Result<CastConfig> {
-    let found_path = find_manifest_path(path.as_ref().map(Utf8PathBuf::as_path));
+    let found_path = get_scarb_manifest(path);
```
maciektr, line 69:
```suggestion
    let found_path = path.unwrap_or_else(|| get_scarb_manifest());
```
Seems more explicit to me, as you are not using the path in `get_scarb_manifest` anyway

crates/cast/src/starknet_commands/declare.rs
```diff
@@ -72,14 +73,18 @@ pub async fn declare(
...
+            result_msg
+        );
     }
 
     // TODO #41 improve handling starknet artifacts
-    let compiled_directory = find_manifest_path(None)
-        .expect("Failed to obtain Scarb.toml file path")
+    let compiled_directory = &manifest_path
         .parent()
         .map(|parent| parent.join("target/release"))
```
maciektr, line 89: This should read target dir from`scarb metadata` instead.

crates/cast/src/helpers/scarb_utils.rs
THenry14: Makes sense, but I also like the way `get_scarb_manifest` checks for scarb existence in PATH (and it is used everywhere where scarb is required). 

By applying you suggestion, I would have to copy this part into multiple places in my code, because it wouldn't be invoked when path to scarb.toml is supplied, right? Also, this is kind of inline with scarb's own `get_manifest_path` where you can override the path, so I thought this is ok to do something similar.

crates/cast/src/starknet_commands/declare.rs
```diff
@@ -57,12 +56,14 @@ pub async fn declare(
...
     max_fee: Option<FieldElement>,
     account: &mut SingleOwnerAccount<&JsonRpcClient<HttpTransport>, LocalWallet>,
+    path_to_scarb_toml: &Option<Utf8PathBuf>,
 ) -> Result<DeclareResponse> {
     let contract_name: String = contract_name.to_string();
-    which::which("scarb")
-        .context("Cannot find `scarb` binary in PATH. Make sure you have Scarb installed https://github.com/software-mansion/scarb")?;
+    let manifest_path =
+        get_scarb_manifest(path_to_scarb_toml).expect("Failed to obtain scarb manifest file path");
     let command_result = Command::new("scarb")
```
maciektr, line 64: vide supra
```suggestion
        let mut cmd = Command::new("scarb");
        if let Some(manifest_path) = path_to_scarb_toml {
            cmd.arg("--manifest-path").arg(manifest_path);
        }
        let command_result = cmd.arg("--release").(...)
```

crates/cast/src/starknet_commands/declare.rs
THenry14: I suppose it should:

1. Run Command::new("scarb").arg("metadata") (...)
2. get stdout out of it
3. transform to String
4. Transform to serialize to json
5. get target_dir
6. append "/release" to it

?

Or is there any other, simpler way? tried to check scarb_metadata, but couldn't find the thing I needed.

crates/cast/src/helpers/scarb_utils.rs
maciektr: > Makes sense, but I also like the way get_scarb_manifest checks for scarb existence in PATH (and it is used everywhere where scarb is required).

I suggest not calling `scarb manifest-path` each time we need to use Scarb, as it often can be redundant. Which makes uncoupling it from ensuring Scarb is installed sensible for me anyway.

> Also, this is kind of inline with scarb's own get_manifest_path where you can override the path, so I thought this is ok to do something similar.

It's probably ok :D
 However, contrary to this function, `find_manifest_path` from Scarb does actually use the variable passed to it. If the variable is not used, I see passing it to the function as unnecessary cognitive complexity.

THenry14: > I suggest not calling scarb manifest-path each time we need to use Scarb, as it often can be redundant. Which makes uncoupling it from ensuring Scarb is installed sensible for me anyway.

Well, if path is supplied it is not called at all. If it's not supplied it is called at most 2 times, and I agree that in this scenario this could be scaled down to 1. I'll try to make that happend in this PR.

> It's probably ok :D
However, contrary to this function, find_manifest_path from Scarb does actually use the variable passed to it. If the variable is not used, I see passing it to the function as unnecessary cognitive complexity.

I want to understand this better, so sorry for silly question - my understanding is that find_manifest_path from Scarb checks for path in cwd or parent dirs if it is supplied with 'None' argument; otherwise it just returns the supplied path, right? So I suppose it does nothing more with the supplied path than get_scarb_manifest here in cast - when supplied with None it searches for path, otherwise it returns it. Or am I missing something?

Thanks for all the tips though! :)

maciektr:
```
        Some(user_override) => Ok(fs::canonicalize(user_override)
            .unwrap_or_else(|_| user_override.into())
            .try_into_utf8()?),
```
Not much, but it is canonicalized and parsed there

crates/cast/src/starknet_commands/declare.rs
THenry14: Changed the way get_scarb_manifest works, so right now it is either a user defined path or found one using scarb, as per https://github.com/foundry-rs/starknet-foundry/pull/340#discussion_r1279390005.

Do you think it can be left as is? Either way the path exists, and my understanding is that scarb just won't look for it for the second time here, but will take the path directly right?

maciektr: Yes, if one present Scarb will used passed path verbatim.

maciektr (approved):

Issue #316, opened 2023-07-28
Fix error handling on user input


THenry14:
It looks like we panic in some places in code, where we depend on user input. we should not do that.
The issue came out when someone set `casm=false` in scarb.toml. We should also allow for this to happen.

PR #350, opened 2023-07-31
Require sierra and casm compilation artifacts

THenry14:
Closes https://github.com/foundry-rs/starknet-foundry/issues/316

## Introduced changes

- Print error when sierra or casm are not enabled for contracts 

## Breaking changes

None

## Checklist

- [X] Linked relevant issue
- [X] Added relevant tests


crates/cast/src/starknet_commands/declare.rs
```diff
@@ -130,29 +130,28 @@ pub async fn declare(
...
             if contract.contract_name == contract_name {
-                return Some(contract.artifacts.sierra.clone());
+                return contract.artifacts.sierra.clone();
             }
             None
         })
-        .unwrap_or_else(|| {
-            panic!("Failed to find contract {contract_name} in starknet_artifacts.json")
+        .ok_or_else(|| -> anyhow::Result<Utf8PathBuf> {
+            anyhow::bail!("Failed to find contract {contract_name} in starknet_artifacts.json")
```
war-in, line 138: Can't we just 
```suggestion
        .ok_or_else(|| -> {
            anyhow::bail!("Failed to find contract {contract_name} in starknet_artifacts.json")
```
?

crates/cast/src/starknet_commands/declare.rs
```diff
@@ -130,29 +130,28 @@ pub async fn declare(
...
         });
-    let sierra_contract_path = &compiled_directory.join(sierra_path);
+    let sierra_contract_path = &compiled_directory.join(sierra_path.expect("Cannot find sierra artifact file - please make sure sierra is set to 'true' under your [[target.starknet-contract]] field in Scarb.toml"));
```
war-in, line 140: I think we should not `expect` here, because it will panic on error. It would be nicer to just print
```
command: declare
error: Cannot find sierra
```

war-in: We can achieve this simply by mapping en error and returning `bail!(...`

crates/cast/tests/data/contracts/v1/no_casm/src/lib.cairo
```diff
@@ -0,0 +1,18 @@
...
+    fn increase_balance(amount: felt252) {
+        balance::write(balance::read() + amount);
+    }
+
+    // Returns the current balance.
+    #[view]
+    fn get_balance() -> felt252 {
+        balance::read()
+    }
+}
```
war-in, line 18: How about an empty contract?

crates/cast/tests/e2e/declare.rs
```diff
@@ -159,3 +159,34 @@ fn test_too_low_max_fee(contract_path: &str, salt: &str, account: &str) {
 
     fs::remove_dir_all(contract_path).unwrap();
 }
+
+#[test_case("/v1/no_sierra", "../../../accounts/accounts.json" ; "dupa")]
```
war-in, line 163: 🍑 👀

crates/cast/tests/e2e/declare.rs
```diff
@@ -159,3 +159,34 @@ fn test_too_low_max_fee(contract_path: &str, salt: &str, account: &str) {
...
+#[test_case("/v1/no_casm", "../../../accounts/accounts.json" ; "dupa2")]
```
war-in, line 164: 🍑 👀

crates/cast/tests/data/contracts/v1/no_casm/src/lib.cairo
THenry14: What do you mean? What happens when there is an empty contract file?

piotmag769 (commented):
Why do we need both casm and sierra, cannot we decide on one format that is required?

crates/cast/src/starknet_commands/declare.rs
war-in: Ok, so how about `map_err` here? I'm not sure if it applies here but it's shorter 😜

crates/cast/tests/data/contracts/v1/no_casm/src/lib.cairo
war-in: I mean there is something like an empty contract in the `cairo` repo
https://github.com/starkware-libs/cairo/blob/main/crates/cairo-lang-starknet/test_data/minimal_contract.cairo

and I wonder if we can't use it there

crates/cast/src/starknet_commands/declare.rs
THenry14: as discussed, will leave as is for now (to be improved in https://github.com/foundry-rs/starknet-foundry/issues/352)

crates/cast/src/starknet_commands/declare.rs
THenry14: as discussed, will leave as is for now (to be improved in https://github.com/foundry-rs/starknet-foundry/issues/352)

crates/cast/tests/data/contracts/v2/no_casm/Scarb.toml
```diff
@@ -0,0 +1,18 @@
...
+sierra = true
+casm = false
+
+[lib]
+sierra = false
+
+[tool.sncast]
+url = "http://127.0.0.1:5055/rpc"
+network = "testnet"
+account = "user2"
```
war-in, line 18: Nitpick, but it probably could be removed

crates/cast/tests/data/contracts/v2/no_sierra/Scarb.toml
```diff
@@ -0,0 +1,18 @@
...
+sierra = false
+casm = true
+
+[lib]
+sierra = false
+
+[tool.sncast]
+url = "http://127.0.0.1:5055/rpc"
+network = "testnet"
+account = "user2"
```
war-in, line 18: Same here

war-in (approved):

PR #354, opened 2023-07-31
[CLI] Fix wait_for_tx

### LINKED ISSUE #318: wait_for_tx does not wait long enough?
from tg:

Fico, [28 Jul 2023 at 10:24:44]:
I've try to declare and deploy the template contract https://github.com/foundry-rs/starknet_forge_template.git, every step returns an error, but it deploys successfully lol

Image

war-in:
Closes #318 

## Introduced changes

- wait for tx listens for a maximum of one minute now and return an error after this time
- added `--wait` flag (we do not wait by default)

## Breaking changes

- users have to pass `--wait` flag to wait for tx

## Checklist

- [ ] Linked relevant issue
- [ ] Added relevant tests


crates/cast/src/lib.rs
```diff
@@ -163,40 +163,50 @@ pub fn get_network(name: &str) -> Result<Network> {
...
-                sleep(Duration::from_secs(5));
-                true
-            }
-            TransactionStatus::AcceptedOnL2 | TransactionStatus::AcceptedOnL1 => {
-                return Ok("Transaction accepted")
+    println!("Transaction hash: {tx_hash:#x}");
+
+    let mut retries = retries;
+    'a: loop {
+        sleep(Duration::from_secs(5));
```
THenry14, line 172: It could perhaps make sense to try and poll for tx status at least once before sleep

crates/cast/src/lib.rs
```diff
@@ -163,40 +163,50 @@ pub fn get_network(name: &str) -> Result<Network> {
...
             }
-            TransactionStatus::Rejected => {
-                return Err(anyhow!("Transaction has been rejected"));
+            Err(ProviderError::StarknetError(StarknetError::TransactionHashNotFound)) => {
+                if retries > 0 {
+                    retries -= 1;
+                    println!("Waiting for transaction to be received. Retries left: {retries}");
+                } else {
+                    bail!("Could not get transaction with hash: {tx_hash:#x}. Transaction rejected or not received.")
+                }
```
THenry14, line 205: I wonder if the syntax:

```
for i in 0..10 {
    // Your code here for each iteration of i (from 0 to 9)
}
```

could simplify things a bit?

crates/cast/tests/e2e/account/deploy.rs
```diff
@@ -141,6 +141,7 @@ async fn test_too_low_max_fee() {
         NETWORK,
         "--accounts-file",
         accounts_file,
+        "--wait",
```
THenry14, line 144: just out of curiousity - does that slow down tests in some significant way? I guess ~5 seconds (since sleep if first in the function)?

crates/cast/src/lib.rs
war-in: I think that would make sense in the case of devnet (and maybe Starknet) in some time but for now all the transactions I've tried are available after (at least) 10 seconds.
However, I think I can change that to be future-proof

crates/cast/tests/e2e/account/deploy.rs
war-in: Yup, they were slowed down but thanks to your comment they are not 🔥

crates/cast/src/helpers/constants.rs
```diff
@@ -15,3 +15,6 @@ inputs = []
 pub const UDC_ADDRESS: &str = "0x041a78e741e5af2fec34b695679bc6891742439f7afb8484ecd7766661ad02bf";
 pub const OZ_CLASS_HASH: &str =
     "0x058d97f7d76e78f44905cc30cb65b91ea49a4b908a76703c54197bca90f81773";
+// used in wait_for_tx. Txs are fetched every 5 seconds
+#[allow(dead_code)]
+pub const DEFAULT_RETRIES: u8 = 12;
```
war-in, line 20: That was the only way I found to resolve the clippy's warning. Please add your suggestions if you have some 🙏 

```
warning: constant `DEFAULT_RETRIES` is never used
  --> crates/cast/src/helpers/constants.rs:19:11
   |
19 | pub const DEFAULT_RETRIES: u8 = 12;
   |           ^^^^^^^^^^^^^^^
   |
   = note: `#[warn(dead_code)]` on by default
```

THenry14 (approved):
lgtm

PR #384, opened 2023-08-02
Export cheatcodes as scarb package

piotmag769:
## Breaking changes

Cheatcodes are now a separate package that can be linked as dependency
Corelib copy is removed

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests

Will update docs in separate PR https://github.com/foundry-rs/starknet-foundry/issues/402
Task to update git dependency test after release https://github.com/foundry-rs/starknet-foundry/issues/403

Did not add more tests using git dependency (almost all of them are using path dependency) due to nondeterministic tests behaviour (that's due to testing on multiple threads that try to fetch from network I guess)


war-in (commented):
I think it's ok but there's a chance I don't have so much RAM in my brain to check it correctly.

crates/cheatable-starknet/Cargo.toml
```diff
@@ -26,5 +26,8 @@ starknet = "0.4.0"
 regex = "1.9.1"
 assert_fs = "1.0.13"
 
+[dev-dependencies]
+indoc.workspace = true
```
cptartur, line 30: Why was this added?

crates/forge/src/scarb.rs
```diff
@@ -165,20 +165,59 @@ pub fn dependencies_for_package(
...
     use super::*;
     use assert_fs::fixture::{FileTouch, FileWriteStr, PathChild, PathCopy};
-    use indoc::indoc;
+    use assert_fs::TempDir;
+    use indoc::{formatdoc, indoc};
     use scarb_metadata::MetadataCommand;
     use std::process::Command;
+    use std::str::FromStr;
+
+    pub fn setup_package(package_name: &str) -> TempDir {
```
cptartur, line 174: I'm assuming this is only visible inside `tests` module anyways, but does this need to be public?

crates/forge/tests/integration/common/corelib.rs
```diff
@@ -1,10 +1,40 @@
...
+        .current_dir(dir.to_path_buf())
+        .inherit_stderr()
+        .exec()
+        .unwrap();
+    let package = &scarb_metadata.workspace.members[0];
+    let (_, _, corelib_path, _, _) =
+        extract_metadata_from_package(&scarb_metadata, package).unwrap();
+
+    corelib_path
 }
```
cptartur, line 38: Let's create an issue to investigate a less hacky way of doing it. Perhaps scarb(-metadata) could implement some feature that would allow requesting for a corelib path. Then we could just use it directly.

But it's okay for now.

crates/cheatable-starknet/Cargo.toml
piotmag769: Needed to use indoc in cheatable-starknet so I decided to hoist this dep here

cptartur (approved):

crates/forge/tests/integration/common/corelib.rs
piotmag769: https://github.com/foundry-rs/starknet-foundry/issues/414

PR #181, opened 2023-07-14
start roll cheatcode

MaksymilianDemitraszek:

PR #385, opened 2023-08-02
Add `mock_call`

abulenok:
Closes #

## Introduced changes

- Add `start_mock_call` cheatcode (with tests and docs)
- Add `stop_mock_call` cheatcode (with tests and docs)

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


docs/src/appendix/cheatcodes/start_mock_call.md
```diff
@@ -0,0 +1,51 @@
+# `start_mock_call`
+
+> `fn start_mock_call(contract_address: ContractAddress, fn_name: felt252, ret_data: Array::<felt252>)`
+
+Mocks external function `fn_name` for a contract at the given address. The function `fn_name` will return data provided in `ret_data` argument.
+This change can be canceled with [`stop_mock_call`](./stop_mock_call.md).
+
+- `contract_address` - target contract address
+- `fn_name` - name of the function in contract `contract_address` that will be mocked
```
war-in, line 9:
```suggestion
- `fn_name` - name of the function in a contract at the `contract_address` that will be mocked
```

crates/cheatable-starknet/src/cheatcodes/mock_call.rs
```diff
@@ -0,0 +1,39 @@
+use crate::{cheatcodes::EnhancedHintError, CheatedState};
+use starknet_api::core::{ContractAddress, EntryPointSelector};
+use starknet_api::hash::StarkFelt;
+use std::collections::HashMap;
+
+impl CheatedState {
+    pub fn start_mock_call(
```
war-in, line 7: I can imagine such a situation - someone has a contract with only one `get_balance` function and mocks the `set_balance` function which originally it does not have. In the current implementation, it would pass but maybe we should be panicking in such a case.

It can be considered as a bug and as a feature. As a feature because someone could create a `virtual` contract which does not exist without mocking.

corelib/src/cheatcodes.cairo
```diff
@@ -118,3 +118,27 @@ fn stop_warp(contract_address: ContractAddress) {
     cheatcode::<'stop_warp'>(array![contract_address_felt].span());
 }
 
+fn start_mock_call(contract_address: ContractAddress, fn_name: felt252, ret_data: Array::<felt252>) {
```
war-in, line 121:
```suggestion
fn start_mock_call<T>(contract_address: ContractAddress, fn_name: felt252, ret_data: T) {
```

I'm not sure if something like that is possible but definitely would be helpful

crates/forge/tests/integration/mock_call.rs
```diff
@@ -0,0 +1,424 @@
...
+    let result = run(
+        &test.path().unwrap(),
+        &test.path().unwrap().join("src/lib.cairo"),
+        &Some(test.linked_libraries()),
+        &Default::default(),
+        Some(&Utf8PathBuf::from_path_buf(corelib().to_path_buf()).unwrap()),
+        &test.contracts(corelib().path()).unwrap(),
+        &Utf8PathBuf::from_path_buf(predeployed_contracts().to_path_buf()).unwrap(),
+    )
+    .unwrap();
```
war-in, line 79: I think we could create some helper function for this one because it is used in every test

crates/forge/tests/integration/mock_call.rs
```diff
@@ -0,0 +1,424 @@
...
+
+            let dispatcher = IMockCheckerDispatcher { contract_address };
+
+            let mut mock_ret_data = ArrayTrait::new();
+            mock_ret_data.append(421);
+            start_mock_call(contract_address, 'get_thing', mock_ret_data);
+
+            let thing = dispatcher.get_thing_wrapper();
+
+            assert(thing == 421, 'Incorrect thing'); // TODO: should this work?
```
war-in, line 278: Yup, I think it should

corelib/src/cheatcodes.cairo
```diff
@@ -118,3 +118,27 @@ fn stop_warp(contract_address: ContractAddress) {
...
+        i += 1;
+    };
+
+    cheatcode::<'start_mock_call'>(inputs.span());
+}
+
+fn stop_mock_call(contract_address: ContractAddress, fn_name: felt252) {
+    let contract_address_felt: felt252 = contract_address.into();
+    cheatcode::<'stop_mock_call'>(array![contract_address_felt, fn_name].span());
+}
```
cptartur, line 144: Newline missing

corelib/src/cheatcodes.cairo
cptartur, line 121: I think we could use nicer sounding names in the user facing interface e.g. `function_name` instead of `fn_name`

crates/cheatable-starknet/src/cheatcodes/mock_call.rs
cptartur: I think mocking methods that do not exist have some benefit though I agree it kind of depends on what we want `mock_call` to do exactly.

crates/cheatable-starknet/src/cheatcodes/mock_call.rs
```diff
@@ -0,0 +1,39 @@
...
+        ret_data: Vec<StarkFelt>,
+    ) -> Result<(), EnhancedHintError> {
+        if let std::collections::hash_map::Entry::Vacant(e) =
+            self.mocked_functions.entry(contract_address)
+        {
+            e.insert(HashMap::from([(fn_name, ret_data)]));
+        } else {
+            let contract_mocked_fns = self.mocked_functions.get_mut(&contract_address).unwrap();
+            contract_mocked_fns.insert(fn_name, ret_data);
+        }
```
cptartur, line 20: I think this could be make less verbose with this method https://doc.rust-lang.org/stable/std/collections/hash_map/enum.Entry.html#method.or_insert_with

crates/cheatable-starknet/src/lib.rs
```diff
@@ -1,5 +1,8 @@
+extern crate core;
```
cptartur, line 1: What's the purpose of this?

crates/cheatable-starknet/src/rpc.rs
```diff
@@ -681,6 +685,30 @@ pub fn execute_inner_call(
...
+        if cheated_state
+            .mocked_functions
+            .contains_key(&contract_address)
+        {
+            let contract_functions = cheated_state
+                .mocked_functions
+                .get(&contract_address)
+                .unwrap_or_else(|| {
+                    panic!("ContractAddress {contract_address:?} not found in mocked functions map")
+                });
```
cptartur, line 702: I think this could be refactored so we do not have to first check if map contains something and then get with unwrap. 

Something like `if let Some(contract_functions) = cheated_state.mocked_functions.get(&contract_address)` or a `map`.

crates/forge/src/cheatcodes_hint_processor.rs
```diff
@@ -218,7 +221,37 @@ impl CairoHintProcessor<'_> {
...
+                let contract_address = ContractAddress(PatriciaKey::try_from(StarkFelt::new(
+                    inputs[0].clone().to_be_bytes(),
+                )?)?);
+                let fn_name = inputs[1].clone();
+                let fn_name = as_cairo_short_string(&fn_name)
+                    .unwrap_or_else(|| panic!("Failed to convert {fn_name:?} to Cairo short str"));
+                let fn_name = selector_from_name(fn_name.as_str());
+
+                self.cheated_state.stop_mock_call(contract_address, fn_name)
+            }
```
cptartur, line 254: Other cheatcodes are now extracted to separate methods, this should be as well

crates/forge/tests/data/contracts/mock_checker.cairo
```diff
@@ -0,0 +1,35 @@
...
+
+        fn get_thing_wrapper(self: @ContractState) -> felt252 {
+            self.get_thing()
+        }
+
+        fn get_other_thing(self: @ContractState) -> felt252 {
+            13
+        }
+    }
+}
```
cptartur, line 35: Newline

crates/forge/tests/integration/mock_call.rs
cptartur: Can you create an issue for that?

crates/forge/tests/integration/mock_call.rs
cptartur, line 1: What should happen in case where we mock a method that returns a value (like in your tests), then make some invokes that change the value that would be returned if the method wasn't mocked and then stop a mock. Should the invokes be reflected? 

something like
```cairo
let mut mock_ret_data = ArrayTrait::new();
mock_ret_data.append(421);
start_mock_call(contract_address, 'get_thing', mock_ret_data);

dispatcher.change_think(200);

let thing = dispatcher.get_thing();
assert(thing == 421, '');

stop_mock_call(...);
assert(thing == 200, ''); <----- 
```

war-in: I think they should be reflected. We are mocking only `get_thing` function

crates/cheatable-starknet/src/cheatcodes/mock_call.rs
MaksymilianDemitraszek: imo it should allow mocking whatever user wants because you don't know what it mocks exactly as there may be no contract under mocked address.

war-in: Ok, so maybe we should include that in the docs?

abulenok: @war-in what exactly do you suggest to include in docs? The fact that it would not panic when the user provides an invalid contract address or fn_name?

war-in: I would document the fact that users can mock any function at any address even if there are no contract under it

abulenok: Yeah, but we can, for example, `prank` contract_address which doesn't exist, same goes for `roll` and `warp`, and there is no info about it in docs AFAIK. So, I think `mock_call` behavior and docs are consistent with other cheatcodes.

abulenok: We can still consider adding a note about the possibility of applying a cheatcode to a contract/function that doesn't exist

piotmag769: Agree with @abulenok here, we should include it everywhere or nowhere

crates/cheatable-starknet/src/cheatcodes/mock_call.rs
piotmag769: +1

crates/cheatable-starknet/src/cheatcodes/mock_call.rs
war-in: Ok, nvm. I just think this is coll stuff and would be nice to have it mentioned somewhere. It's up to you if you want to create some general note about such a possibility.

crates/cheatable-starknet/src/cheatcodes/mock_call.rs
```diff
@@ -0,0 +1,35 @@
...
+use std::collections::HashMap;
+
+impl CheatedState {
+    pub fn start_mock_call(
+        &mut self,
+        contract_address: ContractAddress,
+        function_name: EntryPointSelector,
+        ret_data: Vec<StarkFelt>,
+    ) -> Result<(), EnhancedHintError> {
+        self.mocked_functions.entry(contract_address).or_insert_with(|| HashMap::new());
```
piotmag769, line 13: Doesn't it return `contract_mocked_functions` already? This way we could get rid of the unwrap below

crates/cheatable-starknet/src/cheatcodes/mock_call.rs
```diff
@@ -0,0 +1,35 @@
...
+
+        let contract_mocked_fns = self.mocked_functions.get_mut(&contract_address).unwrap();
```
piotmag769, line 15: Would use `contract_mocked_functions`, I had to think for a while what `fns` mean

crates/forge/tests/integration/mock_call.rs
piotmag769, line 1: I think we should add tests:
- with library call (`mock_call` shouldn't change its behaviour)
- with other syscall (emitting event is fine)
- (optionally) in constructor and mark it with `#[ignore]` since it won't work yet afaik

crates/forge/tests/integration/mock_call.rs
```diff
@@ -0,0 +1,424 @@
...
+            calldata.append(420);
+
+            let class_hash = declare('MockChecker');
+            let prepared = PreparedContract { class_hash: class_hash, constructor_calldata: @calldata };
+            let contract_address = deploy(prepared).unwrap();
+
+            let dispatcher = IMockCheckerDispatcher { contract_address };
+
+            let mut mock_ret_data = ArrayTrait::new();
+            mock_ret_data.append(421);
```
piotmag769, line 35: [tip] You can now use `array!` macro for convenience
```cairo
let mock_ret_data = array![421];
```

crates/forge/src/cheatcodes_hint_processor.rs
abulenok: https://github.com/foundry-rs/starknet-foundry/blob/master/crates/forge/src/cheatcodes_hint_processor.rs#L202-L220
prank, roll, and warp also extract input arguments before passing them to separate methods, so this mock_call behaviour is consistent

crates/cheatable-starknet/src/cheatcodes/mock_call.rs
MaksymilianDemitraszek: I agree with @war-in here. It would be inconsistent but it is more relevant for mock_call as it does not trigger contact execution. We can mention that mocking does not require contract being deployed.

MaksymilianDemitraszek (commented):
images

@Arcticae Can you review it before merginig?

crates/cheatable-starknet/src/lib.rs
abulenok: Removed

crates/forge/tests/integration/mock_call.rs
abulenok: by "in constructor" you mean calling mocked function in a constructor?

abulenok: Just to clarify, it will be reflected in the current implementation.

docs/src/appendix/cheatcodes/start_mock_call.md
```diff
@@ -0,0 +1,51 @@
+# `start_mock_call`
+
+> `fn start_mock_call(contract_address: ContractAddress, function_name: felt252, ret_data: Array::<felt252>)`
+
+Mocks external function `function_name` for a contract at the given address. The function `function_name` will return data provided in `ret_data` argument.
```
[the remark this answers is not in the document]
abulenok: @MaksymilianDemitraszek Maybe we should mention that only external functions can be mocked?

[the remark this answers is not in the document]
MaksymilianDemitraszek: I think we should use phrasing `Mocks contract call` instead of `Mocks function`

crates/forge/tests/integration/mock_call.rs
abulenok: added in https://github.com/foundry-rs/starknet-foundry/pull/385/commits/824fb263391bf2b528cd963322f0a6f41cdf4636

docs/src/appendix/cheatcodes/start_mock_call.md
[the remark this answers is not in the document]
abulenok: Changed in https://github.com/foundry-rs/starknet-foundry/pull/385/commits/8df3ebecb7e30ae022af1774dd025188108337f3

crates/forge/tests/integration/mock_call.rs
abulenok: added https://github.com/foundry-rs/starknet-foundry/issues/432

docs/src/appendix/cheatcodes/start_mock_call.md
```diff
@@ -0,0 +1,52 @@
+# `start_mock_call`
+
+> `fn start_mock_call(contract_address: ContractAddress, function_name: felt252, ret_data: Array::<felt252>)`
+
+Mocks contract call to a `function_name` of a contract at the given address. The function `function_name` will return data provided in `ret_data` argument.
```
MaksymilianDemitraszek, line 5:
```suggestion
Mocks contract call to a `function_name` of a contract at the given address. A call to function `function_name` will return data provided in `ret_data` argument. 
```

crates/forge/tests/integration/mock_call.rs
```diff
@@ -0,0 +1,569 @@
...
+        &corelib_path(),
+        &test.contracts(&corelib_path()).unwrap(),
+        &Utf8PathBuf::from_path_buf(predeployed_contracts().to_path_buf()).unwrap(),
+    )
+    .unwrap();
+    assert_passed!(result);
+}
+
+#[test]
+fn start_mock_call_with_syscall_has_no_effect() {
```
piotmag769, line 519: It should have an effect, it doesn't in this case since it is an inner call. Can we change it so it is not an inner call and check if it work then?

piotmag769 (commented):
Can we also add tests for:
- mocking when contract is not deployed under a given address (just pass contract address that you mocked to a dispatcher)
- mocking when contract is deployed but does not have a certain function implemented (you can define a contract trait with additional function)

piotmag769 (changes requested):
Requesting changes just to make sure the comments are addressed

crates/forge/tests/integration/mock_call.rs
abulenok: So, how do you want this to work?

crates/forge/tests/integration/mock_call.rs
```diff
@@ -0,0 +1,569 @@
...
+        &corelib_path(),
+        &test.contracts(&corelib_path()).unwrap(),
+        &Utf8PathBuf::from_path_buf(predeployed_contracts().to_path_buf()).unwrap(),
+    )
+    .unwrap();
+    assert_passed!(result);
+}
+
+#[test]
+fn start_mock_call_inner_call_has_no_effect() {
```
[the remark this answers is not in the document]
abulenok: yes

crates/forge/tests/integration/mock_call.rs
```diff
@@ -0,0 +1,569 @@
...
+        &test.contracts(&corelib_path()).unwrap(),
+        &Utf8PathBuf::from_path_buf(predeployed_contracts().to_path_buf()).unwrap(),
+    )
+    .unwrap();
+
+    assert_passed!(result);
+}
+
+#[test]
+fn start_mock_call_in_constructor_test_has_no_effect() {
```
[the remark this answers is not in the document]
abulenok: done in https://github.com/foundry-rs/starknet-foundry/pull/385/commits/f938f344267f42213c09f4b3c10f0fa2839a334c

crates/cheatable-starknet/src/cheatcodes/mock_call.rs
war-in: It would be nice to have it in the docs when this PR is merged. Maybe we can write it for `mock_call` and create an issue for the rest?

crates/forge/tests/integration/mock_call.rs
piotmag769: The same way it is described here, just with `start_mock_call`
https://github.com/foundry-rs/starknet-foundry/pull/181#discussion_r1273118750
It is important to emit an event after mocking bcs validation of block (I think? not sure what exactly is validated) is done on next library call, so we are sure this way that we increment stuff correctly

corelib/src/cheatcodes.cairo
piotmag769: Okay, I was wrong, it is definitely possible. An example from my PR

```cairo
trait TxtParser<T, impl TSerde: Serde<T>> {
    fn deserialize_txt(file: @File) -> Option<T>;
}

impl TxtParserImpl<T, impl TSerde: Serde<T>> of TxtParser<T> {
    fn deserialize_txt(file: @File) -> Option<T> {
        let mut content = cheatcode::<'parse_txt'>(array![*file.path].span());
        Serde::<T>::deserialize(ref content)
    }
}
```

crates/forge/tests/integration/mock_call.rs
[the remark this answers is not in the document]
piotmag769: Shouldn't we document this? I think this behaviour may seem unexpected if one doesn't know internals of our code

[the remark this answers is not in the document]
abulenok: Added https://github.com/foundry-rs/starknet-foundry/pull/385/files#diff-0805bdf7251fce3bf814ea4380583dfc76f836cfdcd04b58aff71c23f0e05ab9R14-R15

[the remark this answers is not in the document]
abulenok: Added https://github.com/foundry-rs/starknet-foundry/pull/385/files#diff-0805bdf7251fce3bf814ea4380583dfc76f836cfdcd04b58aff71c23f0e05ab9R14-R15

crates/forge/tests/integration/mock_call.rs
abulenok: Changed in https://github.com/foundry-rs/starknet-foundry/pull/385/commits/86479a098856b4448a3a14cb916509df46b000e2

abulenok:
@piotmag769 Added tests in https://github.com/foundry-rs/starknet-foundry/pull/385/commits/7125bde518bacd94459990c86d485656ea6d30da

crates/cheatable-starknet/src/cheatcodes/mock_call.rs
abulenok: Updated the docs

abulenok: and submitted an issue https://github.com/foundry-rs/starknet-foundry/issues/464

corelib/src/cheatcodes.cairo
abulenok: Added in https://github.com/foundry-rs/starknet-foundry/pull/385/commits/1bade0618e763faa41fd53c94a198010e10d7f3a

war-in (approved):

crates/forge/src/cheatcodes_hint_processor.rs
```diff
@@ -224,7 +224,42 @@ impl CairoHintProcessor<'_> {
...
+                });
+                let function_name = selector_from_name(function_name.as_str());
+
+                let ret_data_length = inputs[2]
+                    .to_usize()
+                    .expect("Missing ret_data len in inputs");
+                let mut ret_data = vec![];
+                for felt in inputs.iter().skip(3).take(ret_data_length) {
+                    ret_data.push(felt_to_stark_felt(&felt.clone()));
+                }
```
piotmag769, line 243:
```suggestion
                let ret_data = Vec::from(&inputs[3..(3 + ret_data_length)]);
```

docs/src/appendix/cheatcodes/start_mock_call.md
```diff
@@ -0,0 +1,55 @@
...
+
+If contract is deployed at given address, mocked function won't be executed. Address with no contract can be mocked as well.
+Mock can be canceled with [`stop_mock_call`](./stop_mock_call.md).
+
+- `contract_address` - target contract address
+- `function_name` - name of the function in a contract at the `contract_address` that will be mocked
+- `ret_data` - data to return by the function `function_name`
+
+> 📝 **Note**
+> The inner call (i.e. when a contract calls a function from within itself) cannot be mocked.
```
piotmag769, line 15:
```suggestion
> The inner call  - a call when a contract calls a function from within itself - cannot be mocked.
```

docs/src/appendix/cheatcodes/start_mock_call.md
```diff
@@ -0,0 +1,55 @@
+# `start_mock_call`
+
+> `fn start_mock_call<T, impl TSerde: serde::Serde<T>, impl TDestruct: Destruct<T>>(contract_address: ContractAddress, function_name: felt252, ret_data: T)`
+
+Mocks contract call to a `function_name` of a contract at the given address. A call to function `function_name` will return data provided in `ret_data` argument. 
+
+If contract is deployed at given address, mocked function won't be executed. Address with no contract can be mocked as well.
```
piotmag769, line 7: `If contract is deployed at given address, mocked function won't be executed`
Wdym? I don't think this sentence is particularly helpful, it confused me at first

piotmag769, line 7:
```suggestion
Mocks contract call to a `function_name` of a contract at the given address. A call to function `function_name` will return data provided in `ret_data` argument. Addresses of non-existing contracts can be mocked as well.
```

docs/src/appendix/cheatcodes/start_mock_call.md
```diff
@@ -0,0 +1,55 @@
...
+
+We can use `start_mock_call` in a test to change the data returned by `get_balance` for a given contract:
+
+```rust
+#[test]
+fn test_mock_call() {
+    // ...
+    
+    let mock_ret_data = 421;
+    start_mock_call(contract_address, 'get_balance', mock_ret_data);
```
piotmag769, line 49: outdated signature

abulenok: Why is this outdated? `start_mock_call` is called the same way in tests.

docs/src/appendix/cheatcodes/start_mock_call.md
abulenok: There could be no contract deployed under the given address. If there is, and the call to the function was mocked, then this function will not be executed.

docs/src/appendix/cheatcodes/start_mock_call.md
abulenok: I don't think this version is more readable. And the word `call` repeats too many times.

docs/src/appendix/cheatcodes/start_mock_call.md
abulenok: Slightly updated in https://github.com/foundry-rs/starknet-foundry/pull/385/commits/dc2461084f0a10b3d556f4b6f51b177251552516

docs/src/appendix/cheatcodes/start_mock_call.md
piotmag769: Agree, can we remove the `i. e.` though? I think it is the only case of an inner call, no?

crates/forge/src/cheatcodes_hint_processor.rs
abulenok: We will still need to convert it into `StarkFelt`

crates/forge/tests/integration/mock_call.rs
```diff
@@ -0,0 +1,776 @@
...
+            let class_hash = declare('MockChecker');
+            let prepared = PreparedContract { class_hash: class_hash, constructor_calldata: @calldata };
+
+            let contract_address: felt252 = 123;
+            let contract_address: ContractAddress = contract_address.try_into().unwrap();
+
+            let dispatcher = IMockCheckerDispatcher { contract_address };
+
+            let mock_ret_data = 421;
+            start_mock_call(contract_address, 'get_thing', mock_ret_data);
```
piotmag769, line 700: I think this test is not finished

crates/forge/tests/integration/mock_call.rs
```diff
@@ -0,0 +1,776 @@
...
+            let calldata = array![420];
+
+            let class_hash = declare('MockChecker');
+            let prepared = PreparedContract { class_hash: class_hash, constructor_calldata: @calldata };
+            let contract_address = deploy(prepared).unwrap();
+
+            let dispatcher = IMockCheckerDispatcher { contract_address };
+
+            let mock_ret_data = 421;
+            start_mock_call(contract_address, 'get_thing_not_implemented', mock_ret_data);
```
piotmag769, line 753: Not finished

docs/src/appendix/cheatcodes/start_mock_call.md
abulenok: It is used here to indicate clarification

piotmag769: Thanks, mb, I was sure was en equivalent of giving an example!

crates/forge/src/cheatcodes_hint_processor.rs
piotmag769: True, can map tho

crates/forge/tests/integration/mock_call.rs
abulenok: Dispatcher will not allow to perform a call, it will throw an error

crates/forge/tests/integration/mock_call.rs
abulenok: Same as above

crates/forge/tests/integration/mock_call.rs
piotmag769: Can we use [`library_call_syscall`](https://docs.starknet.io/documentation/architecture_and_concepts/Smart_Contracts/system-calls-cairo1/#library_call) then? You can obtain `function_selector` using starknet.py https://starknetpy.readthedocs.io/en/latest/api/hash.html#starknet_py.hash.selector.get_selector_from_name

crates/forge/src/cheatcodes_hint_processor.rs
abulenok: yep, but is there any benefit of using `.into_iter().map(...).collect()` instead of the current solution?

crates/forge/tests/integration/mock_call.rs
piotmag769: True, we can try to deploy the contract after that and check if it works, should be fine

crates/forge/src/cheatcodes_hint_processor.rs
piotmag769: More of a nit, I don't mind

crates/forge/tests/integration/mock_call.rs
abulenok: nvm, it works https://github.com/foundry-rs/starknet-foundry/pull/385/commits/48bcc5ae204c2801b32fc71151fa0394a23aaf6c

piotmag769 (approved):
Good to merge, adjust the last test pls

crates/forge/tests/integration/mock_call.rs
abulenok: added an issue https://github.com/foundry-rs/starknet-foundry/issues/465

PR #387, opened 2023-08-03
Add prepare cheatcode design document

### LINKED ISSUE #63: Write a design doc for `prepare` cheatcode
Create a design document for `prepare` cheatcode or alternative solution.

cptartur:
Closes #63

## Introduced changes

- Add design document discussing and proposing a solution in place of `prepare` cheaatcode.

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests

### DIFF
--- a/design_documents/prepare_cheatcode.md
+++ b/design_documents/prepare_cheatcode.md
@@ -7,14 +7,13 @@
   * [Considered Solutions](#considered-solutions)
     * [Require Calling the `prepare` Cheatcode Before Every Deployment](#require-calling-the-prepare-cheatcode-before-every-deployment)
     * [Introducing the `precalculate_address` Cheatcode](#introducing-the-precalculateaddress-cheatcode)
-  * [Proposed Solution 1](#proposed-solution-1)
-    * [Salt "Counter"](#salt-counter)
-    * [`precalculate_address` Cheatcode](#precalculateaddress-cheatcode)
-    * [Known Problems With This Solution](#known-problems-with-this-solution)
-  * [Example Usage](#example-usage)
-  * [Proposed Solution 2](#proposed-solution-2)
+      * [Salt "Counter"](#salt-counter)
+      * [`precalculate_address` Cheatcode](#precalculateaddress-cheatcode)
+      * [Known Problems With This Solution](#known-problems-with-this-solution)
+      * [Example Usage](#example-usage)
+  * [Proposed Solution](#proposed-solution)
     * [`declare` Cheatcode](#declare-cheatcode)
-  * [Salt "Counter"](#salt-counter-1)
+    * [Salt "Counter"](#salt-counter-1)
     * [Known Problems With This Solution](#known-problems-with-this-solution-1)
     * [Example Usage](#example-usage-1)
 <!-- TOC -->
@@ -55,21 +54,14 @@ Users would have to perform an often unnecessary extra step with every deploymen
 
 Introducing the `precalculate_address` cheatcode that would return the contract address of the contract that would be
 deployed with `deploy` cheatcode.
-The problem with this approach is that for it to work, contract addresses would have to be deterministic.
-This would limit the user to only deploying a one instance of the given contact.
 
-## Proposed Solution 1
-
-Improving on the solution [proposed here](#introducing-the-precalculateaddress-cheatcode), allow precalculating the
-address while having the contract address semi-deterministic.
-
-### Salt "Counter"
+#### Salt "Counter"
 
 Introduce an internal "counter" and use its value to salt the otherwise deterministic contract address.
 Every time the `deploy` is called, increment this counter, so subsequent calls of `deploy` with the same `class_hash`
 will yield different addresses.
 
-### `precalculate_address` Cheatcode
+#### `precalculate_address` Cheatcode
 
 Introduce a cheatcode with the signature:
 
@@ -81,7 +73,7 @@ That will use the same method of calculating the contract address as `deploy` us
 This way the user will have an ability to know the address of the contract that will be deployed, and the current
 deployment flow will remain unchanged.
 
-### Known Problems With This Solution
+#### Known Problems With This Solution
 
 For the address returned by `precalculate_address` to match the address from `deploy`, the user will have to
 call `precalculate_address` immediately before the deployment or at least before any other calls to `deploy` as the
@@ -89,7 +81,7 @@ internal counter will then be incremented.
 
 This could be remedied by having separate counters for all `class_hashe`es, but it will still remain a limiting factor.
 
-## Example Usage
+#### Example Usage
 
 ```cairo
 mod HelloStarknet {
@@ -140,7 +132,7 @@ fn call_and_invoke() {
 }
 ```
 
-## Proposed Solution 2
+## Proposed Solution
 
 Change the current deployment flow, so it can better facilitate precalculating of contract addresses.
 
@@ -170,7 +162,7 @@ And remove the `deploy` cheatcode entirely.
 
 Both `precalculate_address` and `deploy` should use the same way of calculating the contract address.
 
-## Salt "Counter"
+### Salt "Counter"
 
 Introduce the same salt counter as [discussed here](#salt-counter).
 This will allow deterministic address calculation and deploying of multiple instances of the same contract.
--- a/design_documents/prepare_cheatcode.md
+++ b/design_documents/prepare_cheatcode.md
@@ -120,3 +120,63 @@ fn call_and_invoke() {
     let dispatcher = IHelloStarknetDispatcher { contract_address };
 }
 ```
+
+## Proposed Solution 2
+
+Change the current deployment flow, so it can better facilitate precalculating of contract addresses.
+
+### `declare` Cheatcode
+
+Change the `declare` cheatcode signature to this:
+
+```cairo
+trait DeclaredContractTrait {
+    fn precalculate_address(constructor_calldata: @Array::<felt252>) -> ContractAddress;
+    fn deploy(constructor_calldata: @Array::<felt252>) -> -> Result::<ContractAddress, RevertedTransaction>;
+}
+
+fn declare(contract: felt252) -> DeclaredContractTrait;
+```
+
+And remove the `deploy` cheatcode entirely.
+
+Both `precalculate_address` and `deploy` should use the same way of calculating the contract address.
+
+## Salt "Counter"
+
+Introduce the same salt counter as [discussed here](#salt-counter).
+This will allow deterministic address calculation and deploying of multiple instances of the same contract.
+
+### Known Problems With This Solution
+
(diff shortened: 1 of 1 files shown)


war-in:
How about such a flow? 

```rust
// pass calldata to the `contract` cheatcode
let contract = HelloStarknet(1, 2, 3);

// from now basic contract properties can be accessed (not sure which should be included)
let address = contract.address;
let salt = contract.salt;
let calldata = contract.calldata;  

// to interact with the defined contract it has to be declared and deployed
let class_hash = contract.declare();
contract.deploy(); 

// interact with the contract
let res = contract.some_method(); 
```

But I don't know if it is possible 🙈

cptartur:
> How about such a flow?
> 
> ```rust
> // pass calldata to the `contract` cheatcode
> let contract = HelloStarknet(1, 2, 3);
> 
> // from now basic contract properties can be accessed (not sure which should be included)
> let address = contract.address;
> let salt = contract.salt;
> let calldata = contract.calldata;  
> 
> // to interact with the defined contract it has to be declared and deployed
> let class_hash = contract.declare();
> contract.deploy(); 
> 
> // interact with the contract
> let res = contract.some_method(); 
> ```
> 
> But I don't know if it is possible 🙈

I think the problematic part of that would be `let contract = HelloStarknet(1, 2, 3);` but we could have something similar to that. 

Two things to consider with this approach:

1. Should `declare` remain or be done implicitly? I think it should still be there.
2. This would completely change the deployment flow from the current one. This might be worth it, but the solution proposed by me doesn't cause any breaking changes for the end user.

war-in:
> 1. Should declare remain or be done implicitly? I think it should still be there.
> 2. This would completely change the deployment flow from the current one. This might be worth it, but the solution proposed by me doesn't cause any breaking changes for the end user.

1. Agree
2. We are at the very early stage of the project so I think such changes are acceptable (especially when they are QOL improvements)

design_documents/prepare_cheatcode.md
```diff
@@ -0,0 +1,122 @@
...
+
+```cairo
+mod HelloStarknet {
+    // ...
+    
+    #[constructor]
+    fn constructor(ref self: ContractState) {
+        let caller_address = starknet::get_caller_address();
+        self.owner.write(caller_address);
+    }
```
piotmag769, line 83: Have been discussing with @Radinyn about this prank use case lately.

This approach to saving owner address is one that you would almost never use on real network, since most of the contracts are deployed through UDC (unless you have an entrypoint calling deploy syscall in your contract address) and `get_caller_address` would return UDC's address. I think we shouldn't use this example in this design doc (and not replicate it anywhere anymore, maybe even state that it is a bad practice and it won't work in real starknet most of the times?), because in real network there is no way you can prank UDC's address.

piotmag769: cc @MaksymilianDemitraszek @war-in

cptartur: Changing this example so it uses `warp` instead.

design_documents/prepare_cheatcode.md
```diff
@@ -0,0 +1,201 @@
...
+Change the current deployment flow, so it can better facilitate precalculating of contract addresses.
+
+### `declare` Cheatcode
+
+Change the `declare` cheatcode signature to this:
+
+```cairo
+trait DeclaredContractTrait {
+    fn precalculate_address(constructor_calldata: @Array::<felt252>) -> ContractAddress;
+    fn deploy(constructor_calldata: @Array::<felt252>) -> -> Result::<ContractAddress, RevertedTransaction>;
```
piotmag769, line 154:
```suggestion
trait DeclaredContractTrait {
    fn precalculate_address(self: @DeclaredContract, constructor_calldata: @Array::<felt252>) -> ContractAddress;
    fn deploy(self: @DeclaredContract, constructor_calldata: @Array::<felt252>) -> Result::<ContractAddress, RevertedTransaction>;
```

design_documents/prepare_cheatcode.md
```diff
@@ -0,0 +1,201 @@
...
+}
+
+fn declare(contract: felt252) -> DeclaredContractTrait;
```
piotmag769, line 157: Is it possible to return statically dispatched objects that implement traits in Cairo?
If not, it would be more like:

```cairo
struct DeclaredContract {
    class_hash: ClassHash,
    // not sure if more fields are needed
}

impl DeclaredContractTraitImpl of DeclaredContractTrait {
    // use `cheatcode` function from starknet to implement the trait
}

fn declare(contract: felt252) -> DeclaredContract;
```

piotmag769 (commented):
We need to decide on what Proposed Solution do we choose

design_documents/prepare_cheatcode.md
```diff
@@ -0,0 +1,201 @@
...
+## Proposed Solution 2
+
+Change the current deployment flow, so it can better facilitate precalculating of contract addresses.
+
+### `declare` Cheatcode
+
+Change the `declare` cheatcode signature to this:
+
+```cairo
+trait DeclaredContractTrait {
```
MaksymilianDemitraszek, line 152:
```suggestion
trait ContractClassTrait {
```
I am actually not sure if you can do impl without trait. If you can this should be tha way.

MaksymilianDemitraszek (commented):
Generally, LGTM. I would use a name `ContractClass` for this entity.

design_documents/prepare_cheatcode.md
cptartur: I think it's not possible, will change it to something similar to suggested.

design_documents/prepare_cheatcode.md
cptartur: I think you cannot, but will verify.

abulenok:
I suggest making users responsible for providing salt. We will have to modify `PreparedContract` struct a little bit:
```cairo
struct PreparedContract {
    class_hash: ...,
    constructor_calldata: ...,
    salt: Option<felt252>,
}
```

The flow with salt would look like this:
```cairo
#[test]
fn call_and_invoke() {
    // Declare the contract
    let class_hash = declare('HelloStarknet');
    
    // Prepare contracts for deployment
    let prepared1 = PreparedContract::new_with_salt( class_hash: class_hash, constructor_calldata: @ArrayTrait::new(), salt: 42 );
    let prepared2 = PreparedContract::new_with_salt( class_hash: class_hash, constructor_calldata: @ArrayTrait::new(), salt: 69 };
    
    // Precalculate the address
    let contract_address1 = precalculate_address(prepared1);
    let contract_address2 = precalculate_address(prepared2);
    
    // Warp the address
    start_warp(contract_address1, 1234);
    
    // Deploy with warped constructor
    let contract_address2 = deploy(prepared2).unwrap();
    let dispatcher2 = IHelloStarknetDispatcher { contract_address2 };

    let contract_address1 = deploy(prepared1).unwrap();
    let dispatcher1 = IHelloStarknetDispatcher { contract_address1 };
}
```

In the meantime, the old flow should also be viable, but we will add the `new` method for PreparedContract:
```cairo
fn new(class_hash: ..., constructor_calldata: ...) {
    PreparedContract { class_hash: class_hash, constructor_calldata: constructor_calldata, salt: Option::None }
}
```

```cairo
#[test]
fn call_and_invoke() {
    // Declare the contract
    let class_hash = declare('HelloStarknet');
    
    // Prepare contract for deployment
    let prepared = PreparedContract::new( class_hash: class_hash, constructor_calldata: @ArrayTrait::new() );
        
    // Deploy
    let contract_address = deploy(prepared).unwrap();
    let dispatcher = IHelloStarknetDispatcher { contract_address };
}
```

design_documents/prepare_cheatcode.md
```diff
@@ -0,0 +1,201 @@
...
+This way the user will have an ability to know the address of the contract that will be deployed, and the current
+deployment flow will remain unchanged.
+
+### Known Problems With This Solution
+
+For the address returned by `precalculate_address` to match the address from `deploy`, the user will have to
+call `precalculate_address` immediately before the deployment or at least before any other calls to `deploy` as the
+internal counter will then be incremented.
+
+This could be remedied by having separate counters for all `class_hashe`es, but it will still remain a limiting factor.
```
abulenok, line 90: If the user wants to deploy multiple instances of the same declared contract, we can enforce them to use different `PreparedContract` instances for it. We can differentiate instances of `PreparedContract` by their references

war-in (approved):

piotmag769 (approved):
What solution do we choose finally?

cptartur:
> WDYT about making users responsible for providing salt. We will have to modify `PreparedContract` struct a little bit:
> 
> ```cairo
> struct PreparedContract {
>     class_hash: ...,
>     constructor_calldata: ...,
>     salt: Option<felt252>,
> }
> ```
> 
> The flow with salt would look like this:
> 
> ```cairo
> #[test]
> fn call_and_invoke() {
>     // Declare the contract
>     let class_hash = declare('HelloStarknet');
>     
>     // Prepare contracts for deployment
>     let prepared1 = PreparedContract::new_with_salt( class_hash: class_hash, constructor_calldata: @ArrayTrait::new(), salt: 42 );
>     let prepared2 = PreparedContract::new_with_salt( class_hash: class_hash, constructor_calldata: @ArrayTrait::new(), salt: 69 };
>     
>     // Precalculate the address
>     let contract_address1 = precalculate_address(prepared1);
>     let contract_address2 = precalculate_address(prepared2);
>     
>     // Warp the address
>     start_warp(contract_address1, 1234);
>     
>     // Deploy with warped constructor
>     let contract_address2 = deploy(prepared2).unwrap();
>     let dispatcher2 = IHelloStarknetDispatcher { contract_address2 };
> 
>     let contract_address1 = deploy(prepared1).unwrap();
>     let dispatcher1 = IHelloStarknetDispatcher { contract_address1 };
> }
> ```
> 
> In the meantime, the old flow should also be viable, but we will add the `new` method for PreparedContract:
> 
> ```cairo
> fn new(class_hash: ..., constructor_calldata: ...) {
>     PreparedContract { class_hash: class_hash, constructor_calldata: constructor_calldata, salt: Option::None }
> }
> ```
> 
> ```cairo
> #[test]
> fn call_and_invoke() {
>     // Declare the contract
>     let class_hash = declare('HelloStarknet');
>     
>     // Prepare contract for deployment
>     let prepared = PreparedContract::new( class_hash: class_hash, constructor_calldata: @ArrayTrait::new() );
>         
>     // Deploy
>     let contract_address = deploy(prepared).unwrap();
>     let dispatcher = IHelloStarknetDispatcher { contract_address };
> }
> ```

I think this solution is quite a bit verbose in case the user wants to deploy multiple instances. The solution with keeping the internal salt counter (and I think protostar also used the similar solution, cc @MaksymilianDemitraszek) is more intuitive to use.

Though the user providing their own salt gives a bit more control over deployment, which could be beneficial, but I think the Proposed Solution 2 could be extended to allow for that as well.

MaksymilianDemitraszek (approved):

PR #408, opened 2023-08-04
Accessing events - design doc

war-in:
## Introduced changes

- `expect_events` cheatcode
- `spy_events` cheatcode


design_documents/expect_events_cheatcode.md
```diff
@@ -0,0 +1,78 @@
...
+}
+
+#[test]
+fn check_emitted_event() {
+    // ...
+    let mut expected_events = ArrayTrait::new();
+    	a.append(HelloEvent::StoredName);
+
+	expect_events(events);
+    let res = contract.emit_store_name(...);  // if the event is not emitted it will panic
```
MaksymilianDemitraszek, line 73: I assume this should be a liberal condition? I mean I can require events if specific order but if function emits more It should pass? Should we provide an option for exact matching as well?

war-in: Yup, definitely a liberal one. I think it is a good idea to add a flag for exact matching 👍

MaksymilianDemitraszek: flags are problematic because you need to pass `None` otherwise. We could maybe add `expect_exact_events` ?

design_documents/expect_events_cheatcode.md
```diff
@@ -0,0 +1,90 @@
...
+            self.emit(Event::StoredName(StoredName { user: get_caller_address(), name: name }));
+        }
+    }
+}
+
+#[test]
+fn check_emitted_event() {
+    // ...
+    let mut expected_events = ArrayTrait::new();
+    	a.append(HelloEvent::StoredName);
```
MaksymilianDemitraszek, line 79: How do you plant to extract event name from this structure?

design_documents/expect_events_cheatcode.md
```diff
@@ -0,0 +1,78 @@
...
```
MaksymilianDemitraszek, line 79: Will the implementation allow to specify event data?
I assume this proposal only covers asserting event names?

Will Cairo type checker accept this? You specified type od `expect_events` seems to take array of event instances.

design_documents/expect_events_cheatcode.md
```diff
@@ -0,0 +1,90 @@
...
+Propose a solution that will allow checking if events were emitted.
+
+## Considered Solutions
+
+### `expect_events` Cheatcode
+
+Introduce a cheatcode with the signature:
+
+```cairo
+fn expect_events(events: Array<Event>)
```
cptartur, line 18: Does the `Event` struct already exist? If not it should be a part of the proposal as well.

design_documents/expect_events_cheatcode.md
```diff
@@ -0,0 +1,90 @@
...
+}
+
+#[test]
+fn check_emitted_event() {
+    // ...
+    let mut expected_events = ArrayTrait::new();
+    	a.append(HelloEvent::StoredName);
+
+	expect_events(expected_events);
+    let res = contract.emit_store_name(...);  // if the event is not emitted it will panic
```
cptartur, line 82: So expect_events gets auto-cancelled after one call to the contract? This is a nice UX but I wonder if this can be reliably implemented.

cptartur (commented):
What about checking the contents of the expected events? Do we need that? The proposed solution doesn't really allow for anything like that.

design_documents/expect_events_cheatcode.md
war-in: I think it can be. My idea is to keep an array of expected events in the `CheatedState`; after the call execution, we will check if those were emitted. If they were we will simply remove them from the array

Arcticae:
I am wondering what's the use case of `stop_spy` since we do have to call `fetch_events` explicitly to get emitted events. The user can simply stop fetching

design_documents/accessing_emitted_events.md
```diff
@@ -0,0 +1,191 @@
...
+    keys: Array<felt252>,
+    data: Array<felt252>
+}
+```
+
+- `name` is a name of an event passed as a shortstring,
+- `keys` are values under `#[key]`-marked fields of an event,
+- `data` are all other values in emitted event.
+
+`expect_event` cheatcode will define events which should be emitted in the next call. Other calls will not be affected.
```
piotmag769, line 49: Just to make sure: let's say we call A, which calls B, B emits an event. We take into consideration the event B emitted, right?

design_documents/accessing_emitted_events.md
```diff
@@ -0,0 +1,191 @@
...
+If provided events will not be emitted it will panic with a detailed message.
+
+`events` are the subsequence of all events emitted by the function, but you can also require function
```
piotmag769, line 52: Subsequence suggests that they have to occur in that particular order, is that what you meant? If not, I would call it `subset`. Either way I would consider implementing both variants if that's possible (I guess we shall implement the `subset` variant now and leave `subsequence` for the future?)

design_documents/accessing_emitted_events.md
```diff
@@ -0,0 +1,191 @@
...
+trait EventFetcher {
+    fn fetch_events(ref self: EventSpy);
+}
+
+trait EventAssertions {
+    fn assert_emitted(ref self: EventSpy, events: Array<snforge_std::Event>);
+}
+```
+
+Users will be responsible for calling `fetch_events` to load emitted events to the `events` property.
```
piotmag769, line 149: Isn't it better to call it under the hood automatically every time `assert_emitted` is called? I don't see a case when someone would like to call `assert_emitted` without calling `fetch_events` first

design_documents/accessing_emitted_events.md
```diff
@@ -0,0 +1,191 @@
...
+    assert(spy.events.at(0).data == data, 'Wrong data');
+    
+    // or use prepared assertions
+    spy.assert_emitted(array![Event {from: ..., name: 'StoredName', ...}, Event {...}]);
+
+    let res = contract.emit_store_name(...);
+    let res = contract.emit_store_name(...);
+    
+    // events will not be present before fetching
+    assert!(spy.events.len() == 1, 'There should be one event');
```
piotmag769, line 183: I think the events array should be emptied every time `assert_emitted` is called. Otherwise `assert_emitted` loses its purpose when used multiple times, as we match against previous events as well

piotmag769 (commented):
I wonder if we should (and need to) have a solution to assert that events occurred in a specific order. Just an idea for the future, wouldn't focus on that for now, maybe create an issue to investigate it?

design_documents/accessing_emitted_events.md
war-in: 👍

design_documents/accessing_emitted_events.md
war-in: What if someone would like to check if an event was emitted and then perform some custom asserts? I'm not sure if it's a valid case but theoretically could happen 🤷

design_documents/accessing_emitted_events.md
war-in: What if someone calls fetch_events, checks the length of the array and then calls `assert_emitted` to check if some event was emitted? We would call `fetch_events` twice

design_documents/accessing_emitted_events.md
piotmag769: Well that depends what do we want to put emphasis on. Imo `assert_emitted` should provide a simple and intuitive way to check for events occurrences, so that user can easily insert it into test flow and bother neither about internal behaviours (like the one mentioned above) nor calling other functions (like `fetch_events`). If someone wants to perform custom asserts, they can use fetch_events and the `events` field

design_documents/accessing_emitted_events.md
```diff
@@ -0,0 +1,218 @@
...
+
+```cario
+struct Event {
+    name: felt252,
+    keys: Array<felt252>,
+    data: Array<felt252>
+}
+```
+
+- `name` is a name of an event passed as a shortstring,
```
piotmag769, line 45: Will it be a hashed name?

design_documents/accessing_emitted_events.md
```diff
@@ -0,0 +1,218 @@
...
+fn check_emitted_event_complex() {
+    // ...
+	let mut spy = spy_events();  // all events emitted after this line will be saved under the `spy` variable
+    let res = contract.emit_store_name(...);
+    
+    spy.fetch_events();
+    
+    // users can assert events on their own
+    assert(spy.events.len() == 1, 'There should be one event');
+    assert(spy.events.at(0).name == sn_keccak('StoredName'), 'Wrong event name');
```
piotmag769, line 202: I understand that the names in `events` field will be hashed names, right? Shouldn't we make a separate struct for that then? I know it will be basically the same but with different field name (`hashed_name` prob) but I think it will be less confusing

piotmag769 (approved):

design_documents/accessing_emitted_events.md
war-in: We could. But it is the `felt` in both cases so it is not necessary

piotmag769: Of course, these 2 structs would be isomorphic, but I think this way the code describes itself and there would be less confusion

war-in: The only moment where users have to worry about hashed names is when they are comparing them on their own. In all other cases, it is just a shortstring. I think this is quite clear, but can change it if you think it would be much better. 
I would also add it in the implementation, not in this design doc because it is rather an overview of the planned changes.

Arcticae (approved):
This is really good. I think most of this stuff can be copied over directly to the docs later on when we produce the actual feature.

One proposition from my side is to have a targeted spy - a `target` parameter for spy creation, which would be an enumeration of:
```
enum SpyOn {
    All,
    One(ContractAddress),
    Multiple(Array<ContractAddress>>)
}
```
This could also be a variation of the spy:
a `TargetedSpy` or something like that.

Issue #71, opened 2023-07-06
Raise gentle error when user tries to use syscalls not supported by `forge` test runner


cptartur:
https://github.com/software-mansion/protostar/issues/1886

Currently we directly use cairo repo test runner. We do not want to include starknet state and other starknet features into our test runner, thus we should remove them

piotmag769:
Is this still relevant? @MaksymilianDemitraszek @cptartur

cptartur:
Not sure

cptartur:
If user tries to use starknet related syscall in tests, we should throw user friendly error

PR #413, opened 2023-08-04
Fails test when user tries to use syscalls not supported by forge tes…

tkumor3:
…t runner

[Raise gentle error when user tries to use syscalls not supported by forge test runner](https://github.com/foundry-rs/starknet-foundry/issues/71)

Example of the error message, after call not supported syscall 
Zrzut ekranu 2023-08-4 o 15 24 34

- [x] Linked relevant issue
- [x] Added relevant tests


crates/forge/tests/integration/syscalls.rs
```diff
@@ -142,3 +142,29 @@ fn library_call_syscall() {
...
+        &test.path().unwrap().join("src/lib.cairo"),
+        &Some(test.linked_libraries()),
+        &Default::default(),
+        Some(&Utf8PathBuf::from_path_buf(corelib().to_path_buf()).unwrap()),
+        &test.contracts(corelib().path()).unwrap(),
+        &Utf8PathBuf::from_path_buf(predeployed_contracts().to_path_buf()).unwrap(),
+    )
+    .unwrap();
+
+    assert_failed!(result);
```
piotmag769, line 169: We could also assert that the output contains "starknet syscalls cannot be used in tests", we have a macro `assert_case_output_contains!` for that

crates/forge/src/cheatcodes_hint_processor.rs
```diff
@@ -282,14 +282,19 @@ fn execute_syscall(
     let mut buffer = MemBuffer::new(vm, system_ptr);
 
     let selector = buffer.next_felt252().unwrap().to_bytes_be();
+
+    if std::str::from_utf8(&selector).unwrap() != "CallContract" {
+        return Err(HintError::CustomHint(Box::from(
+            "starknet syscalls cannot be used in tests".to_string()
+        )));
+    }
```
piotmag769, line 290: I think you are right that it is the only case in which it happens, but let's ask @MaksymilianDemitraszek and @cptartur

cptartur: I think it's only syscalls, but not 100% sure about the internals of cairo code

crates/forge/tests/integration/syscalls.rs
cptartur: I'd definitely use that

piotmag769 (approved):
Let's wait for @MaksymilianDemitraszek to check the previous comment

piotmag769 (approved):

war-in (approved):

Arcticae (approved):

PR #418, opened 2023-08-07
Allow for custom class hash in account creation

THenry14:
## Introduced changes

- Allow to pass custom openzeppelin account contract class hash to account creation
- Fix tests failing due to lack of devnet update

## Breaking changes
None

## Checklist

- [X] Linked relevant issue
- [X] Added relevant tests


docs/src/starknet/account.md
```diff
@@ -75,3 +75,29 @@ command: account create
...
+    --name some-name \
+    --class-hash 0x058d97f7d76e78f44905cc30cb65b91ea49a4b908a76703c54197bca90f81773
+
+Account successfully created. Prefund generated address with at least 432300000000 tokens. It is good to send more in the case of higher demand, max_fee * 2 = 864600000000
+command: account create
+max_fee: 0x64a7168300
+address: 0x7a949e83b243068d0cbedd8d5b8b32fafea66c54de23c40e68b126b5c845b61
+
+$ sncast \
+  account deploy
```
drknzz, line 97:
```suggestion
  account deploy \
```

drknzz (commented):
👌

docs/src/starknet/account.md
THenry14: great catch

crates/cast/src/starknet_commands/account/create.rs
```diff
@@ -31,6 +31,9 @@ pub struct Create {
     #[clap(short, long)]
     pub add_profile: bool,
     // TODO (#253): think about supporting different account providers
+    /// Custom open zeppelin contract class hash of declared contract
+    #[clap(short, long)]
+    pub class_hash: Option<String>,
```
war-in, line 36:
```suggestion
    pub class_hash: Option<FieldElement>,
```
?

crates/cast/src/starknet_commands/account/deploy.rs
```diff
@@ -32,6 +36,7 @@ pub async fn deploy(
     network: &str,
     max_fee: FieldElement,
     wait: bool,
+    class_hash: Option<String>,
```
war-in, line 39: Shouldn't we rather save `class_hash` to the `accounts-file` than require passing it twice?

CHANGELOG.md
```diff
@@ -43,6 +43,16 @@ cheatcodes = { git = "https://github.com/foundry-rs/starknet-foundry", tag = "0.
...
+### Cast
+
+#### Added
+
+- Added `--class-hash` flag to account create/deploy, allowing for custom openzeppelin account contract class hash 
+
+#### Changed
+
+-
+
```
drknzz, line 55: Maybe this could be removed until something appears in the section

drknzz (approved):

crates/cast/src/starknet_commands/account/create.rs
THenry14: Thought about it but wanted to adjust to current hashes types (that are strings)
Maybe one day we will change all hashes to be FieldElements :)

crates/cast/src/starknet_commands/account/deploy.rs
THenry14: The file does not hold such information, and I'm not sure we should modify its structure by our own.
I suppose this is something that could be added to scarb.toml file instead?

war-in: hmm, are you sure we can't change the file structure? In my opinion, it should keep all important account info and class_hash is the one

THenry14: I'm not 100% sure, but I know this file is also created and worked upon by cairo-lang etc, I don't want to take a risk and render it unusable for users that use different tools.

I agree it would be a nice place to put it, but this is somehting that has to be coordinated with starkware.

war-in: Ok, so let's make an issue for this 🙏

war-in (approved):

Issue #305, opened 2023-07-27
Develop a way of testing the cheatnet in isolation


Arcticae:

MaksymilianDemitraszek:
blocked by #283

MaksymilianDemitraszek:
My vision:
- Tests should be in rust using API from #283 
- Contracts used in tests should be stored in repository as well as their compiled versions (to avoid compiling them each time we run tests)
- It should be fairly easy to recompile contracts

cptartur:
Also consider #405

PR #419, opened 2023-08-07
New Cheatcode API (Moved memory management to Forge)

### LINKED ISSUE #405: Decouple cheatable starknet from forge
Cheatable starknet should be more decoupled from forge. It should expose controllers that allow interacting with starknet state but it shouldn't handle reading cheatcode inputs, writing outputs to memory, error handling etc. Instead it should just take correctly typed arguments and return `Result`s. 

Reading inputs and outputs should be handled on the Forge side.

Also, the methods of the controller shouldn't necessarily be coupled directly with specific cheatcodes and should be composable e.g. if Forge decides to introduce a new cheatcode that combines the functionality of two already existing cheatcodes, it should be possible to implement just by calling necessary cheatcodes methods.

Radinyn:
Closes #405
Preparation for #305


crates/forge/src/cheatcodes_hint_processor.rs
```diff
@@ -209,29 +216,82 @@ impl CairoHintProcessor<'_> {
...
+                        Ok(())
+                    }
+                    Err(CheatcodeError::Recoverable(_)) => {
+                        panic!("Declare should not fail recoverably!")
+                    }
+                    Err(CheatcodeError::Unrecoverable(err)) => Err(err),
+                }
+            }
+            "deploy" => {
+                // TODO(#1991) deploy should fail if contract address provided doesn't match calculated
```
war-in, line 262: Are you sure about this number? 😅

war-in (approved):

crates/forge/src/cheatcodes_hint_processor.rs
Radinyn: It's an old protostar issue

war-in: Hmm, I thought that all protostar issues were ported here

crates/cheatable-starknet/src/cheatcodes/declare.rs
```diff
@@ -33,38 +31,40 @@ use starknet::core::types::contract::CompiledClass;
 impl CheatedState {
```
MaksymilianDemitraszek, line 31: Would it be possible to put BlockifierState into CheatedState to abstract the separation for forge devs? `CheatedState` could be renamed to `CheatnetState`

MaksymilianDemitraszek:
@Arcticae can you please look into that before merging?

Arcticae (approved):

crates/cheatable-starknet/src/cheatcodes/declare.rs
Arcticae: Seems like a valid take to me, @Radinyn WDYT?

crates/forge/src/cheatcodes_hint_processor.rs
Radinyn: yes, but this one was closed without addressing this Todo

piotmag769: I think this TODO is no relevant anymore

crates/cheatnet/src/cheatcodes.rs
```diff
@@ -45,15 +44,15 @@ impl From<EnhancedHintError> for HintError {
         }
     }
 }
+pub enum CheatcodeError {
+    Recoverable(Vec<Felt252>),        // Return result in cairo
```
piotmag769, line 48:
```suggestion
    Recoverable(Vec<Felt252>), // Return error result in cairo
```

piotmag769 (approved):

PR #426, opened 2023-08-07
Updated tests printing

### LINKED ISSUE #267: Improve test output printing
Tests found in the `src/lib.cairo` (or other packages when multiple are allowed) should display a better message. `Running X test(s) from src/lib.cairo` may be confusing.

bartekryba:
Closes #267

## Introduced changes
Replaces printing `Running x tests from src/lib.cairo` with `Running x tests from {package_name}`

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests

### DIFF
--- a/crates/forge/src/lib.rs
+++ b/crates/forge/src/lib.rs
@@ -306,13 +306,17 @@ fn run_tests_from_file(
     predeployed_contracts: &Utf8PathBuf,
 ) -> Result<TestFileSummary> {
     let runner = SierraCasmRunner::new(
-        tests.sierra_program.clone(),
+        tests.sierra_program,
         Some(MetadataComputationConfig::default()),
         OrderedHashMap::default(),
     )
     .context("Failed setting up runner.")?;
 
-    pretty_printing::print_running_tests(&tests, package_name);
+    pretty_printing::print_running_tests(
+        &tests.relative_path,
+        package_name,
+        tests.test_cases.len(),
+    );
 
     let mut results = vec![];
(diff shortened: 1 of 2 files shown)


crates/forge/src/lib.rs
```diff
@@ -310,7 +312,14 @@ fn run_tests_from_file(
     )
     .context("Failed setting up runner.")?;
 
-    pretty_printing::print_running_tests(&tests.relative_path, tests.test_cases.len());
+    let tests_source = if tests.relative_path == "src/lib.cairo" {
+        tests.package_name.as_str()
```
war-in, line 316: What do you think about adding a `package` string here?
It could be nice to have such an output
```
Running 1 test(s) from erc20 package
```

war-in: Also, where are those tests coming from? I assume they are unit tests collected from the entire package. Am I right?

bartekryba: Yup, these are the tests from the project tree

bartekryba: > What do you think about adding a package string here?

sounds good

crates/forge/src/lib.rs
```diff
@@ -310,7 +312,14 @@ fn run_tests_from_file(
...
+    } else {
+        tests.relative_path.as_str()
+    };
```
piotmag769, line 319: I would move this logic to `pretty_printing::print_running_tests` and just pass `tests` and `package_name` as args

piotmag769 (approved):

war-in (approved):

PR #437, opened 2023-08-08
State forking design doc

Arcticae:


design_documents/state_forking/state_forking.md
```diff
@@ -0,0 +1,60 @@
...
+while considering future expansion. 
+
+### MVP
+1. Implement fetching data (contracts, state, classes) utilizing a single worker utilizing `ForkDatabase` architecture
+2. Implement executing transaction on top of the forked state (should be able to modify the local state only)
+3. Implement user interface, enabling the forking in the tests
+
+### Further down the line
+3. Implement storing worker cache on disk, loading and clearing it
+4. Implement multifork, fetching utilizing multiple workers + controlling cheatcodes
```
war-in, line 60: 4, 5?

design_documents/state_forking/state_forking.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Considered solutions
+
+For the state to be considered as "forked", we have to swap the storage object 
+to a one which reads/writes on data from the fork. This can be done in two ways:
+
+- Pulling the state from the fork eagerly (use a fork DB dump as storage, clone and use it for reading & writing)
+- Pulling the state from the fork lazily (read from fork on read request, cache locally, write state diffs to local cache)
+
+The latter is preferred as the first is too heavy, inconsistent between node implementations, and generally
```
war-in, line 31: 👍

war-in (approved):

design_documents/state_forking/state_forking.md
```diff
@@ -0,0 +1,60 @@
...
+
+
+## Next steps
+
+In order to make things simple and easy to review, we should approach implementing this iteratively, 
+while considering future expansion. 
+
+### MVP
+1. Implement fetching data (contracts, state, classes) utilizing a single worker utilizing `ForkDatabase` architecture
+2. Implement executing transaction on top of the forked state (should be able to modify the local state only)
```
THenry14, line 55: If I understand correctly, we'll be able to pick a "starting state" (or rather block) from which we'll be getting txs until latest available block, is that right?

If so, does this:
```
Implement executing transaction on top of the forked state
```
mean that we will be able to only modify state in respect to latest block? or will there be a possibility to modify state from previous blocks as well? (eg that could imply forking state from specific block, to specific block - not necessarily the latest one). 

Do you see use cases for such scenarios? I imagine someone would want to have something like this to test multiple scenarios on states that changed in later blocks, but I'm not sure about this and just wanted to know what you think, and if this is something that's been considered.

Arcticae: Okay let me clear this up:

Forking at the specific block number, means we get the state for the specific block number. 
So all of the previous state is considered "included", with the most recent state delta being produced by the specified block.

THenry14: Ok cool, thanks for clarification. The part `(...) and optionally - a block number we want to fork state from.` mislead me a bit.

THenry14 (approved):

Issue #429, opened 2023-08-08
Scarb version parity with forge in tests


THenry14:
We should run sncast tests with the same scarb version forge is running its tests. The problem is, that at the moment devnet does not support starknet 0.12 (nor rpc 0.4.0 - and this won't be supported). In the future there will be devnet-rs that should help with this problem,  but for the time being we do not have many options.

We should explore our possibilities and decide on what to do next. Potential changes can be found on `szymczyk/update-scarb-in-tests` branch.

PR #448, opened 2023-08-09
Bump cairo to 2.1.0

### LINKED ISSUE #431: Bump to 2.1.0

bartekryba:
Closes #431 

## Introduced changes

- Bumped `cairo-lang` to `2.1.0`
- Bumped `scarb` to `0.6.0`

## Checklist

- [x] Linked relevant issue
- [ ] Added relevant tests

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,9 +40,16 @@ cairo-lang-syntax = "2.1.0"
 cairo-vm = "0.8.2"
 cairo-felt = "0.8.2"
 anyhow = "1.0.66"
+assert_fs = "1.0.13"
 camino = { version = "1.1.4", features = ["serde1"] }
 include_dir = "0.7.3"
 indoc = "2"
-tempfile = "3.6.0"
+itertools = "0.11.0"
+num-traits = "0.2.16"
 rayon = "1.7"
-
+regex = "1.9.3"
+serde = { version = "1.0.163", features = ["derive"] }
+serde_json = "1.0.104"
+starknet = "0.5.0"
+tempfile = "3.6.0"
+thiserror = "1.0.43"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1087,7 +1087,7 @@ dependencies = [
  "serde_json",
  "shellexpand",
  "snapbox",
- "starknet",
+ "starknet 0.4.0",
  "tempfile",
  "test-case",
  "tokio",
@@ -1130,7 +1130,7 @@ dependencies = [
  "num-traits 0.2.16",
  "regex",
  "serde_json",
- "starknet",
+ "starknet 0.5.0",
  "starknet_api",
  "tempfile",
  "thiserror",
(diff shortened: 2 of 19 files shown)


Cargo.toml
```diff
@@ -15,28 +15,28 @@ license = "Apache-2.0"
...
-cairo-lang-diagnostics = "2.1.0-rc2"
-cairo-lang-lowering = "2.1.0-rc2"
-cairo-lang-plugins = "2.1.0-rc2"
-cairo-lang-project = "2.1.0-rc2"
-cairo-lang-semantic = "2.1.0-rc2"
-cairo-lang-sierra-ap-change = "2.1.0-rc2"
-cairo-lang-sierra-gas = "2.1.0-rc2"
-cairo-lang-sierra-generator = "2.1.0-rc2"
-cairo-lang-syntax = "2.1.0-rc2"
+blockifier = { git = "https://github.com/starkware-libs/blockifier.git", rev = "df23f46" }
```
war-in, line 18: Shouldn't we use a commit marked in a release tag? Or we just don't care about that

scripts/prepare_for_tests.sh
```diff
@@ -4,7 +4,7 @@ set -e
 COMPILER_DIRECTORY="$(git rev-parse --show-toplevel)/crates/cast/tests/utils/compiler"
 CAIRO_REPO="https://github.com/starkware-libs/cairo/releases/download"
 
-SCARB_VERSIONS=("0.4.1" "0.5.2")
+SCARB_VERSIONS=("0.6.0")
```
THenry14, line 7: I'm not sure this will work, I had troubles updating to 0.6.0-rc1 (looked devnet did not support required libfuncs).
At the moment for sure it will fail, because hashes changed and a 2.0.2 compiler is required for devnet runner (they are set in tests/helpers/constants.rs). Scarb.toml test files also need compiler version bump in order to work.

You can try to run cast tests and see if it works with current devnet from master; alternatively you can revert this and I will try to bump it when new devnet releases.

scripts/prepare_for_tests.sh
```diff
@@ -4,7 +4,7 @@ set -e
 COMPILER_DIRECTORY="$(git rev-parse --show-toplevel)/crates/cast/tests/utils/compiler"
 CAIRO_REPO="https://github.com/starkware-libs/cairo/releases/download"
 
-SCARB_VERSIONS=("0.4.1" "0.5.2")
```
war-in, line 7: I think we should leave 0.4.1 as it was. We want to check a "range" of scarb versions

THenry14 (changes requested):

scripts/prepare_for_tests.sh
THenry14: Forgot to link an issue https://github.com/foundry-rs/starknet-foundry/issues/429

Cargo.toml
piotmag769: They don't have a release tag for 2.1.0 afaik

piotmag769 (approved):
Remember to update `CHANGELOG.md`, LGTM on forge's side

THenry14 (approved):

PR #453, opened 2023-08-10
`spy_events` cheatcode

### LINKED ISSUE #356: Support for `expect_events` cheatcode
Cheatnet should expose a feature so it is possible to check events emitted by the transaction in forge without digging through the internals of the transaction.

war-in:
Closes #356 

## Introduced changes

- `spy_events` cheatcode which creates a handler
- `EventSpy` struct which implements the `EventFetcher` trait and exposes `fetch_events` method
- `EventSpy` struct which implements the `EventAssertions` trait and exposes `assert_emitted` method

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


Radinyn (approved):

crates/forge/src/cheatcodes_hint_processor.rs
```diff
@@ -362,6 +365,46 @@ impl CairoHintProcessor<'_> {
...
+                let spy_on = match inputs.len() {
+                    0 => unreachable!(),
+                    1 => SpyOn::All,
+                    2 => SpyOn::One(contract_address_from_felt252(&inputs[1])?),
+                    _ => {
+                        let addresses_length = inputs[1].to_usize().unwrap();
+                        let addresses = Vec::from(&inputs[2..(2 + addresses_length)])
+                            .iter()
+                            .map(|felt| contract_address_from_felt252(felt).unwrap())
+                            .collect();
```
cptartur, line 389: I'm always unsure about these: Would it make sense to have `expect` with some messages instead of `unwrap`? Or is this overkill?

crates/forge/tests/integration/spy_events.rs
```diff
@@ -0,0 +1,384 @@
...
+        Contract::from_code_path(
+            "SpyEventsChecker".to_string(),
+            Path::new("tests/data/contracts/spy_events_checker.cairo"),
+        )
+        .unwrap()
+    );
+
+    let result = run_test_case(&test);
+
+    assert_failed!(result);
```
cptartur, line 91: You can consider adding `assert_case_output_contains` below `assert_failed` to verify if the message is being set correctly.

crates/forge/tests/integration/spy_events.rs
cptartur, line 1: Would a case where we expect e.g. 2 events and 3 are emitted be useful? 

Also a test where keys are note an empty array could be good to have. Or have I just missed it?

snforge_std/src/event.cairo
```diff
@@ -0,0 +1,183 @@
...
+            };
+
+            if !found {
+                panic(array![*events.at(i).name, 'event was not emitted from', (*events.at(i).from).into()]);
+            }
+
+            i += 1;
+        };
+    }
+}
```
cptartur, line 123: It would be great if this could be split into some smaller functions.

crates/cheatnet/src/cheatcodes/spy_events.rs
```diff
@@ -0,0 +1,47 @@
...
+                flattened.append(&mut event.keys.clone());
+                flattened.push(Felt252::from(event.data.len()));
+                flattened.append(&mut event.data.clone());
+
+                flattened
+            })
+            .collect();
+
+        let emitted_events_len = self.cheatcode_state.emitted_events.len();
+        self.cheatcode_state.emitted_events = vec![];
```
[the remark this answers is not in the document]
war-in: Yup, we are removing them to avoid flattening a large number of events each time. They are stored in the cairo code under the `EventSpy.event` and we append them while calling `spy.fetch_events()`

crates/forge/src/cheatcodes_hint_processor.rs
```diff
@@ -362,6 +365,46 @@ impl CairoHintProcessor<'_> {
                 Ok(())
             }
             "parse_json" => todo!(),
+            "spy_events" => {
+                let spy_on = match inputs.len() {
+                    0 => unreachable!(),
```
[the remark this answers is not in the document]
war-in: I think it is not possible, but we can

crates/forge/src/cheatcodes_hint_processor.rs
war-in: I think this is overkill because those values come directly form the cairo code where they were casted from ContractAddress to felt252

crates/forge/tests/integration/spy_events.rs
war-in: there is one test where keys are not empty
`expect_three_events_while_two_emitted`

cptartur: > there is one test where keys are not empty
expect_three_events_while_two_emitted

Never mind then. What about my first point?

war-in: will add it

crates/cheatnet/src/rpc.rs
```diff
@@ -172,6 +177,57 @@ pub fn call_contract(
...
+            },
+            data: ordered_event
+                .event
+                .data
+                .0
+                .iter()
+                .map(|data| stark_felt_to_felt(*data))
+                .collect(),
+        })
+        .collect();
```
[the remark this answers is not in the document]
war-in: I'm not sure if we want to keep a helper function for this one. It is a very specific case

snforge_std/src/event.cairo
```diff
@@ -0,0 +1,205 @@
...
+        }
+
+        data.append(*event.data.at(i));
+        i += 1;
+    };
+
+    Event { from, name, keys, data }
+}
+
+// Copied from corelib until it will be included in the release
```
Arcticae, line 172: This should be tracked by an issue i think, and we should probably ask the compiler team if it's going to be included

Arcticae (commented):
The event spy is basically a singleton right now, what if the user wants to have 2 spies in the same test (one spies on contract A, second one on contract B) and then assert that appropriate events were emitted from the contracts?

I think it would require to keep the spies in the map instead of a single spy in cheatable state

snforge_std/src/event.cairo
war-in: Afaik it will be included in the 2.2.0 release

crates/cheatnet/src/cheatcodes/spy_events.rs
```diff
@@ -18,33 +20,50 @@ pub enum SpyOn {
...
+        }
+    }
+}
+
 impl CheatnetState {
-    pub fn spy_events(&mut self, spy_on: SpyOn) {
-        self.cheatcode_state.spy_events = Some(spy_on);
+    pub fn spy_events(&mut self, spy_on: SpyOn) -> usize {
+        self.cheatcode_state.spies.push(spy_on);
+        self.cheatcode_state.spies.len() - 1
```
cptartur, line 36: Why `-1`?

crates/cheatnet/src/cheatcodes/spy_events.rs
```diff
@@ -18,33 +20,50 @@ pub enum SpyOn {
...
     }
 
-    pub fn fetch_events(&mut self) -> (usize, Vec<Felt252>) {
+    pub fn fetch_events(&mut self, id: &Felt252) -> (usize, Vec<Felt252>) {
+        let spy_on = &mut self.cheatcode_state.spies[felt_to_usize(id).unwrap()];
```
cptartur, line 40: So in case someone passes an id that is invalid, we just want this to panic entirely?

crates/forge/tests/integration/spy_events.rs
```diff
@@ -384,3 +386,74 @@ fn event_emitted_wrong_data_asserted() {
         "but keys or data are different"
     );
 }
+
+#[test]
+fn use_multiple_spies() {
```
cptartur, line 391: I'd also consider verifying that in case only one contract emits events and other not,`.emitted_events()` are empty for other contracts.

So there isn't some kind of interference between them.

MaksymilianDemitraszek (dismissed):
Please include the docs

crates/cheatnet/src/cheatcodes/spy_events.rs
```diff
@@ -0,0 +1,69 @@
+use crate::CheatnetState;
+use cairo_felt::Felt252;
+use cairo_vm::hint_processor::hint_processor_utils::felt_to_usize;
+use starknet_api::core::ContractAddress;
+
+/// Represents emitted event. It is used in the `CheatnetState` to keep track of events
+/// emitted in the `cheatnet::src::rpc::call_contract`
```
MaksymilianDemitraszek, line 7:
```suggestion
/// Represents an emitted event. It is used in the `CheatnetState` to keep track of events
/// emitted in the `cheatnet::src::rpc::call_contract`
```

crates/cheatnet/src/cheatcodes/spy_events.rs
```diff
@@ -0,0 +1,69 @@
...
+#[derive(Clone)]
+pub struct Event {
+    pub from: ContractAddress,
+    pub name: Felt252,
+    pub keys: Vec<Felt252>,
+    pub data: Vec<Felt252>,
+}
+
+/// Specifies which contract are spied on.
```
MaksymilianDemitraszek, line 16: I assume it works for events emitted by library calls? Can we include a cheatnet test for that?

crates/cheatnet/src/cheatcodes/spy_events.rs
```diff
@@ -0,0 +1,69 @@
...
+pub enum SpyOn {
```
MaksymilianDemitraszek, line 17:
```suggestion
pub enum SpyTarget {
```
Imo a better name

crates/cheatnet/src/cheatcodes/spy_events.rs
```diff
@@ -0,0 +1,69 @@
...
+    All,
+    One(ContractAddress),
```
MaksymilianDemitraszek, line 19: Why to have one target? Isn't it a subset of multiple? Maybe `All` and `Selected` would be better targets.

crates/cheatnet/src/cheatcodes/spy_events.rs
```diff
@@ -0,0 +1,69 @@
...
+    Multiple(Vec<ContractAddress>),
+}
+
+impl SpyOn {
+    pub fn should_be_spied(&mut self, contract_address: ContractAddress) -> bool {
```
MaksymilianDemitraszek, line 24: I don't understand what this function does.

crates/cheatnet/src/state.rs
```diff
@@ -88,6 +89,8 @@ pub struct CheatcodeState {
     pub pranked_contracts: HashMap<ContractAddress, ContractAddress>,
     pub warped_contracts: HashMap<ContractAddress, Felt252>,
     pub mocked_functions: HashMap<ContractAddress, HashMap<EntryPointSelector, Vec<StarkFelt>>>,
+    pub spies: Vec<SpyOn>,
```
MaksymilianDemitraszek, line 92:
```suggestion
    pub event_spy_targets: Vec<SpyOn>,
```

crates/cheatnet/tests/cheatcodes/spy_events.rs
MaksymilianDemitraszek, line 1: Please include tests:
#### validating order of the emitted events 
contract A emits event, calls contract B. B emits event and returns. A emits second event. What will be the order of spied events?

#### duplicate address is passed into multiple spy

### library call emits events

### constructor emits events

Sorry if something exists already and i still mentioned it, this PR is huge and a bit hard to review

snforge_std/src/event.cairo
```diff
@@ -0,0 +1,206 @@
...
+use traits::Into;
+
+use starknet::testing::cheatcode;
+use starknet::ContractAddress;
+
+#[derive(Drop, Serde)]
+enum SpyOn {
+    All: (),
+    One: ContractAddress,
+    Multiple: Array<ContractAddress>
```
MaksymilianDemitraszek, line 15: As I mentioned, one and multiple are not exclusive. I would replace it with `selected` instead.

crates/forge/tests/integration/spy_events.rs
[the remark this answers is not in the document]
MaksymilianDemitraszek: Would be nice to move those tests as cheatnet tests 👍

snforge_std/src/event.cairo
```diff
@@ -0,0 +1,206 @@
...
+    _id: felt252,
+    events: Array<Event>,
+}
+
+trait EventFetcher {
+    fn fetch_events(ref self: EventSpy);
+}
+
+impl EventFetcherImpl of EventFetcher {
+    fn fetch_events(ref self: EventSpy) {
```
MaksymilianDemitraszek, line 50: What if I call this, and then call another spied contract, and then fetch again? Will it fetch duplicates?

crates/cheatnet/src/cheatcodes/spy_events.rs
war-in: Because we add it to the vector and want to get it's index then

crates/cheatnet/src/cheatcodes/spy_events.rs
war-in: Fortunately, users don't have to pass anything here. `EventSpy` structure keeps its id as property and passes it itself

crates/cheatnet/src/cheatcodes/spy_events.rs
war-in: There is a problem with library calls. I noticed it yesterday and didn't have time to mention it anywhere. Basically, library calls don't return the contract address which invoked them. I wonder if it is a bug or a feature. 
However, we can deduct the address because one of the previous calls should have it. 
Will try to fix it asap

crates/cheatnet/src/cheatcodes/spy_events.rs
war-in: Maybe, but now one can do
```
spy_events(SpyOn::One(contract_address))
```

and then will have to 
```
spy_events(SpyOn::Selected(array![contract_address]))
```

The array is not necessary here

crates/cheatnet/src/state.rs
war-in: There is a problem with this approach.
Currently, we can assign an id to the `EventSpy` while creating it with the `spy_events` and then return the wanted events. We have to somehow know which spy wants which events. But it could be nice optimisation 🤔

crates/cheatnet/src/rpc.rs
```diff
@@ -169,6 +174,54 @@ pub fn call_contract(
...
+                        .events
+                        .iter()
+                        .map(|event| (code_address, event)),
+                );
+                break;
+            }
+        }
+
+        stack.extend(current_call.inner_calls.iter().rev());
+    }
```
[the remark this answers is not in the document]
war-in: Yup, they even have an `order` field inside. Which maybe I should consider using 🤔

snforge_std/src/event.cairo
war-in: Nope, we do not have one but you're right. It won't fetch duplicates

snforge_std/src/event.cairo
war-in: #509

crates/forge/tests/integration/spy_events.rs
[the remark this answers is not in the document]
war-in: Unfortunately, those tests are invoking `EventSpy.assert_emitted` which is written completely in Cairo, so I'm not able to test it on the cheatnet side

snforge_std/src/event.cairo
MaksymilianDemitraszek: Why 🤔

crates/cheatnet/src/rpc.rs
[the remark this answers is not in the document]
war-in: They are sorted now 😇

Arcticae (approved):

PR #454, opened 2023-08-10
Add flags to scarb config

### LINKED ISSUE #274: Allow all flags to be declarable in scarb config
Currently, we allow for `network`, `account` and `url` flags to be defined in scarb.toml config under `[tools.cast]`.
We should allow for more (or even all) flags to be declared in this file.

TBD if this also should include subcommands flags, and if yes - how we should approach it.

drknzz:
Closes #274 

## Introduced changes

- Refactor getting properties from config
- Add accounts-file to config

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/cast/src/helpers/scarb_utils.rs
```diff
@@ -12,31 +11,30 @@ pub struct CastConfig {
...
-    tool: &Option<BTreeMap<String, Value>>,
+pub fn get_property(tool: &Value, property: &str) -> String {
+    tool.get(property)
+        .and_then(serde_json::Value::as_str)
+        .map(String::from)
+        .unwrap_or_default()
+}
+
+pub fn get_property_from_profile(
+    tool_sncast: &Value,
```
THenry14, line 25: what is the reason to have two functions?

crates/cast/src/helpers/scarb_utils.rs
```diff
@@ -85,16 +83,47 @@ pub fn parse_scarb_config(
...
+
+    let tool = first_package
+        .manifest_metadata
+        .tool
+        .as_ref()
+        .ok_or_else(|| anyhow!("No field [tool] found in package"))?;
+
+    let tool_sncast = tool
+        .get("sncast")
+        .ok_or_else(|| anyhow!("No field [tool.sncast] found in package"))?;
```
THenry14, line 108: could probably be merged into one?

crates/cast/src/helpers/scarb_utils.rs
```diff
@@ -85,16 +83,47 @@ pub fn parse_scarb_config(
...
+}
+
+pub fn cast_config_from_package_tool_sncast(
+    package_tool_sncast: &Value,
+    profile: &Option<String>,
+) -> Result<CastConfig> {
+    let rpc_url = get_property_from_profile(package_tool_sncast, profile, "url")?;
+    let network = get_property_from_profile(package_tool_sncast, profile, "network")?;
+    let account = get_property_from_profile(package_tool_sncast, profile, "account")?;
+    let accounts_file = get_property_from_profile(package_tool_sncast, profile, "accounts-file")?;
```
THenry14, line 120: it would be cool, if we didn't have to do that, but instead iterate over keys from Scarb.toml and try to match them to CastConfig directly

crates/cast/src/helpers/scarb_utils.rs
```diff
@@ -85,16 +83,47 @@ pub fn parse_scarb_config(
...
 
     Ok(CastConfig {
         rpc_url,
         network,
         account,
+        accounts_file: (!accounts_file.is_empty()).then(|| accounts_file.into()),
```
THenry14, line 126: This should only be dealt with in main, probably

crates/cast/src/helpers/scarb_utils.rs
```diff
@@ -12,31 +11,30 @@ pub struct CastConfig {
     pub rpc_url: String,
     pub network: String,
     pub account: String,
+    pub accounts_file: Option<Utf8PathBuf>,
```
THenry14, line 14: This should just be `UtfPathBuf`, and if the value is not found in the config, we must take care of it in main (as it is with the rest of fields from this struct)

drknzz: Any arguments on why this should be just `Utf8PathBuf`?
I thought it was an optional field, that could either be present in the profile or not.

crates/cast/src/helpers/scarb_utils.rs
drknzz: Definitely could. Just thought it'd look cleaner.

crates/cast/src/helpers/scarb_utils.rs
drknzz: My reasoning was that both are smaller, hence they are easier to comprehend. The abstraction split works great here imo

crates/cast/src/helpers/scarb_utils.rs
THenry14: We do not have to set parameters as Optional, because CastConfig derives `Default` trait (see line 10 in this file), which means if something is not set, it gets the type default.

Now CastConfig is being constructed in main (`parse_scarb_config`), and in there it represents the final config state - and in this context this path is required (as well as other parameters - you can't run cast without rpc url). So this is the place where we currently do all the shenanigans regarding config - so, we look if there is something (anythong) configured in Scarb.toml, we check if something is being set using CLI, pick precedence etc.

We should rather stick to one approach - if you feel things should change, because they could be better - do it, and let's see how it is. Otherwise, let's stick to one approach :+1:

crates/cast/src/helpers/scarb_utils.rs
THenry14: maybe, but it seems to me that in `get_tools` you only return some field from tool, and the function is used in one place? so it perhaps could make sense to just do it in one place?

crates/cast/src/helpers/scarb_utils.rs
THenry14: I see. Wonder what is the memory and speed footprint on this (genuinely don't know).
Let's search for rust good practices, and try to align to them.

crates/cast/src/helpers/scarb_utils.rs
drknzz: If we talk about `get_tools`, it's used in 2 places (parse_scarb_config + add_created_profile_to_configuration). Also, seeing a call to function named `get_package_tool_sncast` immediately tells me that it's getting the reference to the [tool.sncast] and I don't need to know how it gets it and what are the errors it could return, making the code cleaner in my opinion.

Besides, breaking things down into smaller functions (even if they are used once) does not in fact increase the amount of code (besides the lines with the funcition's signature 😸) and makes the outer functions easier to grasp

crates/cast/src/helpers/scarb_utils.rs
drknzz: I don't think there would be any significance in the impact of this approach on the memory / execution speed.
Also, I didn't find any resources about it :/
I think it comes down to the coding style preferences / readability

crates/cast/src/helpers/scarb_utils.rs
drknzz: I was thinking already forward, as if to stick with this approach, other types would likely be of Option type as well.
To me it seemed that only the first 3 (rpc_url, network, account) would be required.

So is CastConfig supposed to be the actual config on which the program executes stuff, or just the config that is contained in Scarb.toml?
The current approach would suggest the second, as later, the config variable is not updated with the arguments passed through CLI, which have precedence over arguments from Scarb.toml.

THenry14: I see, in that case this is ok. Out of those 3, probably network shouldn't be required as well, since we can retrieve this from the provider.

Sure, the second approach is kind of what we have now. It could also make sense if we moved towards the first option though, I suppose. If you have a strong opinion that you would like to share, please do so. I'm honestly fine with both, but haven't given it much thought (and maybe I should :D ).

drknzz: I feel like it would be easier to create a single config that the current execution runs on and then just pass it around (so the first approach - to me it would make sense too).

THenry14: Sure, I like this. Do you want to handle this in this PR, or create another issue? both ways are fine for me

drknzz: I think I'll try to fiddle with it here, and later think about splitting it, if it were to turn out a big, confusing for this pr change

crates/cast/src/starknet_commands/account/create.rs
```diff
@@ -162,24 +153,31 @@ pub fn add_created_profile_to_configuration(
...
         let mut tool_sncast = toml::value::Table::new();
         let mut new_profile = toml::value::Table::new();
 
-        new_profile.insert("network".to_string(), Value::String(network));
-        new_profile.insert("url".to_string(), Value::String(url));
-        new_profile.insert("account".to_string(), Value::String(name.clone()));
+        new_profile.insert("network".to_string(), Value::String(config.network.clone()));
+        new_profile.insert("url".to_string(), Value::String(config.rpc_url.clone()));
+        new_profile.insert("account".to_string(), Value::String(config.account.clone()));
+        new_profile.insert(
```
drknzz, line 175: Should this always be inserted, even if it's pointing to the default file?

crates/cast/src/main.rs
```diff
@@ -1,20 +1,21 @@
...
 
-mod helpers;
+pub mod helpers;
 mod starknet_commands;
 
 #[derive(Parser)]
 #[command(version)]
 #[command(about = "Cast - a Starknet Foundry CLI", long_about = None)]
-struct Cli {
+pub struct Cli {
```
THenry14, line 18: why does it have to be pub?

crates/cast/src/main.rs
```diff
@@ -1,20 +1,21 @@
...
 };
 use anyhow::Result;
 use camino::Utf8PathBuf;
+use cast::helpers::constants::DEFAULT_ACCOUNTS_FILE;
+use cast::helpers::scarb_utils::{parse_scarb_config, CastConfig};
 use cast::{account_file_exists, get_account, get_block_id, get_provider, print_command_result};
 use clap::{Parser, Subcommand};
 
-mod helpers;
+pub mod helpers;
```
THenry14, line 12: why it needs to be `pub` ?

crates/cast/src/main.rs
```diff
@@ -226,3 +234,22 @@ async fn main() -> Result<()> {
         },
     }
 }
+
+pub fn update_cast_config(config: &mut CastConfig, cli: &Cli) {
```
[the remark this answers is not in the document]
THenry14: Let's try to move it to lib

crates/cast/src/main.rs
```diff
@@ -226,3 +234,22 @@ async fn main() -> Result<()> {
...
+        config.network = network.clone();
+    }
+    if let Some(account) = &cli.account {
+        config.account = account.clone();
+    }
+    if let Some(accounts_file) = &cli.accounts_file_path {
+        config.accounts_file = accounts_file.clone();
+    } else if config.accounts_file == Utf8PathBuf::default() {
+        config.accounts_file = DEFAULT_ACCOUNTS_FILE.into();
+    }
```
THenry14, line 252: maybe using something lke `unwrap_or_else` could help us to get rid of ifs here?

crates/cast/src/starknet_commands/account/create.rs
THenry14: yes

crates/cast/src/main.rs
drknzz: Without it, the compiler complains about some functions in scarb_utils.rs not being used anywhere

https://stackoverflow.com/a/67897487

crates/cast/src/main.rs
drknzz: Thought about it, but I don't know if this is better

```rust
fn update_cast_config(config: &mut CastConfig, cli: &Cli) {
    config.rpc_url = cli
        .rpc_url
        .clone()
        .unwrap_or_else(|| config.rpc_url.clone());
    config.network = cli
        .network
        .clone()
        .unwrap_or_else(|| config.network.clone());
    config.account = cli
        .account
        .clone()
        .unwrap_or_else(|| config.account.clone());

    let new_accounts_file = cli
        .accounts_file_path
        .clone()
        .unwrap_or_else(|| Utf8PathBuf::from(DEFAULT_ACCOUNTS_FILE));

}
```

crates/cast/src/main.rs
THenry14: that's odd, what functions are those? are they marked as pub?

crates/cast/src/main.rs
THenry14: hm, you're right. maybe we should declare a macro instead?

```
fn update_cast_config(config: &mut CastConfig, cli: &Cli) {
    macro_rules! clone_or_else {
        ($field:expr, $config_field:expr) => {
            $field.clone().unwrap_or_else(|| $config_field.clone())
        };
    }

    config.rpc_url = clone_or_else!(cli.rpc_url, config.rpc_url);
    config.network = clone_or_else!(cli.network, config.network);
    config.account = clone_or_else!(cli.account, config.account);

    let new_accounts_file = cli
        .accounts_file_path
        .clone()
        .unwrap_or_else(|| Utf8PathBuf::from(DEFAULT_ACCOUNTS_FILE));

}
```

more or less. wdyt?

crates/cast/src/main.rs
drknzz: image

They are

THenry14: I think the root problem is mod helpers is defined twice (in main and in lib), and one of the two is not being used. It would be cool to be able to get rid of one of the two; if this is not possible let's leave your change then. sounds good?

drknzz: It worked 🚀

crates/cast/src/helpers/scarb_utils.rs
```diff
@@ -85,17 +95,31 @@ pub fn parse_scarb_config(
             "Failed to read Scarb.toml manifest file, not found in current nor parent directories",
         )?;
 
-    let package = &metadata.packages[0].manifest_metadata.tool;
+    let package_tool_sncast_result = get_package_tool_sncast(&metadata);
+    if package_tool_sncast_result.is_err() {
+        return Ok(CastConfig::default());
+    }
+
+    CastConfig::from_package_tool_sncast(package_tool_sncast_result.unwrap(), profile)
```
war-in, line 103: I think we can match the result here
```rust
match get_package_tool_sncast(&metadata) {
    Ok(...) => ...,
    Err => ...
}
```

war-in (approved):
🔥

docs/src/appendix/cast/account/deploy.md
```diff
@@ -3,6 +3,7 @@ Deploy previously created account to Starknet.
 
 ## Required common arguments - passed by CLI or specified in Scarb.toml
 
+* [`accounts-file`](../common.html#--accounts-file--f-path_to_accounts_file)
```
war-in, line 6: Is it required now? I think it still has a default value 🤔

docs/src/appendix/cast/account/create.md
```diff
@@ -6,6 +6,7 @@ which is `~/.starknet_accounts/starknet_open_zeppelin_accounts.json` by default.
 
 ## Required common arguments - passed by CLI or specified in Scarb.toml
 
+* [`accounts-file`](../common.html#--accounts-file--f-path_to_accounts_file)
```
THenry14, line 9: This has a default value, so is not required :)

THenry14: Now that I see it, the flag should probably be passed to main sncast, and not to account (create|deploy) right?

THenry14 (approved):

Issue #347, opened 2023-07-31
L1 interoperability testing


MaksymilianDemitraszek:

Arcticae:
@MaksymilianDemitraszek can you elaborate on what is the scope of this?

MaksymilianDemitraszek:
This is a big one, without clearly defined scope. Probably will require extensive research and getting some information from teams developing cross chain solutions like bridges. 

Generally my vision here is a possibility to spawn anvil instance alongside the cheatnet instance and allow l1 <-> l2 messaging between them.

Arcticae:
@MaksymilianDemitraszek we should have a design doc for that, right?

glihm:
Very interesting stuff yes!

For now, this is how I do e2e testing for messaging. Spawning up katana (with messaging feature, currently in PR) and anvil, I use bash scripts to coordinate the state verification on both chains and execute messages on each side.

In the case of Starknet Foundry (`snforge`), we're in a testing environment, which means each time a test runs, we may expect a fresh state on the eth devnode. So how would be the UX for the user to determine which anvil setup he needs for which test, that's interesting part of the design.

Due to the asynchronous nature of messaging, we may also not be able to wait for a specific message in a test. And I am not sure that running `snforge`, we can expect some tests being triggered by anvil generated logs in a spontaneous manner.

But I'm not sure yet @MaksymilianDemitraszek at which level the forge code will "verify" what's going on on anvil (in the case L2->L1). How to express such expectations?

For L1->L2, we need anvil to fire a TX to generate a log, and then getting the logs, we can fire the L1 handler transaction. But this is what we're doing with the `l1_handler_execute` at some point.

So, thinking more about it, is it really in the scope of `snforge`? Or are we talking about integration of starknet devnet / anvil into a two-nodes devnet integrated into Starknet Foundry?

Arcticae:
BTW: Moving this to `general` since it needs design & research

MaksymilianDemitraszek:
@Arcticae I believe you are researching it rn, is it a duplicate of anything?

Arcticae:
#643 was a duplicate

Arcticae:
FYI:
I have some takeaways on this after chats with starkgate team & ex-MakerDAO developer

Arcticae:
Putting in triage in favour of #644 for now

PR #459, opened 2023-08-11
`l1_handler_execute` cheatcode

### LINKED ISSUE #334: Support testing l1_handlers

Forge

### Feature Request

Add support for calling `#[l1_handler]` from test.
While we do not yet have full l1<>l2 message testing, we should at least have the ability to directly call l1_handlers and test their output

glihm:
Closes #334 (and #347 ?)

## Introduced changes

This PR adds a cheatcode to reproduce what the sequencer does when
a message is received from ethereum.

For now, I have used the selector of the `#[l1_handler]` instead
of a cairo short string as in solidity code we use the selector, it looks
more natural to me.
If you prefer cairo short string from a UX perspective, I'll switch it. :+1:

## Breaking changes

New cheatcode.

## Checklist

- [X] Linked relevant issue
- [X] Added relevant tests


crates/cheatnet/src/cheatcodes/l1_handler_call.rs
```diff
@@ -0,0 +1,83 @@
...
+
+use starknet_api::core::{ContractAddress, EntryPointSelector};
+use starknet_api::hash::StarkFelt;
+use starknet_api::transaction::{Calldata, Fee, TransactionHash};
+
+use super::CheatcodeError;
+use crate::panic_data::try_extract_panic_data;
+
+impl CheatnetState {
+    pub fn l1_handler_call(
```
Radinyn, line 24: Perhaps there is a more descriptive name for this cheatcode? Current naming might be misleading. I'm thinking about something like `mock_l1_call`, do you have any other ideas?

crates/cheatnet/src/cheatcodes/l1_handler_call.rs
```diff
@@ -0,0 +1,83 @@
...
+            Err(err) => {
+                if let TransactionExecutionError::ExecutionError(
+                    EntryPointExecutionError::ExecutionFailed { error_data: felts },
+                ) = err
+                {
+                    let reason: String = felts_as_str(&felts);
+                    panic!("L1Transaction error: {reason:?}");
+                } else {
+                    panic!("L1Transaction error: {err:?}");
+                }
```
piotmag769, line 79: I wonder if we should panic the whole runner here or just a single test case. Is there a possibility that it fails due to user error and not error in our code?
cc @MaksymilianDemitraszek @Radinyn

docs/src/appendix/cheatcodes/l1_handler_call.md
```diff
@@ -0,0 +1,73 @@
...
+fn test_l1_handler_call() {
+    // ...
+    let data: Array<felt252> = array![1, 2];
+
+    let mut payload_buffer: Array<felt252> = ArrayTrait::new();
+    data.serialize(ref payload_buffer);
+
+    let l1_handler_prepared = PreparedL1Handler {
+        contract_address,
+        selector: 0x01e6b389ca484cb6fb23cbbcaa2db5581a8d970e3c135e8170c2ea5fdc2d3d8e,
```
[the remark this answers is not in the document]
MaksymilianDemitraszek: I think we should have similar flow here as @war-in is building for `expect-events`. @war-in can you elaborate on it here? Selector is a hashed function name so we probably can use funcionality introduced in https://github.com/foundry-rs/starknet-foundry/pull/466/files/fdd130349316dc45f3826221b034331c6dbf1e27..f386ba42c88e63f7c46bdfa86a9207e950de0f9d ?

The perfect solution would be to take selector as an argument and expose a function to calculate selector from function name.

crates/cheatnet/src/cheatcodes/l1_handler_call.rs
Radinyn: If we aren't sure it's safer to use `CheatcodeError::Unrecoverable` here

docs/src/appendix/cheatcodes/l1_handler_call.md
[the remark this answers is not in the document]
piotmag769: I would be more in favour of using raw name and calculating the entry point selector under the hood. I don't see real UX benefits coming from forcing users to manually calculate selectors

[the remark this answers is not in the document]
war-in: I agree with @piotmag769. We have to use the `event_name_hash` cheatcode because the hash can't be calculated with a standard keccak syscall.
I think it does not applicate here. We can pass shortstring and calculate selector under the hood

crates/cheatnet/src/cheatcodes/l1_handler_call.rs
```diff
@@ -0,0 +1,83 @@
...
+
+        let tx = L1HandlerTransaction {
+            tx: starknet_api::transaction::L1HandlerTransaction {
+                contract_address,
+                entry_point_selector,
+                calldata,
+                ..Default::default()
+            },
+            // TODO: Is tx_hash default always valid?
+            tx_hash: TransactionHash::default(),
```
MaksymilianDemitraszek, line 55: It is not valid, but I think it is not validated anywhere

crates/cheatnet/src/cheatcodes/l1_handler_call.rs
```diff
@@ -0,0 +1,83 @@
...
+            // TODO: In this context, is u128::MAX a good choice? May it be configurable?
+            paid_fee_on_l1: Fee(u128::MAX),
```
MaksymilianDemitraszek, line 57: Would be nice to have it configurable I propose a solution in a different comment

docs/src/appendix/cheatcodes/l1_handler_call.md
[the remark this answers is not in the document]
MaksymilianDemitraszek: Agreed, important to note: this interface should probably be consistent with mock_call interface

MaksymilianDemitraszek (changes requested):
Amazing job 🔥 . I included some suggestions regarding the interface so we make sure it is consistent and flexible

crates/cheatnet/src/cheatcodes/l1_handler_call.rs
piotmag769: Yea but that is included above - if transaction is reverted we bubble RecoverableError

docs/src/appendix/cheatcodes/l1_handler_call.md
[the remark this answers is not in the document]
glihm: All right, I'll then go in that direction with the selector. :+1:

@MaksymilianDemitraszek in that case there is usually no return value. We may include it? Like this if someone wants the `l1_handler` to return a value for testing he can (which is also viable in production, just that the value is never used).

crates/cheatnet/src/cheatcodes/l1_handler_call.rs
glihm: @Radinyn IML `mock_l1_call` sounds more that we're trying to call the `L1`, or the `L1` is calling. Which is I think confusing too.

And also, it's not a `mock`, as we actually call the function.

The terminology `l1_handler` should appear. `call` sounds clear to me, and `l1_handler_mock` would be "we fake the call", which is not what we are doing.

WDYT?

crates/forge/src/cheatcodes_hint_processor.rs
```diff
@@ -302,6 +301,44 @@ impl CairoHintProcessor<'_> {
...
+                let from_address = inputs[2].clone();
+                let payload_length: usize = inputs[3]
+                    .clone()
+                    .to_usize()
+                    .expect("Payload length is expected to fit into usize type");
+
+                let mut payload = vec![];
+                for felt in inputs.iter().skip(4).take(payload_length) {
+                    payload.push(felt.clone());
+                }
```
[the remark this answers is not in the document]
glihm: So clean.. thank you @piotmag769. :+1:

crates/cheatnet/src/cheatcodes/l1_handler_call.rs
MaksymilianDemitraszek: maybe `trigger_l1_handler`? mock sounds off to me as well

docs/src/appendix/cheatcodes/l1_handler_call.md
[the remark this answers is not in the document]
MaksymilianDemitraszek: Not sure if I understood, you would like to provide a way of asserting if certain value is returned by handler? 

I think there is no point in providing such assertion atm as it is not consumed in any way, IMHO triggering is enough

docs/src/appendix/cheatcodes/l1_handler_call.md
```diff
@@ -0,0 +1,73 @@
...
+```
+
+We can use `l1_handler_call` to call a `l1_handler`, which is not accessible using a dispatcher:
+
+```rust
+use snforge_std::l1_handler_call;
+
+#[test]
+fn test_l1_handler_call() {
+    // ...
```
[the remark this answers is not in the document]
glihm: I do like this idea of having something we can build, and then execute, instead of the prepare fashion. :+1: I'll go that way thanks a lot @MaksymilianDemitraszek.

About the name, as commented to @Radinyn I am not sure `mock` is that accurate IMO as the call to the function is done. But it effectively can be more consistent in the naming convention you already have.

crates/cheatnet/src/cheatcodes/l1_handler_call.rs
MaksymilianDemitraszek: Yup, we can stick with it

docs/src/appendix/cheatcodes/l1_handler_call.md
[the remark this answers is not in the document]
MaksymilianDemitraszek: I agree Mock is not accurate what to you think about `trigger`?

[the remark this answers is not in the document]
glihm: `trigger` sounds nice. I was also thinking about `l1_handler_invoke` because it is what we do and Starkware documents it like this.

`call` sounds too much like a regular function so yes it should be discarded.

docs/src/appendix/cheatcodes/l1_handler_call.md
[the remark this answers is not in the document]
glihm: Ok ok we've the same opinion, I was confused by "having the same interface as mock_call" sorry. Let's keep it with the invocation/trigger as it's normally done. :+1:

docs/src/appendix/cheatcodes/l1_handler_call.md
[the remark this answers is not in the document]
glihm: > Maybe we could create a more extensible interface here 🤔 What do you thing about something like this
> 
> ```
> MockL1Call::build(contract_address, function_name, payload).execute()
> ```
> 
> and if somebody wants to include additional configuration
> 
> ```
> let l1mock = MockL1Call::build(contract_address, function_name, payload).execute()
> l1mock.from_address = 0x123123
> l1mock.paid_fee_on_l1 = 0x123123
> ```
> 
> What do you think?

After doing some tests, I am not sure how to use this approach.

We almost always want to set `from_address` for instance. Which mean I'll always have 2 lines of code to init the struct.

```rust
let l1_handler = L1HandlerInvoker::build(contract_address, selector_name, payload);
l1_handler.from_address = 0x1234;
l1_handler.invoke();
```

However, what's is very interesting in your approach is that when we test a contract, we don't want to always repeat the `contract_address` when changing the arguments to see how the contract behaves. Same for the selector name. We want to play with `payload` and `from_address` mostly, to ensure our `l1_handler` implementation is correctly setup.

What about this:

```rust
// Contract 1

#[l1_handler]
fn accept_l1_data(from_address: felt252, value1: felt252, value2: felt252) {
    assert(from_address == 0x123, 'Invalid L1 address');
    assert(value1 != value2, 'Invalid values');
}
```

Then we can:

```rust
// Test function

let contract_address = deploy(...);

// Build the reusable struct.
let l1_handler = L1Handler::build(contract_address, 'accept_l1_data');

// Invoke with given parameters (contract_address, payload, fees).
l1_handler.invoke(0x123, @array![0x11, 0x22], 0); // << will pass
l1_handler.invoke(0x999, @array![0x1, 0x1], 0); // << will revert
```
What's your take on this one?

Also @Radinyn and @MaksymilianDemitraszek for the naming, what do you think about the convention given above:
`L1Handler` struct to build the targetted handler.
`.invoke` to actually fire the TX and the underlying execution flow.

docs/src/appendix/cheatcodes/l1_handler_invoke.md
```diff
@@ -0,0 +1,68 @@
+# `l1_handler_invoke`
+
+> `fn build(contract_address: ContractAddress, selector_name: felt252) -> L1HandlerFn`
+> `fn invoke(self: L1HandlerFn, from_address: felt252, fee: u128, payload: Span::<felt252>)`
```
glihm, line 4: Not sure how to document this @piotmag769 @Radinyn, any suggestion?

docs/src/appendix/cheatcodes/l1_handler_invoke.md
```diff
@@ -0,0 +1,68 @@
...
+#[test]
+fn test_l1_handler_invoke() {
+    // ...
+    let data: Array<felt252> = array![1, 2];
+
+    let mut payload_buffer: Array<felt252> = ArrayTrait::new();
+    data.serialize(ref payload_buffer);
+
+    let l1_handler = L1Handler::build(contract_address, 'process_l1_message');
+    l1_handler.invoke(0x123, 1, payload.span());
```
glihm, line 65: The UX is effectively a lot better. Are you ok with the naming guys?

snforge_std/src/cheatcodes.cairo
```diff
@@ -179,3 +185,39 @@ fn get_class_hash(contract_address: ContractAddress) -> ClassHash {
...
+
+impl L1HandlerImpl of L1Handler {
+    fn build(contract_address: ContractAddress, selector_name: felt252) -> L1HandlerFn {
+        L1HandlerFn {
+            contract_address,
+            selector_name,
+        }
+    }
+
+    fn invoke(self: L1HandlerFn, from_address: felt252, fee: u128, payload: Span::<felt252>) {
```
glihm, line 202: The Fee here must never be 0, as the runner will fail due to the actual fee being higher than 0.
But if the user put at least 1, it's good as I used in the example.

snforge_std/src/cheatcodes.cairo
```diff
@@ -179,3 +185,39 @@ fn get_class_hash(contract_address: ContractAddress) -> ClassHash {
...
     (*buf[0]).try_into().expect('Invalid class hash value')
 }
+
+trait L1Handler {
+    fn build(contract_address: ContractAddress, selector_name: felt252) -> L1HandlerFn;
+    fn invoke(self: L1HandlerFn, from_address: felt252, fee: u128, payload: Span::<felt252>);
+}
+
+impl L1HandlerImpl of L1Handler {
+    fn build(contract_address: ContractAddress, selector_name: felt252) -> L1HandlerFn {
```
glihm, line 195: At the first time using this I was like "why using build function instead of init a struct", but @MaksymilianDemitraszek you were already thinking about "what if we want to do some processing during the build", right? Which.. well, sounds pretty nice and the interface is still elegant IMO.

snforge_std/src/cheatcodes.cairo
```diff
@@ -19,6 +19,12 @@ struct ContractClass {
     class_hash: ClassHash,
 }
 
+#[derive(Drop, Clone)]
+struct L1HandlerFn {
+    contract_address: ContractAddress,
+    selector_name: felt252,
```
glihm, line 25: Is `selector_name` ok, or `function_name` would be better?

snforge_std/src/cheatcodes.cairo
glihm: @MaksymilianDemitraszek I removed the `build` from the trait as it forced someone to directly init from the trait `L1HandlerTrait::build(..)` which look like less natural. For instance the dispatchers for `MockCall` are not built with a `build` function but directly instantiating the struct.

Is this interface ok for you then?

```rust
let l1_handler = L1Handler {
    contract_address,
    selector_name: 'process_l1_message'
};

l1_handler.invoke(0x123, 1, payload.span());
```

docs/src/appendix/cheatcodes/l1_handler_invoke.md
glihm: please see comment above for the `L1Handler` struct initialization.

docs/src/appendix/cheatcodes/l1_handler_invoke.md
glihm: I've rework the interface for only have invoke.

snforge_std/src/cheatcodes.cairo
MaksymilianDemitraszek: `function_name` is more accurate, selector is the value of hash

docs/src/appendix/cheatcodes/l1_handler_invoke.md
```diff
@@ -0,0 +1,68 @@
...
+
+Invokes a `#[l1_handler]` function to mock a
+[message](https://docs.starknet.io/documentation/architecture_and_concepts/L1-L2_Communication/messaging-mechanism/)
+arriving from Ethereum.
+
+```rust
+struct L1Handler {
+    contract_address: ContractAddress,
+    selector_name: felt252,
+}
```
MaksymilianDemitraszek, line 13: I still would prefer the solution I proposed, if we have it this way you need to provide `from_address`, `fee` and `payload` each time even if it is not relevant to the tested code. IMO better interface would be
```
L1HandlerInvoke {
    selector_name: felt252,
    from_address: felt252, 
    fee: u128,
    payload: Span::<felt252>
}


trait L1HandlerTrait {
    fn default(contract_address, selector_name);
    fn execute(self: L1Handler);
}
```
and `default` would set all remaining  fields to `0` or sth.

This way interface would be more extensible as you could add more fields without introducing breaking changes. If somebody has to configure some additional parameters then it is easy to just create the structure by hand.

snforge_std/src/cheatcodes.cairo
MaksymilianDemitraszek: Why build forces you to do that? Imo you can provide `default` as a simplified interface and if somebody needs more customization then they can create the structure by hand.

glihm: Answered in other related comment. :+1:

docs/src/appendix/cheatcodes/l1_handler_invoke.md
glihm: Understood, I've made some tests for spelling and usage. The only point I see is the fact that one must not forget to use `mut` on the instance that is returned from the `new` method to customize it (which is mandatory as default empty payload will never fit a use case as it's the cairo deserialization that will fail, not the user code).

Is this ok for you?

1. New was more "array like", it may sound more familiar to users. Or do you definitely prefer `default`?
2. `execute` sounds nice to me too due to the tx that is executed, let's go with it!

```rust
let data: Array<felt252> = array![1, 2];

let mut payload_buffer: Array<felt252> = ArrayTrait::new();
data.serialize(ref payload_buffer);

let mut l1_handler = L1HandlerTrait::new(contract_address, function_name: 'process_l1_message');

// Testing execute here makes almost no sense to me, or I am perhaps missing
// something important. Except perhaps someone sending message without any payload
// and testing the address 0.

l1_handler.from_address = 0x123;
l1_handler.payload = payload.span();

l1_handler.execute();
```

Radinyn (approved):

docs/src/appendix/cheatcodes/l1_handler_invoke.md
MaksymilianDemitraszek: SGTM

docs/src/appendix/cheatcodes/l1_handler_execute.md
```diff
@@ -0,0 +1,76 @@
...
+
+- `contract_address` - The target contract address
+- `function_name` - Name of the `#[l1_handler]` function
+- `from_address` - Ethereum address of the contract that sends the message
+- `fee` - The fees paid on L1
+- `payload` - The message payload that may contain any Cairo data structure that can be serialized with
+[Serde](https://book.cairo-lang.org/appendix-03-derivable-traits.html?highlight=serde#serializing-with-serde)
+
+It is important to note that when executing the `l1_handler`,
+the `from_address` must always be checked as any L1 contract can call any L2 contract.
```
MaksymilianDemitraszek, line 29: I am not sure what does it mean "must be checked" 🤔

docs/src/appendix/cheatcodes/l1_handler_execute.md
```diff
@@ -0,0 +1,76 @@
...
+
+For contract implementation:
```
MaksymilianDemitraszek, line 31:
```suggestion
For a contract implementation:
```

docs/src/appendix/cheatcodes/l1_handler_execute.md
```diff
@@ -0,0 +1,76 @@
...
+
+#[l1_handler]
+fn process_l1_message(ref self: ContractState, from_address: felt252, data: Span<felt252>) {
+    assert(from_address == self.l1_allowed.read(), 'Unauthorized l1 contract');
+}
+// ...
+```
+
+We can use `execute` method to test the execution of the `#[l1_handler]` function that is
+not available using a dispatcher:
```
MaksymilianDemitraszek, line 51:
```suggestion
We can use `execute` method to test the execution of the `#[l1_handler]` function that is
not available through contracts dispatcher:
```

MaksymilianDemitraszek (approved):
images
Amazing job 🔥 Thanks for the patience in addressing my comments 😄 

I am ok with merging after addressing minor stuff in docs I mentioned, @Arcticae if you want to glance, it is the last chance.

docs/src/appendix/cheatcodes/l1_handler_execute.md
glihm: In the sense that, not checking this can be an important security consideration. But perhaps you're right some people want to do open messaging. :P

glihm: Just reworded with a `may be checked`. If it's not relevant at this level of the documentation, let me know and I'll remove it.

MaksymilianDemitraszek: I don't understand what exactly has to be checked

glihm: The from_address argument, as you contract on L2 can receive L1HandlerTx from any L1 contract. As the official documentation always warn the user on this, I thought it was worthy to have a reminder here.

We can remove this sentence if you think it is not relevant in this context.

Radinyn: I'd keep this sentence, imo it makes sense

PR #460, opened 2023-08-11
Loading data docs + fix minor docs stuff

piotmag769:
## Checklist

- [x] Linked relevant issue
- [ ] Added relevant tests

### DIFF
--- a/docs/src/appendix/forge-library/deserialize_txt.md
+++ b/docs/src/appendix/forge-library/deserialize_txt.md
@@ -1,0 +1,3 @@
+# `deserialize_txt`
+
+
--- a/docs/src/appendix/forge-library/deserialize_txt.md
+++ b/docs/src/appendix/forge-library/deserialize_txt.md
@@ -1,3 +1,53 @@
 # `deserialize_txt`
 
 
+> `trait TxtParser<T, impl TSerde: Serde<T>> {
+>     fn deserialize_txt(file: @File) -> Option<T>;
+> }` 
+
+Parses plain text file content and tries to deserialize it to type `T` that implements `Serde` trait.
+
+- `file` - a snapshot of an instance of the struct `File` that consists of the following fields:
+  - `path` - Cairo shortstring representing a path to a file relative to a package root.
+
+```rust
+use option::OptionTrait;
+use serde::Serde;
+use snforge_std::{ FileTrait, TxtParser };
+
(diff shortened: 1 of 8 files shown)


docs/src/appendix/forge-library/parse_txt.md
```diff
@@ -0,0 +1,34 @@
...
+  - integers in range of `[0, P)` where P is [`Cairo Prime`](https://book.cairo-lang.org/ch02-02-data-types.html?highlight=prime#felt-type)
+  - strings of length `<=31` enclosed in single or double quotation marks
+
+For example, this file content:
+```txt
+1  2   'hello' 10     "world"
+```
+will be parsed to the following array:
+```rust
+array![1, 2, 'hello', 10, 'world]
```
war-in, line 33:
```suggestion
array![1, 2, 'hello', 10, 'world']
```

docs/src/appendix/forge-library/deserialize_txt.md
```diff
@@ -0,0 +1,53 @@
...
+    let my_struct = TxtParser::<MyStruct>::deserialize_txt(@file).unwrap();
+    // ...
+}
+```
+
+File content must consists of elements that:
+- have to be separated with whitespaces
+- have to be either:
+  - integers in range of `[0, P)` where P is [`Cairo Prime`](https://book.cairo-lang.org/ch02-02-data-types.html?highlight=prime#felt-type)
+  - strings of length `<=31` enclosed in single or double quotation marks
```
war-in, line 36: why do we allow for both single and double quotation marks? Shouldn't we wait with double until strings are supported?

war-in (approved):

docs/src/appendix/forge-library/deserialize_txt.md
piotmag769: I did that only to be in line with future json support, as json standard does not allow single quoted strings. Not sure if it is a good idea though cc @MaksymilianDemitraszek

war-in: I think we could add double quotes in the json PR

MaksymilianDemitraszek: I think we should use single quutation for short strings, double quoutations should be used for long strings once those are released.

PR #462, opened 2023-08-14
feat: gas usage estimation report

enitrat:
Closes #

## Introduced changes

- Adds required logic to output gas usage estimation of tests
-

## Checklist

- [ ] Linked relevant issue
- [ ] Added relevant tests

### DIFF
--- a/crates/forge/src/pretty_printing.rs
+++ b/crates/forge/src/pretty_printing.rs
@@ -3,6 +3,7 @@ use crate::TestFileSummary;
 use anyhow::Error;
 use camino::Utf8PathBuf;
 use console::style;
+use num_traits::ToPrimitive;
 
 pub fn print_error_message(error: &Error) {
     let error_tag = style("ERROR").red();
@@ -51,11 +52,23 @@ pub(crate) fn print_test_result(test_result: &TestCaseSummary) {
         | TestCaseSummary::Passed { name, .. } => name,
     };
 
+    let gas_used = match test_result {
+        TestCaseSummary::Passed { available_gas, run_result, .. }
+        | TestCaseSummary::Failed { available_gas, run_result: Some(run_result), .. } if available_gas.is_some() => {
+            run_result.gas_counter.clone().and_then(|gas_counter| {
+                gas_counter.to_usize().map(|gas| available_gas.unwrap() - gas)
+            })
+        },
+        _ => None,
+    };
+
+    let gas_estimation_message = gas_used.map_or(String::new(), |gas| format!(" (gas usage est.: {})", gas));
+
     let result_message = match test_result {
         TestCaseSummary::Passed { msg: Some(msg), .. } => format!("\n\nSuccess data:{msg}"),
         TestCaseSummary::Failed { msg: Some(msg), .. } => format!("\n\nFailure data:{msg}"),
         _ => String::new(),
     };
 
-    println!("{result_header} {result_name}{result_message}");
+    println!("{result_header} {result_name} {gas_estimation_message} {result_message}");
 }
--- a/crates/forge/src/pretty_printing.rs
+++ b/crates/forge/src/pretty_printing.rs
@@ -53,16 +53,25 @@ pub(crate) fn print_test_result(test_result: &TestCaseSummary) {
     };
 
     let gas_used = match test_result {
-        TestCaseSummary::Passed { available_gas, run_result, .. }
-        | TestCaseSummary::Failed { available_gas, run_result: Some(run_result), .. } if available_gas.is_some() => {
-            run_result.gas_counter.clone().and_then(|gas_counter| {
-                gas_counter.to_usize().map(|gas| available_gas.unwrap() - gas)
-            })
-        },
+        TestCaseSummary::Passed {
+            available_gas,
+            run_result,
+            ..
+        }
+        | TestCaseSummary::Failed {
+            available_gas,
+            run_result: Some(run_result),
+            ..
+        } if available_gas.is_some() => run_result.gas_counter.clone().and_then(|gas_counter| {
+            gas_counter
+                .to_usize()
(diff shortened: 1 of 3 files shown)


piotmag769:
@enitrat that's actually a more complex issue I think. From what I see `gas_counter` is always set to `None` in `run_function` at first and then modified in closure, but it does not seem to have any effect...
cc @MaksymilianDemitraszek

enitrat:
> then modified in closure, but it does not seem to have any effect...

Yeah that's the conclusion I had too. But the thing is, I tried to compare it with the current implementation of `cairo-test`, which returns the gas consumption on main. Both of them look very similar in how they handle the invocation of the test functions, however in sn-foundry `gas_counter` is not assigned a value, but in `cairo-test` it is.  🤔

enitrat:
Oh, I think I might understand why.
Starknet foundry uses a dependency cairo-* with release 2.1.0, but this gas estimation change is not included in the latest releases and is only available on main. I will try updating the dependencies for cairo-runner and cairo-test-runner, and see if it provides the expected output.

Anyway, this can probably be put on hold until SW adds this commit in a release.

enitrat:
update: I tried to base the Cairo dependencies on the commit that had this feature, but it still didn't work. So I think the difference is in how starknet foundry invokes tests, compared to cairo-test. Cairo test runs tests using `run_function_with_starknet_context` which takes a parameter of type `StarknetState` which is passed into the `hint_processor`. I won't dig further down into it but I guess it's related to gas calculations

MaksymilianDemitraszek:
@enitrat Sorry for not being top responsive here, had a lot on my plate in the recent days, thank you for persistence in trying to push the task forward 🥇 

Generally It was a good idea to compare it with cairo-test, the problem is starknet foundry and cairo test have different execution models, and those will drive different design and implementation decisions behind gas reporting. Differentiation between sierra gas and starknet gas makes things complicated. Cairo test uses sierra gas only and blockifier which starknet foundry uses for contract execution uses the old starknet gas.


Design wise I am kinda against `available_gas` attribute and I believe gas assertions should be a part of the test control flow, not declarative. Reasonig behind this is that you generally want to assert gas cost of your contract transactions and exclude the costs of the test like setting up calldata. 

In the far future I would imagine having something like this 
```
test {
   let result = ContractAResourcesDispatcher.get_balance()
   assert(result.gas_cost < 15555, `too expensive`)
}
```

Going back to the ground and what we want to build now. I think a good starting point would be to accumulate gas cost of declare, deploy and contract calls inside a test and print it in the result `[PASS] path/to/test (estimated gas: 174917)` instead of asserting it in any way. This way we would create something useful for the present users until sierra and starknet gases are consolidated, and then we could develop a better and more flexible solution as the one proposed above.

Places worth looking into implementation wise as those may be helpful: 
https://github.com/foundry-rs/starknet-foundry/blob/c868a82f97c030374d0b2e3268b9206ad03e648c/crates/cheatnet/src/rpc.rs#L90
https://github.com/foundry-rs/starknet-foundry/blob/c868a82f97c030374d0b2e3268b9206ad03e648c/crates/cheatnet/src/rpc.rs#L806
https://github.com/starkware-libs/blockifier/blob/bf38764a1af0825389179f4df0e9f7e604f94edb/crates/blockifier/src/transaction/transactions_test.rs#L355

We probably want to extend `CheatableState` with `starknet_interactions_gas` variable and bump it with every interaction cost. Then print it at the end.

enitrat:
>  think a good starting point would be to accumulate gas cost of declare, deploy and contract calls inside a test and print it in the result [PASS] path/to/test (estimated gas: 174917)

I agree with the output, however something I don't really understand is why that should be linkeed to a declare/deploy/calls flow. Ideally, I want sn-foundry to be contract agnostic, and to be able to unit tests functions of my code without deploying anything.

There are too many cases where I want to test internal functions that will not be exposed publicly in the contract interface. Solidity is a language whose purpose is to write smart contracts, Cairo is not - and these differences should be taken into account in the way sn-foundry is built.

I think it's one of Cairo's strengths, and where sn-foundry should differ by allowing us to test functions isolated from the starknet environment. I'm just a bit familiar with ethereum foundry, but here if my contract uses a custom `add()` function, I want to be able to test this function without deploying a contract first.

enitrat:
> Then, we will include call_internal calls to the gas calculation, does it make sense?

yep, definitely

> Tests having undelying state (and being effectively a contract) are imho confusing

makes sense. 

> Our plan to workaround this limitation is to introduce call_internal cheatcode which will let users to call all member functions of a contract (and include a dispatcher for it to improve UX).

So I would still need to deploy a contract, right?

For Kakarot, we have very few entrypoints and most of our functionalities are "libraries" that are not exposed externally. Even if we can call internal functions, that would require us to deploy the contract first. 

I want, for example, to be able to test my Stack data-structure without interacting with a contract. This is possible in the actual state of SN-foundry. Will it be possible with the design you're thinking about.

MaksymilianDemitraszek:
>  I think protostar worked this way, right?

Yes it did, we wanted to change that

> For Kakarot, we have very few entrypoints and most of our functionalities are "libraries" that are not exposed externally. Even if we can call internal functions, that would require us to deploy the contract first.

It depends, you will be able to import and run functions, but if your function modifies or reads contract state (storage, caller address etc.) you need it deployed as a contract. Do you think it will be too limiting 🤔 ?

enitrat:
> Do you think it will be too limiting 🤔 ?

It is more limiting than having it possible natively, for sure. That's something I found really appreciated about protostar. But as long as it remains possible throug deployment, it should be fine,especially if we have extra features making it easier.

We should continue this discussion in TG, to have more opinions

cc @eikix @jobez

MaksymilianDemitraszek:
I wouldn't bother about performance too much tbh, there is a lot of work we can do in case it is necessary but right now compilation will be the biggest bottleneck

PR #466, opened 2023-08-17
Support `Keccak` syscall

### LINKED ISSUE #440: Unit testing includes keccak call crash

Forge

### Foundry Version

v0.3.0

Linux

x86

### What happened

I have several unit tests, and the test code calls the keccak-related function interfaces in the cairo corelib. These tests are crashing this time like this:

```
Collected 1 test(s) and 1 test file(s)
Running 1 test(s) from src/lib.cairo
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: ExpectedInteger(Relocatable { segment_index: 10, offset: 2 })', crates/forge/src/cheatcodes_hint_processor.rs:370:50
```

After locating the issue, I found that it is related to the invocation of keccak. 

**Steps of Reproduction**

You can try running the built-in test cases for keccak in the cairo corelib, which can be found [here](https://github.com/starkware-libs/cairo/blob/main/corelib/src/test/keccak_test.cairo#L41). 

Although the `cairo_keccak` method is a new function added in v2.1.0, I still encounter the same issue when manually calling `keccak_syscal` prior to this version.

### Trace

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: ExpectedInteger(Relocatable { segment_index: 10, offset: 2 })', crates/forge/src/cheatcodes_hint_processor.rs:370:50
stack backtrace:
   0: rust_begin_unwind
             at /rustc/8ede3aae28fe6e4d52b38157d7bfe0d3bceef225/library/std/src/panicking.rs:593:5
   1: core::panicking::panic_fmt
             at /rustc/8ede3aae28fe6e4d52b38157d7bfe0d3bceef225/library/core/src/panicking.rs:67:14
   2: core::result::unwrap_failed
             at /rustc/8ede3aae28fe6e4d52b38157d7bfe0d3bceef225/library/core/src/result.rs:1651:5
   3: <forge::cheatcodes_hint_processor::CairoHintProcessor as cairo_vm::hint_processor::hint_processor_definition::HintProcessorLogic>::execute_hint
   4: cairo_vm::vm::vm_core::VirtualMachine::step
   5: cairo_vm::vm::runners::cairo_runner::CairoRunner::run_until_pc
   6: cairo_lang_runner::casm_run::run_function
   7: cairo_lang_runner::SierraCasmRunner::run_function
   8: forge::run
   9: snforge::main
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

- [X] I have searched the existing issues and verified no issue exits for this problem.

war-in:
Closes #440

## Introduced changes

- keccak syscall can be used directly in the tests 

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/cheatnet/src/rpc.rs
```diff
@@ -661,6 +664,69 @@ pub fn library_call_syscall(
     })
 }
 
+pub enum KeccakOutput {
+    Success { ret_data: KeccakResponse },
+    Panic { panic_data: Vec<Felt252> },
+}
+
+pub fn keccak_syscall(
```
piotmag769, line 672: Is it copied from somewhere? If so can we leave a comment with path to a file and repo like in the fn below?

crates/forge/src/cheatcodes_hint_processor.rs
```diff
@@ -423,8 +434,39 @@ fn execute_syscall(
...
+        KeccakOutput::Success { ret_data } => {
+            buffer.write(gas_counter).unwrap();
+            buffer.write(Felt252::from(0)).unwrap();
+
+            buffer.write(ret_data.result_low).unwrap();
+            buffer.write(ret_data.result_high).unwrap();
+        }
+        KeccakOutput::Panic { panic_data } => {
+            buffer.write(gas_counter).unwrap();
+            buffer.write(Felt252::from(0)).unwrap();
```
piotmag769, line 465: I think we should write `1` here (if it is an exit code)

crates/forge/tests/integration/syscalls.rs
```diff
@@ -172,7 +173,184 @@ fn test_call_syscall_fail_in_test_fn() {
...
+        &test.contracts(&corelib_path()).unwrap(),
+        &Utf8PathBuf::from_path_buf(predeployed_contracts().to_path_buf()).unwrap(),
+    )
+    .unwrap();
+
+    assert_failed!(result);
+}
+
+#[test]
+fn test_cairo_keccak() {
```
piotmag769, line 252: What does it test?

piotmag769 (approved):

crates/cheatnet/src/rpc.rs
war-in: Yup, it is changed in few places to adjust errors but should be marked as `inspired`

crates/forge/tests/integration/syscalls.rs
war-in: It tests other common keccak function `cairo_keccak` which uses `keccak_syscall` under the hood

piotmag769 (approved):

crates/forge/src/cheatcodes_hint_processor.rs
```diff
@@ -423,8 +434,39 @@ fn execute_syscall(
...
+    let input_end = buffer.next_addr().unwrap();
+    let request = KeccakRequest {
+        input_start,
+        input_end,
+    };
+
+    let keccak_result = keccak_syscall(
+        &request,
+        buffer.vm(),
+        &mut (gas_counter.to_u64().unwrap() - KECCAK_GAS_COST),
```
Arcticae, line 451: Shouldn't only the inside function mutate the gas?

Arcticae (approved):
🔥

crates/cheatnet/src/rpc.rs
```diff
@@ -661,6 +664,70 @@ pub fn library_call_syscall(
     })
 }
 
+pub enum KeccakOutput {
+    Success { ret_data: KeccakResponse },
+    Panic { panic_data: Vec<Felt252> },
+}
+
+// Inspired by blockifier::execution::syscalls::mod::keccak
+pub fn keccak_syscall(
```
MaksymilianDemitraszek, line 673: Why we cannot redirect execution to `original_cairo_hint_processor`? It has keccak implemented and I would prefer us to avoid maintaining hash functions implementations even it is only for test purposes.

MaksymilianDemitraszek (changes requested):

crates/forge/tests/integration/syscalls.rs
```diff
@@ -151,7 +152,146 @@ fn test_call_syscall_fail_in_test_fn() {
...
+    "#
+    ));
+
+    let result = run_test_case(&test);
+
+    assert_passed!(result);
+}
+
+#[test]
+fn test_keccak_syscall_too_small_input() {
```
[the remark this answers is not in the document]
war-in: done

war-in:
As @MaksymilianDemitraszek suggested, it can be done without copying keccak logic. We just had to pass it to the `original_cairo_hint_processor`

crates/forge/src/cheatcodes_hint_processor.rs
```diff
@@ -400,20 +409,42 @@ fn execute_syscall(
...
-            "starknet syscalls cannot be used in tests".to_string(),
-        )));
-    }
+    let mut system_ptr = get_ptr(vm, cell, &offset)?;
+
+    // We peek into memory to check the selector
+    let selector = DeprecatedSyscallSelector::try_from(felt_to_stark_felt(
+        &vm.get_integer(system_ptr).unwrap(),
+    ))?;
+    system_ptr.offset += 1;
```
MaksymilianDemitraszek, line 424: Why do we increment it here 🤔  I think it will break if you call another syscall after keccak syscall

crates/forge/src/cheatcodes_hint_processor.rs
```diff
@@ -437,8 +468,16 @@ fn execute_syscall(
...
-    Ok(())
+fn execute_keccak(
+    original_cairo_hint_processor: &mut OriginalCairoHintProcessor,
+    vm: &mut VirtualMachine,
+    exec_scopes: &mut ExecutionScopes,
+    hint_data: &Box<dyn Any>,
+    constants: &HashMap<String, Felt252>,
+) -> Result<(), HintError> {
+    original_cairo_hint_processor.execute_hint(vm, exec_scopes, hint_data, constants)
 }
```
MaksymilianDemitraszek, line 481: Do we need this function 🤔 It would be more readable if we directly called `original_cairo_hint_processor.execute_hint` from pattern match. You would know that it is offloaded.

MaksymilianDemitraszek: Otherwise if it stays we could probably mark this `#inline`?

crates/forge/tests/integration/syscalls.rs
```diff
@@ -2,6 +2,7 @@ use crate::integration::common::runner::Contract;
 use crate::integration::common::running_tests::run_test_case;
```
MaksymilianDemitraszek, line 2: Please include a test which calls keccak and another syscall afterwards

MaksymilianDemitraszek (changes requested):

crates/forge/src/cheatcodes_hint_processor.rs
war-in: Why do you think so? In the last test, I call the Keccack syscall and then the CallContract syscall. Everything works fine

piotmag769: I think it's fine since `get_integer` probably does not change the `offset`, but couldn't we just use `next_felt252` which (I assume) increments `offset` under the hood?

crates/forge/tests/integration/syscalls.rs
```diff
@@ -146,7 +147,193 @@ fn test_call_syscall_fail_in_test_fn() {
...
+        .unwrap()
+    );
+
+    let result = run_test_case(&test);
+
+    assert_passed!(result);
+}
+
+#[test]
+fn compare_keccak_from_contract_with_plain_keccak() {
```
piotmag769, line 297: So I assume blockifier keccak == starknet keccak != plain cairo keccak?

piotmag769 (approved):

crates/forge/src/cheatcodes_hint_processor.rs
war-in: We can't do it, because we need to pass the VM to the `original_cairo_hint_processor.execute_hint` which would not be possible if we created the buffer before.

crates/forge/tests/integration/syscalls.rs
war-in: They all are the same. We just want to check if there are no unexpected behaviours when calling it from the contract

crates/forge/src/cheatcodes_hint_processor.rs
piotmag769: Makes sense, we do not want to clone the vm

MaksymilianDemitraszek (approved):

piotmag769:
I can see that increasing the syscall pointer was a mistake - we pass args to cheatcode hint processor without specyfying which cheatcode to execute, so I assume it reads the selector and increases the pointer. How did it work before when we increased the syscall pointer? Why did nothing break?
cc @MaksymilianDemitraszek @war-in

war-in:
Fortunately, it was not a mistake. It was just confusing. We were increasing the pointer but the original processor got values from the previous function. Which means increasing the pointer didn't affect it. We changed that to be as clear as possible.

PR #470, opened 2023-08-17
Fix runner-crashing common cases

### LINKED ISSUE #455: Add an error-handling case for non-existing entrypoint

Forge

### Foundry Version

0.3.0

MacOS

x86

### What happened

We should add handling to a case when entry point fails with:

```
Err(PreExecutionError(EntryPointNotFound(EntryPointSelector(StarkFelt("0x03ee0bfaf5b124501fef19bbd1312e71f6966d186c42eeb91d1bff729b91d1d4")))))
```

And fail a test, outputting a nice message instead of `Unparseable message`

### Trace

_No response_

Arcticae:
Closes #455

## Introduced changes

- Added a few common catches for lvl-1 invocations (test-contract)
- Makes cairo errors with trace, fail just one case with given trace (contract-to-contract)

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/cheatnet/src/rpc.rs
```diff
@@ -164,9 +165,25 @@ pub fn call_contract(
...
+        let selector_hash = selector.0;
+        let msg = format!("Entry point selector not found in contract: {selector_hash}");
+        Ok(CallContractOutput::Error { msg })
+    } else if let Err(EntryPointExecutionError::PreExecutionError(
+        PreExecutionError::UninitializedStorageAddress(contract_address),
+    )) = exec_result
+    {
+        let address = contract_address.0.key().to_string();
+        let msg = format!("Contract not deployed at address: {address}");
+        Ok(CallContractOutput::Error { msg })
```
Radinyn, line 186: It could probably be converted to a match statement

crates/cheatnet/src/rpc.rs
```diff
@@ -164,9 +165,25 @@ pub fn call_contract(
...
+            Ok(CallContractOutput::Panic { panic_data })
+        } else {
+            Ok(CallContractOutput::Error { msg: trace })
+        }
+    } else if let Err(EntryPointExecutionError::PreExecutionError(
+        PreExecutionError::EntryPointNotFound(selector),
+    )) = exec_result
+    {
+        let selector_hash = selector.0;
+        let msg = format!("Entry point selector not found in contract: {selector_hash}");
```
piotmag769, line 178:
```suggestion
        let msg = format!("Entry point selector {selector_hash} not found in contract {contract_address}");
```

piotmag769 (approved):

crates/forge/src/cheatcodes_hint_processor.rs
```diff
@@ -428,9 +428,14 @@ fn execute_syscall(
...
     .unwrap_or_else(|err| panic!("Transaction execution error: {err}"));
 
+    if let CallContractOutput::Error { msg } = call_result {
+        return Err(HintError::CustomHint(Box::from(msg)));
+    }
+
     let (result, exit_code) = match call_result {
         CallContractOutput::Success { ret_data } => (ret_data, 0),
         CallContractOutput::Panic { panic_data } => (panic_data, 1),
+        CallContractOutput::Error { .. } => panic!("Should be unreachable"),
```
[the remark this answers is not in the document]
Radinyn: @piotmag769 If it should be unreachable I think that panicking here is preferable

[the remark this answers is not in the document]
piotmag769: I removed the if with return above, so that we don't match against it twice - it should be reachable in my code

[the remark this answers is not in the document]
Radinyn: Oops! Didn't see that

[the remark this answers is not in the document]
piotmag769: Btw I would use [`unreachable`](https://doc.rust-lang.org/std/macro.unreachable.html) macro if some parts of code should be unreachable in the future

crates/cheatnet/src/rpc.rs
```diff
@@ -83,6 +83,7 @@ type SyscallSelector = DeprecatedSyscallSelector;
 pub enum CallContractOutput {
     Success { ret_data: Vec<Felt252> },
     Panic { panic_data: Vec<Felt252> },
+    Error { msg: String },
```
Radinyn, line 86: This will cause a conflict with #305 assert macros.

Arcticae: Which ones specifically? I think all tests are passing now

Radinyn: It's already fixed

Radinyn (approved):
Perhaps some tests made the Cheatnet way are possible, otherwise LGTM.

PR #472, opened 2023-08-17
Fix forge on windows

piotmag769:
Tested manually on windows, still need to fix some tests as "Windows bad" (absolute paths in toml issues), but in general it works with integration tests and with template

From now on we should use only `join` instead of slashes when creating new paths

## Checklist

- [x] Linked relevant issue
- [ ] Added relevant tests

### DIFF
--- a/crates/forge/src/scarb.rs
+++ b/crates/forge/src/scarb.rs
@@ -231,12 +231,12 @@ mod tests {
         )
         .unwrap();
 
-        let snforge_std_path = Utf8PathBuf::from_str("..")
+        let snforge_std_path = Utf8PathBuf::from_str("../../snforge_std")
             .unwrap()
-            .join("..")
-            .join("snforge_std")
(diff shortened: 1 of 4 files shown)


crates/forge/src/scarb.rs
```diff
@@ -162,7 +162,7 @@ pub fn corelib_for_package(metadata: &Metadata, package: &PackageId) -> Result<U
     let corelib = compilation_unit
         .components
         .iter()
-        .find(|du| du.source_path.to_string().contains("core/src"))
+        .find(|du| du.name == "core")
```
war-in, line 165: What about the `src`?

war-in (approved):

crates/forge/src/scarb.rs
piotmag769: I match the name, not the path

Issue #278, opened 2023-07-25
Get current configuration


THenry14:
We should add a possibility, to quickly determine with what configuration we are currently running. It could be useful to make double sure some calls someone tries to make are done from specific configuration, eg something like:

```
$ sncast show-config
network: example
account: example
url: http://example.com
(...)
```

There should be a new subcommand added - `show-config`, that prints out (if set) currently used:
- profile
- network
- rpc url
- account
- path to scarb toml
- accounts file path

Here's an example, existing command: https://github.com/foundry-rs/starknet-foundry/blob/621a1a5abb49fe41f459f655901cff0c915809e8/crates/cast/src/starknet_commands/call.rs#L12

All commands are parsed and executed from main: https://github.com/foundry-rs/starknet-foundry/blob/621a1a5abb49fe41f459f655901cff0c915809e8/crates/cast/src/main.rs#L81

btc-icosahedron:
can I try this one?

THenry14:
sure :)

PR #476, opened 2023-08-18
Support other networks

### LINKED ISSUE #404: Support other networks
We should think about adding support for other, custom networks (like integration). 
Most custom networks seem to use identifiers like "testnet" or "testnet2." To avoid confusion with actual testnets, it could be a good idea to come up with a solution to clearly mark these as custom networks, maybe in the accounts file.

One thing to keep in mind is that custom networks often have different contract setups, like UDC and OZ account addresses/class hashes. This could potentially lead to confusion for users (since, for example, they won't be able to create an account without providing custom OZ class hash). We already have a feature that lets users use a custom contract's class hash during account creation, and it might make sense to extend this to cover UDC as well. 

DoD:
  - it is possible to run cast on custom networks like so: `sncast --network integration (...)`

drknzz:
Closes #404 

## Introduced changes

- Removed `network` parameter
- Changed logic to depend on chain_id returned by node
- The `starknet_open_zeppelin_accounts.json` file format is modified to contain accounts with chain_id as the network name in case the network is custom (i.e "SN_GOERLI" -> "alpha-goerli", "KATANA" -> "KATANA")
- Wrapped unclear RPC error, when creating an account with an undeclared class hash

## Breaking changes

- Removed `network` parameter

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/cast/src/main.rs
```diff
@@ -91,15 +88,15 @@ async fn main() -> Result<()> {
...
     let account = cli.account.unwrap_or(config.account);
 
-    let provider = get_provider(&rpc_url, &network).await?;
+    let provider = get_provider(&rpc_url)?;
 
     match cli.command {
         Commands::Declare(declare) => {
             account_file_exists(&accounts_file_path)?;
-            let mut account = get_account(&account, &accounts_file_path, &provider, &network)?;
+            let chain_id = provider.chain_id().await?;
```
THenry14, line 98: Looks like chain_id is needed for each subcommand. Maybe we should move this aoutside match block?

crates/cast/src/starknet_commands/account/create.rs
```diff
@@ -87,15 +98,16 @@ pub async fn create(
     let mut items: serde_json::Value = serde_json::from_str(&contents)
         .map_err(|_| anyhow!("Failed to parse accounts file at {accounts_file_path}"))?;
 
-    let network_value = get_network(network)?.get_value();
+    let network_name = chain_id_to_network_name(chain_id);
+    println!("Detected chain_id: {chain_id:#x}");
```
THenry14, line 102: Perhaps this is something, that should be shown globally (and not only for account create subcommand)?

I know we talked about it, but right now I'm also wondering if we should include this info at all - maybe it should rather be available through https://github.com/foundry-rs/starknet-foundry/issues/278 only ? 

what do you think?

crates/cast/src/starknet_commands/account/create.rs
```diff
@@ -87,15 +98,16 @@ pub async fn create(
...
 
-    if !items[network_value][&name].is_null() {
+    if !items[&network_name][&name].is_null() {
         return Err(anyhow!(
-            "Account with provided name already exists in this network"
+            "Account with provided name already exists in network {network_name}"
```
THenry14, line 106: I wonder if we should also map starknet_goerli -> testnet (other nets as well). Perhaps people are more familiar with these networks names, and not their ids?

crates/cast/src/lib.rs
```diff
@@ -426,25 +392,6 @@ mod tests {
...
-    fn test_get_network_invalid() {
-        let net = get_network("mariusz").unwrap_err();
-        assert!(net
-            .to_string()
-            .contains("No such network mariusz! Possible values are testnet, testnet2, mainnet."));
-    }
-
     #[test]
     fn test_generate_salt() {
         let salt = extract_or_generate_salt(None);
```
THenry14, line 397: we also might use some additional tests of `chain_id_to_network_name` to actually see that some random testnames can be used?

crates/cast/tests/data/accounts/accounts.json
```diff
@@ -1,75 +1,82 @@
...
-            "public_key": "0xaaf12b4dd84b99d1fc4356eded2d21dbff5b1e1a26f28ec94640e0b55143f1",
-            "address": "0x321c02f63e477b9b25ca9ab0132fb76f30902943e37ba5ab797b3ae30b29b69"
-        },
-        "user9": {
-            "private_key": "0x5c0883893a3c32b57769f168383c53ee",
-            "public_key": "0x3a00dc633e18cadbcacf7b36b348e7c97827081318f6a08d85ddb546c38d2bf",
-            "address": "0x98459bc7e9740d1b55e5d04c3942badca974f327f7e1bb9d10737c87ea0aec"
-        }
+  "alpha-goerli": {
+    "": {
```
THenry14, line 3: what happens if key is empty?

THenry14 (commented):
overall, looks very good 🔥

crates/cast/src/main.rs
drknzz: It's async, that's why I hesitated

drknzz: I mean, it's not used everywhere and it's async, that's why I hesitated*

drknzz: `multicall new` and `call` do not need it

crates/cast/src/starknet_commands/account/create.rs
drknzz: So you think about putting some non-trivial stuff in there as well? for example `chain_id` that would be fetched from the url

crates/cast/src/starknet_commands/account/create.rs
drknzz: You mean the logic in `chain_id_to_network_name`?

```rust
    match &decoded[..] {
        "SN_GOERLI" => "alpha-goerli".into(),
        "SN_GOERLI2" => "alpha-goerli2".into(),
        "SN_MAIN" => "alpha-mainnet".into(),
        _ => decoded,
    }
```

If so, with the current approach I had in mind that it's still somewhat compatible with the original openzeppelin accounts file format. The network names are the same (`"alpha-mainnet"`, `"alpha-testnet2"`) instead of the chain_ids (`"SN_MAIN"`, `"SN_GOERLI2"`) and for custom ones it's just the decoded chain id.

crates/cast/tests/data/accounts/accounts.json
drknzz: Actually I'm not even sure where this file comes from or why have I modified it somehow 😅

crates/cast/src/starknet_commands/account/create.rs
THenry14: Yup, exactly. Basically _all the important_ things config-wise.

crates/cast/src/starknet_commands/account/create.rs
THenry14: I mean, most people (I imagine) know what is `testnet`, and may not know what is `alpha-goerli`, we do not actually get `alpha-testnet2` anywhere, right? or am I missing something?

crates/cast/tests/data/accounts/accounts.json
THenry14: the file was there, becasue we use it for tests :)

crates/cast/src/starknet_commands/account/create.rs
drknzz: No, only `alpha-goerli`, but this is how it is in the openzeppelin's file though.

THenry14: yup, but here you display it to the user in an error message, right? I suppose most won't be familiar with the oz file, and may not necessarily know why (since they use testnet), the error says 'account exists in alpha-goerli`. 

Maybe it's just a nitpick, and in reality this is a smaller issue than I potray it to be, but I can se this be a bit vague for some users.

drknzz: I see. I actually though about changing the message to something among the lines of `... already exists with network with chain_id {decoded_chain_id}`, where decoded_chain_id would be just the decoded hex-chain_id (so "SN_GOERLI" for example)
wdyt

THenry14: This could be a way to go, yes. This would also be a lot easier to maintain.

drknzz: But we still do want the 3 default nets to be named `alpha-...` in the file, right?

THenry14: yes, actually I think we must still support this, as there are already users with accounts deployed under this schema. We'll have to figure out a way to get rid of this sometime in the future, if there will be a need to do this.

CHANGELOG.md
```diff
@@ -5,6 +5,17 @@ All notable changes to this project will be documented in this file.
...
+
+### Cast
+
+#### Added
+
+- Support for custom networks - accounts created on custom networks are saved in `accounts-file` under network's chain_id
+
+#### Removed
+
+- `--network` option together with the `network` field in Scarb.toml profile
```
war-in, line 17: Shouldn't we deprecate it and then remove it? I wonder if we are scared of breaking changes or not 🤔

crates/cast/tests/data/accounts/accounts.json
war-in, line 1: Don't we want to keep the old formatting?

war-in (approved):

crates/cast/tests/data/accounts/accounts.json
drknzz: We do 💀

CHANGELOG.md
drknzz: It's early enough I'd say 👀

THenry14: I'd say let's just elaborate more in changelog, so it is clear why this change was introduced, and yolo it

THenry14 (approved):

Issue #473, opened 2023-08-18
`found duplicate packages named `build_fails v0.1.0`


eliotstock:
### Which component is your bug related to?

Forge

### Foundry Version

0.4.1

### What operating system are you using?

Linux

### What system architecture are you using?

x86

### What happened

Running `snforge` on code that doesn't compile produces this output. Note the two found locations for the package are the same.

```[ERROR] Scarb build didn't succeed:

    Updating git repository https://github.com/foundry-rs/starknet-foundry
warn: found duplicate packages named `build_fails v0.1.0 (git+https://github.com/foundry-rs/starknet-foundry.git?tag=v0.4.1)`

Found locations:
- git+https://github.com/foundry-rs/starknet-foundry.git?tag=v0.4.1
- git+https://github.com/foundry-rs/starknet-foundry.git?tag=v0.4.1

Because of this, referencing package `build_fails v0.1.0 (git+https://github.com/foundry-rs/starknet-foundry.git?tag=v0.4.1)` will fail.

warn: found duplicate packages named `build_fails v0.1.0 (git+https://github.com/foundry-rs/starknet-foundry.git?tag=v0.4.1)`

Found locations:
- git+https://github.com/foundry-rs/starknet-foundry.git?tag=v0.4.1
- git+https://github.com/foundry-rs/starknet-foundry.git?tag=v0.4.1

Because of this, referencing package `build_fails v0.1.0 (git+https://github.com/foundry-rs/starknet-foundry.git?tag=v0.4.1)` will fail.

warn: found duplicate packages named `constructor_with_params v0.1.0 (git+https://github.com/foundry-rs/starknet-foundry.git?tag=v0.4.1)`

Found locations:
- git+https://github.com/foundry-rs/starknet-foundry.git?tag=v0.4.1
- git+https://github.com/foundry-rs/starknet-foundry.git?tag=v0.4.1

Because of this, referencing package `constructor_with_params v0.1.0 (git+https://github.com/foundry-rs/starknet-foundry.git?tag=v0.4.1)` will fail.

warn: found duplicate packages named `map v0.1.0 (git+https://github.com/foundry-rs/starknet-foundry.git?tag=v0.4.1)`

Found locations:
- git+https://github.com/foundry-rs/starknet-foundry.git?tag=v0.4.1
- git+https://github.com/foundry-rs/starknet-foundry.git?tag=v0.4.1

[...]

PR #477, opened 2023-08-18
Redirect stdout and stderr from scarb build to snforge

tkumor3:
Should we always redirect scarb stdout? Or only in case of warning?

## Introduced changes

-
-

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/cast/tests/helpers/constants.rs
```diff
@@ -11,15 +11,15 @@ pub const CONTRACTS_DIR: &str = "tests/data/contracts";
...
 pub const MAP_CLASS_HASH_V1: &str =
     "0x76e94149fc55e7ad9c5fe3b9af570970ae2cf51205f8452f39753e9497fe849";
 pub const MAP_CLASS_HASH_V2: &str =
-    "0x06eb3fda9eaa7d99bb7c9e72c8d874988a5b9269f3f866773889bc5ab4ec9079";
+    "0x5e8e8e76fcdc72f519ff2697fb5a774f8a029b45f2b61f9664298ab288e4bb1";
 pub const CONTRACT_WITH_CONSTRUCTOR_CLASS_HASH_V1: &str =
     "0x02bb3d35dba2984b3d0cd0901b4e7de5411daff6bff5e072060bcfadbbd257b1";
 pub const CONTRACT_WITH_CONSTRUCTOR_CLASS_HASH_V2: &str =
-    "0x022546fa1ab93bfcd5e8b386a8282ae7d2046e6076a06916608b2e44345b2765";
+    "0x01a8f37a7de9da1336190b6d34a961d6ef881239e85431335c33fa2b95926543";
```
THenry14, line 22: what scarb version was it compiled with?

.github/workflows/ci.yml
```diff
@@ -33,6 +33,22 @@ jobs:
...
+
+      # extract branch name on pull request
+      - name: Print branch name
+        if: github.event_name == 'pull_request'
+        run: echo "BRANCH_NAME=$(echo $GITHUB_HEAD_REF)" >> $GITHUB_ENV
+
+      # print branch name
+      - name: Get branch name
+        run: echo 'The branch name is' $BRANCH_NAME
+
```
cptartur, line 51: I wonder if a more universal solution working exists, but if it works it's okay.

crates/forge/src/main.rs
```diff
@@ -63,6 +63,11 @@ fn main_execution() -> Result<()> {
             String::from_utf8_lossy(&build_output.stdout)
         )
     }
+    let scarb_stdout = String::from_utf8_lossy(&build_output.stdout);
+
+    if scarb_stdout.contains("warn:") {
+        eprintln!("{scarb_stdout}");
+    }
```
cptartur, line 70: I wonder if we should only print warnings or just redirect all stdout and stderr. With this solution, we'll risking hiding some errors in case rust output format changes.

crates/forge/tests/e2e/common/runner.rs
```diff
@@ -48,3 +49,18 @@ pub(crate) fn setup_package(package_name: &str) -> TempDir {
 
     temp
 }
+
+pub(crate) fn gen_current_branch() -> String {
+    let name: &str = "BRANCH_NAME";
+    match env::var(name) {
+        Ok(v) => v,
+        Err(_e) => {
+            let output = SnapboxCommand::new("git")
```
cptartur, line 58: `SnapboxCommand` is mostly designed for testing. Just `Command` would be fine here.

crates/forge/tests/e2e/common/runner.rs
```diff
@@ -48,3 +49,18 @@ pub(crate) fn setup_package(package_name: &str) -> TempDir {
 
     temp
 }
+
+pub(crate) fn gen_current_branch() -> String {
```
cptartur, line 53: I think some `///` comment here explaining what it does would be useful. It might not be clear for future devs that it is trying to use env var set by the github action.

crates/cast/tests/helpers/constants.rs
tkumor3: @war-in?

crates/forge/src/main.rs
tkumor3: I changed it. In the current version, Scarb inherits stdout, so we will write everything

crates/forge/tests/e2e/common/runner.rs
tkumor3: SnapboxCommand is [alias](https://github.com/foundry-rs/starknet-foundry/pull/477/files/0e9c62e8607318f8dbd088f50a10633e875e719b#diff-56dbdcd8d13ff4a2f6a482083b4e3fdadf7d8f0d6246aa8998841eb33cb826e7R5) of Command. It was import before. Should I remove this alias?

cptartur: I mean using Command from std directly. https://doc.rust-lang.org/std/process/struct.Command.html

It's a bit different structure

crates/cast/tests/helpers/constants.rs
war-in: I think it was 0.5.2

THenry14: any idea why hashes changed then? compiler/devnet/scarb didn't, right?

tkumor3: I changed the names of the contract packages,I assumed that it's the reason

.github/workflows/ci.yml
```diff
@@ -31,6 +31,22 @@ jobs:
...
+      - name: Extract branch name
+        if: github.event_name != 'pull_request'
+        shell: bash
+        run: echo "BRANCH_NAME=$(echo ${GITHUB_REF#refs/heads/})" >> $GITHUB_ENV
+        id: extract_branch
+
+      # extract branch name on pull request
+      - name: Print branch name
+        if: github.event_name == 'pull_request'
+        run: echo "BRANCH_NAME=$(echo $GITHUB_HEAD_REF)" >> $GITHUB_ENV
```
cptartur, line 44: You can verify it, but `github.ref_name` https://docs.github.com/en/actions/learn-github-actions/contexts should work without writing specific code for PRs and other events.

crates/forge/tests/e2e/common/runner.rs
```diff
@@ -48,3 +50,22 @@ pub(crate) fn setup_package(package_name: &str) -> TempDir {
 
     temp
 }
+
+/**
+ * In the context of GITHUB actions, get the source branch that triggered the workflow run.
+ * Locally returns current branch.
+ */
```
cptartur, line 57:
```suggestion

/// In the context of GITHUB actions, get the source branch that triggered the workflow run.
/// Locally returns current branch.
```

cptartur (commented):
If the outputs in e2e test changed, I think updating the snippets in the book (docs) would also make sense.

.github/workflows/ci.yml
tkumor3: https://github.com/foundry-rs/starknet-foundry/actions/runs/5903277226/job/16012867783 I tested `GITHUB_REF_NAME ` has the same definition and it doesn't get the name of the branch with processing code
Zrzut ekranu 2023-08-25 o 12 15 07
under GITHUB_SHA if value of `GITHUB_REF_NAME`

Radinyn (approved):

drknzz:
Does this PR fix #473 perhaps?

piotmag769 (commented):
Thanks for fixing the annoying test I wrote 🥇

PR #479, opened 2023-08-18
Added isolated Cheatnet testing

### LINKED ISSUE #305: Develop a way of testing the cheatnet in isolation

Radinyn:
Closes #305 

## Introduced changes

- Cheatnet can now be tested in isolation

## Checklist

- [x] Linked relevant issue
- [ ] Added relevant tests


Radinyn:
#480

war-in (approved):

MaksymilianDemitraszek:
We probably want to migrate as many tests from forge to this way of testing as it is faster. What is the plan for executing it @Arcticae @cptartur @Radinyn ? What will be the rules what should be tested with forge and what with cheatnet?

Radinyn:
I think that everything more complex than simply executing the cheatcodes (like prank_simple) test from Forge should be handled by Cheatnet

cptartur:
I think forge should have some basic tests for each cheatcode verifying if it works at all and perhaps a test for a case where error is expected to see if they get returned correctly from cheatnet. For basic cases, more tests than that are probably not needed, though I'd have to be decided on case by case basis.

crates/cheatnet/tests/README.md
```diff
@@ -0,0 +1,2 @@
+Before testing build with:
+`cd contracts && scarb build`
```
Arcticae, line 2: This should be included somewhere in the development as well. Perhaps we could include this procedure in `scripts/prepare_for_tests.sh` ?

crates/cheatnet/tests/cheatcodes/prank.rs
```diff
@@ -0,0 +1,23 @@
...
+
+#[test]
+fn prank_simple() {
+    let mut state = create_cheatnet_state();
+
+    let contract_address = deploy_contract(&mut state, "PrankChecker", vec![].as_slice());
+
+    state.start_prank(contract_address, ContractAddress::from(123_u128));
+
+    let selector = get_felt_selector_from_name("get_caller_address");
```
Arcticae, line 17: I would pass a "str" to "call_contract" for convenience

crates/cheatnet/tests/common/assert.rs
Arcticae, line 1: can we name this "assertions" ?

crates/cheatnet/tests/common/assert.rs
```diff
@@ -0,0 +1,21 @@
+#[allow(clippy::module_name_repetitions)]
+#[macro_export]
+macro_rules! assert_success {
+    ($call_contract_output:expr,$expected_data:expr) => {
+        assert!(match $call_contract_output {
```
Arcticae, line 5: this, and one below should be `assert_eq!` imho

crates/cheatnet/tests/common/mod.rs
```diff
@@ -0,0 +1,39 @@
...
+    let contracts = get_contracts_map(&artifacts_path);
+
+    let class_hash = state.declare(&contract, &contracts).unwrap();
+    state.deploy(&class_hash, calldata).unwrap()
+}
+
+pub fn get_felt_selector_from_name(name: &str) -> Felt252 {
+    let selector = get_selector_from_name(name).unwrap();
+    Felt252::from_bytes_be(&selector.to_bytes_be())
+}
```
Arcticae, line 39: This probably would be better to have in a separate module, it will be useful for more than tests

crates/cheatnet/tests/common/scarb.rs
```diff
@@ -0,0 +1,80 @@
+use anyhow::{anyhow, Context, Result};
+use camino::Utf8PathBuf;
+use cheatnet::cheatcodes::ContractArtifacts;
+use serde::Deserialize;
+use std::collections::HashMap;
+use std::fs;
+
+// TODO(#41) TAKEN FROM FORGE, REMOVE AFTER SCARB HAS BEEN MOVED TO A SEPARATE PACKAGE
```
Arcticae, line 8: Can we plan to do this soon? Please remind me at planning so that we don't have to copy code like that 😬

Arcticae: I presume that circular imports would be the problem, right?

crates/cheatnet/tests/common/state.rs
```diff
@@ -0,0 +1,25 @@
...
+fn load_predeployed_contracts() -> TempDir {
+    let tmp_dir = tempdir().expect("Failed to create a temporary directory");
+    PREDEPLOYED_CONTRACTS
+        .extract(&tmp_dir)
+        .expect("Failed to copy predeployed contracts to temporary directory");
+    tmp_dir
+}
+
+#[allow(clippy::module_name_repetitions)]
+pub fn create_cheatnet_state() -> CheatnetState {
```
Arcticae, line 18: Isn't that how we always want to instantiate `CheatnetState`? 
If that's the case, can we add the instantiation method to the struct?

crates/cheatnet/tests/contracts/target/CACHEDIR.TAG
Arcticae, line 1: Don't we want to compile those on demand instead of keeping casm/sierra in repo?

Arcticae (changes requested):

crates/cheatnet/tests/common/state.rs
Radinyn: Depends whether we want to keep automatically add those predeployed contracts, perhaps it's okay, I'm not sure which version is better for forge.

crates/cheatnet/tests/common/scarb.rs
Radinyn: Yeah I've planned to to this after this giant chain of test moving PRs

crates/cheatnet/tests/common/assert.rs
Radinyn: Not sure how to go about this if we also want to assert it's variant

crates/cheatnet/tests/common/scarb.rs
cptartur: Isn't it possible to use this code from `forge` crate instead of copying it?

crates/cheatnet/tests/common/state.rs
```diff
@@ -0,0 +1,25 @@
...
+
+static PREDEPLOYED_CONTRACTS: Dir = include_dir!("crates/cheatnet/predeployed-contracts");
+
+fn load_predeployed_contracts() -> TempDir {
+    let tmp_dir = tempdir().expect("Failed to create a temporary directory");
+    PREDEPLOYED_CONTRACTS
+        .extract(&tmp_dir)
+        .expect("Failed to copy predeployed contracts to temporary directory");
+    tmp_dir
+}
```
cptartur, line 15: If this is only used for tests, can't we just use contracts at the path provided? There's no point using `include_dir!` to include contracts in the binary if this code is only for testing.

crates/cheatnet/tests/contracts/src/lib.cairo
```diff
@@ -0,0 +1,2 @@
+mod hello_starknet;
+mod prank_checker;
```
cptartur, line 2: Newline missing

crates/cheatnet/tests/contracts/target/dev/cheatnet_testing_contracts.starknet_artifacts.json
```diff
@@ -0,0 +1 @@
+{"version":1,"contracts":[{"id":"lldvf1koh10qu","package_name":"cheatnet_testing_contracts","contract_name":"HelloStarknet","artifacts":{"sierra":"cheatnet_testing_contracts_HelloStarknet.sierra.json","casm":"cheatnet_testing_contracts_HelloStarknet.casm.json"}},{"id":"r22odu2r2tgug","package_name":"cheatnet_testing_contracts","contract_name":"PrankChecker","artifacts":{"sierra":"cheatnet_testing_contracts_PrankChecker.sierra.json","casm":"cheatnet_testing_contracts_PrankChecker.casm.json"}}]}
```
cptartur, line 1: Do we wan't to commit compiled contracts artifacts? It will be quite impractical to maintain.

crates/forge/src/cheatcodes_hint_processor.rs
```diff
@@ -415,7 +416,15 @@ fn execute_syscall(
...
 
     let gas_counter = buffer.next_usize().unwrap();
+
     let contract_address = buffer.next_felt252().unwrap().into_owned();
+    let contract_address = ContractAddress(
+        PatriciaKey::try_from(
+            StarkFelt::new(contract_address.to_be_bytes()).expect("Felt conversion failed"),
+        )
+        .expect("PatriciaKey failed"),
+    );
```
cptartur, line 426: I think it would make sense to introduce helpers handling this conversions. This can be done in separate PR.

crates/cheatnet/tests/cheatcodes/prank.rs
Radinyn: That would require forge to do a reverse transformation from the felt selector they have back to the string.

crates/cheatnet/tests/contracts/target/dev/cheatnet_testing_contracts.starknet_artifacts.json
Radinyn: done

crates/cheatnet/tests/common/state.rs
Radinyn: done

crates/cheatnet/tests/contracts/target/CACHEDIR.TAG
Radinyn: done

crates/cheatnet/tests/common/state.rs
Radinyn: I'd keep it as is on this PR

crates/cheatnet/tests/README.md
Radinyn: Found a better solution

crates/cheatnet/tests/common/mod.rs
Radinyn: Moved to conversions

crates/forge/src/cheatcodes_hint_processor.rs
Radinyn: #489

cptartur (approved):

Arcticae (approved):

PR #482, opened 2023-08-18
Add `spoof` cheatcode

abulenok:
Closes #

## Introduced changes

- Add `start_spoof` and `stop_spoof`
- Add docs and tests

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/cheatnet/src/rpc.rs
```diff
@@ -358,6 +359,65 @@ impl CheatableSyscallHandler<'_> {
...
+        original_tx_info: &[MaybeRelocatable],
+        contract_address: &ContractAddress,
+    ) -> Relocatable {
+        // create a new segment with replaced tx info
+        let ptr_cheated_tx_info = vm.add_memory_segment();
+
+        let mut new_tx_info = original_tx_info.to_owned();
+
+        if let Some(tx_info_mock) = self.cheatcode_state.spoofed_contracts.get(contract_address) {
+            let TxInfoMock {
```
cptartur, line 374: Shouldn't we panic in case it's `None`? `get_cheated_tx_info_ptr` is only executed if contract was in fact spoofed, so I'm assuming it should be impossible for this function to execute with it being `None`.

crates/cheatnet/src/cheatcodes/spoof.rs
```diff
@@ -0,0 +1,47 @@
...
+    pub fn start_spoof(
+        &mut self,
+        contract_address: ContractAddress,
+        version: Option<Felt252>,
+        account_contract_address: Option<Felt252>,
+        max_fee: Option<Felt252>,
+        signature: Option<Vec<Felt252>>,
+        transaction_hash: Option<Felt252>,
+        chain_id: Option<Felt252>,
+        nonce: Option<Felt252>,
```
cptartur, line 26: This isn't necessarly a better approach, but what if `TxInfoMock` implemented `Default` trait, and this function accepted this struct?

This way in case we just want to change one value (e.g. in tests) we could do something along the lines of 
```rust
let mocked = TxInfoMock { version: Some(Felt252::from(1), ..TxInfoMock::default() };
start_spoof(mocked);
```

crates/forge/src/cheatcodes_hint_processor.rs
```diff
@@ -265,6 +265,55 @@ impl CairoHintProcessor<'_> {
...
+                    } else {
+                        Option::None
+                    }
+                };
+
+                let contract_address = ContractAddress(PatriciaKey::try_from(StarkFelt::new(
+                    inputs[0].clone().to_be_bytes(),
+                )?)?);
+
+                let version = make_option(inputs[1].clone().is_one(), inputs[2].clone());
```
cptartur, line 281: What could be done here is just 

```rust
let version = inputs[1].is_one().then(|| inputs[2].clone());
```

No need to constructor a helper that basically achieves the same.

crates/forge/src/cheatcodes_hint_processor.rs
```diff
@@ -265,6 +265,55 @@ impl CairoHintProcessor<'_> {
...
+                let account_contract_address =
+                    make_option(inputs[3].clone().is_one(), inputs[4].clone());
+                let max_fee = make_option(inputs[5].clone().is_one(), inputs[6].clone());
+                let transaction_hash = make_option(inputs[7].clone().is_one(), inputs[8].clone());
+                let chain_id = make_option(inputs[9].clone().is_one(), inputs[10].clone());
+                let nonce = make_option(inputs[11].clone().is_one(), inputs[12].clone());
+
+                let signature_len = inputs[14]
+                    .to_usize()
+                    .expect("Missing signature len in inputs");
```
cptartur, line 291: I think this error is not from a missing signature, but from type conversion failure.

docs/src/appendix/cheatcodes/start_spoof.md
```diff
@@ -0,0 +1,71 @@
...
+    // All other fields of `TxInfo` remain unchanged
+    let mut tx_info = TxInfoMockTrait::default();
+    tx_info.transaction_hash = Option::Some(1234);
+    
+    start_spoof(contract_address, tx_info);
+    
+    dispatcher.store_tx_hash(13);
+    
+    let tx_hash = dispatcher.get_stored_tx_hash();
+    assert(tx_hash == 1234, 'Wrong tx_hash'); // this assert passes
```
cptartur, line 69: I'd change the error message in `assert` to more descriptive instead of adding a comment after.

docs/src/appendix/cheatcodes/start_spoof.md
```diff
@@ -0,0 +1,71 @@
...
+    // ...
+    
+    // Change transaction_hash to 1234
+    // All other fields of `TxInfo` remain unchanged
+    let mut tx_info = TxInfoMockTrait::default();
+    tx_info.transaction_hash = Option::Some(1234);
+    
+    start_spoof(contract_address, tx_info);
+    
+    dispatcher.store_tx_hash(13);
```
cptartur, line 66: This doesn't take any arguments.

crates/cheatnet/tests/cheatcodes/spoof.rs
```diff
@@ -0,0 +1,309 @@
...
+use cheatnet::{
+    conversions::{
+        class_hash_to_felt, contract_address_to_felt, felt_from_short_string,
+        felt_selector_from_name,
+    },
+    rpc::call_contract,
+};
+
+#[test]
+fn spoof_simple() {
```
cptartur, line 15: We should also verify that if spoofing a one argument, other remain unchanged (not spoofed).

crates/cheatnet/src/rpc.rs
Radinyn: I think unwrapped lookup is enough.

crates/cheatnet/tests/cheatcodes/spoof.rs
Radinyn: Preferably in a different test IMO

crates/cheatnet/tests/cheatcodes/spoof.rs
```diff
@@ -0,0 +1,309 @@
...
+    let selector = felt_selector_from_name("get_tx_hash_and_emit_event");
+
+    let output =
+        call_contract(&contract_address, &selector, vec![].as_slice(), &mut state).unwrap();
+
+    assert_success!(output, vec![Felt252::from(123)]);
+}
+
+#[test]
+#[ignore = "TODO(#254)"]
```
Arcticae, line 184: This issue is now closed

snforge_std/src/cheatcodes.cairo
```diff
@@ -39,6 +39,42 @@ impl RevertedTransactionImpl of RevertedTransactionTrait {
     }
 }
 
+#[derive(Drop)]
+struct TxInfoMock {
+    version: Option<felt252>,
+    account_contract_address: Option<felt252>,
```
Arcticae, line 45: This should be `ContractAddress` for this to align with `TxInfo`

snforge_std/src/cheatcodes.cairo
```diff
@@ -39,6 +39,42 @@ impl RevertedTransactionImpl of RevertedTransactionTrait {
...
+    max_fee: Option<u128>,
+    signature: Option<Array<felt252>>,
```
Arcticae, line 47: And this should be a `Span`

snforge_std/src/cheatcodes.cairo
```diff
@@ -39,6 +39,42 @@ impl RevertedTransactionImpl of RevertedTransactionTrait {
     }
 }
 
+#[derive(Drop)]
```
Arcticae, line 42: It should derive `Copy` as well for users convenience

snforge_std/src/cheatcodes.cairo
```diff
@@ -39,6 +39,42 @@ impl RevertedTransactionImpl of RevertedTransactionTrait {
...
+            max_fee: Option::None(()),
+            signature: Option::None(()),
+            transaction_hash: Option::None(()),
+            chain_id: Option::None(()),
+            nonce: Option::None(()),
+        }
+    }
+}
+
+fn is_set_and_value<T, impl TDrop: Drop::<T>>(option: Option<T>, default: T) -> (bool, T) {
```
Arcticae, line 71: The naming is somewhat confusing. Also, please consider moving it closer to usage

snforge_std/src/cheatcodes.cairo
```diff
@@ -179,3 +215,55 @@ fn get_class_hash(contract_address: ContractAddress) -> ClassHash {
...
+        nonce
+    } = tx_info_mock;
+
+    let (is_version_set, version) = is_set_and_value(version, 0);
+    let (is_acc_address_set, account_contract_address) = is_set_and_value(account_contract_address, 0);
+    let (is_max_fee_set, max_fee) = is_set_and_value(max_fee, 0_u128);
+    let (is_tx_hash_set, transaction_hash) = is_set_and_value(transaction_hash, 0);
+    let (is_chain_id_set, chain_id) = is_set_and_value(chain_id, 0);
+    let (is_nonce_set, nonce) = is_set_and_value(nonce, 0);
+    let (is_signature_set, signature) = is_set_and_value(signature, ArrayTrait::new());
```
Arcticae, line 238: I would just do `serialize_input` (returning a tuple of felts mby) so that it's more explicit about what actually it does

docs/src/appendix/cheatcodes/start_spoof.md
```diff
@@ -0,0 +1,71 @@
+# `start_spoof`
+
+> `fn start_spoof(contract_address: ContractAddress, tx_info_mock: TxInfoMock)`
+Changes `TxInfo` returned by `get_tx_info()` for the targeted contract until the spoof is stopped
+with [stop_spoof](./stop_spoof.md).
+
+- `contract_address` address of the contract for which `get_tx_info()` result will be mocked.
+- `TxInfoMock` - a struct with same structure as `TxInfo` (returned by `get_tx_info()`), 
+
+To mock the field of `TxInfo`, set the corresponding field of `TxInfoMock` to `Some(mocked_value)`. Setting the field to `None` will use a default value - the field will not be mocked. Using `None` will also cancel current mock for that field. See below for practical example.
```
Arcticae, line 10: We should add an example about using `None` as a cancel of the spoof

crates/forge/tests/integration/spoof.rs
```diff
@@ -0,0 +1,240 @@
...
+                );
+                let version = dispatcher.get_version();
+                assert(version == version_before_mock, 'Invalid version');
+                let max_fee = dispatcher.get_max_fee();
+                assert(max_fee == max_fee_before_mock, 'Invalid max_fee');
+                let chain_id = dispatcher.get_chain_id();
+                assert(chain_id == chain_id_before_mock, 'Invalid chain_id');
+                let signature = dispatcher.get_signature();
+                assert(signature.len() == signature_before_mock.len(), 'Invalid signature');
+
```
Arcticae, line 67: We should compare signature as well as its' length

crates/forge/tests/integration/spoof.rs
```diff
@@ -0,0 +1,240 @@
+use crate::integration::common::runner::Contract;
```
Arcticae, line 1: We should add a case where we cancel some entries with a `None` (and a `.default()` object)

Arcticae (changes requested):

crates/forge/tests/integration/spoof.rs
abulenok: Added in https://github.com/foundry-rs/starknet-foundry/pull/482/commits/a0f058d80db7f7a297faa8e2ece8bd8b39b61cb2

crates/cheatnet/src/cheatcodes/spoof.rs
abulenok: In the spirit of https://github.com/foundry-rs/starknet-foundry/issues/506 `Cheatnet` `start_mock` accepts individual fields instead of the `TxInfoMock` structure.  So in the current implementation, I don't see clear benefits of using `TxInfoMock::default()`

docs/src/appendix/cheatcodes/start_spoof.md
abulenok: added in https://github.com/foundry-rs/starknet-foundry/pull/482/commits/17bb23850739c6f871a542eae54947828c146bb0

crates/cheatnet/src/cheatcodes/spoof.rs
cptartur: To me the benefit is mostly readability in tests + not having to pass x number of arguments if only one changes. But if you think your approach is better, we can keep it as it is.

Arcticae (approved):
Good job m8

Radinyn (approved):
Looks great! BTW we should consider splitting `cheatcodes.cairo` into multiple files as it starts to grow pretty big.

Arcticae:
@Radinyn https://github.com/foundry-rs/starknet-foundry/issues/561

PR #484, opened 2023-08-18
Workspaces support

### LINKED ISSUE #250: Support scarb workspace package selection

Forge

### Feature Request

We will be pushing workspaces support to _Scarb_ in the near future. We would love to see `starknet-foundry` provide a coherent behaviour inside a Scarb workspace consisting of multiple packages. 

Current behaviour: 
`snforge` tries to run for all workspace members. 

Expected behaviour:
- When run inside a package, run for this package only (in particular, when run in a workspace root package, run for this package only and none of the workspace members below in the directory tree).
- When run inside a virtual workspace root, run for all workspace members.
- Provide a cli interface to select a package (`--package / --workspace` arguments or similar).

Glossary: 
root package - package in the workspace root, i.e. one denoted by a manifest file with both `[workspace]` and `[package]` sections 
virtual workspace - a workspace without a root package

Motivation:
The suggested behaviour is consistent with how  commands interact with _Cargo_ workspaces the _Scarb_ implementation is inspired on. 
It also plays well with  `scarb test` command implementation, which needs to be able to run `snforge` tests for a single package only.

Example implementation: 
[`scarb-cairo-test`](https://github.com/software-mansion/scarb/blob/be0e6e924e33644faa709fcc120d1777d6af7c0e/extensions/scarb-cairo-test/src/main.rs#L50) extension uses the [`PackagesFilter`](https://docs.rs/scarb-metadata/latest/scarb_metadata/packages_filter/struct.PackagesFilter.html) from `scarb-metadata` crate which we intend to understand workspace package selection as of https://github.com/software-mansion/scarb/pull/465 .

bartekryba:
## Introduced changes

- Added support for scarb workspaces
   - test specific package by running `snforge` in package dir or explicitly adding `--package package_name` flag 
   - test all packages with `--workspace` flag

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests

Issues that arised: #674 #675 #671


crates/forge/src/lib.rs
```diff
@@ -72,6 +74,48 @@ struct TestsFromFile {
...
+            Box::new(|package| {
+                package.manifest_path == metadata.workspace.manifest_path || is_virtual_workspace
+            })
+        }
+    };
+
+    metadata
+        .packages
+        .iter()
+        .filter(|package| workspace_members.contains(&package.id))
```
piotmag769, line 113: Cannot we just filter `workspace_members`? We return ids only anyways

crates/forge/src/main.rs
```diff
@@ -29,6 +29,14 @@ struct Args {
...
     #[arg(short = 'x', long)]
     exit_first: bool,
+
+    /// Run tests for specific package
+    #[arg(short = 'p', long)]
+    package: Option<String>,
+
+    /// Run tests for all packages in the workspace
+    #[arg(short, long)]
+    workspace: bool,
```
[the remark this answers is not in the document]
bartekryba: That was my first idea too, but then I checked how it works in cargo. If both args are passed it's accepted, but the worspace flag takes precedence

crates/forge/src/lib.rs
bartekryba: We use metedata of packages in the filtering function, so we need additional fields, which aren't present in the `workspace_members` vector.

crates/forge/src/lib.rs
```diff
@@ -99,11 +143,15 @@ fn find_cairo_root_files_in_directory(
     let src_path = lib_path.parent().with_context(|| {
         format!("Failed to get parent directory of a package at path = {lib_path}")
     })?;
+    let crates_path = src_path
+        .parent()
+        .with_context(|| format!("Failed to get root directory of a package at path = {lib_path}"))?
+        .join(Utf8PathBuf::from("crates"));
```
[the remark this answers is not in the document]
bartekryba: Good catch. Do you think ignoring subtrees that have a Scarb.toml file in root should be enough?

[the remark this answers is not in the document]
piotmag769: I think it makes sense, but let's ask @mkaput

crates/forge/src/lib.rs
piotmag769: Right, my bad

crates/forge/tests/e2e/workspaces.rs
```diff
@@ -0,0 +1,306 @@
...
+        Failure data:
+            original value: [8111420071579136082810415440747], converted to a string: [failing check]
+        
+        [SKIP] test_failing::test_another_failing
+        Tests: 1 passed, 1 failed, 1 skipped
+        "#});
+}
+
+#[test]
+fn run_missing_package() {
```
piotmag769, line 231: I think it should result in an error message, that's how Cargo works

crates/forge/tests/e2e/workspaces.rs
piotmag769, line 1: Can we duplicate the tests from workspace with root package for the virtual workspace?

crates/forge/src/lib.rs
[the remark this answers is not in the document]
mkaput: If you run `scarb metadata` over a workspace, it should yield you precise paths of all workspace members. You do not have to scarb fs manually.

crates/cast/Cargo.toml
```diff
@@ -15,11 +15,11 @@ starknet = "0.4.0"
 tokio = { version = "1.28.2", features = ["full"] }
 url = "2.2.2"
 rand = "0.8.5"
-scarb-metadata = "1.4.2"
 primitive-types = "0.12.1"
 shellexpand = "3.1.0"
 toml = "0.5.2"
 project-root = "0.2.2"
+scarb-metadata_1_4_2 = { git = "https://github.com/software-mansion/scarb.git", rev="4b87413", package = "scarb-metadata" }
```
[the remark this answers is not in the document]
cptartur: Any info on this? @THenry14

crates/forge/src/lib.rs
```diff
@@ -72,6 +75,48 @@ struct TestsFromFile {
...
+            Box::new(|package| {
+                package.manifest_path == metadata.workspace.manifest_path || is_virtual_workspace
+            })
+        }
+    };
+
+    metadata
+        .packages
+        .iter()
+        .filter(|package| workspace_members.contains(&package.id))
```
cptartur, line 114: Why do we need this filter?

crates/forge/src/lib.rs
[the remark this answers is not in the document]
cptartur: Are we changing this then?

crates/forge/src/main.rs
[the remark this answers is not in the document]
cptartur: Perhaps we shouldn't do it like cargo and an error should be returned when both flags are passed?

crates/forge/src/main.rs
```diff
@@ -66,7 +76,22 @@ fn main_execution() -> Result<()> {
...
     }
 
-    for package in &scarb_metadata.workspace.members {
+    let manifest_path = Command::new("scarb")
+        .current_dir(current_dir)
+        .arg("manifest-path")
+        .output()
+        .context("Failed to fetch manifest-path")?
+        .stdout;
+    let manifest_path = PathBuf::from(String::from_utf8(manifest_path)?.trim_end());
```
cptartur, line 85: Shouldn't it be possible to get this value from scarb metadata as well?

crates/forge/tests/e2e/workspaces.rs
```diff
@@ -0,0 +1,306 @@
...
+
+    snapbox
+        .current_dir(&temp)
+        .assert()
+        .success()
+        .stdout_matches(indoc! {r#"
+        [..]Compiling[..]
+        [..]Finished[..]
+        Collected 3 test(s) and 2 test file(s)
+        Running 1 test(s) from hello_workspaces package
```
cptartur, line 24: I think these outputs are a little bit confusing if we say that only some of tests are run from the xyz package, if in fact all of them are (just that some are not in the src, but in tests for example).

crates/forge/tests/e2e/workspaces.rs
```diff
@@ -0,0 +1,306 @@
...
+        Failure data:
+            original value: [8111420071579136082810415440747], converted to a string: [failing check]
+        
+        [SKIP] test_failing::test_another_failing
+        Tests: 1 passed, 1 failed, 1 skipped
+        "#});
+}
+
+#[test]
+fn run_for_entire_workspace_and_specific_package() {
```
cptartur, line 189: I think we should have an error about arguments being exclusive here. I understand that it is not how cargo works, but in this case I don't think it useful to mimic its behavior.

cc @piotmag769

docs/src/testing/running-tests.md
```diff
@@ -59,3 +59,69 @@ Failure data:
 [SKIP] package_name::yet_another_test
 Tests: 3 passed, 1 failed, 2 skipped
 ```
+
+## Running tests in Scarb Workspace
```
cptartur, line 63: I think we should link scarb docs about workspaces somewhere in this section

docs/src/testing/running-tests.md
```diff
@@ -59,3 +59,69 @@ Failure data:
...
+[PASS] test_simple::simple_case
+Tests: 4 passed, 0 failed, 0 skipped
+Running 1 test(s) from fibonacci package
+[PASS] fibonacci::tests::it_works
+Tests: 1 passed, 0 failed, 0 skipped
+Collected 3 test(s) and 2 test file(s)
+Running 1 test(s) from hello_workspaces package
+[PASS] hello_workspaces::test_simple
+Tests: 1 passed, 0 failed, 0 skipped
+```
```
cptartur, line 125: I think a separate subsection showcasing how scarb behaves in virtual workspaces could be useful as well.

docs/src/testing/running-tests.md
```diff
@@ -59,3 +59,69 @@ Failure data:
...
+Collected 4 test(s) and 3 test file(s)
+Running 1 test(s) from addition package
+[PASS] addition::tests::it_works
+Running 2 test(s) from tests/nested/test_nested.cairo
+[PASS] test_nested::test_two
+[PASS] test_nested::test_two_and_two
+Running 1 test(s) from tests/test_simple.cairo
+[PASS] test_simple::simple_case
+Tests: 4 passed, 0 failed, 0 skipped
+```
```
cptartur, line 103: This and the output below could be changed a little bit so there are less tests. It's quite hard to read with so many outputs.

crates/forge/src/lib.rs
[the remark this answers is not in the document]
bartekryba: Yes, I'm working on it

crates/forge/src/main.rs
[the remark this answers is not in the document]
bartekryba: Yeah, I can do it that way

crates/cheatnet/src/constants.rs
```diff
@@ -67,6 +67,7 @@ pub fn build_block_context() -> BlockContext {
         block_timestamp: BlockTimestamp::default(),
         sequencer_address: ContractAddress(patricia_key!(TEST_SEQUENCER_ADDRESS)),
         fee_token_address: ContractAddress(patricia_key!(TEST_ERC20_CONTRACT_ADDRESS)),
+        deprecated_fee_token_address: ContractAddress(patricia_key!(TEST_ERC20_CONTRACT_ADDRESS)),
```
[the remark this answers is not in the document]
bartekryba: That is what it's set to in Blockifier: https://github.com/starkware-libs/blockifier/blob/f206ccac6aea57b05ef1bd4894cbdb11c42b2987/crates/blockifier/src/test_utils.rs#L346

crates/forge/src/lib.rs
```diff
@@ -72,6 +75,48 @@ struct TestsFromFile {
...
+    workspace: bool,
+) -> Vec<PackageId> {
+    let workspace_members = &metadata.workspace.members;
+
+    let is_virtual_workspace = metadata
+        .packages
+        .iter()
+        .filter(|package| package.manifest_path == metadata.workspace.manifest_path)
+        .collect_vec()
+        .is_empty();
```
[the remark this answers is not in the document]
bartekryba: Exactly, although I'll check again if there's more direct way to check it in the metadata.

crates/forge/src/lib.rs
bartekryba: To limit packages to only those that are members of the workspace, excluding dependencies.

From the scarb docs of `metadata.packages` field:
```
Metadata of all packages used in this workspace, or just members of it if this is an output
of `scarb metadata --no-deps`.

In the former case, use [`WorkspaceMetadata::members`](`WorkspaceMetadata::members`) to filter workspace members.
```

crates/forge/src/main.rs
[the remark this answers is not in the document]
piotmag769: I'm fine with that approach too

crates/forge/src/lib.rs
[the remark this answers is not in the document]
piotmag769: I asked Scarb team before and I don't think there is a direct method for that

crates/forge/tests/e2e/workspaces.rs
piotmag769: You mean `from XYZ package` part? We did this change way before this PR, it is basically an alternative for `from src/lib.cairo` which is even more missleading, since we look for tests in `src` folder only using the `lib.cairo` entrypoint (and then tests are found using the module tree)

crates/forge/tests/e2e/workspaces.rs
piotmag769: I'm fine with that if it suits you better

crates/forge/src/lib.rs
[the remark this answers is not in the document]
cptartur: I think it could make sense for scarb metadata to expose this information. I'm assuming scarb is doing some checks if the workspace is virtual or not anyways.

cc @mkaput

crates/forge/src/lib.rs
cptartur: Okay, makes sense

cptartur: Should scarb metadata allow getting this information more directly? @mkaput

crates/forge/tests/e2e/workspaces.rs
cptartur: Yes, the `from XYZ package` part. The problem I have with this is this just suggest that tests outside `src` are not really part of the package which is kind of true but also it isn't.

docs/src/testing/running-tests.md
cptartur: Yeah, sorry. I meant `snforge`.

crates/forge/src/main.rs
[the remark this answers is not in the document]
mkaput: Again, we made https://github.com/software-mansion/scarb/blob/d98fca8190ff330109b414ed70634b569732a6b5/utils/scarb-ui/src/args/packages_filter.rs#L1-L231 specially for you, why not use it here?

crates/cast/Cargo.toml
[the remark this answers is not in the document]
THenry14: should be ok, as long as tests passes

crates/forge/tests/e2e/workspaces.rs
bartekryba: Still, I guess it's outside of the scope of this pr

bartekryba: What about moving package name to the previous line, and adding some kind of separator like:
```
Collected N test(s) and N test file(s) from XYZ package
Running N' inline test(s)
...
Running N" test(s) from tests/test.cairo
Tests: N passed, 0 failed, 0 skipped

Collected M test(s) and M test file(s) from ABC package
Running M' inline test(s)
...
Running M" test(s) from tests/nested/test.cairo
Tests: M passed, 0 failed, 0 skipped
```

cptartur: I think this could clear up the confusion. What do you thing @piotmag769?

crates/cast/Cargo.toml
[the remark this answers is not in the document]
piotmag769: I would really prefer to resolve this before merging

crates/forge/tests/e2e/workspaces.rs
piotmag769: Fine for me

crates/cast/Cargo.toml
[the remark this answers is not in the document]
bartekryba: Alright, I thought about resolving this in another pr, but I'll give it another try here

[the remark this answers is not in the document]
piotmag769: If it is complex we can do that in another PR, don't mind that, just remember to create an issue for that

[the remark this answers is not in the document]
THenry14: What is the problem? maybe I can help

crates/forge/src/main.rs
```diff
@@ -74,16 +79,29 @@ fn main_execution() -> Result<()> {
...
-        let forge_config = forge::scarb::config_from_scarb_for_package(&scarb_metadata, package)?;
-        let (package_path, lib_path) = paths_for_package(&scarb_metadata, package)?;
+    let packages: Vec<PackageMetadata> = args
+        .packages_filter
+        .match_many(&scarb_metadata)
+        .context("Failed to find any packages matching the specified filter")?;
+
+    for (index, package) in packages.iter().enumerate() {
+        let forge_config =
+            forge::scarb::config_from_scarb_for_package(&scarb_metadata, &package.id)?;
```
piotmag769, line 89: I would just import this one

piotmag769 (commented):
Shouldn't we document that now we can use globs to match packages as stated in `PackagesFilter` docs?

crates/forge/Cargo.toml
```diff
@@ -39,9 +39,10 @@ parity-scale-codec = "3.5.0"
 parity-scale-codec-derive = "3.1.4"
 num-bigint.workspace = true
 cairo-felt = "0.8.1"
-clap = { version = "4.2.7", features = ["derive"] }
+clap = { version = "4.3.23", features = ["derive"] }
 walkdir = "2.3.3"
-scarb-metadata = "1.4.2"
+scarb-ui = { git = "https://github.com/software-mansion/scarb.git", rev = "58cc88e" }
+scarb-metadata = { git = "https://github.com/software-mansion/scarb.git", rev = "58cc88e" }
```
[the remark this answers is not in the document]
bartekryba: Scarb isn't released on crates.io anymore, so it has to be fetched from git. And it comes with its own `scarb-metadata`dependency. So if `scarb-metadata` is fetched from crates, then there are two copies of `1.7.0` conflicting with each other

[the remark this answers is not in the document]
mkaput: https://github.com/software-mansion/scarb/issues/667

piotmag769 (approved):

crates/forge/src/scarb.rs
```diff
@@ -179,12 +179,21 @@ pub fn paths_for_package(
...
 
     Ok((package_path, lib_path))
 }
 
+pub fn target_dir_for_package(package_path: &Utf8PathBuf) -> Result<Utf8PathBuf> {
+    let scarb_metadata = MetadataCommand::new().inherit_stderr().exec()?;
+
+    let target_dir = scarb_metadata
+        .target_dir
+        .unwrap_or_else(|| package_path.join("target/dev"));
```
cptartur, line 192: We also join `dev` in `try_get_starknet_artifacts_path`. Won't this cause problems?

crates/forge/tests/e2e/common/runner.rs
```diff
@@ -51,6 +51,117 @@ pub(crate) fn setup_package(package_name: &str) -> TempDir {
     temp
 }
 
+pub(crate) fn setup_hello_workspace() -> TempDir {
```
cptartur, line 54: Why are we reconstructing whole scarb tomls here?

crates/forge/tests/e2e/workspaces.rs
cptartur, line 1: Perhaps some kind of test with `--exact` match could be useful here as well? Though I'm not sure.

docs/src/testing/running-tests.md
```diff
@@ -60,3 +60,93 @@ Failure data:
...
+
+$ snforge
+Collected 1 test(s) and 1 test file(s) from hello_workspaces package
+Running 1 inline test(s)
+[PASS] hello_workspaces::test_simple
+Tests: 1 passed, 0 failed, 0 skipped
+```
+
+To select the specific package to test, pass a `--package package_name` (or `-p package_name` for short) flag.
+You can also run `snforge` from the package directory to achieve the same effect.
```
cptartur, line 107: I know what you mean but I'm not sure if this is explicit enough.

docs/src/testing/running-tests.md
```diff
@@ -60,3 +60,93 @@ Failure data:
...
+Tests: 1 passed, 0 failed, 0 skipped
+```
+
+`--package` and `--workspace` flags are mutually exclusive, adding both of them to a `snforge` command will result in an error.
+
+### Virtual workspaces
+
+Running `snforge` command in a virtual workspace (a workspace without a root package)
+outside any package will by default run tests for all the packages. 
+Using the `--workspace` flag has the same effect.
```
cptartur, line 149:
```suggestion
It is equivalent to running `snforge` with the `--workspace` flag.
```

crates/forge/tests/e2e/common/runner.rs
piotmag769: I have to inject absolute path to `snforge_std` and I refuse to use toml_edit, it's unintuitive and has terrible API

crates/forge/src/scarb.rs
piotmag769: My mistake here

crates/forge/tests/e2e/workspaces.rs
piotmag769: Don't see a point, I left this test since Bartek added it, but it is not the logic we test here

crates/forge/src/lib.rs
```diff
@@ -101,16 +102,19 @@ fn collect_tests_from_directory(
...
     })?;
-
     for entry in WalkDir::new(package_path)
         .sort_by_file_name()
         .into_iter()
-        .filter_entry(|e| !e.path().starts_with(src_path))
+        .filter_entry(|e| {
+            !e.path().starts_with(src_path)
+                && !subpackages.iter().any(|path| e.path().starts_with(path))
+        })
```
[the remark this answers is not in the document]
piotmag769: Yup, that's the reason + we avoid collecting tests from unwanted packages. It is really only useful when in the root package of the workspace root. But actually now that I think about it I can ditch it entirely and just implement #654 here, it will help us get rid of this weird logic, wdyt?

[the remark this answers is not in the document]
piotmag769: Yup, did it like I described and it work without this weird subpackages logic

crates/forge/src/main.rs
```diff
@@ -70,36 +74,41 @@ fn main_execution() -> Result<bool> {
...
+    for package in &packages {
+        let forge_config = config_from_scarb_for_package(&scarb_metadata, &package.id)?;
+        let (package_path, lib_path) = paths_for_package(&scarb_metadata, &package.id)?;
         std::env::set_current_dir(package_path.clone())?;
 
-        let package_name = name_for_package(&scarb_metadata, package)?;
-        let dependencies = dependencies_for_package(&scarb_metadata, package)?;
-        let target_name = target_name_for_package(&scarb_metadata, package)?;
-        let corelib_path = corelib_for_package(&scarb_metadata, package)?;
+        let target_dir = target_dir_for_package(&package_path)?; // TODO(#671)
```
cptartur, line 88: Let's move the todo above the lines and include an short explanation alongside the issue number

crates/forge/src/scarb.rs
```diff
@@ -179,12 +179,21 @@ pub fn paths_for_package(
...
 
     Ok((package_path, lib_path))
 }
 
+pub fn target_dir_for_package(package_path: &Utf8PathBuf) -> Result<Utf8PathBuf> {
+    let scarb_metadata = MetadataCommand::new().inherit_stderr().exec()?;
+
+    let target_dir = scarb_metadata
+        .target_dir
+        .unwrap_or_else(|| package_path.join("target"));
```
cptartur, line 192: I wonder in which case can the `target_dir` be None in scarb metadata. Perhaps it makes more sense to just bubble the error here instead of using a default?

docs/src/testing/running-tests.md
```diff
@@ -60,3 +60,93 @@ Failure data:
...
+
+Forge supports Scarb Workspaces.
+To make sure you know how workspaces work,
+check Scarb documentation [here](https://docs.swmansion.com/scarb/docs/reference/workspaces.html).
+
+### Workspaces with root package
+
+When running `snforge` in a Scarb workspace with a root package, it will only run tests inside the root package.  
+
+I.e., for a project structure like this
```
cptartur, line 74:
```suggestion
For a project structure like this
```

cptartur (approved):

crates/forge/src/main.rs
piotmag769: It is really well explained in the issue imo, will move it up though

crates/forge/src/scarb.rs
maciektr: 👋 It is an option for backward compatibility with older scarb versions only. I do not see an issue with using a default here, although this should be workspace root, not package root.

PR #490, opened 2023-08-21
[CI] Skip irrelevant checks

### LINKED ISSUE #483: [CI] Disable checks from other crates

Cast

### Feature Request

Forge tests could be omitted on pull requests with the `Cast` label, or if the files (the hash of files maybe?) from other crates are left unchanged.

drknzz:
Closes #483 

## Introduced changes

- Detect changes in crates (cast, forge, cheatnet) and run tests only if their files have changed
- Changes outside of the crates count as global changes, resulting in all tests being ran

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


THenry14 (approved):
LGTM! let's wait for someone from forge and cheatnet though, but this is lit 🔥

.github/workflows/ci.yml
```diff
@@ -134,3 +143,26 @@ jobs:
...
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: read
+    outputs:
+      cast: ${{ steps.filter.outputs.cast }}
+      cheatnet: ${{ steps.filter.outputs.cheatnet }}
+      forge: ${{ steps.filter.outputs.forge }}
+      global: ${{ steps.filter.outputs.global }}
+    steps:
+    - uses: dorny/paths-filter@v2
```
cptartur, line 157: I think we can fix a specific commit here instead of just `@v2`.

cptartur (commented):
I wonder if we should be always running all checks on master branch (and perhaps in the merge queue)? I see the benefit of only checking related changes in the PR itself, but I worry that always running checks conditional we'll make us miss some errors.

drknzz:
> I wonder if we should be always running all checks on master branch (and perhaps in the merge queue)? I see the benefit of only checking related changes in the PR itself, but I worry that always running checks conditional we'll make us miss some errors.

Unless somehow changes from one crate would impact another one (in which case it would be some giant flaw in the repo structure) it's safe to run only conditional checks, but I'm down with the idea of running the full suite on merged PRs (as it does not slow down the development process anyway).

cptartur:
> > I wonder if we should be always running all checks on master branch (and perhaps in the merge queue)? I see the benefit of only checking related changes in the PR itself, but I worry that always running checks conditional we'll make us miss some errors.
> 
> Unless somehow changes from one crate would impact another one (in which case it would be some giant flaw in the repo structure) it's safe to run only conditional checks, but I'm down with the idea of running the full suite on merged PRs (as it does not slow down the development process anyway).

The split is cleared with cast and forge, but in case of cheatnet there very much can be a case where some part of the code gets changed there and it starts affecting the behavior of forge as well.

drknzz:
> very much can be a case

Why so?
What kind of changes to cheatnet crate would affect forge crate?

drknzz:
Maybe we should just leave the cast crate separation from other crates then?

THenry14:
It may be a good idea to run all the checks for master, as Artur suggested :+1: this shouldn't affect anyone, and could provide us with additional layer of 'security' - compatibility wise. It also shouldn't be too hard to accomplish.

Arcticae:
I would also run all checks on merge queue as well as on master, so we don't have to revert the merges

cptartur:
Wonder if this should still be included considering that #505 introduces a considerate speedup anyway.

PR #503, opened 2023-08-22
Use cheatable entrypoint in constructors

### LINKED ISSUE #254: Support cheatable state in contract constructors

Forge

### Feature Request

We should allow mocking the constructor block info (and not only) state. 
This should include `roll`, `warp` etc. 
The process will include breaking up the `deploy` into smaller functions and using lower-level ones instead of transaction-level blockifier logic, to allow modifying execution context.

Arcticae:
Closes #254 

## Introduced changes

- Use `call_contract` instead of invoking transaction in deploy
- Catch all deploy syscalls, so that cheating should work for all depths
- Clone deploy logic from blockifier + redirect constructor execution to our cheatable execution

## Checklist

- [x] Add test for checking deploy cheating for call depth > 1
- [x] Linked relevant issue
- [x] Added relevant tests


crates/cheatnet/tests/cheatcodes/prank.rs
```diff
@@ -43,7 +43,6 @@ fn prank_with_other_syscall() {
 }
 
 #[test]
-#[ignore = "TODO(#254)"]
```
Radinyn, line 46: There are more constructor tests

crates/cheatnet/src/constants.rs
```diff
@@ -112,16 +113,30 @@ pub fn build_invoke_transaction(
...
     ContractClassV0::try_from_json_string(&raw_contract_class).unwrap()
 }
 
+fn load_v1_contract_class(
+    predeployed_contracts: &Utf8PathBuf,
+    contract_path: &str,
+) -> ContractClassV1 {
+    let raw_contract_class = read_predeployed_contract_file(predeployed_contracts, contract_path);
+    ContractClassV1::try_from_json_string(&raw_contract_class).unwrap()
+}
```
war-in, line 138: Those two can be combined in one using generics

crates/cheatnet/src/rpc.rs
```diff
@@ -649,6 +653,128 @@ pub fn call_contract_syscall(
...
+
+    execute_call_entry_point(
+        &mut constructor_call,
+        state,
+        cheatcode_state,
+        resources,
+        context,
+    )
+}
+
```
war-in, line 777: Is it entirely copied?

docs/src/testing/using-cheatcodes.md
```diff
@@ -167,3 +167,33 @@ Failure data:
...
+    
+    // Precalculate the address to obtain the contract address before the constructor call (deploy) itself
+    let contract_address = contract.precalculate_address(constructor_arguments); 
+    
+    // Change the caller address to 123 before the call to contract.deploy
+    start_prank(contract_address, 123.try_into().unwrap());
+    
+    // The constructor will have 123 set as the caller address 
+    contract.deploy(constructor_arguments).unwrap();
+}
```
war-in, line 198: I'm not sure if we should show such behaviour in our tests. I think calling `get_caller_address` inside the constructor is not a good practice because contracts are often deployed with UDC

crates/cheatnet/src/rpc.rs
Arcticae: Good point. I'll mark the modified call

docs/src/testing/using-cheatcodes.md
Arcticae: That's just an example on how to use that pattern, i don't think that's a big deal tbh

crates/cheatnet/src/rpc.rs
Arcticae: [Mark modifed region](https://github.com/foundry-rs/starknet-foundry/pull/503/commits/d35aa372f6abeebf96f69384781a41004908bf23)

crates/cheatnet/tests/cheatcodes/prank.rs
Arcticae: Added

crates/cheatnet/src/constants.rs
Arcticae: `try_from_json_string` isn't in any trait and those classes aren't in our codebase - perhaps i don't know how, but it might not be doable?

crates/cheatnet/tests/cheatcodes/prank.rs
Arcticae: BTW the mock call one was broken, so i fixed it

war-in (approved):

docs/src/testing/using-cheatcodes.md
Radinyn: I'd change it as well

Radinyn (approved):

Issue #489, opened 2023-08-21
Extract conversions to a separate crate


Radinyn:
The conversions listed in `conversions.rs` file 

https://github.com/foundry-rs/starknet-foundry/blob/621a1a5abb49fe41f459f655901cff0c915809e8/crates/cheatnet/src/conversions.rs#L8

Should be moved to a separate crate (like `cheatnet` and `forge`). 
They also should be unified to use a reference to an object, and return a new object (so that they don't `drop` the used value, for convenience). Adding more unit tests is strongly encouraged.

war-in:
there is one 
```
contract_address_from_felt252
```

Arcticae:
@war-in i think it's no longer the case and it's been changed

PR #504, opened 2023-08-22
Migrated rest of the cheatcode tests to Cheatnet

Radinyn:


crates/cheatnet/src/cheatcodes/mock_call.rs
```diff
@@ -1,15 +1,23 @@
...
 use std::collections::HashMap;
 
 impl CheatnetState {
     pub fn start_mock_call(
         &mut self,
         contract_address: ContractAddress,
-        function_name: EntryPointSelector,
-        ret_data: Vec<StarkFelt>,
+        function_name: &Felt252,
+        ret_data: &[Felt252],
```
war-in, line 15: We have to set up some guides on what we want to pass to cheatcodes

crates/cheatnet/tests/cheatcodes/get_class_hash.rs
```diff
@@ -0,0 +1,46 @@
...
+    rpc::call_contract,
+};
+
+#[test]
+fn get_class_hash_simple() {
+    let mut state = create_cheatnet_state();
+
+    let contracts = get_contracts();
+    let contract_name = felt_from_short_string("HelloStarknet");
+    let class_hash = state.declare(&contract_name, &contracts).unwrap();
```
war-in, line 13: I would extract the `declare` method which will take shortstring as an argument

war-in (approved):

crates/cheatnet/src/cheatcodes/mock_call.rs
Radinyn: #506

crates/cheatnet/tests/cheatcodes/get_class_hash.rs
Radinyn: Should be considered in #489

PR #505, opened 2023-08-22
Fix CI cache

Radinyn:


.github/workflows/ci.yml
```diff
@@ -19,13 +19,11 @@ jobs:
...
       - uses: Swatinem/rust-cache@dd05243424bd5c0e585e4b55eb2d7615cdd32f1f
-        with:
-          workspaces: starknet-foundry
       - uses: software-mansion/setup-scarb@v1.0.1
         with:
           scarb-version: ${{ env.SCARB_VERSION }}
-      - run: cargo test --lib -p forge
-      - run: cargo test integration -p forge
+      - run: cargo test --release --lib -p forge
+      - run: cargo test --release integration -p forge
```
drknzz, line 26: What does the `release` flag do in this context?
Does it speed up the test execution in exchange for compile time?
Also, does it not disable some assertion checks?

Radinyn: > Does it speed up the test execution in exchange for compile time?
Yes

The test behaviour doesn't change

war-in (approved):

.github/workflows/ci.yml
```diff
@@ -18,14 +18,12 @@ jobs:
         with:
           submodules: true
       - uses: dtolnay/rust-toolchain@stable
-      - uses: Swatinem/rust-cache@dd05243424bd5c0e585e4b55eb2d7615cdd32f1f
-        with:
-          workspaces: starknet-foundry
+      - uses: Swatinem/rust-cache@v2
```
cptartur, line 21: I'd still use specific commit instead of just `@v2`, we can bump it to most recent.

cptartur (approved):

.github/workflows/ci.yml
drknzz: What about this?
https://stackoverflow.com/a/29822963

PR #508, opened 2023-08-22
Add version check in release workflow

### LINKED ISSUE #488: Release workflow should verify if project version in `Cargo.toml` matches that from tag name

Other (Please describe below)

### Feature Request

To avoid situations like the release 0.4.0 where version wasn't bumped, some verification should be added to the workflow. The workflow should fail if the version doesn't match.

cptartur:
Closes #488 

## Introduced changes

- In release workflow, if version in `Cargo.toml` doesn't match that from the tag, the release workflow should now fail.

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -22,7 +22,7 @@ jobs:
 
       - name: Version check
         if: steps.lookupVersion.outputs.result != github.ref_name
-        run: exit 1
+        run: echo "Project version in Cargo.toml does not match the tag" && exit 1
 
   build-binaries:
     name: Build ${{ matrix.target }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -9,8 +9,27 @@ permissions:
   contents: write
 
 jobs:
+  verify-version:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Get version from Cargo.toml
+        id: lookupVersion
+        uses: mikefarah/yq@master
+        with:
+          cmd: yq -oy '"v" + .workspace.package.version' 'Cargo.toml'
+
+      - name: Version check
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -9,8 +9,27 @@ permissions:
...
+        uses: mikefarah/yq@6609ed76ecb69f9d8254345292d90ea72f641715
+        with:
+          cmd: yq -oy '"v" + .workspace.package.version' 'Cargo.toml'
+
+      - name: Version check
+        if: steps.lookupVersion.outputs.result != github.ref_name
+        uses: actions/github-script@v3
+        with:
+          script: |
+            core.setFailed('Version in Cargo.toml does not match the tag')
```
THenry14, line 28: 🤯 , why not just shell exit 1? XD

cptartur: Weeeeeel, idk. Let's change it 😅

cptartur:
Here's a link to a run that fails if version doesn't match https://github.com/cptartur/starknet-foundry/actions/runs/5948973933

THenry14:
> Here's a link to a run that fails if version doesn't match https://github.com/cptartur/starknet-foundry/actions/runs/5948973933

Cool! I know it is pretty self-explanatory, but maybe we could also use some additional error messege there, instead of just `exit 1`? It is just a nitpick though.

cptartur:
> > Here's a link to a run that fails if version doesn't match https://github.com/cptartur/starknet-foundry/actions/runs/5948973933
> 
> Cool! I know it is pretty self-explanatory, but maybe we could also use some additional error messege there, instead of just `exit 1`? It is just a nitpick though.

Makes sense, added that!

THenry14 (approved):
🔥

PR #513, opened 2023-08-23
Get rid of expects

### LINKED ISSUE #352: Get rid of expects
https://github.com/foundry-rs/starknet-foundry/pull/340#discussion_r1279265345

drknzz:
Closes #352 

## Introduced changes

- 
- 

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests

### DIFF
--- a/crates/cast/src/helpers/scarb_utils.rs
+++ b/crates/cast/src/helpers/scarb_utils.rs
@@ -96,14 +96,7 @@ pub fn parse_scarb_config(
         return Ok(CastConfig::default());
     }
 
-    let metadata = scarb_metadata::MetadataCommand::new()
-        .inherit_stderr()
-        .manifest_path(manifest_path)
-        .no_deps()
-        .exec()
-        .context(
-            "Failed to read Scarb.toml manifest file, not found in current nor parent directories",
-        )?;
+    let metadata = get_scarb_metadata(&manifest_path)?;
 
     match get_package_tool_sncast(&metadata) {
         Ok(package_tool_sncast) => {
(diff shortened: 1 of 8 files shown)


crates/cast/src/starknet_commands/declare.rs
```diff
@@ -116,15 +111,12 @@ pub async fn declare(
...
-            return Err(anyhow!(
-                "Failed to read starknet_artifacts.json contents: {}",
-                err
-            ))
-        }
-    };
+    let starknet_artifacts = std::fs::read_to_string(&starknet_artifacts).with_context(|| {
+        format!(
+            "Failed to read {} contents",
+            starknet_artifacts.into_os_string().into_string().unwrap()
```
THenry14, line 117: not sure we should use unwrap here, what if this fails?

drknzz: Would doing this be okay?
```rust
            "Failed to read {:?} contents",
            starknet_artifacts
```

crates/cast/src/helpers/scarb_utils.rs
```diff
@@ -66,6 +66,17 @@ pub fn get_scarb_manifest() -> Result<Utf8PathBuf> {
...
 }
 
+pub fn get_scarb_metadata(manifest_path: &Utf8PathBuf) -> Result<scarb_metadata::Metadata> {
+    scarb_metadata::MetadataCommand::new()
+        .inherit_stderr()
+        .manifest_path(manifest_path)
+        .no_deps()
+        .exec()
+        .context(
+            "Failed to read Scarb.toml manifest file, not found in current nor parent directories",
```
war-in, line 76: Do we have a test for that? 🤔

crates/cast/src/starknet_commands/declare.rs
THenry14: yes, I think this is acceptable :+1:

war-in (commented):
I wonder if those contexts will be displayed as follows

```
command: ...
error: ...
```

And if so, we should include some tests

war-in (approved):

PR #522, opened 2023-08-24
adds reading from json

tkumor3:
Closes #

## Introduced changes

-
-

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


snforge_std/src/file_operations.cairo
```diff
@@ -37,13 +37,34 @@ fn parse_txt(file: @File) -> Array<felt252> {
...
+
+trait Parser<T, impl TSerde: Serde<T>> {
     fn deserialize_txt(file: @File) -> Option<T>;
+    fn deserialize_json(file: @File) -> Option<T>;
 }
 
-impl TxtParserImpl<T, impl TSerde: Serde<T>> of TxtParser<T> {
+impl ParserImpl<T, impl TSerde: Serde<T>> of Parser<T> {
     fn deserialize_txt(file: @File) -> Option<T> {
         let mut content = cheatcode::<'parse_txt'>(array![*file.path].span());
```
MaksymilianDemitraszek, line 63:
```suggestion
        let mut content = cheatcode::<'read_txt'>(array![*file.path].span());
```

snforge_std/src/file_operations.cairo
```diff
@@ -37,13 +37,34 @@ fn parse_txt(file: @File) -> Array<felt252> {
...
         Serde::<T>::deserialize(ref content)
     }
+    fn deserialize_json(file: @File) -> Option<T> {
+        let mut content = cheatcode::<'parse_json'>(array![*file.path].span());
```
MaksymilianDemitraszek, line 67:
```suggestion
        let mut content = cheatcode::<'read_json'>(array![*file.path].span());
```

crates/forge/src/cheatcodes_hint_processor/file_operations.rs
```diff
@@ -25,6 +27,43 @@ pub(super) fn parse_txt(file_path: &Felt252) -> Result<Vec<Felt252>, EnhancedHin
         })
 }
 
+pub(super) fn parse_json(file_path: &Felt252) -> Result<Vec<Felt252>, EnhancedHintError> {
```
MaksymilianDemitraszek, line 30:
```suggestion
pub(super) fn read_json(file_path: &Felt252) -> Result<Vec<Felt252>, EnhancedHintError> {
```

crates/forge/src/cheatcodes_hint_processor/file_operations.rs
```diff
@@ -59,6 +98,34 @@ mod tests {
     use num_bigint::BigUint;
     use num_traits::One;
 
+    use super::json_values_sorted_by_keys;
+
+    #[test]
+    fn test_parse_json() {
+        let string = r#"
+        {
```
MaksymilianDemitraszek, line 106: What if there are:
- arrays
- string overflowing a short string

crates/forge/src/cheatcodes_hint_processor/file_operations.rs
```diff
@@ -25,6 +27,43 @@ pub(super) fn parse_txt(file_path: &Felt252) -> Result<Vec<Felt252>, EnhancedHin
...
+    felts_in_results
+        .iter()
+        .cloned()
+        .collect::<Result<Vec<Felt252>, ()>>()
+        .map_err(|_| FileParsing {
+            path: file_path_str,
+        })
+}
+
+fn json_values_sorted_by_keys(content: &str) -> Result<Vec<String>, EnhancedHintError> {
```
MaksymilianDemitraszek, line 52: Can we unit test this one?

docs/src/appendix/forge-library/deserialize_json.md
```diff
@@ -0,0 +1,61 @@
...
+    "a": 12,
+    "d": {
+        "e": 1234
+    },
+    "c": "123"
+}
+```
+could be parsed to the following struct:
+
+```rust
```
MaksymilianDemitraszek, line 52: can we provide struct type definition instead?

docs/src/appendix/forge-library/deserialize_json.md
```diff
@@ -0,0 +1,61 @@
...
+> }` 
+
+Parses json file content and tries to deserialize it to type `T` that implements `Serde` trait.
+
+- `file` - a snapshot of an instance of the struct `File` that consists of the following fields:
+  - `path` - Cairo shortstring representing a path to a file relative to a package root.
+
+> ⚠️ **Warning**
+>
+>  JSON object is an unordered data, we had to somehow give it order. Therefore, the values in the array are sorted alphabetically by JSON key. That means that in order to decode the JSON object correctly, you will need to define attributes of the struct with types that correspond to the values of the alphabetical order of the keys of the JSON.
```
MaksymilianDemitraszek, line 15:
```suggestion
>  JSON object is an unordered data. To give it an order, the values in the array are sorted alphabetically by JSON key. That means that in order to decode the JSON object correctly, you will need to define attributes of the struct with types that correspond to the values of the alphabetical order of the keys of the JSON.
```

crates/forge/src/cheatcodes_hint_processor/file_operations.rs
```diff
@@ -25,6 +27,43 @@ pub(super) fn parse_txt(file_path: &Felt252) -> Result<Vec<Felt252>, EnhancedHin
...
+    let json: Map<String, Value> =
+        serde_json::from_str(content).map_err(|_| anyhow!("Wrong format"))?;
+
+    let data = flatten(&json);
+
+    let mut keys: Vec<String> = data.keys().map(std::string::ToString::to_string).collect();
+    keys.sort_by_key(|a| a.to_lowercase());
```
MaksymilianDemitraszek, line 59: I assume those keys are in the flatten format, namely `a.b.c`. I don't think it is explained anywhere in the docs how alphanumeric ordering works for nested objects

crates/forge/src/cheatcodes_hint_processor/file_operations.rs
```diff
@@ -25,6 +27,43 @@ pub(super) fn parse_txt(file_path: &Felt252) -> Result<Vec<Felt252>, EnhancedHin
...
+
+    Ok(keys
+        .into_iter()
+        .map(|key| data.get(&key).unwrap().to_string().replace('\"', "\'"))
```
MaksymilianDemitraszek, line 63: What this line does?

MaksymilianDemitraszek (changes requested):

crates/forge/src/cheatcodes_hint_processor/file_operations.rs
tkumor3: in the case of an array `serde_json::from_str` will return an error, so I will add a test.  Protection against string overflowing has been implemented in `string_into_felt` and is tested.

snforge_std/src/file_operations.cairo
```diff
@@ -37,13 +37,34 @@ fn parse_txt(file: @File) -> Array<felt252> {
...
 }
 
-impl TxtParserImpl<T, impl TSerde: Serde<T>> of TxtParser<T> {
+impl ParserImpl<T, impl TSerde: Serde<T>> of Parser<T> {
     fn deserialize_txt(file: @File) -> Option<T> {
-        let mut content = cheatcode::<'parse_txt'>(array![*file.path].span());
+        let mut content = cheatcode::<'read_txt'>(array![*file.path].span());
+        Serde::<T>::deserialize(ref content)
+    }
+    fn deserialize_json(file: @File) -> Option<T> {
```
Arcticae, line 66: The name is not really fitting here, it loads + deserializes (deserialization seems a bit extra since it's only a call to serde, but i can live with that)

Arcticae (changes requested):

docs/src/appendix/forge-library/parse_json.md
```diff
@@ -0,0 +1,42 @@
...
+    "d": {
+        "e": 1234
+    },
+    "c": "123"
+}
+```
+will be parsed to the following array:
+> ⚠️ **Warning**
+>
+>  JSON object is an unordered data, we had to somehow give it order. Therefore, the values in the array are sorted alphabetically by JSON key
```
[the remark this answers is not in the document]
tkumor3: I copied this solution from [solidity foundry](https://book.getfoundry.sh/cheatcodes/parse-json?highlight=json#decoding-json-objects-into-solidity-structs).  I'm not convinced to backward solution, because we can't ensure that order in the JSON wouldn't change in some place

CHANGELOG.md
```diff
@@ -15,6 +15,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `l1_handler_execute` cheatcode
 - support for `roll`ing/`warp`ing/`prank`ing the constructor logic (precalculate address, prank, assert pranked state in constructor)
 - `spy_events` cheatcode
+- Functions `read_json` and `Parser<T>::deserialize_json` to load data from json files and serialize it
+
+#### Changed
+
+- rename `TxtParser` trait to `Parser`
```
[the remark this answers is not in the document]
tkumor3: IMO FileDeserializer would be the best trait name and it could implement two methods `json` and `txt`

crates/forge/src/cheatcodes_hint_processor/file_operations.rs
```diff
@@ -25,6 +27,43 @@ pub(super) fn parse_txt(file_path: &Felt252) -> Result<Vec<Felt252>, EnhancedHin
...
+    let file_path_str = as_cairo_short_string(file_path)
+        .with_context(|| format!("Failed to convert {file_path} to str"))?;
+    let content = std::fs::read_to_string(&file_path_str)?;
+    let split_content = json_values_sorted_by_keys(&content).map_err(|_| FileParsing {
+        path: file_path_str.clone(),
+    })?;
+
+    let felts_in_results: Vec<Result<Felt252, ()>> = split_content
+        .iter()
+        .map(|string| string_into_felt(string))
```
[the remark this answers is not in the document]
tkumor3: Zrzut ekranu 2023-08-28 o 11 28 19
Any idea how to fix it?

crates/forge/src/cheatcodes_hint_processor/file_operations.rs
```diff
@@ -1,9 +1,11 @@
...
+use anyhow::{anyhow, Context};
 use cairo_felt::Felt252;
 use cairo_lang_runner::short_string::as_cairo_short_string;
 use cheatnet::cheatcodes::EnhancedHintError;
 use cheatnet::cheatcodes::EnhancedHintError::FileParsing;
+use flatten_serde_json::flatten;
 use num_bigint::BigUint;
+use serde_json::{Map, Value};
 
 pub(super) fn parse_txt(file_path: &Felt252) -> Result<Vec<Felt252>, EnhancedHintError> {
```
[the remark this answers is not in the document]
tkumor3: What do you think about create trait `FileSerializer` with 2 methods `json`, `txt`

CHANGELOG.md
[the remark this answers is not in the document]
Arcticae: But it does 2 things, reads & deserializes

crates/forge/src/cheatcodes_hint_processor/file_operations.rs
tkumor3: `string_into_felt` demanding string starting and finishing with a single quote. I talked with @piotmag769 and I will change it, so I will remove this replacement.

docs/src/appendix/forge-library/deserialize_json.md
```diff
@@ -0,0 +1,61 @@
+# `deserialize_json`
```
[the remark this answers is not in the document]
tkumor3: we can document this, and link to cairo because Serde doesn't support it https://book.cairo-lang.org/appendix-03-derivable-traits.html?highlight=serde#store

crates/forge/src/cheatcodes_hint_processor/file_operations.rs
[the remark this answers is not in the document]
Arcticae: Not really, if you don't know how to fix leave as-is

docs/src/appendix/forge-library/deserialize_json.md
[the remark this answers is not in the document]
Arcticae: Fine by me, although wouldn't that be doable via a separate method? 
Something like, `load_multiple`
Up to discussion i guess, if you could make an issue it'd be great.

docs/src/appendix/forge-library/deserialize_json.md
tkumor3: added struct but also left result, what do you think about it?

crates/forge/src/cheatcodes_hint_processor/file_operations.rs
```diff
@@ -25,6 +27,43 @@ pub(super) fn parse_txt(file_path: &Felt252) -> Result<Vec<Felt252>, EnhancedHin
...
+        .map(|string| string_into_felt(string))
+        .collect();
+
+    felts_in_results
+        .iter()
+        .cloned()
+        .collect::<Result<Vec<Felt252>, ()>>()
+        .map_err(|_| FileParsing {
+            path: file_path_str,
+        })
```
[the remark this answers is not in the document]
cptartur: This can be simplified like this

crates/forge/tests/data/file_reading/data/json/invalid.json
```diff
@@ -0,0 +1 @@
+231232
```
cptartur, line 1: Newline missing here

crates/forge/tests/data/file_reading/data/json/nested_valid.json
```diff
@@ -0,0 +1,10 @@
+{
+    "b": {
+        "d": 1,
+        "c": "test",
+        "e": {
+            "a": 2
+        }
+    },
+    "a": 23
+}
```
cptartur, line 10: Newline

crates/forge/tests/data/file_reading/test.cairo
```diff
@@ -48,35 +48,78 @@ struct D {
...
+    let content = read_json(@file);
     compare_with_expected_content(content);
 }
 
+#[test]
+fn invalid_json() {
+    let file = FileTrait::new('data/json/invalid.json');
+    let content = read_json(@file);
+    assert(1 == 1, '');
+}
```
cptartur, line 105: Could use use `#[should_panic]` with a proper error message here instead?

crates/forge/tests/data/file_reading/test.cairo
```diff
@@ -48,35 +48,78 @@ struct D {
...
+#[test]
+fn json_deserialization() {
+    let file = FileTrait::new('data/json/nested_valid.json');
+    let content = Parser::<E>::deserialize_json(@file).unwrap();
+
+    let mut output_array = ArrayTrait::new();
+    let serialized = content.serialize(ref output_array);
+    assert(content.a == 23, '');
+    assert(content.b.c == 'test', '');
+    assert(content.b.e.c == 2, '');
```
cptartur, line 116: Shouldn't 2 be equal to `b.e.a`? Or are the fields of structs defined in intentionally different than these in `nested_valid.json`?

docs/src/appendix/forge-library/deserialize_json.md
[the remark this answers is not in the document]
cptartur: I think we add an note to docs for now, and open an issue before this PR is merged so we can follow up.

crates/forge/tests/data/file_reading/test.cairo
cptartur, line 1: I wonder if we should add `parse_json` tests for cases like "non_existent", "invalid_quotes" etc. so exactly the same as for `parse_txt`. Or are we assuming that the logic is shared anyways (is it?) so we don't have to duplicate these tests?

snforge_std/src/file_operations.cairo
cptartur: Previously we had `parse_txt` (so it'd be `parse_json` here) which is also not ideal, because we are also loading.

Perhaps there's some other option than parse/load/deserialize, though I don't have any idea atm.

crates/forge/tests/data/file_reading/test.cairo
tkumor3: yep I did it intentionally because the names of JSON keys has only impact on values order

docs/src/appendix/forge-library/deserialize_json.md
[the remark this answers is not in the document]
tkumor3: Ok, I read sth wrong, and we can support Array. I added it.

crates/forge/src/cheatcodes_hint_processor/file_operations.rs
tkumor3: I added note in docs

docs/src/appendix/forge-library/parse_json.md
```diff
@@ -0,0 +1,82 @@
...
+> }`
+
+Parses json file content and tries to deserialize it to type `T` that implements `Serde` trait.
+
+- `file` - a snapshot of an instance of the struct `File` that consists of the following fields:
+    - `path` - Cairo shortstring representing a path to a file relative to a package root.
+
+> ⚠️ **Warning**
+>
+>  JSON object is an unordered data. To give it an order, the values in the array are sorted alphabetically by JSON key. That means that in order to decode the JSON object correctly, you will need to define attributes of the struct with types that correspond to the values of the alphabetical order of the keys of the JSON.
```
piotmag769, line 15: The last sentence is too long and I don't really get the meaning of this part:
"to define attributes of the struct with types that correspond to the values of the alphabetical order of the keys of the JSON"

docs/src/appendix/forge-library/parse_json.md
```diff
@@ -0,0 +1,82 @@
...
+}
+```
+could be parsed to the following struct:
+
+```rust
+struct A {
+    ab: u8,
+    bc: u8,
+    cda: felt252,
+    d: B
```
piotmag769, line 58: I would name the fields the same way the keys are named in the json - it is the most common use case and a good practice. This example already shows the importance of alphabetical order without this change in fields' names

docs/src/appendix/forge-library/parse_json.md
```diff
@@ -0,0 +1,82 @@
...
+    ab: 12,
+    bc: 1,
+    cda: "123"
+    d: B {
+        e: 1234
+    }
+}
+```
+
+> 📝 **Note**
```
piotmag769, line 80: I think this info should be in a warning, it seems like a crucial info (I was not aware of that when reading tbh)

docs/src/appendix/forge-library/read_json.md
```diff
@@ -0,0 +1,47 @@
...
+    "b": 1,
+    "a": 12,
+    "d": {
+        "e": 1234
+    },
+    "c": "123"
+}
+```
+will be parsed to the following array:
+> ⚠️ **Warning**
```
piotmag769, line 38: I think it should be higher on the page, just like in `parse_json`. Please address other comments from `parse_json` here as well

piotmag769 (commented):
Can we test json with array containing both strings and numbers?

docs/src/appendix/forge-library/parse_json.md
tkumor3: I'm agree that we have to show good practice in our snippets!

docs/src/appendix/forge-library/parse_json.md
```diff
@@ -0,0 +1,42 @@
+# `parse_json`
+
+> `fn parse_json(file: @File) -> Array<felt252>`
+
+Parses json file content to an array of felts.
+
+- `file` - a snapshot of an instance of the struct `File` that consists of the following fields:
+  - `path` - Cairo shortstring representing a path to a file relative to a package root.
```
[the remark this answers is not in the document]
tkumor3: path is the argument of File, that the reason why it's nested

piotmag769 (approved):

crates/forge/src/cheatcodes_hint_processor/file_operations.rs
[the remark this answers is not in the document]
cptartur: It built successfully on my computer though not sure if this has exactly the same behavior, should be checked.

crates/forge/tests/data/file_reading/test.cairo
cptartur: Okay , if serde_json doest this validation, it doesn't make sense to re-check.

crates/forge/src/cheatcodes_hint_processor/file_operations.rs
```diff
@@ -37,10 +86,9 @@ fn string_into_felt(string: &str) -> Result<Felt252, ()> {
         let length = string.len();
         let first_char = string.chars().next();
         let last_char = string.chars().nth(length - 1);
-
         if length >= 2
             && length - 2 <= 31
-            && first_char == Some('\'')
+            && (first_char == Some('\'') || first_char == Some('\"'))
```
cptartur, line 91: What's the purpose of this check?

crates/forge/src/cheatcodes_hint_processor/file_operations.rs
```diff
@@ -59,6 +107,53 @@ mod tests {
...
+
+        let string = r#"
+        {
+            "ad": "string",
+            "test": ["1",2,"3",4]
+        }"#
+        .to_owned();
+        let result = json_values_sorted_by_keys(&string).unwrap();
+        let expected_result = ["\"string\"", "4", "\"1\"", "2", "\"3\"", "4"];
+        assert_eq!(result, expected_result);
```
cptartur, line 138: How should this sorting behave if we had an array like `["1", 2, "2", "3", 4]`?

crates/forge/src/cheatcodes_hint_processor/file_operations.rs
```diff
@@ -1,11 +1,12 @@
...
 use cairo_felt::Felt252;
 use cairo_lang_runner::short_string::as_cairo_short_string;
 use cheatnet::cheatcodes::EnhancedHintError;
-use cheatnet::cheatcodes::EnhancedHintError::FileParsing;
+use flatten_serde_json::flatten;
 use num_bigint::BigUint;
+use serde_json::{Map, Value};
 
-pub(super) fn parse_txt(file_path: &Felt252) -> Result<Vec<Felt252>, EnhancedHintError> {
+pub(super) fn read_txt(file_path: &Felt252) -> Result<Vec<Felt252>, EnhancedHintError> {
```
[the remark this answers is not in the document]
tkumor3: @piotmag769 could you answer the question? tbh I did the same visablility as it was for a similar function `parse_txt`

crates/forge/src/cheatcodes_hint_processor/file_operations.rs
tkumor3: the think is, serde_json returns String as String("sth"), in plain file reading we probably have String('sth') or String("sth") (@piotmag769 could you confirm?). For strings, we remove the first and last char before mapping to Felts. So here we make sure that we remove proper chars

crates/forge/src/cheatcodes_hint_processor/file_operations.rs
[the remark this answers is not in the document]
tkumor3: I will try again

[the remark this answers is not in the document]
tkumor3: I checked it second time and it works, thx

crates/forge/src/cheatcodes_hint_processor/file_operations.rs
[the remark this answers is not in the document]
piotmag769: Why not? It is only needed by the module that is directly above the module this function is defined in
https://doc.rust-lang.org/reference/visibility-and-privacy.html#pubin-path-pubcrate-pubsuper-and-pubself

crates/forge/src/cheatcodes_hint_processor/file_operations.rs
piotmag769: Felts can be written in 2 formats:
- integers
- shortstrings

If the parsed string can be parsed to integer, we check if it in is range [0, P).
If not, we check if it is a shortstring, which obviously starts with `'`. However, JSON grammar requires `"` for strings, therefore we allow both formats, since in txt files it is cleaner to just write using shortstrings using `'`e.g. `1 'hello' 3`

docs/src/appendix/forge-library/deserialize_json.md
piotmag769: I disagree, the struct definition can be easily derived from the result + in `parse_txt.md` we also give just a result, we should stay consistent

crates/forge/src/cheatcodes_hint_processor/file_operations.rs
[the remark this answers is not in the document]
cptartur: Okay, it makes sense

crates/forge/src/cheatcodes_hint_processor/file_operations.rs
cptartur: Thanks

crates/forge/src/cheatcodes_hint_processor/file_operations.rs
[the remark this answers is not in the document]
tkumor3: sure, I've already changed in both places!

docs/src/appendix/forge-library/read_json.md
```diff
@@ -0,0 +1,46 @@
...
+{
+    "b": 1,
+    "a": 12,
+    "d": {
+        "e": 1234
+    },
+    "c": "123"
+}
+```
+will be parsed to the following array:
```
MaksymilianDemitraszek, line 42:
```suggestion
will be read to the following array:
```

MaksymilianDemitraszek (approved):

PR #523, opened 2023-08-24
Add `class_hash` to accounts file

### LINKED ISSUE #421: Include clash hash in accounts file
https://github.com/foundry-rs/starknet-foundry/pull/418#discussion_r1285748707

drknzz:
Closes #421 

## Introduced changes

-
-

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


war-in (commented):
I think we should also use saved class_hash in `account deploy` command

crates/cast/src/starknet_commands/account/create.rs
```diff
@@ -116,6 +116,11 @@ pub async fn create(
         "deployed": false,
     });
 
+    if let Some(class_hash_) = class_hash {
```
THenry14, line 119: let's also modify an Account struct in src/lib.rs

THenry14 (changes requested):

crates/cast/src/starknet_commands/account/deploy.rs
```diff
@@ -50,27 +50,33 @@ pub async fn deploy(
...
+        if let Some(class_hash_) = &class_hash {
+            class_hash_.as_str()
+        } else if let Some(class_hash_) = account
+            .get("class_hash")
+            .and_then(serde_json::Value::as_str)
+        {
+            class_hash_
+        } else {
+            OZ_CLASS_HASH
+        }
```
THenry14, line 75: it would be cool to have some tests for this :+1:

drknzz: I've tried to add some tests but didn't manage to since we use devnet.
The only test I think could be viable here would to be check if `account deploy` fails with the `undeclared class hash` error, if some arbitrary class hash is passed.

THenry14: technically we could declare different OZ contract and deploy from it, but I can see that this is maybe an overkill for what we want to test. 
>The only test I think could be viable here would to be check if account deploy fails with the undeclared class hash error, if some arbitrary class hash is passed.

maybe this is a way to go then

THenry14 (approved):
great job man 🔥

PR #525, opened 2023-08-25
Mark `blockifier` modified regions

### LINKED ISSUE #392: Document exactly what the cheated methods are changing
To make sure we are capable of keeping `Cheatnet` up to date we need to have a clear understanding of the modified foreign code.

Arcticae:
Closes #392

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests

### DIFF
--- a/crates/cheatnet/src/rpc.rs
+++ b/crates/cheatnet/src/rpc.rs
@@ -258,15 +258,17 @@ fn collect_emitted_events_from_spied_contracts(
         .collect::<Vec<Event>>()
 }
 
-// Copied over (with modifications) from blockifier/src/execution/entry_point.rs:144
+// blockifier/src/execution/entry_point.rs:180 (CallEntryPoint::execute)
 fn execute_call_entry_point(
-    entry_point: &mut CallEntryPoint,
+    entry_point: &mut CallEntryPoint, // Instead of 'self'
     state: &mut dyn State,
-    cheatcode_state: &CheatcodeState,
+    cheatcode_state: &CheatcodeState, // Added parameter
     resources: &mut ExecutionResources,
     context: &mut EntryPointExecutionContext,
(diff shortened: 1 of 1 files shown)


crates/cheatnet/src/rpc.rs
```diff
@@ -846,50 +854,76 @@ pub fn execute_constructor_entry_point(
...
+    call_to_external: bool,
+    entry_point_selector: EntryPointSelector,
+    calldata: Calldata,
+    remaining_gas: &mut u64,
+) -> SyscallResult<ReadOnlySegment> {
+    let entry_point_type = if call_to_external {
+        EntryPointType::External
+    } else {
+        EntryPointType::L1Handler
+    };
```
Arcticae, line 894: I've added this method so the code mirrors the blockifier code directly, so it can be easier to update @war-in

MaksymilianDemitraszek:
I understand the intentions but I am not a big fan of this PR, imo it encourages blind copying of the blockifier code without understanding 😞.

war-in:
> I get the intentions but I am not a big fan of this PR, imo it encourages developers to blind copy the blockifier code without understanding 😞.

I can't agree. The main advantage of this solution is that developers know which parts should be changed while bumping the cairo compiler. Without those comments, it is tough to do it efficiently

Arcticae:
> I get the intentions but I am not a big fan of this PR, imo it encourages developers to blind copy the blockifier code without understanding 😞.

Well i would be happy to upstream all of this into blockifier but that seems impossible ATM. Perhaps if we have clear definitions of what we would need, and what we're modifying internally we could one day consider upstreaming, and the comments would help with defining requirements as well.

Radinyn (approved):

war-in (approved):

PR #541, opened 2023-08-29
Forking MVP

### LINKED ISSUE #344: Starknet state forking - MVP
It should be possible to load starknet state locally and run tests on it instead of the fresh local instance

war-in:
## Introduced changes

- contracts deployed on real networks can be accessed in the tests now

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/cheatnet/src/forking/state.rs
```diff
@@ -0,0 +1,73 @@
...
+use starknet_api::state::StorageKey;
+
+#[derive(Debug)]
+pub struct ForkStateReader {
+    pub dict_state_reader: DictStateReader,
+    pub worker: Worker,
+}
+
+impl StateReader for ForkStateReader {
+    fn get_storage_at(
```
Arcticae, line 16: I think we should store the values in dictionaries (a simple cache without storing anything on disk)

Arcticae: I think we're reading right now from the dictionary, but we're not writing to it on response before returning the value

Arcticae: Or does it cache automatically after returning? I might be wrong here

war-in: Yup, it is cached automatically in the CachedState which calls the `CustomStateReader`

crates/cheatnet/tests/common/state.rs
```diff
@@ -1,8 +1,19 @@
...
-    CheatnetState::new(&predeployed_contracts)
+    CheatnetState::new(&predeployed_contracts, None)
+}
+
+#[allow(clippy::module_name_repetitions)]
+pub fn create_cheatnet_fork_state() -> CheatnetState {
+    let predeployed_contracts = Utf8PathBuf::from("predeployed-contracts");
+    CheatnetState::new(
+        &predeployed_contracts,
+        Some(("http://188.34.188.184:9545/rpc/v0.4", BlockId::Tag(Latest))),
```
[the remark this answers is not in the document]
war-in: So should I use `.env` file or maybe some other approach?

crates/cheatnet/src/state.rs
```diff
@@ -50,6 +119,15 @@ impl StateReader for DictStateReader {
         Ok(nonce)
     }
 
+    fn get_class_hash_at(&mut self, contract_address: ContractAddress) -> StateResult<ClassHash> {
```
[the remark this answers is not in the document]
war-in: They are in the correct order now

crates/cheatnet/tests/cheatcodes/mod.rs
```diff
@@ -5,3 +5,5 @@ mod precalculate_address;
 mod roll;
 mod spy_events;
 mod warp;
+
+mod forking;
```
Radinyn, line 9: 1. Forking should be moved to a different testing module or at least integrated with the starknet module as it is not a cheatcode from the Cheatnet's perspective and we should be able to run it's test separately.

Radinyn (approved):

crates/cheatnet/src/lib.rs
```diff
@@ -1,30 +1,46 @@
...
 
 impl CheatnetState {
     #[must_use]
-    pub fn new(predeployed_contracts: &Utf8PathBuf) -> Self {
+    pub fn new(predeployed_contracts: &Utf8PathBuf, rpc_config: Option<(&str, BlockId)>) -> Self {
+        let dict_state_reader = build_testing_state(predeployed_contracts);
         CheatnetState {
             cheatcode_state: CheatcodeState::new(),
-            blockifier_state: build_testing_state(predeployed_contracts),
+            blockifier_state: CachedState::new(
```
Arcticae, line 33: I still think that `StateReader` should be provided from outside rather than passing optional RPC conf

Arcticae: If you dislike that approach i suggest as well an alternative (which i kind of dislike as well, but not as much as this one):

Add another constructor method which instantiates the whole thing, but name it `from_rpc` or `new_forked` and then we can differentiate on how we instantiate those.

That said - I still prefer injecting such deps since it's cleaner and this code will actually compose more nicely with changes that will inevitably be made to it (multifork)

crates/cheatnet/src/state.rs
```diff
@@ -52,6 +83,15 @@ impl StateReader for DictStateReader {
...
```
Arcticae, line 86: This is still moved

crates/cheatnet/tests/common/state.rs
```diff
@@ -1,8 +1,24 @@
...
 pub fn create_cheatnet_state() -> CheatnetState {
     let predeployed_contracts = Utf8PathBuf::from("predeployed-contracts");
-    CheatnetState::new(&predeployed_contracts)
+    CheatnetState::new(&predeployed_contracts, None)
+}
+
+#[allow(clippy::module_name_repetitions)]
+pub fn create_cheatnet_fork_state() -> CheatnetState {
+    let predeployed_contracts = Utf8PathBuf::from("predeployed-contracts");
+    dotenv().ok();
```
Arcticae, line 16: It doesn't seem to belong in this method

.gitignore
```diff
@@ -3,3 +3,4 @@ pyproject.toml
 .DS_Store
 .tool-versions
 target
+.env
```
Arcticae, line 6: We should add info about this to developers guide

Arcticae (changes requested):
Please refrain from merging before Maks has taken a look

crates/cheatnet/tests/starknet/forking.rs
```diff
@@ -0,0 +1,59 @@
+use crate::assert_success;
```
Arcticae, line 1: We are lacking tests which check the state differences between different block heights

Arcticae:
Also - let's not close this issue with this PR, we will need to follow-up on forge's side, etc.

piotmag769 (changes requested):
What is the planned approach for testing it? Will we host our own RPC node?

Radinyn:
> What is the planned approach for testing it? Will we host our own RPC node?

We already do for testing SDK's

crates/cheatnet/src/forking/worker.rs
```diff
@@ -0,0 +1,143 @@
...
+#[derive(Debug)]
+pub struct Worker {
+    client: JsonRpcClient<HttpTransport>,
+    block_id: BlockId,
+}
+
+impl Worker {
+    #[must_use]
+    pub fn new(url: &str, block_id: BlockId) -> Self {
+        Worker {
```
MaksymilianDemitraszek, line 31: It would be probably better to keep one runtime, as afaik creating new runtime spawns a thread each time.

MaksymilianDemitraszek (dismissed):
LGTM, amazing job with scoping the MVP, we probably can progress with forge implementation rn and keep working on the optimization on the cheatnet side. 
I will finish review Monday morning as I am a bit tired rn 😄 

Aside, @piotmag769 @Radinyn and maybe @THenry14 (i think you considered it for cast) we have to be careful with testing using hosted node as it will be very problematic for contributors to replicate 😢 . We probably should consider other options first.

crates/cheatnet/src/forking/state.rs
piotmag769: Unresolving, since this looks like a major concern to me
cc @MaksymilianDemitraszek

crates/cheatnet/src/forking/worker.rs
```diff
@@ -0,0 +1,143 @@
...
+use starknet::providers::jsonrpc::HttpTransport;
+use starknet::providers::{JsonRpcClient, Provider};
+use starknet_api::core::{ClassHash, ContractAddress, Nonce};
+use starknet_api::hash::{StarkFelt, StarkHash};
+use starknet_api::state::StorageKey;
+use tokio::runtime::Runtime;
+use url::Url;
+
+#[derive(Debug)]
+pub struct Worker {
```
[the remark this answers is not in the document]
war-in: I'm not sure. We keep runtime inside it so `Worker` is ok now

crates/cheatnet/src/forking/state.rs
MaksymilianDemitraszek: > low prob that someone will want to use contract deployed at address e.g. 0x101

@piotmag769 what do you mean by that?

crates/cheatnet/src/forking/worker.rs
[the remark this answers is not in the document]
MaksymilianDemitraszek: I mean it is not a strong opinion and it is fine name but imo it is a little weird

crates/cheatnet/src/state.rs
```diff
@@ -18,6 +18,37 @@ use starknet_api::{
 };
 use std::collections::HashMap;
 
+pub struct CustomStateReader(pub Box<dyn StateReader>);
```
MaksymilianDemitraszek, line 21: StateReaderProxy/Router maybe would be a better name?

MaksymilianDemitraszek: What is the performance overhead of keeping it on the heap? I assume nothing significant but I wanted to raise it so you can think about it.

MaksymilianDemitraszek (commented):
Do we have tests wich verify that all possible state modifications work as intended?

.github/workflows/ci.yml
```diff
@@ -71,7 +71,10 @@ jobs:
       - uses: software-mansion/setup-scarb@v1.0.1
         with:
           scarb-version: ${{ env.SCARB_VERSION }}
-      - run: cargo test -p cheatnet
+      - name: Run Cheatnet tests
+        run: |
+          echo CHEATNET_RPC_URL=${{ secrets.CHEATNET_RPC_URL }} >> .env
+          cargo test -p cheatnet
```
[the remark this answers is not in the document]
cptartur: Also will this work for external PRs at all?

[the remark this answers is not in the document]
THenry14: > Also will this work for external PRs at all?
Depends on what you count as an external PR. If someone forks and runs CI on the fork, the secret is not available.
When someone gets an approval to run CI on our repo - this is where things start to get tricky. We must make sure the output is censored (I'm pretty positive it is) and that no shenanigans in the form of redirecting the secret to file and downloading it works. Pretty solid PRs before allowing the runs should suffice, but we must make sure that those are needed after every new commit (otherwise someone that got such approval could try to do the file redirect thing).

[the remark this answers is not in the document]
cptartur: I think the current flow is that only approval for the first run is needed (actually its for the first PR, but it can be changed) and then the subsequent runs just work. So this is not ideal.

Though [the docs say](https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks) "Although workflows from forks do not have access to sensitive data such as secrets, they can be an annoyance for maintainers if they are modified for abusive purposes." so it seems this will not work with PRs from forks at all. But it should be verified probably.

crates/cheatnet/tests/starknet/forking.rs
```diff
@@ -0,0 +1,213 @@
...
+use starknet::core::types::BlockId;
+use std::str::FromStr;
+
+#[test]
+fn fork_simple() {
+    let mut state = create_cheatnet_fork_state();
+
+    let contract_address = contract_address_from_felt(&Felt252::from(
+        BigUint::from_str(
+            "3216637956526895219277698311134811322769343974163380838558193911733621219342",
```
piotmag769, line 27: What address is it? Does calling `increase_balance` not mutate the state? Won't it cause the test to fail when a new block is minted?

crates/cheatnet/tests/starknet/forking.rs
```diff
@@ -0,0 +1,213 @@
...
+fn test_forking_at_block_number() {
+    let predeployed_contracts = Utf8PathBuf::from("predeployed-contracts");
+    let node_url =
+        std::env::var("CHEATNET_RPC_URL").expect("CHEATNET_RPC_URL must be set in the .env file");
+
+    let mut state_before_deploy =
+        CheatnetState::new(CustomStateReader(Box::new(ForkStateReader {
+            dict_state_reader: build_testing_state(&predeployed_contracts),
+            worker: Worker::new(&node_url, BlockId::Number(309_780)),
+        })));
```
piotmag769, line 88: Wouldn't it be better to have a constructor method that takes block number as arg?

crates/cheatnet/tests/starknet/forking.rs
```diff
@@ -0,0 +1,213 @@
...
+    assert!(match state.deploy(&class_hash, &[]) {
+        Err(CheatcodeError::Unrecoverable(EnhancedHintError::State(
+            StateError::UndeclaredClassHash(undeclared_class_hash),
+        ))) => undeclared_class_hash == class_hash,
+        _ => false,
+    });
+}
+
+#[test]
+fn test_forking_at_block_number() {
```
piotmag769, line 79: Do I understand correctly that you manually checked the block number of the block that contract with the address was included in?

crates/cheatnet/tests/starknet/forking.rs
war-in: This is the `HelloStarknet` contract address deployed by me for this specific test suite.
`increase_balance` mutates the state but only locally so it will not affect future runs

crates/cheatnet/tests/starknet/forking.rs
war-in: Do you mean the `ForkStateReader` constructor?

crates/cheatnet/tests/starknet/forking.rs
war-in: Yup

crates/cheatnet/tests/starknet/forking.rs
piotmag769: Oh true, forgot we only mutate it locally, thanks!

.github/workflows/ci.yml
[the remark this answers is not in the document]
war-in: Conclusion:
- every new commit requires approval
- amended commits too
- force-pushed branch also

We can safely keep our node address in the secrets

crates/cheatnet/src/state.rs
war-in: I don't think it will be much more heavier than the one without forking

MaksymilianDemitraszek (approved):

PR #548, opened 2023-08-29
Fix events

piotmag769:
## Breaking changes

Does not require to assert events in any specific order when looking for 2 events with same `from` and `name` but different `data` or `keys

## Checklist

- [x] Linked relevant issue
- [ ] Added relevant tests

### DIFF
--- a/docs/src/appendix/cheatcodes/spy_events.md
+++ b/docs/src/appendix/cheatcodes/spy_events.md
@@ -26,7 +26,7 @@ enum SpyOn {
 
 > ⚠️ **Warning**
 >
-> Sharing contracts to spy on between spies can result in unexpected behavior — avoid it if possible.
+> Spying on the same contract with multiple spies can result in unexpected behavior — avoid it if possible.
 
 `EventSpy` implements `EventFetcher` and `EventAssertions` traits.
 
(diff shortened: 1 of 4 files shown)


docs/src/appendix/cheatcodes/spy_events.md
```diff
@@ -24,6 +24,10 @@ enum SpyOn {
 }
 ```
 
+> ⚠️ **Warning**
+>
+> Sharing contracts to spy on between spies can result in unexpected behavior — avoid it if possible.
```
war-in, line 29:
```suggestion
> Spying on the same contract with multiple spies can result in unexpected behavior — avoid it if possible.
```
I think it is clearer that way

snforge_std/src/events.cairo
```diff
@@ -102,21 +103,10 @@ fn assert_if_emitted(ref self: EventSpy, event: Event) -> bool {
...
-                        'event was emitted from',
-                        event.from.into(),
-                        'but keys or data are different'
-                    ]
-                );
-            }
-
+          && event.from == *emitted_events.at(j).from
+          && @event.keys == emitted_events.at(j).keys
+          && @event.data == emitted_events.at(j).data {
```
war-in, line 108: I liked the old one because it was more descriptive but this way is good too.
We can consider adding a new method to the `EventAssertions` trait: `assert_names_emitted` which would check if there are events with specific names

war-in (approved):

PR #555, opened 2023-08-30
Fixed docs typo in testing smart contracts

tiagofneto:


piotmag769:
Thanks a lot! That's a good catch, I'm going to fix that in another PR, since `PreparedContract` is no longer a thing, we probably forgot to change it in the docs. Thanks once again, that would be really confusing for users

PR #559, opened 2023-08-30
Fix docs and minor stuff + bump version before release

piotmag769:
Closes #

## Introduced changes

-
-

## Checklist

- [ ] Linked relevant issue
- [ ] Added relevant tests

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -26,29 +26,35 @@ To run tests scripts, you have to install:
 Before creating a contribution, make sure your code passes the following checks
 
 ```shell
-./scripts/test_forge.sh
-./scripts/test_cast.sh
-cargo fmt --check
-cargo lint
+$ cargo test
+$ cargo fmt --check
+$ cargo lint
 ```
 
 Otherwise, it won't be possible to merge your contribution.
 
-You can also run specific set of tests, by directly running `cargo test`.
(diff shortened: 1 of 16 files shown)


Arcticae (approved):

Radinyn (approved):

CONTRIBUTING.md
```diff
@@ -26,29 +26,35 @@ To run tests scripts, you have to install:
 Before creating a contribution, make sure your code passes the following checks
 
 ```shell
-./scripts/test_forge.sh
-./scripts/test_cast.sh
```
war-in, line 30: Why it is removed? Those scripts have some nice features like downloading correct Scarb and Cairo compiler versions.

CONTRIBUTING.md
```diff
@@ -26,29 +26,35 @@ To run tests scripts, you have to install:
...
-cargo fmt --check
-cargo lint
+$ cargo test
```
THenry14, line 29: I think we should include `scripts/prepare-for-tests.sh` here, otherwise cast tests may fail :(

piotmag769: It is mentioned above

THenry14: Ahh yes, sorry didn't check what is above, only saw the changes. You're right, that should suffice.

THenry14 (approved):

PR #562, opened 2023-08-30
Allow specifying contract address while deploying

### LINKED ISSUE #543: Support deploying contract at a specified address

war-in:
Closes #543 

## Introduced changes

- cheatnet allows for predefined contract address in `deploy` cheatcode

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests

### DIFF
--- a/crates/cheatnet/src/cheatcodes/deploy.rs
+++ b/crates/cheatnet/src/cheatcodes/deploy.rs
@@ -4,7 +4,7 @@ use anyhow::Result;
 use blockifier::abi::abi_utils::selector_from_name;
 use blockifier::execution::execution_utils::{felt_to_stark_felt, stark_felt_to_felt};
 
-use blockifier::state::state_api::StateReader;
+use blockifier::state::state_api::{State, StateReader};
 use cairo_felt::Felt252;
 use cairo_vm::vm::errors::hint_errors::HintError::CustomHint;
 use starknet::core::utils::get_selector_from_name;
@@ -30,16 +30,12 @@ impl CheatnetState {
         let account_address = ContractAddress(patricia_key!(TEST_ACCOUNT_CONTRACT_ADDRESS));
         let entry_point_selector = selector_from_name("deploy_contract");
 
-        if self
-            .blockifier_state
-            .state
-            .address_to_class_hash
-            .get(&contract_address)
-            .is_some()
-        {
-            return Err(CheatcodeError::Unrecoverable(EnhancedHintError::from(
-                CustomHint(Box::from("Address is already taken")),
-            )));
+        if let Ok(class_hash) = self.blockifier_state.get_class_hash_at(contract_address) {
+            if class_hash != ClassHash::default() {
+                return Err(CheatcodeError::Unrecoverable(EnhancedHintError::from(
+                    CustomHint(Box::from("Address is already taken")),
+                )));
+            }
         }
 
         let contract_class = self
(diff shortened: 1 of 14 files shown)


Radinyn (approved):
We could also add a test where we are calling a contract at a predefined address from within another contract.

crates/cheatnet/src/cheatcodes/deploy.rs
```diff
@@ -25,15 +25,39 @@ impl CheatnetState {
...
+            }
+
+            contract_address
+        } else {
+            self.precalculate_address(class_hash, calldata)
+        };
         self.increment_deploy_salt_base();
 
         let blockifier_state: &mut CachedState<DictStateReader> = &mut self.blockifier_state;
+        if predefined_contract_address.is_some() {
```
piotmag769, line 55: Do we need the `if`? Cannot we always insert that as we already assigned to `contract_address`?

piotmag769 (approved):

crates/cheatnet/src/cheatcodes/deploy.rs
```diff
@@ -25,15 +25,39 @@ impl CheatnetState {
         &mut self,
         class_hash: &ClassHash,
         calldata: &[Felt252],
+        predefined_contract_address: Option<ContractAddress>,
```
Arcticae, line 28: I would create another method on `CheatnetState` and use composition, instead of adding a parameter here:
Public methods:
`deploy_at(address)` -> deploys or panics if address is taken
`deploy()` -> generates address for you, uses `deploy_at` afterwards

crates/cheatnet/src/cheatcodes/deploy.rs
```diff
@@ -25,15 +25,39 @@ impl CheatnetState {
...
+        let contract_address = if let Some(contract_address) = predefined_contract_address {
+            if self
+                .blockifier_state
+                .state
+                .address_to_class_hash
+                .get(&contract_address)
+                .is_some()
+            {
+                return Err(CheatcodeError::Recoverable(vec![felt_from_short_string(
+                    "Address is already taken",
```
Arcticae, line 44: This shouldn't be recoverable

Arcticae (changes requested):

crates/cheatnet/src/cheatcodes/deploy.rs
war-in: Unfortunately not. If we do so there will be an error in `call_contract` that the address is already taken

piotmag769 (approved):

crates/cheatnet/tests/cheatcodes/mod.rs
```diff
@@ -1,3 +1,4 @@
+mod deploy;
```
Radinyn, line 1: I think we should make an issue for migrating `deploy` tests to `Cheatnet`.

crates/cheatnet/src/cheatcodes/deploy.rs
```diff
@@ -70,6 +88,27 @@ impl CheatnetState {
...
+        let salt = self.get_salt();
+        let account_address = ContractAddress(patricia_key!(TEST_ACCOUNT_CONTRACT_ADDRESS));
+
+        let contract_address = calculate_contract_address(
+            salt,
+            *class_hash,
+            &Calldata(Arc::new(calldata.iter().map(felt_to_stark_felt).collect())),
+            account_address,
+        )
+        .unwrap();
```
Radinyn, line 106: We might consider using `precalculate_address` cheatcode here in case we want to add some behavior there.

crates/cheatnet/tests/cheatcodes/deploy.rs
Radinyn, line 1: It would be great  to add a test with a contract that has a hardcoded address inside that we later deploy at.

crates/cheatnet/tests/cheatcodes/mod.rs
war-in: #569

PR #565, opened 2023-08-31
<!-- Reference any GitHub issues resolved by this PR -->

tkumor3:
Closes #

## Introduced changes

-
-

## Checklist

- [ ] Linked relevant issue
- [ ] Added relevant tests

### DIFF
--- a/crates/forge/src/main.rs
+++ b/crates/forge/src/main.rs
@@ -3,7 +3,8 @@ use camino::Utf8PathBuf;
 use clap::Parser;
 use include_dir::{include_dir, Dir};
 use scarb_metadata::MetadataCommand;
-use std::path::PathBuf;
+use std;
+use std::path::{Path, PathBuf};
 use tempfile::{tempdir, TempDir};
 
 use forge::run;
@@ -20,8 +21,8 @@ static PREDEPLOYED_CONTRACTS: Dir = include_dir!("crates/cheatnet/predeployed-co
 #[derive(Parser, Debug)]
 #[command(version)]
 struct Args {
-    /// Name used to filter tests
-    test_name: Option<String>,
+    /// Name used to filter tests and to create init destination dir
+    name: Option<String>,
     /// Use exact matches for `test_filter`
     #[arg(short, long)]
     exact: bool,
@@ -42,18 +43,41 @@ fn load_predeployed_contracts() -> Result<TempDir> {
     Ok(tmp_dir)
 }
 
+pub fn copy_recursively(
+    source: impl AsRef<Path>,
+    destination: impl AsRef<Path>,
+) -> std::io::Result<()> {
+    std::fs::create_dir_all(&destination)?;
+    for entry in std::fs::read_dir(source)? {
+        let entry = entry?;
+        let filetype = entry.file_type()?;
+        if filetype.is_dir() {
+            copy_recursively(entry.path(), destination.as_ref().join(entry.file_name()))?;
+        } else {
+            std::fs::copy(entry.path(), destination.as_ref().join(entry.file_name()))?;
+        }
+    }
+    Ok(())
+}
+
 fn main_execution() -> Result<()> {
     let args = Args::parse();
-
     if args.init {
-        Command::new("git")
-            .args([
-                "clone",
-                "https://github.com/foundry-rs/starknet_forge_template.git",
-                &args.test_name.clone().unwrap(),
-            ])
-            .output()
-            .unwrap();
+        let project_name = args.name.unwrap_or("starknet_forge_template".to_string());
+        let project_path = std::env::current_dir().unwrap().join(project_name);
+
+        if project_path.exists() {
+            bail!(
+                "Destination {} already exists, new project couldn't be created",
+                &project_path.display().to_string()
+            )
+        }
+
+        let template_path = project_root::get_project_root()
+            .unwrap()
+            .join("starknet_forge_template");
+        copy_recursively(template_path, project_path.display().to_string());
+
         return Ok(());
     }
 
@@ -92,7 +116,7 @@ fn main_execution() -> Result<()> {
         let target_name = target_name_for_package(&scarb_metadata, package)?;
         let corelib_path = corelib_for_package(&scarb_metadata, package)?;
         let runner_config = RunnerConfig::new(
-            args.test_name.clone(),
+            args.name.clone(),
             args.exact,
             args.exit_first,
             &forge_config,
--- a/crates/forge/src/main.rs
+++ b/crates/forge/src/main.rs
@@ -26,6 +26,9 @@ struct Args {
     #[arg(short, long)]
     exact: bool,
 
+    #[clap(short = 'i', long)]
+    init: bool,
+
     /// Stop test execution after first failed test
     #[arg(short = 'x', long)]
     exit_first: bool,
@@ -41,6 +44,18 @@ fn load_predeployed_contracts() -> Result<TempDir> {
 
 fn main_execution() -> Result<()> {
     let args = Args::parse();
+
+    if args.init {
+        Command::new("git")
+            .args([
+                "clone",
+                "https://github.com/foundry-rs/starknet_forge_template.git",
+                &args.test_name.clone().unwrap(),
+            ])
+            .output()
+            .unwrap();
+        return Ok(());
+    }
 
     let predeployed_contracts_dir = load_predeployed_contracts()?;
     let predeployed_contracts_path: PathBuf = predeployed_contracts_dir.path().into();
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -52,6 +52,7 @@ main() {
       '--help')           set -- "$@" '-h'   ;;
       '--no-modify-path') set -- "$@" '-p'   ;;
       '--version')        set -- "$@" '-v'   ;;
+      '--init')        set -- "$@" '-i'   ;;
       *)                  set -- "$@" "$arg" ;;
     esac
   done
--- a/starknet_forge_template/README.md
+++ b/starknet_forge_template/README.md
@@ -1,0 +1,3 @@
+# Starknet Forge Template
+
+This repository is a basic project for Starknet Forge - testing tool that is a part of Starknet Foundry. 
--- a/starknet_forge_template/Scarb.toml
+++ b/starknet_forge_template/Scarb.toml
@@ -1,0 +1,16 @@
+[package]
+name = "starknet_forge_template"
+version = "0.1.0"
+
+# See more keys and their definitions at https://docs.swmansion.com/scarb/docs/reference/manifest
+
+[[target.starknet-contract]]
+sierra = true
+casm = true
+
+[dependencies]
+starknet = "2.2.0"
+snforge_std = { git = "https://github.com/foundry-rs/starknet-foundry.git", tag = "v0.5.0" }
+
+[tool.snforge]
+# exit_first = true
--- a/starknet_forge_template/.gitignore
+++ b/starknet_forge_template/.gitignore
@@ -1,0 +1,2 @@
+target
+.DS_Store
--- a/starknet_forge_template/src/lib.cairo
+++ b/starknet_forge_template/src/lib.cairo
@@ -1,0 +1,25 @@
+#[starknet::interface]
+trait IHelloStarknet<TContractState> {
+    fn increase_balance(ref self: TContractState, amount: felt252);
+    fn get_balance(self: @TContractState) -> felt252;
+}
+
+#[starknet::contract]
+mod HelloStarknet {
+    #[storage]
+    struct Storage {
+        balance: felt252, 
+    }
+
+    #[external(v0)]
+    impl HelloStarknetImpl of super::IHelloStarknet<ContractState> {
+        fn increase_balance(ref self: ContractState, amount: felt252) {
+            assert(amount != 0, 'Amount cannot be 0');
+            self.balance.write(self.balance.read() + amount);
+        }
+
+        fn get_balance(self: @ContractState) -> felt252 {
+            self.balance.read()
+        }
+    }
+}
--- a/starknet_forge_template/tests/test_contract.cairo
+++ b/starknet_forge_template/tests/test_contract.cairo
@@ -1,0 +1,48 @@
+use array::ArrayTrait;
+use result::ResultTrait;
+use option::OptionTrait;
+use traits::TryInto;
+use starknet::ContractAddress;
+use starknet::Felt252TryIntoContractAddress;
+
+use snforge_std::{declare, ContractClassTrait};
+
+use starknet_forge_template::IHelloStarknetSafeDispatcher;
+use starknet_forge_template::IHelloStarknetSafeDispatcherTrait;
+
+fn deploy_contract(name: felt252) -> ContractAddress {
+    let contract = declare(name);
+    contract.deploy(@ArrayTrait::new()).unwrap()
+}
+
+#[test]
+fn test_increase_balance() {
+    let contract_address = deploy_contract('HelloStarknet');
+
+    let safe_dispatcher = IHelloStarknetSafeDispatcher { contract_address };
+
+    let balance_before = safe_dispatcher.get_balance().unwrap();
+    assert(balance_before == 0, 'Invalid balance');
+
+    safe_dispatcher.increase_balance(42).unwrap();
+
+    let balance_after = safe_dispatcher.get_balance().unwrap();
+    assert(balance_after == 42, 'Invalid balance');
+}
+
+#[test]
+fn test_cannot_increase_balance_with_zero_value() {
+    let contract_address = deploy_contract('HelloStarknet');
+
+    let safe_dispatcher = IHelloStarknetSafeDispatcher { contract_address };
+
+    let balance_before = safe_dispatcher.get_balance().unwrap();
+    assert(balance_before == 0, 'Invalid balance');
+
+    match safe_dispatcher.increase_balance(0) {
+        Result::Ok(_) => panic_with_felt252('Should have panicked'),
+        Result::Err(panic_data) => {
+            assert(*panic_data.at(0) == 'Amount cannot be 0', *panic_data.at(0));
+        }
+    };
+}


scripts/install.sh
```diff
@@ -52,6 +52,7 @@ main() {
       '--help')           set -- "$@" '-h'   ;;
       '--no-modify-path') set -- "$@" '-p'   ;;
       '--version')        set -- "$@" '-v'   ;;
+      '--init')        set -- "$@" '-i'   ;;
```
war-in, line 55:
```suggestion
      '--version')        set -- "$@" '-v'   ;;
      '--init')           set -- "$@" '-i'   ;;
```

crates/forge/src/main.rs
```diff
@@ -20,12 +21,15 @@ static PREDEPLOYED_CONTRACTS: Dir = include_dir!("crates/cheatnet/predeployed-co
...
 struct Args {
-    /// Name used to filter tests
-    test_name: Option<String>,
+    /// Name used to filter tests and to create init destination dir
+    name: Option<String>,
     /// Use exact matches for `test_filter`
     #[arg(short, long)]
     exact: bool,
 
+    #[clap(short = 'i', long)]
```
war-in, line 30: I think we do not have to specify short. It should be done automatically

war-in (commented):
Overall good but we should add some tests

crates/forge/src/main.rs
cptartur, line 1: I think the approach of just using an `--init` flag is not ideal. What in case someone passes both `--init` and some other arguments? The way `snforge` command works kind of suggest that it also should then run tests in the newly created packages.

How about introducing a `snforge-init` as an separate command (crate) instead? 

cc @piotmag769 @MaksymilianDemitraszek

cptartur: And btw I really really don't want to introduce `snforge test` as an solution for this. My idea is that `snforge == cargo test` already.

crates/forge/src/init.rs
```diff
@@ -0,0 +1,122 @@
...
+            "the path `{path}` contains invalid PATH characters (usually `:`, `;`, or `\"`)\n\
+            It is recommended to use a different name to avoid problems."
+        ));
+    }
+    Ok(())
+}
+
+/// Inspired by scarb package name validation
+/// https://github.com/software-mansion/scarb/blob/main/scarb/src/core/package/name.rs#L57
+fn check_name(name: &str) -> Result<()> {
```
cptartur, line 54: Perhaps instead of duplicating this logic, scarb should create a package with utilites like that? We could do a PR for this. 

cc @mkaput

crates/forge/src/init.rs
```diff
@@ -0,0 +1,122 @@
...
+            }
+        }
+    }
+
+    Ok(())
+}
+
+fn replace_project_name(contents: &[u8], project_name: &str) -> Vec<u8> {
+    // SAFETY: We control these files, and we know that they are UTF-8.
+    let contents = unsafe { std::str::from_utf8_unchecked(contents) };
```
cptartur, line 35: If we know that these are utf-8 why don't we just use `from_utf8` and unwrap/rethrow the error. I don't like the unsafe block here.

crates/forge/src/init.rs
```diff
@@ -0,0 +1,122 @@
...
+
+use include_dir::{include_dir, Dir, DirEntry};
+use std::ffi::OsStr;
+use std::fs;
+use std::path::{Path, PathBuf};
+
+static TEMPLATE: Dir = include_dir!("starknet_forge_template");
+
+fn create_with_template(dir: &Dir<'_>, base_path: &PathBuf, project_name: &str) -> Result<()> {
+    fs::create_dir_all(base_path)?;
```
cptartur, line 11: Is recursively creating a project dir a safe approach? I think we could assume that the dir from which `snforge` is run already exists anyway so we only should need to create a directory with one level of nesting here.

crates/forge/src/init.rs
```diff
@@ -0,0 +1,122 @@
...
+    // warn if the path contains characters that will break `env::join_paths`
+    if std::env::join_paths(std::slice::from_ref(&OsStr::new(path))).is_err() {
+        let path = path.to_string_lossy();
+        bail!(format!(
+            "the path `{path}` contains invalid PATH characters (usually `:`, `;`, or `\"`)\n\
+            It is recommended to use a different name to avoid problems."
+        ));
+    }
+    Ok(())
+}
```
cptartur, line 50: Wouldn't using https://doc.rust-lang.org/stable/std/path/struct.Path.html#method.canonicalize  or  this and checking for errors be enough of path validation? 

I'm not sure if this kind of validation is a good/necessary approach.

crates/forge/tests/e2e/running.rs
```diff
@@ -440,6 +440,26 @@ fn exit_first_flag_takes_precedence() {
...
+        .arg("test_name")
+        .assert()
+        .success();
+
+    let paths = std::fs::read_dir(temp.path().join(Path::new("test_name"))).unwrap();
+
+    for path in paths {
+        dbg!(path.unwrap().file_name());
+    }
+    dbg!(temp.display());
```
cptartur, line 460: `dbg!` 👀

starknet_forge_template/README.md
```diff
@@ -0,0 +1,3 @@
+# Starknet Forge Template
+
+This repository is a basic project for Starknet Forge - testing tool that is a part of Starknet Foundry.
```
cptartur, line 3: This is no longer a repository

crates/forge/src/init.rs
mkaput: That's one of the reasons I suggested that you run `scarb new/init` and then patch what came out of it. Especially as @MrDenkoV is currently changing this logic 😅

mkaput: I am generally unopposed to extract `PackageName` type _somewhere_, but I think it's going to be more painful to maintain, than you relying on these commands

cptartur: My idea was that if scarb has this validation somewhere already, we could just make it a separate crate we could use. Problem with `new/init` is that we have limited control on template structure and patching the outputs is more error prone than just copying.

mkaput: > Problem with new/init is that we have limited control on template structure and patching the outputs is more error prone than just copying.

The trick is that I think the opposite 😆 Primarily, you would have to remember to upgrade this crate whenever stuff changes. On the other hand, we can basically guarantee that `scarb new/init` only produces `src/lib.cairo` with some stuff, and it's `Scarb.toml` won't grow beyond what you have currently; except we may end up adding editions support (in `[package]` section) when they will happen in Cairo. So your patching would actually boil down to appending to `Scarb.toml` and overwriting `src/lib.cairo`. We can huddle to discuss this further.

scripts/install.sh
```diff
@@ -52,6 +53,7 @@ main() {
       '--help')           set -- "$@" '-h'   ;;
       '--no-modify-path') set -- "$@" '-p'   ;;
       '--version')        set -- "$@" '-v'   ;;
+      '--init')           set -- "$@" '-i'   ;;
```
maciektr, line 56: Should this go into the installer? 🤔

crates/forge/src/main.rs
MaksymilianDemitraszek: imo `snforge test` is not bad, maybe in the future we will introduce other testing commands?

cptartur: If we do, then we can consider such change, but right now it would just be an extra annoyance.

crates/forge/src/init.rs
piotmag769: Strong +1

scripts/install.sh
piotmag769: I don't think so, we shouldn't touch this script in this PR

crates/forge/src/main.rs
cptartur: If that's the two option we have I think `--init` is better than introducing `test` already. But you can try convincing me otherwise 😅

war-in: I agree with @cptartur. Adding `test` to the snforge command will be just annoying

tkumor3: IMO both approaches are ok, but its annoying when it change

crates/forge/src/init.rs
tkumor3: copied from scarb, I can try change it

crates/forge/src/init.rs
tkumor3: I realize that I validate it double, because I also check if `characters are ASCII lowercase letters, ASCII numbers or underscore"`

starknet_forge_template/Scarb.toml
```diff
@@ -0,0 +1,16 @@
...
+
+# See more keys and their definitions at https://docs.swmansion.com/scarb/docs/reference/manifest
+
+[[target.starknet-contract]]
+sierra = true
+casm = true
+
+[dependencies]
+starknet = "2.2.0"
+snforge_std = { git = "https://github.com/foundry-rs/starknet-foundry.git", tag = "v0.5.0" }
```
piotmag769, line 13: Git tag fallacy xD It is of course desired to have this tag for users, no doubt. However it will make testing the template impossible if something changes in basic declare-deploy flow. We should probably reuse the solution we have for git dependency (@cptartur is improving it rn) and replace the tag with branch in our test instead.

Also it would be great to fill this tag with current version of forge from the main `Cargo.toml`, although not sure if it's possible together with `include_dir` (it would have to be modified before `include_dir` executes)

tkumor3: I changed how it works, so currently I use `scarb add -git .../starknet-foundry.git` to get the the lastest released version

crates/forge/src/init.rs
```diff
@@ -0,0 +1,104 @@
...
+    let mut casm = Table::new();
+    let mut contract = Table::new();
+    contract.set_implicit(true);
+
+    casm.insert("casm", Item::Value(true.into()));
+    array_of_tables.push(casm);
+    contract.insert("starknet-contract", Item::ArrayOfTables(array_of_tables));
+    doc.insert("target", Item::Table(contract));
+
+    fs::write(path, doc.to_string())?;
```
piotmag769, line 61: Yea, their API is terrible, now I understand XD

crates/forge/src/init.rs
```diff
@@ -0,0 +1,104 @@
...
+        .stdout(Stdio::inherit())
+        .output()
+        .context("Failed to initial new project")?;
+
+    Command::new("scarb")
+        .current_dir(&project_path)
+        .arg("add")
+        .arg("snforge_std")
+        .arg("--git")
+        .arg("https://github.com/foundry-rs/starknet-foundry.git")
```
piotmag769, line 84: The problem here is that we add `snforge_std` version from main branch - we should add it with a tag pointing to the version of forge from Cargo.toml. 

The approach for tests should be different though - we should overwrite this dependency in tests to point to the current branch and repository (we should duplicate the logic from git dependency test, check updates made by @cptartur)

crates/forge/src/init.rs
```diff
@@ -0,0 +1,104 @@
...
+
+static TEMPLATE: Dir = include_dir!("starknet_forge_template");
+
+fn overwrite_files_from_template(
+    dir_to_overwrite: &str,
+    base_path: &PathBuf,
+    project_name: &str,
+) -> Result<()> {
+    let copy_form_dir = TEMPLATE.get_dir(dir_to_overwrite);
+    match copy_form_dir {
```
piotmag769, line 18: Do we have a guarantee that Scarb basic template will have files only in `src` and `tests` folders so we replace them all with our files? cc @mkaput

piotmag769 (changes requested):

crates/forge/src/init.rs
```diff
@@ -0,0 +1,104 @@
...
+    fs::write(path, doc.to_string())?;
+
+    Ok(())
+}
+
+pub fn init(name: Option<String>) -> Result<()> {
+    let project_name = name.unwrap_or("starknet_forge_template".to_string());
+    let project_path = std::env::current_dir()?.join(&project_name);
+
+    Command::new("scarb")
```
maciektr, line 70: Consider using `ScarbCommand` from `scarb-metadata` crate for such simple examples. https://docs.rs/scarb-metadata/latest/scarb_metadata/struct.ScarbCommand.html

crates/forge/src/init.rs
mkaput: You can take this as guaranteed

crates/forge/src/init.rs
```diff
@@ -0,0 +1,104 @@
...
+        .arg("https://github.com/foundry-rs/starknet-foundry.git")
+        .stderr(Stdio::inherit())
+        .stdout(Stdio::inherit())
+        .output()
+        .context("Failed to add snforge_std")?;
+
+    Command::new("scarb")
+        .current_dir(&project_path)
+        .arg("add")
+        .arg("starknet@2.2.0")
```
maciektr, line 93: Can this be read from `scarb --version` output?

crates/forge/src/init.rs
maciektr: słowo harcerza

crates/forge/src/init.rs
tkumor3: I can't update scarb-metadata to 1.7.0 because there dependency conflict with the blockifier. Blockifier demand older version of serde. I assume that solution exists, but we talked with @piotmag769, and ` Command::new("scarb")` is ok

maciektr: Oh, yes, that's totally fine.

crates/forge/src/init.rs
```diff
@@ -0,0 +1,115 @@
...
+use toml_edit::{ArrayOfTables, Document, Item, Table};
+
+static TEMPLATE: Dir = include_dir!("starknet_forge_template");
+
+fn overwrite_files_from_template(
+    dir_to_overwrite: &str,
+    base_path: &PathBuf,
+    project_name: &str,
+) -> Result<()> {
+    let copy_from_dir = TEMPLATE.get_dir(dir_to_overwrite);
```
cptartur, line 18:
```suggestion
    let copy_from_dir = TEMPLATE.get_dir(dir_to_overwrite).ok_or_else(|| anyhow!(...))?;
```

Nit but this way we don't have to deal with match below

crates/forge/src/init.rs
```diff
@@ -0,0 +1,115 @@
...
+    match copy_from_dir {
+        Some(dir) => {
+            for file in dir.files() {
+                fs::create_dir_all(base_path.join(Path::new(dir_to_overwrite)))?;
```
cptartur, line 22: Wouldn't just `create_dir` work here?

crates/forge/src/init.rs
```diff
@@ -0,0 +1,115 @@
...
+    Command::new("scarb")
+        .current_dir(std::env::current_dir().context("Failed to get current directory")?)
+        .arg("new")
+        .arg(&project_path)
+        .stderr(Stdio::inherit())
+        .stdout(Stdio::inherit())
+        .output()
+        .context("Failed to initial new project")?;
+
+    let version = env!("CARGO_PKG_VERSION");
```
cptartur, line 80: Does this work for compiled packages? So running `snforge` and not `cargo run`? This should be verified.

crates/forge/src/main.rs
```diff
@@ -14,17 +15,21 @@ use forge::scarb::{
...
 
 static PREDEPLOYED_CONTRACTS: Dir = include_dir!("crates/cheatnet/predeployed-contracts");
 
 #[derive(Parser, Debug)]
 #[command(version)]
 struct Args {
-    /// Name used to filter tests
-    test_name: Option<String>,
+    /// Name used to filter tests or project name during init
+    name: Option<String>,
```
cptartur, line 26: I wouldn't reuse this argument like this. It is confusing to have it do two things.

Instead we should do something like `init: Option<String>`, and handle it being `Some` or `None`.

starknet_forge_template/Scarb.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "{{ PROJECT_NAME }}"
+version = "0.1.0"
```
cptartur, line 3: Nit, perhaps this should be `0.0.0` just to better show that this is a placeholder value?

crates/forge/src/init.rs
maciektr: The env is checked at compile time, so it would work. 

https://doc.rust-lang.org/std/macro.env.html

starknet_forge_template/Scarb.toml
tkumor3: this file should be removed

crates/forge/src/init.rs
```diff
@@ -0,0 +1,115 @@
...
+
+    Ok(())
+}
+
+fn replace_project_name(contents: &[u8], project_name: &str) -> Vec<u8> {
+    // SAFETY: We control these files, and we know that they are UTF-8.
+    let contents = unsafe { std::str::from_utf8_unchecked(contents) };
+    let contents = contents.replace("{{ PROJECT_NAME }}", project_name);
+    contents.into_bytes()
+}
```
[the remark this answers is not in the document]
tkumor3: Zrzut ekranu 2023-09-7 o 13 00 21

rust-analizer demanding it 🤔 should I do sth with that?

crates/forge/src/init.rs
tkumor3: the problem is, that we have to overwrite `src/`  and `create_dir` returns error in this case `path already exists.`

crates/forge/src/init.rs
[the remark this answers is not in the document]
piotmag769: We shouldn't use `unchecked` imo, just use `from_utf8` and then unpack result with `expect` and appropriate panic message

crates/forge/src/init.rs
```diff
@@ -0,0 +1,115 @@
...
+use scarb_metadata::MetadataCommand;
+
+use std::fs;
+use std::path::{Path, PathBuf};
+use std::process::{Command, Stdio};
+use toml_edit::{ArrayOfTables, Document, Item, Table};
+
+static TEMPLATE: Dir = include_dir!("starknet_forge_template");
+
+fn overwrite_files_from_template(
```
piotmag769, line 13:
```suggestion
fn overwrite_files_from_scarb_template(
```
nit, but can get lost in what template it is

crates/forge/tests/e2e/running.rs
```diff
@@ -440,6 +440,37 @@ fn exit_first_flag_takes_precedence() {
...
+        .current_dir(&temp)
+        .arg("--init")
+        .arg("test_name")
+        .assert()
+        .success();
+
+    let snapbox = runner();
+    snapbox
+        .current_dir(temp.child(Path::new("test_name")))
+        .arg("--exit-first")
```
piotmag769, line 458: nit, but this arg shouldn't be passed here

crates/forge/tests/e2e/running.rs
```diff
@@ -440,6 +440,37 @@ fn exit_first_flag_takes_precedence() {
         "#});
 }
 
+#[test]
+fn init_new_project_test() {
+    let temp = TempDir::new().unwrap();
+
+    let snapbox = runner();
```
piotmag769, line 447: There is still a problem here I mentioned before: it will create a template with `snforge_std` from the latest release as dependency (as we use tag when adding git dependency), so it doesn't test the template if `snforge_std` changes. 

We don't need to (and we shouldn't) modify `init` logic - we should just change the content of Scarb.toml created by this command in this test before running `snforge` to use current branch and repo (if someone makes a PR from fork). Feel free to just copy paste from https://github.com/foundry-rs/starknet-foundry/blob/master/crates/forge/tests/e2e/running.rs#L55

crates/forge/tests/e2e/running.rs
tkumor3: I added it because it will run fewer tests if any fail and it will break the test. I can remove it

crates/forge/src/main.rs
piotmag769: Nvm, your way is way cleaner @tkumor3, good job

crates/forge/tests/e2e/running.rs
```diff
@@ -453,11 +453,57 @@ fn init_new_project_test() {
...
+
+            # See more keys and their definitions at https://docs.swmansion.com/scarb/docs/reference/manifest.html
+
+            [dependencies]
+            snforge_std = {{ git = "https://github.com/foundry-rs/starknet-foundry", tag = "v{}" }}
+            starknet = "2.2.0"
+
+            [[target.starknet-contract]]
+            casm = true
+            # foo = {{ path = "vendor/foo" }}
```
piotmag769, line 474: Could we get rid of that? I think we can just remove commented lines in `init`, I guess `toml_edit` should make it easy

crates/forge/src/main.rs
```diff
@@ -41,6 +46,9 @@ fn load_predeployed_contracts() -> Result<TempDir> {
 
 fn main_execution() -> Result<()> {
     let args = Args::parse();
+    if args.init.is_some() {
+        return init::run(&args.init.unwrap());
+    }
```
piotmag769, line 51:
```suggestion
    if let Some(project_name) = args.init {
        return init::run(&project_name);
    }
```

docs/src/getting-started/first-steps.md
```diff
@@ -27,11 +27,10 @@ $ tree . -L 1
 * `tests/` contains tests. Note that test can also be included in any file or directory.
 * `Scarb.toml` contains configuration of the project as well as of `snforge`, `sncast` etc.
 
-Ensures that `sierra` and `casm` codegen are enabled in the `Scarb.toml` file.
+Ensures that `casm` codegen are enabled in the `Scarb.toml` file.
```
piotmag769, line 30:
```suggestion
Ensures that `casm` codegen is enabled in the `Scarb.toml` file.
```

docs/src/getting-started/first-steps.md
```diff
@@ -40,6 +39,9 @@ And run tests with `snforge`
 
 ```shell
 $ snforge
+[..]Updating[..]
+[..]Compiling[..]
+[..]Finished[..]
 Collected 2 test(s) and 2 test file(s)
 Running 0 test(s) from package_name package
```
piotmag769, line 46:
```suggestion
Running 0 test(s) from project_name package
```

piotmag769 (approved):

crates/forge/tests/e2e/running.rs
tkumor3: @mkaput told that this comment will be removed from scarb

piotmag769 (approved):

cptartur (approved):

PR #570, opened 2023-09-01
Tests design doc

piotmag769:
Closes #

## Introduced changes

-
-

## Checklist

- [x] Linked relevant issue
- [ ] Added relevant tests

### DIFF
--- a/design_documents/test_design.md
+++ b/design_documents/test_design.md
@@ -1,0 +1,30 @@
+# Test design
+
+## Context
+
+Current test design in Starknet Forge has some issues:
+- gas estimation is problematic since we have to distinguish between pure Cairo gas
+(e.g., creating an array in test) and Starknet gas
+- starknet functions like `get_caller_address` are unavailable in tests directly
+- starknet system calls like `library_call` are unavailable in tests directly
+- testing internal contract functions requires exposing them as `external`
+- operating on the contract execution layer while tests being pure Cairo code can create confusion
+
+## Goal
+
+Propose a solution of test design that solves most of the aforementioned issues.
+
+## Proposed Solution
+
+Tests should be contracts (just like in the early days of Protostar).
+Basically, the code in test should behave as if it was a function called in an invoke transaction.
+This clears up the confusion concerning gas estimation — we just calculate the cost of the whole test 
+as it is all executed in the Starknet context. Moreover, starknet functions and system calls now can be made available
+in tests directly. Last but not least, the confusion concerning what actually happens when using dispatchers will be cleared:
+we just call another contract as if we were in invoke transaction.
+
+## Concerns
+
+We need to clearly communicate the changes to the users and explain to them that Starknet Forge is operating on
+execution layer (in fact, account layer) rather than on transaction layer. We need to put an emphasis on the fact
+that all transactions are equivalent to `invoke` in tests.
--- a/design_documents/test_design.md
+++ b/design_documents/test_design.md
@@ -3,12 +3,11 @@
 ## Context
 
 Current test design in Starknet Forge has some issues:
-- gas estimation is problematic since we have to distinguish between pure Cairo gas
+1. Gas estimation is problematic since we have to distinguish between pure Cairo gas
 (e.g., creating an array in test) and Starknet gas
-- starknet functions like `get_caller_address` are unavailable in tests directly
-- starknet system calls like `library_call` are unavailable in tests directly
-- testing internal contract functions requires exposing them as `external`
-- operating on the contract execution layer while tests being pure Cairo code can create confusion
+2. Starknet functions like `get_caller_address` and system calls like `library_call` are unavailable in tests directly 
+3. Testing internal contract functions requires exposing them as `external`
+4. Operating on the contract execution layer while tests being pure Cairo code can create confusion
 
 ## Goal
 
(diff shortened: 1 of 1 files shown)


design_documents/test_design.md
```diff
@@ -0,0 +1,30 @@
+# Test design
+
+## Context
+
+Current test design in Starknet Forge has some issues:
+- gas estimation is problematic since we have to distinguish between pure Cairo gas
+(e.g., creating an array in test) and Starknet gas
+- starknet functions like `get_caller_address` are unavailable in tests directly
+- starknet system calls like `library_call` are unavailable in tests directly
+- testing internal contract functions requires exposing them as `external`
```
cptartur, line 10: So it is just not possible to test internal at all.

design_documents/test_design.md
```diff
@@ -0,0 +1,30 @@
...
+- operating on the contract execution layer while tests being pure Cairo code can create confusion
```
cptartur, line 11: You mean with the current approach? We aren't really operating at this layer if I'm understanding correctly

design_documents/test_design.md
```diff
@@ -0,0 +1,30 @@
...
+
+## Goal
+
+Propose a solution of test design that solves most of the aforementioned issues.
+
+## Proposed Solution
+
+Tests should be contracts (just like in the early days of Protostar).
```
cptartur, line 19: Nit, but referencing protostar is not really useful for people that just started with starknet foundry already.

design_documents/test_design.md
```diff
@@ -0,0 +1,30 @@
...
+Basically, the code in test should behave as if it was a function called in an invoke transaction.
+This clears up the confusion concerning gas estimation — we just calculate the cost of the whole test 
+as it is all executed in the Starknet context. Moreover, starknet functions and system calls now can be made available
+in tests directly. Last but not least, the confusion concerning what actually happens when using dispatchers will be cleared:
```
cptartur, line 23: I think what is still unclear is that what state is used for the system calls made directly in tests. And how would one influence it.

E.g. we can allow using `get_caller_address`, but what it would return? A default value? How is it useful at all? Should be able to apply cheatcodes to the test state and how would that work?

design_documents/test_design.md
```diff
@@ -0,0 +1,30 @@
...
+- operating on the contract execution layer while tests being pure Cairo code can create confusion
+
+## Goal
+
+Propose a solution of test design that solves most of the aforementioned issues.
+
+## Proposed Solution
+
+Tests should be contracts (just like in the early days of Protostar).
+Basically, the code in test should behave as if it was a function called in an invoke transaction.
```
cptartur, line 20: Should we still be allowing the patter of `declare & deploy` and then using dispatchers if this change is introduced?  If so, should the contract-test state be applied to this calls? E.g. should the caller address be that of whatever it was set in the test itself (if we allow it)? 

code example
```cairo
#[test]
fn test() {
  cheatcode_that_changes_the_caller_address_in_test(1234);
  assert(get_caller_address() == 1234, 'this passes');
  // ...
  other_dispatcher.function();
}

// ...
#[starknet::contract]
mod Other {
  // ...
  fn function() {
    // should get_caller_address here be whatever it was set in the test above i.e. 1234?
    get_caller_address() == ???
  }
}

**Review by cptartur — DISMISSED** [review]
I've listed some cases that came to my mind that I think are unclear with the proposed solution. This is more than likely an incomplete list: Perhaps the document should be restructured so all points like this are clearly listed with numbers and then corresponding solution for each one is proposed.

Introducing this kind of changes to the tests is quite serious.

File: design_documents/test_design.md (hunk 2 above)
Comment (piotmag769) [reply design_documents/test_design.md:11]: We are currently operating on execution layer (layer 2) with weird mix of pure Cairo (layer 1), am I right @MaksymilianDemitraszek?

File: design_documents/test_design.md (hunk 1 above)
Comment (piotmag769) [reply design_documents/test_design.md:10]: Discussed offline, will add an example

File: design_documents/test_design.md [hunk 6]
```diff
@@ -0,0 +1,30 @@
...
+This clears up the confusion concerning gas estimation — we just calculate the cost of the whole test 
+as it is all executed in the Starknet context. Moreover, starknet functions and system calls now can be made available
+in tests directly. Last but not least, the confusion concerning what actually happens when using dispatchers will be cleared:
+we just call another contract as if we were in invoke transaction.
+
```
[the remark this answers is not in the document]
piotmag769: Discussed offline, will add an example

design_documents/test_design.md
piotmag769: State of the test (as each test is a contract) will be used - basically we will have storage implementation for the test. When it comes to functions like `get_caller_address` that depend on the "who called the test" we will prob just use default values. Not sure what to do with cheatcodes - I guess it depends on the implementation that@MaksymilianDemitraszek is working on. From what I understood from our conversation offline some cheatcodes won't make any sense (like prank, roll etc.) since we just want to give a test a state (storage access) but don't want to include it in blockifier's address map

design_documents/test_design.md
piotmag769: Discussed offline - all tests using the declare & deploy flow will still work, these functions will basically be syscalls with better UX.
In your example `get_caller_address` inside `function` should return the address of the test (since each test is a contract), so chaging caller address inside the test would only change the address of the artificial "caller of the `test` function". We can think of a test as a function on starknet - basically everything happens on execution layer (or inside `invoke` when looking from 3rd layer perspective)

design_documents/test_design.md
```diff
@@ -18,13 +17,33 @@ Propose a solution of test design that solves most of the aforementioned issues.
...
+3. See next [section](#testing-internal-functions).
+4. The confusion concerning what actually happens when using dispatchers is cleared: we just call another contract as if we were in an invoke transaction.
+
+## Testing internal functions
+
+When implementing contracts, a function called `contract_state_for_testing` is created just before compilation to Sierra/CASM, just like with dispatchers.
+It can be used to obtain the state (storage) of the current (to be tested) contract. Since each test is a contract, now internal functions can be tested
+using the storage of the test.
+
+```rust
```
cptartur, line 33:
```suggestion
```cairo
```

design_documents/test_design.md
```diff
@@ -18,13 +17,33 @@ Propose a solution of test design that solves most of the aforementioned issues.
...
+use pkg::HelloStarknet::debtContractMemberStateTrait;
+use pkg::HelloStarknet::balanceContractMemberStateTrait;
+
+#[test]
+fn test_essa() {
```
cptartur, line 38: Nit but let's name it something more serious.

design_documents/test_design.md
```diff
@@ -18,13 +17,33 @@ Propose a solution of test design that solves most of the aforementioned issues.
...
+    let mut state = HelloStarknet::contract_state_for_testing();
+    state.balance.read();
+    state.debt.write(5);
+}
+```
```
cptartur, line 43: Should this snippet show the interfaces as well?

design_documents/test_design.md
```diff
@@ -18,13 +17,33 @@ Propose a solution of test design that solves most of the aforementioned issues.
...
 
 ## Concerns
 
 We need to clearly communicate the changes to the users and explain to them that Starknet Forge is operating on
 execution layer (in fact, account layer) rather than on transaction layer. We need to put an emphasis on the fact
-that all transactions are equivalent to `invoke` in tests.
+that everything that they should treat tests as if they were happening in an `invoke` transaction.
```
cptartur, line 49:
```suggestion
that tests should be treated as if they were happening in an `invoke` transaction.
```

design_documents/test_design.md
piotmag769: Leftover, thanks

MaksymilianDemitraszek:
PR introducing this changes has been already merged. Can we merge/close this one?

design_documents/test_design.md
MaksymilianDemitraszek: We are operating on layer 2

MaksymilianDemitraszek (approved):

Issue #18, opened 2023-07-06
Deployment scripts in cairo


THenry14:
https://github.com/software-mansion/protostar/issues/2174

@MaksymilianDemitraszek please share your thoughts 🙏

MaksymilianDemitraszek:
Hi, my broad idea is something like this
```
fn some_contract_stuff (cli_arg: felt, cli_arg2: felt) {
    let class_hash = declared('ContractName');
    let address = deploy(class_hash);
   
    let dispatcher = DispatcherContractA { contract_address: address}

    dispatcher.increase_balance(5);

    let val = dispatcher.get_balance();

    let multicall = MulticallDispatcherContractA { contract_address: address};

    multicall.increase_balance(5);
    multicall.get_balance();

    multicall.execute();


}
```

This is some general idea I have in mind but details are TBD.

ClementWalter:
This is `forge script --broadcast` I guess

PR #579, opened 2023-09-04
Add deployment scripts design doc

THenry14:
## Introduced changes

This PR introduces the idea for cairo deployment scripts. After it is approved, new issues for MVP should be created.
https://github.com/foundry-rs/starknet-foundry/issues/18

@glihm I know you were interested in this

## Checklist

- [X] Linked relevant issue
- [X] Added relevant tests

### DIFF
--- a/design_documents/cairo_deployment_scripts.md
+++ b/design_documents/cairo_deployment_scripts.md
@@ -74,7 +74,8 @@ We could have a flag to suppress outputting to stdout or to output with differen
 By default, the scripts could run in some kind of "dry run mode", meaning all the transactions would be simulated with 
 specific rpc calls (like `starknet_simulateTransactions`). An interesting idea would be to research an approach similar 
 to cheatnet - fork a runner and test the scripts against it (to be assessed if it is feasible). We could also provide 
-a way to show a trace of said transactions if required.
+a way to show a trace of said transactions if required. In later iterations, we could also provide a helper function
+that sets up the devnet to be used for dry running for users.
 
 The behaviour would be changed with `--broadcast` flag, that would actually execute needed transactions.
 
@@ -311,7 +312,7 @@ If it makes sense, some of the flags could be included in Scarb.toml config file
 ## Proposed steps/tasks
 
 MVP:
-  - allow for writing scripts using dispatchers and imported cast subcommand functions
+  - allow for writing scripts using dispatchers and imported cast subcommand functions:
     - no idempotency required at this stage
     - all cast subcommand functions return every information necessary (tx hashes, addresses etc) as importable structs
     - no cast subcommand requires CastConfig directly (eg as parameter)
@@ -331,3 +332,4 @@ next iteration:
 next iteration:
   - add support for flags from [Miscellaneous](#miscellaneous)
   - script tests
+  - helper for user to set up devnet for dry_running / tests
--- a/design_documents/cairo_deployment_scripts.md
+++ b/design_documents/cairo_deployment_scripts.md
@@ -36,6 +36,8 @@ We must make sure our functions return specific types, that make retrieving esse
 we should be in a pretty good shape, for most commands we return specific structs (defined in `cast/src/helpers/response_structs.rs`),
 but let's double-check everything necessary is included there, and all importable commands are using such structs.
 
+The commands would have to be implemented in the same manner as forge's cheatcodes.
+
 ### sncast config
 We should allow for all flags passed to sncast to be propagated to the script. CastConfig should also be importable
 and usable from script if someone wishes so, but we should not require it. That means, we might have to review our subcommands
@@ -211,25 +213,39 @@ Picking those would ensure us that:
 An example deployment script could look like this:
 
 ```cairo
-// we might want to rename account functions to avoid confusion
+// we might need to rename account functions to avoid confusion
 use cast::starknet_commands::account::create::create as create_account
+use cast::starknet_commands::account::deploy::deploy as deploy_account
+use cast::{get_provider, get_account};
+use cast::starknet_commands::{declare, deploy, invoke, call};
 (...)
 
-fn make_account(...) -> Result<InvokeResponse> {
-  let user = create_account(...);
-  let prefund_account = invoke(...);
-  deploy_account(...);
+fn make_account(provider: &JsonRpcClient<HttpTransport>) -> Result<SingleOwnerAccount<&'a JsonRpcClient<HttpTransport>, LocalWallet>> {
+  let mut prefunder = get_account("user", &provider)?; // the user that will prefund new account 
+  let user = create_account("user1", &provider);
+  let prefund_account = invoke("0x123", "deposit", [1234, user.address], &prefunder);
+  deploy_account("user1", &provider);
+  get_account("user1", &provider)?
 }
 
 fn main() {
-  let user = make_account(...);
+  ler provider = get_provider("http://127.0.0.1:5050/rpc")?;
+  let user = make_account(provider);
+
+  let declared_contract = declare("mycontract", &user, "/path/to/Scarb.toml");
+  let contract = deploy(&declared_contract.class_hash, [], &user);
+
+  let dispatcher = DispatcherMyContract { contract.contract_address };
+
   skip_if_done(|| {
-        dispatcher.increase_balance(...);
+        dispatcher.increase_balance(5);
     });
   let get_balance = skip_if_done(|| {
-        dispatcher.get_balance(...);
+        dispatcher.get_balance();
     });
-}  
+
+  let called_balance = call(&contract.contract_address, "get_balance", [], "latest"); 
+}
 ```
 
 ### example state file
@@ -263,11 +279,16 @@ Its schema could look like this:
 }
 ```
 
-Having this, we could allow users to `--resume`, `--retry` or `--verify` transactions if something unexpected happens,
+Having this, we could allow users to `--retries` or `--verify` transactions if something unexpected happens,
 some transactions were rejected because of faulty script, too low user eth balance etc.
 
 If we were to implement dry run mode that runs by default, the state should only be written to a file when we pass `--broadcast`
 flag; otherwise the output should be directed to stdout.
+
+The file should be used to see, if given action/transaction was already taken - it should take under account the function
+name as well as its parameters, to distinguish functions from one another (eg. user creates two accounts, we should check
+state file records, one by one, and check if a function `account new` was already invoked with specified parameter `username` 
+(and others, if supplied)).
 
 ### Miscellaneous
 
@@ -275,11 +296,11 @@ There would be a new subcommand `script` that would be invoked with required par
 `*.s.cairo` extension. Proposed interface:
 
 - `--gas-multiplier` - relative percentage by which to multiply add gas estimates (perhaps passed directly to sncast)
-- `--boradcast` - boradcasts the transactions
+- `--broadcast` - broadcasts the transactions
 - `--no-state-file` - do not write to state file
-- `--to-state-file` - specify path to custom state file
+- `--state-file` - specify path to custom state file
 - `--log-file` - output logs to a file
-- `--silent` - do not output logs to stdout
+- `--quiet` - do not output logs to stdout
 - `--slow` - makes sure a transaction is sent, only after its previous one has been confirmed and succeeded (possible reuse of `--wait` flag)
 - `--verify` - find a matching broadcast in state file, and try to verify transactions
 - `--delay` - optional timeout to apply in between attempts in seconds (perhaps passed directly to sncast)
--- a/design_documents/cairo_deployment_scripts.md
+++ b/design_documents/cairo_deployment_scripts.md
@@ -19,7 +19,7 @@
 ## Context
 
 There should be a possibility to write a script in cairo, that would enable users to make transactions and send them 
-to chain. It should allow to declare and deploy contracts as well as apply state transitions on already deployed
+to starknet chain. It should allow to declare and deploy contracts as well as apply state transitions on already deployed
 contracts.
 
 ## Goal
@@ -32,9 +32,8 @@ This section is split into smaller subsections describing things we will need to
 
 ### sncast commands
 Specific sncast commands (declare, deploy, account) could be imported as regular functions to the scipts, and called as such.
-We must make sure our functions return specific types, that make retrieving essential information easier. At the moment 
-we should be in a pretty good shape, for most commands we return specific structs (defined in `cast/src/helpers/response_structs.rs`),
-but let's double-check everything necessary is included there, and all importable commands are using such structs.
+Our functions return specific types (structs defined in `cast/src/helpers/response_structs.rs`), that make retrieving
+essential information easier, so we should be good on that front.
 
 The commands would have to be implemented in the same manner as forge's cheatcodes.
 
@@ -255,28 +254,30 @@ Its schema could look like this:
 
 ```json
 {
-  "create_account": {
-    "arguments": {
-      "name": "whatever",
-      (...)
+  "version": 1,
+  "transactions" : {
+    "create_account": {
+      "arguments": {
+        "name": "whatever",
+        (...)
+      },
+      "output": {
+        "address": "0x123",
+        "max_fee": "0x321"
+      },
+      "status": "accepted",
+      "timestamp": (...)
     },
-    "output": {
-      "address": "0x123",
-      "max_fee": "0x321",
-    },
-    "status": "accepted",
-    "timestamp": (...),
-  },
-  "prefund_account": {
-    "arguments": {
-      ...
-    },
-    "output": {
-      "error": "..."
-    }
-  },
-  "status": "rejected",
-  "timestamp": (...),
+    "prefund_account": {
+      "arguments": {
+        (...)
+      },
+      "output": {
+        "error": "..."
+      },
+      "status": "rejected",
+      "timestamp": (...)
+  }
 }
 ```
 
--- a/design_documents/cairo_deployment_scripts.md
+++ b/design_documents/cairo_deployment_scripts.md
@@ -87,7 +87,7 @@ log information.
 
 In case of failed transaction, in the "output" field in state file we could include an error message. After fixing the issue
 we should allow users to just re-run the script - all the previous (succeeded) transactions should not be replayed, the
-erroneous transaction should be retried and its output should be put into state file.
+erroneous transaction should be retried and its output should be put into [state file](#example-state-file).
 
 ### idempotency
 At the later stages we will want to have the script to be able to track the on chain state using a state file, which would 
--- a/design_documents/cairo_deployment_scripts.md
+++ b/design_documents/cairo_deployment_scripts.md
@@ -1,0 +1,312 @@
+# Cairo deployment scripts
+
+## Table of contents
+
+* [Context](#context)
+* [Goal](#goal)
+* [Considered Solutions](#considered-solution)
+  * [sncast commands](#sncast-commands)
+  * [sncast config](#sncast-config)
+  * [interacting with contract](#interacting-with-contract)
+  * [running the script](#running-the-script)
+  * [error handling](#error-handling)
+  * [idempotency](#idempotency)
+  * [example script](#example-script)
+  * [example state file](#example-state-file)
+  * [miscellaneous](#miscellaneous)
+* [Proposed steps and tasks](#proposed-stepstasks)
+
+## Context
+
+There should be a possibility to write a script in cairo, that would enable users to make transactions and send them 
+to chain. It should allow to declare and deploy contracts as well as apply state transitions on already deployed
+contracts.
+
+## Goal
+
+Propose a solution with an example syntax, that would allow to write deployment scripts in cairo.
+
+## Considered solution
+
+This section is split into smaller subsections describing things we will need to tackle while implementing the solution.
+
+### sncast commands
+Specific sncast commands (declare, deploy, account) could be imported as regular functions to the scipts, and called as such.
+We must make sure our functions return specific types, that make retrieving essential information easier. At the moment 
+we should be in a pretty good shape, for most commands we return specific structs (defined in `cast/src/helpers/response_structs.rs`),
+but let's double-check everything necessary is included there, and all importable commands are using such structs.
+
+### sncast config
+We should allow for all flags passed to sncast to be propagated to the script. CastConfig should also be importable
+and usable from script if someone wishes so, but we should not require it. That means, we might have to review our subcommands
+and if some of them requires it directly (`account create` is one), we should try to get rid of this dependency. 
+
+### interacting with contract
+We will use contracts dispatchers to be able to call/invoke its functions directly 
+(e.g. if contract named `mycontract` has a function `myfunction`, we should be able to just do `mycontract.myfunction();`).
+Later on we could also support our subcommands (`invoke`, `call`) to call/invoke contracts without dispatchers.
+
+### running the script
+The script would essentially run an entrypoint function - `main`. Inside our script subcommand, we will have to compile 
+the cairo script file using `scarb build`, and then run it using `scarb cairo-run`. 
+The main function would be required in the deployment script, and we should return an error if it is not found.
+
+In order to build the script, there'll need to be a `Scarb.toml` file present in the working directory. We should either:
+- look for a Scarb.toml in the same directory script is (this is scarb's default behaviour - we just must make sure scarb
+command is executed from this directory - Command::new(...).current_dir(...) should do); if does not exist we could just create it
+- create a hidden Scarb.toml with a custom name (eg `.script_name.toml`) and override path to scarb toml using scarb flag;
+if a file does not exist let's just create it
+
+Option 1 is probably more in line with scarb ethos, but the option 2 would probably be more practical for us as it would allow
+for multiple Scarb.toml files in one directory, where multiple scripts could reside + Scarb.toml file present in cwd would not
+confuse sncast subcommands (eg declare also looks for Scarb.toml in cwd, which could lead to errors/user confusion).
+
+The deployment script could then be run like so:
+```bash
+$ sncast script /path/to/myscript.s.cairo
+```
+
+Log info should be outputted to stdout and optionally written to some log file. 
+We could have a flag to suppress outputting to stdout or to output with different formats (int, hex).
+
+By default, the scripts could run in some kind of "dry run mode", meaning all the transactions would be simulated with 
+specific rpc calls (like `starknet_simulateTransactions`). An interesting idea would be to research an approach similar 
+to cheatnet - fork a runner and test the scripts against it (to be assessed if it is feasible). We could also provide 
+a way to show a trace of said transactions if required.
+
+The behaviour would be changed with `--broadcast` flag, that would actually execute needed transactions.
+
+### error handling
+If the transaction fails, we should put all the relevant info about it to the state file, output log information and stop
+script execution.
+If the script fails (for any reason that is not connected to transactions - eg rpc node down), we should output relevant
+log information.
+
+In case of failed transaction, in the "output" field in state file we could include an error message. After fixing the issue
+we should allow users to just re-run the script - all the previous (succeeded) transactions should not be replayed, the
+erroneous transaction should be retried and its output should be put into state file.
+
+### idempotency
+At the later stages we will want to have the script to be able to track the on chain state using a state file, which would 
+in turn allow the script to be called multiple times without any modifications, to ensure desired state. To achieve that
+we must ensure idempotency of calls to the network.
+
+The proposed solution would be to use an external json file, that holds hashes of transactions that were already performed.
+While executing a command we could first check this file to see if given transaction exists/succeeded, and based on this
+either perform and an action (make a transaction) or skip it and proceed to the next thing in script.
+
+To achieve this there are a few possible solutions:
+- adjust our existing cast subcommand functions to handle checking the file
+  - pros: 
+    - just import a subcommand function and use it without a fuss
+    - easy to implement for cast subcommands functions
+  - cons:
+    - we will have to modify currently existing functions to enable state checking
+    - this solution does not work for dispatcher functions
+
+- just try to execute functions, parse the output and compare it with the state file
+  - pros:
+    - no modifications needed for subcommands functions
+    - just import a function and use it
+  - cons:
+    - kind of brute force solution - we should not have to talk to chain since we have all the necessary info in local file
+    - executing dispatcher functions can cause us all kinds of trouble (eg - someone could invoke 'send_eth' function multiple times, where in fact they only wanted to do it once)
+    - not every cast subcommand is idempotent
+
+- implement a higher order function (or function wrapper) to conditionally execute functions based on the information in state file
+  - pros:
+    - no modifications needed to cast subcommands functions
+    - works with dispatcher functions just fine
+    - added flexibility - user can decide which function should be replayed each time, and which shouldn't
+  - cons:
+    - boilerplate code - wrapping each function call will be ugly
+    - no option to just annotate main function (procedural macros could achieve this, but they do some compile-time modifications)
+    - potentially more complex
+
+An example pseudo-implementation of this could look like this:
+```rust
+fn skip_if_done<F>(func: F)
+where
+    F: Fn() + 'static,
+{
+    // Decorator logic here
+    // Check the state file and decide whether to execute the function or not
+
+    if should_execute {
+        func();
+        // write func outputs to state file
+    } else {
+        // Do nothing or return an appropriate value
+    }
+}
+
+fn main() {
+    skip_if_done(|| {
+        dispatcher.send_eth(...);
+    });
+}
+```
+
+- Use a procedural macro to automatically apply decorator-like behaviour to functions, which will decide on whether
+given function should be executetd or not (based on state file)
+  - pros:
+    - elegant solution
+    - could be applied selectively (per function) or globally (to main)
+  - cons:
+    - more complex solution
+    - probably not very suitable for runtime based behaviours (like reading json file) - requires more research
+
+An example pseudo-implementation of this could look like this:
+```rust
+// skip_if_done.rs
+use proc_macro_hack::proc_macro_hack;
+
+#[proc_macro_hack]
+pub use skip_if_done_impl::skip_if_done;
+
+
+// skip_if_done_impl.rs
+use quote::quote;
+use std::env;
+
+fn should_execute() -> bool {
+    // Your logic to check whether the function should be executed goes here
+    true
+}
+
+#[macro_export]
+macro_rules! skip_if_done {
+    ($($tokens:tt)*) => {
+        {
+            if should_execute() {
+                $($tokens)*
+            }
+        }
+    };
+}
+
+// main.rs
+#[path = "skip_if_done.rs"]
+mod skip_if_done;
+
+fn main() {
+    skip_if_done! {
+        dispatcher.send_eth(...);
+    }
+}
+```
+
+Given all this, the most obvious solutions seem to be:
+  - for cast subcommands functions:
+    - adjust them to handle checking the file
+  - for dispatcher functions:
+    - implement a higher order function
+
+Picking those would ensure us that:
+  - syntax is relatively boiler-code free (only dispatcher functions are annotated)
+  - idempotency is achieved
+  - it is quite easy to write a script
+
+### example script
+An example deployment script could look like this:
+
+```cairo
+// we might want to rename account functions to avoid confusion
+use cast::starknet_commands::account::create::create as create_account
+(...)
+
+fn make_account(...) -> Result<InvokeResponse> {
+  let user = create_account(...);
+  let prefund_account = invoke(...);
+  deploy_account(...);
+}
+
+fn main() {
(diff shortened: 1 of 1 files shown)


design_documents/cairo_deployment_scripts.md
```diff
@@ -0,0 +1,312 @@
...
+If we were to implement dry run mode that runs by default, the state should only be written to a file when we pass `--broadcast`
+flag; otherwise the output should be directed to stdout.
+
+### Miscellaneous
+
+There would be a new subcommand `script` that would be invoked with required parameter - a path to script file, with
+`*.s.cairo` extension. Proposed interface:
+
+- `--gas-multiplier` - relative percentage by which to multiply add gas estimates (perhaps passed directly to sncast)
+- `--boradcast` - boradcasts the transactions
```
drknzz, line 278:
```suggestion
- `--broadcast` - broadcasts the transactions
```

design_documents/cairo_deployment_scripts.md
```diff
@@ -0,0 +1,312 @@
...
+Having this, we could allow users to `--resume`, `--retry` or `--verify` transactions if something unexpected happens,
+some transactions were rejected because of faulty script, too low user eth balance etc.
+
+If we were to implement dry run mode that runs by default, the state should only be written to a file when we pass `--broadcast`
+flag; otherwise the output should be directed to stdout.
+
+### Miscellaneous
+
+There would be a new subcommand `script` that would be invoked with required parameter - a path to script file, with
+`*.s.cairo` extension. Proposed interface:
```
drknzz, line 275: The `.s` would be just conventional though.

design_documents/cairo_deployment_scripts.md
```diff
@@ -0,0 +1,312 @@
...
+
+- `--gas-multiplier` - relative percentage by which to multiply add gas estimates (perhaps passed directly to sncast)
+- `--boradcast` - boradcasts the transactions
+- `--no-state-file` - do not write to state file
+- `--to-state-file` - specify path to custom state file
```
drknzz, line 280:
```suggestion
- `--state-file` - specify path to custom state file
```

design_documents/cairo_deployment_scripts.md
```diff
@@ -0,0 +1,312 @@
...
+- `--log-file` - output logs to a file
+- `--silent` - do not output logs to stdout
```
drknzz, line 282:
```suggestion
- `--quiet` - do not output logs to stdout
```
And we could do `-q` for short

design_documents/cairo_deployment_scripts.md
```diff
@@ -0,0 +1,312 @@
...
+    "output": {
+      "error": "..."
+    }
+  },
+  "status": "rejected",
+  "timestamp": (...),
+}
+```
+
+Having this, we could allow users to `--resume`, `--retry` or `--verify` transactions if something unexpected happens,
```
drknzz, line 266: What exactly would these flags do?

design_documents/cairo_deployment_scripts.md
```diff
@@ -0,0 +1,312 @@
...
+The deployment script could then be run like so:
+```bash
+$ sncast script /path/to/myscript.s.cairo
+```
+
+Log info should be outputted to stdout and optionally written to some log file. 
+We could have a flag to suppress outputting to stdout or to output with different formats (int, hex).
+
+By default, the scripts could run in some kind of "dry run mode", meaning all the transactions would be simulated with 
+specific rpc calls (like `starknet_simulateTransactions`). An interesting idea would be to research an approach similar
```
war-in, line 73: Is it possible to simulate a few transactions? I mean, one can be dependent on the other

design_documents/cairo_deployment_scripts.md
```diff
@@ -0,0 +1,312 @@
...
+$ sncast script /path/to/myscript.s.cairo
+```
+
+Log info should be outputted to stdout and optionally written to some log file. 
+We could have a flag to suppress outputting to stdout or to output with different formats (int, hex).
+
+By default, the scripts could run in some kind of "dry run mode", meaning all the transactions would be simulated with 
+specific rpc calls (like `starknet_simulateTransactions`). An interesting idea would be to research an approach similar 
+to cheatnet - fork a runner and test the scripts against it (to be assessed if it is feasible). We could also provide 
+a way to show a trace of said transactions if required.
```
war-in, line 75: It would be nice to have such a flow:
- run a script
- create devnet instance
- fork the state
- run script against forked devnet

design_documents/cairo_deployment_scripts.md
```diff
@@ -0,0 +1,312 @@
...
+}
+
+fn main() {
+  let user = make_account(...);
+  skip_if_done(|| {
+        dispatcher.increase_balance(...);
+    });
+  let get_balance = skip_if_done(|| {
+        dispatcher.get_balance(...);
+    });
```
war-in, line 231: What would happen if get_balance was called? Will it return an old value? What if the state changed between calls?

design_documents/cairo_deployment_scripts.md
```diff
@@ -0,0 +1,312 @@
...
+If it makes sense, some of the flags could be included in Scarb.toml config file.
+
+## Proposed steps/tasks
+
+MVP:
+  - allow for writing scripts using dispatchers and imported cast subcommand functions
+    - no idempotency required at this stage
+    - all cast subcommand functions return every information necessary (tx hashes, addresses etc) as importable structs
+    - no cast subcommand requires CastConfig directly (eg as parameter)
+    - it is possible to create and deploy an account, create and deploy a contract
```
war-in, line 297: I assume importing a contract would be possible too, right?

design_documents/cairo_deployment_scripts.md
```diff
@@ -0,0 +1,312 @@
...
+- `--delay` - optional timeout to apply in between attempts in seconds (perhaps passed directly to sncast)
+- `--retries` - number of attempts for retrying (perhaps passed directly to sncast)
+
+If it makes sense, some of the flags could be included in Scarb.toml config file.
+
+## Proposed steps/tasks
+
+MVP:
+  - allow for writing scripts using dispatchers and imported cast subcommand functions
+    - no idempotency required at this stage
```
war-in, line 294: Let's adjust the tabs

design_documents/cairo_deployment_scripts.md
THenry14: afaik this endpoint allows for that

design_documents/cairo_deployment_scripts.md
THenry14: I suppose this would look like this:

- someone runs devnet locally with forked testnet state
- in the deployment script uses `get_provider` with devnet url
- profit


would that be ok, or did you have something else in mind?

design_documents/cairo_deployment_scripts.md
THenry14: If this is called with 'latest' block, that would potentially change every next function that depends on this value, so they will be done again (as they are treated as "new").
If it was called with fixed block number, this will always return the same value (correct me if I'm wrong), so nothing changes.

design_documents/cairo_deployment_scripts.md
THenry14: actually this was intended as "subsection" to the general "allow for writing scripts". But if you feel this should be remodeled, that's not a problem.

design_documents/cairo_deployment_scripts.md
THenry14: yes: `allow for writing scripts using dispatchers`

design_documents/cairo_deployment_scripts.md
war-in: I think it is fine and we should allow for both. My idea was to do it `under the hood` so a user would not have to setup devnet by itself

design_documents/cairo_deployment_scripts.md
war-in: Ok, but if it was done once and is marked with `skip_if_done` would it be called again?

design_documents/cairo_deployment_scripts.md
war-in: Sorry, I meant the `account`

design_documents/cairo_deployment_scripts.md
war-in: Oh, ok I missed that. How about adding `:` at the end of the first line?

design_documents/cairo_deployment_scripts.md
THenry14: this could be good, but I think we should wait for this for devnet-rs. Will add info about this to the doc.

design_documents/cairo_deployment_scripts.md
THenry14: Yes, this would be doable using `get_account` function (as in example script)

glihm (commented):
Very interesting stuff, good job for the proposal.

As you already mentioned, forking + run locally on forked devnet is crucial as we can detect some errors prior broadcasting.

A quick question to be sure I understood the topic with cast, the idea is to use existing cheatcodes to then (once in rust), use cast utils to easily create and send TXs, right?

design_documents/cairo_deployment_scripts.md
THenry14: if marked with skip_if_done, it won't be called again. to have it called again, you'd need to "un-mark" it. Do you think this behaviour should change?

design_documents/cairo_deployment_scripts.md
```diff
@@ -0,0 +1,335 @@
...
+
+  skip_if_done(|| {
+        dispatcher.increase_balance(5);
+    });
+  let get_balance = skip_if_done(|| {
+        dispatcher.get_balance();
+    });
+
+  let called_balance = call(&contract.contract_address, "get_balance", [], "latest"); 
+}
```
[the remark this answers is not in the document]
THenry14: Yes, this will work in cheatcode-like fashion, in order to allow rust code to be used in cairo. I didn't describe the mechanisms in detail, jsut mentioned it should work like that, because the team already has some experience and good practices developed for this.

THenry14:
> A quick question to be sure I understood the topic with cast, the idea is to use existing cheatcodes to then (once in rust), use cast utils to easily create and send TXs, right?

We'll use the idea of cheatcodes for this, but we'll need to write new cheatcodes for this to work.

design_documents/cairo_deployment_scripts.md
war-in: I think marking `calls` (not `invokes`) should be disabled entirely. Maybe not if the block is specified. 
I can think of a situation where:
- someone calls a contract and wants to use this value in the next call
- something is wrong with the script and it fails
- in the meantime state changes
- someone runs the script but the value is already `"fetched"`
- wrong value is saved on the chain

THenry14: To be clear - I think skipping should be allowed for calls with fixed blocks (not latest/pending). I see that pending/latest blocks can be problematic. Your concerns are very valid.

However, wouldn't it be too limiting for some people? It all may be remedied with just not using `skip_if_done` for certain calls. And keeping it the way it is currently proposed, it would allow users to have a script that can actually update on chain state based on some external dependency (call). I can imagine such behaviour would be desired by _some_ people. I'd like to provide the users with options, and let them decide.

I am aware that that could prove to be problematic, or even confusing to some. If we go that way, it is crucial to make it really clear what is the behaviour, and what users should do to achieve a behaviour they desire. This owuld have to be described very well in the docs.

Any thoughts?

war-in: I agree that options are overall better but as you said if we go that way the docs would have to be `very` clear.
We can stick to your proposal but I think sometimes skipping the `calls` can lead to unexpected behaviour. We should definitely include that in the docs.

design_documents/cairo_deployment_scripts.md
```diff
@@ -0,0 +1,335 @@
...
+
+- implement a higher order function (or function wrapper) to conditionally execute functions based on the information in state file
+  - pros:
+    - no modifications needed to cast subcommands functions
+    - works with dispatcher functions just fine
+    - added flexibility - user can decide which function should be replayed each time, and which shouldn't
+  - cons:
+    - boilerplate code - wrapping each function call will be ugly
+    - no option to just annotate main function (procedural macros could achieve this, but they do some compile-time modifications)
+    - potentially more complex
```
cptartur, line 127: I think cast commands, when run from the command line should just behave as they did - without checking the state. But it's just my opinion. 

If we followed this approach, I think the solution with wrappers on the basic functions or macros seems to be a preferred solution.

cptartur: Also, if it possible at all, these functions could have the same names as the command line commands, but have this slightly different behavior where they check the state.

design_documents/cairo_deployment_scripts.md
```diff
@@ -0,0 +1,335 @@
...
+
+### error handling
+If the transaction fails, we should put all the relevant info about it to the state file, output log information and stop
+script execution.
+If the script fails (for any reason that is not connected to transactions - eg rpc node down), we should output relevant
+log information.
+
+In case of failed transaction, in the "output" field in state file we could include an error message. After fixing the issue
+we should allow users to just re-run the script - all the previous (succeeded) transactions should not be replayed, the
+erroneous transaction should be retried and its output should be put into state file.
```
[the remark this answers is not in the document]
cptartur: I see it is described below, I think it could be linked here.

design_documents/cairo_deployment_scripts.md
```diff
@@ -0,0 +1,335 @@
...
+  let get_balance = skip_if_done(|| {
+        dispatcher.get_balance();
+    });
+
+  let called_balance = call(&contract.contract_address, "get_balance", [], "latest"); 
+}
+```
+
+### example state file
+The state file by default should be written to the current working directory, with a name `<script file name>.state.json`.
```
cptartur, line 253: What if we used an in-memory db like sqlite? This way we don't have to create and parse our own format and we could even do some kind of migrations in case the format changes (though this might not be necessary?)?

design_documents/cairo_deployment_scripts.md
```diff
@@ -0,0 +1,335 @@
...
+`*.s.cairo` extension. Proposed interface:
+
+- `--gas-multiplier` - relative percentage by which to multiply add gas estimates (perhaps passed directly to sncast)
+- `--broadcast` - broadcasts the transactions
+- `--no-state-file` - do not write to state file
+- `--state-file` - specify path to custom state file
+- `--log-file` - output logs to a file
+- `--quiet` - do not output logs to stdout
+- `--slow` - makes sure a transaction is sent, only after its previous one has been confirmed and succeeded (possible reuse of `--wait` flag)
+- `--verify` - find a matching broadcast in state file, and try to verify transactions
```
cptartur, line 306: Can you explain this flag a bit more in depth? What do you mean by verify?

cptartur (commented):
Like the idea in general, left some comments asking for clarification

design_documents/cairo_deployment_scripts.md
```diff
@@ -0,0 +1,335 @@
...
+
+### sncast config
+We should allow for all flags passed to sncast to be propagated to the script. CastConfig should also be importable
+and usable from script if someone wishes so, but we should not require it. That means, we might have to review our subcommands
+and if some of them requires it directly (`account create` is one), we should try to get rid of this dependency. 
+
+### interacting with contract
+We will use contracts dispatchers to be able to call/invoke its functions directly 
+(e.g. if contract named `mycontract` has a function `myfunction`, we should be able to just do `mycontract.myfunction();`).
+Later on we could also support our subcommands (`invoke`, `call`) to call/invoke contracts without dispatchers.
```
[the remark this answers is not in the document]
THenry14: I'd rather this is something that user needs to do, at least for first iterations. If we gain enough traction and requests from community, we can think about it.

design_documents/cairo_deployment_scripts.md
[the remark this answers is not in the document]
THenry14: Linked the state file :+1:

Makes sense, how would you name such flag? `--force`, `--replay` or maybe something else?

design_documents/cairo_deployment_scripts.md
THenry14: > I think cast commands, when run from the command line should just behave as they did - without checking the state. But it's just my opinion.

Yup, that's the plan - these will be the same functions. The logic would be there in already present cast functions, but it will only be used in deployment scripts. Probably some input parameter should suffice.

design_documents/cairo_deployment_scripts.md
THenry14: Basically - for each transaction found in state file, we get a receipt and see if it exists, is accepted etc.

design_documents/cairo_deployment_scripts.md
```diff
@@ -0,0 +1,335 @@
...
+  * [idempotency](#idempotency)
+  * [example script](#example-script)
+  * [example state file](#example-state-file)
+  * [miscellaneous](#miscellaneous)
+* [Proposed steps and tasks](#proposed-stepstasks)
+
+## Context
+
+There should be a possibility to write a script in cairo, that would enable users to make transactions and send them 
+to chain. It should allow to declare and deploy contracts as well as apply state transitions on already deployed
```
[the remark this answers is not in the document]
cptartur: It makes the document 1% more explicit so I guess it wouldn't hurt us 😅

design_documents/cairo_deployment_scripts.md
```diff
@@ -0,0 +1,335 @@
...
+Log info should be outputted to stdout and optionally written to some log file. 
+We could have a flag to suppress outputting to stdout or to output with different formats (int, hex).
+
+By default, the scripts could run in some kind of "dry run mode", meaning all the transactions would be simulated with 
+specific rpc calls (like `starknet_simulateTransactions`). An interesting idea would be to research an approach similar 
+to cheatnet - fork a runner and test the scripts against it (to be assessed if it is feasible). We could also provide 
+a way to show a trace of said transactions if required. In later iterations, we could also provide a helper function
+that sets up the devnet to be used for dry running for users.
+
+The behaviour would be changed with `--broadcast` flag, that would actually execute needed transactions.
```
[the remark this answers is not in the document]
cptartur: I don't have that much experience with devops tools of this kind. If you think that dry run being default is more standard/more expected by users, we can do it the way suggests. 

For me it just seemed kind of impractical to have to always type a specific flag when I want to run my scripts, but I guess it depends on how complex your scripts are.

Don't have a strong opinion on this as well.

design_documents/cairo_deployment_scripts.md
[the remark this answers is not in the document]
cptartur: Out of this two I think `--replay` sounds better, but perhaps even something verbose like `--clean-run` or similar would be okay?

design_documents/cairo_deployment_scripts.md
cptartur: If you think that using json would be simpler and more practical, I'm fine with this approach. I'm just a bit worried that eventually, this json format will become quite difficult to maintain.

Also do we want to support some kind of versioning of this file? In case the format changes between cast versions. I'm not sure for how long periods are we expecting the users to keep this files.

design_documents/cairo_deployment_scripts.md
cptartur: Okay, so the default run would not compare anything agains the state file? Or would it just check if the transaction was sent without verifying it passed?

design_documents/cairo_deployment_scripts.md
THenry14: Good catch. I have added `version` field to example state file. 

Will leave this unresolved for others to have a say on json vs db.

design_documents/cairo_deployment_scripts.md
THenry14: If there is a state file we will always check it, and based on what's inside we'll execute functions or not. So for example:

```
create_account()
deploy_account()
invoke_something()
```

say we run the script that does those 3 things. For each of them we check the state file, and if it was already done (and suceeded!) we skip it. If it is not, we execute it and write to state file.

A `--verify` flag would add to this flow - for each of them functions we check the state file, and if it was already done we get its tx hash and check against the chain, to see if all out information from state file matches the status on chain (eg, is the status changed etc). If given function was not executed yet, we do it and write to state file.

Hope that clarifies it :)

design_documents/cairo_deployment_scripts.md
```diff
@@ -0,0 +1,336 @@
...
+and if some of them requires it directly (`account create` is one), we should try to get rid of this dependency. 
+
+### interacting with contract
+We will use contracts dispatchers to be able to call/invoke its functions directly 
+(e.g. if contract named `mycontract` has a function `myfunction`, we should be able to just do `mycontract.myfunction();`).
+Later on we could also support our subcommands (`invoke`, `call`) to call/invoke contracts without dispatchers.
+
+### running the script
+The script would essentially run an entrypoint function - `main`. Inside our script subcommand, we will have to compile 
+the cairo script file using `scarb build`, and then run it using `scarb cairo-run`.
```
Arcticae, line 52: Is this going to be doable? Custom hint logic will require us to probably handle both compilation and running

design_documents/cairo_deployment_scripts.md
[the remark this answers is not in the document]
Arcticae: I think adding a warning about incuring costs would also work

design_documents/cairo_deployment_scripts.md
```diff
@@ -0,0 +1,336 @@
...
+
+The behaviour would be changed with `--broadcast` flag, that would actually execute needed transactions.
+
+### error handling
+If the transaction fails, we should put all the relevant info about it to the state file, output log information and stop
+script execution.
+If the script fails (for any reason that is not connected to transactions - eg rpc node down), we should output relevant
+log information.
+
+In case of failed transaction, in the "output" field in state file we could include an error message. After fixing the issue
```
Arcticae, line 87: We should probably ask the user if he wants to re-run from the end of the last txn (i.e. `--continue`)  or run a new flow (y/n confirmation?).

design_documents/cairo_deployment_scripts.md
cptartur: Makes sense now, would be good to include part of this explanation in the document.

design_documents/cairo_deployment_scripts.md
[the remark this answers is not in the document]
MaksymilianDemitraszek: I mean in traditonal devops maybe, but here things are not reversible. I like the idea of the defensive approach

design_documents/cairo_deployment_scripts.md
MaksymilianDemitraszek: Wouldn't be better to let users handle errors themselves in cairo with Result?

MaksymilianDemitraszek (commented):
First part of the review

design_documents/cairo_deployment_scripts.md
```diff
@@ -0,0 +1,336 @@
...
+
+### sncast config
+We should allow for all flags passed to sncast to be propagated to the script. CastConfig should also be importable
+and usable from script if someone wishes so, but we should not require it. That means, we might have to review our subcommands
+and if some of them requires it directly (`account create` is one), we should try to get rid of this dependency. 
+
+### interacting with contract
+We will use contracts dispatchers to be able to call/invoke its functions directly 
+(e.g. if contract named `mycontract` has a function `myfunction`, we should be able to just do `mycontract.myfunction();`).
+Later on we could also support our subcommands (`invoke`, `call`) to call/invoke contracts without dispatchers.
```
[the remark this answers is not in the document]
THenry14: I thought about just requiring users to write their own dispatchers, at least for first iterations of deployment scripts. That would mean one should know when to supply max_fee. Do you guys think this is a no-go?

design_documents/cairo_deployment_scripts.md
THenry14: Yes, I had a meeting with Artur and he outlined to me how it works now in forge (which is a bit different than I imagined). We'll be doing custom build/extension and a runner similarly to forge. Updated this section in newest commit. Sorry I didn't do it sooner.

design_documents/cairo_deployment_scripts.md
THenry14: > We should probably ask the user if he wants to re-run from the end of the last txn (i.e. --continue) or run a new flow (y/n confirmation?).

re-running from last tx would be a default action, as the script aims to be idempotent. There is a flag `--replay` that runs a new flow. Is that ok?

> Wouldn't be better to let users handle errors themselves in cairo with Result?

Hmm, makes sense, but I would also like to keep current behaviour as a possibility. I can imagine a scenario, where someone wants the script to just fail if a transaction fails, and be able to act on it later. would that be ok? + included your suggestion in newest commit.

design_documents/cairo_deployment_scripts.md
```diff
@@ -0,0 +1,336 @@
...
+erroneous transaction should be retried and its output should be put into [state file](#example-state-file).
+
+### idempotency
+At the later stages we will want to have the script to be able to track the on chain state using a state file, which would 
+in turn allow the script to be called multiple times without any modifications, to ensure desired state. To achieve that
+we must ensure idempotency of calls to the network.
+
+The proposed solution would be to use an external json file, that holds hashes of transactions that were already performed.
+While executing a command we could first check this file to see if given transaction exists/succeeded, and based on this
+either perform and an action (make a transaction) or skip it and proceed to the next thing in script.
```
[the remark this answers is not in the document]
THenry14: to see if the transaction was already executed, the proposal is to check for the function name and its arguments (making it some kind of a function signature), and the hash is just its result. So, if I understand you correctly, the transaction would be replayed only if something before it changed, causing the arguments change. An example would be that we do a `call` and mark it to be re-done each script invokation (so eg do not use `skip_if_done`), then you're right and the tx would be re-done. But, this would have to be deliberate. Otherwise, the arguments should not change, so effectively the tx won't be replayed.

Does that address your concern?

design_documents/cairo_deployment_scripts.md
```diff
@@ -0,0 +1,336 @@
...
+Specific sncast commands (declare, deploy, account) could be imported as regular functions to the scipts, and called as such.
+Our functions return specific types (structs defined in `cast/src/helpers/response_structs.rs`), that make retrieving
+essential information easier, so we should be good on that front.
+
+The commands would have to be implemented in the same manner as forge's cheatcodes.
+
+### sncast config
+We should allow for all flags passed to sncast to be propagated to the script. CastConfig should also be importable
+and usable from script if someone wishes so, but we should not require it. That means, we might have to review our subcommands
+and if some of them requires it directly (`account create` is one), we should try to get rid of this dependency.
```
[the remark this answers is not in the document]
THenry14: sncast config - as the section name states. This is our struct we're using to hold current config; I just don't think it should be required to construct it in deployment scripts.

design_documents/cairo_deployment_scripts.md
```diff
@@ -0,0 +1,336 @@
...
+
+### interacting with contract
+We will use contracts dispatchers to be able to call/invoke its functions directly 
+(e.g. if contract named `mycontract` has a function `myfunction`, we should be able to just do `mycontract.myfunction();`).
+Later on we could also support our subcommands (`invoke`, `call`) to call/invoke contracts without dispatchers.
+
+### running the script
+The script would essentially run an entrypoint function - `main`. Inside our script subcommand, we will have to compile 
+the cairo script file using `scarb build`, and then run it using `scarb cairo-run`. 
+The main function would be required in the deployment script, and we should return an error if it is not found.
```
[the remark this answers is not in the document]
THenry14: Added as an idea for later iterations :+1:

design_documents/cairo_deployment_scripts.md
```diff
@@ -0,0 +1,341 @@
...
+use cast::starknet_commands::account::create::create as create_account
+use cast::starknet_commands::account::deploy::deploy as deploy_account
+use cast::{get_provider, get_account};
+use cast::starknet_commands::{declare, deploy, invoke, call};
+(...)
+
+fn make_account(provider: &JsonRpcClient<HttpTransport>) -> Result<SingleOwnerAccount<&'a JsonRpcClient<HttpTransport>, LocalWallet>> {
+  let mut prefunder = get_account("user", &provider)?; // the user that will prefund new account 
+  let user = create_account("user1", &provider);
+  let prefund_account = invoke("0x123", "deposit", [1234, user.address], &prefunder);
```
amanusk, line 229: Do we necessarily want to support funding from the script? If so, shouldn't it be with a dispatcher as well?

design_documents/cairo_deployment_scripts.md
```diff
@@ -0,0 +1,341 @@
...
+  let dispatcher = DispatcherMyContract { contract.contract_address };
+
+  skip_if_done(|| {
+        dispatcher.increase_balance(5);
+    });
+  let get_balance = skip_if_done(|| {
+        dispatcher.get_balance();
+    });
+
+  let called_balance = call(&contract.contract_address, "get_balance", [], "latest");
```
amanusk, line 250: Guess it is a similar questions, but why is this call different than the one with the dispatcher?

design_documents/cairo_deployment_scripts.md
THenry14: This is not necessary, I just thought it was a low hanging fruit to show one can achieve this by simply invoking a contract. For first interation(s) I thought we would just require users to provide the dispatchers, so it would be doable but with additional input from users. would that make sense?

design_documents/cairo_deployment_scripts.md
THenry14: this is just to show that you can use either a dispatcher or call directly (call being a sncast subcommand)

design_documents/cairo_deployment_scripts.md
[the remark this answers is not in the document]
Arcticae: Should we allow creating an account from within the script? Nothing is really doable except for querying the state without the account (querying the complex state can actually be useful, but it'd require implementing some I/O utilities)

design_documents/cairo_deployment_scripts.md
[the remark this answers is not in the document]
Arcticae: What if somebody wants to do another deploy (i.e. second system) with the same parameters? I think it might be a valid use case

design_documents/cairo_deployment_scripts.md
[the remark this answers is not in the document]
Arcticae: I think that this should be a part of the functionality of the scripts, but can probably be postponed and implemented as an improvement. The difficulty of maintaining it would be increased as well, since the compiler (scarb) is not really pluggable now.

design_documents/cairo_deployment_scripts.md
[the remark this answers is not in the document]
THenry14: At the moment, there are a few options:
- one makes sure there is nothing prior to deploy that changes state (if one does not desire such things - eg would be an account creation), then a `--replay` flag can be used to create a second system
- one can modify the state file to exclude txs they want to do (this is veery janky, and not really a solution but rather a workaround)

I understand that those solutions may not be optimal, but I'm not sure how else we should approach it. We could provide another argument (something like `--replay-since` that would indicate from which tx one would want to redo the txs), but I'm not sure this is something that will we desired - at least for first iterations. If you have any ideas on that matter, please feel free to share.

design_documents/cairo_deployment_scripts.md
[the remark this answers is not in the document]
THenry14: I don't see a clear reason why we should not allow it. Given that deployment scripts would act like some kind of 'terraform for starknet', it just makes sense to me to require as little prerequisites as possible, and allow for as much things to be done from within a script as possible. I am, however, ready to have my mind changed on this one :)

design_documents/cairo_deployment_scripts.md
amanusk: Might be pushing it for an initial release, but we can have some kind of `abigen` to create super simple dispatchers

ClementWalter:
jumping after seeing the tweet. simple comment without to much thinking, but it doesn't look like forge scripts.
Forge scripts are plain solidity completely hiding the concept of  `deploy` for example. In a forge script, you do `contract = new MyContract()` as you would do in a contract. forge turn this into a deployment

THenry14:
> jumping after seeing the tweet. simple comment without to much thinking, but it doesn't look like forge scripts. Forge scripts are plain solidity completely hiding the concept of `deploy` for example. In a forge script, you do `contract = new MyContract()` as you would do in a contract. forge turn this into a deployment

Thanks for chipping in! yes, cairo deployment scripts in snfoundry are not 1:1 to forge script in foundry - there are differences, but the goal is roughly the same. 

Regarding deploy, starknet requires users to compile -> declare -> deploy a contract, and the flow looks a bit different in ethereum (compile -> deploy, afaik). Because of such minor differences it just made sense to me to allow users to pick the desired deployment behaviours - I may want to declare and deploy a contract, but also I may want to just deploy an existing contract using its class hash. I know this process could be "hidden" from users, but I think transparency and expliciticy of this is a plus actually. If you think this should be done differently though, I'm more than happy to discuss :) cheers!

Arcticae (approved):

design_documents/cairo_deployment_scripts.md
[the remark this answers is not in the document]
cptartur: Wouldn't just having separate state files per system enough to solve this?

design_documents/cairo_deployment_scripts.md
```diff
@@ -0,0 +1,341 @@
...
+
+  skip_if_done(|| {
+        dispatcher.increase_balance(5);
+    });
+  let get_balance = skip_if_done(|| {
+        dispatcher.get_balance();
+    });
+
+  let called_balance = call(&contract.contract_address, "get_balance", [], "latest"); 
+}
```
cptartur, line 251: So re-running this script would not increase balance? Since it would skip the increase_balance step after first successful run?

design_documents/cairo_deployment_scripts.md
```diff
@@ -0,0 +1,341 @@
...
+name as well as its parameters, to distinguish functions from one another (eg. user creates two accounts, we should check
+state file records, one by one, and check if a function `account new` was already invoked with specified parameter `username` 
+(and others, if supplied)).
+
+### Miscellaneous
+
+There would be a new subcommand `script` that would be invoked with required parameter - a path to script file, with
+`*.s.cairo` extension. Proposed interface:
+
+- `--gas-multiplier` - relative percentage by which to multiply add gas estimates (perhaps passed directly to sncast)
```
cptartur, line 303: Should we somehow allow explicitly specifying the maximum gas for the transaction in code? Perhaps this could be done with `#[...]` annotation so at least it is limited for a specific function. This way the user could split their code into functions with specific limits.

Though an ideal solution would probably be limiting it per function call (but how exactly? and another argument or some other way of configuring it?).

MaksymilianDemitraszek (approved):
Looks good, well thought design:

- Please introduce a notion of Transaction request and Transaction request id we discussed offline
- Please add a line about modular design (pass separation)
- We could consider adding a declarative frontend in the future but I think we should leave it as a future work as it might be much more compliated.

design_documents/cairo_deployment_scripts.md
```diff
@@ -0,0 +1,341 @@
...
+### sncast config
+We should allow for all flags passed to sncast to be propagated to the script. CastConfig should also be importable
+and usable from script if someone wishes so, but we should not require it. That means, we might have to review our subcommands
+and if some of them requires it directly (`account create` is one), we should try to get rid of this dependency. 
+
+### interacting with contract
+We will use contracts dispatchers to be able to call/invoke its functions directly 
+(e.g. if contract named `mycontract` has a function `myfunction`, we should be able to just do `mycontract.myfunction();`).
+We should also support our subcommands (`invoke`, `call`) to call/invoke so dispatchers are able to use them, and for calling/invoking 
+contracts without dispatchers.
```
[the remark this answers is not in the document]
THenry14: yup :+1: our subcommands return structs with returned fields with specified types (for invoke there is tx_hash of type felt for example)

design_documents/cairo_deployment_scripts.md
```diff
@@ -0,0 +1,341 @@
...
+
+### error handling
+If the transaction fails, we should put all the relevant info about it to the state file, output log information and stop
+script execution.
+If the script fails (for any reason that is not connected to transactions - eg rpc node down), we should output relevant
+log information.
+
+In case of failed transaction, we should allow to:
+- let the users handle the errors in cairo with Result
+- let the script fail - in the "output" field in state file we should then include an error message. After fixing the issue
```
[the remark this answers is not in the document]
THenry14: I want to allow users to handle Result:Err by their own if they want to. Otherwise the script should fail.

design_documents/cairo_deployment_scripts.md
THenry14: yes

design_documents/cairo_deployment_scripts.md
THenry14: This would be a goal for later iterations - basically I think we should allow setting max_fee per function, so it would probably be another argument to the function. I view this as more of a implementation issue though, that could be solved while implementing it.

PR #586, opened 2023-09-04
`deploy_at`

piotmag769:
## Introduced changes

- `deploy_at` added to `snforge_std`

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/forge/src/cheatcodes_hint_processor.rs
```diff
@@ -339,25 +338,29 @@ impl CairoHintProcessor<'_> {
...
-                    Err(CheatcodeError::Unrecoverable(err)) => Err(err),
-                }
+                let calldata_length = inputs[1].to_usize().unwrap();
+                let calldata = Vec::from(&inputs[2..(2 + calldata_length)]);
+
+                let contract_address_felt = inputs[2 + calldata_length].clone();
+                let contract_address = ContractAddress(
+                    PatriciaKey::try_from(felt_to_stark_felt(&contract_address_felt))
+                        .expect("StarkFelt to PatriciaKey conversion failed"),
+                );
```
war-in, line 357: `contract_class_from_felt` is here too

crates/forge/src/cheatcodes_hint_processor.rs
```diff
@@ -489,6 +492,30 @@ impl CairoHintProcessor<'_> {
     }
 }
 
+fn handle_deploy_result(
+    deploy_result: Result<ContractAddress, CheatcodeError>,
+    buffer: &mut MemBuffer,
+) -> Result<(), EnhancedHintError> {
+    match deploy_result {
+        Ok(contract_address) => {
+            let felt_contract_address: Felt252 = stark_felt_to_felt(*contract_address.0.key());
```
war-in, line 501: `contract_address_to_felt` 😉

crates/forge/tests/data/contracts/deploy_at_checker.cairo
```diff
@@ -0,0 +1,15 @@
...
+}
+
+#[starknet::contract]
+mod DeployAtChecked {
+    use starknet::ContractAddress;
+    use traits::Into;
+
+    #[external(v0)]
+    fn get_caller_address(ref self: ContractState, contract_address: ContractAddress) -> felt252 {
+        IPrankChecker{ contract_address }.get_caller_address
```
war-in, line 13: Shouldn't it be called? `()`

crates/forge/tests/integration/deploy_at.rs
```diff
@@ -0,0 +1,71 @@
...
+        use array::ArrayTrait;
+        use starknet::ContractAddress;
+        
+        #[starknet::interface]
+        trait IProxy<TContractState> {
+            fn get_caller_address(ref self: TContractState, checker_address: ContractAddress) -> felt252;
+        }
+
+        #[test]
+        fn test_deploy_error_handling() {
```
war-in, line 22: Let's change the name

crates/forge/tests/data/contracts/deploy_at_checker.cairo
war-in: Actually, I think it is just a leftover. We should remove it

docs/src/appendix/forge-library/deploy_at.md
```diff
@@ -0,0 +1,42 @@
...
+#[test]
+fn test_deploy() {
+    let contract = declare('HelloStarknet');
+
+    let mut constructor_calldata = ArrayTrait::new();
+    constructor_calldata.append(42_u8.into());
+    constructor_calldata.append(21);
+    constructor_calldata.append(37);
+    
+    let contract_address = 123.try_into().unwrap();
```
war-in, line 37: How about using `contract_address_const`?

docs/src/appendix/forge-library/deploy_at.md
```diff
@@ -0,0 +1,42 @@
...
+    fn deploy_at(
+        self: @ContractClass,
+        constructor_calldata: @Array::<felt252>,
+        contract_address: ContractAddress
+    ) -> Result<ContractAddress, RevertedTransaction>;
+    // ...
+}
+```
+
+Deploys a contract at given address and returns the address.
```
Radinyn, line 14:
```suggestion
Deploys a contract at a given address and returns the address.
```

snforge_std/src/cheatcodes/contract_class.cairo
```diff
@@ -31,6 +31,11 @@ trait ContractClassTrait {
     fn deploy(
         self: @ContractClass, constructor_calldata: @Array::<felt252>
     ) -> Result<ContractAddress, RevertedTransaction>;
+    fn deploy_at(
```
Radinyn, line 34:
```suggestion
    ) -> Result<ContractAddress, RevertedTransaction>;
    
    fn deploy_at(
```

crates/forge/tests/integration/deploy_at.rs
war-in, line 1: I wonder if we need more tests here 🤔 We do have most of them on the cheatnet side but maybe checking an error message wouldn't be such a bad idea. Wdyt?

war-in (changes requested):

CHANGELOG.md
```diff
@@ -11,6 +11,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 #### Added
 
+- `deploy_at` function
```
Radinyn, line 14:
```suggestion
- `deploy_at` cheatcode
```

is this a cheatcode?

Radinyn (approved):

crates/cheatnet/src/cheatcodes/deploy.rs
```diff
@@ -23,9 +23,11 @@ impl CheatnetState {
         &mut self,
         class_hash: &ClassHash,
         calldata: &[Felt252],
-        salt: ContractAddressSalt,
         contract_address: ContractAddress,
     ) -> Result<ContractAddress, CheatcodeError> {
+        let salt = self.get_salt();
+        self.increment_deploy_salt_base();
```
cptartur, line 29: I'm not sure why was this moved here. Won't it influence the contract address?

crates/forge/tests/data/contracts/deploy_at_checker.cairo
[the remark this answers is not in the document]
cptartur: Because it doesn't seem like it

crates/forge/tests/integration/deploy_at.rs
cptartur: Though perhaps some test for unsuccessful case could be useful

docs/src/appendix/forge-library/deploy_at.md
```diff
@@ -0,0 +1,42 @@
...
+
+    let mut constructor_calldata = ArrayTrait::new();
+    constructor_calldata.append(42_u8.into());
+    constructor_calldata.append(21);
+    constructor_calldata.append(37);
+    
+    let contract_address = 123.try_into().unwrap();
+
+    let address = contract.deploy_at(@constructor_calldata, contract_address).unwrap();
+    // ...
```
cptartur, line 40: Perhaps adding an assert for `address` here would make it extra explicit what this function does?

crates/cheatnet/src/cheatcodes/deploy.rs
war-in: I think it is ok to move it here. Previously `deploy_at` didn't change the salt so I think deploying two same contracts with the `deploy_at` function would panic. 
@piotmag769 could you please add an issue (or a test) for checking it?

CHANGELOG.md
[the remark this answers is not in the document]
piotmag769: Hmmm, is it a cheatcode though? I guess you are right

crates/forge/src/cheatcodes_hint_processor.rs
```diff
@@ -339,25 +338,29 @@ impl CairoHintProcessor<'_> {
...
-                        let felt_contract_address: Felt252 =
-                            stark_felt_to_felt(*contract_address.0.key());
+                handle_deploy_result(
+                    self.cheatnet_state.deploy(&class_hash, &calldata),
+                    &mut buffer,
+                )
+            }
+            "deploy_at" => {
+                let class_hash = inputs[0].clone();
+                let class_hash = ClassHash(StarkFelt::new(class_hash.to_be_bytes()).unwrap());
```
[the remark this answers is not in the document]
piotmag769: Yup but in cheatnet, I cannot create cyclic dependency

crates/forge/src/cheatcodes_hint_processor.rs
piotmag769: Same as above

crates/cheatnet/src/cheatcodes/deploy.rs
piotmag769: It is to avoid having to pass salt to `deploy_at` so it can be used directly from `forge` as salt is stored in cheatnet

I think we already have a test for that, don't we? `deploy_two_at_the_same_address` in cheatnet

crates/forge/tests/integration/deploy_at.rs
piotmag769: I can add a test with deploying 2 contracts at the same address, WDYT?

docs/src/appendix/forge-library/deploy_at.md
piotmag769: We use the pattern I used everywhere in docs so I would leave it this way

war-in (approved):

PR #592, opened 2023-09-05
Moved conversions to a seperate crate

### LINKED ISSUE #489: Extract conversions to a separate crate

Radinyn:
Closes #489

## Introduced changes

- Moved conversions to a separate crate
- Modified existing usage
- Switched to trait-based approach

## Checklist

- [x] Linked relevant issue
- [ ] Added relevant tests


war-in (approved):
🔥 
I thought there were more usages on the forge side 🤔

Radinyn:
@war-in 
>I thought there were more usages on the forge side 🤔

With that merge commit more usages were included, it seems that Forge mostly converted it manually to that point.

PR #604, opened 2023-09-06
Allow for keystore usage

### LINKED ISSUE #499: Support starkli-style keystore
We should take a look at starkli's keystore, and decide if we should support it.

drknzz:
Closes #499 

## Introduced changes

-
-

## Checklist

- [ ] Linked relevant issue
- [ ] Added relevant tests


crates/cast/src/main.rs
```diff
@@ -30,14 +37,18 @@ struct Cli {
     #[clap(short = 'u', long = "url")]
     rpc_url: Option<String>,
 
-    /// Account name to be used for contract declaration; overrides account from Scarb.toml
+    /// Account name to be used for contract declaration; overrides account from Scarb.toml; should be a path if used with --keystore
```
THenry14, line 40: Let's expand the description a bit more more
```suggestion
    /// Account to be used for contract declaration; 
    /// When using keystore (`--keystore`), this should be a path to account file
    /// When using accounts file, this should be an account name
```

Something like this?

crates/cast/src/main.rs
```diff
@@ -240,3 +244,51 @@ fn update_cast_config(config: &mut CastConfig, cli: &Cli) {
 
     config.accounts_file = Utf8PathBuf::from(shellexpand::tilde(&new_accounts_file).to_string());
 }
+
+async fn get_account<'a>(
```
THenry14, line 248: I'd rather move this function away to lib.

All the keystore functionality could be placed in `helpers/keystore.rs` and imported where needed, this function should (ideally) just pick keystore or accounts file based on user input.

crates/cast/src/main.rs
```diff
@@ -84,44 +95,45 @@ async fn main() -> Result<()> {
...
-            account_file_exists(&config.accounts_file)?;
-            let chain_id = get_chain_id(&provider).await?;
-            let mut account =
-                get_account(&config.account, &config.accounts_file, &provider, chain_id)?;
-
             let mut result = starknet_commands::declare::declare(
                 &declare.contract,
                 declare.max_fee,
-                &mut account,
+                &mut account.unwrap(),
```
THenry14, line 115: It is generlly not very good idea to use unwrap() in prod code, as this will panic if the Option is None, which can lead to runtime errors and crashes. 

Panics are not a graceful way to handle errors, and they can make an app unreliable. It is better to do something like:

```
 if let Some(whatever) = &mut account {
     (...)
 } else {
     (...)
}
```

or something like `match`. You can look through our codebase to see examples of that.

crates/cast/tests/data/keystore/my_account.json
```diff
@@ -0,0 +1,13 @@
+{
```
THenry14, line 1: Can you give me some more context on those files?

If I understand correctly, we have a keystore that stores private keys and then we have account files that hold "public" info? or did I mix up something?

THenry14 (commented):
More thorough review will come after the `get_account` function is split (if we agree on that). For now, overall, looking good :+1:

crates/cast/src/main.rs
drknzz: Problem with this is it takes in `Cli` and `CastConfig`, which makes it work only in main. I could tweak the inputs however

crates/cast/src/main.rs
drknzz: Agree with everything. Decided to use unwrap here, since by the logic of the code flow, account shouldn't be `None` at this point in code.

crates/cast/tests/data/keystore/my_account.json
drknzz: my_key.json - keystore storing private / public key
my_account.json - obtained through `starkli account oz init my_account.json --keystore my_key.json`, holds publicly available account data (the file itself can be fetched using something like `starkli account fetch <address>`). one more thing is that when undeployed, the account json file holds the salt, instead of address, which is unrecoverable after deployment as far as i know
my_account_undeployed.json - the undeployed account file

crates/cast/tests/helpers/fixtures.rs
```diff
@@ -18,7 +18,7 @@ use url::Url;
 pub async fn declare_contract(account: &str, path: &str) -> FieldElement {
     let provider = get_provider(URL).expect("Could not get the provider");
     let chain_id = provider.chain_id().await.expect("Could not get chain_id");
-    let account = get_account(
+    let account = get_account_from_accounts_file(
```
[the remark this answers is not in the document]
drknzz: Currently, `get_account` uses `get_account_from_accounts_file`. There should probably be a separate function `get_account_from_keystore` instead of having the logic in `get_account`

crates/cast/src/main.rs
THenry14: We could just have separate `get_account` function per type (file/keystore), import it in main and do a simple if or something.

crates/cast/src/main.rs
THenry14: Sure, I understand. Still, I guess we should avoid using it if we can.

crates/cast/tests/data/keystore/my_account.json
THenry14: Cool, thanks for clarification. Is `starkli account fetch <address>` necessary to get this, or can users just use them accounts without this fetch? 

To be clear - say I have an account created by starkli. Can I just point sncast to it and call it a day, or do I have to do this `fetch` thing first?

drknzz: The starkli flow is like this:
1. create the keystore (or not, if you want to use plaintext private key)
2. create the account json file (my_account_undeployed.json as an example) - it will contain the salt instead of the address
3. prefund the account and deploy it - after that the salt field will change to address field in the account json file. the file then can be removed and completely fetched again from the network

THenry14: cool, so we can just use a file directly, without fetching it (if it exists), right?

drknzz: Roger that

crates/cast/src/main.rs
```diff
@@ -84,44 +95,45 @@ async fn main() -> Result<()> {
...
+
+    match &cli.command {
+        Commands::Call(_) | Commands::Account(_) => {}
+        Commands::Multicall(value)
+            if matches!(
+                &value.command,
+                starknet_commands::multicall::Commands::New(_)
+            ) => {}
+        _ => account = Some(get_account(&cli, &config, &provider).await?),
+    }
```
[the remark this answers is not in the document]
drknzz: Removing repetition, account is being fetched in one place only now

crates/cast/src/lib.rs
```diff
@@ -95,12 +98,69 @@ pub fn decode_chain_id(chain_id: FieldElement) -> String {
...
+    provider: &'a JsonRpcClient<HttpTransport>,
+    chain_id: FieldElement,
+    keystore_path: &Utf8PathBuf,
+    account: &str,
+) -> Result<SingleOwnerAccount<&'a JsonRpcClient<HttpTransport>, LocalWallet>> {
+    if !keystore_path.exists() {
+        bail!("keystore file does not exist");
+    }
+    if account.is_empty() {
+        bail!("--account is required when using keystore");
```
THenry14, line 126: Maybe something like?

```suggestion
        bail!("Path passed with --account cannot be empty!");
```

crates/cast/src/lib.rs
```diff
@@ -95,12 +98,69 @@ pub fn decode_chain_id(chain_id: FieldElement) -> String {
...
+    }
+    let path_to_account = Utf8PathBuf::from(account);
+    if !path_to_account.exists() {
+        bail!("account file does not exist; when using --keystore, --account argument should be a path to the starkli JSON account file");
+    }
+
+    let password = match env::var(KEYSTORE_PASSWORD_ENV_VAR) {
+        Ok(password) => {
+            println!(
+                "Getting keystore password from {KEYSTORE_PASSWORD_ENV_VAR} evironment variable"
```
THenry14, line 136: maybe?
```suggestion
                "{KEYSTORE_PASSWORD_ENV_VAR} environment variable found and will be used"
```

crates/cast/tests/helpers/fixtures.rs
THenry14, line 21: techniaclly, we could just do `get_account` and it should use the file right? so maybe there's no need to invoke `get_account_from_accounts_file` directly?

docs/src/appendix/cast/common.md
```diff
@@ -26,11 +26,19 @@ Account name used to interact with the network, aliased in open zeppelin account
 
 Overrides account from `Scarb.toml`.
 
+If used with `--keystore`, should be a path to starkli account JSON file.
```
THenry14, line 29: can we link to some starkli docs, where this file is described or something?

THenry14: it would be cool if we could do the same with an account file (if possible)

THenry14 (commented):
🔥

crates/cast/src/main.rs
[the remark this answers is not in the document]
THenry14: I suppose, but now we have two match statements that basically do the same, except the first one sets up an account if commands are different than call or multicall new. This is especially confusing since the second match actually executes the commands, so the first one seems rather redundant. I think that getting an account in places it is needed, in only one match statement (as it was previously), in this case can greatly improve the code readability and maintainability. I know the `get_account` will be repeated, but I think this is a small price + we can get rid of `expects` to the accounts. Does that make sense?

docs/src/appendix/cast/common.md
```diff
@@ -19,18 +19,26 @@ Starknet RPC node url address.
...
 ## `--accounts-file, -f <PATH_TO_ACCOUNTS_FILE>`
 Optional.
 
 Path to the open zeppelin accounts file holding accounts info. Defaults to `~/.starknet_accounts/starknet_open_zeppelin_accounts.json`.
 
+## `--keystore, -k <PATH_TO_KEYSTORE_FILE>`
+Optional.
+
+Path to [keystore file](https://book.starkli.rs/signers#encrypted-keystores).
+If specified, --account should be a path to [starkli account JSON file](https://book.starkli.rs/accounts#accounts).
```
THenry14, line 40:
```suggestion
When specified, the --account argument must be a path to [starkli account JSON file](https://book.starkli.rs/accounts#accounts).
```

docs/src/starknet/account.md
```diff
@@ -99,3 +99,19 @@ $ sncast \
 command: account deploy
 transaction_hash: 0x20b20896ce63371ef015d66b4dd89bf18c5510a840b4a85a43a983caa6e2579
 ```
+
+### Using keystore and starkli account
+
+Accounts created and deployed with [starkli](https://book.starkli.rs/accounts#accounts) can be used by specifying the [`--keystore` argument](../appendix/cast/common.md#--keystore--k-path_to_keystore_file).
+
+> 💡 **Info**
+> When passing the `--keystore` argument, `--account` argument should be a path to the starkli account JSON file.
```
THenry14, line 108:
```suggestion
> When passing the `--keystore` argument, `--account` argument must be a path to the starkli account JSON file.
```

THenry14 (approved):

PR #608, opened 2023-09-06
Random fuzz testing

### LINKED ISSUE #313: Fuzz testing

cptartur:
Closes #313 

## Introduced changes

- Added an initial fuzzer
- Tested `RunnerConfig::new` behavior and argument precedence

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


cptartur:
I think the PR is ready for initial round of reviews. Perhaps some more tests are still needed.

cptartur:
> This is more of a "random testing" approach, not really fuzzing per-se It should be clearly communicated and we should not mix this up with fuzzing via a dedicated engine

I've updated the docs a bit so it is clear that this is an initial fuzzing implementation and it just uses random values.

crates/forge/src/lib.rs
```diff
@@ -47,12 +55,18 @@ impl RunnerConfig {
...
+        fuzzer_seed: Option<u64>,
         forge_config_from_scarb: &ForgeConfig,
     ) -> Self {
         Self {
             test_name_filter,
             exact_match,
             exit_first: forge_config_from_scarb.exit_first || exit_first,
+            fuzzer_runs: fuzzer_runs
+                .or(forge_config_from_scarb.fuzzer_runs)
+                .unwrap_or(256),
```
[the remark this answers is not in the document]
cptartur: We can do that

docs/src/SUMMARY.md
```diff
@@ -38,17 +40,17 @@
...
+    * [stop_prank](appendix/cheatcodes/stop_prank.md)
+    * [start_roll](appendix/cheatcodes/start_roll.md)
+    * [stop_roll](appendix/cheatcodes/stop_roll.md)
+    * [start_warp](appendix/cheatcodes/start_warp.md)
+    * [stop_warp](appendix/cheatcodes/stop_warp.md)
+    * [get_class_hash](appendix/cheatcodes/get_class_hash.md)
+    * [l1_handler_execute](appendix/cheatcodes/l1_handler_execute.md)
+    * [start_mock_call](appendix/cheatcodes/start_mock_call.md)
+    * [stop_mock_call](appendix/cheatcodes/stop_mock_call.md)
+    * [spy_events](appendix/cheatcodes/spy_events.md)
```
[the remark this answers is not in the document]
cptartur: It's just the intellij autorformatter 😅

crates/forge/src/lib.rs
```diff
@@ -269,10 +290,72 @@ fn run_tests_from_file(
...
+        } else {
+            pretty_printing::print_fuzz_running(case_name, runner_config);
+
+            if contains_non_felt252_args(&args, &builtins) {
+                bail!("Test {case_name} requires arguments that are not felt252 type");
+            }
+
+            let mut results = vec![];
+
+            for _ in 1..runner_config.fuzzer_runs {
```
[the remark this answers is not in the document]
cptartur: Right, good catch

crates/forge/src/lib.rs
```diff
@@ -269,10 +290,72 @@ fn run_tests_from_file(
...
+                if let TestCaseSummary::Failed { .. } = result {
+                    // Fuzz failed
+                    break;
+                }
+            }
+
+            let (result, args) = results
+                .last()
+                .expect("Test should always run at least once")
+                .clone();
```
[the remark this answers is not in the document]
cptartur: We could, but I though that the case of test not running at all just means that there's some serious problem in the logic, so we should just fail completely

crates/forge/src/main.rs
```diff
@@ -36,6 +36,14 @@ struct Args {
...
     #[arg(short = 'x', long)]
     exit_first: bool,
+
+    /// Number of fuzzer runs
+    #[arg(short = 'r', long)]
+    fuzzer_runs: Option<u32>,
+
+    /// Seed for the fuzzer
+    #[arg(short = 's', long)]
+    fuzzer_seed: Option<u64>,
```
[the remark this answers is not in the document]
cptartur: For the seed? Or runs? I think it's okay for the latter but the first probably not

crates/forge/src/fuzzer.rs
```diff
@@ -0,0 +1,42 @@
...
+
+    pub fn from_seed(seed: u64) -> Self {
+        Random {
+            rng: Box::new(StdRng::seed_from_u64(seed)),
+        }
+    }
+}
+
+impl Fuzzer for Random {
+    fn next_argument(&mut self, type_name: &str) -> Vec<Felt252> {
```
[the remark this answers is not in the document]
piotmag769: It feels weird tbh, I would get rid of that and just add another methods while we develop fuzzing. E.g. we could have `next_felt252_argument`, `next_u8_argument` etc.

crates/forge/src/main.rs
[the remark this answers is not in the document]
piotmag769: Not the other way round?

crates/forge/src/lib.rs
```diff
@@ -269,10 +290,72 @@ fn run_tests_from_file(
...
+            "RangeCheck",
+            "Bitwise",
+            "EcOp",
+            "Poseidon",
+            "SegmentArena",
+            "GasBuiltin",
+            "System",
+        ];
+
+        let result = if args.is_empty() {
```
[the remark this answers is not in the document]
cptartur: I've extracted the method for fuzzing run, code for running without fuzzing is already simple enough

crates/forge/src/fuzzer.rs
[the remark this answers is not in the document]
cptartur: I've removed this abstraction entirely since we don't know our future needs anyway

crates/forge/src/lib.rs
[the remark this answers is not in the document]
piotmag769: Nvm, result is out of scope here already

crates/forge/src/main.rs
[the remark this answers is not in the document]
cptartur: Oh yeah, I meant the other way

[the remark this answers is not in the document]
cptartur: Logic is updated anyways

crates/forge/src/main.rs
```diff
@@ -36,6 +36,24 @@ struct Args {
...
+}
+
+fn validate_runs(val: &str) -> Result<u32> {
+    let parsed_val: u32 = val
+        .parse()
+        .map_err(|_| anyhow!("Failed to parse {val} as u32"))?;
+    if parsed_val < 1 {
+        bail!("Value must be greater than or equal to 1")
```
piotmag769, line 54:
```suggestion
        bail!("Number of fuzzer runs must be greater than or equal to 1")
```

crates/forge/tests/e2e/fuzzing.rs
piotmag769, line 1: Can we add a test with incorrect `fuzzer_runs` value passed?

docs/src/testing/advanced/fuzz-testing.md
```diff
@@ -0,0 +1,59 @@
+# Fuzz Testing
+
+In many cases, a test needs to verify function behavior for multiple possible values.
+While it is possible to come up with these cases on your own, it is often impractical, especially when you want to test
+for a large number of possible arguments.
```
piotmag769, line 5: not sure if `test for` or `test against` is correct

docs/src/testing/advanced/fuzz-testing.md
```diff
@@ -0,0 +1,59 @@
...
+$ snforge
+Collected 1 test(s) and 1 test file(s)
+Running 1 test(s) from package_name package
+Running fuzzer for package_name::test_sum, 256 runs:
+[PASS] package_name::test_sum
+```
+
+## Configuring The Fuzzer
+
+It is possible to configure the number of runs of the random fuzzer as well as its seed too by a command line argument:
```
piotmag769, line 46:
```suggestion
It is possible to configure the number of runs of the random fuzzer as well as its seed too with command line arguments:
```

piotmag769 (approved):

docs/src/testing/advanced/fuzz-testing.md
cptartur: I think the second one sounds, better. Changing

piotmag769 (approved):

crates/forge/src/scarb.rs
```diff
@@ -43,6 +43,10 @@ pub struct ForgeConfig {
     #[serde(default)]
     /// Should runner exit after first failed test
     pub exit_first: bool,
+    /// How many runs should fuzzer execute
+    pub fuzzer_runs: Option<u32>,
```
Arcticae, line 47: We should allow configuring this per-case, can we add a task/note for that?

crates/forge/src/main.rs
```diff
@@ -36,6 +36,24 @@ struct Args {
...
+
+    /// Seed for the fuzzer
+    #[arg(short = 's', long)]
+    fuzzer_seed: Option<u64>,
+}
+
+fn validate_runs(val: &str) -> Result<u32> {
+    let parsed_val: u32 = val
+        .parse()
+        .map_err(|_| anyhow!("Failed to parse {val} as u32"))?;
```
Arcticae, line 52: This will result in an overflow error instead of a more human-readable message when it exceeds u32 size. Although this might be not be a concern since users should not use values that large anyway.

crates/forge/tests/e2e/fuzzing.rs
```diff
@@ -0,0 +1,122 @@
...
+        .args(["--fuzzer-runs", "10"])
+        .assert()
+        .code(1)
+        .stdout_matches(indoc! {r#"
+        [..]Compiling[..]
+        [..]Finished[..]
+        Collected 5 test(s) and 1 test file(s)
+        Running 5 test(s) from fuzzing package
+        [PASS] fuzzing::tests::adding
+        Running fuzzer for fuzzing::tests::fuzzed_argument, 10 runs:
```
Arcticae, line 54: The output is a tad confusing, can we perhaps print the message under the case and maybe indented?
(ran 10 cases etc..)

docs/src/testing/advanced/fuzz-testing.md
```diff
@@ -0,0 +1,59 @@
...
+> ℹ️**Info**
+> Currently, Forge only supports using randomly generated values for fuzzing.
+> This way of fuzzing doesn't support any kind of value generation based on code analysis, test coverage or results of
+> other fuzzer runs.
+> In the future, more advanced fuzzing execution modes will be added.
+
+## Random Fuzzing
+
+> ⚠️ **Warning**
+> Currently Forge only supports fuzz testing arguments with `felt252` type. Trying to use different argument types will
```
Arcticae, line 16: There was a section in the docs about the type of test functions. It should be adjusted in this PR

Arcticae (approved):
I won't block since my comments are mostly nitpicks, but [this](https://foundry-rs.github.io/starknet-foundry/testing/testing.html#test-collecting) info needs to be adjusted before merging

crates/forge/src/scarb.rs
cptartur: #661

crates/forge/src/lib.rs
```diff
@@ -269,10 +287,37 @@ fn run_tests_from_file(
 
     let mut results = vec![];
     for (i, case) in tests.test_cases.iter().enumerate() {
-        let result = run_from_test_case(&runner, case, contracts, predeployed_contracts)?;
+        let case_name = case.name.as_str();
+        let function = runner.find_function(case_name)?;
+        let args = args_for_function(function);
+
+        let result = if args.is_empty() {
```
MaksymilianDemitraszek, line 294: It is very poorly abstracted here, it will be a hell to parallelize here (which we plan to do very soon i suppose). It doesn't need to be done in this PR.
Maybe instead we should create a lazy scheduler which will create a list of tuples (function, args) , for normal functions it will be an single element array. 

```
trait TestScheduler {
   fn get_schedule(function) -> Vec<ScheduledTestCase>;
}
```
and store a map of those schedules which will extract their cases from the resulting list. This way it would be trivially parallelizable. On the other hand we should consider that threads running cased for the same function have to share the state. Just leaving as a heads up.

crates/forge/src/main.rs
```diff
@@ -36,6 +36,24 @@ struct Args {
...
+    /// Number of fuzzer runs
+    #[arg(short = 'r', long, value_parser = validate_runs)]
+    fuzzer_runs: Option<u32>,
+
+    /// Seed for the fuzzer
+    #[arg(short = 's', long)]
+    fuzzer_seed: Option<u64>,
+}
+
+fn validate_runs(val: &str) -> Result<u32> {
```
MaksymilianDemitraszek, line 49:
```suggestion
fn validate_runs(val: &str) -> Result<u32> {
```
The name suggests validation of finished runs

crates/forge/tests/e2e/fuzzing.rs
MaksymilianDemitraszek: I don't se the point in printing ` Running fuzzer for ` it is inconsistent with other test cases.

MaksymilianDemitraszek (dismissed):

crates/forge/src/main.rs
cptartur: In case of the overflow we still get an error like
```shell
snforge --fuzzer-runs 99999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999
error: invalid value '99999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999' for '--fuzzer-runs <FUZZER_RUNS>': Failed to parse 99999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999 as u32
```

Isn't it okay?

crates/forge/src/fuzzer.rs
```diff
@@ -0,0 +1,33 @@
+use cairo_felt::Felt252;
+use num_bigint::{BigUint, RandBigInt};
+use num_traits::Zero;
+use rand::rngs::StdRng;
+use rand::{thread_rng, RngCore, SeedableRng};
+
+pub struct Random {
+    rng: Box<dyn RngCore>,
```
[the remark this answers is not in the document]
piotmag769: I guess we could change the type to [ThreadRng](https://docs.rs/rand/latest/rand/rngs/struct.ThreadRng.html#) since that's the only type we use when creating rng field

docs/src/testing/advanced/fuzz-testing.md
```diff
@@ -0,0 +1,59 @@
...
+
+In many cases, a test needs to verify function behavior for multiple possible values.
+While it is possible to come up with these cases on your own, it is often impractical, especially when you want to test
+against a large number of possible arguments.
+
+> ℹ️**Info**
+> Currently, Forge only supports using randomly generated values for fuzzing.
+> This way of fuzzing doesn't support any kind of value generation based on code analysis, test coverage or results of
+> other fuzzer runs.
+> In the future, more advanced fuzzing execution modes will be added.
```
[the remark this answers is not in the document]
cptartur: I don't think this note makes a lot of sense if you remove `for fuzzing`.

crates/forge/src/lib.rs
```diff
@@ -295,6 +340,67 @@ fn run_tests_from_file(
     })
 }
 
+fn run_test_case_with_fuzzing(
```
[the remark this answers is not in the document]
cptartur: It's not really running a test case so it's an unfortunate name choice. I'll rename it to something differnet.

crates/forge/src/main.rs
cptartur: It's in the main.rs which only handles entry point to the program. I don't think it would confuse anyone. It's used in the `Args` struct only and even accepts `str` as argument.

MaksymilianDemitraszek: I mean i mentioned it beacuse it confused me and i have to check the function body to see what it does. I didin't came up with some arbitrary "confusion possibility" I suggest the name`validate_fuzzer_runs_value`

crates/forge/src/fuzzer.rs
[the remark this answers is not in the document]
cptartur: Actually, I'll change it

crates/forge/src/lib.rs
piotmag769: Will be done with #662 to make changes related to parallelism in one PR

crates/forge/src/lib.rs
```diff
@@ -269,15 +297,40 @@ fn run_tests_from_file(
...
+                contracts,
+                predeployed_contracts,
+                &runner,
+                case,
+                &args,
+                fuzzer,
+            )?;
+            pretty_printing::print_test_result(&result, Some(runs));
+
+            if let TestCaseSummary::Failed { .. } = result {}
```
Arcticae, line 322: what's that for?

crates/forge/src/lib.rs
```diff
@@ -295,6 +348,68 @@ fn run_tests_from_file(
...
+fn run_with_fuzzing(
+    runner_config: &RunnerConfig,
+    contracts: &HashMap<String, StarknetContractArtifacts>,
+    predeployed_contracts: &Utf8PathBuf,
+    runner: &SierraCasmRunner,
+    case: &TestCase,
+    args: &Vec<&ConcreteTypeId>,
+    fuzzer: &mut fuzzer::Random,
+) -> Result<(TestCaseSummary, u32)> {
+    if contains_non_felt252_args(args, &BUILTINS) {
```
Arcticae, line 360: This will never happen since you filter out the non-felt252 arguments earlier

crates/forge/src/lib.rs
```diff
@@ -319,6 +434,61 @@ mod tests {
...
+    #[test]
+    fn runner_config_default_arguments() {
+        let config = RunnerConfig::new(None, false, false, None, None, &Default::default());
+        assert_eq!(
+            config,
+            RunnerConfig {
+                test_name_filter: None,
+                exact_match: false,
+                exit_first: false,
+                fuzzer_runs: FUZZER_RUNS_DEFAULT,
```
Arcticae, line 446: I don't think we should provide default for runs amount, it always should be configured by the user

crates/forge/tests/e2e/running.rs
```diff
@@ -44,7 +44,8 @@ fn simple_package() {
         Running 1 test(s) from tests/without_prefix.cairo
         [PASS] without_prefix::five
         Tests: 9 passed, 2 failed, 0 skipped
-        
+        Fuzzer seed: [..]
```
Arcticae, line 47: Is it always printed? If fuzzing was not enabled, it shouldn't be

Arcticae (changes requested):

crates/forge/src/lib.rs
piotmag769: Good point

crates/forge/src/lib.rs
piotmag769: I think it is fine for now, we cannot really force user to pass it, at least for now. Maybe can be done in #661 with introduction of new compiler plugin

MaksymilianDemitraszek: I think we should provide a default value

crates/forge/tests/e2e/fuzzing.rs
```diff
@@ -91,7 +91,7 @@ fn fuzzing_set_seed() {
         [PASS] fuzzing::tests::fuzzed_argument (fuzzer runs = 256)
         [PASS] fuzzing::tests::fuzzed_both_arguments (fuzzer runs = 256)
         [PASS] fuzzing::tests::passing
-        [FAIL] fuzzing::tests::failing_fuzz (fuzzer runs = 1, arguments = [2176913491924129583795008547758153887527303407267805286823327310126940911830, 2350787482052932408706155311404918127641790603096372587257662840043562009418])
+        [FAIL] fuzzing::tests::failing_fuzz (fuzzer runs = 1, arguments = [[..], [..]])
```
MaksymilianDemitraszek, line 94: Are those args printed?

crates/forge/src/fuzzer.rs
```diff
@@ -1,11 +1,35 @@
...
+#[allow(clippy::module_name_repetitions)]
+pub struct FuzzerRunParams {
+    /// Inclusive value
+    low: BigUint,
+    /// Exclusive value
+    high: BigUint,
+    /// Number of arguments
+    args_number: usize,
+    /// Total number of runs
+    runs_all: u32,
```
MaksymilianDemitraszek, line 18:
```suggestion
    runs_number: u32,
```

crates/forge/src/fuzzer.rs
```diff
@@ -1,11 +1,35 @@
...
+    runs_with_min_value: Vec<u32>,
+    /// Runs in which arguments have max value e.g. `run_with_min_value[0] = 5`
+    /// means that the first argument has the highest possible value in 5th run
+    runs_with_max_value: Vec<u32>,
+}
 
 pub struct Random {
     rng: StdRng,
     seed: u64,
+    fuzzer_run_params: FuzzerRunParams,
```
MaksymilianDemitraszek, line 32: I think it should be other way around, Fuzzer should own random, do we actually still need this structure? Woudn't be better to just have `RandomFuzzer` struct?

namely something like `RandomFuzzer::new(seed, runs_number)`

crates/forge/src/fuzzer.rs
```diff
@@ -1,11 +1,35 @@
...
+use rand::{thread_rng, Rng, RngCore, SeedableRng};
+use std::default::Default;
+
+#[derive(Default)]
+#[allow(clippy::module_name_repetitions)]
+pub struct FuzzerRunParams {
+    /// Inclusive value
+    low: BigUint,
+    /// Exclusive value
+    high: BigUint,
```
MaksymilianDemitraszek, line 14: We will always generate felts, those have to be configurable?

crates/forge/src/lib.rs
```diff
@@ -203,13 +223,19 @@ pub fn run(
 
     let mut tests_iterator = tests.into_iter();
 
+    let mut fuzzer = match runner_config.fuzzer_seed {
+        None => fuzzer::Random::new(),
+        Some(seed) => fuzzer::Random::from_seed(seed),
+    };
```
MaksymilianDemitraszek, line 229: Why it is created per file not per test case? number of runs should be per test case

docs/src/projects/configuration.md
```diff
@@ -27,12 +27,12 @@ by either supplying cast with those parameters directly [see more detailed CLI d
...
 ```toml
-(...)
+# ...
 [tool.sncast.myprofile]
 account = "user"
 accounts-file = "~/my_accounts.json"
 url = "http://127.0.0.1:5050/rpc"
-(...)
+# ...
 ```
```
MaksymilianDemitraszek, line 36: Unrelated change

docs/src/testing/advanced-testing.md
```diff
@@ -0,0 +1,5 @@
+# Advanced Testing
```
MaksymilianDemitraszek, line 1: I think we don't need this subsection, we can just include fuzz testing

docs/src/testing/testing.md
```diff
@@ -34,7 +34,9 @@ Tests: 1 passed, 0 failed, 0 skipped
 ## Test collecting
 
 Forge considers all functions in your project marked with `#[test]` attribute as tests.
-Test functions cannot return any values and cannot take any arguments.
+By default, test functions run without any arguments.
+However, adding any arguments to function signature will enable [fuzz testing](./advanced/fuzz-testing.md) for this
+function.
```
MaksymilianDemitraszek, line 39:
```suggestion
test case.
```

crates/forge/src/fuzzer.rs
```diff
@@ -0,0 +1,192 @@
...
+
+impl Random {
+    pub fn set_fuzzer_run_params(
+        &mut self,
+        fuzzer_runs: u32,
+        args_number: usize,
+        low: BigUint,
+        high: BigUint,
+    ) {
+        // sanity checks
```
MaksymilianDemitraszek, line 69:
```suggestion

```

crates/forge/tests/e2e/fuzzing.rs
piotmag769: Yup, actually I can use the exact values (they changed since we use `get_random` more times before, but the seed is still fixed)

crates/forge/src/fuzzer.rs
piotmag769: I guess we can just do another PR with enabling other types of args. Then we can think about how to solve this - we can have args of types `u8`, `u16` and `felt252` in one function and then simple generic is not enough

piotmag769: https://github.com/foundry-rs/starknet-foundry/issues/693

crates/forge/src/lib.rs
piotmag769: Actually it is one fuzzer per package. I agree, but then seeds could be different for each test case - do we want that? If we do I can do it this way and print the seed after each test case

docs/src/getting-started/first-steps.md
```diff
@@ -29,10 +29,10 @@ $ tree . -L 1
 
 Ensures that `casm` codegen is enabled in the `Scarb.toml` file.
 ```toml
-(...)
+# ...
 [[target.starknet-contract]]
 casm = true
-(...)
+# ...
```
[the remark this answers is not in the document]
piotmag769: Ik, but would leave it as it is

docs/src/SUMMARY.md
```diff
@@ -38,17 +40,17 @@
...
-  * [start_roll](appendix/cheatcodes/start_roll.md)
-  * [stop_roll](appendix/cheatcodes/stop_roll.md)
-  * [start_warp](appendix/cheatcodes/start_warp.md)
-  * [stop_warp](appendix/cheatcodes/stop_warp.md)
-  * [get_class_hash](appendix/cheatcodes/get_class_hash.md)
-  * [l1_handler_execute](appendix/cheatcodes/l1_handler_execute.md)
-  * [start_mock_call](appendix/cheatcodes/start_mock_call.md)
-  * [stop_mock_call](appendix/cheatcodes/stop_mock_call.md)
-  * [spy_events](appendix/cheatcodes/spy_events.md)
+    * [start_prank](appendix/cheatcodes/start_prank.md)
```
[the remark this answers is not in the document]
piotmag769: Not my changes, but can revert

crates/forge/src/fuzzer.rs
piotmag769: Can do if we want to have fuzzer per test case

piotmag769: Decided to leave the struct but renamed Random to RandomFuzzer

crates/forge/src/fuzzer.rs
```diff
@@ -0,0 +1,171 @@
...
+            .zip(&runs_with_min_value)
+            .map(|(_, &run_with_min)| {
+                let run_with_max = self.rng.gen_range(1..=self.fuzzer_run_params.runs_number);
+                if run_with_max == run_with_min {
+                    run_with_min + 1 % runs_number
+                } else {
+                    run_with_max
+                }
+            })
+            .collect();
```
Arcticae, line 85: Wouldn't it be easier to just throw in `[max, min, rand1, rand2..., rand(n-2)]` ?

crates/forge/src/fuzzer.rs
```diff
@@ -0,0 +1,171 @@
...
+
+        self.fuzzer_run_params.runs_with_max_value = runs_with_max_value;
+        self.fuzzer_run_params.runs_with_min_value = runs_with_min_value;
+    }
+
+    pub fn next_felt252_args(&mut self) -> Vec<Felt252> {
+        assert!(self.fuzzer_run_params.runs_executed < self.fuzzer_run_params.runs_number);
```
cptartur, line 92: With this the fuzzer can now fail in runtime if it was misconfigured. I wonder if it shouldn't be handled more gracefully.

crates/forge/src/fuzzer.rs
```diff
@@ -0,0 +1,171 @@
...
+                    } else {
+                        self.rng.gen_biguint_range(
+                            &self.fuzzer_run_params.low,
+                            &self.fuzzer_run_params.high,
+                        )
+                    },
+                )
+            })
+            .collect();
+
```
cptartur, line 115: We're constructing an iterator every time someone calls `next_felt252_args` + this code is quite confusing. Isn't there a way to handle these in without enumerating and mapping (just a loop maybe or some other approach)?

crates/forge/src/fuzzer.rs
piotmag769: Actually true, will change it

crates/forge/src/fuzzer.rs
piotmag769: It is just a sanity check, this check should always pass if the API of fuzzer is used properly by us. I think we should panic if the check fails

crates/forge/src/fuzzer.rs
piotmag769: Creating an iterator doesn't hurt I think, but I can try to make the code less confusing

crates/forge/src/fuzzer.rs
MaksymilianDemitraszek: The idea was to be able to test it in different combinations to create stronger test cases, this way you will only test 0 with other 0s.

piotmag769: Not really, we want to have max and min values in random runs. The proposed solution is fine for `args_number == 1`, otherwise we will have first run will all args set to max, then 2nd one with all args set to min.

TL;DR: for 3 args, 3 runs, `low=0`, `high=10` currently we can have sth like:
```
[10, 0, 1]
[0, 2, 0]
[5, 10, 10]
```
Using the proposed solution we will have
```
[10, 10,  10]
[0, 0, 0]
[5, 3, 2]
```
so max and min values are always tested together :/

crates/forge/src/fuzzer.rs
```diff
@@ -0,0 +1,171 @@
...
+    /// Total number of runs
+    runs_number: u32,
+    /// Number of already executed runs
+    runs_executed: u32,
+    /// Runs in which arguments have min value e.g. `run_with_min_value[0] = 5`
+    /// means that the first argument has the lowest possible value in 5th run
+    runs_with_min_value: Vec<u32>,
+    /// Runs in which arguments have max value e.g. `run_with_min_value[0] = 5`
+    /// means that the first argument has the highest possible value in 5th run
+    runs_with_max_value: Vec<u32>,
```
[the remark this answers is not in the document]
piotmag769: The point is to always have at least one run for each arg where it is set to max and min value, it was requested by @MaksymilianDemitraszek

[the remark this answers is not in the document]
cptartur: I get the idea, but perhaps it can be handled in less complicated way than creating these vectors. Something like just a counter and if in `next_felt252`.

crates/forge/src/fuzzer.rs
```diff
@@ -0,0 +1,223 @@
...
+    /// e.g. `run_with_min_value_argument[0] = 5`
+    /// means that the first argument will have the lowest possible value in 5th run
+    run_with_min_value_for_argument: Vec<u32>,
+    /// Run in which argument has a max value
+    /// e.g. `run_with_max_value_for_argument[0] = 5`
+    /// means that the first argument will have the highest possible value in 5th run
+    run_with_max_value_for_argument: Vec<u32>,
+}
+
+impl Default for RunParams {
```
piotmag769, line 29: Do we use it anywhere?

piotmag769 (approved):
Thanks for refactoring, indeed looks better!

crates/forge/src/fuzzer.rs
cptartur: Only in unit tests, but it makes the code a little bit shorter.

crates/forge/src/fuzzer.rs
```diff
@@ -0,0 +1,223 @@
...
+                    self.rng
+                        .gen_biguint_range(&self.run_params.low, &self.run_params.high)
+                })
+            })
+            .collect()
+    }
+
+    fn next_run(&mut self) {
+        self.run_params.executed_runs += 1;
+    }
```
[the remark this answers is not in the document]
cptartur: I wanted to abstract any operations on these internal counters or structures from the main function.

crates/forge/src/fuzzer.rs
piotmag769: Then I think this code should be in `mod tests`

crates/forge/src/fuzzer.rs
```diff
@@ -0,0 +1,223 @@
...
+        Self {
+            low: BigUint::zero(),
+            high: Felt252::prime(),
+            arguments_number: 0,
+            total_runs: 256,
+            executed_runs: 0,
+            run_with_min_value_for_argument: vec![],
+            run_with_max_value_for_argument: vec![],
+        }
+    }
```
piotmag769, line 40:
```suggestion
        fn default() -> Self {
            Self {
                low: BigUint::zero(),
                high: Felt252::prime(),
                total_runs: 256,
                ..Default::default()
            }
        }
```

Arcticae (approved):

PR #619, opened 2023-09-07
Make deploy panic `Unrecoverably` on invalid calldata

abulenok:
Closes #

## Introduced changes

- deploy fails the test on invalid calldata

## Checklist

- [x] Linked relevant issue
- [ ] Added relevant tests

### DIFF
--- a/crates/cheatnet/src/cheatcodes/deploy.rs
+++ b/crates/cheatnet/src/cheatcodes/deploy.rs
@@ -6,6 +6,7 @@ use blockifier::execution::execution_utils::{felt_to_stark_felt, stark_felt_to_f
 
 use blockifier::state::state_api::{State, StateReader};
 use cairo_felt::Felt252;
+use cairo_lang_runner::short_string::as_cairo_short_string;
 use cairo_vm::vm::errors::hint_errors::HintError::CustomHint;
 use conversions::StarknetConversions;
 
@@ -45,9 +46,9 @@ impl CheatnetState {
             .get_compiled_contract_class(class_hash)
             .map_err::<EnhancedHintError, _>(From::from)?;
         if contract_class.constructor_selector().is_none() && !calldata.is_empty() {
-            return Err(CheatcodeError::Recoverable(vec![
-                "No constructor in contract".to_owned().to_felt252(),
-            ]));
+            return Err(CheatcodeError::Unrecoverable(EnhancedHintError::from(
+                CustomHint(Box::from("No constructor in contract")),
+            )));
         }
 
         let execute_calldata = create_execute_calldata(
@@ -77,6 +78,19 @@ impl CheatnetState {
                     ))))
                 }),
             CallContractOutput::Panic { panic_data } => {
+                let err_str = as_cairo_short_string(&panic_data[0]).unwrap();
+                for invalid_calldata_msg in [
+                    "Failed to deserialize param #",
+                    "Input too long for arguments",
+                    "INVALID_CALLDATA_LEN",
+                ] {
+                    if err_str.contains(invalid_calldata_msg) {
+                        return Err(CheatcodeError::Unrecoverable(EnhancedHintError::from(
+                            CustomHint(Box::from(err_str)),
+                        )));
+                    }
+                }
+
                 Err(CheatcodeError::Recoverable(panic_data))
             }
             CallContractOutput::Error { msg } => Err(CheatcodeError::Unrecoverable(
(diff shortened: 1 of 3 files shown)


crates/cheatnet/src/cheatcodes/deploy.rs
```diff
@@ -77,6 +78,19 @@ impl CheatnetState {
                     ))))
                 }),
             CallContractOutput::Panic { panic_data } => {
+                let err_str = as_cairo_short_string(&panic_data[0]).unwrap();
```
war-in, line 81: Should we check the first element only?

crates/cheatnet/src/cheatcodes/deploy.rs
```diff
@@ -77,6 +78,19 @@ impl CheatnetState {
...
+                for invalid_calldata_msg in [
+                    "Failed to deserialize param #",
+                    "Input too long for arguments",
+                    "INVALID_CALLDATA_LEN",
```
war-in, line 85: I can't see this error in the test. How to obtain it?

war-in (approved):

crates/cheatnet/src/cheatcodes/deploy.rs
abulenok: `INVALID_CALLDATA_LEN` appears when no `constructor` exists in a contract and the calldata.len() != 0

We have a separate `if` for this case in the deploy cheatcode (https://github.com/foundry-rs/starknet-foundry/pull/619/files#diff-acb75fb617fd5f282a6613f7233f2efeb1adc8b67c4bcb4f709d863da752e92aR48-R51)

abulenok: So actually there is no need to include it here

crates/cheatnet/src/cheatcodes/deploy.rs
abulenok: Maybe we should iterate the whole array so as not to miss anything

abulenok: In the case, there are multiple error messages in the panic data

crates/cheatnet/src/cheatcodes/deploy.rs
war-in: Hmm, now I wonder if we should check constructor existence then. Shouldn't we allow blockifier to do it for us?

piotmag769 (approved):

crates/cheatnet/src/cheatcodes/deploy.rs
```diff
@@ -77,6 +78,20 @@ impl CheatnetState {
...
             CallContractOutput::Panic { panic_data } => {
+                for panic_felt in &panic_data {
+                    let err_str = as_cairo_short_string(panic_felt).unwrap();
+                    for invalid_calldata_msg in [
+                        "Failed to deserialize param #",
+                        "Input too long for arguments",
+                    ] {
+                        if err_str.contains(invalid_calldata_msg) {
+                            return Err(CheatcodeError::Unrecoverable(EnhancedHintError::from(
+                                CustomHint(Box::from(err_str)),
```
piotmag769, line 89: We only pass a part of an error here though, what if `panic_data.len() > 1`? We should convert all felts to shortstrings and concatenate them imo

piotmag769 (changes requested):

crates/cheatnet/src/cheatcodes/deploy.rs
abulenok: We can do that, this will simplify the logic making blockifier the only one responsible for the calldata validation

piotmag769 (approved):

crates/cheatnet/src/cheatcodes/deploy.rs
```diff
@@ -77,6 +68,23 @@ impl CheatnetState {
...
             CallContractOutput::Panic { panic_data } => {
+                let panic_data_str = panic_data
+                    .iter()
+                    .map(|x| as_cairo_short_string(x).unwrap())
+                    .collect::<Vec<String>>()
+                    .join("\n");
+
+                for invalid_calldata_msg in [
+                    "Failed to deserialize param #",
+                    "Input too long for arguments",
```
war-in, line 79: I think there was one more (in the case of nonexistent constructor)

abulenok: yes, but it has occurred that `CallContractOutput::Error` is returned from `call_contract`, not `CallContractOutput::Panic`

war-in (approved):

PR #632, opened 2023-09-11
Fix exit code

piotmag769:
## Introduced changes

- Extracted `TestFileSummary` to a separate file

## Breaking changes

- Return non-zero exit code when any tests fail

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests

### DIFF
--- a/crates/forge/src/main.rs
+++ b/crates/forge/src/main.rs
@@ -14,7 +14,9 @@ use forge::scarb::{
     corelib_for_package, dependencies_for_package, get_contracts_map, name_for_package,
     paths_for_package, target_name_for_package, try_get_starknet_artifacts_path,
 };
+use forge::test_case_summary::TestCaseSummary;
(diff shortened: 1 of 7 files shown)


crates/forge/src/main.rs
```diff
@@ -68,12 +79,10 @@ fn main_execution() -> Result<()> {
         .output()
         .context("Failed to build contracts with Scarb")?;
     if !build_output.status.success() {
-        bail!(
-            "Scarb build didn't succeed:\n\n{}",
-            String::from_utf8_lossy(&build_output.stdout)
```
Arcticae, line 73: Why is this removed?

piotmag769: Nope, we redirect both stdout and stderr, that's why I removed it

Arcticae (approved):

Issue #434, opened 2023-08-08
Add `account add` command


war-in:
### To what component is your bug related to?

Cast

### Feature Request

It would be nice to have:
- `account add` 

Someone could import an existing account to the `accounts-file` with such a command. A new file named `add.rs` should be added to https://github.com/foundry-rs/starknet-foundry/tree/c04585092890f3cd351b36b411a4dae651c35feb/crates/cast/src/starknet_commands/account and included in mod.rs placed there. 

This new command should take additional arguments in the form of flags:
- `--name` - a name of an account to be added (required)
- `--private-key` - a private key of an account to be added (required)
- `--public-key` - a public key of an account to be added (required)
- `--address` - an address of an account to be added (required)
- `--network` - a chain id of the network where account to be added was deployed (optional - if not passed we should require --url or --profile with configured rpc url to be passed to cast)
- `--salt` - a salt of an account to be added (optional)
- `--class-hash` - a class hash of an account to be added (optional)

The command should take all the necassary info and put it to the accounts file specified by the user using flag `--accounts-file` passed to cast. An example of a command would be:

```
sncast --accounts-file ~/myfile.json account add --name mariusz --private-key 0x123 --public-key 0x321 --address 0x456 --salt 0x789 --class-hash 0x987 --network alpha-goerli
```

and the resulting record in myfile.json should look like this:

```
{
  "alpha-goerli": {
    "devnet1": {
      "address": "0x456",
      "deployed": true,
      "private_key": "0x123",
      "public_key": "0x321",
      "salt": "0x789",
      "class_hash": "0x987",
    },
}
```

If optional flags were not passed, they should not be set in the file, eg if someone did not pass --class-hash:

```
sncast --accounts-file ~/myfile.json account add --name mariusz --private-key 0x123 --public-key 0x321 --address 0x456 --salt 0x789 --network alpha-goerli
```

should give us:
```
{
  "alpha-goerli": {
    "devnet1": {
      "address": "0x456",
      "deployed": true,
      "private_key": "0x123",
      "public_key": "0x321",
      "salt": "0x789",
    },
}
```

PR #634, opened 2023-09-11
Keystore create

drknzz:
Closes #

## Introduced changes

- **no tests written yet** ⚠️ 
- Allow for creating an account from keystore and starkli account file

### Undeployed account flow

*my_account_undeployed.json*
```
{
    "version": 1,
    "variant": {
        "type": "open_zeppelin",
        "version": 1,
        "public_key": "0xe2d3d7080bfc665e0060a06e8e95c3db3ff78a1fec4cc81ddc87e49a12e0a"
    },
    "deployment": {
        "status": "undeployed",
        "class_hash": "0x646a72e2aab2fca75d713fbe4a58f2d12cbd64105621b89dc9ce7045b5bf02b",
        "salt": "0x14df438ac6825165c7a0af29decd5892528b763a333f93a5f6b12980dbddd9f"
    }
}
```

*command*
```bash
> sncast --url http://127.0.0.1:5050/rpc --accounts-file accounts.json --keystore my_key.json --account my_account_undeployed.json account create --name starkli --from-keystore --add-profile             
Enter password: 
Account successfully created. Prefund generated address with at least 432300000000 tokens. It is good to send more in the case of higher demand, max_fee * 2 = 864600000000
command: account create
add_profile: Profile successfully added to Scarb.toml
address: 0xcce3217e4aea0ab738b55446b1b378750edfca617db549fda1ede28435206c
max_fee: 0x64a7168300
```

*accounts.json*
```
{
  "alpha-goerli": {
    "starkli": {
      "address": "0xcce3217e4aea0ab738b55446b1b378750edfca617db549fda1ede28435206c",
      "class_hash": "0x646a72e2aab2fca75d713fbe4a58f2d12cbd64105621b89dc9ce7045b5bf02b",
      "deployed": false,
      "private_key": "0x55ae34c86281fbd19292c7e3bfdfceb4",
      "public_key": "0xe2d3d7080bfc665e0060a06e8e95c3db3ff78a1fec4cc81ddc87e49a12e0a",
      "salt": "0x14df438ac6825165c7a0af29decd5892528b763a333f93a5f6b12980dbddd9f"
    },
  }
}
```

### Deployed account flow

*my_account.json*
```
{
    "version": 1,
    "variant": {
        "type": "open_zeppelin",
        "version": 1,
        "public_key": "0xe2d3d7080bfc665e0060a06e8e95c3db3ff78a1fec4cc81ddc87e49a12e0a"
    },
    "deployment": {
        "status": "deployed",
        "class_hash": "0x646a72e2aab2fca75d713fbe4a58f2d12cbd64105621b89dc9ce7045b5bf02b",
        "address": "0xcce3217e4aea0ab738b55446b1b378750edfca617db549fda1ede28435206c"
    }
}
```

*command*
```bash
> sncast --url http://127.0.0.1:5050/rpc --accounts-file accounts.json --keystore my_key.json --account my_account.json account create --name starkli_deployed --from-keystore --add-profile
Enter password: 
command: account create
add_profile: Profile successfully added to Scarb.toml
address: 0xcce3217e4aea0ab738b55446b1b378750edfca617db549fda1ede28435206c
max_fee: 0x0
```

*accounts.json*
```
{
  "alpha-goerli": {
    "starkli_deployed": {
      "address": "0xcce3217e4aea0ab738b55446b1b378750edfca617db549fda1ede28435206c",
      "class_hash": "0x646a72e2aab2fca75d713fbe4a58f2d12cbd64105621b89dc9ce7045b5bf02b",
      "deployed": true,
      "private_key": "0x55ae34c86281fbd19292c7e3bfdfceb4",
      "public_key": "0xe2d3d7080bfc665e0060a06e8e95c3db3ff78a1fec4cc81ddc87e49a12e0a"
    }
  }
}
```

## Checklist

- [ ] Linked relevant issue
- [ ] Added relevant tests


crates/cast/src/starknet_commands/account/create.rs
```diff
@@ -44,100 +51,53 @@ pub async fn create(
...
-    let public_key = private_key.verifying_key();
-    let salt = extract_or_generate_salt(maybe_salt);
-    let oz_class_hash: &str = if let Some(value) = &class_hash {
-        value
+    let (account_json, max_fee) = if from_keystore {
+        let keystore_path_ = keystore_path
+            .ok_or_else(|| anyhow!("--keystore must be passed when using --from-keystore"))?;
+        let account_path_ = account_path
+            .ok_or_else(|| anyhow!("--account must be passed when using --from-keystore"))?;
+        import_from_keystore(provider, keystore_path_, account_path_).await?
```
THenry14, line 66: I think `account create` should actually create an account; this looks to me that this is only importing, so should rather be done under https://github.com/foundry-rs/starknet-foundry/issues/434

crates/cast/src/starknet_commands/account/create.rs
```diff
@@ -156,6 +116,110 @@ pub async fn create(
...
+            (None, false) => bail!("No salt field in account JSON file"),
+            (None, true) => None,
+        }
+    };
+
+    let class_hash: FieldElement = {
+        let ch = deployment
+            .get("class_hash")
+            .and_then(serde_json::Value::as_str)
+            .ok_or_else(|| anyhow!("No class_hash field in account JSON file"))?;
```
THenry14, line 148: are those accounts by default OZ? if yes, we could just use their class hash, and not fail.

drknzz: They are 👀

crates/cast/src/starknet_commands/account/create.rs
```diff
@@ -156,6 +116,110 @@ pub async fn create(
...
+        Some("deployed") => true,
+        Some("undeployed") => false,
+        _ => bail!("Unknown deployment status value"),
+    };
+
+    let salt: Option<FieldElement> = {
+        let salt = deployment.get("salt").and_then(serde_json::Value::as_str);
+        match (salt, deployed) {
+            (Some(salt), _) => Some(parse_number(salt)?),
+            (None, false) => bail!("No salt field in account JSON file"),
```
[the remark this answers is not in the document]
drknzz: I think a clear error is better here instead of implicit behaviour.
I don't think this would happen, if someone uses starkli to generate the account json file anyway.

crates/cast/src/starknet_commands/account/create.rs
```diff
@@ -44,100 +51,53 @@ pub async fn create(
...
+        account_json["address"]
+            .as_str()
+            .ok_or_else(|| anyhow!("Invalid address"))?,
+    )?;
 
-    if let Some(class_hash_) = class_hash {
-        items[&network_name][&config.account]["class_hash"] =
-            serde_json::Value::String(format!("{:#x}", parse_number(&class_hash_)?));
-    }
+    write_account_to_accounts_file(
```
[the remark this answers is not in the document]
drknzz: Isn't that the whole point of importing an account?

crates/cast/src/starknet_commands/account/create.rs
drknzz: `account add --from-keystore` might be the place

crates/cast/src/lib.rs
```diff
@@ -74,6 +74,16 @@ fn get_account_info(name: &str, chain_id: FieldElement, path: &Utf8PathBuf) -> R
     account.ok_or_else(|| anyhow!("Account {} not found under network {}", name, network_name))
 }
 
+pub fn get_keystore_password() -> std::io::Result<String> {
```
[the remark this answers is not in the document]
drknzz: Ah, so the merge did change some stuff :S
Good catch

crates/cast/src/starknet_commands/account/create.rs
```diff
@@ -36,6 +39,10 @@ pub struct Create {
     /// Custom open zeppelin contract class hash of declared contract
     #[clap(short, long)]
     pub class_hash: Option<String>,
+
+    /// If passed, create account from keystore and starkli account JSON file
+    #[clap(short, long)]
+    pub from_keystore: bool,
```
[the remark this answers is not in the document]
drknzz: There already is a `--keystore` common flag, wouldn't it be confusing?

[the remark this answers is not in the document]
THenry14: We could re-use it for exactly this purpose right? if `--keystore` was passed and `--account` is a path, we know we import starkli account; otherwise it is standard oz account

crates/cast/src/starknet_commands/account/create.rs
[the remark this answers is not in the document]
THenry14: maybe I misunderstand what is `accounts_file` - for now we used this name for our json account file (~/.starknet_accounts/starknet_open_zeppelin_accounts.json by default). My understanding is that starkli has one account per json file? so we would have multiple json files with different accounts (and different schema to starknet_open_zeppelin_accounts). Is that true?

crates/cast/src/starknet_commands/account/create.rs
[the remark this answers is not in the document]
THenry14: I guess we should take under account also this thread https://github.com/foundry-rs/starknet-foundry/pull/634/files#r1322706719

If we put everything into one file, I don't think we should alter `--account` to be a path in case of starkli account, but rather an account name (as it is right now). If that's the case, only `--keystore` flag would indicate that we're using starkli account.

crates/cast/src/main.rs
```diff
@@ -27,21 +27,21 @@ struct Cli {
...
     /// RPC provider url address; overrides url from Scarb.toml
-    #[clap(short = 'u', long = "url")]
+    #[clap(short, long = "url")]
     rpc_url: Option<String>,
 
     /// Account to be used for contract declaration;
     /// When using keystore (`--keystore`), this should be a path to account file    
     /// When using accounts file, this should be an account name
-    #[clap(short = 'a', long)]
+    #[clap(long, visible_alias = "acc", global = true)]
```
drknzz, line 36: I'll think about this

crates/cast/src/starknet_commands/account/add.rs
```diff
@@ -1,33 +1,50 @@
...
+    /// Name of the account to be added
     #[clap(short, long)]
+    pub name: String,
+
+    /// Address of the account
+    #[clap(
+        short,
+        long,
+        default_value = "0",
+        required_unless_present = "from_keystore"
```
drknzz, line 25: The `default_value` needs to be present for `required_unless_present` to work for some reason
(it doesn't change anything though)

crates/cast/src/starknet_commands/account/add.rs
```diff
@@ -37,6 +54,10 @@ pub struct Add {
     /// If passed, a profile with corresponding data will be created in Scarb.toml
     #[clap(long)]
     pub add_profile: bool,
+
+    /// If passed, import account from keystore and starkli account JSON file
+    #[clap(short, long, conflicts_with_all = ["address", "class_hash", "deployed", "private_key", "public_key", "salt"])]
```
drknzz, line 59: `conflicts_with` makes it not possible to pass from_keystore with any of the arguments

crates/cast/src/starknet_commands/account/create.rs
```diff
@@ -119,28 +119,3 @@ async fn generate_account(
 
     Ok((account_json, max_fee))
 }
-
-async fn get_account_deployment_fee(
```
drknzz, line 123: moved to mod.rs

crates/cast/src/starknet_commands/account/create.rs
drknzz: Thought about it and I think explicit error is better here.
The starkli account json is something that shouldn't be modified manually, hence it should be of the desired schema (and have the class hash among other fields).

crates/cast/src/starknet_commands/account/add.rs
THenry14: why not just do whatever is done with this flag with just `--keystore` argument?

crates/cast/src/starknet_commands/account/add.rs
```diff
@@ -99,3 +122,119 @@ pub async fn add(
...
 }
+
+fn import_from_keystore(
+    keystore_path: &Utf8PathBuf,
+    account_path: &Utf8PathBuf,
+) -> Result<serde_json::Value> {
+    let account_info: serde_json::Value = serde_json::from_str(&fs::read_to_string(account_path)?)?;
+    let deployment = account_info
+        .get("deployment")
+        .ok_or_else(|| anyhow!("No deployment field in account JSON file"))?;
```
THenry14, line 133: theoritically - is it possible? or is it that if it's not present, it means the json file is corrupt/invalid?

crates/cast/src/starknet_commands/account/add.rs
```diff
@@ -99,3 +122,119 @@ pub async fn add(
...
+    let public_key: FieldElement = {
+        let pk = account_info
+            .get("variant")
+            .and_then(|v| v.get("public_key"))
+            .and_then(serde_json::Value::as_str)
+            .ok_or_else(|| anyhow::anyhow!("No public_key in account JSON file"))?;
+        parse_number(pk)?
+    };
+    if public_key != private_key.verifying_key().scalar() {
+        bail!("Public key mismatch");
```
THenry14, line 168:
```suggestion
        bail!("The private key does not match the public key");
```

crates/cast/src/starknet_commands/account/add.rs
```diff
@@ -99,3 +122,119 @@ pub async fn add(
...
+                    "class_hash": "0x646a72e2aab2fca75d713fbe4a58f2d12cbd64105621b89dc9ce7045b5bf02b",
+                    "deployed": false,
+                    "private_key": "0x55ae34c86281fbd19292c7e3bfdfceb4",
+                    "public_key": "0xe2d3d7080bfc665e0060a06e8e95c3db3ff78a1fec4cc81ddc87e49a12e0a",
+                    "salt": "0x14df438ac6825165c7a0af29decd5892528b763a333f93a5f6b12980dbddd9f",
+                }
+            )
+        );
+    }
+}
```
THenry14, line 240: it would be cool to have more tests. I know the description says they will be added later, just want to make sure there will be some more :+1:

CHANGELOG.md
```diff
@@ -23,6 +23,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - `account add` command for importing accounts to the accounts file
 
+### Changed
+
+- `--add-profile` no longer accepts `-a` for short
```
THenry14, line 28: this was done in different PR, right?

crates/cast/src/main.rs
```diff
@@ -214,6 +214,8 @@ async fn main() -> Result<()> {
                     &cli.path_to_scarb_toml,
                     &provider,
                     &add,
+                    cli.keystore,
+                    cli.account.map(Utf8PathBuf::from),
```
THenry14, line 218: this (account) is a part of CastConfig, right? maybe we should not map it here, but rather in the add function itself, when it's needed, and it could be taken from the config?

crates/cast/src/starknet_commands/account/add.rs
```diff
@@ -1,48 +1,63 @@
...
+        default_value = "0",
+        required_unless_present = "keystore",
+        conflicts_with = "keystore"
+    )]
     pub private_key: FieldElement,
 
     /// Account public key
-    #[clap(long)]
+    #[clap(long, visible_alias = "pub")]
     pub public_key: Option<FieldElement>,
```
THenry14, line 57: in your examples in description I can see that public key is included in starkli account file. Wouldn't it make sense to also set is as conflicted with keystore, and just take what's in the file?

crates/cast/src/starknet_commands/account/add.rs
```diff
@@ -55,30 +70,36 @@ pub async fn add(
...
-    } else if provider
-        .get_class_hash_at(BlockId::Tag(Latest), add.address)
-        .await
-        .is_ok()
-    {
-        println!("Contract detected as deployed on chain");
-        true
+    let account_json = if keystore_path.is_some() {
+        let keystore_path_ = keystore_path
+            .ok_or_else(|| anyhow!("--keystore must be passed when using --keystore"))?;
```
THenry14, line 78: this does not say much, let's improve the message

crates/cast/tests/e2e/account/add.rs
```diff
@@ -222,3 +223,101 @@ pub async fn test_missing_arguments() {
...
+        "--private-key",
+        "0x123",
+    ];
+
+    let snapbox = runner(&args);
+    snapbox.assert().stderr_matches(indoc! {r#"
+        error: the argument '--keystore <KEYSTORE>' cannot be used with '--private-key <PRIVATE_KEY>'
+        ...
+    "#});
+}
```
THenry14, line 323: do you think we should also add a test for malformed accound file?

THenry14 (commented):
I have reviewed the code and it looks ok, but I have a one question - didn't we discuss the issues about complying to one schema, and decided not to include starkli accounts in our current accounts file, but rather just use one file-one acocunt approach to simplify things? Perhaps I mixed something up but I was under the impression we would go this way?

PR #650, opened 2023-09-13
[tests] Use dynamically calculated hashes and addresses instead of hard-coded ones

THenry14:
This PR does not close any issue, but is a part of https://github.com/foundry-rs/starknet-foundry/issues/429
Hashes and addresses changed every time we updated devnet/compilers. This is a bit tiresome to update them each time, plus there probably is no need for that. This PR makes the tests to use hashes and addresses taken directly from our code.

## Introduced changes

- do not use hard coded hashes and addresses

## Breaking changes

None

## Checklist

- [X] Linked relevant issue
- [X] Added relevant tests

### DIFF
--- a/crates/cast/tests/helpers/fixtures.rs
+++ b/crates/cast/tests/helpers/fixtures.rs
@@ -1,6 +1,8 @@
-use crate::helpers::constants::{ACCOUNT_FILE_PATH, CONTRACTS_DIR, URL};
+use crate::helpers::constants::{ACCOUNT_FILE_PATH, CONTRACTS_DIR, DEVNET_ENV_FILE, URL};
 use camino::Utf8PathBuf;
+use cast::helpers::constants::UDC_ADDRESS;
 use cast::{get_account, get_provider, parse_number};
+use primitive_types::U256;
 use serde_json::{json, Value};
 use starknet::accounts::{Account, Call};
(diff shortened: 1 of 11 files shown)


war-in (approved):

crates/cast/tests/helpers/fixtures.rs
```diff
@@ -68,7 +74,24 @@ pub async fn declare_deploy_contract(account: &str, path: &str) {
...
+                if event.from_address == FieldElement::from_hex_be(UDC_ADDRESS).unwrap() {
+                    address = event.data.first().copied();
+                    break;
+                }
+            }
+        }
+        _ => {
+            panic!("Unexpected TransactionReceipt variant");
+        }
+    };
```
drknzz, line 93: Wouldn't it be more straight forward to just calculate the address ourselves using `get_contract_address`?

THenry14: It probably would, but I actually wanted to avoid depending on this. No specific reasons, I just thought it may be more lightweight and overall a better idea to just get the values straight from devnet.

THenry14:
> I don't see the .env file anywhere - would it be created when running the tests for the first time?

They are created only for tests and then are removed automatically

PR #660, opened 2023-09-14
Account add

### LINKED ISSUE #434: Add `account add` command

drknzz:
Closes #434 

## Introduced changes

-
-

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/cast/src/starknet_commands/account/add.rs
```diff
@@ -0,0 +1,82 @@
...
+    #[clap(short, long)]
+    pub address: FieldElement,
+
+    /// Class hash of the account
+    #[clap(short, long)]
+    pub class_hash: Option<FieldElement>,
+
+    /// Account deployment status
+    #[clap(short, long)]
+    pub deployed: bool,
```
THenry14, line 23: I'm worried this might be too vague for users. I don't think we should require this to be passed; I suppose we could try to fetch the account to see if it was deployed and include the info by ourselves, or alternatively just don't include this and assume it was deployed.

crates/cast/src/starknet_commands/account/add.rs
```diff
@@ -0,0 +1,82 @@
...
+    chain_id: FieldElement,
+    add: &Add,
+) -> Result<AccountAddResponse> {
+    let private_key = &SigningKey::from_secret_scalar(add.private_key);
+    if let Some(public_key) = &add.public_key {
+        ensure!(
+            public_key == &private_key.verifying_key().scalar(),
+            "public key mismatch"
+        );
+    }
```
THenry14, line 55: What if we just calculated public key from private key?

docs/src/starknet/account.md
```diff
@@ -43,7 +43,7 @@ Do the following to start interacting with the Starknet:
```
THenry14, line 43: We could also add some examples to docs

crates/cast/src/starknet_commands/account/add.rs
drknzz: The argument is to allow passing the public key, but it's being asserted that the public key is actually from the private key. The public key is optional to pass.

crates/cast/src/starknet_commands/account/add.rs
drknzz: I'll think some alternative

crates/cast/src/starknet_commands/account/add.rs
THenry14: yup. but what if we just simplify things, and just calculate the key? then we can get rid of the parameter, and still make sure the public key is always there

drknzz: What prompts me to leave it is that currently all the keys that are in a single account in our accounts file, are present as a flag and available to pass. So if someone were to use the account json with the "public_key" field in there, it wouldn't throw errors, and they wouldn't need to modify the dict 🤔

drknzz: Also it's some layer that just assures the user the public and private key they passed correspond to each other

THenry14: I see your point, I just wanted to make sure UX is the best possible. Maybe we could make this flag optional, and if it's not present calculate it?

>So if someone were to use the account json with the "public_key" field in there, it wouldn't throw errors, and they wouldn't need to modify the dict 🤔

not really sure what you mean by that

drknzz: > Maybe we could make this flag optional, and if it's not present calculate it?

That's exactly what is happening right now 😄


> > So if someone were to use the account json with the "public_key" field in there, it wouldn't throw errors, and they wouldn't need to modify the dict 🤔

> not really sure what you mean by that

I thought that if someone were to take the whole account from the openzeppelin_accounts.json file, and automate it to pass the contents to `account add` it wouldn't complain about public_key being an unknown key

THenry14: >That's exactly what is happening right now 😄

Perhaps I'm blind, but I cannot see where we actually calculate the key here :P I can see that we try to verify it. Where is it done?

>it wouldn't complain about public_key being an unknown key

sure, but what you mean by "it"? I don't think our nor someone's code should do that, if we calculate the key

crates/cast/src/starknet_commands/account/add.rs
```diff
@@ -0,0 +1,101 @@
...
+    pub class_hash: Option<FieldElement>,
+
+    /// Account deployment status
+    /// If not passed, sncast will check whether the account is deployed or not
+    #[clap(short, long)]
+    pub deployed: bool,
+
+    /// Account private key
+    #[clap(long)]
+    pub private_key: FieldElement,
```
THenry14, line 38: I wonder if we should allow for getting this from file as well. Potentially it can be a security issue to have private keys in plain text in terminal, so some people may want to be able to provide it in some more secure way

crates/cast/src/starknet_commands/account/add.rs
```diff
@@ -0,0 +1,101 @@
...
+    config: &CastConfig,
+    path_to_scarb_toml: &Option<Utf8PathBuf>,
+    provider: &JsonRpcClient<HttpTransport>,
+    add: &Add,
+) -> Result<AccountAddResponse> {
+    let private_key = &SigningKey::from_secret_scalar(add.private_key);
+    if let Some(public_key) = &add.public_key {
+        ensure!(
+            public_key == &private_key.verifying_key().scalar(),
+            "public key mismatch"
```
THenry14, line 63:
```suggestion
            "The private key does not match the public key"
```

crates/cast/src/starknet_commands/account/add.rs
```diff
@@ -0,0 +1,101 @@
...
+        );
+    }
+
+    let deployed = if add.deployed {
+        true
+    } else if provider
+        .get_class_hash_at(BlockId::Tag(Latest), add.address)
+        .await
+        .is_ok()
```
THenry14, line 72: can this potentially break cast somehow? I mean some random problems with rpc node, or something

crates/cast/tests/e2e/account/add.rs
```diff
@@ -0,0 +1,177 @@
...
+                    "private_key": "0x5",
+                    "public_key": "0x788435d61046d3eec54d77d25bd194525f4fa26ebe6575536bc6f656656b74c"
+                  }
+                }
+            }
+        )
+    );
+
+    fs::remove_dir_all(Utf8PathBuf::from(accounts_file).parent().unwrap()).unwrap();
+}
```
THenry14, line 177: let's add some more tests, to for example see what happens when one of the arguments is missing/malformed

crates/cast/src/starknet_commands/account/add.rs
drknzz: Maybe in the future? 😅

crates/cast/src/starknet_commands/account/add.rs
drknzz: What exactly do you mean by breaking cast? If the provider request fails for some reason, deployed will be set to false

crates/cast/tests/helpers/constants.rs
```diff
@@ -26,5 +26,7 @@ pub const FEE_CONTRACT_ADDRESS: &str =
     "0x049d36570d4e46f48e99674bd3fcc84644ddd6b96f7c741b1562b82f9e004dc7";
 pub const DEVNET_OZ_CLASS_HASH: &str =
     "0x646a72e2aab2fca75d713fbe4a58f2d12cbd64105621b89dc9ce7045b5bf02b";
+pub const DEVNET_PREDEPLOYED_ACCOUNT_ADDRESS: &str =
+    "0x98459bc7e9740d1b55e5d04c3942badca974f327f7e1bb9d10737c87ea0aec";
```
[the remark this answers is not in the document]
drknzz: I feel like it might be helpful to have it available as const for future tests.

crates/cast/src/starknet_commands/account/add.rs
THenry14: hmm, sure, but let's create an issue for this so it won't get lost

crates/cast/src/starknet_commands/account/add.rs
THenry14: will it panic? we should avoid that

drknzz: It won't, `await` returns Result, and here we just check if the Result is Ok

THenry14 (approved):

PR #683, opened 2023-09-18
Stop requiring sierra artifacts

### LINKED ISSUE #658: Stop requiring sierra contract artifacts
We do not use them anyways

cptartur:
Closes #658

## Introduced changes

- Sierra contract artifacts are no longer required for running `snforge`.

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/cheatnet/tests/contracts/Scarb.toml
```diff
@@ -5,7 +5,6 @@ version = "0.1.0"
 # See more keys and their definitions at https://docs.swmansion.com/scarb/docs/reference/manifest
 
 [[target.starknet-contract]]
-sierra = true
 casm = true
```
piotmag769, line 8: sierra is enabled by default, you have to do `sierra = false` to truly check it. I think we should leave some tests with sierra enable too (at least one) to make sure we always parse the artifacts properly

crates/cheatnet/tests/common/scarb.rs
```diff
@@ -73,7 +73,13 @@ pub fn get_contracts_map(path: &Utf8PathBuf) -> HashMap<String, ContractArtifact
         let casm_path = base_path.join(contract.artifacts.casm);
         let sierra = fs::read_to_string(sierra_path).unwrap();
         let casm = fs::read_to_string(casm_path).unwrap();
-        map.insert(name, ContractArtifacts { sierra, casm });
+        map.insert(
```
piotmag769, line 76: We unwrap above, basically still requiring sierra

crates/cheatnet/tests/contracts/Scarb.toml
cptartur: Some scarb unit tests still use `sierra = true`, I've added a comment explaining that to `scarb::setup_package`.

crates/cheatnet/tests/common/scarb.rs
```diff
@@ -69,11 +69,9 @@ pub fn get_contracts_map(path: &Utf8PathBuf) -> HashMap<String, ContractArtifact
     let mut map = HashMap::new();
     for contract in artifacts.contracts {
         let name = contract.contract_name;
-        let sierra_path = base_path.join(contract.artifacts.sierra);
         let casm_path = base_path.join(contract.artifacts.casm);
-        let sierra = fs::read_to_string(sierra_path).unwrap();
         let casm = fs::read_to_string(casm_path).unwrap();
-        map.insert(name, ContractArtifacts { sierra, casm });
+        map.insert(name, ContractArtifacts { sierra: None, casm });
```
piotmag769, line 74: We can remove this field entirely if we don't use it

piotmag769 (approved):

crates/cheatnet/tests/common/scarb.rs
piotmag769: Same with `StarknetContractArtifacts` from forge

cptartur:
Calculating class_hash in cheatnet should still require sierra. Let's investigate this before this is merged.

crates/cheatnet/src/cheatcodes/mod.rs
```diff
@@ -67,6 +67,6 @@ impl From<EnhancedHintError> for CheatcodeError {
 
 #[derive(Debug, PartialEq, Clone)]
 pub struct ContractArtifacts {
-    pub sierra: String,
+    pub sierra: Option<String>,
```
MaksymilianDemitraszek, line 70: How class hash is calculated without sierra?

piotmag769: Didn't know that, if we need both we can just get both I guess

piotmag769: Okay, so it seems that actually `get_class_hash` is in fact `get_compiled_class_hash` and we use the `compiled_class_hash` instead of `class_hash` for contract identification  xD

cptartur: I'll try changing these so we calculate the correct values

crates/cheatnet/src/cheatcodes/declare.rs
```diff
@@ -96,9 +85,15 @@ impl CheatnetState {
...
 
-fn get_class_hash(casm_contract: &str) -> Result<ClassHash> {
-    let compiled_class = serde_json::from_str::<CompiledClass>(casm_contract)?;
-    let class_hash = compiled_class.class_hash()?;
+fn get_class_hash(sierra_contract: &str) -> Result<ClassHash> {
+    let sierra_class: ContractClass = serde_json::from_str(sierra_contract)?;
+    let abi_flattened = sierra_class.abi.unwrap().json();
+    let mut sierra_contract_map: HashMap<String, Value> = serde_json::from_str(sierra_contract)?;
+    sierra_contract_map.insert("abi".to_string(), Value::String(abi_flattened));
+    let sierra_contract = serde_json::to_string_pretty(&sierra_contract_map)?;
```
cptartur, line 93: This is necessary, because scarb outputs abi as json, and blockifier expects it to just be a plain string (abi format is not enforced for cairo 1 contracts, this is the case on starknet as well).

piotmag769 (approved):

abulenok (approved):

Arcticae (approved):

PR #694, opened 2023-09-19
0.6.1

piotmag769:
TODO: still have to fix the failing test. It is not on Scarb, that was my mistake.


cptartur (changes requested):
If we are adding a new feature as big as Scarb workspaces + changing the way test collecting behaves, it should be 0.7.0 not 0.6.1

Also, #683 could be included in this one

MaksymilianDemitraszek:
@cptartur good call, I think we should postpone it until next wednesday. (It was my call to release it)

PR #698, opened 2023-09-20
Fix test failing on release

piotmag769:
Checked manually, tests work when changing forge's version
## Checklist

- [x] Linked relevant issue
- [ ] Added relevant tests

### DIFF
--- a/crates/forge/src/init.rs
+++ b/crates/forge/src/init.rs
@@ -1,11 +1,12 @@
 use anyhow::{anyhow, Context, Ok, Result};
 
 use include_dir::{include_dir, Dir};
-use scarb_metadata::MetadataCommand;
 
+use regex::Regex;
 use std::fs;
 use std::path::{Path, PathBuf};
 use std::process::{Command, Stdio};
+use std::str::from_utf8;
 use toml_edit::{ArrayOfTables, Document, Item, Table};
 
 static TEMPLATE: Dir = include_dir!("starknet_forge_template");
@@ -75,6 +76,7 @@ pub fn run(project_name: &str) -> Result<()> {
 
     Command::new("scarb")
         .current_dir(&project_path)
+        .arg("--offline")
         .arg("add")
         .arg("snforge_std")
         .arg("--git")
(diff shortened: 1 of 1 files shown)


crates/forge/src/init.rs
```diff
@@ -86,14 +88,33 @@ pub fn run(project_name: &str) -> Result<()> {
...
+        .range();
+    let cairo_plus_version = &version_output.to_owned()[cairo_plus_version_range];
+
+    let cairo_version_regex =
+        Regex::new(r"[0-9]+.[0-9]+.[0-9]+").expect("Could not create cairo version matching regex");
+    let cairo_version_range = cairo_version_regex
+        .find(cairo_plus_version)
+        .expect("Could not find cairo version")
+        .range();
+    let cairo_version = &cairo_plus_version[cairo_version_range];
```
cptartur, line 113:
```suggestion
    let cairo_version_regex =
        Regex::new(r"(?:cairo:?\s*)([0-9]+.[0-9]+.[0-9]+)").expect("Could not create cairo version matching regex");
    let cairo_version_capture = cairo_version_regex.captures(version_output)
        .expect("Could not find cairo version");
    let cairo_version = cairo_version_capture.get(1).expect("Could not find cairo version).as_str();
```

Something alone this lines should work and is a bit simpler

piotmag769: Good point, thx

crates/forge/src/init.rs
```diff
@@ -96,21 +96,15 @@ pub fn run(project_name: &str) -> Result<()> {
...
-        Regex::new(r"[0-9]+.[0-9]+.[0-9]+").expect("Could not create cairo version matching regex");
-    let cairo_version_range = cairo_version_regex
-        .find(cairo_plus_version)
+    let cairo_version_regex = Regex::new(r"(?:cairo:?\s*)([0-9]+.[0-9]+.[0-9]+)")
+        .expect("Could not create cairo version matching regex");
+    let cairo_version_capture = cairo_version_regex
+        .captures(version_output)
+        .expect("Could not find cairo version");
+    let cairo_version = cairo_version_capture
+        .get(1)
```
cptartur, line 105: An `expect` here would be good I think - we should panic if there is no version captured.

PR #701, opened 2023-09-20
Account create + deploy keystore

### LINKED ISSUE #641: Create new starkli-style account

drknzz:
Closes #641 

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/cast/src/helpers/constants.rs
```diff
@@ -24,3 +24,4 @@ pub const DEFAULT_RETRIES: u8 = 12;
 pub const DEFAULT_ACCOUNTS_FILE: &str = "~/.starknet_accounts/starknet_open_zeppelin_accounts.json";
 
 pub const KEYSTORE_PASSWORD_ENV_VAR: &str = "KEYSTORE_PASSWORD";
+pub const TEST_KEYSTORE_PASSWORD_ENV_VAR: &str = "TEST_KEYSTORE_PASSWORD";
```
THenry14, line 27: why do we need a test variable in cast?

crates/cast/src/lib.rs
```diff
@@ -102,14 +127,14 @@ pub async fn get_account<'a>(
     account: &str,
     accounts_file: &Utf8PathBuf,
     provider: &'a JsonRpcClient<HttpTransport>,
-    keystore: &Option<Utf8PathBuf>,
+    keystore: &Utf8PathBuf,
```
THenry14, line 130: shouldn't it stay optional? this also takes non-starkli accounts

crates/cast/src/starknet_commands/account/create.rs
```diff
@@ -43,34 +45,56 @@ pub async fn create(
...
-        &config.rpc_url,
-        &config.account,
-        &config.accounts_file,
-        chain_id,
-        account_json.clone(),
-        add_profile,
-    )?;
+    if let Some(keystore_path_) = &keystore_path {
+        let account_path_ = account_path
+            .ok_or_else(|| anyhow!("--account must be passed when using --keystore"))?;
```
THenry14, line 69:
```suggestion
            .ok_or_else(|| anyhow!("--account must be passed and be a path when using --keystore"))?;
```
?

crates/cast/src/starknet_commands/account/mod.rs
```diff
@@ -135,7 +138,16 @@ pub fn add_created_profile_to_configuration(
             Value::String(config.accounts_file.to_string()),
         );
 
-        tool_sncast.insert(config.account.clone(), Value::Table(new_profile));
```
THenry14, line 138: should accounts-file be present in the toml file if we used keystore? I think that, unless specifically added by multiple, various account create commands, we should only add what's needed to toml file (so eg no keystore when creating "classic" account, and probably no accounts-file otherwise)

docs/src/starknet/account.md
```diff
@@ -132,3 +132,17 @@ $ sncast \
...
+
+```shell
+$ sncast \
+    --url http://127.0.0.1:5050 \
+    --keystore my_key.json \
+    --account my_account.json \
+    account create
+```
+
+The command above will generate a keystore file containing the private key, as well as an account file containing the openzeppelin account info.
```
THenry14, line 148: let's maybe also include an info, that such accounts can later be used with starkli?

crates/cast/src/starknet_commands/account/deploy.rs
```diff
@@ -30,9 +35,143 @@ pub struct Deploy {
...
+    let salt = FieldElement::from_hex_be(
+        items
+            .get("deployment")
+            .and_then(|deployment| deployment.get("salt"))
+            .and_then(serde_json::Value::as_str)
+            .ok_or_else(|| anyhow::anyhow!("Failed to get salt from account JSON file"))?,
+    )?;
+    let oz_class_hash = FieldElement::from_hex_be(
+        items
+            .get("deployment")
```
THenry14, line 118: it seems to me we could just have a `deployment` variable that we can take various things from, instead of doing .get("deployment") multiple times. this is just a nitpick though, I leave the decision on changing this up to you.

crates/cast/src/starknet_commands/account/deploy.rs
```diff
@@ -30,9 +35,143 @@ pub struct Deploy {
...
+        provider,
+        oz_class_hash,
+        private_key,
+        salt,
+        chain_id,
+        max_fee,
+        wait,
+    )
+    .await?;
+
```
THenry14, line 165: it would be cool if this could be more atomic. deploy_oz_account could fail, leaving us with an account file that says the account was deployed, even though it wasn't. There are multiple ways we could approach this problem, let me know if you need any help figuring this out.

crates/cast/src/starknet_commands/account/deploy.rs
```diff
@@ -30,9 +35,143 @@ pub struct Deploy {
...
+    std::fs::write(&account_path, serde_json::to_string_pretty(&items).unwrap())
+        .context("Couldn't write to account file")?;
```
THenry14, line 167: what are possible scenarios here? file already exists? path is a directory? something else? 
We should get ready for at least some of the scenarios, and prevent the once we can prevent. it would be a shame if users lots some funds, because we failed to write necessary info to a file. And my understanding is that it could happen, since we removed salt and it is required for the deployment. right?

crates/cast/src/starknet_commands/account/mod.rs
```diff
@@ -57,11 +57,13 @@ pub fn prepare_account_json(
     account_json
 }
 
+#[allow(clippy::too_many_arguments)]
 pub fn write_account_to_accounts_file(
     path_to_scarb_toml: &Option<Utf8PathBuf>,
     rpc_url: &str,
     account: &str,
     accounts_file: &Utf8PathBuf,
+    keystore: &Utf8PathBuf,
```
THenry14, line 66: sorry, I'm a bit confused - up until now I was under the impression we use separate files for starkli accounts, and that is the only scenario we want to support. If that's true, why we need keystore in accounts file?

crates/cast/tests/e2e/account/create.rs
```diff
@@ -234,3 +235,131 @@ pub async fn test_account_already_exists() {
...
+
+    let snapbox = runner(&args);
+
+    snapbox.assert().stderr_matches(indoc! {r#"
+        command: account create
+        error: --account must be passed and be a path when using --keystore
+    "#});
+
+    _ = fs::remove_file(keystore_path);
+}
```
THenry14, line 365: maybe we should also test when there is already an account/keystore file?

crates/cast/tests/e2e/account/deploy.rs
```diff
@@ -266,3 +267,117 @@ pub async fn create_account(salt: &str, add_profile: bool) -> (Utf8PathBuf, &str
...
+    ];
+
+    let snapbox = Command::new(cargo_bin!("sncast")).args(args);
+    snapbox.assert().stderr_matches(indoc! {r#"
+        command: account deploy
+        error: Public key and private key from keystore do not match
+    "#});
+
+    _ = fs::remove_file(account_path);
+}
```
THenry14, line 383: let's also test: when files are corrupt, when one of the files does not exist

THenry14: maybe even some/all the optional flags (I mean keystore flags + classic accounts flags) are present, let's make sure evertything works as expected

docs/src/starknet/account.md
```diff
@@ -132,3 +132,17 @@ $ sncast \
 ```
 
 For a detailed CLI description, see [account add command reference](../appendix/cast/account/add.md).
+
+### Creating account with keystore
```
THenry14, line 136:
```suggestion
### Creating an account with starkli-style keystore
```

THenry14 (changes requested):

THenry14 (commented):
let's fix tests :+1:

crates/cast/src/starknet_commands/account/create.rs
```diff
@@ -144,3 +169,54 @@ async fn get_account_deployment_fee(
...
+    private_key: FieldElement,
+    salt: FieldElement,
+    class_hash: FieldElement,
+    keystore_path: &Utf8PathBuf,
+    account_path: &Utf8PathBuf,
+    add_profile: bool,
+    config: &CastConfig,
+    path_to_scarb_toml: &Option<Utf8PathBuf>,
+) -> Result<()> {
+    let password = get_test_keystore_password()?;
```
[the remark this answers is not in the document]
drknzz: If you have any other idea, I'm down to change this.
The problem is that, when creating a keystore, in tests, the password needs to be passed somehow, and it's done through the TEST_KEYSTORE_PASSWORD env var.
The idea is that a user could use the KEYSTORE_PASSWORD env var themself, but shouldn't use the TEST_KEYSTORE_PASSWORD env var.

crates/cast/src/starknet_commands/account/create.rs
```diff
@@ -144,3 +169,54 @@ async fn get_account_deployment_fee(
...
+            "status": "undeployed",
+            "class_hash": format!("{class_hash:#x}"),
+            "salt": format!("{salt:#x}"),
+        }
+    });
+
+    if add_profile {
+        add_created_profile_to_configuration(path_to_scarb_toml, config)?;
+    }
+    write_account_to_file(&oz_account_json, account_path)
```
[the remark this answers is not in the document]
drknzz: I thought the idea was to do it in this order, as it was done like this for non keystore accounts.
Good, it will actually simplify things 👍

crates/cast/src/starknet_commands/account/deploy.rs
drknzz: 🤔 not sure I am following - if `deploy_oz_account` fails the error gets propagated through `?` and no changes in files are made

crates/cast/src/starknet_commands/account/deploy.rs
drknzz: The salt has been removed and address added, yes.
If the write fails here though, the file wouldn't be changed though, would it?

crates/cast/src/starknet_commands/account/mod.rs
drknzz: My mistake, it's not needed here.

crates/cast/src/helpers/constants.rs
THenry14: Sure, but shouldn't it be declared in tests then? we could allow for variable name manipulation, which would also allow us to remvoe the duplicated code you've linked. Or, we could even set env variable in test and unset it when we're done, if that's more preferable.

crates/cast/src/lib.rs
THenry14: Ok, I see that this is only used in main, so I guess it's passable :+1:

crates/cast/src/starknet_commands/account/create.rs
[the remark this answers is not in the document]
THenry14: Sure, I've pitched some ideas in https://github.com/foundry-rs/starknet-foundry/pull/701/files#r1332646434
Allowing for custom env name would probably be the simplest, setting and unsetting per test maybe a bit more elegant

crates/cast/src/starknet_commands/account/deploy.rs
THenry14: yes you're right, sorry, brainfart
obraz

crates/cast/src/starknet_commands/account/deploy.rs
THenry14: yup, but since we already deployed an account, doesn't that mean we loose all the information about deployed account, because it was not written to a file? and the user cannot re-deploy it, because it already exists? and in order to deploy, user had to prefund, right? so one kind of looses money

drknzz: The account file does not change, so it's still a valid file for account deployment - but you are right, the deployment would fail because it's already deployed.
What we could do is maybe check if the account is already deployed, and if so just adjust the file without sending a second tx.

THenry14 (approved):
🔥

PR #709, opened 2023-09-21
Define Scarb version for devs

### LINKED ISSUE #590: Scarb version for snfoundry devs
We should have one source of truth for scarb version used by CI.

We should warn dev when running tests that local version is different than one on CI.

cptartur:
Closes #590 

## Introduced changes

- Defined top-level `.tool-versions` file specifying scarb version
- Simplified devnet-rs caching logic
- Renamed `prepare_for_tests.sh` to `install_devnet.sh` 

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


cptartur:
> @cptartur is this going to be worked on, or can be closed?

I think I'll be able to salvage this after (I think) #812 is merged.

DelevoXDG (approved):
👍🏻

.github/workflows/ci.yml
```diff
@@ -89,26 +89,13 @@ jobs:
...
-        run: |
-          devnetrev=$(grep "DEVNET_REV=" ./scripts/prepare_for_tests.sh | awk -F '=' '{print $2}')
-          echo "devnet_rev=${devnetrev//\"}" >> $GITHUB_OUTPUT
-      - name: Cache starknet-devnet-rs
-        id: cache-devnet
-        uses: actions/cache@v3
         with:
-          path: ./crates/cast/tests/utils/devnet
-          key: starknet-devnet-rs-${{ steps.devnet-rev.outputs.devnet_rev }}
+          cache-directories: crates/cast/tests/utils/devnet
```
THenry14, line 93: how does this work? I can't see a cache ket, so is it somehow automatically generated? will this cache change when any of the other cached deps changes? will this change when we up the rev of devnet?

cptartur: We don't define keys for any other usages of `Swatinem/rust-cache` action so I think automatic generation should be fine. I'll test how it behaves when deps change and when devnet rev changes.

THenry14: ok, so there is automatic key generation, I assume based on hash? 
that would be great, I'd really like to avoid a situation where changing one of the deps forces rebuild of devnet (and vice-versa), as this compilation takes quite some time

THenry14: that's awesome, thanks for taking the time and checking this!

THenry14 (approved):

Arcticae (approved):

PR #712, opened 2023-09-22
Extracted test runner into a seperate crate

### LINKED ISSUE #679: Move test running logic and hint processor to a separate crate, under `snforge-core` team responsibility

Radinyn:
Closes #679


Radinyn:
In my opinion moving the hint processor to Cheatnet is a better solution

PR #720, opened 2023-09-25
Write a test which uses get_block_hash_syscall

### LINKED ISSUE #685: Investigate if `get_block_hash_syscall` behaves correctly
Investigate if it behaves correctly in contracts deployed on cheatnet

Arcticae:
Closes #685 

## Introduced changes

- Adds a test which verifies that this function works inside of a contract

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


piotmag769 (approved):

MaksymilianDemitraszek (commented):
Can we also add a test in integration/test_state?

Arcticae:
@MaksymilianDemitraszek [Add integration test](https://github.com/foundry-rs/starknet-foundry/pull/720/commits/193bf0e558d7137f7f4489872b6064849b62be4e)

crates/cheatnet/tests/contracts/src/starknet/blocker.cairo
```diff
@@ -29,6 +32,9 @@ mod Blocker {
             self.block_number.write(block_info.block_number);
             self.block_timestamp.write(block_info.block_timestamp);
             self.sequencer_address.write(block_info.sequencer_address);
+
+            let block_hash = get_block_hash_syscall(block_info.block_number - 10).unwrap_syscall();
```
cptartur, line 36: Why are we getting the block hash of 10 blocks before and not the current one?

Arcticae: https://docs.starknet.io/documentation/architecture_and_concepts/Smart_Contracts/system-calls-cairo1/#get_block_hash

cptartur: Should we have some test verifying that we throw an error if the block number is greater than `current_block - 10` then?

Arcticae: I don't think that testing syscalls behavior is in scope here

cptartur (approved):

crates/cheatnet/tests/contracts/src/starknet/blocker.cairo
cptartur: Should it be done at all? Or is the logic just that blockifier should be handling this properly so we don't have to check?

Arcticae: yes, this part is performed by blockifier

Issue #721, opened 2023-09-25
Add a cheatcode for mocking `get_block_hash_syscall`


Arcticae:
Not requested by anybody, not high prio

Arcticae:
Duplicate of #684

PR #725, opened 2023-09-26
Add support for `get_block_hash_syscall` in the test state

### LINKED ISSUE #686: Make `get_block_hash` behaviour consistent with cheatnet behaviour in tests
blocked by #685 
Currently this syscall is turned off in tests but considering the behaviour in cheatnet we should make it work.

Arcticae:
Closes #686 

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/forge/src/cheatcodes_hint_processor.rs
```diff
@@ -543,20 +553,40 @@ fn execute_syscall(
...
         ))),
         DeprecatedSyscallSelector::ReplaceClass => Err(HintError::CustomHint(Box::from(
-            "Replace class can't be used in tests"
-                .to_string(),
+            "Replace class can't be used in tests".to_string(),
         ))),
-        DeprecatedSyscallSelector::GetBlockHash => Err(HintError::CustomHint(Box::from(
-            "Get block hash is temporarily disabled in tests: https://github.com/foundry-rs/starknet-foundry/issues/686"
-                .to_string(),
-        ))),        _ => blockifier_syscall_handler.execute_hint(vm, exec_scopes, hint_data, constants),
```
piotmag769, line 556: Simply redirecting it doesn't work? Or we want to explicitly set the `BlockHash` to 0?

crates/forge/src/cheatcodes_hint_processor.rs
```diff
@@ -543,20 +553,40 @@ fn execute_syscall(
...
+    let _selector = stark_felt_from_ptr(vm, system_ptr)?;
+    let SyscallRequestWrapper {
+        gas_counter,
+        request: _,
+    } = SyscallRequestWrapper::<GetBlockHashRequest>::read(vm, system_ptr)?;
+
+    let sc_response = SyscallResponseWrapper::Success {
+        gas_counter: gas_counter - GET_BLOCK_HASH_GAS_COST,
+        response: GetBlockHashResponse {
+            block_hash: BlockHash(StarkFelt::from(0_u32)),
```
war-in, line 583: Why block_hash is always 0? 👀

crates/forge/src/cheatcodes_hint_processor.rs
Arcticae: Doesn't work since the storage is uninitialized (block hash is being stored on-chain)
We don't really have the concept of a block, and it always returns 0 in the contracts, so i'm just aligning it with the implementation here.

(I think it's related to [this one](https://github.com/foundry-rs/starknet-foundry/issues/732))
So when the above is fixed, we can just redirect, but the result will be the same still. 
To mock the hash, we can implement a cheatcode: https://github.com/foundry-rs/starknet-foundry/issues/721

crates/forge/src/cheatcodes_hint_processor.rs
Arcticae: https://github.com/foundry-rs/starknet-foundry/pull/725#discussion_r1338419739

piotmag769 (approved):

war-in (approved):

PR #705, opened 2023-09-21
Add cheatcode for reading from environment variables

cptartur:
Closes #638 

## Introduced changes

- Added a cheatcode for reading `felt252` values from environment variables

## Checklist

- [x] Linked relevant issue
- [x] Updated relevant documentation
- [x] Added relevant tests
- [x] Performed self-review of the code
- [x] Added changes to `CHANGELOG.md`

PR #727, opened 2023-09-26
Remove reimport

### LINKED ISSUE #718: Do not reimport functions from modules in `snforge_std` `lib.cairo`
Functions from modules like `io` (`fs` #716) should not be reimported in the `lib.cairo` file, so we enforce rust-like importing of these e.g.

```cairo
use snforge_std::io;

// ...
io::read_txt(...);
```

or 

```cairo
use snforge_std::io::read_txt;
```

Instead of current

```cairo
use snforge_std::read_txt;
```

To resolve this, just remove unnecessary `use` clauses from the `lib.cairo` file (e.g. remove `use io::file_operations::read_json;`, leaving just `mod io` definition).

Utilitycoder:
Closes #718

## Introduced changes

- Removed function reimport from snforge_std's lib.cairo
- Updated relevant documents to effect the changes

## Checklist

- [x] Linked relevant issue
- [ ] Added relevant tests


crates/forge/Cargo.toml
```diff
@@ -46,6 +46,7 @@ scarb-metadata = { git = "https://github.com/software-mansion/scarb.git", rev =
 project-root = "0.2.2"
 console = "0.15.7"
 snapbox = "0.4.11"
+starknet = "0.5.0"
```
piotmag769, line 49: Please revert this change + changes made to `Cargo.lock`

piotmag769 (commented):
I'm pretty sure we want the imports from `cheatcodes` module to stay

crates/forge/Cargo.toml
Utilitycoder: I assumed the cheactcode would follow the same style as io. I will fix up.

crates/forge/Cargo.toml
```diff
@@ -64,4 +64,4 @@ name = "forge"
 
 [[bin]]
 name = "snforge"
-path = "src/main.rs"
+path = "src/main.rs"
```
cptartur, line 67: Newline missing here

docs/src/testing/contracts.md
```diff
@@ -53,7 +53,7 @@ Note that the name after `mod` will be used as the contract name for testing pur
 Let's write a test that will deploy the `HelloStarknet` contract and call some functions.
 
 ```rust
-use snforge_std::{ declare, ContractClassTrait };
+use snforge_std::contract_class::{ declare, ContractClassTrait };
```
cptartur, line 56: I think `declare` and `ContractClassTrait` should be available directly from `snforge_std`

snforge_std/src/lib.cairo
```diff
@@ -33,13 +36,4 @@ use cheatcodes::stop_warp;
...
-
 mod io;
-
-use io::file_operations::File;
-use io::file_operations::FileTrait;
-use io::file_operations::read_txt;
-use io::file_operations::read_json;
-use io::file_operations::FileParser;
-
-use io::forge_print::PrintTrait;
```
cptartur, line 45: Additionally to these changes, lines like this should be added to `io.cairo` file:

```cairo
use io::file_operations::File;
use io::file_operations::FileTrait;
// ... etc
```

This way you will be able to use imports like this in tests code:

```cairo
use io::File;
```

So we are enforcing using the module name in path, but do not force the user to write the whole file name for each function.

snforge_std/src/lib.cairo
```diff
@@ -15,6 +15,9 @@ use cheatcodes::tx_info::stop_spoof;
 use cheatcodes::l1_handler::L1Handler;
 use cheatcodes::l1_handler::L1HandlerTrait;
 
+use cheatcodes::fork::BlockTag;
+use cheatcodes::fork::BlockId;
```
piotmag769, line 19: Good catch, I don't think we want to import those. We can just fix the imports instead

snforge_std/src/lib.cairo
cptartur: You can check out my code in this PR: https://github.com/foundry-rs/starknet-foundry/pull/705/files#diff-65daba458203fdcab6033fc02561c249c5c8f796ba4938e1741185434837e9fe to see how this can be approached.

cptartur (changes requested):

snforge_std/src/lib.cairo
cptartur: Here's an example usage: https://github.com/foundry-rs/starknet-foundry/pull/705/files#diff-ed0161129846964bde0ae2888f9ec23fbed5498cb4e733d8f6cfb9118107b678R11

piotmag769 (commented):
Thanks a lot for the contribution! Last thing to do is to fix the imports in forge tests (that's why they fail now). Lemme know if you want to try to fix the tests as well

Utilitycoder (commented):
I believe the changes have been effected.

docs/src/testing/using-cheatcodes.md
```diff
@@ -100,7 +100,7 @@ so it passes our validation.
 ### Pranking the Address
 
 ```rust
-use snforge_std::{ declare, ContractClassTrait, start_prank };
+use snforge_std::{ declare, ContractClassTrait, start_prank};
```
cptartur, line 103:
```suggestion
use snforge_std::{ declare, ContractClassTrait, start_prank };
```

cptartur (approved):
👍

PR #730, opened 2023-09-26
Support different integer types in fuzzer

### LINKED ISSUE #693: Support for different arg types in fuzzing
Currently we only support `felt252`

cptartur:
Closes #693 

## Introduced changes

-
-

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/forge/src/lib.rs
```diff
@@ -65,6 +62,19 @@ pub struct RunnerConfig {
...
+        Self {
+            test_name_filter: None,
+            exact_match: false,
+            exit_first: false,
+            fork_targets: vec![],
+            fuzzer_runs: 256,
+            fuzzer_seed: 12345,
+        }
+    }
+}
```
piotmag769, line 76: Do we use it anywhere?

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 #### Added
 
 - `var` library function for reading environmental variables
+- Support for `u8`, `u16`, `u32`, `u64`, `u128`, `u256` types to fuzzer
```
piotmag769, line 15: Can we add support for i8, i16 etc. as well?

crates/forge/tests/integration/fuzzing.rs
piotmag769, line 1: Is there a point of having the same tests in both integration and e2e? I think e2e tests are enough, they test the correctness of the behaviour anyways

piotmag769 (commented):
Really well abstracted, gj

crates/forge/src/fuzzer.rs
```diff
@@ -1,16 +1,124 @@
...
+    U256,
+    Felt252,
+}
+
+trait Argument {
+    fn low(&self) -> BigUint;
+    fn high(&self) -> BigUint;
+    fn gen(&self, rng: &mut StdRng) -> Vec<Felt252>;
+    fn min(&self) -> Vec<Felt252>;
+    fn max(&self) -> Vec<Felt252>;
```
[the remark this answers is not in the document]
cptartur: For the most part yes, but e.g. for u256 we are both interested in it's max possible value as BigUint as well as it's max Felt252 value (a vector with two large felts).

If you have a better idea for abstracting these I'm open to suggestions though.

crates/forge/src/fuzzer.rs
```diff
@@ -138,12 +235,42 @@ mod tests {
...
+            run_params,
+        };
+
+        let args1 = fuzzer.next_args();
+        let args2 = fuzzer.next_args();
+        let args3 = fuzzer.next_args();
+
+        assert!(all_values_different(&args1, &args2));
+        assert!(all_values_different(&args1, &args3));
+        assert!(all_values_different(&args2, &args3));
```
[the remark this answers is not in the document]
cptartur: Just testing if we are generating different values between runs. I know it seems kind of stupid, but I had a logic bug (i was cloning the rng somewhere) that caused the values to be exactly the same between runs.

This tests would catch a situation like this.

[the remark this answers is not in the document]
piotmag769: Would add a comment, other than that it's fine

crates/forge/src/fuzzer.rs
```diff
@@ -1,16 +1,124 @@
...
+
+impl CairoType {
+    fn from_name(name: &str) -> Result<Self> {
+        match name {
+            "u8" => Ok(Self::U8),
+            "u16" => Ok(Self::U16),
+            "u32" => Ok(Self::U32),
+            "u64" => Ok(Self::U64),
+            "u128" => Ok(Self::U128),
+            "u256" | "core::integer::u256" => Ok(Self::U256),
```
piotmag769, line 109: I assume it cannot happen in other cases as those are builtin types, right?

crates/forge/src/lib.rs
cptartur: Only in tests, I'll remove it

CHANGELOG.md
cptartur: I think we can do that, but I'd do it in a separate PR, since it will require some minimal changes to the logic so we handle negative values.

crates/forge/tests/integration/fuzzing.rs
cptartur: I've added some integration tests for debugging purposes, it's not easy with e2e tests. Maybe it should be simplified though?

crates/forge/src/fuzzer.rs
cptartur: I think so. Actually `u256` is I think always represented as `"core::integer::u256"` only. But I've added `u256` for consistency sake + it doesn't hurt us to support that as well.

crates/forge/tests/integration/fuzzing.rs
piotmag769: I guess we can simplify it if it is just for debugging purposes

crates/forge/src/fuzzer.rs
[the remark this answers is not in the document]
cptartur: I've updated this code so these are not the same anymore. `High` is and exclusive upper bound now so it works good with the rng. `Max` is actual maximum value that can be used.

piotmag769 (approved):

CHANGELOG.md
cptartur: #755

PR #739, opened 2023-09-27
Trigger docs build only on released release

### LINKED ISSUE #452: Don't build docs when prerelease is published

cptartur:
Closes #452 

## Introduced changes

- Changed the `release` trigger type for docs build from `published` to `released`. This way docs should only build once we actually publish the release and not when draft is created.

## Checklist

- [ ] Linked relevant issue
- [ ] Added relevant tests

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -5,7 +5,7 @@ name: Deploy mdBook site to Pages
 on:
   # Runs on release
   release:
-    types: [ published ]
+    types: [ released ]
 
   # Allows you to run this workflow manually from the Actions tab
   workflow_dispatch:


.github/workflows/docs.yml
```diff
@@ -5,7 +5,7 @@ name: Deploy mdBook site to Pages
 on:
   # Runs on release
   release:
-    types: [ published ]
+    types: [ released ]
```
cptartur, line 8: A possible problem (change?) with this approach: Docs will only start building after we manually publish the release. Previously the docs workflow triggered immediately after the release one successfully created a pre-release.

piotmag769 (approved):

.github/workflows/docs.yml
THenry14: should be fine; do you happen to know if that works for draft releases (or how do they call it) as well?

Arcticae (approved):

.github/workflows/docs.yml
cptartur: [The docs](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#release) say that 
```
Note: Workflows are not triggered for the created, edited, or deleted activity types for draft releases. When you create your release through the GitHub browser UI, your release may automatically be saved as a draft.
```
So I'm assuming drafts won't trigger anything.

THenry14 (approved):

PR #743, opened 2023-09-27
Add delete account feature to remove an account from the accounts file.

### LINKED ISSUE #498: Add `account delete` command
### To what component is your feature request related to?

Cast

### Feature Request

It would be nice to have:
- `account delete` 

Someone could delete an existing account from the `accounts-file` with such a command. A new file named `delete.rs` should be added to https://github.com/foundry-rs/starknet-foundry/tree/c04585092890f3cd351b36b411a4dae651c35feb/crates/cast/src/starknet_commands/account and included in mod.rs placed there. 

This new command should take additional arguments in the form of flags:
- `--name` - a name of an account to be added (required)
- `--network` - a chain id of the network where account to be added was deployed (optional - if not passed we should require --url or --profile with configured rpc url to be passed to cast)

The command should take all the necassary info and remove the account (with all its data) from the accounts file specified by the user using flag `--accounts-file` passed to cast. An example of a command would be:

```
sncast --accounts-file ~/myfile.json account delete --name mariusz --network alpha-goerli
```

devnet0x:
Closes #498 

## Introduced changes

Add account delete feature to remove an account from a network in the accounts file.

## Breaking changes

N/A

## Checklist

- [ X] Linked relevant issue
- [ X] Updated relevant documentation
- [ X] Added relevant tests
- [ X] Performed self-review of the code
- [ X] Added changes to `CHANGELOG.md`


crates/cast/src/starknet_commands/account/delete.rs
```diff
@@ -0,0 +1,57 @@
...
+use serde_json::Value;
+
+#[derive(Args, Debug)]
+#[command(about = "Delete an account with all important secrets")]
+pub struct Delete {
+    /// Name of the account to be deployed
+    #[clap(short, long)]
+    pub name: Option<String>,
+
+    /// Custom open zeppelin contract class hash of declared contract
```
THenry14, line 16: I don't think this is the correct description 😅

crates/cast/src/starknet_commands/account/delete.rs
```diff
@@ -0,0 +1,57 @@
+use anyhow::{anyhow, bail, Context, Result};
+use camino::Utf8PathBuf;
+use cast::helpers::response_structs::AccountDeleteResponse;
+use cast::parse_number;
+use clap::Args;
+use serde_json::Map;
+use serde_json::Value;
+
+#[derive(Args, Debug)]
+#[command(about = "Delete an account with all important secrets")]
```
THenry14, line 10:
```suggestion
#[command(about = "Delete an account information from the accounts file")]
```

crates/cast/src/starknet_commands/account/delete.rs
```diff
@@ -0,0 +1,57 @@
...
+    if items[&network_name][&name].is_null() {
+        bail!("Account with name {name} does not exist")
+    }
+
+    let address = parse_number(
+        items[&network_name][&name]["address"]
+            .as_str()
+            .ok_or_else(|| anyhow!("Invalid address"))?,
+    )?;
+
```
THenry14, line 39: I don't think we should care about this, at the end of the day we just want to delete this so we don't really care what was the address (and if it was valid).

crates/cast/src/starknet_commands/account/delete.rs
```diff
@@ -0,0 +1,57 @@
...
+        .as_object_mut()
+        .expect("should be an object");
+
+    // now remove the child from there
+    nested.remove(name);
+
+    std::fs::write(path.clone(), serde_json::to_string_pretty(&items).unwrap())?;
+
+    println!("Account successfully removed");
+    Ok(AccountDeleteResponse { address })
```
THenry14, line 56: Let's maybe just return an idication that the account was removed (un)successfully instead? AccoundDeleteResponse could just hold a bool value indicating the status.

crates/cast/src/starknet_commands/account/delete.rs
```diff
@@ -0,0 +1,57 @@
...
+    #[clap(short, long)]
+    pub name: Option<String>,
+
+    /// Custom open zeppelin contract class hash of declared contract
+    #[clap(long)]
+    pub network: Option<String>,
+}
+
+#[allow(clippy::too_many_arguments)]
+pub fn delete(name: &str, path: &Utf8PathBuf, network_name: &str) -> Result<AccountDeleteResponse> {
```
THenry14, line 22: let's maybe also add an option to remove the account from a Scarb.toml profile (if it exists and is indeed included). You can consult `account create` command to see how it is added using `add_profile` toggle, and do something similar (but reversed) here :)

crates/cast/src/starknet_commands/account/delete.rs
```diff
@@ -0,0 +1,57 @@
...
+
+    // get to the nested object "nested"
+    let nested = items
+        .get_mut(network_name)
+        .expect("should exist")
+        .as_object_mut()
+        .expect("should be an object");
+
+    // now remove the child from there
+    nested.remove(name);
```
THenry14, line 51: Let's ask for user confirmation upon removal, to make sure some accounts are not deleted by mistake - this should be doable with either a prompt ("do you wan to remove an account <name> that's on network <network>? y/n") OR a flag to `account delete` - maybe something like `--yes`, that would just skip the prompt and remove the account straight away. What do you think?

crates/cast/src/starknet_commands/account/delete.rs
```diff
@@ -0,0 +1,57 @@
...
+
+    let mut items: Map<String, Value> =
+        serde_json::from_str(&contents).expect("failed to read file");
+
+    // get to the nested object "nested"
+    let nested = items
+        .get_mut(network_name)
+        .expect("should exist")
+        .as_object_mut()
+        .expect("should be an object");
```
THenry14, line 48: Let's improve on error messages in `expect`s to include more information on what really went wrong

crates/cast/src/starknet_commands/account/delete.rs
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    let address = parse_number(
+        items[&network_name][&name]["address"]
+            .as_str()
+            .ok_or_else(|| anyhow!("Invalid address"))?,
+    )?;
+
+    let mut items: Map<String, Value> =
+        serde_json::from_str(&contents).expect("failed to read file");
```
THenry14, line 41: Let's state what's the file we failed to read :+1:

crates/cast/tests/e2e/account/delete.rs
```diff
@@ -0,0 +1,70 @@
+use crate::helpers::constants::DEVNET_OZ_CLASS_HASH;
+use crate::helpers::fixtures::default_cli_args;
+use crate::helpers::runner::runner;
+use indoc::indoc;
+
+#[tokio::test]
+pub async fn test_account_network_should_exist() {
```
THenry14, line 7:
```suggestion
pub async fn test_no_accounts_in_network() {
```

crates/cast/tests/e2e/account/delete.rs
```diff
@@ -0,0 +1,70 @@
...
+    let snapbox = runner(&args);
+
+    snapbox.assert().stderr_matches(indoc! {r#"
+    command: account delete
+    error: No accounts defined for network goerli0-network
+    "#});
+}
+
+#[tokio::test]
+pub async fn test_account_delete_should_exist() {
```
THenry14, line 27:
```suggestion
pub async fn test_account_does_not_exist() {
```

crates/cast/tests/e2e/account/delete.rs
```diff
@@ -0,0 +1,70 @@
...
+    args.append(&mut vec![
+        "account",
+        "create",
+        "--name",
+        "user99",
+        "--salt",
+        "0x1",
+        "--class-hash",
+        DEVNET_OZ_CLASS_HASH,
+    ]);
```
THenry14, line 52: let's not use sncast account create for this - we do not need to create users on devnet chain to test this functionality. Let's add a helper that just writes some random data to the file instead, and use it here. Preferably, this would be done on some new, temporary accounts file (and not tests/data/accounts/accounts.json) to avoid messing up with the rest of the tests

docs/src/starknet/account.md
```diff
@@ -6,6 +6,8 @@ entire account management flow with the `sncast account create` and `sncast acco
 Difference between those two commands is that the first one creates account information (private key, address and more)
 and the second one deploys it to the network. After deployment, account can be used to interact with Starknet.
 
+Also, you can use  `sncast account delete` command to delete account information from accounts file.
+
```
THenry14, line 10:
```suggestion
To remove an account from the accounts file, you can use  `sncast account delete`. Please note this only removes the account information stored locally - this will not remove the account from Starknet.

```

THenry14 (changes requested):
Looking very good, thanks for an amazing contribution 💪 A few smaller things to change/add but apart from that good job!

One think I'd like you to also add is `delete` subcommand appendix in docs - with the description of its arguments etc. Something very similar to what's already done (you can find appendix pages for account command in `docs/src/appendix/cast/account - you can add `delete.md` there :) )

crates/cast/src/starknet_commands/account/delete.rs
devnet0x: i prefer to ask in case of mistakes

THenry14:
Hi @devnet0x , was this closed on purpose?

devnet0x:
Hi, no, i think it was my mistake

devnet0x:
gm..al conversations solved, please recheck. Thanks.

crates/cast/tests/e2e/account/delete.rs
```diff
@@ -0,0 +1,243 @@
...
+        "--url",
+        "http://127.0.0.1:5050/",
+        "--accounts-file",
+        "temp_accounts2.json",
+        "account",
+        "delete",
+        "--name",
+        "user1",
+        "--network",
+        "alpha-goerli",
```
THenry14, line 230: it would be easier for users to provide network with their user-friendly names - eg `testnet` or `mainnet`.

crates/cast/src/starknet_commands/account/delete.rs
```diff
@@ -0,0 +1,118 @@
...
+use std::io::Read;
+use std::io::Write;
+use std::str::Split;
+
+#[derive(Args, Debug)]
+#[command(about = "Delete an account information from the accounts file")]
+pub struct Delete {
+    /// Name of the account to be deployed
+    #[clap(short, long)]
+    pub name: Option<String>,
```
THenry14, line 19: should this be optional? This might lead to confusing situation, when you delete an account (specified by Scarb.toml) without removing it from Scarm.toml itself. I'd suggest we require users provided the name.

Especially so that I see you've marked it as required in the docs :+1:

crates/cast/src/starknet_commands/account/delete.rs
```diff
@@ -0,0 +1,118 @@
...
+use std::fs::{File, OpenOptions};
+use std::io;
+use std::io::Read;
+use std::io::Write;
+use std::str::Split;
+
+#[derive(Args, Debug)]
+#[command(about = "Delete an account information from the accounts file")]
+pub struct Delete {
+    /// Name of the account to be deployed
```
THenry14, line 17: 👀 I think you meant `deleted`

crates/cast/tests/e2e/account/delete.rs
```diff
@@ -0,0 +1,243 @@
...
+    // Run test with an affirmative user input
+    let snapbox = Command::new(cargo_bin!("sncast")).args(args).stdin("Y");
+    let bdg = snapbox.assert();
+    let out = bdg.get_output();
+    let stdout_str =
+        std::str::from_utf8(&out.stdout).expect("failed to convert command output to string");
+    let _ = std::fs::remove_file("temp_accounts2.json");
+    let _ = std::fs::remove_file("Scarb.toml");
+    assert!(stdout_str.contains("Account successfully removed"));
+}
```
THenry14, line 243: let's also add a test without specifying `--network`

docs/src/appendix/cast/account/delete.md
```diff
@@ -0,0 +1,21 @@
+# `delete`
+Delete an account from `accounts-file` and optionally his associated Scarb profile.
```
THenry14, line 2:
```suggestion
Delete an account from `accounts-file` and optionally its associated Scarb profile.
```

docs/src/appendix/cast/account/delete.md
```diff
@@ -0,0 +1,21 @@
...
+
+## `--name, -n <ACCOUNT_NAME>`
+Required.
+
+Account name which is going to be deleted.
+
+## `--delete-profile`
+Optional.
+
+If passed, his associated profile will be removed from Scarb.toml.
```
THenry14, line 16:
```suggestion
If passed, the account's associated profile will be removed from Scarb.toml.
```

docs/src/appendix/cast/account/delete.md
```diff
@@ -0,0 +1,21 @@
...
+
+## `--network`
+Optional.
+
+Network in `accounts-file` associated with the account.
```
THenry14, line 21:
```suggestion
Network in `accounts-file` associated with the account. By default the network of rpc node.
```

docs/src/starknet/account.md
```diff
@@ -73,6 +75,25 @@ command: account create
 max_fee: 0x64a7168300
 address: 0x7a949e83b243068d0cbedd8d5b8b32fafea66c54de23c40e68b126b5c845b61
 ```
+### `account delete`
+
+Delete an account from `accounts-file` and optionally his associated Scarb profile.
```
THenry14, line 80:
```suggestion
Delete an account from `accounts-file` and optionally its associated Scarb profile.
```

crates/cast/tests/e2e/account/delete.rs
devnet0x: yes but accounts file use this notation for networks so i think it would be better to keep this names. Also i can reuse chain_id_to_network_name function which already exists and don't have to add another mapping for testnet/mainnet. Please tell me if you still prefers to change to testnet/mainnet names or we are ok with alpha-goerli, etc?

crates/cast/src/starknet_commands/account/delete.rs
devnet0x: i think we must leave as default behavior to delete the profile as you suggest. and optionally a --no_delete_profile flag. are you ok?

crates/cast/src/starknet_commands/account/delete.rs
```diff
@@ -0,0 +1,118 @@
...
+                    get_scarb_manifest().context("Failed to obtain manifest path from scarb")?
+                }
+            };
+            println!("Removing profile {profile_name} from {manifest_path}.");
+
+            // Open Scarb file.
+            let mut file = File::open(manifest_path.clone())?;
+            let mut contents = String::new();
+            file.read_to_string(&mut contents)?;
+            let eol = "\n";
```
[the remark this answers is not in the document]
devnet0x: i tried to use toml in many ways to remove the profile but i can't find any solution. Remove lines manually was the simplest way i found to do this. I don't know if you can help me with this or approve this solution and in the future open a new issue?

crates/cast/src/starknet_commands/account/delete.rs
drknzz, line 17:
```suggestion
    /// Name of the account to be deleted
```

docs/src/appendix/cast/account/delete.md
drknzz: I see a lot of @THenry14's suggestions are resolved but not applied, although they seem accurate. Is there any reason for it? 😅

crates/cast/src/helpers/response_structs.rs
```diff
@@ -36,6 +36,11 @@ pub struct AccountAddResponse {
     pub add_profile: String,
 }
 
+#[derive(Serialize)]
+pub struct AccountDeleteResponse {
+    pub result: String,
```
drknzz, line 41: Should there also be a notification tied to `--delete-profile`? i.e when it's not passed, maybe there should be a notification that it has been left in the Scarb.toml - and when passed - that it has been deleted from there?

crates/cast/src/starknet_commands/account/delete.rs
drknzz: Maybe let's make it an explicit `--delete-profile false` flag and set it to true as default (if not passed).

crates/cast/src/starknet_commands/account/delete.rs
```diff
@@ -0,0 +1,118 @@
...
+            let mut file = OpenOptions::new()
+                .write(true)
+                .truncate(true)
+                .open(manifest_path)?;
+            file.write_all(new_contents.as_bytes())?;
+            file.flush()?;
+        }
+    } else {
+        result = "Delete cancelled".to_string();
+    };
```
drknzz, line 113: Let's maybe revert the condition, so this short block is executed first, and inside the if we could **return** the AccountDeleteResponse. Then we could have one less indentation for the actual logic deletion logic (above)

So
```
if not delete profile:
  return ...
// rest of the logic
```

crates/cast/src/starknet_commands/account/delete.rs
```diff
@@ -0,0 +1,118 @@
...
+    // Let's ask confirmation
+    println!("Do you want to remove account {name} from network {network_name}? (Y/n)");
+
+    let mut input = String::new();
+    io::stdin()
+        .read_line(&mut input)
+        .expect("Failed to read input");
+
+    let mut result = "Account successfully removed".to_string();
+    if input.starts_with('Y') {
```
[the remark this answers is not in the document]
drknzz: I think this is a valid comment - let's try to do something like

```
if not input.starts_with('Y'):
   return
   
// rest of the logic
```

crates/cast/src/starknet_commands/account/delete.rs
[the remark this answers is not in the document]
drknzz: [See this function](https://github.com/foundry-rs/starknet-foundry/blob/6db50de92474535f18267682ef83ab4a377f637f/crates/cast/src/starknet_commands/account/mod.rs#L94) - the logic here probably should be similar
especially the `get_package_tool_sncast` [could be helpful](https://github.com/foundry-rs/starknet-foundry/blob/6db50de92474535f18267682ef83ab4a377f637f/crates/cast/src/starknet_commands/account/mod.rs#L104)

crates/cast/tests/e2e/account/delete.rs
drknzz: I think the alpha-goerli-like names are fine

crates/cast/src/starknet_commands/account/delete.rs
devnet0x: yes, Thenry14 already made this suggestion so it will be resolved in the new version. thanks.

crates/cast/src/starknet_commands/account/delete.rs
devnet0x: ok, then this is the behavior:
--delete-profile false : to not delete profile in scarb
--delete-profile true : to delete profile in scarb
ommited --delete-profile:  to delete profile in scarb

crates/cast/src/starknet_commands/account/delete.rs
```diff
@@ -0,0 +1,121 @@
...
+use std::fs::File;
+use std::io::Read;
+use std::io::Write;
+
+#[derive(Args, Debug)]
+#[command(about = "Delete account information from the accounts file")]
+pub struct Delete {
+    /// Name of the account to be deleted
+    #[clap(short, long)]
+    pub name: Option<String>,
```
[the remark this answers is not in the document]
devnet0x: maybe, but I left it this way to be consistent with the create command where is declared as option<tring>

devnet0x:
Thanks for your advice, conversations solved and changes commited.

drknzz (commented):
Great 💯
Let's add the sections to the docs like in [this pr](https://github.com/foundry-rs/starknet-foundry/pull/922) before merging

devnet0x:
Done

drknzz (approved):
lgtm 👍

crates/cast/src/starknet_commands/account/delete.rs
```diff
@@ -0,0 +1,114 @@
...
+    if items[&network_name][&name].is_null() {
+        bail!("Account with name {name} does not exist")
+    }
+
+    let mut items: Map<String, serde_json::Value> =
+        serde_json::from_str(&contents).expect("failed to read file { path }");
+
+    // Let's ask confirmation
+    let prompt_text =
+        format!("Do you want to remove account {name} from network {network_name}? (Y/n)");
```
THenry14, line 52: This implies the account will be removed from network, and we only remove it from local file. Let's change to something like:
```suggestion
        format!("Do you want to remove the account {name} deployed to network {network_name} from local file {path}? (Y/n)");
```

wdyt?

devnet0x: 👍

THenry14 (approved):
great job, thanks again for the contribution 🔥

PR #752, opened 2023-09-28
`#fuzzer` attribute for test cases

### LINKED ISSUE #661: Allow configuring random fuzzer per test
Right now it is only possible to set a global random fuzzer config with Scarb.toml or cli arguments. 

It should be possible to configure the fuzzer in tests with a custom attribute like `#[fuzzer ...]`

bartekryba:
Closes #661 

## Introduced changes

- Added `#fuzzer` attribute, that makes it possible to use custom fuzzer configuration for a single test case.

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/forge/tests/data/fuzzing/tests/incorrect_args.cairo
```diff
@@ -6,6 +6,13 @@ fn correct_args(b: felt252) {
     assert(result == 2 + b, '2 + b == 2 + b');
 }
 
+#[test]
+#[fuzzer(test: 10)]
+fn incorrect_attribute_ignored(b: felt252) {
```
cptartur, line 11: Shouldn't we throw an error in case like this?

crates/forge/src/pretty_printing.rs
```diff
@@ -60,19 +64,24 @@ pub(crate) fn print_test_result(test_result: &TestCaseSummary, fuzzer_runs: Opti
...
+                format!(" (fuzzer runs = {runs}{fuzzer_seed_string}, arguments = {arguments:?})")
             } else {
-                format!(" (fuzzer runs = {runs})")
+                format!(" (fuzzer runs = {runs}{fuzzer_seed_string})")
             }
         }
     };
 
-    println!("{result_header} {result_name}{fuzzer_runs}{result_message}");
+    println!("{result_header} {result_name}{fuzzer_report}{result_message}");
```
cptartur, line 84: I'm not sure if it makes sense to print the seed in case the user configured it manually in the test. I understand that the logic was to emphasise that it is different than the seed output at the end of test run, but not sure if it's necessary.

wdyt @MaksymilianDemitraszek @piotmag769

docs/src/testing/fuzz-testing.md
```diff
@@ -59,3 +59,13 @@ fuzzer_runs = 1234
...
+
+It is also possible to specify custom fuzzer configuration for a single test case, using a `#fuzzer` attribute.
+
+```rust
+#[test]
+#[fuzzer(runs: 22, seed: 38)]
+fn test_sum(x: felt252, y: felt252) {
+    assert(sum(x, y) == x + y, 'sum incorrect');
+}
+```
```
cptartur, line 71: I think it should be shown before cmd and scarb.toml arguments. This is more flexible way that will probably be preferred by many developers. At least for the runs.

crates/forge/src/pretty_printing.rs
piotmag769: I think it's fine to print it

CHANGELOG.md
```diff
@@ -9,11 +9,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Forge
 
-#### Changed
+#### Added
 
+- `#fuzzer` attribute to specify fuzzer configuration for a single test case. 
 - Spying events interface is updated to enable the use of events defined inside contracts in assertions
```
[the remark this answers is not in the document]
piotmag769: Disagree, it is a new feature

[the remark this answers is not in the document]
cptartur: I mean spying events got moved to `added` from `changed`.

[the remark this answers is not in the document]
piotmag769: Oh, sorry xD

crates/forge/tests/data/fuzzing/tests/incorrect_args.cairo
bartekryba: Yeah, i'll update it as soon as #760 is merged

CHANGELOG.md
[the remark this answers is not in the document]
bartekryba: Ah it must've slipped in when i resolved merge conflict

crates/forge/src/pretty_printing.rs
```diff
@@ -41,7 +41,11 @@ pub(crate) fn print_test_seed(seed: u64) {
     println!("{}: {seed}", style("Fuzzer seed").bold());
 }
 
-pub(crate) fn print_test_result(test_result: &TestCaseSummary, fuzzer_runs: Option<u32>) {
+pub(crate) fn print_test_result(
+    test_result: &TestCaseSummary,
+    fuzzer_runs: Option<u32>,
+    fuzzer_seed: Option<u64>,
```
piotmag769, line 47: I understand the logic that we only print manually configured seed here, but I think we should always print the seed after a fuzzed test, no matter if configured globally or in code. We can then get rid of printing the seed at the end of the execution (and of this weird `was_fuzzed` variable I introduced 😅)

Therefore we should pass a struct here I think - either we fuzzed or not. There should be no case when one of those args is None and the other is not

piotmag769 (changes requested):

crates/forge/src/pretty_printing.rs
piotmag769: I think we can just return `seed` from `run_with_fuzzing`, but that's up to u how to solve it

cptartur (approved):

PR #760, opened 2023-09-28
Print test diagnostics only once

piotmag769:
Diagnostics from `src/` that pass `scarb build` (like test attribute errors) will still be printed twice since we compile everything in parallel so the dependencies are compiled exactly two times - prints come from different threads. We can use channels to kill one thread earlier if compilation in other one fails, but it is out of scope of this PR - https://github.com/foundry-rs/starknet-foundry/issues/761

## Introduced changes

- `test-collector` refactoring, using only full crates

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


bartekryba (approved):
Looks good to me, didn't find anything that would catch my eye.

bartekryba (approved):
Looks good, i like the test-collector changes.

PR #762, opened 2023-09-29
Store/Load design doc

### LINKED ISSUE #512: Write a design document on `store` and `load` cheatcodes
Create a design document proposing the interface of the cheatcodes `store` and `load` for purpose of modifying the contract storage.

Arcticae:
Closes #512


design_documents/store_load_cheatcodes.md
```diff
@@ -0,0 +1,118 @@
...
+}
+
+#[test]
+fn test_felt252() {
+    let contract_address = deploy_hello_contract();
+    let state = HelloStarknet::contract_state_for_testing();
+    let variable_address = state.balance.address();
+    
+    let new_balance = 420;
+    store::<felt252>(contract_address, variable_address, new_balance);
```
war-in, line 79: Would it be possible to replace `variable_address` with `variable_name`? Maybe we could do some rust magic and handle this outside the cairo. Wdyt?

Arcticae: I think this way it's more intuitive to use and type-safe, which seems to be important

MaksymilianDemitraszek: I think we extract variable address per name from the state so imo it is better as it is.

Radinyn (approved):

design_documents/store_load_cheatcodes.md
```diff
@@ -0,0 +1,118 @@
...
+```
+### Test
+
+```cairo
+use array::ArrayTrait;
+use result::ResultTrait;
+use option::OptionTrait;
+use traits::TryInto;
+use starknet::ContractAddress;
+use starknet::Felt252TryIntoContractAddress;
```
piotmag769, line 60: Those imports are not needed anymore

piotmag769 (approved):
Really like the interface, I think a little bit of magic is worth for achieving type safety

design_documents/store_load_cheatcodes.md
```diff
@@ -0,0 +1,118 @@
...
+enable the users to get the address of the variable, via the `<var_name>.address(<params>)` function, and 
+implement a `store/load` functions, which use the: 
+- Contract address
+- Calculated variable address
+- Value (in case of `store`)
+
+## Caveats & pitfalls of the approach
+
+1. Constructing `contract_state_for_testing` to only calculate the address of the variable
+2. Implementing & importing `storage_access` for the objects we want to write (needs to be implemented anyway, just a bit inconvenient)
```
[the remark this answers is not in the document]
Arcticae: Yes but that's not the typical use-case

PR #763, opened 2023-09-29
fix: display max_fee as int

### LINKED ISSUE #410: Display max_fee as decimal

Cast

### Feature Request

```bash
$ sncast account create --name my_account

Account successfully created. Prefund generated address with at least 4323008256930 tokens. It is good to send more in the case of higher demand, max_fee * 2 = 8646016513860
command: account create
add_profile: --add-profile flag was not set. No profile added to Scarb.toml
address: 0x3e5b320468ce80d69fe69e2011f80065cf41298ac18a76a124f14849c3d29e9
max_fee: 0x3ee875f1ba2
```

max_fee should be displayed in decimal form, the same way it is displayed in the message above

```bash
address: 0x3e5b320468ce80d69fe69e2011f80065cf41298ac18a76a124f14849c3d29e9
max_fee: 4323008256930
```

Possibly a flag similar to `--int-format`, should be added to display every output in hex format.

lambda-0x:
Closes #410 

## Introduced changes

- display `max_fee` as int unless `--hex-format` is specified for it

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


lambda-0x:
@THenry14 i see there is a `--int-format` flag in `sncast`, wouldn't it be confusing to use `--hex-format` along with that?

THenry14:
> @THenry14 i see there is a `--int-format` flag in `sncast`, wouldn't it be confusing to use `--hex-format` along with that?

They should be mutually exclusive - if someone passes `--int-format` all outputs should be in int format; when `--hex-format` all outputs should be in hex format; when someone passes both flags we should error out (I believe clap has such feature) and if no flag is passed we output addresses/hashes as hex and fees as ints. Does that make sense?

lambda-0x:
when `--int-format` is not specified we are already using hex format. 

i think the problem is that even when using hex format (i.e. `--int-format` is not specified) we want to print `max_fee` as int, and give an option for hex format (just for `max_fee`) inside `sncast account create` subcommand

lambda-0x:
note: `--int-format` that i mention is argument available on `sncast` and `--hex-format` i have added in this PR is on `sncast account create` subcommand

THenry14:
> when --int-format is not specified we are already using hex format.

yes, and we want to have this default behavioour changed to display addresses/hashes as hex, but fees as int - not hex

>i think the problem is that even when using hex format (i.e. --int-format is not specified) we want to print max_fee as int, and give an option for hex format (just for max_fee) inside sncast account create subcommand

>note: --int-format that i mention is argument available on sncast and --hex-format i have added in this PR is on sncast account create subcommand

I suppose this flag should be global and not to an account subcommand exclusively - plus it should change every output value to hex, as I described in previous comment. This way, by default, we have all the addresses/hashes in hex and fees as int (as it probably should be), but if someone wishes so one could change the outputs to ints/hexes exclusively (be it for scripting, to avoid manual conversion for contracts input etc). I believe that should be a behaviour that would suit more usage scenarios, and it could even potentially be simpler to implement. 

so, a tldr:
- `--int-format` - show everything as int
- `--hex-format` - show everything as hex
- no flags - show addresses and hashes as hex, fees as int
- both flags - error

Does that clarify things?

lambda-0x:
okay got it, thanks for the detailed summary!

This will be bigger change then i initially thought it to be as it would affect every part of sncast. So to test out things i would need to setup account contracts and stuff so i can run each and every subcommand.

for e.g. i was trying to run `sncast account create` subcommand by specifying my own keystore and account (that was created using starkli) but it was giving some errors. I also tried using katana as local devnet which was giving some other error.  (i didn't look to deeply into it)

is there any guide on setting up accounts and local devnet to test out this things?

THenry14:
> okay got it, thanks for the detailed summary!
> 
> This will be bigger change then i initially thought it to be as it would affect every part of sncast. So to test out things i would need to setup account contracts and stuff so i can run each and every subcommand.
> 
> for e.g. i was trying to run `sncast account create` subcommand by specifying my own keystore and account (that was created using starkli) but it was giving some errors. I also tried using katana as local devnet which was giving some other error. (i didn't look to deeply into it)
> 
> is there any guide on setting up accounts and local devnet to test out this things?

I think just running the script `prepare_for_tests.sh` as detailed in contribution guidelines https://github.com/foundry-rs/starknet-foundry/blob/master/CONTRIBUTING.md should be enough to get you going. Make sure you have installed asdf (for scarb versioning) and starknet-devnet first, and you can then just run tests (eg cargo test  -p cast, you can also filter out tests as per cargo docs). Let me know if you run at any issues :)

lambda-0x:
i tried installing `starknet-devnet` via pip but it failed with some error. will the new `starknet-devnet-rs` work instead?

THenry14:
> i tried installing `starknet-devnet` via pip but it failed with some error. will the new `starknet-devnet-rs` work instead?

Make sure you have libgmp3-dev (linux) or gmp (mac) installed, and then please try again. the new starknet-devnet-rs still has some missing features and issues - it could work for you for some simple tests, but most of our tests from the test suite will fail

lambda-0x:
i do have it installed, i think the problem is that i have newer python version then what `starknet-devnet` supports. i will try this on a different system in sometime.

THenry14:
> i do have it installed, i think the problem is that i have newer python version then what `starknet-devnet` supports. i will try this on a different system in sometime.

I see. you can also try installing `pyenv` for python version management.

lambda-0x:
pyenv with virtualenv did the trick!

crates/cast/src/lib.rs
```diff
@@ -307,22 +318,40 @@ pub async fn handle_wait_for_tx<T>(
...
+                if let Ok(int_value) = U256::from_dec_str(&value) {
+                    (key, format!("{int_value:#x}"))
+                } else if let Ok(field) = FieldElement::from_str(value.as_str()) {
+                    (key, format!("{field:#x}"))
+                } else {
+                    (key, value)
+                }
+            }
+        })
+        .collect();
```
lambda-0x, line 354: does this logic make sense @THenry14? `Default` is when neither `int-format` or `hex-format` is specified.

lambda-0x: and i assume that when neither is specified we receive values in appropriate format from source

crates/cast/src/main.rs
```diff
@@ -42,10 +44,14 @@ struct Cli {
     #[clap(short, long)]
     keystore: Option<Utf8PathBuf>,
 
-    /// If passed, values will be displayed as integers, otherwise as hexes
-    #[clap(short, long)]
+    /// If passed, values will be displayed as integers
+    #[clap(long, conflicts_with = "hex_format")]
     int_format: bool,
 
+    /// If passed, values will be displayed as integers
```
drknzz, line 51:
```suggestion
    /// If passed, values will be displayed as hex
```

crates/cast/src/starknet_commands/account/create.rs
```diff
@@ -94,7 +97,13 @@ pub async fn create(
     let mut output = vec![("address", format!("{address:#x}"))];
     if account_json["deployed"] == json!(false) {
         println!("Account successfully created. Prefund generated address with at least {max_fee} tokens. It is good to send more in the case of higher demand, max_fee * 2 = {}", max_fee * 2);
-        output.push(("max_fee", format!("{max_fee:#x}")));
+        // For default and Int it should be in int
+        let max_fee_str = if value_format == ValueFormat::Hex {
+            format!("{max_fee:#x}")
+        } else {
+            format!("{max_fee:#}")
+        };
```
drknzz, line 105: Could we maybe have some trait for ValueFormat, so it would be possible to call `value_format.format(max_fee)`?

crates/cast/src/lib.rs
drknzz: The logic seems okay. If we had a trait implementing .format(), this maybe could be simplified though

drknzz (commented):
Looking good 👍 
Could there be some test added?

crates/cast/src/starknet_commands/account/create.rs
lambda-0x: don't think a trait is required, directly added a method on the type.

crates/cast/tests/e2e/account/create.rs
```diff
@@ -389,3 +389,87 @@ pub fn test_keystore_already_exists(keystore_path: &str, account_path: &str, err
...
+
+    let snapbox = runner(&args);
+
+    snapbox.assert().stdout_matches(indoc! {r#"
+        CREATE_KEYSTORE_PASSWORD environment variable found and will be used for keystore password
+        Account successfully created[..]
+        command: account create
+        add_profile: --add-profile flag was not set. No profile added to Scarb.toml
+        address: [..]
+        max_fee: [..]
```
lambda-0x, line 423: i couldn't find a way to enforce this `[..]` (this wildcard is part of snapbox) to be just numbers. currently this would pass even if those are in format `0x[..]`.

is there anyway to enforce this?

CHANGELOG.md
```diff
@@ -22,6 +22,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 #### Added
 
 - `show-config` subcommand to display currently used configuration
+- `--hex-format` flag has been added
+
+#### Removed
+- `-i` for `--int-format` is removed
```
THenry14, line 28: Let's make this message more explicit - for a moment I thought the `--int-format` flag was removed altogether, and not only the abbreviation :D

crates/cast/tests/e2e/account/create.rs
THenry14: not sure tbh, we can leave it like that for now I guess

THenry14 (commented):
looking good 🔥

CHANGELOG.md
lambda-0x: haha, fixed it!

lambda-0x:
(rebased changes to master)

THenry14:
@lambda-0x please fix the issues found by clippy 🙏 after that, I think this PR is ready :+1:

THenry14 (approved):
great job, thanks for the contribution!

PR #764, opened 2023-10-01
feat: implement show-config subcommand

### LINKED ISSUE #278: Get current configuration

0xicosahedron:
Closes #278

## Introduced changes

- current its printing fields that are not present (empty strings) as `0x0` due to the way `print_command_result` works, should i create a new duplicate function to not print this if they are not set?
- any other suggestions?

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/cast/src/main.rs
```diff
@@ -273,6 +280,19 @@ async fn main() -> Result<()> {
                 Ok(())
             }
         },
+        Commands::ShowConfig(_) => {
+            let chain_id = get_chain_id(&provider).await?;
+            let chain_id = chain_id_to_network_name(chain_id);
```
THenry14, line 285: let's maybe rename one of the variables?

crates/cast/src/starknet_commands/show_config.rs
```diff
@@ -0,0 +1,26 @@
...
+    scarb_path: Utf8PathBuf,
+    network: String,
+) -> Result<ShowConfigResponse> {
+    Ok(ShowConfigResponse {
+        profile,
+        scarb_path,
+        network,
+        rpc_url: cast_config.rpc_url,
+        account: cast_config.account,
+        account_file_path: cast_config.accounts_file,
```
THenry14, line 24: what if users use keystore instead of accounts file?

crates/cast/src/starknet_commands/show_config.rs
```diff
@@ -0,0 +1,26 @@
...
+    })
+}
```
THenry14, line 26: Let's also add some tests :) After they're added I will access the pipeline to go through

THenry14 (dismissed):
great job, thanks for the contribution 💪 a few smaller things to change

crates/cast/src/starknet_commands/show_config.rs
0xicosahedron: before adding test, i wanted to know your thoughts on:
> current its printing fields that are not present (empty strings) as 0x0 due to the way print_command_result works, should i create a new duplicate function to not print this if they are not set?

THenry14: Ideally, the current function should be modified to not print these fields. How much work do you think this would be?

crates/cast/src/starknet_commands/show_config.rs
0xicosahedron: added keystore

crates/cast/src/starknet_commands/show_config.rs
0xicosahedron: that function calls another function (`print_formatter`) that does this conversion, we can pass in a boolean flag depending on which it can choose to convert values to numeric. does that sound good?

https://github.com/foundry-rs/starknet-foundry/blob/302627ec8ef83bd8d80626d143a34f794d3a95dc/crates/cast/src/lib.rs#L308-L325

war-in: How about passing `Options` as `ShowConfigResponse` fields? We could modify the print_formatted function to not print `None`'s. wdyt?

0xicosahedron: thats what i tried initially, but `print_command_result` was panicking somwhere here: https://github.com/foundry-rs/starknet-foundry/blob/302627ec8ef83bd8d80626d143a34f794d3a95dc/crates/cast/src/lib.rs#L361

when value is `None` i think `as_str` returns `None` https://docs.rs/serde_json/1.0.107/serde_json/value/enum.Value.html#method.as_str

war-in: Any updates on `print_command_result` are welcome, so feel free to do it. 

> Is it a good idea to ignore value's that are not convertable to string?

Are there any?

0xicosahedron: > Are there any?

Not atm but in future we might want to print more complicated nested values.

considering that i think for now adding a boolean to not convert values to numeric make more sense.

THenry14: > How about passing Options as ShowConfigResponse fields? We could modify the print_formatted function to not print None's. wdyt?

that would be a way to go imo

> actually then we would print empty values for those, so would need to remove it somehow so would have to make it optional and ignore None values

I think we should just not print fields that are None. Not sure their omission would make debugging harder - in contrary, you know they were just empty (not set at all) so it is a hint that maybe something went not as you expected it to. what do you think?

0xicosahedron: yeah i think making it optional and removing None make sense, will implement that.

0xicosahedron: fixed this, moving conversation to new thread

0xicosahedron:
now the output looks like:
```
command: show-config
account: 0x0
account_file_path: /path/to/starknet_open_zeppelin_accounts.json
keystore: 0x0
network: KATANA
rpc_url: http://localhost:5050
```

`account` and `keystore` still show `0x0` because they are set to empty value by `update_cast_config`, how should this be handled? https://github.com/foundry-rs/starknet-foundry/blob/33bd7a7c8247ad360fae1b47e69b1cd744199d10/crates/cast/src/main.rs#L286-L288

THenry14:
> now the output looks like:
> 
> ```
> command: show-config
> account: 0x0
> account_file_path: /path/to/starknet_open_zeppelin_accounts.json
> keystore: 0x0
> network: KATANA
> rpc_url: http://localhost:5050
> ```
> 
> `account` and `keystore` still show `0x0` because they are set to empty value by `update_cast_config`, how should this be handled?
> 
> https://github.com/foundry-rs/starknet-foundry/blob/33bd7a7c8247ad360fae1b47e69b1cd744199d10/crates/cast/src/main.rs#L286-L288

I think the best solution would be to just check if some fields are empty strings or None and omit them altogether. We could just add a simple check to either `print_command_result` or to `show_config`, the latter is probably more preferable. More fields could also be marked as Optional in `ShowConfigResponse`.  Let me know if you need some additional hints/help :)

0xicosahedron:
i got `account`, `keystore` and `rpc_url` from `CastConfig` which are non-optional, is there any reason we add default values there? 

my guess is because previously `print_command_result` wasn't able to handle `None` values. but now since `None` values are skipped we can make this values optional in `CastConfig` itself?

THenry14:
> i got `account`, `keystore` and `rpc_url` from `CastConfig` which are non-optional, is there any reason we add default values there?

not sure, don't hink so.

> my guess is because previously `print_command_result` wasn't able to handle `None` values. but now since `None` values are skipped we can make this values optional in `CastConfig` itself?

this is a good idea :+1:

0xicosahedron:
> my guess is because previously print_command_result wasn't able to handle None values. but now since None values are skipped we can make this values optional in CastConfig itself?

tried doing this but its since its used in alot of other commands which i don't understand i am going with the design that @THenry14 previously suggested.

now what kind of tests and documentation update do you suggest?

THenry14:
> > my guess is because previously print_command_result wasn't able to handle None values. but now since None values are skipped we can make this values optional in CastConfig itself?
> 
> tried doing this but its since its used in alot of other commands which i don't understand i am going with the design that @THenry14 previously suggested.
> 
> now what kind of tests and documentation update do you suggest? any examples would help.

That's fine, thanks for trying out different approaches - I really appreciate it :) 

Regarding tests - some simple ones like:
- checking if the config is as expected when we have as much things as possible defined in Scarb.toml
- checking if the config is as expected when we provide everything via cli (not in Scarb.toml)
- checking if the config is as expected when we mix Scarb.toml with cli
- checking if the config is as expected when we provide/do not provide `--network` parameter

I think that should suffice. Regarding docs, we will need to have a simple overiew page that show example usage (something like [this](https://github.com/foundry-rs/starknet-foundry/blob/master/docs/src/starknet/account.md)) and an appendix with all the possible flags to the subcommand ([eg](https://github.com/foundry-rs/starknet-foundry/blob/master/docs/src/appendix/cast/call.md)).

0xicosahedron:
added a test, but when trying to run:
> cargo test e2e

i keep getting a panic in a test which is unrelated to this changes. getting the same error on master.

THenry14:
> added a test, but when trying to run:
> 
> > cargo test e2e
> 
> i keep getting a panic in a test which is unrelated to this changes. getting the same error on master.

Can you show me the error? tests on master work on our CI, so my bet there's some issues with your local env

0xicosahedron:
merge master still seeing this:
```
thread '<unnamed>' panicked at 'scarb build failed,
stderr: 

stdout: 
error: Version solving failed:
- cheatnet_testing_contracts v0.1.0 (/path/to/starknet-foundry/crates/cheatnet/tests/contracts/Scarb.toml) cannot use starknet v1.1.1 (std), because cheatnet_testing_contracts requires starknet ^2.2.0

Scarb does not have real version solving algorithm yet.
Perhaps in the future this conflict could be resolved, but currently,
please upgrade your dependencies to use latest versions of their dependencies.
', crates/cheatnet/tests/main.rs:20:9
fatal runtime error: failed to initiate panic, error 5
error: test failed, to rerun pass `-p cheatnet --test main`

Caused by:
  process didn't exit successfully: `/path/to/starknet-foundry/target/debug/deps/main-d1a34fc2f3c08e65 e2e` (signal: 6, SIGABRT: process abort signal)
```

i have already ran `prepare_for_tests.sh` script before running the test.

> $ scarb --version
> scarb 0.4.1 (f26a2a4e7 2023-06-28)
> cairo: 1.1.1 (bdbfe6f26

> $ starknet-devnet --version
> 0.5.5

0xicosahedron:
switching to latest version of scarb using asdf fixes the issue. 

why is `prepare_for_tests.sh` is setting global scarb version to `0.4.0`, it can use `.tool-version` to locally set the scarb version for this repo. it shouldn't set the global version in the script.

THenry14:
> merged with master still seeing this:
> 
> ```
> thread '<unnamed>' panicked at 'scarb build failed,
> stderr: 
> 
> stdout: 
> error: Version solving failed:
> - cheatnet_testing_contracts v0.1.0 (/path/to/starknet-foundry/crates/cheatnet/tests/contracts/Scarb.toml) cannot use starknet v1.1.1 (std), because cheatnet_testing_contracts requires starknet ^2.2.0
> 
> Scarb does not have real version solving algorithm yet.
> Perhaps in the future this conflict could be resolved, but currently,
> please upgrade your dependencies to use latest versions of their dependencies.
> ', crates/cheatnet/tests/main.rs:20:9
> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
> fatal runtime error: failed to initiate panic, error 5
> error: test failed, to rerun pass `-p cheatnet --test main`
> 
> Caused by:
>   process didn't exit successfully: `/path/to/starknet-foundry/target/debug/deps/main-d1a34fc2f3c08e65 e2e` (signal: 6, SIGABRT: process abort signal)
> ```
> 
> i have already ran `prepare_for_tests.sh` script before running the test.
> 
> > $ scarb --version
> > scarb 0.4.1 (f26a2a4e7 2023-06-28)
> > cairo: 1.1.1 (bdbfe6f26
> 
> > $ starknet-devnet --version
> > 0.5.5

This looks like you have global/local for v1 contracts scarb version set to newer scarb, and so v1 contracts needed for tests are being compiled with it, which will always fail.|

> why is prepare_for_tests.sh is setting global scarb version to 0.4.0, it can use .tool-version to locally set the scarb version for this repo. it shouldn't set the global version in the script.

this is something we will change [here](https://github.com/foundry-rs/starknet-foundry/pull/814), it was needed for a few reasons for our test setup. scarb uses local .tool-versions from test directories, it seems that for some reason you may have those files set to wrong versions

0xicosahedron:
added the tests you mentioned above

> checking if the config is as expected when we provide/do not provide --network parameter
i am guessing you meant `--url` here, this is already tested by the first case in which we don't provide url but read it from scarb toml, when neither is specified we print an error.

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 #### Added
 
+- `show-config` subcommand to display currently used configuration
```
THenry14, line 14: This should go under `Cast` subsection, not `Forge` :)

crates/cast/tests/e2e/show_config.rs
```diff
@@ -0,0 +1,77 @@
...
+        "--profile",
+        "profile1",
+        "show-config",
+    ];
+
+    let snapbox = runner(&args);
+
+    snapbox.assert().success().stdout_eq(indoc! {r#"
+        command: show-config
+        account: user1
```
THenry14, line 18: when there is `keystore`, the account is a path to account file (this is something different than accounts file)

show-config should:

- not show accounts_file_path when keystore is non empty, the account then should be a path to account file
- show accouns_file_path when keystore is empty, the account then is an account name from accounts file

let's adjust the outputs to this :+1:

THenry14:
I can see clippy CI check is failing with unrelated changes - can you make sure you are rebased/merged to master?
if yes I will try to figure out what is the deal with it

let's also fox formatting (cargo fmt)

CHANGELOG.md
0xicosahedron: oops didn't see it

crates/cast/tests/e2e/show_config.rs
0xicosahedron: > show accouns_file_path when keystore is empty, the account then is an account name from accounts file

do i need to do validation that `account_name` is present inside `accounts_file`?

THenry14: no, this is already done elsewhere :)

0xicosahedron:
added docs as well, since `show-config` dont take any arguments of its own i am not sure what to write in appendix. for other section i have added the motivation of adding this subcommand.

docs/src/starknet/show_config.md
```diff
@@ -0,0 +1,30 @@
+# Printing Current Configuration
+
+## Overview
+
+Many times before executing any other cast command you want to make sure that you are passing the right
+configuration settings and not mixing up commands that needs to be sent to development environment to mainnet :)
```
THenry14, line 6:
```suggestion
Sometimes, before executing any other cast command, one may want to make sure that the right
configuration settings are being used (eg proper network or account is used).
```

wydt? :)

docs/src/starknet/show_config.md
```diff
@@ -0,0 +1,30 @@
...
+
+For those cases you can call `show-config` with all the same configuration related arguments, you can also just
+replace subcommand and anything after it with `show-config` and it should show all the configuration that will be
+used.
```
THenry14, line 10:
```suggestion
To see just that, a `show-config` subcommand  can be used. You can just
replace any subcommand (and its parameters) with `show-config` and it will show currently used configuration.
```

docs/src/starknet/show_config.md
```diff
@@ -0,0 +1,30 @@
...
+
+Anything configuration that you cannot see there means its not set properly or is not being detected.
```
THenry14, line 12:
```suggestion

```

docs/src/starknet/show_config.md
```diff
@@ -0,0 +1,30 @@
...
+  --rpc_url http://127.0.0.1:5050 \
+  --acount user1
+  show-config
+
+command: show-config
+account: user1
+account_file_path: ../account-file
+chain_id: alpha-goerli
+keystore: ../keystore
+rpc_url: http://127.0.0.1:5050/rpc
```
THenry14, line 29: let's remember to adjust the output after changes with exclusive keystore <-> accounts_file_path items

0xicosahedron: yup also added separate tests for those.

docs/src/starknet/show_config.md
0xicosahedron: looks good!

0xicosahedron:
everything should be fixed now

CHANGELOG.md
```diff
@@ -50,6 +50,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
```
THenry14, line 53: this is added in wrong place, uder previous release. Please move it to the top, under `## [Unreleased]` section. 

A
```
### Cast
#### Added
(...)
```

should be added there, since this is a first added PR scheduled for next release :+1:

crates/cast/tests/e2e/show_config.rs
```diff
@@ -0,0 +1,116 @@
...
+        "--keystore",
+        "../keystore",
+        "show-config",
+    ];
+
+    let snapbox = runner(&args);
+
+    snapbox.assert().success().stdout_eq(indoc! {r#"
+        command: show-config
+        account: user2
```
THenry14, line 43: is this a path, or username?

docs/src/starknet/show_config.md
```diff
@@ -0,0 +1,28 @@
+# Printing Current Configuration
+
+## Overview
+
+Sometimes, before executing any other cast command, one may want to make sure that the right
+configuration settings are being used (eg proper network or account is used).
+
+To see just that, a `show-config` subcommand  can be used. You can just
```
THenry14, line 8:
```suggestion
To see just that, a `show-config` subcommand can be used. You can just
```

docs/src/starknet/show_config.md
THenry14: this still shows both keystore and accounts_file_path, which are exclusive

THenry14:
Regarding failing clippy, there was a rust update - a fix should be on master soon. sorry!

CHANGELOG.md
0xicosahedron: my bad, i just searched for Cast and added it below it...

docs/src/starknet/show_config.md
0xicosahedron: sorry forgot to update docs.

crates/cast/tests/e2e/show_config.rs
0xicosahedron: i misunderstood what you meant, fixed!

CHANGELOG.md
THenry14: no worries ;)

0xicosahedron:
will rereview myself after CI bug is fixed

THenry14:
> will rereview myself after CI bug is fixed

should be any moment now, as the fix is already in merge queue https://github.com/foundry-rs/starknet-foundry/pull/833 :) 
please rebase to/merge with master and I will unblock the CI for this PR once again

0xicosahedron:
should be good to go!

crates/cast/src/helpers/response_structs.rs
```diff
@@ -41,3 +41,14 @@ pub struct MulticallNewResponse {
...
 }
+
+#[derive(Serialize)]
+pub struct ShowConfigResponse {
+    pub profile: Option<String>,
+    pub chain_id: String,
+    pub rpc_url: Option<String>,
+    pub account: Option<String>,
+    pub scarb_path: Option<Utf8PathBuf>,
+    pub account_file_path: Option<Utf8PathBuf>,
```
drknzz, line 52: Let's name this `accounts_file_path`

crates/cast/src/starknet_commands/show_config.rs
```diff
@@ -0,0 +1,42 @@
...
+    provider: &JsonRpcClient<HttpTransport>,
+    cast_config: CastConfig,
+    profile: Option<String>,
+    scarb_path: Option<Utf8PathBuf>,
+) -> Result<ShowConfigResponse> {
+    let chain_id_field = get_chain_id(provider).await?;
+    let chain_id = chain_id_to_network_name(chain_id_field);
+    let rpc_url = Some(cast_config.rpc_url).filter(|p| !p.is_empty());
+    let account = Some(cast_config.account).filter(|p| !p.is_empty());
+    let mut account_file_path =
```
drknzz, line 25: Same here

crates/cast/src/helpers/response_structs.rs
0xicosahedron: done

CHANGELOG.md
```diff
@@ -40,6 +40,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 
+### Cast
+
+#### Added
+
+- `show-config` subcommand to display currently used configuration
+
+### Changed
+
+
```
drknzz, line 51: Let's update the changelog, so the changes are on top, as a new version has come out

drknzz (commented):
Looking good 👍 
just the CHANGELOG comment

CHANGELOG.md
0xicosahedron: done!

drknzz (approved):
Great job 💯

PR #775, opened 2023-10-02
663 parallelise test execution tokio

tkumor3:
Closes #

## Introduced changes

-
-

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/forge/src/lib.rs
```diff
@@ -285,23 +289,25 @@ fn try_close_tmp_dir(maybe_tmp_dir: Option<TempDir>) -> Result<()> {
 /// * `contracts` - Map with names of contract used in tests and corresponding sierra and casm artifacts
 /// * `predeployed_contracts` - Absolute path to predeployed contracts used by starknet state e.g. account contracts
 ///
-#[allow(clippy::implicit_hasher)]
-pub fn run(
-    package_root: &Utf8PathBuf,
+
+#[allow(clippy::implicit_hasher, clippy::too_many_arguments)]
+pub async fn run(
+    package_root: Arc<Utf8PathBuf>,
```
cptartur, line 295: I'm not sure if it makes sense to use `Arc` for all possible types. It is more efficient than copying complex structs like `RunnerParams` etc. but for simpler stuff, I'd just use normal references (`&`) and clone where necessary.

crates/forge/src/lib.rs
```diff
@@ -352,97 +361,144 @@ pub fn run(
...
+
+    for thread in tasks {
+        let (result, runs) = thread.await.unwrap()?;
+        if runs.is_some() {
+            was_fuzzed = true;
         }
+
+        pretty_printing::print_test_result(&result, runs);
+        results.push((result, runs));
     }
```
cptartur, line 474: Instead of doing this we should add `futures` dependency and

```rust
use futures::future::join_all;

join_all(tasks).await()
```

This loop is not an efficient way of waiting for multiple threads.

crates/forge/src/lib.rs
```diff
@@ -352,97 +361,144 @@ pub fn run(
...
+                }
+                Ok((result, None))
+            }
+            Err(e) => Err(e),
+        }
 
-fn run_tests_from_crate(
-    package_root: &Utf8PathBuf,
+        // pretty_printing::print_test_result(&result, None);
+        // result
```
cptartur, line 397: Commented out code

crates/forge/src/lib.rs
```diff
@@ -352,97 +361,144 @@ pub fn run(
...
+            runner_params,
+            runner,
+            case,
+            &args,
+            cancellation_token,
+        )
+        .await
+    }
+}
+//TODO rename
```
cptartur, line 411: TODO

crates/forge/src/lib.rs
```diff
@@ -462,26 +518,63 @@ fn run_with_fuzzing(
...
+            let runner_params = runner_params.clone();
+            let runner_config = runner_config.clone();
+            async move {
+                tokio::select! {
+                    _ = cancellation_token.cancelled() => {
+                        // The token was cancelled
+                        Ok(TestCaseSummary::Failed { name: "fuzzing".to_string(), run_result: None, msg: None, arguments: vec![] })
+                    },
+                    _ = cancellation_fuzzing_token.cancelled() => {
+                        Ok(TestCaseSummary::Failed { name: "fuzzing".to_string(), run_result: None, msg: None, arguments: vec![] })
```
cptartur, line 541: Why do we need separate tokens for these?

crates/forge/src/lib.rs
```diff
@@ -462,26 +518,63 @@ fn run_with_fuzzing(
...
+        };
+
+        if let TestCaseSummary::Failed { .. } = &result {
+            if results.is_empty() {
+                token.cancel();
+                if runner_config.exit_first {
+                    cancellation_token.cancel();
+                };
+                results.push(result);
+            }
```
cptartur, line 575: I'm not sure if I understand this cancellation logic. This could be explained or preferably explained better.

crates/cheatnet/src/forking/state.rs
```diff
@@ -97,10 +97,11 @@ impl StateReader for ForkStateReader {
...
         }
 
-        match self.runtime.block_on(
+        match self.runtime.block_on(async {
             self.client
-                .get_class_hash_at(self.block_id, contract_address.to_field_element()),
-        ) {
+                .get_class_hash_at(self.block_id, contract_address.to_field_element())
+                .await
+        }) {
```
cptartur, line 104: This looks a bit weird to have block on and then async inside. Why was this added?

crates/forge/src/running.rs
```diff
@@ -68,17 +69,39 @@ fn build_hints_dict<'b>(
...
+            &package_root,
+            &runner,
+            &case,
+            &runner_config,
+            &runner_params,
+            args,
+        )
+    })
+    .await?
+}
```
cptartur, line 91: We have a async threads (created in `lib.rs`) spawning blocking threads inside them. This doesn't seem like a correct solution. Why do we need this?

cptartur (changes requested):
I only reviewed the code, not the tests.

crates/forge/src/lib.rs
tkumor3: cancellation_token -  is to stop threads in case of `exit-first` so it stops all tests in package
cancellation_fuzzing_token - it cancels all threads related to a single fuzzing test. I assume, that for fuzzing test if one generated test fails, we can assume that it failed at all

crates/forge/src/lib.rs
```diff
@@ -352,97 +361,144 @@ pub fn run(
...
-                        test_crate_type: tests.test_crate_type,
+        let task = task::spawn({
+            let package_root = package_root.clone();
+            let case = Arc::new(case.clone());
+            let c = case.clone();
+            let runner = runner.clone();
+            let args: Vec<ConcreteTypeId> = args.into_iter().cloned().collect();
+            let cancellation_token = cancellation_token.clone();
+            let runner_params = runner_params.clone();
+            let runner_config = runner_config.clone();
```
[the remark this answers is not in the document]
tkumor3: probably no, 
```cannot move out of `runner`, a captured variable in an `FnMut` closure
`runner` is moved here
```
I get this error

crates/cheatnet/src/forking/state.rs
tkumor3: good catch! I test sth and add it by mistake

crates/forge/src/running.rs
tkumor3: Because cast block thread to call externals

crates/forge/src/lib.rs
tkumor3: I answer in prev comment

crates/forge/src/lib.rs
cptartur: Okay, makes sense

crates/forge/src/lib.rs
cptartur: I wonder if this can be refactored a bit to make it more clear/readable?

crates/forge/src/lib.rs
MaksymilianDemitraszek: We should generally avoid keeping stuff on the heap if not necessary

MaksymilianDemitraszek:
I will be reviewing this one please don't merge beforehand 🙏

crates/forge/src/lib.rs
MaksymilianDemitraszek: I don't think it much matters performance wise, but is better to do it this way (`join_all` or `select_all`) as it is more idiomatic and easier to reason about.

crates/forge/src/lib.rs
```diff
@@ -352,97 +360,150 @@ pub fn run(
 
     Ok(summaries)
 }
+async fn chose_strategy_and_run_test(
```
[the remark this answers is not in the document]
MaksymilianDemitraszek: And it should be slplit into two functions, chossing strategy and running should be separate steps, also shouldn't forking be a strategy as well?

MaksymilianDemitraszek (changes requested):
Partial review, waiting for some remaining changes (discussed on slack) to be applied before doing a proper one

crates/forge/src/lib.rs
piotmag769: @MaksymilianDemitraszek isn't it blocking in ever loop run though? Unless tokio is smart and somehow compiles it to wait for all threads at once under the hood (like Django with for loop and db requests)

MaksymilianDemitraszek: Yea, it would block in single threaded async setting, but we use `task::spawn` which is immediately executed on a separate thread.

crates/forge/src/lib.rs
```diff
@@ -352,97 +360,150 @@ pub fn run(
...
+                }
             }
+        });
+
+        tasks.push(task);
+    });
+
+    let mut results = vec![];
+
+    for thread in tasks {
```
MaksymilianDemitraszek, line 471: This is a task not a thread, please change the name as it is misleading.

crates/forge/tests/e2e/workspaces.rs
```diff
@@ -238,12 +246,17 @@ fn root_workspace_for_entire_workspace() {
         Failure data:
             original value: [8111420071579136082810415440747], converted to a string: [failing check]
         
-        [SKIP] tests::test_failing::test_another_failing
-        Tests: 1 passed, 1 failed, 1 skipped
+        [FAIL] tests::test_failing::test_another_failing
```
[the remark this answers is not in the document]
piotmag769: I guess execution order could change when executing in parallel. We should only make one test with exit first with one short test failing and a few really long tests being skipped

crates/forge/tests/integration/setup_fork.rs
```diff
@@ -53,77 +47,81 @@ fn fork_simple_decorator() {
...
-            test.contracts(&corelib_path()).unwrap(),
-            Utf8PathBuf::from_path_buf(predeployed_contracts().to_path_buf()).unwrap(),
-            Default::default(),
-        ),
-    )
-    .unwrap();
-
-    assert_passed!(result);
-}
+//TODO:: fix test
```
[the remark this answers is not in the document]
tkumor3: yes, but problem is with test execution, it use default `run` (not the test one)currently it's async. So I have to handle it

crates/forge/tests/e2e/workspaces.rs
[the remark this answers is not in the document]
tkumor3: In prev version workspace tests were running with flag exit-first

In the current version we run tests parallel, so runs with flag exit-first are not deterministic. Before finished we can't know how many tests will finished with success. So I resign with running it with this flag.


I added test to check if --exit-first works good.

crates/forge/src/lib.rs
tkumor3: I found `use futures::stream::{FuturesOrdered, FuturesUnordered};` and it seems has better performance. Mostly FuturesUnordered which we are able to use in fuzzing tests

crates/forge/src/lib.rs
tkumor3: I think currently. it's better

crates/forge/src/running.rs
```diff
@@ -68,17 +70,28 @@ fn build_hints_dict<'b>(
     (hints_dict, string_to_hint)
 }
 
-#[allow(clippy::too_many_arguments)]
+pub(crate) async fn blocking_run_from_test(
```
piotmag769, line 73: Misleading name, it is not actually blocking if it is `async`. Moreover I think we can just copy paste the line from this func to the two places it is used, but that's up to you

crates/forge/src/lib.rs
```diff
@@ -462,35 +487,101 @@ fn run_with_fuzzing(
...
+                    Ok(TestCaseSummary::None {  })
+                },
+                () = runner_config.cancellation_token.cancelled() => {
+                    //Stop executing all tests because flag --exit-first'
+                    //has been set and one test FAIL
+                    Ok(TestCaseSummary::skipped(&case_copy))
+                },
+                () = cancellation_fuzzing_token.cancelled() => {
+                    //Stop executing all tests because flag --exit-first'
+                    //has been set and one test FAIL
```
piotmag769, line 518: Nit: Spaces at starts of comments

crates/forge/src/main.rs
```diff
@@ -117,65 +118,67 @@ fn main_execution() -> Result<bool> {
...
-        let build_output = Command::new("scarb")
-            .arg("build")
-            .stderr(Stdio::inherit())
-            .stdout(Stdio::inherit())
-            .output()
-            .context("Failed to build contracts with Scarb")?;
-        if !build_output.status.success() {
-            bail!("Scarb build did not succeed")
+    let rt = Runtime::new()?;
+    let all_failed_tests = rt.block_on(async {
```
piotmag769, line 122: This doesn't cause parallel execution, does it?

crates/forge/src/lib.rs
```diff
@@ -462,35 +487,101 @@ fn run_with_fuzzing(
     };
     let mut fuzzer = RandomFuzzer::create(fuzzer_seed, fuzzer_runs, &args)?;
 
-    let mut results = vec![];
+    let mut tasks = FuturesUnordered::new();
+
+    let (send, mut recv) = channel(1);
```
piotmag769, line 492: Do we use it anywhere?

crates/forge/src/lib.rs
```diff
@@ -462,35 +487,101 @@ fn run_with_fuzzing(
     };
     let mut fuzzer = RandomFuzzer::create(fuzzer_seed, fuzzer_runs, &args)?;
 
-    let mut results = vec![];
+    let mut tasks = FuturesUnordered::new();
```
piotmag769, line 490: I wonder if we should use `FuturesOrdered` maybe to make fuzzing fail with the same args when using the same seed
cc @MaksymilianDemitraszek

crates/forge/src/lib.rs
```diff
@@ -462,35 +487,101 @@ fn run_with_fuzzing(
...
+                () = runner_config.cancellation_token.cancelled() => {
+                    //Stop executing all tests because flag --exit-first'
+                    //has been set and one test FAIL
+                    Ok(TestCaseSummary::skipped(&case_copy))
+                },
+                () = cancellation_fuzzing_token.cancelled() => {
+                    //Stop executing all tests because flag --exit-first'
+                    //has been set and one test FAIL
+                    Ok(TestCaseSummary::skipped(&case_copy))
+                },
```
piotmag769, line 520: This description is missleading, it is not cancelled bcs of `exit_first`, but because of fuzzing not being run further after the first failed run. 

Also I would introduce a `TestCaseSummary::SkippedDuringFuzzing` or sth like that. This ofc works, but the test is not actually skipped in the sense of `exit_first` flag

crates/forge/src/lib.rs
```diff
@@ -462,35 +487,101 @@ fn run_with_fuzzing(
...
         results.push(result.clone());
 
-        if let TestCaseSummary::Failed { .. } = result {
-            // Fuzz failed
-            break;
+        match &result {
+            TestCaseSummary::Failed { .. } | TestCaseSummary::None {} => {
+                break;
+            }
+            _ => (),
```
piotmag769, line 561: I think we may not yield some tests that were executed bcs we break, but imo it's fine

crates/forge/src/main.rs
piotmag769: cc @MaksymilianDemitraszek @mkaput

mkaput: https://docs.rs/tokio/latest/tokio/runtime/struct.Runtime.html#method.block_on

I'm not reading the code, but if the async block here is doing awaits, joins, whatever, then Tokio will parallelize this.

crates/forge/src/lib.rs
```diff
@@ -285,146 +288,116 @@ fn try_close_tmp_dir(maybe_tmp_dir: Option<TempDir>) -> Result<()> {
...
-    for tests_from_file in tests_iterator {
-        let skipped: Vec<TestCaseSummary> = tests_from_file
-            .test_cases
-            .iter()
-            .map(TestCaseSummary::skipped)
-            .collect();
-
-        for test_case_summary in &skipped {
-            pretty_printing::print_test_result(test_case_summary, None);
+        let (summary, was_fuzzed) = task.await??;
```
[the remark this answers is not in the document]
tkumor3: I can think about it, but the problem is, that in this place I create a tuple to have the contest of a single task.

crates/forge/src/running.rs
tkumor3: I prefer to leave it as a separate function, because I use it in tokio::select!, and it already has quite a complex structure. Any idea for better name?

crates/forge/src/lib.rs
tkumor3: because we waiting for things to finish shutting down - https://tokio.rs/tokio/topics/shutdown

crates/forge/src/lib.rs
tkumor3: the reason why I used FuturesUnordered is faster :)

crates/forge/src/lib.rs
tkumor3: Thx for the catch! I  overlooked it! `cancellation_fuzzing_token.cancelled()` is executed only if signle fuzzing test fails. So the user will get `TestCaseSummary::Failed` after the use of this token!

crates/forge/src/pretty_printing.rs
```diff
@@ -79,12 +85,14 @@ pub fn print_failures(all_failed_tests: &[TestCaseSummary]) {
...
         return;
     }
+    let empty_string = String::new();
     let failed_tests_names: Vec<&String> = all_failed_tests
         .iter()
         .map(|test_case_summary| match test_case_summary {
             TestCaseSummary::Passed { name, .. }
             | TestCaseSummary::Failed { name, .. }
             | TestCaseSummary::Skipped { name, .. } => name,
+            TestCaseSummary::None {} => &empty_string,
```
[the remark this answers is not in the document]
tkumor3: Zrzut ekranu 2023-10-9 o 09 31 08
Zrzut ekranu 2023-10-9 o 09 30 47

crates/forge/src/lib.rs
```diff
@@ -285,146 +288,116 @@ fn try_close_tmp_dir(maybe_tmp_dir: Option<TempDir>) -> Result<()> {
...
+            sierra_program.clone(),
+            Some(MetadataComputationConfig::default()),
+            OrderedHashMap::default(),
+        )
+        .context("Failed setting up runner.")?,
+    );
 
     let mut was_fuzzed = false;
-    let mut results = vec![];
+    let mut tasks = FuturesOrdered::new();
```
cptartur, line 365: I wonder if we need this to be ordered.

crates/forge/src/lib.rs
```diff
@@ -285,146 +288,116 @@ fn try_close_tmp_dir(maybe_tmp_dir: Option<TempDir>) -> Result<()> {
...
-                    },
-                    was_fuzzed,
-                ));
-            }
+    let mut results = vec![];
+
+    while let Some(task) = tasks.next().await {
+        let await_task = task??;
+
+        let result = await_task;
```
cptartur, line 392: This lines could be combined I think

crates/forge/src/lib.rs
```diff
@@ -285,146 +288,116 @@ fn try_close_tmp_dir(maybe_tmp_dir: Option<TempDir>) -> Result<()> {
...
+
+        if result.runs().is_some() {
+            was_fuzzed = true;
```
cptartur, line 395: I'm not sure if we should be detecting if the task was fuzzed like this. This kind of looks like a thing we should just be aware of here in code.

crates/forge/src/lib.rs
```diff
@@ -435,14 +408,67 @@ fn run_tests_from_crate(
...
+    args: Vec<ConcreteTypeId>,
+) -> JoinHandle<Result<TestCaseSummary>> {
+    if args.is_empty() {
+        run_single_test(case, runner, runner_config)
+    } else {
+        tokio::task::spawn(async move {
+            run_with_fuzzing(runner_config.clone(), runner, case, &args).await
+        })
+    }
+}
```
cptartur, line 424: This is inconsistent. We should either be spawning a new task in both if and else arms or in none.

cptartur: This suggests that `run_single_test` is not actually running within a task

crates/forge/src/lib.rs
```diff
@@ -435,14 +408,67 @@ fn run_tests_from_crate(
...
+
+fn run_single_test(
+    case: Arc<TestCase>,
+    runner: Arc<SierraCasmRunner>,
+    runner_config: Arc<RunnerConfig>,
+) -> JoinHandle<Result<TestCaseSummary>> {
+    let exit_first = runner_config.exit_first;
+    tokio::task::spawn(async move {
+        tokio::select! {
+            () = runner_config.exit_first_cancellation_token.cancelled() => {
```
cptartur, line 434: I don't like cancellation tokens being part of runner config. I think they should be passed separately, these are more like implementation details of the task execution.

crates/forge/src/lib.rs
```diff
@@ -435,14 +408,67 @@ fn run_tests_from_crate(
...
+                //Stop executing all tests because flag --exit-first'
+                //has been set and one test FAIL
```
cptartur, line 436:
```suggestion
                // Stop executing all tests because flag --exit-first'
                // has been set and one test FAIL
```

crates/forge/src/lib.rs
```diff
@@ -435,14 +408,67 @@ fn run_tests_from_crate(
...
+                Ok(TestCaseSummary::skipped(&case))
+            },
+            () = runner_config.error_cancellation_token.cancelled() => {
+                //Stop executing all tests because
+                //one of a test returns Err
```
cptartur, line 441:
```suggestion
                // Stop executing all tests because
                // one of a test returns Err
```

crates/forge/src/lib.rs
```diff
@@ -462,35 +488,101 @@ fn run_with_fuzzing(
...
+        let args = args.clone();
+        let cancellation_fuzzing_token = token.clone();
+        let runner_config = runner_config.clone();
+        let send = send.clone();
+
+        tasks.push(task::spawn(async move {
+            tokio::select! {
+                () = runner_config.error_cancellation_token.cancelled() => {
+                    //Stop executing all tests because
+                    //one of a test returns Err
```
cptartur, line 509:
```suggestion
                    //Stop executing all tests because
                    //one of a test returns Err
```
```suggestion
                    // Stop executing all tests because
                    // one of a test returns Err
```

crates/forge/src/lib.rs
```diff
@@ -462,35 +488,101 @@ fn run_with_fuzzing(
...
+                    Ok(TestCaseSummary::None {  })
+                },
+                () = runner_config.exit_first_cancellation_token.cancelled() => {
+                    //Stop executing all tests because flag --exit-first'
+                    //has been set and one test FAIL
```
cptartur, line 514:
```suggestion
                    // Stop executing all tests because flag --exit-first'
                    // has been set and one test FAIL
```

crates/forge/src/lib.rs
```diff
@@ -462,35 +488,101 @@ fn run_with_fuzzing(
...
+                    Ok(TestCaseSummary::skipped(&case_copy))
+                },
+                () = cancellation_fuzzing_token.cancelled() => {
+                    //Stop executing all single fuzzing tests
+                    //because one of fuzzing test has been FAIL
```
cptartur, line 519:
```suggestion
                    // Stop executing all single fuzzing tests
                    // because one of fuzzing test has been FAIL
```

crates/forge/src/lib.rs
```diff
@@ -435,14 +408,67 @@ fn run_tests_from_crate(
...
+                }
+            }
+        }
+    })
+}
+
+async fn run_with_fuzzing(
+    runner_config: Arc<RunnerConfig>,
+    runner: Arc<SierraCasmRunner>,
+    case: Arc<TestCase>,
```
cptartur, line 467: I don't think this needs to be `Arc`, we can just consume the test case in the task.

crates/forge/src/lib.rs
```diff
@@ -462,35 +488,101 @@ fn run_with_fuzzing(
...
-            case,
-            runner_config.fork_targets.as_ref(),
-            &runner_params.contracts,
-            &runner_params.predeployed_contracts,
-            args.clone(),
-            &runner_params.environment_variables,
-        )?;
+    let mut results = vec![];
+
+    drop(send);
```
cptartur, line 552: This could use some explanation/linking to tokio docs, so someone doesn't accidentally remove this line in the future.

crates/forge/src/lib.rs
```diff
@@ -462,35 +488,101 @@ fn run_with_fuzzing(
...
+                    TestCaseSummary::Passed { .. } | TestCaseSummary::Failed { .. }
+                )
+            })
+            .count(),
+    )?;
+
+    let result: TestCaseSummary = results
         .last()
         .expect("Test should always run at least once")
         .clone();
```
cptartur, line 583: This check should probably be executed before counting runs.

crates/forge/src/lib.rs
```diff
@@ -542,63 +657,77 @@ mod tests {
...
-                test_name_filter: None,
-                exact_match: false,
-                exit_first: true,
-                fork_targets: vec![],
-                fuzzer_runs: 100,
-                fuzzer_seed: 32,
-            }
-        );
-    }
+    // #[test]
```
cptartur, line 688: What about this test?

crates/forge/src/main.rs
```diff
@@ -111,71 +112,73 @@ fn main_execution() -> Result<bool> {
...
```
cptartur, line 122: Any reason we're creating an runtime on our own here instead of `tokio::main`? Just to be more explicit?

crates/forge/src/pretty_printing.rs
```diff
@@ -41,17 +41,23 @@ pub(crate) fn print_test_seed(seed: u64) {
...
+pub(crate) fn print_test_result(test_result: &TestCaseSummary) {
+    if let TestCaseSummary::None {} = test_result {
+        return;
+    }
+
     let result_header = match test_result {
         TestCaseSummary::Passed { .. } => format!("[{}]", style("PASS").green()),
         TestCaseSummary::Failed { .. } => format!("[{}]", style("FAIL").red()),
         TestCaseSummary::Skipped { .. } => format!("[{}]", style("SKIP").yellow()),
+        TestCaseSummary::None {} => unreachable!(),
```
cptartur, line 53: Why do we have a test case like this at all if we are expecting to never handle it anyway?

crates/forge/src/lib.rs
tkumor3: we need it in our e2e tests to have repetitivet output. That's the only one reason

crates/forge/src/pretty_printing.rs
tkumor3:
```
pub(crate) fn print_test_result(test_result: &TestCaseSummary) {
    if let TestCaseSummary::None {} = test_result {
        return;
    }

    let result_header = match test_result {
        TestCaseSummary::Passed { .. } => format!("[{}]", style("PASS").green()),
        TestCaseSummary::Failed { .. } => format!("[{}]", style("FAIL").red()),
        TestCaseSummary::Skipped { .. } => format!("[{}]", style("SKIP").yellow()),
        TestCaseSummary::None {} => unreachable!(),
    };
```

because I handled it in this `if` but rust still demand to have `TestCaseSummary::None` in the match statement

crates/forge/src/main.rs
tkumor3: https://github.com/foundry-rs/starknet-foundry/pull/775#discussion_r1345516498 -
It was @MaksymilianDemitraszek suggestion

crates/forge/src/lib.rs
tkumor3: I'm not sure what do you mean

crates/forge/src/lib.rs
cptartur: Okay, perhaps this should be a CLI configuration then? So you can enforce ordered execution if a flag (and we can use it tests) but the default should be unordered?

crates/forge/src/pretty_printing.rs
cptartur: What I meant is what's the purpose of the `TestCaseSummary::None`. I'm not sure why it was introduced.

crates/forge/src/lib.rs
cptartur: We should first try to get `result` from `results` before we count the runs. So we panic early if no result is present.

crates/forge/src/lib.rs
tkumor3: @MaksymilianDemitraszek ?

crates/forge/src/lib.rs
tkumor3: but on the other hand cancellation is part of the runner, `exit_first` works in the context of all tasks

crates/forge/src/pretty_printing.rs
tkumor3: I will recheck purpose of it, because it was introduced before final version

crates/forge/src/lib.rs
cptartur: It is just kind of weird to me to include it as part of the config.

crates/forge/src/lib.rs
tkumor3: I could move this `tokio::task::spawn` in to run_with_fuzzing but this function has 130lines and is very messy

crates/forge/src/pretty_printing.rs
tkumor3: It handled the situation when one of a test returned Err. There are only tests finieshed before Err
Zrzut ekranu 2023-10-9 o 12 33 22

crates/forge/src/lib.rs
tkumor3: in the context of fuzzer test has to pass all "subtests", otherwise is Failed of skiped

crates/forge/src/lib.rs
tkumor3: `run_from_test_case` get `case` as params, and in case of fuzzing we create in default 256 task which use `case`

crates/forge/src/lib.rs
tkumor3: IMO in the first version we can leave `FuturesOrdered ` and we can change it in the next version

crates/forge/src/lib.rs
```diff
@@ -124,36 +155,11 @@ struct TestCrate {
     crate_type: TestCrateType,
 }
 
-pub struct RunnerParams {
-    corelib_path: Utf8PathBuf,
-    contracts: HashMap<String, StarknetContractArtifacts>,
-    predeployed_contracts: Utf8PathBuf,
-    environment_variables: HashMap<String, String>,
-}
```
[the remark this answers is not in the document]
MaksymilianDemitraszek: It shouldn't be done as a part of this PR, it is unrelated and cheanges introduced here are alrady to big.

crates/forge/src/lib.rs
MaksymilianDemitraszek: I don't think it should be ordered, we should write e2e in such way that not requires exact order

crates/forge/src/pretty_printing.rs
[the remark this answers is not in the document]
piotmag769: Still `unreachable!()` though, no?

crates/forge/src/running.rs
piotmag769: Actually it is blocking since it uses `spawn_blocking` 🤦  mb

crates/forge/src/pretty_printing.rs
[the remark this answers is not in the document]
tkumor3: isn't it reachable only for  TestCaseSummary::Failed?

crates/forge/src/lib.rs
piotmag769: Looking from the user perspective I still think it should be ordered, the output will be really messy otherwise. We can create an issue for that and investigate `FuturesUnordered` makes us much faster - if it does ofc speed >>> UI

crates/forge/src/pretty_printing.rs
piotmag769: I suggested renaming `None` to `InterruptedByError` or sth like that

crates/forge/src/test_case_summary.rs
```diff
@@ -28,20 +29,67 @@ pub enum TestCaseSummary {
...
         arguments: Vec<Felt252>,
+        runs: Option<u32>,
     },
     /// Test case skipped (did not run)
     Skipped {
         /// Name of the test case
         name: String,
     },
+    SkippedFuzzing {},
+    Interrupted {},
```
piotmag769, line 40: Can we add docs `///`?

crates/forge/src/lib.rs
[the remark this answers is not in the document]
tkumor3: It's can effect performance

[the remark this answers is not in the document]
piotmag769: But in positive or negative way?

crates/forge/src/lib.rs
piotmag769: +1, issue and investigation if <5% speedup is worth it

crates/forge/src/test_case_summary.rs
```diff
@@ -28,20 +29,69 @@ pub enum TestCaseSummary {
...
         /// Arguments used in the test case run
         arguments: Vec<Felt252>,
+        runs: Option<u32>,
     },
     /// Test case skipped (did not run)
     Skipped {
         /// Name of the test case
         name: String,
     },
+    /// Fuzzing test subtests skipped (did not run)
```
MaksymilianDemitraszek, line 39: Not sure what is the semantics of this result

crates/forge/src/test_case_summary.rs
```diff
@@ -28,20 +29,69 @@ pub enum TestCaseSummary {
...
+    SkippedFuzzing {},
+    /// Tests execution interrupted (by Err)
+    Interrupted {},
```
MaksymilianDemitraszek, line 42: Not sure what is the sementics of this result

crates/forge/src/test_case_summary.rs
```diff
@@ -17,6 +17,7 @@ pub enum TestCaseSummary {
         msg: Option<String>,
         /// Arguments used in the test case run
         arguments: Vec<Felt252>,
+        runs: Option<u32>,
```
MaksymilianDemitraszek, line 20: Imo this being option is weird. Either it should be 1 (which is not complete as fuzzing can fail after one run) if not fuzzed or we should have `Option<FuzzingStatistics>``

crates/forge/src/test_case_summary.rs
```diff
@@ -28,20 +29,69 @@ pub enum TestCaseSummary {
...
+    pub(crate) fn runs(&self) -> Option<u32> {
+        match self {
+            TestCaseSummary::Failed { runs, .. } | TestCaseSummary::Passed { runs, .. } => *runs,
+            TestCaseSummary::Skipped { .. }
+            | TestCaseSummary::Interrupted {}
+            | TestCaseSummary::SkippedFuzzing {} => None,
+        }
+    }
+
+    pub(crate) fn update_runs(self, runs: u32) -> Self {
```
MaksymilianDemitraszek, line 64: I don't know what this functions does

MaksymilianDemitraszek (changes requested):
Test veryfing the fuzz cancellation token is missing, am I correct?

crates/forge/src/lib.rs
piotmag769: #841, out of scope of this PR

crates/forge/src/lib.rs
```diff
@@ -53,6 +63,7 @@ static BUILTINS: Lazy<Vec<&str>> = Lazy::new(|| {
 /// Configuration of the test runner
 #[derive(Deserialize, Debug, PartialEq)]
 pub struct RunnerConfig {
+    workspace_root: Utf8PathBuf,
```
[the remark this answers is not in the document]
piotmag769: Cache

crates/forge/src/test_case_summary.rs
piotmag769: I don't understand what you mean

MaksymilianDemitraszek: @piotmag769 markdown cut the code snippet, fixed

crates/forge/src/lib.rs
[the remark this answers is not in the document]
MaksymilianDemitraszek: We shouldn't pass any path, we should have a separate entity responsible for cache reading and it should be injected to the runner

crates/forge/src/lib.rs
MaksymilianDemitraszek: I mean we don't have to do it here, but we should create an issue to investiagte it

piotmag769: #843

cptartur (commented):
Looks very good after the changes 👍 Left some minor comments

crates/forge/src/pretty_printing.rs
```diff
@@ -79,12 +85,15 @@ pub fn print_failures(all_failed_tests: &[TestCaseSummary]) {
...
         .iter()
         .map(|test_case_summary| match test_case_summary {
-            TestCaseSummary::Passed { name, .. }
-            | TestCaseSummary::Failed { name, .. }
-            | TestCaseSummary::Skipped { name, .. } => name,
+            TestCaseSummary::Failed { name, .. } => name,
+            TestCaseSummary::Passed { .. }
+            | TestCaseSummary::Skipped { .. }
+            | TestCaseSummary::Interrupted {}
+            | TestCaseSummary::SkippedFuzzing {} => unreachable!(),
```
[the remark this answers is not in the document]
piotmag769: Disagree, if we ever pass one of those here it is a bug

crates/forge/src/main.rs
```diff
@@ -111,71 +112,80 @@ fn main_execution() -> Result<bool> {
...
+            let tests_file_summaries = run(
+                &package_path,
+                &package_name,
+                &package_source_dir_path,
+                &dependencies,
+                runner_config,
+                runner_params,
+                cancellation_tokens,
+            )
+            .await?;
```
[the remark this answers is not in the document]
piotmag769: #845

crates/forge/src/lib.rs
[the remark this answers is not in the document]
piotmag769: Now that I think about it the current solution should be optimal (or almost optimal, better would be with `FuturesUnordered`)
cc @MaksymilianDemitraszek

crates/forge/src/pretty_printing.rs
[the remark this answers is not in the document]
cptartur: I see your point, resolving.

crates/forge/src/lib.rs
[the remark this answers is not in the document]
tkumor3: could we improved it in https://github.com/foundry-rs/starknet-foundry/issues/842?

crates/forge/src/lib.rs
```diff
@@ -92,6 +106,49 @@ impl RunnerConfig {
...
+        let error = CancellationToken::new();
+        Self { exit_first, error }
+    }
+}
+
+impl Default for CancellationTokens {
+    fn default() -> Self {
+        Self::new()
+    }
+}
```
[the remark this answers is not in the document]
tkumor3: cargo linter suggest - [clippy](https://rust-lang.github.io/rust-clippy/master/index.html#/new_without_default)

crates/forge/src/pretty_printing.rs
```diff
@@ -41,17 +41,23 @@ pub(crate) fn print_test_seed(seed: u64) {
     println!("{}: {seed}", style("Fuzzer seed").bold());
 }
 
-pub(crate) fn print_test_result(test_result: &TestCaseSummary, fuzzer_runs: Option<u32>) {
+pub(crate) fn print_test_result(test_result: &TestCaseSummary) {
+    if let TestCaseSummary::Interrupted {} | TestCaseSummary::SkippedFuzzing {} = test_result {
+        return;
+    }
```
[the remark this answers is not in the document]
tkumor3: ok, so `unreachable!()` is enough?

MaksymilianDemitraszek (approved):

crates/forge/src/lib.rs
```diff
@@ -435,62 +451,222 @@ fn run_tests_from_crate(
...
+                // Stop executing all tests because flag --exit-first'
+                // has been set and one test FAIL
+                Ok(TestCaseSummary::skipped(&case))
+            },
+            () = cancellation_tokens.error.cancelled() => {
+                // Stop executing all tests because
+                // one of a test returns Err
+                Ok(TestCaseSummary::Interrupted {  })
+            },
+            result = blocking_run_from_test(vec![], case.clone(),runner,  runner_config.clone(), runner_params.clone() , None) => {
```
cptartur, line 502: Making sure, do we need clones here?

cptartur (approved):

crates/forge/src/lib.rs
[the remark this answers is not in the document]
piotmag769: https://github.com/foundry-rs/starknet-foundry/issues/853

PR #780, opened 2023-10-02
Refactor test collecting

cptartur:
Closes #

## Introduced changes

- This PR separates collecting tests with compiling test cases, so they can be tested separately. Also some of the code has been converted from free functions to struct methods.
- Replaced all occurrences of `&Utf8PathBuf` with `&Utf8Path`, same for `&PathBuf`.

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/forge/src/lib.rs
```diff
@@ -502,28 +377,27 @@ fn function_args<'a>(function: &'a Function, builtins: &[&str]) -> Vec<&'a Concr
...
+//         if exact_match {
+//             if test.name == test_name_filter {
+//                 result.push(test);
+//             }
+//         } else if test.name.contains(test_name_filter) {
+//             result.push(test);
+//         }
+//     }
+//     result
+// }
```
piotmag769, line 396: commented out code

piotmag769 (approved):

cptartur:
> I would like to review this one but after it is adapted to changes of #775

@MaksymilianDemitraszek it is ready for review now

piotmag769 (approved):

crates/forge/src/collecting.rs
```diff
@@ -0,0 +1,588 @@
...
+    let test_dir_crate = if tests_dir_path.exists() {
+        let lib_path = tests_dir_path.join("lib.cairo");
+        if lib_path.exists() {
+            Some(TestCrate {
+                crate_root: tests_dir_path,
+                crate_name: "tests".to_string(),
+                crate_type: TestCrateType::Tests,
+            })
+        } else {
+            Some(pack_tests_into_one_file(temp_dir, &tests_dir_path)?)
```
MaksymilianDemitraszek, line 92: Why this is done inside of the collecting function?

crates/forge/src/collecting.rs
```diff
@@ -0,0 +1,588 @@
...
+        tests_from_crates
+            .into_iter()
+            .map(|tc| tc.filter_by_name(test_name_filter, runner_config.exact_match))
+            .collect()
+    } else {
+        tests_from_crates
+    }
+}
+
+pub fn pack_tests_into_one_file(
```
MaksymilianDemitraszek, line 136: Isn't it packing into one crate not file?

crates/forge/src/collecting.rs
```diff
@@ -0,0 +1,588 @@
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::scarb::ForgeConfig;
+    use assert_fs::fixture::PathCopy;
+    use assert_fs::TempDir;
+    use test_collector::ExpectedTestResult;
+
+    #[test]
+    fn runner_config_argument_precedence() {
```
MaksymilianDemitraszek, line 195: Why this is tested here, not in the RunnerConfig file?

crates/forge/src/collecting.rs
```diff
@@ -0,0 +1,588 @@
...
+                exit_first: true,
+                fork_targets: vec![],
+                fuzzer_runs: 100,
+                fuzzer_seed: 32,
+            }
+        );
+    }
+
+    #[test]
+    fn collecting_test_crates() {
```
MaksymilianDemitraszek, line 226: collecting_test_compilation_targets

crates/forge/src/collecting.rs
```diff
@@ -0,0 +1,588 @@
...
+
+        Ok(TestsFromCrate {
+            sierra_program,
+            test_cases,
+            test_crate_type: self.crate_type,
+        })
+    }
+}
+
+pub fn collect_test_crates(
```
MaksymilianDemitraszek, line 75:
```suggestion
pub fn collect_test_compilation_targets(
```

crates/forge/src/collecting.rs
MaksymilianDemitraszek: Test cases:
How it behaves when there is no lib
How it behaves where there are no tests only lib

crates/forge/src/collecting.rs
```diff
@@ -0,0 +1,588 @@
...
+
+        let temp_dir = TempDir::new().unwrap();
+        let tests = pack_tests_into_one_file(&temp_dir, &tests_path).unwrap();
+        let virtual_lib_path = tests.crate_root.join("lib.cairo");
+        let virtual_lib_u8_content = std::fs::read(&virtual_lib_path).unwrap();
+        let virtual_lib_content = std::str::from_utf8(&virtual_lib_u8_content).unwrap();
+
+        assert!(virtual_lib_path.try_exists().unwrap());
+        assert!(virtual_lib_content.contains("mod contract;"));
+        assert!(virtual_lib_content.contains("mod ext_function_test;"));
```
MaksymilianDemitraszek, line 266: Test case: what if names are duplicates? It can't happen because files in dir always have unique names, right?

crates/forge/src/collecting.rs
```diff
@@ -0,0 +1,588 @@
...
+                available_gas: None,
+                expected_result: ExpectedTestResult::Success,
+                fork_config: None,
+                fuzzer_config: None,
+            },]
+        );
+    }
+
+    #[test]
+    fn filtering_tests_works_without_crate_in_test_name() {
```
MaksymilianDemitraszek, line 531: I don't understand what this test verifies

crates/forge/src/collecting.rs
```diff
@@ -0,0 +1,588 @@
...
+        Self {
+            sierra_program: self.sierra_program.clone(),
+            test_cases: result.into_iter().cloned().collect(),
+            test_crate_type: self.test_crate_type,
+        }
+    }
+}
+
+#[derive(Debug, PartialEq, Clone, Copy)]
+pub enum TestCrateType {
```
MaksymilianDemitraszek, line 38:
```suggestion
pub enum TestCrateLocation {
```

crates/forge/src/collecting.rs
```diff
@@ -0,0 +1,588 @@
...
+            libfunc_declarations: vec![],
+            statements: vec![],
+            funcs: vec![],
+        }
+    }
+
+    #[test]
+    #[allow(clippy::too_many_lines)]
+    fn filtering_tests() {
+        let mocked_tests = TestsFromCrate {
```
MaksymilianDemitraszek, line 283: What if has no tests?

crates/forge/src/collecting.rs
```diff
@@ -0,0 +1,588 @@
...
+        let path = entry.path();
+
+        if path.is_file() && path.extension().unwrap_or_default() == "cairo" {
+            let mod_name = path
+                .strip_prefix(tests_folder_path)
+                .expect("Each test file path should start with package path")
+                .to_str()
+                .context("Unable to convert test file path to string")?
+                .strip_suffix(".cairo")
+                .expect("Each test file path should have .cairo extension");
```
MaksymilianDemitraszek, line 168: should be extracted to a separate function

MaksymilianDemitraszek (changes requested):
Generally looks good, it is a positive change , It is ok to include it to the codebase after we remove cloning on large data, make it more idiomatic using iterators and improve names.

crates/forge/src/collecting.rs
```diff
@@ -0,0 +1,588 @@
...
+            if exact_match {
+                if test.name == filter {
+                    result.push(test);
+                }
+            } else if test.name.contains(filter) {
+                result.push(test);
+            }
+        }
+        Self {
+            sierra_program: self.sierra_program.clone(),
```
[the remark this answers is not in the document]
piotmag769: +1, changed it in #801

crates/forge/src/collecting.rs
```diff
@@ -0,0 +1,588 @@
...
+) -> Vec<TestsFromCrate> {
+    if let Some(test_name_filter) = &runner_config.test_name_filter {
+        tests_from_crates
+            .into_iter()
+            .map(|tc| tc.filter_by_name(test_name_filter, runner_config.exact_match))
+            .collect()
+    } else {
+        tests_from_crates
+    }
+}
```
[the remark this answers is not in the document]
piotmag769: Would leave it, I extend this function in #886

crates/forge/src/collecting.rs
```diff
@@ -0,0 +1,588 @@
...
+use camino::Utf8PathBuf;
+use rayon::prelude::{IntoParallelRefIterator, ParallelIterator};
+use test_collector::{collect_tests, LinkedLibrary, TestCase};
+use walkdir::WalkDir;
+
+pub struct TestsFromCrate {
+    pub sierra_program: Program,
+    pub test_cases: Vec<TestCase>,
+    pub test_crate_type: TestCrateType,
+}
```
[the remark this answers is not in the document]
cptartur: Both names are a bit confusing, but it's not really a direct representation of a crate.

crates/forge/src/collecting.rs
piotmag769: > How it behaves where there are no tests only lib

I don't think we need to test that, it is kinda testing the behaviour of the compiler

> How it behaves when there is no lib

It is already tested in e2e tests: `collection_with_lib` and `collection_without_lib` packages

crates/forge/src/collecting.rs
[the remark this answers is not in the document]
piotmag769: What about `CompiledTestCrate`?

crates/forge/src/collecting.rs
piotmag769: It is kinda a part of test collecting (if we want to return test crates from this function)

crates/forge/src/collecting.rs
```diff
@@ -0,0 +1,588 @@
...
+
+impl TestsFromCrate {
+    pub fn filter_by_name(&self, filter: &str, exact_match: bool) -> Self {
```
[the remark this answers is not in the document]
cptartur: This way it's easier to test the filtering behavior, because we have a specific functions that can be called in unit tests. Otherwise we would have to test by calling a higher level function.

crates/forge/src/collecting.rs
cptartur: > Test cases: How it behaves when there is no lib 
> How it behaves where there are no tests only lib

so in both these cases the returned struct will contain a `TestCompilationTarget` representing a lib. It's hardcoded in the assumption that this always exists. 
The path to the lib is passed by the user as an argument, so I'm unsure if we should verify it's existence inside this method.

crates/forge/src/collecting.rs
cptartur: It was meant to verify that even for tests which somehow doesn't have a crate name attached (no part before `::`) the filtering logic works correctly. I'll combine it with other tests though.

crates/forge/src/collecting.rs
piotmag769: Still imo e2e tests are enough there

crates/forge/src/collecting.rs
cptartur: So a test case for that doesn't make much sense

crates/forge/src/collecting.rs
cptartur: Tests stay an empty vector, I'll add an unit test for that.

crates/forge/src/collecting.rs
MaksymilianDemitraszek: I disagree, this is detail of compilation not collection, please rename the function or move it to compilation phase.

crates/forge/src/collecting.rs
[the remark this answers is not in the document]
MaksymilianDemitraszek: If you would implemented it idomatically with iterator you could possibly only test matching function.

crates/forge/src/collecting.rs
```diff
@@ -0,0 +1,590 @@
...
+use rayon::prelude::{IntoParallelRefIterator, ParallelIterator};
+use test_collector::{collect_tests, LinkedLibrary, TestCase};
+use walkdir::WalkDir;
+
+#[derive(Debug, Clone)]
+pub struct CompiledTests {
+    pub sierra_program: Program,
+    pub test_cases: Vec<TestCase>,
+    pub tests_location: CrateLocation,
+}
```
MaksymilianDemitraszek, line 16: This is also a bad name. It suggest that it contains only compiled code. It also doesn't give an information of the scope, is this an arbitrary set of tests? Or tests from a module.  Please rename it to `TestCrate`

piotmag769: `CompiledTestCrate`, we have `TestCrate` struct already

MaksymilianDemitraszek: It has been renamed.

MaksymilianDemitraszek: I am ok with: `TestCrate`, `CrateWithTests` or `CompiledTestCrate`, your call @cptartur

piotmag769 (approved):

crates/forge/src/collecting.rs
MaksymilianDemitraszek: > >   How it behaves where there are no tests only lib
> 
> I don't think we need to test that, it is kinda testing the behaviour of the compiler

I am not convinced, compiler doesn't know about test dir, we implement that on the top of it. How collection is going to behave if we call it on a dir without such directory?

crates/forge/src/collecting.rs
MaksymilianDemitraszek: I am not sure it this test actually test something useful. It is kinda testing logic of `name.contains` which is part of the std lib

MaksymilianDemitraszek (approved):
I am ok with merging this one after you remove packing tests from collecting phase

cptartur:
> I am ok with merging this one after you remove packing tests from collecting phase

I'll do that

crates/forge/src/collecting.rs
cptartur: Yeah, it's been removed, I've just extanded mocked_tests in other unit tests

crates/forge/src/collecting.rs
```diff
@@ -0,0 +1,682 @@
...
+use test_collector::{collect_tests, LinkedLibrary, TestCase};
+use walkdir::WalkDir;
+
+// TODO docstrings in file
+
+#[derive(Debug, Clone)]
+pub(crate) struct CompiledTestCrate {
+    pub sierra_program: Program,
+    pub test_cases: Vec<TestCase>,
+    pub tests_location: CrateLocation,
```
MaksymilianDemitraszek, line 17:
```suggestion
    pub tests_source_location: CrateLocation,
```

crates/forge/src/collecting.rs
```diff
@@ -0,0 +1,682 @@
...
+use anyhow::{anyhow, Context, Result};
+use assert_fs::fixture::{FileTouch, PathChild, PathCopy};
+use assert_fs::TempDir;
+use cairo_lang_sierra::program::Program;
+use camino::{Utf8Path, Utf8PathBuf};
+use rayon::prelude::{IntoParallelRefIterator, ParallelIterator};
+use test_collector::{collect_tests, LinkedLibrary, TestCase};
+use walkdir::WalkDir;
+
+// TODO docstrings in file
```
piotmag769, line 11: todo

piotmag769 (approved):
We merge thoooose

cptartur:
> We merge thoooose

Waiting for @MaksymilianDemitraszek re-review

MaksymilianDemitraszek (approved):
I would consider spliting test compilation target candidate (one without lib) and test compilation target instead of keeping it as a single type, imo this is a reason behind your problems in describing this contract, how `lib.cairo` is ensured is an implementation detail so there could be alternative strategies.  

Overall it improves state of the codebase good job, LGTM feel free to merge.

Issue #757, opened 2023-09-28
Add printing used block id in fork when `latest` to allow reproducing


MaksymilianDemitraszek:

piotmag769:
Changing the title as we plan to get rid of `Pending` block #938

PR #783, opened 2023-10-02
Refactor for printing with `BlockTag::Latest` and `BlockTag::Pending`

piotmag769:
Refactors changes for #757 (blocked by #795)
Raised when working on this PR: #791

## Introduced changes

- `RawForgeConfig` to serialize data from metadata. RawForgeConfig is parsed into ForgeConfig using `TryFrom` trait, allowing us to early catch e.g. URL parsing errors
- test-collector: `RawForkConfig` parsed to `ForkConfig` to parse URL early and return errors during compilation.
 
## Checklist

- [ ] Linked relevant issue
- [x] Added relevant tests


crates/forge/src/pretty_printing.rs
```diff
@@ -72,7 +79,34 @@ pub(crate) fn print_test_result(test_result: &TestCaseSummary, fuzzer_runs: Opti
...
+        match block_id {
+            BlockId::Tag(_) => {
+                let client = JsonRpcClient::new(HttpTransport::new(url.clone()));
+                let runtime = Runtime::new().expect("Could not instantiate Runtime");
+                match runtime.block_on(client.get_block_with_txs(block_id))? {
+                    Block(BlockWithTxs {
+                        block_number,
+                        block_hash,
+                        ..
+                    }) => format!("\nBlock number: {block_number}, block hash: {block_hash}"),
```
MaksymilianDemitraszek, line 97: I don't think this will always match the block used in tests, does it?

piotmag769: #791

crates/test-collector/src/lib.rs
```diff
@@ -63,10 +64,30 @@ pub enum ExpectedTestResult {
...
 pub enum ForkConfig {
+    Id(String),
+    Params(Url, BlockId),
+}
+
+#[derive(Debug, Clone, PartialEq)]
+enum RawForkConfig {
     Id(String),
     Params(String, BlockId),
 }
```
war-in, line 75: Do we really need to have another struct for handling parsing URL errors?

crates/forge/tests/e2e/diagnostics_and_plugins.rs
```diff
@@ -37,7 +42,7 @@ fn print_error_if_attributes_incorrect() {
              --> test_fuzzer.cairo:8:9
             #[fuzzer()]
                     ^^
-
+
```
war-in, line 45: 👀

war-in (approved):

crates/test-collector/src/lib.rs
piotmag769: I found it the easiest to do while retaining the current interface. Can refactor it if you want to

crates/forge/tests/e2e/diagnostics_and_plugins.rs
piotmag769: Was just a newline here before and I used tab on multiple lines, it doesn't change anything

crates/forge/src/lib.rs
```diff
@@ -31,7 +30,7 @@ pub mod scarb;
 pub mod test_case_summary;
 
 mod fuzzer;
-mod running;
+pub mod running;
```
cptartur, line 33: Does it need to be pub? We can explicitly reexport some functions with `pub use ...` if they are needed somewhere.

crates/forge/src/pretty_printing.rs
```diff
@@ -72,7 +79,34 @@ pub(crate) fn print_test_result(test_result: &TestCaseSummary, fuzzer_runs: Opti
...
+                    }
+                }
+            }
+            _ => String::new(),
+        }
+    } else {
+        String::new()
+    };
+
+    Ok(result)
```
cptartur, line 109: This is not a way to go. We shouldn't be spawning any runtimes in the code that just does printing. If it's impossible to get the actual block id used, code of cheatnet should be changed so this value is simply returned/accessible to us.

crates/forge/src/scarb.rs
```diff
@@ -38,9 +41,28 @@ pub struct StarknetContractArtifacts {
...
+pub struct ForkTarget {
+    pub name: String,
+    pub url: Url,
+    pub block_id: BlockId,
+}
+
 /// Represents forge config deserialized from Scarb.toml
 #[derive(Deserialize, Debug, PartialEq, Default)]
-pub struct ForgeConfig {
+pub struct RawForgeConfig {
```
cptartur, line 65: Do then need to still be public? Also documentation is no longer valid for these, should be updated.

crates/forge/src/scarb.rs
```diff
@@ -50,16 +72,63 @@ pub struct ForgeConfig {
...
+            let [block_id] = block_id[..] else {
+                unreachable!()
+            };
+
+            fork_targets.push(ForkTarget {
+                name: raw_fork_target.name,
+                url: Url::parse(&raw_fork_target.url).map_err(|parse_error| {
+                    anyhow!(
+                        "Could not parse the \"{}\" URL from Scarb.toml: {parse_error}",
+                        raw_fork_target.url
```
cptartur, line 116: Nit but we usually use constructs like `url = {}` in these errors instead of `""` or `:`.

crates/forge/src/scarb.rs
```diff
@@ -50,16 +72,63 @@ pub struct ForgeConfig {
...
 
 #[derive(Deserialize, Debug, PartialEq, Default, Clone)]
-pub struct ForkTarget {
+pub struct RawForkTarget {
     pub name: String,
     pub url: String,
     pub block_id: HashMap<String, String>,
 }
 
+impl TryFrom<RawForgeConfig> for ForgeConfig {
```
cptartur, line 85: I wonder if it would be possible to resolve by having some kind of custom deserializer just for the `ForkTarget` fields. Without duplicating all structs etc.

Should be investigated if `serde` provides such a functionality.

crates/forge/tests/e2e/forking.rs
```diff
@@ -0,0 +1,40 @@
...
+        [..]Finished[..]
+
+
+        Collected 4 test(s) from forking package
+        Running 0 test(s) from src/
+        Running 4 test(s) from tests/
+        [PASS] tests::test_forking::latest
+        Block number: [..], block hash: [..]
+        [PASS] tests::test_forking::pending
+        Block parent hash: [..]
```
cptartur, line 26: I wonder if naming it "block parent" is clear enough. Maybe just "previous block"?

cptartur (changes requested):
Didn't manage to review whole PR yet, will finish later

MaksymilianDemitraszek:
I don't think there is a point merging this one before chatnet fixes their issue

crates/forge/src/lib.rs
piotmag769: Will try that, but I believe they have to be pub in the first place

crates/forge/src/pretty_printing.rs
piotmag769: There is an issue for that, it is a temporary solution #791

crates/forge/tests/e2e/forking.rs
piotmag769: I used the name from blockifier, but can change it

crates/forge/src/scarb.rs
piotmag769: Was wondering about implementing `Serde` manually too, however I believe I would have to use `Serde` api instead of manually converting the already deserialized output. Can investigate on that

piotmag769: https://serde.rs/deserialize-struct.html it is quite complex. It is probably easier to leave it this way

crates/forge/src/lib.rs
piotmag769: Yep, it doesn't work, have to be pub in the first place
https://stackoverflow.com/questions/69275034/what-is-the-difference-between-use-and-pub-use

MaksymilianDemitraszek (changes requested):
There is no issue linked.

Does the PR name actually describe introduces changes 🤔 ?

crates/test-collector/src/lib.rs
```diff
@@ -63,10 +64,30 @@ pub enum ExpectedTestResult {
...
 
+fn fork_config_from_raw_fork_config(
```
MaksymilianDemitraszek, line 77: why not `impl ForkConfig::from`?

crates/test-collector/src/lib.rs
```diff
@@ -63,10 +64,30 @@ pub enum ExpectedTestResult {
...
+    maybe_raw_fork_config: Option<RawForkConfig>,
+) -> Result<Option<ForkConfig>> {
+    let result = match maybe_raw_fork_config {
+        Some(RawForkConfig::Params(raw_url, block_id)) => {
+            let url = Url::parse(&raw_url)?;
+            Some(ForkConfig::Params(url, block_id))
+        }
+        Some(RawForkConfig::Id(name)) => Some(ForkConfig::Id(name)),
```
MaksymilianDemitraszek, line 85: Why do we need to leak ID logic to the deeper parts of the codebase instead of resolving it somewhere here?

crates/test-collector/src/lib.rs
piotmag769: Should use `TryFrom`, true

crates/test-collector/src/lib.rs
piotmag769: I would need to pass the Scarb config to the compilation crate, idk if that 's a good idea, but can do that

cptartur (commented):
I'll review this once the split into smaller PRs is done, please re-request review then

piotmag769:
Closing, starting again with #919

PR #793, opened 2023-10-03
521 Docs Cheatcodes - update code example for roll/warp

### LINKED ISSUE #521: Add a more relatable example for using cheatcodes in the constructor

Forge

### Feature Request

For now, we present all of the cheatcodes behavior using `prank` as an example, but using `roll`/`warp` would be more practical for users to comprehend the nature of the "mocking constructor" flow.

As a part of this task you should:
- Delete the old example
- Write a contract and a test that uses this behavior (with comments preferably)
- Explain to the user in documentation how this works and what it can be used for

robertkodra:
Closes #521

## Introduced changes

- removed old code example for roll/warp cheatcode
- added a new code example that mocks the constructor for roll/warp cheatcode

## Checklist

- [x] Linked relevant issue
- [ ] Added relevant tests


Arcticae (changes requested):
Hey @robertkodra! 
Thanks for the contribution :) 
What i meant in the task is to make a dedicated example for this file:
`docs/src/testing/using-cheatcodes.md:177`
There is a description of using cheatcodes, but it does not include a special contract for that purpose, it's discussed "in theory". The contracts you provided are fine, but we should present them in the file above. 

We could also make it that it doesn't require 3 separate contracts, but rather tests all 3 cheats in the same deploy.

The cheatcode appendix should not be changed. 

Let me know if you want more details or explanation :)

piotmag769:
@robertkodra do you plan to continue working on this?

robertkodra:
@piotmag769 Sending a PR by tomorrow. Apologize for the holdup.

robertkodra:
@Arcticae Can you review and let me know if there are any changes to be made. I've incorporated all of them in one contract. One thing that might be worth asking is if we want to keep the same contract name 'HelloStarknet' or have it as something else. 

Also I've worked on the latest version of `sn-foundry` and bumped the version from the top of the page.  I'm mentioning this because on the rest of the docs, I believe the version is  `0.7.1`.

docs/src/testing/using-cheatcodes.md
```diff
@@ -174,32 +197,81 @@ Failures:
...
     start_prank(contract_address, 123.try_into().unwrap());
-    
-    // The constructor will have 123 set as the caller address 
+
+    // The constructor will have 123 set as the caller address
     contract.deploy(constructor_arguments).unwrap();
 }
 ```
+
+### Mocking the constructor with `roll`
```
Arcticae, line 225: This is a little excessive, since if we wanted to be consistent with it, we'd have to document all of the cheatcodes that work in constructors in this way.

I would delete everything after this line for brevity.

docs/src/testing/using-cheatcodes.md
```diff
@@ -174,32 +197,81 @@ Failures:
     package_name::call_and_invoke
 ```
 
-### Pranking the constructor
+### Mocking the constructor with `prank`
```
Arcticae, line 200: This was meant to be a "generic" section which it should stay

docs/src/testing/using-cheatcodes.md
```diff
@@ -15,45 +16,68 @@ using [cheatcodes](../appendix/cheatcodes.md).
...
             self.balance.read()
         }
+        // Gets the block number
+        fn get_block_number(self: @ContractState) -> u64 {
+            self.blk_nb.read()
+        }
+        // Gets the block timestamp
+        fn get_block_timestamp(self: @ContractState) -> u64 {
+            self.blk_timestamp.read()
+        }
```
Arcticae, line 69: The code does not really say which block number we are getting, this is specifically the `constructor` value so it's rather 
`get_block_timestamp_at_constructor_invocation` or `get_block_timestamp_at_construction`

docs/src/testing/using-cheatcodes.md
```diff
@@ -174,32 +197,81 @@ Failures:
...
-
+To `prank` the constructor, you need to `start_prank` before it is invoked, with the right address. To achieve this, you need to precalculate the address of the contract by using the `precalculate_address` function of `ContractClassTrait` on the declared contract, and then use it in `start_prank` as an argument:
 
 ```rust
 use snforge_std::{ declare, ContractClassTrait, start_prank };
 
 #[test]
-fn prank_the_constructor() {
+fn mock_constructor_with_prank() {
+    // declaring contract
```
Arcticae, line 213:
```suggestion
```

docs/src/testing/using-cheatcodes.md
```diff
@@ -174,32 +197,81 @@ Failures:
...
+    // set the block number to 234 for the precalculated address
+    start_roll(contract_address, 234);
+    // deploying contract
+    let contract_address = contract.deploy(@ArrayTrait::new()).unwrap();
+    // set dispatcher
+    let dispatcher = IHelloStarknetDispatcher { contract_address };
+    // retrieving the block number
+    let block_number = dispatcher.get_block_number();
+    // asserting if block number is 234 set by start_roll()
+    assert(block_number == 234, 'Wrong block number');
```
Arcticae, line 249: The comments on this are a little excessive. Please concentrate on "why" instead of "what" when explaining the code to the user

Arcticae (changes requested):

robertkodra:
Thank you for the feedback @Arcticae . Made the changes, lmk.

Arcticae (approved):
Thanks for the contribution @robertkodra!

PR #799, opened 2023-10-03
Test not panickable error message changed

piotmag769:


Arcticae (approved):

crates/test-collector/src/lib.rs
```diff
@@ -595,7 +595,10 @@ fn validate_tests(
         }
         if let Some(return_type_name) = maybe_return_type_name {
             if !return_type_name.starts_with("core::panics::PanicResult::") {
-                anyhow::bail!("Test function {} must be panicable but it's not", test.name);
+                anyhow::bail!(
+                    "The test function {} cannot fail, thus it always passes - make sure to include panickable statements in your test",
```
cptartur, line 599:
```suggestion
                    "The test function {} always succeeds and cannot be used as a test. Make sure to include panickable statements such as `assert` in your test",
```

PR #801, opened 2023-10-04
Support #[ignore] attribute

piotmag769:
## Introduced changes

- Support ignored attribute

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/forge/src/collecting.rs
```diff
@@ -8,16 +8,17 @@ use rayon::prelude::{IntoParallelRefIterator, ParallelIterator};
...
+#[cfg_attr(test, derive(Clone))]
 pub struct TestsFromCrate {
     pub sierra_program: Program,
     pub test_cases: Vec<TestCase>,
     pub test_crate_type: TestCrateType,
 }
 
 impl TestsFromCrate {
-    pub fn filter_by_name(&self, filter: &str, exact_match: bool) -> Self {
+    pub fn filter_by_name(self, filter: &str, exact_match: bool) -> Self {
```
piotmag769, line 19: We do not want to copy that sierra if we consume it anyways

crates/forge/src/collecting.rs
```diff
@@ -123,13 +137,23 @@ pub fn filter_tests_from_crates(
...
+    let filtered_by_name = if let Some(test_name_filter) = &runner_config.test_name_filter {
         tests_from_crates
             .into_iter()
             .map(|tc| tc.filter_by_name(test_name_filter, runner_config.exact_match))
             .collect()
     } else {
         tests_from_crates
+    };
+
+    // TODO: add tests for filter_tests_from_crate
```
piotmag769, line 149: Will do on Friday

crates/forge/src/collecting.rs
```diff
@@ -8,16 +8,17 @@ use rayon::prelude::{IntoParallelRefIterator, ParallelIterator};
 use test_collector::{collect_tests, LinkedLibrary, TestCase};
 use walkdir::WalkDir;
 
+#[cfg_attr(test, derive(Clone))]
```
piotmag769, line 11: To be able to use clone, but only in tests

cptartur: I think we could derive it for all cases, doesn't matter that much.

crates/forge/src/collecting.rs
```diff
@@ -27,8 +28,21 @@ impl TestsFromCrate {
...
         Self {
-            sierra_program: self.sierra_program.clone(),
-            test_cases: result.into_iter().cloned().collect(),
+            sierra_program: self.sierra_program,
+            test_cases: result,
+            test_crate_type: self.test_crate_type,
+        }
+    }
+
+    pub fn extract_ignored(self) -> Self {
```
cptartur, line 37: Does it need to be `pub` if we only use it in this file? I'd say it's mostly an implementation detail - does not need to be a part of the public api.

crates/forge/src/lib.rs
```diff
@@ -56,11 +56,13 @@ static BUILTINS: Lazy<Vec<&str>> = Lazy::new(|| {
...
 /// Configuration of the test runner
 #[derive(Deserialize, Debug, PartialEq)]
+#[allow(clippy::struct_excessive_bools)]
 pub struct RunnerConfig {
     test_name_filter: Option<String>,
     exact_match: bool,
     exit_first: bool,
-    fork_targets: Vec<ForkTarget>,
+    only_ignored: bool,
+    include_ignored: bool,
```
cptartur, line 65: I guess this could be changed to something like
```rust
// temp name
enum Run {
  All,
  OnlyIgnored,
  OnlyNonIgnored
}

pub struct RunnerConfig {
  // ...
  run: Run
  // ...
}
```

but I'm not sure if it is better at all.

crates/forge/src/lib.rs
```diff
@@ -124,13 +131,15 @@ impl RunnerParams {
         predeployed_contracts: Utf8PathBuf,
         environment_variables: HashMap<String, String>,
         linked_libraries: Vec<LinkedLibrary>,
+        forge_config: ForgeConfig,
```
cptartur, line 134: I'd take `Vec<ForkTarget>` directly here and extract at caller site.

crates/forge/src/main.rs
```diff
@@ -53,6 +54,13 @@ struct Args {
     /// Clean forge cache directory
     #[arg(short, long)]
     clean_cache: bool,
+
+    /// Run only tests marked with `#[ignore]` attribute
+    #[arg(long = "ignored")]
+    only_ignored: bool,
+    /// Run all tests regardless of `#[ignore]` attribute
+    #[arg(long, conflicts_with = "only_ignored")]
+    include_ignored: bool,
```
cptartur, line 63:
```suggestion

    /// Run all tests regardless of `#[ignore]` attribute
    #[arg(long, conflicts_with = "only_ignored")]
    include_ignored: bool,
```

crates/forge/src/running.rs
```diff
@@ -68,17 +69,24 @@ fn build_hints_dict<'b>(
...
 ) -> Result<TestCaseSummary> {
+    // if only_ignored == true then tests without `#[ignore]` are filtered earlier
+    let should_be_run =
+        runner_config.include_ignored || runner_config.only_ignored || !case.ignored;
+
+    if !should_be_run {
+        return Ok(TestCaseSummary::Ignored {
+            name: case.name.clone(),
+        });
+    }
```
cptartur, line 88: This should be handler somewhere higher. If we enter `run_from_test_case` at all, the assumption should be that we do run the function.

docs/src/testing/testing.md
```diff
@@ -91,3 +91,27 @@ Running 1 test(s) from src/
...
+```shell
+$ snforge
+Collected 1 test(s) from package_name package
+Running 1 test(s) from src/
+[IGNORE] src::expensive_test
+Tests: 1 passed, 0 failed, 0 skipped
+```
+
+You can run only the ignored tests using `snforge --ignored`. 
+To run all tests regardless of the `#[ignore]` attribute use `snforge --include-ignored`.
```
cptartur, line 117: Perhaps an example output with `--include-ignored` or `--ignored` should be added here? Though I'm not sure.

crates/forge/src/main.rs
MaksymilianDemitraszek: What is the purpose of `only_ignored` flag? imo is not that much useful

crates/forge/src/running.rs
MaksymilianDemitraszek: +1 We should just have a filter which passes a list of relevant test cases

MaksymilianDemitraszek (changes requested):
Logic wise it seems correct, `--ignored` flag is useless imo don't see a point including it. We should also abstract filtering in the eariler stages of execution as @cptartur mentioned

crates/forge/src/main.rs
cptartur: `cargo test` has `--ignored` and `--include-ignored` as well. Not sure if there was any other motivation to include these in `snforge.

MaksymilianDemitraszek: Discussed with @piotmag769 we can leave it as it is but we should not recommend using it for expensive tests in docs

docs/src/testing/testing.md
piotmag769: I don't think it is that important to clog the docs with it

crates/forge/src/collecting.rs
piotmag769: I would leave it as it is

crates/forge/src/collecting.rs
piotmag769: Actually it is tested in e2e tests, the function filtering ignore is so simple that I won't unit test it

crates/forge/src/collecting.rs
cptartur: This just adds complication to the code without any benefit.

crates/forge/src/lib.rs
```diff
@@ -54,13 +53,41 @@ static BUILTINS: Lazy<Vec<&str>> = Lazy::new(|| {
...
 
+#[derive(Debug, PartialEq)]
+enum TestsToRun {
+    All,
+    Ignored,
+    NonIgnored,
+}
+
+impl TestsToRun {
+    fn from_flags(only_ignored: bool, include_ignored: bool) -> Self {
```
cptartur, line 64: What in case both `only_ignored` and `include_ignored` are true. Should we panic/return an error?

crates/forge/src/lib.rs
```diff
@@ -54,13 +53,41 @@ static BUILTINS: Lazy<Vec<&str>> = Lazy::new(|| {
...
+    }
+}
+
+fn should_be_run(tests_to_run: &TestsToRun, case_ignored: bool) -> bool {
+    match tests_to_run {
+        TestsToRun::All => true,
+        TestsToRun::Ignored => case_ignored,
+        TestsToRun::NonIgnored => !case_ignored,
+    }
+}
```
cptartur, line 81: What this function does is kind of unclear just looking at its implementation. Perhaps this should be defined on test case?

crates/forge/src/lib.rs
piotmag769: They cannot be, guaranteed by clap
```rust
    #[arg(long, conflicts_with = "only_ignored")]
    include_ignored: bool,
```

crates/forge/src/lib.rs
piotmag769: True, will change that

piotmag769: `TestCase` is declared in `test-collector`, so it's not possible, but I can pass a test case here as arg

crates/forge/src/lib.rs
cptartur: I understand this is configured in clap, but I think having an extra assert here could prevent some weird bugs in case logic elsewhere changes.

crates/forge/src/lib.rs
cptartur: That could be better I think

crates/forge/src/lib.rs
piotmag769: Can add `assert` then

cptartur (approved):

piotmag769:
Closing in favour of #886

PR #804, opened 2023-10-04
Fixed declare

### LINKED ISSUE #786: Make declare test work
https://github.com/foundry-rs/starknet-foundry/blob/4a5b36e4d755eec31ba886ef61c485f3a25726dd/crates/cheatnet/tests/cheatcodes/declare.rs#L40

We shouldn't execute it as a transaction from now

Radinyn:
Closes #786

## Introduced changes

- Declare now works the same way as before, by setting the compiled class hash  value to 0.


crates/cheatnet/src/cheatcodes/declare.rs
```diff
@@ -39,45 +40,31 @@ impl BlockifierState<'_> {
...
-        //         Err(TransactionExecutionError::DeclareTransactionError { class_hash })
-        //     }
-        // }
-
-        Ok(class_hash)
+                // NOTE: Compiled class hash is being set to 0 here
+                blockifier_state
+                    .set_compiled_class_hash(class_hash, Default::default())
+                    .unwrap_or_else(|err| panic!("Failed to set compiled class hash: {err:?}"));
+                Ok(class_hash)
```
piotmag769, line 55: Cannot we compute real compiled class hash from CASM using blockifier's api?

war-in: Do we need this for anything?

war-in (approved):

crates/cheatnet/src/cheatcodes/declare.rs
```diff
@@ -39,45 +40,31 @@ impl BlockifierState<'_> {
...
-        //     Err(error) => Err(error).map_err(TransactionExecutionError::from),
-        //     Ok(_) => {
-        //         // Class is already declared, cannot redeclare
-        //         // (i.e., make sure the leaf is uninitialized).
-        //         Err(TransactionExecutionError::DeclareTransactionError { class_hash })
-        //     }
-        // }
-
-        Ok(class_hash)
+                // NOTE: Compiled class hash is being set to 0 here
```
Arcticae, line 51: Please comment "why" - instead of "what"

crates/cheatnet/src/cheatcodes/declare.rs
```diff
@@ -39,45 +40,31 @@ impl BlockifierState<'_> {
...
+            Err(error) => Err(CheatcodeError::Unrecoverable(EnhancedHintError::State(
+                error,
+            ))),
+            Ok(_) => {
+                // Class is already declared, cannot redeclare
+                // (i.e., make sure the leaf is uninitialized).
+                Err(CheatcodeError::Unrecoverable(EnhancedHintError::Anyhow(
+                    anyhow!("Class hash {} is already declared", class_hash),
+                )))
+            }
```
Arcticae, line 66: Are those errors tested? I think intuitively, errors should be different now than the ones from the txn flow, so perhaps we should test more thoroughly the handling

Arcticae (changes requested):

crates/cheatnet/src/cheatcodes/declare.rs
Radinyn: The "already declared" error is tested

crates/cheatnet/src/cheatcodes/declare.rs
piotmag769: Since we communicate that we are consistent with starknet - it would be nice

Arcticae (approved):

PR #808, opened 2023-10-04
Events emitted in a test can be spied

### LINKED ISSUE #612: Cheatcodes should work in test state
If we target address `0`, cheatcode should target test state

abulenok:
Closes #612

## Introduced changes

- spy events cheatcode works in the test state
-

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/cheatnet/src/execution/syscalls.rs
```diff
@@ -297,6 +298,17 @@ impl CheatableSyscallHandler<'_> {
             // Increment, since the selector was peeked into before
             self.syscall_handler.syscall_ptr += 1;
             return self.execute_syscall(vm, deploy_syscall, constants::DEPLOY_GAS_COST);
+        } else if SyscallSelector::EmitEvent == selector {
```
war-in, line 301: Should we increment the pointer here too?
We do it in every other case

crates/cheatnet/src/execution/syscalls.rs
```diff
@@ -297,6 +298,17 @@ impl CheatableSyscallHandler<'_> {
...
+            let result = self.syscall_handler.execute_next_syscall(vm, hint);
+            let event = Event::from_ordered_event(
+                self.syscall_handler.events.last().unwrap(),
```
war-in, line 304: Are we sure there will be only one event? (Or if we want to catch only the last one)

crates/cheatnet/src/execution/events.rs
war-in, line 1: Previously we collected only those events which were emitted from spied addresses. I think we lack this feature now

Arcticae: Marcin is right, even if we filter the events out, when we create a spy after some events will be emitted, the spy will catch also the ones that were emitted before that.
(also, we should add a test case)

crates/cheatnet/src/execution/syscalls.rs
Arcticae: It's a good thing to check, but i doubt that they are batched

Arcticae (changes requested):
Blocking until Marcin's comments are addressed

crates/cheatnet/src/execution/syscalls.rs
abulenok: We should not since `self.syscall_handler.execute_next_syscall` reads the selector and increments `syscall_ptr` by itself.
Moreover, adding `syscall_ptr` incrementation makes cheatnet spy_events tests fail.

crates/cheatnet/src/execution/events.rs
abulenok: Good point! fixed it and added a test case

crates/cheatnet/src/execution/syscalls.rs
abulenok: Added a comment to the code with clarification

crates/cheatnet/src/execution/syscalls.rs
abulenok: Added an assert

war-in (approved):

Arcticae (approved):

PR #814, opened 2023-10-05
Drop v1 tests, use only one scarb version

THenry14:
Closes https://github.com/foundry-rs/starknet-foundry/issues/813

## Introduced changes

- drop v1 tests, simplify setup, do not require asdf for scarb version management

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


CHANGELOG.md
```diff
@@ -19,6 +19,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Spying events interface is updated to enable the use of events defined inside contracts in assertions
 
+### Cast
+
+#### Changed
+
+- dropped official support for cairo 1 contracts. while they still should be working without any problems, from now on 
+the only officially supported cairo version is 2
```
war-in, line 27: I would rephrase it because I'm not sure that we are now on Cairo 2 😅 It would be better to call it `cairo compiler version 2` or something like that

crates/cast/tests/e2e/account/add.rs
```diff
@@ -59,10 +59,14 @@ pub async fn test_happy_case() {
...
-        CONTRACTS_DIR.to_string() + "/v1/map",
+        CONTRACTS_DIR.to_string() + "/map",
         "put",
-        "3",
+        "30",
     ));
+    //     let tmp_dir = TempDir::new()?;
+    //     tmp_dir
+    //         .copy_from(&tests_folder_path, &["**/*.cairo"])
+    //         .context("Unable to copy files to temporary directory")?;
```
war-in, line 69: 👀

crates/cast/tests/e2e/account/create.rs
```diff
@@ -91,9 +91,9 @@ pub async fn test_happy_case_generate_salt() {
 #[tokio::test]
 pub async fn test_happy_case_add_profile() {
     let current_dir = Utf8PathBuf::from(duplicate_directory_with_salt(
-        CONTRACTS_DIR.to_string() + "/v1/balance",
+        CONTRACTS_DIR.to_string() + "/map",
         "put",
-        "1",
+        "10",
```
war-in, line 96: Hmm, why is it changed? 🤔

crates/cast/tests/utils/build_contracts.sh
```diff
@@ -1,31 +1,19 @@
...
-        pushd "$contract_dir"
-        scarb build
-        popd
-      fi
-    done
-
-  else
-    echo "Please run tests/utils/prepare_for_tests.sh script first!"
+for contract_dir in "$CONTRACTS_DIRECTORY"/*; do
+  if ! test -d "$contract_dir"/target && [[ "$contract_dir" != *"fails"* ]]; then
```
war-in, line 14: Now I wonder if we should exclude already built directories. If someone is running this script he usually want to build them

THenry14: This script is being called by our test suite if I remember correctly, not sure compiling these contracts each time we want to run test is preferable (especially so that these contracts generally don't change)

crates/cast/tests/e2e/account/create.rs
THenry14: two weeks passed, and I don't remember 😅 my guess is a few more tests used this account and something failed because of too few funds, but not sure. If you want, I can try anc check it, but I don't think this is that relevant tbh :+1:

crates/cast/tests/e2e/account/add.rs
THenry14: sorry

war-in (approved):

PR #827, opened 2023-10-05
Contract verify

### LINKED ISSUE #830: Contract verification command in `cast` tool

Cast

### Feature Request

Proposed design : https://github.com/foundry-rs/starknet-foundry/blob/c1a3324c35aa68593e064ef5cb816928eb7c487f/design_documents/contract_verification.md

rjnrohit:
Closes #830 

## Introduced changes

This pull request proposes the design for the contract verification through sncast command

## Checklist

- [x] Linked relevant issue
- [ ] Added relevant tests


design_documents/contract_verification.md
```diff
@@ -0,0 +1,118 @@
...
+
+#### `--verifier-url <URL>`
+
+Optional.
+Specifies the Blockchain Explorer's Verification API URL ().  
+Default: the default API URL for selected verifier (eg. https://goerli.voyager.online/) (???)
+
+Options are: 
+ - `https://voyager.online/` - for verification at mainnet
+ - `https://goerli.voyager.online/` -  for verification on testnet
```
THenry14, line 63: I suppose urls won't generally change, right? if so, maybe we should take a `--network` parameter instead? That would be easier UX-wise, as users wouldn't need to know the urls - they could just provide the network they want to verify against. wdyt?

THenry14 (commented):
looking good! 🔥

design_documents/contract_verification.md
```diff
@@ -0,0 +1,118 @@
...
+                """
+                #[starknet::interface]
+                ...
+                #[starknet::contract]
+                mod Balance {
+                    ...
+                }
+                """
+            },
+            "src/forty_two.cairo" : {
```
THenry14, line 100: how do you plan to determine which files will be sent in the request? just include all files from `src`?

rjnrohit: Yes, right now we're thinking to include all files that comes under a Scarb workspace

design_documents/contract_verification.md
rjnrohit: yes, we must be able to provide/override the API URL. While we can have a "network" parameter, and assume hardcoded Voyager API endpoints for different networks, we must also give the ability to provide own URL

THenry14: sure, let's do that then. 

In that case, I think this shouldn't be limited to only those two urls, but any url right?

rjnrohit: yes

cptartur (changes requested):

design_documents/contract_verification.md
```diff
@@ -0,0 +1,118 @@
...
+Default: the default API URL for selected verifier (eg. https://goerli.voyager.online/) (???)
+
+Options are: 
+ - `https://voyager.online/` - for verification at mainnet
+ - `https://goerli.voyager.online/` -  for verification on testnet
+
+### Voyager API plugin
+
+A sample request to the voyager API end point will look as follows: 
+```rust
```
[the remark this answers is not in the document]
THenry14: I'm curious if that would be possible, but that would indeed be cool

design_documents/contract_verification.md
```diff
@@ -0,0 +1,118 @@
...
+- select the verifiation logic to use (based on `--verifier` parameter)
+- pick the selected scarb workspace source code from local filesystem (based on `Scarb.toml` file)
+- for the selected verifier it will upload the Scarb workspace contracts source code to the verifier's API
+- call the verifier's API to trigger verification of the uploaded source code
+- wait for results of the verification API call
+- respond to the user with the results 
+
+#### Parameters
+
+#### `--path-to-scarb-toml, -s <PATH>`
```
Arcticae, line 29: I think this is a top-level argument, so this will not be specific to this command, i don't think we should mention it here

Arcticae: (might be confusing to the person who implements)

design_documents/contract_verification.md
```diff
@@ -0,0 +1,118 @@
...
+
+#### `--verifier <VERIFIER NAME>`
+
+Optional.
+Specifies the Blockchan Explorer to verify with.  
+Default: `voyager`
+
+Options are: 
+ - voyager
+ - starkscan
```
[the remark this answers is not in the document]
Arcticae: I think this should be either one of those, or a URL, instead of 2 separate arguments.
When we want to go for 'goerli' we could provide a predefined "voyager-goerli" option.
This way if something changes, users are left with one simple change in the command.

design_documents/contract_verification.md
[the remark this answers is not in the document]
Arcticae: It would be better on our end, maintaining would be easier that way.

design_documents/contract_verification.md
```diff
@@ -0,0 +1,118 @@
...
+#### `--path-to-scarb-toml, -s <PATH>`
+
+Optional.
+Path to `Scarb.toml` file.
+If supplied, cast will not look for `Scarb.toml`` file in current (or parent) directory, but will use this path instead.
+
+#### `--contract-name`
+
+Required.
+Name of the contract to be submitted for verification.
```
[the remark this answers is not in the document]
rjnrohit: yes

design_documents/contract_verification.md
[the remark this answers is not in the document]
rjnrohit: let's go with the `--network` argument for now! I agree with you @Arcticae

design_documents/contract_verification.md
rjnrohit: Good catch! thanks @Arcticae removing it!

design_documents/contract_verification.md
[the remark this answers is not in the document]
rjnrohit: okay!

design_documents/contract_verification.md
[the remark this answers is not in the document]
rjnrohit: will work on that what type of responses the sncast will `expect`, will update the same in this PR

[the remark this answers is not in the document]
stranger80: @cptartur @THenry14 we are assuming we have no influence on the APIs provided by any of the block explorers - ie. Voyager verification APIs are there already and are different form StarkScan APIs, etc. Hence - the idea of "adapters" in `sncast`, that encapsulate the nuances of each specific API and expose a generic interface (trait?) to the `verify` command implementation. And for this iteration if the feature, we would be implementing Voyager adapter only. Does that make sense?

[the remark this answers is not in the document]
cptartur: Perhaps enforcing some api standard is a bit overkill and we can go with "adapters" approach. Preferably we'd still have some minimal interface that explorers should implement if they want to integrate with `sncast`. Maybe it can be just at a trait level and explorers should be allowed to send whatever request format they want as long as they can then return it as whatever rust structure we expect. 

My concern is just if we have multiple explorers supported I'd like the outputs presented to the user in the CLI to be exactly the same, no matter which one is used.

I'm understand that initially the response will basically be `verified/not verified` but I'm assuming this might get expanded in the future?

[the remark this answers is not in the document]
stranger80: Added `### Contract Verification interface` section to propose a generic request/response format. I hope this makes the idea more clear?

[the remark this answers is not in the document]
cptartur: It does, thanks. Resolving this discussion, will add comments to the updated section if I have any.

design_documents/contract_verification.md
```diff
@@ -0,0 +1,131 @@
...
+#### `--contract-address`
+
+Required.
+Address of the contract to be submitted for verification.
+
+#### `--verifier <VERIFIER NAME>`
+
+Optional.
+Specifies the Blockchan Explorer to verify with.  
+Default: `voyager`
```
cptartur, line 43: I'd prefer we don't have a default defined in the document itself

design_documents/contract_verification.md
```diff
@@ -0,0 +1,131 @@
...
+
+Options as of writing of this document: 
+ - voyager
+ - starkscan
+
+#### `--network <NETWORK_NAME>`
+
+Optional.
+Specifies the network on which Blockchain Explorers will do the verification
+Default: the default network on which verfication would be done is `goerli` testnet
```
cptartur, line 53: @THenry14 does `sncast` have default networks anywhere else? If so, are they goerli or mainnet?

design_documents/contract_verification.md
```diff
@@ -0,0 +1,131 @@
...
+
+### Contract Verification interface
+
+To implement contract verification for a specific explorer, it is required to implement a generic interface (request/response), as described below. The data structures proposed can be eventually extended to cater for detailed requirements of subsequent explorer adapters.
+
+#### Request
+
+- `ContractAddress` - `ContractAddress`
+- `ContractName` - `String`
+- `SourceCode` - collection of file records:
```
cptartur, line 67: Just to be clear, we want to include what kind of files. `.cairo` + the `Scarb.toml`? I think this could be defined here.

cptartur (commented):
Looks mostly good 👍 Left some minor comments

design_documents/contract_verification.md
rjnrohit: So you want to make this argument a required one ?

cptartur: I think it might be good for the user to explicitly select the block explorer. cc @Arcticae @THenry14 what do you think?

design_documents/contract_verification.md
cptartur: I've verified manually and we aren't defaulting to any network in other places. I guess `goerli` is fine as default. We should make sure that docs and/or output of the implemented command makes it clear what network is being used and how to change it.

cptartur (approved):
Looks good to me 👍 
Not resolving my comments until other reviewers have a chance to look at them

design_documents/contract_verification.md
Arcticae: Yes i think that is reasonable, since there might be other available ones in the future and we don't want to be biased :)

design_documents/contract_verification.md
Arcticae: I think explicitly providing the network is preferable

cptartur: Maybe let's just remove the default here then

rjnrohit:
Now, we can merge this PR, right?

PR #849, opened 2023-10-11
Do not print ascii control characters in `snforge_std::PrintTrait`

### LINKED ISSUE #844: Don't print "weird" ascii characters
We shouldn't print invisible characters like `Esc` or `DC1`. It results in a part of input not being printed sometimes

cptartur:
Closes #844 

## Introduced changes

- Changed printing logic so we do not print shortstrings containing ascii control characters

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


docs/src/testing/debugging.md
```diff
@@ -13,7 +13,7 @@ Starknet Foundry standard library adds a utility method for printing inside test
 
 ## In tests
 
-Here's a test with example use of print method:
+Here's a test with example use of [`print`](../appendix/forge-library/print.md) method:
 
 ```rust
 // Make sure to import Starknet Foundry PrintTrait
```
piotmag769, line 19: I think the import below is incorrect, rn it should be `use snforge_std::io::PrintTrait` I believe

piotmag769 (approved):

docs/src/testing/debugging.md
cptartur: PrintTrait is reexported in lib.cairo so it should be fine. We show it like this in couple other places.

PR #869, opened 2023-10-12
Fix not deteministic tests

tkumor3:
Closes #

## Introduced changes

-
-

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


piotmag769 (approved):

crates/forge/src/lib.rs
```diff
@@ -603,6 +605,14 @@ fn run_with_fuzzing(
...
+            .iter()
+            .find(|item| matches!(item, TestCaseSummary::Failed { .. }))
+        {
+            let result = result.clone().with_runs(runs);
+            return Ok(result);
+        }
+
         if let Some(result) = results.iter().find(|item| {
             matches!(
                 item,
```
piotmag769, line 618: Just realised: we shouldn't match for `Interrupted` below - if we ever return interrupted from this function it is a bug (it should return Failed instead)

piotmag769 (changes requested):

crates/forge/src/lib.rs
```diff
@@ -612,14 +622,6 @@ fn run_with_fuzzing(
...
-            .iter()
-            .find(|item| matches!(item, TestCaseSummary::Failed { .. }))
-        {
-            let result = result.clone().with_runs(runs);
-            return Ok(result);
-        }
-
         let result: TestCaseSummary = results
             .last()
             .expect("Test should always run at least once")
```
piotmag769, line 627: We could assert that this result is `Passed` too, other cases are covered above

crates/forge/src/lib.rs
```diff
@@ -582,7 +582,9 @@ fn run_with_fuzzing(
             results.push(result.clone());
 
             match &result {
-                TestCaseSummary::Failed { .. } | TestCaseSummary::Interrupted {} => {
+                TestCaseSummary::Failed { .. }
+                | TestCaseSummary::Interrupted {}
+                | TestCaseSummary::Skipped { .. } => {
```
piotmag769, line 587: We should match for `SkippedFuzzing` too

crates/forge/src/lib.rs
```diff
@@ -603,6 +605,14 @@ fn run_with_fuzzing(
...
         )?;
 
+        if let Some(result) = results
+            .iter()
+            .find(|item| matches!(item, TestCaseSummary::Failed { .. }))
+        {
+            let result = result.clone().with_runs(runs);
+            return Ok(result);
+        }
+
```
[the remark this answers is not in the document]
tkumor3: it's possible, that a failing test will execute cancelationToken.cancel(), and before this task is ended other fuzzing task will return as skip! So if we do not change the priority of FAIL/SKIP it's possible that someone will get only skipped tests

crates/forge/src/main.rs
```diff
@@ -223,6 +223,7 @@ fn main_execution() -> Result<bool> {
             Ok(all_failed_tests)
         })
     })??;
+    rt.shutdown_background();
```
piotmag769, line 226: Why do we need that? Don't we wait for all tasks to shutdown anyways?

tkumor3: good catch, It has to be added during merge PRs :o

piotmag769 (approved):

crates/forge/src/lib.rs
```diff
@@ -587,7 +587,7 @@ fn run_with_fuzzing(
             results.push(result.clone());
 
             match &result {
-                TestCaseSummary::Failed { .. } | TestCaseSummary::Interrupted {} => {
+                TestCaseSummary::Failed { .. } | TestCaseSummary::InterruptedByError {} => {
```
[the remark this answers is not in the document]
tkumor3: about skip, we can't do that because https://github.com/foundry-rs/starknet-foundry/pull/869#discussion_r1362030442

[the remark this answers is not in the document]
tkumor3: some_dummy_value -> it should be Failed ich we do not get results from test we have to tread it as fails

crates/forge/src/lib.rs
```diff
@@ -581,13 +581,19 @@ fn run_with_fuzzing(
         }
 
         let mut results = vec![];
+        let mut result = TestCaseSummary::Failed {
+            name: case.name.to_string(),
+            msg: None,
+            arguments: vec![],
+            fuzzing_statistic: None,
+        };
```
cptartur, line 589: Is it possible at all for this default to be returned? Because if yes, we will sometimes show quite weird outputs from the fuzzer, where the runs will be non 0, but arguments will be empty. Perhaps `Skipped` is a safer default?

tkumor3: I have no idea, how to trigger a situation where we will return default. IMO it will be weird if someone will create a fuzzing test without subtests 🤔  I assume that if this happened FAIL is a more accurate message than skip. 

IMO we should panic in this case, but no idea how to do that

cptartur (approved):

PR #874, opened 2023-10-12
change the way of killing cancelled tasks

tkumor3:
Closes #

## Introduced changes

-
-

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/forge/src/lib.rs
```diff
@@ -552,7 +559,6 @@ fn run_with_fuzzing(
         let mut tasks = FuturesUnordered::new();
         // Pattern in order to waiting for things to finish shutting down
         // https://tokio.rs/tokio/topics/shutdown
```
cptartur, line 561: This comments should be moved/copied as well.

crates/forge/src/lib.rs
```diff
@@ -441,6 +443,8 @@ async fn run_tests_from_crate(
         results.push(result);
     }
 
+    rec.close();
```
cptartur, line 446: Why do we close instead of dropping?

crates/forge/src/running.rs
```diff
@@ -70,25 +71,17 @@ fn build_hints_dict<'b>(
...
     tokio::task::spawn_blocking(move || {
-        run_test_case(
-            args,
-            &case,
-            &runner,
-            &runner_config,
-            &runner_params,
-            &sender,
-        )
+        run_test_case(args, &case, &runner, &runner_config, &runner_params, &send)
```
cptartur, line 83: Do we need to pass the sender down to the blocking task?

crates/forge/src/main.rs
```diff
@@ -192,6 +193,7 @@ fn main_execution() -> Result<bool> {
             Ok(all_failed_tests)
         })
     })??;
+    rt.shutdown_background();
```
cptartur, line 196: The docs say that this can cause resource leak because blocking tasks are not actually getting cancelled. https://docs.rs/tokio/latest/tokio/runtime/struct.Runtime.html#method.shutdown_background

cptartur (changes requested):

crates/forge/src/lib.rs
tkumor3: https://docs.rs/tokio/latest/tokio/sync/mpsc/struct.Receiver.html - I want to close a channel. Not remove single sender

crates/forge/src/running.rs
tkumor3: I can refactor it a little bit

crates/forge/src/main.rs
tkumor3: I can remove it! It makes only the feeling that our tests finished a little bit faster! This task which is alive only checks if the chanel exists

crates/forge/src/lib.rs
tkumor3: remove It use other pattern

crates/forge/src/running.rs
```diff
@@ -70,25 +71,20 @@ fn build_hints_dict<'b>(
...
     tokio::task::spawn_blocking(move || {
-        run_test_case(
-            args,
-            &case,
-            &runner,
-            &runner_config,
-            &runner_params,
-            &sender,
-        )
+        if send.is_closed() {
```
Arcticae, line 86: What if the channel is closed after we are past this section?

cptartur: I think we'll get an error like previously. Perhaps this should be approached the other way round - "main" task should wait for all blocking tasks to terminate? Channel or some other messaging could be used for that (or maybe a simpler solution).

tkumor3: in most cases, this test will be executed but user will not get the result/error. I think in extreme situations we will get >= x error like the previous, where x is the number of user cores.

tkumor3: @Arcticae @cptartur  https://github.com/foundry-rs/starknet-foundry/pull/874/commits/792532dd89e4eaa10bc6ba5c4d67cea9a7bc7a2b

I fixed this problem

cptartur: I think this should resolve the issue. We should add a comment somewhere in the code (probably in `run()`) explaining the logic of using two channels for this.

crates/forge/src/running.rs
```diff
@@ -75,25 +76,31 @@ fn build_hints_dict<'b>(
...
-) -> Result<TestCaseSummary> {
+    send: Sender<()>,
+    send_shut_down: Sender<()>,
+) -> JoinHandle<Result<TestCaseSummary>> {
     tokio::task::spawn_blocking(move || {
+        // Due to the inability of spawn_blocking to be abruptly cancelled,
+        // a channel is used to receive information indicating
+        // that the execution of the task is no longer necessary.
+        if send.is_closed() {
+            return Err(anyhow::anyhow!("stop spawn_blocking"));
```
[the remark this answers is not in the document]
tkumor3: No because this task has no parent

crates/forge/src/running.rs
tkumor3: 👍

cptartur (approved):

crates/forge/tests/integration/setup_fork.rs
```diff
@@ -94,6 +94,7 @@ fn fork_aliased_decorator() {
     ).as_str());
 
     let rt = Runtime::new().expect("Could not instantiate Runtime");
+
```
MaksymilianDemitraszek, line 97: Unrelated change

crates/forge/tests/integration/common/running_tests.rs
```diff
@@ -12,6 +12,7 @@ use tokio::runtime::Runtime;
 
 pub fn run_test_case(test: &TestCase) -> Vec<TestCrateSummary> {
     let rt = Runtime::new().expect("Could not instantiate Runtime");
+
```
MaksymilianDemitraszek, line 15: Unrelated change

crates/forge/src/lib.rs
```diff
@@ -364,7 +364,7 @@ pub async fn run(
         )
         .await?;
 
-        summaries.push(summary.clone());
+        summaries.push(summary);
```
MaksymilianDemitraszek, line 367: Unrelated change

PR #878, opened 2023-10-13
Catch network errors while forking

### LINKED ISSUE #791: Forking: ambigious errors when passing incorrect url or being offline
We do not inform that the connection couldn't be established - it is prob bcs in `impl StateReader for ExtendedStateReader` we always return `Ok` even on network errors (we just return `Default`).
Instead we get sth like:

```console
[FAIL] tests::test_forking::hash

Failure data:
    Got an exception while executing a hint:
    Contract not deployed at address: 0x071c8d74edc89330f314f3b1109059d68ebfa68874aa91e9c425a6378ffde00e
```

We should also add tests for cases like:
- [ ] network error
- [ ] invalid URL (parsable, but not pointing to any node)

war-in:
Closes #791

## Introduced changes

- default storage values are not returned when network errors occur

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests

### DIFF
--- a/crates/cheatnet/src/state.rs
+++ b/crates/cheatnet/src/state.rs
@@ -52,13 +52,7 @@ impl StateReader for ExtendedStateReader {
                 self.fork_state_reader
                     .as_mut()
                     .map_or(Ok(StarkFelt::default()), |reader| {
-                        match reader.get_storage_at(contract_address, key) {
-                            Ok(value) => Ok(value),
-                            Err(StateError::StateReadError(msg)) if msg.contains("node") => {
-                                Err(StateError::StateReadError(msg))
-                            }
-                            _ => Ok(Default::default()),
-                        }
+                        match_node_response(reader.get_storage_at(contract_address, key))
                     })
             })
     }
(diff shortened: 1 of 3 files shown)


Arcticae (approved):

crates/cheatnet/src/state.rs
```diff
@@ -222,3 +216,13 @@ impl CheatnetState {
...
     }
 }
+
+fn match_node_response<T: Default>(result: StateResult<T>) -> StateResult<T> {
+    match result {
+        Ok(class_hash) => Ok(class_hash),
+        Err(StateError::StateReadError(msg)) if msg.contains("node") => {
+            Err(StateError::StateReadError(msg))
+        }
+        _ => Ok(Default::default()),
```
piotmag769, line 226: Why do we return default here?

war-in: Because we have to

war-in: Imagine a situation when someone wants to deploy a new contract when forking.
The state is asked this way:
- dict_state_reader (which does not have a value)
- fork_state_reader:
  - node is fetched and if there is no value, an error is returned
  - we have to return default because blockifier handles the state this way

Issue #887, opened 2023-10-15
Timeout when running Satoru tests


akhercha:
### Which component is your bug related to?

Forge

### Foundry Version

0.8.2

### What operating system are you using?

Linux

### What system architecture are you using?

x86

### What happened

When launching unit tests using `snforge`, it works locally on my machine but:
- it does not work for someone else on a computer with less performance, it timeouts,
- it does not work for [Satoru's CI](https://github.com/keep-starknet-strange/satoru/actions/runs/6523211064/job/17713611608), we also have a timeout here.

For a quick fix solution I tried adding a `--cores` parameter which sets the number of threads for the tests. Setting it to 1 should not run the tests in parallel & work.

### Trace

_(... see [CI linked](https://github.com/keep-starknet-strange/satoru/actions/runs/6523211064/job/17713611608) in issue)_

```rust
[PASS] tests::referral::test_referral_utils::given_normal_conditions_when_get_referral_info_then_works
[PASS] tests::referral::test_referral_utils::given_refferal_discountshare_when_get_referral_info_then_works
[PASS] tests::referral::test_referral_utils::given_normal_conditions_when_claim_affiliate_reward_then_works
Error: The operation was canceled.
```

### Is there an existing issue for this?

- [x] I have searched the existing issues and verified no issue exits for this problem.

PR #888, opened 2023-10-15
feat: Added --cores argument to snforge

akhercha:
Related: #887

(would not really _close_ the issue, it's more a temporary solution since it does not work with parallel tests for now)

## Introduced changes

- Added a `--cores` argument to the `snforge` command, allows to choose the number of threads for the test executions.

## Comments

If you suggestions for a better name etc let me know!

## Checklist

- [X] Linked relevant issue
- [X] Added relevant tests

### DIFF
--- a/crates/forge/src/main.rs
+++ b/crates/forge/src/main.rs
@@ -139,7 +139,7 @@ fn main_execution() -> Result<bool> {
     if args.clean_cache {
         clean_cache(&workspace_root).context("Failed to clean snforge cache")?;
     }
-    let cores_approx = if let Some(cores) = args.cores {
+    let cores = if let Some(cores) = args.cores {
         cores
     } else {
         match available_parallelism() {
@@ -151,7 +151,7 @@ fn main_execution() -> Result<bool> {
         }
     };
     let rt = Builder::new_multi_thread()
-        .max_blocking_threads(cores_approx)
+        .max_blocking_threads(cores)
         .enable_all()
         .build()?;
     let all_failed_tests = rt.block_on({
--- a/crates/forge/src/main.rs
+++ b/crates/forge/src/main.rs
@@ -71,7 +71,7 @@ fn validate_cores_number(val: &str) -> Result<usize> {
     let cores_approx = available_parallelism()?.get();
     if parsed_val > cores_approx {
         bail!(
-            "Number of cores must be less than or equal to the number of cores available on the machine ({})",
+            "Number of cores must be less than or equal to the number of cores available on the machine = {}",
             cores_approx
         );
     }
--- a/crates/forge/src/main.rs
+++ b/crates/forge/src/main.rs
@@ -55,6 +55,27 @@ struct Args {
     /// Clean forge cache directory
     #[arg(short, long)]
     clean_cache: bool,
+
+    /// Cores used for parallel tests execution
+    #[arg(short = 't', long, value_parser = validate_cores_number)]
+    cores: Option<usize>,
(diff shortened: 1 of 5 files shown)


crates/forge/tests/e2e/cores.rs
```diff
@@ -0,0 +1,112 @@
...
+        .arg("--cores")
+        .arg("99999")
+        .assert()
+        .code(2);
+
+    let stderr =
+        String::from_utf8(assert.get_output().stderr.clone()).expect("stderr is not valid UTF-8");
+
+    assert!(stderr.contains("error: invalid value '99999' for '--cores <CORES>': Number of cores must be less than or equal to the number of cores available on the machine"));
+}
```
piotmag769, line 112: You can use `stderr_matches` with `[..]` pattern to match unknown part of the stderr here

piotmag769 (approved):
LGTM, thanks a lot for the contribution!

crates/forge/tests/e2e/cores.rs
akhercha: Updated!

piotmag769 (approved):

crates/forge/src/main.rs
```diff
@@ -55,6 +55,27 @@ struct Args {
     /// Clean forge cache directory
     #[arg(short, long)]
     clean_cache: bool,
+
+    /// Cores used for parallel tests execution
```
cptartur, line 59:
```suggestion
    /// Number of cores used for test execution
```

crates/forge/src/main.rs
```diff
@@ -55,6 +55,27 @@ struct Args {
...
+    let parsed_val: usize = val
+        .parse()
+        .map_err(|_| anyhow::anyhow!("Failed to parse '{}' as usize", val))?;
+    if parsed_val == 0 {
+        bail!("Number of cores must be greater than 0");
+    }
+    let cores_approx = available_parallelism()?.get();
+    if parsed_val > cores_approx {
+        bail!(
+            "Number of cores must be less than or equal to the number of cores available on the machine ({})",
```
cptartur, line 74:
```suggestion
            "Number of cores must be less than or equal to the number of cores available on the machine = {}",
```
We usually format these error like this

crates/forge/src/main.rs
```diff
@@ -118,9 +139,18 @@ fn main_execution() -> Result<bool> {
...
-    let cores_approx = available_parallelism()?.get();
+
+    let cores = if let Some(cores) = args.cores {
+        cores
+    } else if let Ok(available_cores) = available_parallelism() {
+        available_cores.get()
+    } else {
+        eprintln!("Failed to get the number of available cores, defaulting to 1");
+        1
+    };
```
cptartur, line 150: I'm not sure if we should be defaulting to 1 at all. Previous approach where we bubbled an error seemed reasonable.

cptartur: I wonder in which cases `available_parallelism` can return an error at all. I guess there is some benefit of having a default.

docs/src/appendix/forge.md
```diff
@@ -41,6 +41,10 @@ Seed for the fuzzer.
 
 Clean forge cache directory.
 
+## `--cores`, `<NUM_CORES>`
+
+Number of cores to use for parallel execution.
```
cptartur, line 46:
```suggestion
Number of cores to use for test execution.
```

cptartur (approved):

PR #893, opened 2023-10-16
Return correct timestamp in a forked block

### LINKED ISSUE #748: Return correct block info while forking

Forge

### Foundry Version

0.7.1

MacOS

arm

### What happened

When a state is forked in a test, `get_block_timestamp` returns 0 even if block_id is set to the `Latest`.
We should improve `get_cheated_exec_info_ptr` in the `execution_info.rs` to use block info from the fork

### Trace

_No response_

- [x] I have searched the existing issues and verified no issue exits for this problem.

abulenok:
Closes #748

## Introduced changes

- Apply forked block info to the execution context
-

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests

### DIFF
--- a/crates/cheatnet/src/forking/cache.rs
+++ b/crates/cheatnet/src/forking/cache.rs
@@ -54,17 +54,13 @@ impl ForkCacheContent {
                     .insert(other_contract_address.clone(), other_storage.clone());
             }
         }
-        // nonce_at
+
         self.nonce_at.extend(other.nonce_at.clone());
-        // class_hash_at
         self.class_hash_at.extend(other.class_hash_at.clone());
-        // compiled_contract_class
         self.compiled_contract_class
             .extend(other.compiled_contract_class.clone());
-        // compiled_class_hash
         self.compiled_class_hash
             .extend(other.compiled_class_hash.clone());
-        // block_info
         if other.block_info.is_some() {
             self.block_info = other.block_info;
         }
--- a/crates/cheatnet/src/forking/cache.rs
+++ b/crates/cheatnet/src/forking/cache.rs
@@ -1,3 +1,4 @@
+use crate::state::CheatnetBlockInfo;
 use camino::Utf8PathBuf;
 use conversions::StarknetConversions;
 use fs2::FileExt;
@@ -21,6 +22,7 @@ struct ForkCacheContent {
     class_hash_at: HashMap<String, String>,
     compiled_contract_class: HashMap<String, String>,
     compiled_class_hash: HashMap<String, String>,
+    block_info: Option<CheatnetBlockInfo>,
 }
 
 impl ForkCacheContent {
@@ -32,6 +34,7 @@ impl ForkCacheContent {
             class_hash_at: HashMap::new(),
             compiled_contract_class: HashMap::new(),
             compiled_class_hash: HashMap::new(),
+            block_info: None,
         }
     }
     fn from_str(serialized: &str) -> Self {
@@ -61,6 +64,10 @@ impl ForkCacheContent {
         // compiled_class_hash
         self.compiled_class_hash
             .extend(other.compiled_class_hash.clone());
+        // block_info
+        if other.block_info.is_some() {
+            self.block_info = other.block_info;
+        }
     }
 }
 
@@ -265,6 +272,14 @@ impl ForkCache {
         self.fork_cache_content
             .compiled_contract_class
             .insert(class_hash_str, contract_class_str);
+    }
+
+    pub(crate) fn get_block_info(&self) -> Option<CheatnetBlockInfo> {
+        self.fork_cache_content.block_info
+    }
+
+    pub(crate) fn cache_get_block_info(&mut self, block_info: CheatnetBlockInfo) {
+        self.fork_cache_content.block_info = Some(block_info);
     }
 }
 
--- a/crates/cheatnet/src/forking/state.rs
+++ b/crates/cheatnet/src/forking/state.rs
@@ -55,7 +55,7 @@ impl ForkStateReader {
 
 fn get_pending_block_parent(
     state_reader: &ForkStateReader,
-    pending_block: PendingBlockWithTxHashes,
+    pending_block: &PendingBlockWithTxHashes,
 ) -> StateResult<CheatnetBlockInfo> {
     let parent_block_id = BlockId::Hash(pending_block.parent_hash);
 
@@ -100,7 +100,7 @@ impl BlockInfoReader for ForkStateReader {
                 Ok(block_info)
             }
             Ok(MaybePendingBlockWithTxHashes::PendingBlock(pending_block)) => {
-                get_pending_block_parent(self, pending_block)
+                get_pending_block_parent(self, &pending_block)
             }
             Err(err) => Err(StateReadError(format!(
                 "Unable to get block with tx hashes from fork, err: {err:?}"
--- a/crates/cheatnet/src/forking/state.rs
+++ b/crates/cheatnet/src/forking/state.rs
@@ -67,9 +67,9 @@ impl BlockInfoReader for ForkStateReader {
                 "Block with id {:?} is pending",
                 self.block_id
             ))),
-            Err(_) => Err(StateReadError(format!(
-                "Unable to get block with tx hashes form fork"
-            ))),
+            Err(_) => Err(StateReadError(
+                "Unable to get block with tx hashes form fork".to_string(),
+            )),
         }
     }
 }
--- a/crates/cheatnet/src/forking/state.rs
+++ b/crates/cheatnet/src/forking/state.rs
@@ -73,9 +73,30 @@ impl BlockInfoReader for ForkStateReader {
 
                 Ok(block_info)
             }
-            Ok(MaybePendingBlockWithTxHashes::PendingBlock(_)) => Err(StateReadError(
-                "Unable to get block info of the pending block".to_string(),
-            )),
+            Ok(MaybePendingBlockWithTxHashes::PendingBlock(pending_block)) => {
+                let parent_block_id = BlockId::Hash(pending_block.parent_hash);
+
+                match self
+                    .runtime
+                    .block_on(self.client.get_block_with_tx_hashes(parent_block_id))
+                {
+                    Ok(MaybePendingBlockWithTxHashes::Block(parent_block)) => {
+                        Ok(CheatnetBlockInfo {
+                            block_number: BlockNumber(parent_block.block_number + 1),
+                            timestamp: BlockTimestamp(pending_block.timestamp),
+                            sequencer_address: ContractAddress(patricia_key!(
+                                pending_block.sequencer_address
+                            )),
+                        })
+                    }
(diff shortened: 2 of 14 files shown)


crates/forge/tests/integration/setup_fork.rs
```diff
@@ -160,3 +162,67 @@ fn fork_cairo0_contract() {
...
+
+            #[test]
+            #[fork(url: "{}", block_id: BlockId::Number(315057))]
+            fn test_fork_get_block_timestamp_contract_on_testnet() {{
+                let dispatcher = IERC20CamelDispatcher {{
+                    contract_address: contract_address_const::<1825832089891106126806210124294467331434544162488231781791271899226056323189>()
+                }};
+
+                let timestamp = dispatcher.return_block_timestamp();
+                assert(timestamp != 0, timestamp);
```
war-in, line 197: can we assert a real timestamp?

war-in:
What about the other information? Does `block_number` work correctly?

crates/cheatnet/src/forking/state.rs
```diff
@@ -45,6 +52,28 @@ impl ForkStateReader {
     }
 }
 
+impl BlockInfoReader for ForkStateReader {
+    fn get_block_info(&self) -> StateResult<CheatnetBlockInfo> {
+        match self
```
Arcticae, line 57: This will fetch every time that we request the block info, instead we should cache that with the `block_id` as a cache key

Arcticae (changes requested):

crates/cheatnet/src/state.rs
```diff
@@ -27,11 +32,42 @@ pub struct ExtendedStateReader {
...
 
+#[derive(Copy, Clone)]
+pub struct CheatnetBlockInfo {
+    pub block_number: BlockNumber,
+    pub timestamp: BlockTimestamp,
+    pub sequencer_address: ContractAddress,
+}
+
+impl Default for CheatnetBlockInfo {
+    fn default() -> Self {
```
piotmag769, line 62: Why do we need that?

crates/forge/tests/integration/setup_fork.rs
```diff
@@ -160,3 +162,67 @@ fn fork_cairo0_contract() {
...
 }
+
+#[test]
+fn get_block_timestamp_in_forked_block() {
+    let test = test_case!(formatdoc!(
+        r#"
+            use result::ResultTrait;
+            use array::ArrayTrait;
+            use option::OptionTrait;
+            use traits::TryInto;
```
piotmag769, line 173: No need to import these

crates/forge/tests/integration/setup_fork.rs
```diff
@@ -160,3 +162,67 @@ fn fork_cairo0_contract() {
 
     assert_passed!(result);
 }
+
+#[test]
+fn get_block_timestamp_in_forked_block() {
```
piotmag769, line 167: Can we add a test verifying the error returned when using `Pending` block and trying to get block info? Can use `assert_case_output_contains`

One test with `Latest` would be nice too

crates/forge/tests/integration/setup_fork.rs
piotmag769: Shouldn't we also verify the whole block info instead of just the timestamp?

crates/cheatnet/src/forking/state.rs
```diff
@@ -45,6 +52,28 @@ impl ForkStateReader {
...
+                block_number: BlockNumber(block.block_number),
+                timestamp: BlockTimestamp(block.timestamp),
+                sequencer_address: ContractAddress(patricia_key!(block.sequencer_address)),
+            }),
+            Ok(MaybePendingBlockWithTxHashes::PendingBlock(_)) => Err(StateReadError(format!(
+                "Block with id {:?} is pending",
+                self.block_id
+            ))),
+            Err(_) => Err(StateReadError(
+                "Unable to get block with tx hashes form fork".to_string(),
```
piotmag769, line 71: typo

crates/cheatnet/src/state.rs
```diff
@@ -27,11 +32,42 @@ pub struct ExtendedStateReader {
...
 }
 
+pub trait BlockInfoReader {
+    fn get_block_info(&self) -> StateResult<CheatnetBlockInfo>;
+}
+
+impl BlockInfoReader for ExtendedStateReader {
+    fn get_block_info(&self) -> StateResult<CheatnetBlockInfo> {
+        if self.fork_state_reader.is_some() {
+            return self.fork_state_reader.as_ref().unwrap().get_block_info();
```
piotmag769, line 42: `if let Some(fork_state_reader) = ...` would let us avoid the `unwrap`

abulenok: `if let Some(fork_state_reader) = self.fork_state_reader` will require for `self.fork_state_reader` to be moved, which is forbidden.

piotmag769: What about `if let Some(ref fork_state_reader) = ...`?

MaksymilianDemitraszek:
Please next time update the changelog before opening a PR 🙏

crates/cheatnet/src/forking/state.rs
abulenok: updated

crates/cheatnet/src/state.rs
abulenok: good point! done

crates/cheatnet/src/forking/cache.rs
```diff
@@ -21,6 +22,7 @@ struct ForkCacheContent {
     class_hash_at: HashMap<String, String>,
     compiled_contract_class: HashMap<String, String>,
     compiled_class_hash: HashMap<String, String>,
+    block_info: Option<CheatnetBlockInfo>,
```
MaksymilianDemitraszek, line 25: Why this is optional?

crates/cheatnet/src/forking/cache.rs
```diff
@@ -61,6 +64,10 @@ impl ForkCacheContent {
         // compiled_class_hash
         self.compiled_class_hash
             .extend(other.compiled_class_hash.clone());
+        // block_info
```
MaksymilianDemitraszek, line 67:
```suggestion
```
Can you remove comments above as well? Those are kinda useless and only introduce unecessary mess.

crates/cheatnet/src/forking/state.rs
```diff
@@ -45,6 +52,38 @@ impl ForkStateReader {
...
+                };
+
+                self.cache.cache_get_block_info(block_info);
+
+                Ok(block_info)
+            }
+            Ok(MaybePendingBlockWithTxHashes::PendingBlock(_)) => Err(StateReadError(format!(
+                "Block with id {:?} is pending",
+                self.block_id
+            ))),
```
MaksymilianDemitraszek, line 79: I am not sure if this is correct semantics, we should investigate it

crates/forge/tests/integration/setup_fork.rs
abulenok: added a test for the Latest block

The problem with the Pending block error is that this error fails the runner

crates/cheatnet/src/forking/cache.rs
abulenok: Because it is cached when the `fork_state_reader.get_block_info` is called for the first time

crates/cheatnet/src/forking/state.rs
piotmag769: Wdym not correct, the match?

MaksymilianDemitraszek: I mean returning the error.

crates/forge/src/running.rs
```diff
@@ -188,17 +188,22 @@ pub(crate) fn run_test_case(
             &case.fork_config,
         )?,
     };
-    let mut context = build_context();
+    let block_info = state_reader.get_block_info().unwrap();
```
piotmag769, line 191: It should result in a failed test, not panic the runner. The best option would be make `run_test_case` return the result of `run_function` and match the error in `blocking_run_from_test`. This way we can `map_err` here and bubble the error

piotmag769 (changes requested):
Requesting changes to make sure to rereview after changes are made

crates/cheatnet/src/forking/state.rs
abulenok: changed it, now `pending_block.number = pending_block.parent_block.number + 1`

abulenok: Although I didn't add a test for forking a pending block, because it fails to fetch it sometimes: `StarknetError(StarknetErrorWithMessage { code: Known(BlockNotFound), message: \"Block not found\" })`

Probably at certain moments, there is no `Pending` block

piotmag769: Same with `Latest`, I think it's fine to include those

abulenok: I'm afraid such a test will make the CI fail too often

piotmag769: If that actually happens more often than once in a while (the one with `Latest` happened once for me in CI, locally a few times) then it's fine to not include it

war-in (approved):
Should we add proxy and library call tests too?

crates/cheatnet/src/forking/state.rs
abulenok: after an internal discussion, we have decided that it would be better to remove `pending` block forking https://github.com/foundry-rs/starknet-foundry/issues/938

crates/cheatnet/src/forking/state.rs
```diff
@@ -45,6 +52,58 @@ impl ForkStateReader {
...
+                    block_number: BlockNumber(block.block_number),
+                    timestamp: BlockTimestamp(block.timestamp),
+                    sequencer_address: ContractAddress(patricia_key!(block.sequencer_address)),
+                };
+
+                self.cache.cache_get_block_info(block_info);
+
+                Ok(block_info)
+            }
+            Ok(MaybePendingBlockWithTxHashes::PendingBlock(pending_block)) => {
```
Arcticae, line 76: This is barely comprehensible, can we extract getting the block info into a method?

Arcticae (approved):

crates/cheatnet/src/forking/state.rs
abulenok: updated

piotmag769 (approved):
Gj!

crates/forge/src/running.rs
```diff
@@ -215,26 +234,44 @@ pub(crate) fn run_test_case(
...
+                name: case.name.clone(),
+                msg: Some(format!(
+                    "\n    {}\n",
+                    error.to_string().replace(" Custom Hint Error: ", "\n    ")
+                )),
+                arguments: args,
+                fuzzing_statistic: None,
+            })
+        }
+        Err(TestCaseRunError::StateError(error)) => Ok(TestCaseSummary::Failed {
```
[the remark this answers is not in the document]
abulenok: Added

PR #903, opened 2023-10-17
Add correct class hashes to tests and fix get_class_hash impl

### LINKED ISSUE #689: Add a test that verifies `class_hash`es are calculated correctly
We only verify that `class_hash` is not 0, but there is no test that actually compares the calculated value agains an expected, precalculated one. This way we can have values for class hashes that are completely inconsistent with starknet but not be aware of it at all.

abulenok:
Closes #689 

## Introduced changes

- Add expected class hashes to `declare` tests. Class hashes are obtained by declaring contracts with `sncast`
- Fix `get_class_hash` implementation

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/cheatnet/src/cheatcodes/declare.rs
```diff
@@ -71,14 +69,8 @@ impl BlockifierState<'_> {
...
-    let sierra_class: ContractClass = serde_json::from_str(sierra_contract)?;
-    let abi_flattened = sierra_class.abi.unwrap().json();
-    let mut sierra_contract_map: HashMap<String, Value> = serde_json::from_str(sierra_contract)?;
-    sierra_contract_map.insert("abi".to_string(), Value::String(abi_flattened));
-    let sierra_contract = serde_json::to_string_pretty(&sierra_contract_map)?;
-
-    let sierra_class: FlattenedSierraClass = serde_json::from_str(&sierra_contract)?;
-    let class_hash = sierra_class.class_hash();
+    let sierra_class: SierraClass = serde_json::from_str(sierra_contract)?;
+    let class_hash = sierra_class.class_hash().unwrap();
```
piotmag769, line 73: `expect` with an error message would be better

crates/cheatnet/tests/cheatcodes/declare.rs
```diff
@@ -1,8 +1,15 @@
...
 use cheatnet::cheatcodes::{CheatcodeError, EnhancedHintError};
 use conversions::StarknetConversions;
+use starknet_api::core::ClassHash;
+use starknet_api::hash::StarkFelt;
+use starknet_api::stark_felt;
+
+static HELLO_STARKNET_CLASS_HASH: &str =
+    "0x298f80e468953d1e65314b6bc63347c7a3fe454a89c2b15387dd52ee281d103";
+static CONSTRUCTOR_SIMPLE_CLASS_HASH: &str =
+    "0x02dbeae7583f3dd4af0bc2da4d58611433165fec7e31245bfa2f1378fbff6a4c";
```
piotmag769, line 12: Calculated using starknet py I guess?

piotmag769 (approved):

crates/cheatnet/tests/cheatcodes/declare.rs
abulenok: with `sn_cast`, mentioned it in PR description

crates/cheatnet/src/cheatcodes/declare.rs
abulenok: Added error propagation since `get_class_hash` returns `Result`

crates/cheatnet/src/cheatcodes/declare.rs
```diff
@@ -71,14 +69,8 @@ impl BlockifierState<'_> {
...
 
 fn get_class_hash(sierra_contract: &str) -> Result<ClassHash> {
-    let sierra_class: ContractClass = serde_json::from_str(sierra_contract)?;
-    let abi_flattened = sierra_class.abi.unwrap().json();
-    let mut sierra_contract_map: HashMap<String, Value> = serde_json::from_str(sierra_contract)?;
-    sierra_contract_map.insert("abi".to_string(), Value::String(abi_flattened));
-    let sierra_contract = serde_json::to_string_pretty(&sierra_contract_map)?;
-
-    let sierra_class: FlattenedSierraClass = serde_json::from_str(&sierra_contract)?;
-    let class_hash = sierra_class.class_hash();
```
war-in, line 81: What happened here before? Did we extract abi and then insert it again? Does anyone have an idea?

war-in (approved):

crates/cheatnet/src/cheatcodes/declare.rs
piotmag769: cc @cptartur

crates/cheatnet/tests/cheatcodes/declare.rs
```diff
@@ -1,8 +1,15 @@
...
 use crate::common::{get_contracts, state::create_cached_state};
-use cairo_felt::Felt252;
 use cheatnet::cheatcodes::{CheatcodeError, EnhancedHintError};
 use conversions::StarknetConversions;
+use starknet_api::core::ClassHash;
+use starknet_api::hash::StarkFelt;
+use starknet_api::stark_felt;
+
+static HELLO_STARKNET_CLASS_HASH: &str =
+    "0x298f80e468953d1e65314b6bc63347c7a3fe454a89c2b15387dd52ee281d103";
```
MaksymilianDemitraszek, line 10: Oh I think I misguided you, I thought you mean hardcoding declared contracts in cairo code as those stay on chain forever. In such case as here it will change with each sierra version bump.

MaksymilianDemitraszek: I think it is still good to have such test, though it will be pain to re-calculate it each time with sncast. Maybe we could use `starknet.rs` to automate it here?

crates/cheatnet/src/cheatcodes/declare.rs
cptartur: Scarb artifacts represent ABIs as json structures, where blockifier expected the ABI to be a flat string (this is how actual network works as well for cairo1 contracts), so I had to flatten the abi, otherwise serialization would fail.

I see a different type is used for sierra class now that seems to accept abis like this.

Arcticae (approved):

crates/cheatnet/tests/cheatcodes/declare.rs
MaksymilianDemitraszek: It is ok to merge this one and add another issue for it

abulenok: added https://github.com/foundry-rs/starknet-foundry/issues/929

PR #915, opened 2023-10-18
`StarkCurveKeyPair` implementation

### LINKED ISSUE #879: Facilitate generating signatures using `ecdsa` and `ecdsa` keypairs
It could be both implemented using an external library & cheatcodes or pure cairo

war-in:
Closes #879

## Introduced changes

- implemented proposal from `design_documents/stark_curve_signatures.md`

## Checklist

- [ ] Linked relevant issue
- [ ] Added relevant tests

### DIFF
--- a/crates/cheatnet/src/cheatcodes/ecdsa_utils.rs
+++ b/crates/cheatnet/src/cheatcodes/ecdsa_utils.rs
@@ -1,0 +1,25 @@
+use cairo_felt::Felt252;
+use conversions::StarknetConversions;
+use starknet::signers::SigningKey;
+
+#[must_use]
+pub fn generate_ecdsa_keys() -> (Felt252, Felt252) {
+    let key_pair = SigningKey::from_random();
+
+    (
+        key_pair.secret_scalar().to_felt252(),
+        key_pair.verifying_key().scalar().to_felt252(),
+    )
+}
+
+pub fn ecdsa_sign_message(
+    private_key: &Felt252,
+    message_hash: &Felt252,
+) -> Result<(Felt252, Felt252), String> {
+    let key_pair = SigningKey::from_secret_scalar(private_key.to_field_element());
+
+    match key_pair.sign(&message_hash.to_field_element()) {
+        Ok(signature) => Ok((signature.r.to_felt252(), signature.s.to_felt252())),
+        Err(_) => Err("message_hash out of range".to_string()),
+    }
+}
--- a/crates/cheatnet/src/cheatcodes/mod.rs
+++ b/crates/cheatnet/src/cheatcodes/mod.rs
@@ -11,6 +11,7 @@ use thiserror::Error;
 
 pub mod declare;
 pub mod deploy;
+pub mod ecdsa_utils;
 pub mod get_class_hash;
 pub mod l1_handler_execute;
 pub mod mock_call;
--- a/crates/forge/tests/integration/signing.rs
+++ b/crates/forge/tests/integration/signing.rs
@@ -1,0 +1,51 @@
+use crate::integration::common::running_tests::run_test_case;
+use crate::{assert_passed, test_case};
+use indoc::indoc;
+
+#[test]
+fn generate_keys_sing_verify() {
+    let test = test_case!(indoc!(
+        r#"
+            use snforge_std::{ StarkCurveKeyPair, StarkCurveKeyPairTrait };
+
+            #[test]
+            fn test() {
+                let mut key_pair = StarkCurveKeyPairTrait::generate();
+                let message_hash = 123456;
+
+                let signature = key_pair.sign(message_hash);
+                assert(key_pair.verify(message_hash, signature), 'Wrong signature');
+            }
+        "#
+    ));
+
+    let result = run_test_case(&test);
+
+    assert_passed!(result);
+}
+
+#[test]
+fn try_to_sign_max_felt_plus_1() {
+    let test = test_case!(indoc!(
+        r#"
+            use snforge_std::{ StarkCurveKeyPair, StarkCurveKeyPairTrait };
+
+            #[test]
+            fn test() {
+                let mut key_pair = StarkCurveKeyPairTrait::generate();
+                let max_felt = 3618502788666131213697322783095070105623107215331596699973092056135872020480;
+
+                match key_pair.sign(max_felt) {
+                    Result::Ok(_) => panic_with_felt252('shouldve panicked'),
+                    Result::Err(msg) => {
+                        assert(msg == 'message_hash out of range', msg);
+                    }
+                }
+            }
+        "#
+    ));
+
+    let result = run_test_case(&test);
+
+    assert_passed!(result);
+}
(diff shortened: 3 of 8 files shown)


crates/cheatnet/src/cheatcodes/ecdsa_utils.rs
Arcticae, line 1: i wouldn't place this file there, and call it with a postfix `_utils`
This makes the structure more confusing

crates/forge/tests/integration/signing.rs
```diff
@@ -0,0 +1,51 @@
+use crate::integration::common::running_tests::run_test_case;
+use crate::{assert_passed, test_case};
+use indoc::indoc;
+
+#[test]
+fn generate_keys_sing_verify() {
```
Arcticae, line 6:
```suggestion
fn simple_signing_flow() {
```

Arcticae (changes requested):
I would also add a complex test which would:
1. Deploy an account with generated keys
2. Generate a call to sample contract through the account
3. Sign the call
4. Call the accounts' `__execute__` with `spoof`ed signature and see if it passed the verification

MaksymilianDemitraszek (dismissed):
I would like to review this one before merging

crates/forge/tests/integration/signing.rs
MaksymilianDemitraszek, line 1: I assume that purpose of this implementation is to allow to sign something which can be verified later with corelib builtins? Could we have a test which checks if generated signature can be verified with corelib builtins?

snforge_std/src/lib.cairo
```diff
@@ -43,3 +43,5 @@ use forge_print::PrintTrait;
 mod fs;
 
 mod env;
+
+mod signing;
```
MaksymilianDemitraszek, line 47:
```suggestion
mod signature;
```
Other module names are consistent with rust, so this one should be as well.

snforge_std/src/cheatcodes/l1_handler.cairo
```diff
@@ -17,9 +17,7 @@ trait L1HandlerTrait {
 
 impl L1HandlerImpl of L1HandlerTrait {
     fn new(contract_address: ContractAddress, function_name: felt252) -> L1Handler {
-        L1Handler {
-            contract_address, function_name, from_address: 0, payload: array![].span(),
-        }
+        L1Handler { contract_address, function_name, from_address: 0, payload: array![].span(), }
```
MaksymilianDemitraszek, line 20: Unrelated change, makes it hard to review

crates/forge/tests/e2e/signing.rs
```diff
@@ -0,0 +1,26 @@
...
+        [..]Finished[..]
+
+
+        Collected 1 test(s) from signing package
+        Running 1 test(s) from src/
+        [PASS] signing::test
+        Tests: 1 passed, 0 failed, 0 skipped
+        "#}
+    );
+}
```
MaksymilianDemitraszek, line 26: I don't think we need an e2e test here 🤔 What does it verify above things we verify already in the integration?

snforge_std/src/signing/stark_curve.cairo
```diff
@@ -0,0 +1,56 @@
+use starknet::testing::cheatcode;
+use ecdsa::check_ecdsa_signature;
+
+#[derive(Copy, Drop)]
+struct StarkCurveKeyPair {
+    private_key: felt252,
+    public_key: felt252
+}
+
+trait StarkCurveKeyPairTrait {
```
MaksymilianDemitraszek, line 10: Traits should abstract over properties, not be struct specific. Also please make it consistent with rust `signature` crate.
Please split it into `Signer` `Verifier`. `generate` and `from_private_key` shouldn't be part of the trait.

crates/cheatnet/src/cheatcodes/mod.rs
```diff
@@ -17,6 +17,7 @@ pub mod mock_call;
 pub mod prank;
 pub mod precalculate_address;
 pub mod roll;
+pub mod signing;
```
MaksymilianDemitraszek, line 20:
```suggestion
pub mod signature;
```
For consistency with rust please create signature module keeping the general traits.
Please create module with specific crypto primitive name and keep the implementations of this primitive there.

MaksymilianDemitraszek (changes requested):
LGTM but please make the module structure consistent with Rust.

crates/forge/tests/e2e/signing.rs
war-in: It gets oz account and verifies the signature generated by us. But we could probably get rid of that because it uses basically the same logic under the hood

crates/forge/tests/integration/signing.rs
war-in: We can, but we use corelib function to perform verification, so it will work

crates/forge/tests/e2e/signing.rs
MaksymilianDemitraszek: Oh ok, could we do this as integration test instead?

crates/forge/tests/integration/signing.rs
MaksymilianDemitraszek: Oh, I see, my bad i thought we are using starknet-rs for verification as well. Then I think integration tests we have now are sufficient!

crates/forge/tests/e2e/signing.rs
war-in: Yup, but we would have to copy the OZ account 💀

Arcticae (approved):

piotmag769 (approved):

crates/forge/tests/e2e/common/runner.rs
```diff
@@ -32,25 +33,17 @@ pub(crate) fn setup_package_with_file_patterns(
...
-                casm = true
 
-                [dependencies]
-                starknet = "2.2.0"
-                snforge_std = {{ path = "{}" }}
-                "#,
-            package_name,
-            snforge_std_path
-        ))
+    let mut scarb_toml = fs::read_to_string(&manifest_path)
```
MaksymilianDemitraszek, line 37: [[Nitpick]] I would remove it from this PR as it is completely unrelated and makes commit history messy, though knowing the reasons why its here I assume it would be a bit annoying so marking it as nit.

docs/src/appendix/forge-library/signature/stark_curve_key_pair.md
```diff
@@ -0,0 +1,55 @@
...
+    let max_felt = 3618502788666131213697322783095070105623107215331596699973092056135872020480;
+
+    match key_pair.sign(max_felt) {
+        Result::Ok(_) => panic_with_felt252('shouldve panicked'),
+        Result::Err(msg) => {
+            assert(msg == 'message_hash out of range', msg);
+        }
+    }
+}
+```
```
MaksymilianDemitraszek, line 55: This is definitely too much info about message overflow, can we just add 
> When signing and the `message_hash` is too big
(`> 0x800000000000000000000000000000000000000000000000000000000000000`)
the `Result::Err('message_hash out of range')` error will be returned.

As a warning callout? Example is not necessary imo

MaksymilianDemitraszek (approved):

PR #919, opened 2023-10-19
Move fork config parsing to `lib.rs`

piotmag769:
Refactors for #757 

## Introduced changes

-

## Checklist

- [ ] Linked relevant issue
- [ ] Added relevant tests


crates/forge/src/lib.rs
```diff
@@ -193,6 +194,58 @@ fn try_close_tmp_dir(temp_dir: TempDir) -> Result<()> {
     Ok(())
 }
 
+fn find_or_parse_raw_fork_config(
+    fork_config: ForkConfig,
```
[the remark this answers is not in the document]
MaksymilianDemitraszek: We could make TestCase generic to achieve it.

MaksymilianDemitraszek (changes requested):

crates/forge/src/lib.rs
```diff
@@ -193,6 +194,58 @@ fn try_close_tmp_dir(temp_dir: TempDir) -> Result<()> {
...
+        let fork_config = if let Some(fc) = case.fork_config {
+            Some(find_or_parse_raw_fork_config(fc, runner_config)?)
+        } else {
+            None
+        };
+        test_cases.push(TestCase {
+            fork_config,
+            ..case
+        });
+    }
```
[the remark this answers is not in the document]
piotmag769: Cannot do that bcs `map` on option cannot return a result, therefore I cannot bubble the error

crates/forge/src/lib.rs
```diff
@@ -194,6 +195,61 @@ fn try_close_tmp_dir(temp_dir: TempDir) -> Result<()> {
...
+            block_id,
+        }),
+        RawForkConfig::Id(name) => {
+            let fork_target_from_runner_config = runner_config
+                .fork_targets
+                .iter()
+                .find(|fork| fork.name == name)
+                .ok_or_else(|| {
+                    anyhow!("Fork configuration named = {name} not found in the Scarb.toml")
+                })?; // TODO essa test it
```
abulenok, line 214: TODO

piotmag769: Ik, adding it rn

crates/test-collector/src/lib.rs
```diff
@@ -61,8 +62,18 @@ pub enum ExpectedTestResult {
...
 }
 
+pub trait ForkConfig {}
+
+impl ForkConfig for ValidatedForkConfig {}
+impl ForkConfig for RawForkConfig {}
+
 #[derive(Debug, Clone, PartialEq)]
-pub enum ForkConfig {
+pub struct ValidatedForkConfig {
```
piotmag769, line 71: It is used only in forge, but I would keep it in this crate for consistency

crates/forge/src/lib.rs
```diff
@@ -194,6 +195,61 @@ fn try_close_tmp_dir(temp_dir: TempDir) -> Result<()> {
...
+        });
+    }
+
+    Ok(CompiledTestCrate {
+        sierra_program: compiled_test_crate.sierra_program,
+        test_cases,
+        tests_location: compiled_test_crate.tests_location,
+    })
+}
+
```
cptartur, line 252: I think this should be done somewhere in the `collecting.rs` module and treated as part of test collection or compilation. Details of fork configs should not concern us later. 

We should only be dealing with ValidatedForkConfig post that stages.

crates/forge/src/collecting.rs
```diff
@@ -5,13 +5,13 @@ use assert_fs::TempDir;
...
 use walkdir::WalkDir;
 
 #[derive(Debug, Clone)]
-pub(crate) struct CompiledTestCrate {
+pub(crate) struct CompiledTestCrate<T: ForkConfig> {
     pub sierra_program: Program,
-    pub test_cases: Vec<TestCase>,
+    pub test_cases: Vec<TestCase<T>>,
     pub tests_location: CrateLocation,
 }
```
cptartur, line 16: This parametrization only introduces complexity without any direct benefit. As I wrote below, we should always parse fork configs and only return test cases with it parsed, no parametrization needed, especially since basically all code post test collection is coded to expect `TestCase<ValidatedForkConfig>>`.

crates/forge/src/scarb.rs
```diff
@@ -832,8 +785,37 @@ mod tests {
...
+            .inherit_stderr()
+            .current_dir(temp.path())
+            .exec()
+            .unwrap();
+
+        let err =
+            config_from_scarb_for_package(&scarb_metadata, &scarb_metadata.workspace.members[0])
+                .unwrap_err();
+        assert!(format!("{err:?}")
+            .contains("Could not parse the url = unparsable_url from Scarb.toml:"));
```
cptartur, line 819: `err.to_string()` should be enough here.

crates/forge/src/scarb/config.rs
```diff
@@ -0,0 +1,127 @@
...
+    pub block_id: HashMap<String, String>,
+}
+
+pub(super) fn validate_raw_fork_config(raw_config: &RawForgeConfig) -> Result<()> {
+    let forks = &raw_config.fork;
+    let names: Vec<String> = forks.iter().map(|fork| fork.name.clone()).collect();
+    let removed_duplicated_names: Vec<String> = names.clone().into_iter().unique().collect();
+
+    if names.len() != removed_duplicated_names.len() {
+        return Err(anyhow!("Some fork names are duplicated"));
```
cptartur, line 57:
```suggestion
        bail("Some fork names are duplicated"));
```

cptartur: Same below

crates/forge/src/scarb/config.rs
```diff
@@ -0,0 +1,127 @@
...
+    for fork in forks {
+        let block_id_items: Vec<(&String, &String)> = fork.block_id.iter().collect();
+        let [(block_id_key, block_id_value)] = block_id_items[..] else {
+            return Err(anyhow!("block_id should be set once per fork"));
+        };
+
+        if !["number", "hash", "tag"].contains(&&**block_id_key) {
+            return Err(anyhow!(
+                "block_id has only three variants: number, hash and tag"
+            ));
```
cptartur, line 69: This errors should be structured more like `"block_id = {.. user input ..} is not valid. Possible values = are "number", "hash" and "tag"`. 

With errors like this, it's hard for the users to find where exactly in the fork configuration the error was made.

crates/test-collector/src/lib.rs
cptartur: I'd move it to forge, this structure doesn't represent something returned from test collector, it is our abstraction over what collector can return.

cptartur (changes requested):

crates/forge/src/scarb.rs
piotmag769: It is not since I use `context` above this error

crates/forge/src/collecting.rs
```diff
@@ -5,13 +5,18 @@ use assert_fs::TempDir;
...
 
+pub(crate) type CompiledTestCrateRaw = CompiledTestCrate<RawForkConfig>;
+pub(crate) type CompiledTestCrateRunnable = CompiledTestCrate<ValidatedForkConfig>;
+
 #[derive(Debug, Clone)]
-pub(crate) struct CompiledTestCrate {
+pub(crate) struct CompiledTestCrate<T: ForkConfig> {
     pub sierra_program: Program,
-    pub test_cases: Vec<TestCase>,
+    pub test_cases: Vec<TestCase<T>>,
```
cptartur, line 19: Let's also add a type `TestCaseRunnable` or `RunnableTestCase` that is `TestCase<CompiledTestCrateRunnable>` and use it in methods in `lib.rs` and `running.rs`.

cptartur (approved):

crates/forge/src/scarb/config.rs
```diff
@@ -0,0 +1,123 @@
...
+    /// Seed to be used by fuzzer
+    pub fuzzer_seed: Option<u64>,
+
+    pub fork: Vec<ForkTarget>,
+}
+
+#[derive(Debug, PartialEq, Clone)]
+pub struct ForkTarget {
+    pub name: String,
+    pub url: Url,
```
MaksymilianDemitraszek, line 25:
```suggestion
    pub url: String,
```
we should keep validation in one place

crates/test-collector/src/lib.rs
```diff
@@ -61,8 +61,12 @@ pub enum ExpectedTestResult {
...
 
+pub trait ForkConfig {}
+
+impl ForkConfig for RawForkConfig {}
+
 #[derive(Debug, Clone, PartialEq)]
-pub enum ForkConfig {
+pub enum RawForkConfig {
     Id(String),
     Params(String, BlockId),
```
MaksymilianDemitraszek, line 71:
```suggestion
    Params(RawForkParams),
```
please create a separate type and use it for ForkTargets as well. In other comment i suggested making it a pipeline, first function should map from `RawForkConfig` to `RawForkParams` (querying scarb config). And then other function should parse url in params.

MaksymilianDemitraszek (changes requested):
Much better, sorry for the long review time, last two days were a bit messy :(. It is close to being merged, couple last changes as we have two types representing the same thing and it is a bit messy.

crates/forge/src/scarb/config.rs
piotmag769: If we don't parse it when parsing a config it won't be obvious to the user that the unparsable url is in Scarb. We don't do that in compilation (since we want to split compilation and url parsing) but I think if we can provide better UX without a lot of cost in terms of code abstraction we should do that

crates/test-collector/src/lib.rs
piotmag769: Okay, that convinces me to provide worse UX xD

crates/forge/src/scarb/config.rs
piotmag769: Nvm, see below

crates/forge/src/running.rs
```diff
@@ -279,70 +274,29 @@ fn extract_test_case_summary(
...
-    match &fork_config {
-        Some(ForkConfig::Params(url, mut block_id)) => {
+    match fork_config {
+        Some(ValidatedForkConfig { url, mut block_id }) => {
             if let BlockId::Tag(Latest) = block_id {
                 block_id = get_latest_block_number(url)?;
             }
             Ok(Some(ForkStateReader::new(
-                url,
+                url.clone(),
```
MaksymilianDemitraszek, line 285: Why do we need to clone it?

MaksymilianDemitraszek (approved):

crates/forge/src/running.rs
piotmag769: It is due to the fact that inside `ForkStateReader::new` we call `HttpTransport::new` of signature `pub fn new(url: impl Into<Url>) -> Self`. `&Url` doesn't implement `Into<Url>` so we would have to clone it anyways. 

Therefore I decide to ensure that `Url` is passed to `ForkStateReader::new`, not `&Url`, so the copy is not hidden somewhere

PR #923, opened 2023-10-19
Allow passing calldata without quotes

### LINKED ISSUE #920: Allow passing calldata without quotes

Cast

### Feature Request

There should be a clap config parameter to set this

drknzz:
Closes #920


crates/cast/src/starknet_commands/call.rs
```diff
@@ -19,7 +19,7 @@ pub struct Call {
     pub function: String,
 
     /// Arguments of the called function (list of hex)
-    #[clap(short, long, value_delimiter = ' ')]
+    #[clap(short, long, value_delimiter = ' ', num_args = 1..)]
```
THenry14, line 22: how does this work? will it be possible to change an order of arguments (eg do soething like `--calldata 0x1 0x2 0x3 --function whatever`) or does this enforces calldata being a positional argument?

drknzz: [[docs for reference]](https://docs.rs/clap/latest/clap/builder/struct.Arg.html#method.num_args)

`num_args` sets the number of arguments for a specific flag
it is still be possible to use other flags after it ([parsing of arguments](https://docs.rs/clap/latest/clap/builder/struct.Arg.html#method.num_args:~:text=It%20finds%20another%20flag%20or%20option))

I've adjusted the tests to check for the concerns you've raised up 👍.
Previously, if calldata was passed to `Command::new()` like this

```
args.append(&mut vec![
    "deploy",
    "--constructor-calldata",
    "0x1",
    "0x1",
    "0x0",
    "--class-hash",
    &class_hash,
]);
```

the tests would fail with:

```sh
running 1 test
thread 'e2e::deploy::test_happy_case_with_constructor' panicked at 'Expected success, was 2
stderr:
error: unexpected argument '0x1' found

Usage: sncast deploy [OPTIONS] --class-hash <CLASS_HASH>

For more information, try '--help'.

', crates/cast/tests/e2e/deploy.rs:56:35
```

After adding `num_args`, the tests do pass.

THenry14: Perfect, thanks for checking this

THenry14 (approved):
Let's add info about this change to CHANGELOG. After that, this can me merged :+1:

PR #950, opened 2023-10-24
Automate release procedure

### LINKED ISSUE #859: Automate release procedure
We should add as script to the repository which would:
- Bump the versions (minor/patch as options perhaps?) in Cargo.toml
- Mark unreleased changelog sections with the version + date, add a new "template" for next release
- Create a tag with the new version
- Commit the changes
- Push the tag to the repo
- Emit a message that a release has been started. 

Display at the end (if possible):
- a link to release action
- a link to the draft release the changelog in the console with the instructions to copy it into the release notes on GH

Bonus idea:
If the release action could create a draft release with the input already pre-filled with the contents of latest changelog, this would be even better

cptartur:
Closes #859 

## Introduced changes

- Added a script for updating `Cargo.toml`, `Cargo.lock` and `CHANGELOG.md` files
- Release workflow is now also triggered from the master branch if version in `Cargo.toml` increased

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


.github/workflows/release.yml
```diff
@@ -10,23 +13,71 @@ permissions:
...
+          excludes: prerelease, draft
+
+      - name: Compare versions
+        id: validVersion
+        run: |
+          RELEASE_VERSION=${{ steps.lookupVersionRelease.outputs.release }}
+          COMMIT_VERSION=${{ steps.lookupVersion.outputs.result }}
+          echo "Project version from newest release = $RELEASE_VERSION"
+          echo "Project version from this commit = $COMMIT_VERSION"
+          IS_VALID=$(node ./scripts/compareVersions.js $RELEASE_VERSION $COMMIT_VERSION)
```
THenry14, line 52: was there a specific reason to write js script, and not just use bash?

scripts/release.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+VERSION=$1
+
+sed -i.bak "s/## \[Unreleased\]/## \[Unreleased\]\n\n## \[${VERSION}\] - $(TZ=Europe/Krakow date '+%Y-%m-%d')/" CHANGELOG.md
+rm CHANGELOG.md.bak 2> /dev/null
```
THenry14, line 6: it seems that we are not using .bak for anything? so maybe we can just not create it, and change in place?

cptartur: MacOS's version of sed doesn't play nicely with just `sed -i '....'` and I've found this to be a solution: https://stackoverflow.com/questions/5694228/sed-in-place-flag-that-works-both-on-mac-bsd-and-linux

.github/workflows/release.yml
cptartur: Handling semver versions in bash seemed much more complicated than using `semver` library with node. If there's some simple solution for bash, we can change to that.

THenry14: not a problem, I was just curious

THenry14 (approved):
lgtm

PR #954, opened 2023-10-24
Add derive Copy for ContractClass

tensojka:
## Introduced changes

1-line change to derive Copy for ContractClass, this is needed to make our tests run. I don't see any downsides to this.

## Breaking changes

None


piotmag769 (commented):
I think you should be able to call `clone` on this struct and your tests will work 🤔 I wonder what is the case in which you need to clone it anyways thought, it seems weird that you need it

cptartur (approved):
Don't see any immediate downsides of this.

PR #955, opened 2023-10-24
Bump Cairo and Scarb to 2.3.0

piotmag769:
## Introduced changes

Self-explanatory.

## Checklist

- [x] Linked relevant issue
- [x] Updated relevant documentation
- [x] Added relevant tests
- [x] Performed self-review of the code
- [ ] Added changes to `CHANGELOG.md`

PR #960, opened 2023-10-25
Remove pending block forking

### LINKED ISSUE #938: Remove `Pending` block forking
It doesn't make much sense, since the test with the `Pending` block forked, will change the block context too often

drknzz:
Closes #938 

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - separated `snforge` to subcommands: `snforge test`, `snforge init` and `snforge clean-cache`. 
 Read more [here](https://foundry-rs.github.io/starknet-foundry/appendix/forge.html).
 - `starknet::get_block_info` now returns correct block info in a forked block
+- forking `Pending` block has become disallowed
```
war-in, line 24: Should we add this to the `Breaking` section?

docs/src/SUMMARY.md
```diff
@@ -72,7 +72,7 @@
         * [parse_json](appendix/forge-library/fs/parse_json.md)
     * [env](appendix/forge-library/env.md)
         * [var](appendix/forge-library/env/var.md)
-    * [signature](appendix/forge-library/signature)
+    * [signature](appendix/forge-library/signature.md)
```
war-in, line 75: 👀

docs/src/testing/fork-testing.md
```diff
@@ -57,16 +61,11 @@ block_id.tag = "Latest"
 [[tool.snforge.fork]]
 name = "SOME_SECOND_NAME"
 url = "http://your.second.rpc.url"
-block_id.tag = "Pending"
+block_id.number = "123"
```
war-in, line 64: I believe we have to change the parsing scarb.toml functions (not sure which ones) because allow passing `block_id.tag = "Pending"` at the moment.
Afaik @piotmag769 refactored it recently, so he can have the best knowledge about it

CHANGELOG.md
piotmag769: It is not automated, we can consider adding it @cptartur

docs/src/SUMMARY.md
piotmag769: Fixed in #955 too, it prevented docs from building xD

drknzz: Docs didn't build because of this 👀 from what I see they are not checked anywhere in CI

docs/src/testing/fork-testing.md
piotmag769: Yep, you can change it when #919 is merged in `crates/forge/src/scarb/config.rs` and in `crates/test-collector/src/lib.rs`

piotmag769 (changes requested):
Please wait with merge until #919 is merged (should be soon)

docs/src/SUMMARY.md
piotmag769: #961

crates/cheatnet/src/forking/state.rs
```diff
@@ -99,9 +73,9 @@ impl BlockInfoReader for ForkStateReader {
 
                 Ok(block_info)
             }
-            Ok(MaybePendingBlockWithTxHashes::PendingBlock(pending_block)) => {
-                get_pending_block_parent(self, &pending_block)
-            }
+            Ok(MaybePendingBlockWithTxHashes::PendingBlock(_)) => Err(StateReadError(
+                "Forking Pending block is not available".to_string(),
+            )),
```
piotmag769, line 78: It should be `unreachable!` I think

drknzz: Probably yes 👍 didn't know about that macro

CHANGELOG.md
MaksymilianDemitraszek: We haven't been explicit about breaking changes in the past, we can consider it to simplify migrations for users.

crates/cheatnet/src/forking/state.rs
MaksymilianDemitraszek: Is it always unreachable though? We removed the tag but isn't it possible to fetch pending giving specific block number? If it is I would keep the human error.

docs/src/testing/fork-testing.md
```diff
@@ -30,6 +30,10 @@ enum BlockTag {
 }
 ```
 
+> 📝 **Note**
```
MaksymilianDemitraszek, line 33: I think two lines above we have type including pending. Also I think it will be enough to not list it instead of explicitly excluding it, it is harder to maintain as we then should list everything we don't include.

MaksymilianDemitraszek (changes requested):

crates/cheatnet/src/forking/state.rs
drknzz: > Is it always unreachable though? We removed the tag but isn't it possible to fetch pending giving specific block number? If it is I would keep the human error.

It seems it's not possible to fetch the pending block with it's block number.

piotmag769: @MaksymilianDemitraszek I think pending block does not have a number, only parent hash

crates/cheatnet/src/forking/state.rs
```diff
@@ -99,9 +73,9 @@ impl BlockInfoReader for ForkStateReader {
 
                 Ok(block_info)
             }
-            Ok(MaybePendingBlockWithTxHashes::PendingBlock(pending_block)) => {
-                get_pending_block_parent(self, &pending_block)
-            }
+            Ok(MaybePendingBlockWithTxHashes::PendingBlock(_)) => Err(StateReadError(
```
Arcticae, line 76: Shouldn't this be resolved on the configuration level? I think it's a bit late in code to throw an error like this (if it's not strictly for validation)

crates/cheatnet/src/forking/state.rs
MaksymilianDemitraszek: So I assume it is poor type encoding of JSONRpcClient?

piotmag769: Nope, their type encoding is fine, `self.block_id` is of type `BlockId`. It is probably a general arg that has to be passed to methods like this (doubt there are separate methods where you pass the number explicitly)

```rust
pub enum BlockId {
    Hash(FieldElement),
    Number(u64),
    Tag(BlockTag),
}
```

crates/cheatnet/src/forking/state.rs
piotmag769: See the conv above^

docs/src/testing/fork-testing.md
piotmag769: @drknzz #919 merged already, feel free to change stuff in the files in mentioned

CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Forge
+
+#### Breaking
```
cptartur, line 12: Let's just call it `Removed`, we've had a section named like this in a couple of previous releases. 

```suggestion
#### Removed
```

CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+
+- forking `Pending` block has become disallowed
```
cptartur, line 14:
```suggestion
- forking of the `Pending` block
```

cptartur (changes requested):
What about fork configurations in code? The code in `test-collector` still allows the block id value to be `pending`.

Shouldn't the code there be changed so we fail at test compilation level and not in runtime? I think having `pending` for block id, should be treated equally to passing some random value to it like `qwerty`.

piotmag769 (commented):
Missing changes in `fn block_id_to_string`

drknzz:
> Missing changes in `fn block_id_to_string`

I know, couldn't it stay that way, so it encodes string for each `BlockId`?

piotmag769:
> I know, couldn't it stay that way, so it encodes string for each BlockId?

I think removing it can help us get rid of random bugs since it should never appear there. Just use `unreachable!` imo

crates/cheatnet/src/forking/state.rs
```diff
@@ -99,8 +73,8 @@ impl BlockInfoReader for ForkStateReader {
 
                 Ok(block_info)
             }
-            Ok(MaybePendingBlockWithTxHashes::PendingBlock(pending_block)) => {
-                get_pending_block_parent(self, &pending_block)
+            Ok(MaybePendingBlockWithTxHashes::PendingBlock(_)) => {
+                unreachable!("Pending block should not be allowed at configuration level")
```
MaksymilianDemitraszek, line 77:
```suggestion
                unreachable!("Pending block is not be allowed at the configuration level")
```

MaksymilianDemitraszek (approved):
Good added/deleted ratio 👍

piotmag769 (approved):

PR #962, opened 2023-10-25
Test docs building in CI

### LINKED ISSUE #961: Test docs building in CI

drknzz:
Closes #961

## Introduced changes

- Add docs build job to CI workflow

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


piotmag769 (approved):
Looks good, can we somehow cache that? @THenry14

THenry14:
> Looks good, can we somehow cache that? @THenry14

Potentially the mdbook install could be cached (~3m atm), the check would then last significantly less time. There is a rust caching done in this file which I suppose could work. 

let's remember to change repo settings after this is merged, to mark the check as required. I'd also suggest adding a `name` label for the check, not a fan of displaying `build-docs` but it's just a prefference and nothing I'm particularly strong about.

drknzz:
> > Looks good, can we somehow cache that? @THenry14
> 
> Potentially the mdbook install could be cached (~3m atm), the check would then last significantly less time. There is a rust caching done in this file which I suppose could work.
> 
> let's remember to change repo settings after this is merged, to mark the check as required. I'd also suggest adding a `name` label for the check, not a fan of displaying `build-docs` but it's just a prefference and nothing I'm particularly strong about.

Added the name label and rust-cache 👍

.github/workflows/ci.yml
```diff
@@ -149,3 +149,21 @@ jobs:
...
+  build-docs:
+    name: Test Building Docs
+    runs-on: ubuntu-latest
+    env:
+      MDBOOK_VERSION: 0.4.31
+    steps:
+      - uses: dtolnay/rust-toolchain@439cf607258077187679211f12aa6f19af4a0af7
+        with:
+          toolchain: stable
+      - uses: Swatinem/rust-cache@v2
```
THenry14, line 162: this is a nitpick, but I think we should stick to hashes when it goes for non-official actions like this one; tag can be easily overwritten by a maintainer and injected with mallicious code. I understand that someone already did the same in this file (I mean marked this action with a tag), so I won't block the review, but this is something that in my opinion should be changed (again - not necessarily in this PR).

THenry14 (approved):

.github/workflows/ci.yml
drknzz: Noticed it as well, but thought maybe there was already some discussion about it previously, since it's used in a few places in the file - will update to point to a commit 👍

Issue #970, opened 2023-10-25
Global cheatcodes


piotmag769:
Cheatcodes cheating everything globally e.g. `start_global_warp` changing the timestamp everywhere in the execution

Open questions: what should take priority, global or local cheatcode? Or should it depend on the order in which the cheatcodes were called?

Arcticae:
Closing in favour of 
- #1011 
- #1010 
- #1009

PR #972, opened 2023-10-25
feat: `start_warp_global` and `end_warp_global`

bllu404:
This PR partially addresses  #970 by adding `start_warp_global` and `end_warp_global` cheatcodes. 

`start_warp_global` overrides `start_warp`, meaning if you call `start_warp` for contract A with `timestamp == 500`, and call `start_warp_global` with `timestamp == 400`, then regardless of the order in which they are called, the timestamp for contract A will be `400`, together with all other contracts.


Arcticae:
@bllu404 can you add docs for the cheatcode?

bllu404:
Should calling `start_warp_global` clear the `CheatnetState.warped_contracts` hash map? 

My concern is that users may intuitively do something like this, leading to confusion:
```cairo 
#[test]
fn test_foo () {
   // do stuff
 
   start_warp(foo.contract_address, ts); 

   // do stuff 

   // now we need to change the timestamp for all contracts for some reason
   start_warp(different_ts)

   // do stuff

   stop_warp_global() // User cancels the global warp, thinking it resets all contract timestamps to zero

   foo.bar(x, y, z); // < -- Fails due to incorrect timestamp

}
   
```

Perhaps it should even be enforced that if `start_warp` has been called, then `start_global_warp` can't be called until `stop_warp` has been called for ALL warped contracts, and vice versa (If you called `start_global_warp`, then you can't call `start_warp` until you call `stop_global_warp`. 

What do you think?

Arcticae:
@bllu404 we would like to have a detailed design doc about this. Is this something that you would be willing to do? 
We do have a slightly different vision on the implementation and its' details
We would like to have something like the `Events.spyOn()` interface, where we'd pass an enum value of `All`, `One`, `Multiple` and would have that behave accordingly.

As for the cancellation it also would be clearer, since it'd accept similar thing, and if `All` would be passed, it'd just clear all the mocks.

WDYT?
Let me know if you want to push it further :)

bllu404:
@Arcticae, I implemented the interface as you described it. 

I can also create a design doc, I would just need to know what you're looking for specifically. Do you want a new document specifically on `warp`?

crates/cheatnet/src/cheatcodes/warp.rs
```diff
@@ -2,12 +2,56 @@ use crate::CheatnetState;
...
+    Multiple(Vec<(ContractAddress, Felt252)>),
+}
+
+// Specifies which contracts to stop warping
+#[derive(Debug)]
+pub enum StopWarpTarget {
+    All,
+    One(ContractAddress),
+    Multiple(Vec<ContractAddress>),
+}
```
Arcticae, line 19: It's really similar to what i thought, but 
"Target" should be the 1st argument 
and 2nd argument of warp should be the target timestamp

And the "Target" enum would have associated addresses values, so that multiple would have an Vec of `ContractAddress`es, `One` would have a singular `ContractAddress` and `All` would be without extra arguments. 

Also as for the struct, I think that it should be named something more generic and placed higher up in the file hierarchy.

The changes i mentioned should be applied both to `snforge_std` and the rust code

Arcticae (changes requested):

crates/forge/src/test_execution_syscall_handler.rs
```diff
@@ -266,20 +269,56 @@ impl TestExecutionSyscallHandler<'_, '_> {
...
+                } else if inputs[0] == Felt252::from(2) {
+                    let warp_targets: Vec<_> = inputs[2..]
+                        .iter()
+                        .map(Felt252::to_contract_address)
+                        .collect();
+                    self.contract_execution_syscall_handler
+                        .cheatable_syscall_handler
+                        .cheatnet_state
+                        .stop_warp(StopWarpTarget::Multiple(warp_targets));
+                }
```
piotmag769, line 320: `match` statement would be more idiomatic, look at the way `"spy_events"` arm is implemented

bllu404: I had to use match guards. Let me know if it looks okay.

crates/cheatnet/src/state.rs
```diff
@@ -233,7 +247,31 @@ impl CheatnetState {
...
+    #[must_use]
+    pub fn get_address_block_timestamp(&self, address: &ContractAddress) -> Option<Felt252> {
+        /* The Logic
+         * - If there is a global warp and no contract warp, use the global warp
+         * - If there is a global warp and a contract warp (either Warped or Unwarped), use the contract warp
+         * - If there is no global warp and a contract warp, use the contract warp (either Warped or Unwarped)
+         * - If there is no global warp and no contract warp, return None (no warp)
+         *
+         * This logic ensures that `start_warp` and `stop_warp` will always behave as expected,
+         */
```
piotmag769, line 266: It can be simplified to `a specific contract being cheated takes precedence over the global warp`

crates/cheatnet/src/state.rs
```diff
@@ -233,7 +247,31 @@ impl CheatnetState {
...
+
+        match (&self.global_warp, self.warped_contracts.get(address)) {
+            (_, Some(warped_contract)) => match warped_contract {
+                WarpedContract::Warped(contract_timestamp) => Some(contract_timestamp.clone()),
+                WarpedContract::Unwarped => None,
+            },
+            (any_global, None) => any_global.clone(),
+        }
```
piotmag769, line 274: I think we can express the comment above with a bit more expressive code thus removing the need to comment entirely.
```suggestion
if let Some(warped_contract) = self.warped_contracts.get(address) {
    match warped_contract {
           WarpedContract::Warped(contract_timestamp) => Some(contract_timestamp.clone()),
            WarpedContract::Unwarped => None,
        },
} else {
    self.global_warp
}
```

crates/forge/src/test_execution_syscall_handler.rs
```diff
@@ -266,20 +266,54 @@ impl TestExecutionSyscallHandler<'_, '_> {
...
+                // The last element in `inputs` should be the timestamp in all cases
+                let warp_timestamp = inputs.last().unwrap().clone();
+
+                // First element encodes the variant of CheatTarget
+                let target = match &inputs[0] {
+                    x if *x == Felt252::from(0) => CheatTarget::All,
+                    x if *x == Felt252::from(1) => {
+                        CheatTarget::One(inputs[1].to_contract_address())
+                    }
+                    x if *x == Felt252::from(2) => {
```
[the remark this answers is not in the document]
bllu404: Doesn't work. The error is: `expected tuple struct or tuple variant, found associated function Felt252::from`

crates/cheatnet/src/state.rs
bllu404: I think this kind of comment might be a bit misleading because, while it is true in this context, in the greater context of how `snforge_std::start_warp` works, calling `start_warp::(CheatTarget::All,...)` overrides individual warps by clearing the hash map.

docs/src/appendix/cheatcodes/start_warp.md
```diff
@@ -1,13 +1,21 @@
...
+- `target` - target contract address(es).
 - `block_timestamp` - block timestamp to be set
 
+```rust
+enum CheatTarget {
+    All: (),
+    One: ContractAddress,
+    Multiple: Span<ContractAddress>
+}
+```
```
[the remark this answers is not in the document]
bllu404: I added it. I felt that the part explaining what each variant does was a bit redundant, let me know if it should be removed.

docs/src/appendix/cheatcodes/start_warp.md
```diff
@@ -31,19 +39,51 @@ impl IContractImpl of IContract<ContractState> {
...
-We can use `start_warp` in a test to change the block timestamp for a given contract:
+We can use `start_warp` in a test to change the block timestamp for contracts:
 
 ```rust
-use snforge_std::start_warp;
+use snforge_std::{start_warp, CheatTarget};
 
 #[test]
-fn test_warp() {
+fn test_warp_one() {
```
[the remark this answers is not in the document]
Arcticae: I think we should add an example in `cheat_target.md`

crates/cheatnet/tests/cheatcodes/warp.rs
```diff
@@ -318,3 +319,144 @@ fn warp_library_call() {
     .unwrap();
     assert_outputs(before_warp_output, after_warp_cancellation_output);
 }
+
+#[test]
+fn warp_all_simple() {
```
Arcticae, line 324: Please add a test with `multiple` as well
And an e2e test (or a few cases) with the actual cairo interface to test its' interface for errors/changes

crates/cheatnet/src/state.rs
```diff
@@ -202,12 +210,18 @@ impl StateReader for DictStateReader {
     }
 }
 
+pub enum WarpedContract {
+    Warped(Felt252),
+    Unwarped,
+}
```
Arcticae, line 216: Can we come up with a different name? We'll be reusing this probably

Arcticae (changes requested):
Also - please add the changes to the changelog.md so that users know that it's a breaking change

crates/forge/src/test_execution_syscall_handler.rs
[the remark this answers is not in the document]
piotmag769: This seems to be the best:
```rust
           let target = match inputs[0].to_u8() {
                    Some(0) => CheatTarget::All,
                    Some(1) => CheatTarget::One(inputs[1].to_contract_address()),
                    Some(2) => {
                        let contract_addresses: Vec<_> = inputs[2..inputs.len() - 1]
                            .iter()
                            .map(Felt252::to_contract_address)
                            .collect();
                        CheatTarget::Multiple(contract_addresses)
                    }
                    _ => panic!("Invalid CheatTarget variant"),
                };
```

docs/src/appendix/cheatcodes/start_warp.md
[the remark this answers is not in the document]
piotmag769: I would link the page in every page `CheatTarget` is used

docs/src/appendix/cheatcodes/start_warp.md
```diff
@@ -1,11 +1,11 @@
...
 
-> `fn start_warp(contract_address: ContractAddress, block_timestamp: u64)`
+> `fn start_warp(target: CheatTarget, block_timestamp: u64)`
 
-Changes the block timestamp for a contract at the given address.
+Changes the block timestamp for the given target.
 The change can be canceled with [`stop_warp`](./stop_warp.md).
 
-- `contract_address` - target contract address
+- `target` - target contract address(es). See [`CheatTarget`](./cheat_target.md).
```
piotmag769, line 8:
```suggestion
- `target` - instance of [`CheatTarget`](./cheat_target.md)  specifying which contracts to warp
```
We can reuse this pattern in `stop_warp`

crates/cheatnet/src/state.rs
bllu404: In that case would it have to be generic? As in, 
```rust
pub enum CheatedContract<T> {
   Cheated(T), 
   Uncheated,
}
```

Because some of the cheats (`prank`, `spoof`) require types other than Felt.

bllu404: Alternative name: `CheatInfo`. Wdyt?

docs/src/appendix/cheatcodes/cheat_target.md
```diff
@@ -0,0 +1,24 @@
...
+    Multiple: Span<ContractAddress>
+}
+```
+
+`CheatTarget` is an enum used to designate the contracts to which a warp should be applied. 
+- `All` applies the cheatcode to all contract addresses. 
+- `One` applies the cheatcode to the given contract address. 
+- `Multiple` applies the cheatcode to the given contract addresses. 
+
+
```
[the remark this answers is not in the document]
bllu404: Is this necessary when there is already an example with `start_warp` in `cheat_target.md`?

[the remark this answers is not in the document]
piotmag769: The comment is related to `cheat_target.md` though, wdym?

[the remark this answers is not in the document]
bllu404: Ahhh got it, I misunderstood what you were saying.

crates/forge/src/test_execution_syscall_handler.rs
```diff
@@ -268,20 +268,50 @@ impl TestExecutionSyscallHandler<'_, '_> {
...
+                    Some(1) => CheatTarget::One(inputs[1].to_contract_address()),
+                    Some(2) => {
+                        let contract_addresses: Vec<_> = inputs[2..]
+                            .iter()
+                            .map(Felt252::to_contract_address)
+                            .collect();
+                        CheatTarget::Multiple(contract_addresses)
+                    }
+                    _ => panic!("Invalid CheatTarget variant"),
+                };
```
piotmag769, line 308: Can be extracted to a function since it is reused

docs/src/appendix/cheatcodes/cheat_target.md
```diff
@@ -0,0 +1,37 @@
...
+}
+```
+
+`CheatTarget` is an enum used to designate the contracts to which a cheat should be applied. 
+- `All` applies the cheatcode to all contract addresses. 
+- `One` applies the cheatcode to the given contract address. 
+- `Multiple` applies the cheatcode to the given contract addresses. 
+
+
+An example with the [`start_warp`](./start_warp.md) cheatcode.
```
piotmag769, line 17: `:` instead of `.`

docs/src/appendix/cheatcodes/start_warp.md
```diff
@@ -31,19 +31,51 @@ impl IContractImpl of IContract<ContractState> {
...
 
-    start_warp(contract_address, 1000);
+    start_warp(CheatTarget::One(contract_address), 1000);
 
     dispatcher.set_block_timestamp();
     let new_timestamp = dispatcher.get_block_timestamp();
     assert(new_timestamp == 1000, 'Wrong timestamp');
 }
+
+#[test]
```
piotmag769, line 50: I would just remove those, it is well explained in the example in `cheat_target.md` I think

piotmag769 (commented):
Looking really good! 🔥

docs/src/appendix/cheatcodes/start_warp.md
Arcticae: Yes - please do remove those

Arcticae (changes requested):
Almost done - some cosmetic changes left :)

crates/forge/src/lib.rs
```diff
@@ -659,6 +662,22 @@ fn function_args<'a>(function: &'a Function, builtins: &[&str]) -> Vec<&'a Concr
...
+        Some(2) => {
+            let contract_addresses: Vec<_> = inputs[2..]
+                .iter()
+                .map(Felt252::to_contract_address)
+                .collect();
+            CheatTarget::Multiple(contract_addresses)
+        }
+        _ => panic!("Invalid CheatTarget variant"),
+    }
+}
```
piotmag769, line 679: Should be in `crates/forge/src/test_execution_syscall_handler.rs` 🙏

docs/src/appendix/cheatcodes/stop_warp.md
```diff
@@ -1,19 +1,19 @@
 # `stop_warp`
 
-> `fn stop_warp(contract_address: ContractAddress)`
+> `fn stop_warp(target: CheatTarget)`
 
-Cancels the [`start_warp`](./start_warp.md) for the contract at the given address.
+Cancels the [`start_warp`](./start_warp.md) for the given target.
 
-- `contract_address` - target contract address
+- `target` - target contract address(es)
```
piotmag769, line 7:
```suggestion
- `target` - instance of [`CheatTarget`](./cheat_target.md) specifying which contracts to warp.
```

docs/src/appendix/cheatcodes/start_warp.md
```diff
@@ -31,19 +31,21 @@ impl IContractImpl of IContract<ContractState> {
...
-    start_warp(contract_address, 1000);
+    start_warp(CheatTarget::One(contract_address), 1000);
 
     dispatcher.set_block_timestamp();
     let new_timestamp = dispatcher.get_block_timestamp();
     assert(new_timestamp == 1000, 'Wrong timestamp');
 }
 ```
+
+
```
piotmag769, line 51: Unrelated change

CHANGELOG.md
```diff
@@ -16,6 +16,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - diagnostic paths referring to `tests` folder
 - caching `get_class_hash_at` in forking test mode (credits to @jainkunal for catching the bug)
 
+#### Changed
+- `start_warp` and `stop_warp` now take `CheatTarget` as the first argument instead of `ContractAddress`. Read more [here](./docs/src/appendix/cheatcodes/start_warp.md). 
+
```
Arcticae, line 21: Before merging: move this section to the `Unreleased` segment

crates/cheatnet/tests/cheatcodes/warp.rs
```diff
@@ -318,3 +319,247 @@ fn warp_library_call() {
...
+    )
+    .unwrap();
+    let changed_back_block_timestamp = recover_data(output);
+
+    assert_eq!(old_block_timestamp, changed_back_block_timestamp);
+}
+
+#[test]
+fn warp_multiple() {
+    println!("blah blah blah");
```
Arcticae, line 466: Please delete this logging

Arcticae (commented):
LGTM
FYI: We will wait until 0.10 comes out (next release) before merging this though

docs/src/appendix/cheatcodes/start_warp.md
```diff
@@ -31,16 +31,16 @@ impl IContractImpl of IContract<ContractState> {
...
-We can use `start_warp` in a test to change the block timestamp for a given contract:
+We can use `start_warp` in a test to change the block timestamp for contracts:
 
 ```rust
-use snforge_std::start_warp;
+use snforge_std::{start_warp, CheatTarget};
 
 #[test]
-fn test_warp() {
+fn test_warp_one() {
```
piotmag769, line 40: `test_warp`

piotmag769 (approved):

piotmag769:
Really good job on that, thanks a lot 🙏 🚀

crates/cheatnet/src/cheatcodes/warp.rs
```diff
@@ -1,13 +1,45 @@
...
-        self.warped_contracts.remove(&contract_address);
+    pub fn stop_warp(&mut self, target: CheatTarget) {
+        match target {
+            CheatTarget::All => {
+                self.global_warp = None;
+                self.warped_contracts.clear();
+            }
+            CheatTarget::One(contract_address) => {
+                self.warped_contracts
+                    .insert(contract_address, CheatStatus::Uncheated);
```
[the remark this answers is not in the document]
piotmag769: Not really, if you cheat globally first and then uncheat locally you will have no way of knowing that the single contract is uncheated

docs/src/appendix/cheatcodes/cheat_target.md
```diff
@@ -0,0 +1,37 @@
+# 'CheatTarget`
+
+```rust
+enum CheatTarget {
+    All: (),
+    One: ContractAddress,
+    Multiple: Span<ContractAddress>
```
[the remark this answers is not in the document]
piotmag769: +1, we should be consistent here

crates/cheatnet/src/state.rs
```diff
@@ -233,7 +247,24 @@ impl CheatnetState {
...
+        self.global_warp.is_some()
+            || matches!(
+                self.warped_contracts.get(contract_address),
+                Some(CheatStatus::Cheated(_))
+            )
+    }
+
+    #[must_use]
+    pub fn get_cheated_block_timestamp(&self, address: &ContractAddress) -> Option<Felt252> {
+        // An individual warp for a contract overrides any global warp
```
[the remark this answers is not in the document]
piotmag769: It should be more like `warps from the self.warped_contract take priority over self.global_warp`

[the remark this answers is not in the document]
bllu404: > That's not always true

Yes...  I realized the way I wrote it was confusing but I wasn't sure how to phrase it better. What I actually meant is that the existence of a value for a given contract in the `warped_contracts` mapping will always override a global warp. 

But of course, calling `start_warp` or `stop_warp` with `All` will clear the mapping, which from the user's perspective looks like "overriding".

war-in (approved):
🔥 🥇

war-in:
@bllu404 could you merge the master branch? I think it will fix the failing test

Arcticae:
Closing in favour of #1116

PR #985, opened 2023-10-26
Syscall chaining

### LINKED ISSUE #916: Research implementing a pattern for chaining syscalls
It could involve a trait which would include middleware-like behavior for syscall handlers.
The result we're looking for is cleaner and more comprehensible code, where flow/order of the syscalls is well-defined.

Arcticae:
Closes #916

## Introduced changes

-
-

## Checklist

- [ ] Linked relevant issue
- [ ] Added relevant tests


crates/cheatnet/src/execution/syscall_interceptor.rs
```diff
@@ -0,0 +1,206 @@
...
+    pub hint_data: &'a Box<dyn Any>,
+    pub constants: &'a HashMap<String, Felt252>,
+}
+
+pub trait ChainableHintProcessor {
+    fn get_child(&self) -> Option<&dyn HintProcessorLogicInterceptor>;
+    fn get_child_mut(&mut self) -> Option<&mut dyn HintProcessorLogicInterceptor>;
+}
+
+pub trait HintExecutionInterceptor: ChainableHintProcessor {
```
MaksymilianDemitraszek, line 24: I wonder about it because `HintExecutionInterceptor` interface is kinda a part of the `HintProcessorExtension`, shouldn't we merge them into one trait?

crates/cheatnet/src/execution/syscall_interceptor.rs
```diff
@@ -0,0 +1,206 @@
...
+}
+
+pub struct CompileHintRequest<'a> {
+    pub hint_code: &'a str,
+    pub ap_tracking_data: &'a ApTracking,
+    pub reference_ids: &'a HashMap<String, usize>,
+    pub references: &'a [HintReference],
+}
+
+pub trait HintCompilationInterceptor: ChainableHintProcessor {
```
MaksymilianDemitraszek, line 70: I think instead of the abstract traitology with `HintCompilationInterceptor` and `ResourceTrackerInterceptor` we could just define custom procedural macro #derive instead? would make it less abstract thus simpler. Also probably for now we can write implementations by hand, and work on in a separate PR, so this one will remain reasonably sized.

crates/cheatnet/src/execution/syscall_interceptor.rs
```diff
@@ -0,0 +1,206 @@
...
+use std::collections::HashMap;
+
+pub struct ExecuteHintRequest<'a> {
+    pub vm: &'a mut VirtualMachine,
+    pub exec_scopes: &'a mut ExecutionScopes,
+    pub hint_data: &'a Box<dyn Any>,
+    pub constants: &'a HashMap<String, Felt252>,
+}
+
+pub trait ChainableHintProcessor {
```
MaksymilianDemitraszek, line 19: I am not sure if chainable is a good name, as chain suggest sequencial operations. Maybe something like HintProcessorExtension?

crates/cheatnet/src/execution/syscall_interceptor.rs
```diff
@@ -0,0 +1,206 @@
...
+    fn intercept_execute_hint(
+        &mut self,
+        execute_hint_request: &mut ExecuteHintRequest,
+    ) -> Option<Result<(), HintError>>;
+
+    fn next_execute_hint(
+        &mut self,
+        execute_hint_request: &mut ExecuteHintRequest,
+    ) -> Option<Result<(), HintError>> {
+        if let Some(result) = self.intercept_execute_hint(execute_hint_request) {
```
MaksymilianDemitraszek, line 34: Can we split execution and interception? Interception should return boolean instead and dont have access to self, to prevent any side effects by mistake. Then if true we should call logic from here.

crates/cheatnet/src/execution/syscall_interceptor.rs
```diff
@@ -0,0 +1,206 @@
...
+    ) -> Result<(), HintError> {
+        if let Some(result) = self.next_execute_hint(&mut ExecuteHintRequest {
+            vm,
+            exec_scopes,
+            hint_data,
+            constants,
+        }) {
+            return result;
+        }
+        unreachable!("execute_hint failed to terminate with a result")
```
MaksymilianDemitraszek, line 59: Why this branching is possible on the type level? If we separated Extension and main syscall handler we could gurantee on the type level that any hint will be handled and we could avoid unreachable branches.

MaksymilianDemitraszek (commented):
Leaving some comments and suggestions

crates/cheatnet/src/execution/syscall_interceptor.rs
Arcticae: Macro for making any interception logic "generic"? i don't think that's worth the effort in our case tbh, apart from being a rust programming exercise

crates/cheatnet/src/execution/syscall_interceptor.rs
Arcticae: I doubt that we'll be able to avoid this kind of panic/unreachable branch, since we want all the handlers to handle a subset of the hints

crates/cheatnet/src/execution/syscall_interceptor.rs
Arcticae: It doesn't make sense to me and would be less maintainable, as we'd have to:
1. Assess the hint type based on the request, and return whether it should be intercepted
2. Match based on the hint type, and redirect to execution (if intercepted)

So hint types would have to be synced in 1 and 2 match arms, which seems like a handful to keep in mind when modifying something

crates/forge/src/running.rs
```diff
@@ -220,16 +220,16 @@ pub(crate) fn run_test_case(
         ..Default::default()
     };
     let mut cheatable_syscall_handler =
-        CheatableSyscallHandler::wrap(syscall_handler, &mut cheatnet_state);
-    let contract_execution_syscall_handler =
+        CheatableSyscallHandler::wrap(&mut syscall_handler, &mut cheatnet_state);
```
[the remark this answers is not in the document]
Arcticae: That's not that simple, 
since in other file `cairo1_execution` we have the use of syscall handler as well with the `finalize_execution` function from blockifier. That function requires the ownership of the handler, thus it can't be moved to any other wrapper

crates/cheatnet/src/execution/syscall_interceptor.rs
Arcticae: I was worried that would make a mess of the trait, since it'd not provide clear separation between different interceptors for different traits.
Right now, there is one interceptor trait per one real trait from the `HintProcessorLogic` so that `HintProcessorLogic` interceptor is consistent in its' counterparts

crates/forge/src/running.rs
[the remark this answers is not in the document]
MaksymilianDemitraszek: I mean I don't think we need to transfer ownership for any reason here so its fine, I am just not sure why it is a problem? You can move out a field from struct with
```
let z = structure.a;
```

crates/cheatnet/src/execution/syscall_interceptor.rs
MaksymilianDemitraszek: Discussed offline

crates/cheatnet/src/execution/syscall_interceptor.rs
MaksymilianDemitraszek: Syscall Handlers handles the full set, so each hint should be eventually handled by the core handler.

crates/forge/src/running.rs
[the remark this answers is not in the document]
MaksymilianDemitraszek: Though, transferring the ownership would make possible to move some logic later to a functions `build_testing_syscall_handler` `build_contracts_sycall_handler`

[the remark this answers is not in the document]
MaksymilianDemitraszek: And you would avoid a b c d :D

PR #988, opened 2023-10-26
Allow disabling colors in forge output

### LINKED ISSUE #528: Allow disabling colors in forge output
Add `--color` flag to forge, similar to `cargo test` flag, that allows disabling or enabling coloring the output of `snforge.
It should have the following options:

- always - always use color
- never - never use color, disable all coloring

This flag should be added to `TestRunnerConfig` https://doc.rust-lang.org/nightly/unstable-book/compiler-flags/report-time.html 
Argument and necessary logic configuring if coloring is used should be added to `print_test_result` function https://github.com/foundry-rs/starknet-foundry/blob/ceaf6c507d04922aecafe8ac1efb684a029d34ea/crates/forge/src/pretty_printing.rs#L41 and passed accordingly to usages of it.

ptisserand:
Add `--color` option for test subcommand.

Closes #528

## Introduced changes

- `--color` option have been added to `test` subcomand to control when colored control is used.
- When `--color` is set to `never`, the environment variable [`CLICOLOR`](https://docs.rs/console/latest/console/fn.colors_enabled.html) will be set to `0`

## Checklist

- [x] Linked relevant issue
- [ ] Added relevant tests


ptisserand:
According to [console documentation](https://docs.rs/console/latest/console/fn.colors_enabled.html) , colors will be disabled when the program is piped, so I don't know how we can check the `--color never` behavior in CI.

crates/forge/src/main.rs
```diff
@@ -77,6 +83,9 @@ struct TestArgs {
     /// Run all tests regardless of `#[ignore]` attribute
     #[arg(long, conflicts_with = "only_ignored")]
     include_ignored: bool,
+
+    #[arg(value_enum, long, help = "Control when colored output is used", default_value_t = ColorOption::Always, value_name="WHEN")]
```
piotmag769, line 87: You can just a comment starting with `///`, it will be parsed by `clap` and then you can remove the `help` section

Also the better description would be 

`Control whether colored output is used`

crates/forge/src/main.rs
```diff
@@ -116,6 +125,9 @@ fn extract_failed_tests(tests_summaries: Vec<TestCrateSummary>) -> Vec<TestCaseS
 }
 
 fn test_workspace(args: TestArgs) -> Result<bool> {
+    if let ColorOption::Never = args.color {
+        env::set_var("CLICOLOR", "0");
+    }
```
piotmag769, line 130: I think we should also set `CLICOLOR_FORCE != 0` when `ColorOption::Always` is used. We can then introduce a `match` statement here.

I think it will also allow us to test this feature since `CLICOLOR_FORCE != 0` enforces the colored output no matter what

ptisserand: I agree that `always` should force the color output.

Maybe I should also add `auto` value to match `cargo test`:
```
--color when
    Control when colored output is used. Valid values:
        auto (default): Automatically detect if color support is available on the terminal.
        always: Always display colors.
        never: Never display colors.
```

What do you think about `auto` ?

piotmag769: I think we can skip the `auto` option, since it basically does nothing

ptisserand: > I think we can skip the `auto` option, since it basically does nothing

I think "do nothing" should be the default: as a user I expect that software by default should not change my colored output setting.

Here is my proposal:
```diff
-    if let ColorOption::Never = args.color {
-        env::set_var("CLICOLOR", "0");
+    match args.color {
+        ColorOption::Always => env::set_var("CLICOLOR_FORCE", "1"),
+        ColorOption::Never => env::set_var("CLICOLOR", "0"),
+        ColorOption::Auto => (),
     }
+

```

piotmag769: I forgot we set `Always` as default, I thought the arg is `Option<ColorOption>`. I think it is fine to add `Auto` then to be consistent with Cargo, it makes the code a bit clearer too I think since wrapping with `Option` adds additional complexity. Good call imo!
cc @cptartur

cptartur: > Here is my proposal:
> 
> ```diff
> -    if let ColorOption::Never = args.color {
> -        env::set_var("CLICOLOR", "0");
> +    match args.color {
> +        ColorOption::Always => env::set_var("CLICOLOR_FORCE", "1"),
> +        ColorOption::Never => env::set_var("CLICOLOR", "0"),
> +        ColorOption::Auto => (),
>      }
> +
> ```

Looks reasonable to me, we can do that.

crates/forge/src/main.rs
ptisserand: `Control when colored output is used` is used in [`cargo test` documentation](https://doc.rust-lang.org/cargo/commands/cargo-test.html).

Should I change the description ?

crates/forge/src/main.rs
ptisserand: `auto` option and test cases added.

ptisserand:
Branch rebased on master.

crates/forge/tests/e2e/color.rs
```diff
@@ -0,0 +1,32 @@
...
+use crate::e2e::common::runner::setup_package;
+
+fn runner_color(value: &str) -> SnapboxCommand {
+    SnapboxCommand::new(cargo_bin!("snforge"))
+        .arg("test")
+        .arg("--color")
+        .arg(value)
+}
+
+fn has_escape(output: &OutputAssert) -> bool {
```
piotmag769, line 12: `is_colored` is a better name I think

crates/forge/tests/e2e/color.rs
```diff
@@ -0,0 +1,32 @@
...
+        .unwrap()
+        .contains("\x1b[")
+}
+
+#[test]
+fn color_always() {
+    let temp = setup_package("simple_package");
+    let snapbox = runner_color("always");
+    let output = snapbox.current_dir(&temp).assert().code(1);
+    assert!(has_escape(&output), "Stdout doesn't contains escape code");
```
piotmag769, line 23: Better error message would be `"output expected to be colored but it is not"`. Analogous above

piotmag769 (approved):
Only small changes left, lgtm 🔥

crates/forge/tests/e2e/color.rs
ptisserand: I agree

crates/forge/tests/e2e/color.rs
ptisserand: Done

cptartur (approved):

ptisserand:
I have merged master and fixed the cargo fmt issue.

piotmag769:
@ptisserand thanks a lot for your valuable contribution! 🚀

Issue #953, opened 2023-10-24
Do not print skipped tests outputs


cptartur:
Skipped tests should only be shown in summary as number of skipped tests. Specific test cases shouldn't be reported as `SKIP`.

jobez:
can i be assigned this?

tkumor3:
@jobez it is resolved in #989

PR #989, opened 2023-10-27
Simplify cancelation logic in the context of parallel execution

tkumor3:
The main aid of this PR is to simplify cancelation logic. The previous version was base on the cancellation token. In order to optimize cases when tests could be skipped. Because tokio::select doesn't work with `spawn_blocking` logic with channels was added, to gently close the blocking task, after cancelled tokio::tasks. After deeper research, we realized that communication-based on channels and "closing" blocking spawn task is enough. 

Currently, if we use `--exit-first` flag, the failing test will close the channel, so all next spawn_blocking tasks will return `Skipped` instead of executing test

## Introduced changes

-
-

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/forge/src/lib.rs
```diff
@@ -578,72 +494,22 @@ fn run_with_fuzzing(
...
         )?;
 
+        if let Some(TestCaseSummary::Passed { .. }) = final_result {
+            if runs != fuzzer_runs {
+                final_result = results
+                    .iter()
+                    .cloned()
+                    .find(|item| matches!(item, TestCaseSummary::Skipped {}));
+            };
+        };
```
cptartur, line 504: We have the result extraction logic defined in couple different places, I think this can be simplified. 

1. If there is some `Failed` test, return `Failed`
2. Else if there are some skipped tests, return `Skipped`
3. Else if there are some passed tests, return `Passed`
4. Else if there's no result, panic

Is this flow missing any edge cases?

crates/forge/src/pretty_printing.rs
```diff
@@ -42,22 +42,24 @@ pub(crate) fn print_test_seed(seed: u64) {
 }
 
 pub(crate) fn print_test_result(test_result: &TestCaseSummary) {
+    if let TestCaseSummary::Skipped { .. } = test_result {
+        return;
+    }
```
cptartur, line 47: Isn't this resolving #953 as well?

crates/forge/src/lib.rs
```diff
@@ -290,22 +267,21 @@ pub async fn run(
...
-        .await?;
+        let summary =
+            run_tests_from_crate(compiled_test_crate, runner_config, runner_params).await?;
+
+        let interrupted = summary.interrupted;
 
         summaries.push(summary);
+        if interrupted {
+            break;
+        }
```
cptartur, line 284: Why was this introduced (an here specifically)?

crates/forge/src/lib.rs
tkumor3: 3. if there are only passed tests return `Passed`

crates/forge/src/pretty_printing.rs
tkumor3: It's resolve it! It was easier to remove skip before my changes

crates/forge/src/lib.rs
tkumor3: because we create cancellation channel for each test crate, and each test_crate is executed sync, so if we get info that one of test crate was interupted we can stop process all tests

crates/forge/src/pretty_printing.rs
cptartur: Okay, please add a comment in #953 and assign yourself, since there's already someone assigned there.

crates/forge/src/lib.rs
cptartur: Should we add a comment here?

crates/forge/src/lib.rs
MaksymilianDemitraszek: Why on the type level it is posssible to final result be none 🤔

tkumor3:
```
        let mut final_result = None;

        while let Some(task) = tasks.next().await {
            let result = task??;

            results.push(result.clone());
            final_result = Some(result.clone());

            if let TestCaseSummary::Failed { .. } = result {
                rec.close();
                break;
            }
        }
```
TBH, I have no idea, how to do it without None. If I remove the option, the compiler claims that None is possible.I think the reason is `tasks` could be empty

crates/forge/src/pretty_printing.rs
cptartur: Let's link this issue to this PR as well.

MaksymilianDemitraszek (changes requested):
Please split it into two separate PRs 🙏 It is hard to review otherwise because I don't know what relates to what.

crates/forge/src/running.rs
```diff
@@ -103,6 +138,10 @@ pub(crate) fn blocking_run_from_test(
             &send_shut_down,
         );
 
+        if send.is_closed() {
+            return Ok(TestCaseSummary::Skipped {});
+        }
```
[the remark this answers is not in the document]
tkumor3: To fix out tests, other wise no way to test --exit-fist flag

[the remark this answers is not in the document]
MaksymilianDemitraszek: This makes no sense, it basically means our tests describe a property poorly, because they don't test if execution time is reduced (which should be the goal of exit_first) but what is printed.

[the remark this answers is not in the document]
tkumor3: 👍 so I think that I will remove this code and ignore those tests. + Create an issue on proper testing of --exit-first flag.

crates/forge/src/lib.rs
```diff
@@ -404,17 +383,13 @@ async fn run_tests_from_crate(
...
-    // If TestCaseSummary::Interrupted is returned by a test,
-    // this implies that there should be another test than returned an Err.
-    assert!(!interrupted, "Tests were interrupted");
-
     let contained_fuzzed_tests = results.iter().any(|summary| summary.runs().is_some());
     Ok(TestCrateSummary {
         test_case_summaries: results,
         runner_exit_status: RunnerStatus::Default,
         test_crate_type: tests.tests_location,
         contained_fuzzed_tests,
```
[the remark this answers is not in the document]
tkumor3: Now issue has to be created

crates/forge/src/running.rs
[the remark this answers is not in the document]
cptartur: Testing based on execution time sounds like a bad idea. Most likely it will still fail periodically in case CI decides to slow down for whatever reason etc.

crates/forge/src/lib.rs
```diff
@@ -291,22 +268,25 @@ pub async fn run(
...
+        let interrupted = summary.interrupted;
 
         summaries.push(summary);
+
+        if interrupted {
+            // Handle scenario for --exit-first flag.
+            // Because snforge runs test crates one by one synchronously.
+            // In case of test FAIL with --exit-first flag stops processing the next crates
+            break;
+        }
```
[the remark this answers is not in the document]
tkumor3:
```
    if interrupted {
        Ok(TestCrateRunResult::Interrupted(TestCrateSummary {
            test_case_summaries: results,
            runner_exit_status: RunnerStatus::Default,
            test_crate_type: tests.tests_location,
            contained_fuzzed_tests,
        }))
    } else {
        Ok(TestCrateRunResult::OK(TestCrateSummary {
            test_case_summaries: results,
            runner_exit_status: RunnerStatus::Default,
            test_crate_type: tests.tests_location,
            contained_fuzzed_tests,
        }))
    }
    
```
end in run function 
```
        match summary {
            TestCrateRunResult::OK(result) => {
                summaries.push(summary);
            },
            TestCrateRunResult::Interrupted(_) => {
                summaries.push(summary);
                break;

            },
        }
```
I doesn't like this

crates/forge/src/running.rs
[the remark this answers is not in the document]
MaksymilianDemitraszek: @cptartur Disagreed, If you make one test run basically infinitely, and one run in a seconds it will always work. What matters is the stastistical significance of the time difference.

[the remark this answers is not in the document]
tkumor3: I created the [issue](https://github.com/orgs/foundry-rs/projects/3/views/2?pane=issue&itemId=43725188)

crates/forge/src/lib.rs
tkumor3: I have no idea, how to write it better each solution has some disadvantages

crates/forge/src/lib.rs
```diff
@@ -291,22 +269,27 @@ pub async fn run(
...
+                summaries.push(summary);
+            }
+            TestCrateRunResult::Interrupted(summary) => {
+                summaries.push(summary);
+                // Handle scenario for --exit-first flag.
+                // Because snforge runs test crates one by one synchronously.
+                // In case of test FAIL with --exit-first flag stops processing the next crates
+                break;
+            }
+        }
```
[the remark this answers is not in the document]
tkumor3: But I also have to unwrap `summary` from ` TestCrateRunResult`

crates/forge/src/lib.rs
```diff
@@ -421,67 +421,23 @@ fn choose_test_strategy_and_run(
...
     runner_config: Arc<RunnerConfig>,
     runner_params: Arc<RunnerParams>,
-    cancellation_tokens: Arc<CancellationTokens>,
+    send: Sender<()>,
     send_shut_down: Sender<()>,
 ) -> JoinHandle<Result<TestCaseSummary>> {
     tokio::task::spawn(async move {
-        let cancellation_fuzzing_token = CancellationToken::new();
-        let (send, mut rec) = channel(1);
+        let (fuzzing_send, mut fuzzing_rec) = channel(1);
```
piotmag769, line 440: I think we could check if `send.is_closed()` as the beginning of this function. Example: we have 100 fuzzing test functions, each running 1000 tests, all in `tests` folder (which means one crate). We use `--exit-first`, the first test fails. Then we have to wait for the 99 * 1000 tests to skip instead of returning `Skipped` here, which will happen 99 times instead of 99000 times. There is prob an exaggeration when it comes to number, it is about the idea. WDYT? cc @cptartur

piotmag769 (approved):

crates/forge/src/lib.rs
tkumor3: I added

crates/forge/src/running.rs
```diff
@@ -94,6 +93,42 @@ pub(crate) fn blocking_run_from_test(
...
+            &case,
+            &runner,
+            &runner_config,
+            &runner_params,
+            &send_shut_down,
+        );
+
+        if send.is_closed() {
+            return Ok(TestCaseSummary::Skipped {});
+        }
```
cptartur, line 107: This is duplicated now

crates/forge/src/running.rs
cptartur, line 143: Same here

crates/forge/src/running.rs
tkumor3: yes, but I had to add it to fix our --exit-first tests. We can remove it when we will change the way of testing it.

crates/forge/src/running.rs
```diff
@@ -94,6 +93,45 @@ pub(crate) fn blocking_run_from_test(
...
+            &case,
+            &runner,
+            &runner_config,
+            &runner_params,
+            &send_shut_down,
+        );
+
+        // TODO: code below is added to fix snforge tests
+        // remove it after improve exit-first tests
+        // issue #1043
```
cptartur, line 107: Let's mention it in the #1043 description as well.

cptartur (approved):

PR #997, opened 2023-10-27
Remove breaking changes section from pull request template

cptartur:
Closes #

## Introduced changes

-Removed breaking changes section: In most cases it's left empty even if it's relevant to the PR. It doesn't make sense to keep it if it's not being enforced.

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -9,10 +9,6 @@ Closes #
 -
 -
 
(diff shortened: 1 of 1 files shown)


MaksymilianDemitraszek (approved):
I think we should get back to it after we stop be in alpha.

Issue #839, opened 2023-10-10
Split forge crate


MaksymilianDemitraszek:
`forge` crate encapsulates to many responsibilities

We should split it into `cli` and `test-runner`

cptartur:
I think we should just split `lib.cairo` into smaller modules. I don't see any benefit of making it entriely separate crates.

MaksymilianDemitraszek:
Disagreed, It is better to keep responsibilities of crates small

cptartur:
I just don't see how this split would go exactly. The CLI part is mostly just resolving things from scarb which is a separate module already, or just some code in `main.rs`. Also, the running code is also printing outputs so there's no clear separation between these.

This split can be done, but in my opinion it would require significant refactor (#175 should probably be done after all).

MaksymilianDemitraszek:
I think this loop (and everything below it) should go into test runnet package 
https://github.com/foundry-rs/starknet-foundry/blob/1b897056179545d37f7b9f8b13d665d952d31459/crates/forge/src/lib.rs#L290

There is no need for refactor, printing can be part of the test running as well, cli responsibility should be define user interface, fetch args from different soruces and consolidate them.

MaksymilianDemitraszek:
test collector is also thin, I don't see it as an major issue. I am not sure what would be the benefits of printing refactor if there are some but I don't also understand the goal you think about clearly.

MaksymilianDemitraszek:
I see, not sure if we have any benefits of such control right now, would be useful if we consider implementing machine readable output. One way or another I don't see it blocking or conflicing suggested crate split in any what. those are separate concerns

PR #1007, opened 2023-10-30
Print latest block number and improve error handling

### LINKED ISSUE #757: Add printing used block id in fork when `latest` to allow reproducing

piotmag769:
Closes #757 

## Introduced changes

- added printing block number when `BlockId::Tag(Latest)` is specified
- improved error handling
- added `unwrap` in places where the error will be only our fault

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/forge/src/running.rs
```diff
@@ -229,35 +234,42 @@ pub(crate) fn run_test_case(
...
+        Err(RunnerError::CairoRunError(error)) => Ok(TestCaseSummary::Failed {
+            name: case.name.clone(),
+            msg: Some(format!(
+                "\n    {}\n",
+                error.to_string().replace(" Custom Hint Error: ", "\n    ")
+            )),
+            arguments: args,
+            fuzzing_statistic: None,
+        }),
+        Err(err) => Err(err.into()),
```
piotmag769, line 255: Those are other variants of [`RunnerError`](https://github.com/starkware-libs/cairo/blob/main/crates/cairo-lang-runner/src/lib.rs#L47), I don't think they should ever happen but I'm not sure. I decided to bubble an error, it kills the whole runner anyways but with a nice error message

crates/forge/src/running.rs
```diff
@@ -229,35 +234,42 @@ pub(crate) fn run_test_case(
         &string_to_hint,
     );
 
-    Ok(runner.run_function(
+    match runner.run_function(
```
cptartur, line 237: We're doing `find_function` again here even though we have `func` variable available.

crates/forge/src/running.rs
```diff
@@ -229,35 +234,42 @@ pub(crate) fn run_test_case(
...
-    )?)
+    ) {
+        Ok(result) => Ok(TestCaseSummary::from_run_result(result, case, args)),
+        // CairoRunError comes from VirtualMachineError which may come from HintException that originates in the TestSyscallExecutionHandler
+        Err(RunnerError::CairoRunError(error)) => Ok(TestCaseSummary::Failed {
+            name: case.name.clone(),
+            msg: Some(format!(
+                "\n    {}\n",
+                error.to_string().replace(" Custom Hint Error: ", "\n    ")
+            )),
```
cptartur, line 251: I know we had this code for a while now, but what in case the error is something different than `Custom Hint Error`. Shouldn't we kill the runner as well, as we do with other errors?

crates/forge/src/running.rs
```diff
@@ -229,35 +234,42 @@ pub(crate) fn run_test_case(
...
+            arguments: args,
+            fuzzing_statistic: None,
+        }),
+        Err(err) => Err(err.into()),
+    }
 }
 
 fn extract_test_case_summary(
```
cptartur, line 259: That's a weird flow to me. `TestCaseSummary` is already extracted in this function argument. We should have this handling in one place, there's no benefit to wrapping the error in two different places.

cptartur (changes requested):

crates/forge/src/running.rs
piotmag769: The reason for that change is to be able to add  additional info to `TestCaseSummary`, e.g. number of the block used (that's why I did this PR, see description), if we got to `run_function` call . It is a compromise between doing weird `match` statements in every place `?` is used in `run_test_case` and passing additional info to `TestCrateSummary`

crates/forge/src/running.rs
piotmag769: Not really, for example when reading from a file and the file does not exists this logic does not apply

crates/forge/src/running.rs
cptartur: I don't understand why cannot the arms of match in `run_test_case` and in `extract_test_case_summary` be combined. I'd rather have a one large match handling all errors than artificially splitting it into two functions.

crates/forge/src/running.rs
```diff
@@ -291,12 +303,15 @@ fn get_fork_state_reader(
...
+fn get_latest_block_number(
+    url: &Url,
+) -> Result<BlockId, ProviderError<JsonRpcClientError<HttpTransportError>>> {
     let client = JsonRpcClient::new(HttpTransport::new(url.clone()));
     let runtime = Runtime::new().expect("Could not instantiate Runtime");
 
     match runtime.block_on(client.get_block_with_tx_hashes(BlockId::Tag(Latest))) {
         Ok(MaybePendingBlockWithTxHashes::Block(block)) => Ok(BlockId::Number(block.block_number)),
-        _ => Err(anyhow!("Could not get the latest block number".to_string())),
+        Ok(MaybePendingBlockWithTxHashes::PendingBlock(_)) => unreachable!(),
```
Arcticae, line 314: This is out of scope

crates/forge/src/running.rs
```diff
@@ -229,35 +234,42 @@ pub(crate) fn run_test_case(
...
-    Ok(runner.run_function(
+    match runner.run_function(
         runner.find_function(case.name.as_str())?,
         &mut test_execution_syscall_handler,
         hints_dict,
         instructions,
         builtins,
-    )?)
+    ) {
+        Ok(result) => Ok(TestCaseSummary::from_run_result(result, case, args)),
```
Arcticae, line 244: Why move the parsing here? It makes whole thing a lot more confusing

Arcticae (changes requested):

crates/forge/src/running.rs
piotmag769: Will change it, but it requires introducing additional intermediate struct. I need to be able to pass block number with a `RunResult` somehow

crates/forge/src/running.rs
piotmag769: It is not, before we returned a general error that would kill the whole runner. Now I want to return a specific error that occurs on rpc call, so I have to add this variant to the match statement to be able to get the error arm below. It is impossible for us to ever get `PendingBlock` here.

crates/forge/src/running.rs
piotmag769: > I generally don't understand those classes of errors. What is RunnerError? What is StateError?

External libraries errors, decided to remove them and just make sure everything that fails in `run_from_test` results in failed test (does not kill the runner)

crates/forge/src/running.rs
piotmag769: > I don't think running.rs should know anything about TestCaseSummary (stuff like fuzzing statistics and block number) it should be handled in an upper layer (cli layer). We should differentiate between TestCaseResult and summary and don't leak this stuff here as this feeds the complexity.

@MaksymilianDemitraszek out of scope of this PR, should be done with #839

crates/forge/src/running.rs
```diff
@@ -151,18 +151,12 @@ fn build_syscall_handler<'a>(
...
-pub(crate) enum TestCaseRunError {
-    #[error("{0}")]
-    RunnerError(#[from] RunnerError),
-    #[error("{0}")]
-    StateError(#[from] StateError),
-    #[error("{0}")]
-    Error(#[from] anyhow::Error),
+pub(crate) struct RunResultWithInfo {
+    pub(crate) run_result: Result<RunResult, RunnerError>,
+    pub(crate) block_number_of_latest: Option<BlockNumber>,
```
Arcticae, line 159:
```suggestion
    pub(crate) latest_block_number: Option<BlockNumber>,
```
I know that this isn't exactly the same semantically, but wouldn't it make it more readable?

Arcticae (changes requested):

crates/forge/src/test_case_summary.rs
```diff
@@ -23,6 +24,8 @@ pub enum TestCaseSummary {
         arguments: Vec<Felt252>,
         /// Statistic for fuzzing test
         fuzzing_statistic: Option<FuzzingStatistics>,
+        /// Number of block used if BlockId::Tag(Latest) was specified
+        block_number_of_latest: Option<BlockNumber>,
```
cptartur, line 28: I think we should follow the suit of `FuzzingStatistics` and introduce a struct for a forking parameters (e.g. name it `ForkConfigUsed`) and store this details there.

cptartur: Perhaps we could always store the block number and have a separate boolean flag for if `latest` block was used. Could make the logic simpler.

piotmag769: Out of scope: #1049

crates/forge/src/running.rs
```diff
@@ -229,46 +227,66 @@ pub(crate) fn run_test_case(
...
-        runner.find_function(case.name.as_str())?,
+    let block_number_of_latest = if let Some(ValidatedForkConfig {
+        url: _,
+        block_id: BlockId::Tag(Latest),
+    }) = &case.fork_config
+    {
+        Some(block_info.block_number)
+    } else {
+        None
+    };
```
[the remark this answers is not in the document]
piotmag769: Out of scope: #1049

crates/forge/src/test_case_summary.rs
piotmag769: Agree, but imo `Option` expresses it better for now

piotmag769: I would leave it for now and wrap it in a struct when additional info will be needed to not introduce unnecessary abstraction layer

Arcticae (approved):

cptartur (approved):

Issue #675, opened 2023-09-15
Bump `scarb metadata` for cast and hoist it to common `Cargo.toml`


piotmag769:
Currently the version is fixed to `1.4.2`, some breaking changes happened.

PR #1008, opened 2023-10-31
Refactor `get_contracts_map`

### LINKED ISSUE #1005: Refactor `get_contracts_map` to take `Metadata` and `PackageId`
I think this function should just take `Metadata` and `PackageId` so we don't leak abstraction like in forge main (see below). Also the description is incorrect - path is not a path to scarb package, it is a path to `{target_name}.starknet_artifacts.json` file

https://github.com/foundry-rs/starknet-foundry/blob/master/crates/forge/src/main.rs#L168-L175

_Originally posted by @piotmag769 in https://github.com/foundry-rs/starknet-foundry/pull/957#discussion_r1371573537_

drknzz:
Closes #1005

## Introduced changes

- change input of **get_contracts_map**:
  - `Utf8Path` path to *{target_name}.starknet_artifacts.json*  --->  `Metadata` and `PackageId`

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/cast/Cargo.toml
```diff
@@ -16,7 +16,7 @@ url.workspace = true
 rand.workspace = true
 project-root.workspace = true
 which.workspace = true
-scarb-metadata = "=1.4.2"
+scarb-metadata = { git = "https://github.com/software-mansion/scarb.git", rev = "58cc88e" }
```
THenry14, line 19: I'd prefer this included changes suggested in https://github.com/foundry-rs/starknet-foundry/issues/675 instead

MaksymilianDemitraszek: Why do we need to set it to commit instead of version?

MaksymilianDemitraszek (changes requested):
Please split version bump and refactor as a separate PRs 🙏 . Such PRs are harder to review, merge quickly and create messy history. If we for example detected a regression later, we wouldn't be able to find out quickly if it happend from version bump or we messed somenting up doing refactor.

crates/cast/Cargo.toml
piotmag769: @MaksymilianDemitraszek we don't need to anymore, we needed it before with Scarb 0.7.0 bcs scarb-ui was not published on crates.io and is dependent on metadata. As @THenry14 says, #675 is related to that

drknzz: Done in #1019

piotmag769 (approved):

THenry14 (approved):

crates/scarb-artifacts/src/lib.rs
```diff
@@ -84,25 +84,34 @@ pub fn try_get_starknet_artifacts_path(
...
+                .ok_or_else(|| anyhow!("Failed to get parent for path = {}", &contracts_path))?;
+            let artifacts = artifacts_for_package(&contracts_path)?;
+            let mut map = HashMap::new();
+            for contract in artifacts.contracts {
+                let name = contract.contract_name;
+                let sierra_path = base_path.join(contract.artifacts.sierra);
+                let casm_path = base_path.join(contract.artifacts.casm);
+                let sierra = fs::read_to_string(sierra_path)?;
+                let casm = fs::read_to_string(casm_path)?;
+                map.insert(name, StarknetContractArtifacts { sierra, casm });
```
MaksymilianDemitraszek, line 109: Please add method `StarknetContractArtifacts::from_scarb_contract_artifact(StarknetContract)` I think it is the only way we read it, we can abstract this away.

Please move the whole logic under this pattern to `load_contract_artifacts` function, right now you need to read the whole code to understand what it does and cognitive complexity is high.

MaksymilianDemitraszek (approved):
LGTM, Please address the requested changes before merging 🙏 🥺

PR #1018, opened 2023-10-31
remove TestCrateSummary::Interrupted and don't print Skipped

tkumor3:
Closes #

## Introduced changes

-
-

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/forge/src/test_case_summary.rs
```diff
@@ -41,17 +41,7 @@ pub enum TestCaseSummary {
         name: String,
     },
     /// Test case skipped due to exit first
```
cptartur, line 43: Is this only possible reason for `Skipped` status now?

tkumor3: In general, it's returned when we do not care about the test result, in the case when there is no test that returns Err we use it to know how many tests was skipped

cptartur: I'd update this comment then

MaksymilianDemitraszek (approved):
Please update the comment about skipped before merging.

PR #1030, opened 2023-11-02
Move running related code to separate crate

### LINKED ISSUE #839: Split forge crate

cptartur:
Closes #839 

## Introduced changes

- Split forge into two crates. Note that this PR will not pass checks as it is incomplete. See related PRs linked in comments for continuation of this work.

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


cptartur:
Continuation: 
#1032 
#1046

crates/forge-runner/src/fuzzer.rs
```diff
@@ -1,8 +1,9 @@
-use crate::fuzzer::arguments::CairoType;
 use anyhow::Result;
 use rand::rngs::StdRng;
 use rand::Rng;
 
+use crate::fuzzer::arguments::CairoType;
+
```
Arcticae, line 6: plz revert this one

crates/forge-runner/src/lib.rs
Arcticae, line 1: Sidenote: This whole file encapsulates so many different things that it's hard to grasp what's really happening here, i would strongly advise refactoring this

crates/forge-runner/src/test_crate_summary.rs
```diff
@@ -8,28 +8,31 @@ pub struct TestCrateSummary {
     pub test_case_summaries: Vec<TestCaseSummary>,
     /// Status of the runner after executing tests in the file
     pub runner_exit_status: RunnerStatus,
-    /// Type of the test crate
-    pub test_crate_type: CrateLocation,
+    // /// Type of the test crate
+    // pub test_crate_type: CrateLocation,
```
Arcticae, line 12: Either delete or uncomment this

crates/forge-runner/src/test_execution_syscall_handler.rs
```diff
@@ -542,9 +542,9 @@ impl TestExecutionSyscallHandler<'_, '_> {
                         write_cheatcode_panic(&mut buffer, &panic_data);
                         Ok(())
                     }
-                    CallContractResult::Failure(CallContractFailure::Error { msg }) => Err(
-                        EnhancedHintError::from(HintError::CustomHint(Box::from(msg))),
-                    ),
+                    CallContractResult::Failure(CallContractFailure::Error { msg }) => {
+                        Err(EnhancedHintError::from(CustomHint(Box::from(msg))))
+                    }
```
Arcticae, line 547:
```suggestion
                    CallContractResult::Failure(CallContractFailure::Error { msg }) => 
                        Err(EnhancedHintError::from(CustomHint(Box::from(msg))))

```

crates/forge-runner/src/test_execution_syscall_handler.rs
Arcticae, line 1: All those changes are unrelated

crates/forge-runner/src/lib.rs
```diff
@@ -0,0 +1,594 @@
...
+        let error = CancellationToken::new();
+        Self { exit_first, error }
+    }
+}
+
+impl Default for CancellationTokens {
+    fn default() -> Self {
+        Self::new()
+    }
+}
```
piotmag769, line 168: To be rebased after we get rid of them xD

crates/forge-runner/src/lib.rs
```diff
@@ -0,0 +1,594 @@
...
+        Self {
+            sierra_program,
+            test_cases,
+        }
+    }
+}
+
+pub trait TestCaseFilter {
+    fn should_be_run(&self, test_case: &TestCase) -> bool;
+}
```
piotmag769, line 227: Everything from the beginning of the file to this line should be moved to separate files

crates/forge-runner/src/test_crate_summary.rs
piotmag769: This should stay uncommented for sure

crates/forge/src/pretty_printing.rs
```diff
@@ -41,57 +42,6 @@ pub(crate) fn print_test_seed(seed: u64) {
     println!("{}: {seed}", style("Fuzzer seed").bold());
 }
 
-pub(crate) fn print_test_result(test_result: &TestCaseSummary) {
```
piotmag769, line 44: What happened with that?

crates/forge-runner/src/test_case_summary.rs
```diff
@@ -124,7 +127,7 @@ impl TestCaseSummary {
     #[must_use]
     pub(crate) fn from_run_result_and_info(
         run_result: RunResult,
-        test_case: &TestCaseRunnable,
+        test_case: &TestCase,
```
piotmag769, line 130: It should stay the way it was

crates/forge-runner/src/lib.rs
```diff
@@ -0,0 +1,594 @@
...
+        Ok(ForkConfig {
+            url: value.url.parse()?,
+            block_id: value.block_id,
+        })
+    }
+}
+
+impl ForkConfigTrait for ForkConfig {}
+
+pub type TestCase = CollectedTestCase<ForkConfig>;
```
piotmag769, line 207: Randomly changed name, making it hard to review

crates/forge-runner/src/lib.rs
```diff
@@ -0,0 +1,594 @@
...
+
+    function
+        .signature
+        .param_types
+        .iter()
+        .filter(|pt| !builtins.contains(&pt.debug_name))
+        .collect()
+}
+
+fn print_test_result(test_result: &TestCaseSummary) {
```
piotmag769, line 545: Why is it here?

crates/forge-runner/src/lib.rs
```diff
@@ -0,0 +1,594 @@
...
+use futures::StreamExt;
+use once_cell::sync::Lazy;
+use scarb_artifacts::StarknetContractArtifacts;
+use smol_str::SmolStr;
+use starknet::core::types::BlockId;
+use std::collections::HashMap;
+use std::sync::Arc;
+use test_collector::FuzzerConfig;
+use test_collector::TestCase as CollectedTestCase;
+use test_collector::{ForkConfig as ForkConfigTrait, LinkedLibrary, RawForkParams};
```
piotmag769, line 24: Those aliases make it unreadable, please remove them. If you want to rename them do it in separate PR

crates/forge-runner/src/lib.rs
```diff
@@ -0,0 +1,594 @@
...
+    /// Runner exited without problems
+    Default,
+    /// Some test failed
+    TestFailed,
+    /// Runner did not run, e.g. when test cases got skipped
+    DidNotRun,
+}
+
+#[derive(Debug, Clone)]
+pub struct ForkConfig {
```
piotmag769, line 182: randomly changed name

piotmag769 (commented):
Please revert the name changes, it creates a lot of mess when moving code between files

crates/forge-runner/src/test_crate_summary.rs
cptartur: We no longer pass CrateLocation to forge_runner, because it's irrelevant for running tests, only for printing statistics, which is handled in a different way. Removing this.

crates/forge-runner/src/test_execution_syscall_handler.rs
```diff
@@ -825,9 +825,7 @@ fn write_call_contract_response(
                     .map(StarknetConversions::to_stark_felt)
                     .collect(),
             },
-            CallContractFailure::Error { msg, .. } => {
-                return Err(HintError::CustomHint(Box::from(msg)))
-            }
+            CallContractFailure::Error { msg, .. } => return Err(CustomHint(Box::from(msg))),
```
cptartur, line 828: This change is introduced by `cargo fmt`

crates/forge-runner/src/lib.rs
cptartur: It's either these aliases or introducing weird names to structs defined in this file

crates/forge-runner/src/lib.rs
cptartur: In a separate PR

crates/forge-runner/src/lib.rs
cptartur: I can either copy this code to this file or copy the whole `pretty_printing` to `forge-runner` which doesn't make much sense, because we cannot have a cyclic dependency between `forge-runner` and `forge` crates.

Ideally we'd move it to a separate crate or introduce some different abstraction, but that's for future PRs.

crates/forge-runner/src/lib.rs
```diff
@@ -0,0 +1,491 @@
...
+    #[must_use]
+    pub fn name(&self) -> &str {
+        &self.name
+    }
+
+    #[must_use]
+    pub fn params(&self) -> &RawForkParams {
+        &self.params
+    }
+}
```
Arcticae, line 70: This impl is not really useful, can't we just mark the fields `pub` and avoid adding `new` method if it doesn't wrap any logic/add any context?

crates/forge-runner/src/lib.rs
```diff
@@ -0,0 +1,491 @@
...
+    }
+    #[must_use]
+    pub fn corelib_path(&self) -> &Utf8PathBuf {
+        &self.corelib_path
+    }
+}
+
+impl RunnerParams {
+    #[must_use]
+    pub fn new(
```
Arcticae, line 130: Same as above

crates/forge-runner/src/lib.rs
```diff
@@ -0,0 +1,491 @@
...
+
+#[derive(Debug, Clone)]
+pub struct TestCrate {
+    sierra_program: Program,
+    test_cases: Vec<TestCaseRunnable>,
+}
+
+impl TestCrate {
+    #[must_use]
+    pub fn new(sierra_program: Program, test_cases: Vec<TestCaseRunnable>) -> Self {
```
Arcticae, line 194: same as above

crates/forge-runner/src/lib.rs
Arcticae: It should be at least in a separate file

Arcticae (changes requested):

crates/forge-runner/src/lib.rs
cptartur: #1091

crates/forge-runner/src/lib.rs
cptartur: I want to keep fields of this struct and other completely private/inaccessible, but we still have to allow constructing instances of it.

crates/forge-runner/src/lib.rs
cptartur: Moved

crates/forge/src/lib.rs
```diff
@@ -298,265 +164,26 @@ pub async fn run(
...
-        let mocked_tests = CompiledTestCrate {
+        let mocked_tests = CompiledTestCrateRaw {
             sierra_program: Program {
                 type_declarations: vec![],
                 libfunc_declarations: vec![],
                 statements: vec![],
                 funcs: vec![],
             },
-            test_cases: vec![TestCase {
+            test_cases: vec![TestCase::<RawForkConfig> {
```
piotmag769, line 186: Is it needed?

crates/forge/src/lib.rs
```diff
@@ -298,265 +164,26 @@ pub async fn run(
...
+    use forge_runner::ForkTarget;
     use starknet::core::types::BlockId;
     use starknet::core::types::BlockTag::Latest;
-    use test_collector::ExpectedTestResult;
+    use test_collector::{ExpectedTestResult, TestCase};
 
     #[test]
     fn to_runnable_unparsable_url() {
-        let mocked_tests = CompiledTestCrate {
+        let mocked_tests = CompiledTestCrateRaw {
```
piotmag769, line 179: Is it needed?

crates/forge/src/lib.rs
```diff
@@ -243,11 +113,6 @@ pub async fn run(
     let not_filtered: usize = test_crates.iter().map(|tc| tc.test_cases.len()).sum();
     let filtered = all_tests - not_filtered;
 
-    let test_crates = test_crates
-        .into_iter()
-        .map(|ctc| to_runnable(ctc, &runner_config))
-        .collect::<Result<Vec<_>>>()?;
```
piotmag769, line 249: Unrelated + not needed imo

crates/forge/src/lib.rs
```diff
@@ -167,32 +38,32 @@ fn replace_id_with_params(
...
+            Ok(fork_target_from_runner_config.params().clone())
         }
     }
 }
 
 fn to_runnable(
     compiled_test_crate: CompiledTestCrateRaw,
     runner_config: &RunnerConfig,
-) -> Result<CompiledTestCrateRunnable> {
+) -> Result<TestCrate> {
```
piotmag769, line 54: Why?

crates/forge-runner/src/lib.rs
```diff
@@ -0,0 +1,442 @@
...
+        })
+    }
+}
+
+impl ForkConfig for ValidatedForkConfig {}
+
+pub type TestCaseRunnable = TestCase<ValidatedForkConfig>;
+
+#[derive(Debug, Clone)]
+pub struct TestCrate {
```
piotmag769, line 189: Can we keep the previous name? It is not actually a `TestCrate` itself, it has test configs too

crates/forge/src/lib.rs
cptartur: It's needed: The `TestCrate` struct used by forge-runner doesn't have the crate location unnecessary since it's unused there at all. We need to first call `print_running_tests` and only then convert to runnable.

crates/forge-runner/src/lib.rs
cptartur: It doesn't make sense to name it `CompiledTestCrateRunnable` if it's the only variant of this struct present in forge-runner. It's no longer related to struct from the test collector except for the similar contents.

crates/forge/src/lib.rs
cptartur: The forge-runner now owns the type it runs and it's named like that, so it had to be changed.

crates/forge-runner/src/lib.rs
Arcticae: Let's discuss offline

Arcticae (approved):

crates/forge-runner/src/lib.rs
piotmag769: I disagree, when I browse the code following the reference (cmd + leftclick) I won't check every time in what package I am, it introduces additional complexity

crates/forge/src/lib.rs
piotmag769: Okay, makes sense

crates/forge-runner/src/lib.rs
cptartur: We can use `forge_runner::TestCrate` syntax instead of importing it directly if that's more readable. I don't think we names in other crates should heavily influence the names in this one, that's kind of agains the whole point of "namespacing" our code with different crates.

crates/forge-runner/src/lib.rs
cptartur: #1123

crates/forge-runner/src/lib.rs
```diff
@@ -0,0 +1,437 @@
...
+                rec.close();
+            }
+        }
+
+        results.push(result);
+    }
+
+    // Waiting for things to finish shutting down
+    drop(send_shut_down);
+    let _ = rec_shut_down.recv().await;
```
tkumor3, line 284: I think this logic is not required anymore. We wait for all tasks in prev loop - I will create ticket to remove and test it

crates/forge-runner/src/lib.rs
```diff
@@ -0,0 +1,437 @@
+use crate::fuzzer::RandomFuzzer;
```
tkumor3, line 1: This file is messy, it has +400 lines and is related to many things. IMO we should split it

crates/forge-runner/src/lib.rs
piotmag769: I understand your point, but I think it just adds cognitive complexity for no reason. Those crates are really tightly coupled so keeping names unique across both of them makes a lot of sense imo

piotmag769 (approved):

crates/forge-runner/src/lib.rs
MaksymilianDemitraszek: Agree with @piotmag769 here, those are important information

crates/forge-runner/src/lib.rs
cptartur: #1091 

It will be done but in a separate PR

crates/forge-runner/src/lib.rs
cptartur: These arguments are valid in general case, though I think that we'd never name this struct `CompiledTestCrateRunnable` or whatever if we had two separate crates from the very beginning. 

This name was introduced just because we had multiple versions of a similar struct and needed to differentiate between them which is no longer the case. 

If you look at the usages of it in the code, it's pretty much constructed in one place in forge and all other usages of it are in the forge-runner, where the name doesn't matter that much.

Issue #275, opened 2023-07-25
`snfoundryup` for updating snfoundry


THenry14:
We should think about adding some mechanism, to simply update forge and cast binaries.
One approach could be to copy `rustup` behaviour.

kontrol-apa:
Hello, i would like to take over if this is still open. I am thinking about modifying the foundryup from [foundry-rs](https://github.com/foundry-rs/foundry/tree/master/foundryup) to provide the same functionality

kontrol-apa:
@THenry14 this is addressed by https://github.com/foundry-rs/starknet-foundry/pull/1041

PR #1041, opened 2023-11-06
Adding snfoundryup support 🔨

partychad:
Resolves: [Issue #275](https://github.com/foundry-rs/starknet-foundry/issues/275)

## Introduced changes

* Added a new script under `scripts/install.sh` which installs `snfoundryup` script under `$HOME/.local/bin` and adds the path to the profile file of the associated shell. `snfoundryup` can be used to update starknet foundry binaries.

* Added a new CLI option to snfoundryup, which allows to build the binaries from any commit hash with `snfoundryup --commit <COMMIT HASH>`

* Added a new banner to snfoundryup script for cosmetic purposes,because why not :)
image

* Minor fixes and improvements
  * Added progress bar to curl commands for better visualization
  * Encapsulated snforge and sncast under BINARIES variable
  * The installation directory is determined by the resolved version rather than the requested version. This approach avoids using ambiguous directory names like "latest."
  
  
  ## Testing
  
  Since the download URL `SNFOUNDRYUP_URL="https://raw.githubusercontent.com/foundry-rs/starknet-foundry/master/scripts/snfoundryup"` isnt a part of the master branch yet, changes can be tested by replacing `line 11` in `scripts/install.sh` with `SNFOUNDRYUP_URL="https://raw.githubusercontent.com/partychad/starknet-foundry/fork-master/scripts/snfoundryup"` so that snfoundryup can be downloaded.

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -4,11 +4,9 @@ set -e
 echo Installing snfoundryup...
 
 
-REPO="https://github.com/foundry-rs/starknet-foundry"
 XDG_DATA_HOME="${XDG_DATA_HOME:-"${HOME}/.local/share"}"
 INSTALL_ROOT="${XDG_DATA_HOME}/starknet-foundry-install"
 LOCAL_BIN="${HOME}/.local/bin"
-LOCAL_BIN_ESCAPED="\$HOME/.local/bin"
 
 SNFOUNDRYUP_URL="https://raw.githubusercontent.com/partychad/starknet-foundry/fork-master/scripts/snfoundryup"
 SNFOUNDRYUP_PATH="${LOCAL_BIN}/snfoundryup"
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -8,7 +8,7 @@ XDG_DATA_HOME="${XDG_DATA_HOME:-"${HOME}/.local/share"}"
 INSTALL_ROOT="${XDG_DATA_HOME}/starknet-foundry-install"
 LOCAL_BIN="${HOME}/.local/bin"
 
-SNFOUNDRYUP_URL="https://raw.githubusercontent.com/partychad/starknet-foundry/fork-master/scripts/snfoundryup"
+SNFOUNDRYUP_URL="https://raw.githubusercontent.com/foundry-rs/starknet-foundry/master/scripts/snfoundryup"
 SNFOUNDRYUP_PATH="${LOCAL_BIN}/snfoundryup"
 
 # Check for curl
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -3,9 +3,6 @@ set -e
 
 echo Installing snfoundryup...
 
-
-XDG_DATA_HOME="${XDG_DATA_HOME:-"${HOME}/.local/share"}"
-INSTALL_ROOT="${XDG_DATA_HOME}/starknet-foundry-install"
 LOCAL_BIN="${HOME}/.local/bin"
 
 SNFOUNDRYUP_URL="https://raw.githubusercontent.com/foundry-rs/starknet-foundry/master/scripts/snfoundryup"
@@ -19,9 +16,9 @@ fi
 
 
 # Create the ${HOME}/.local/bin bin directory and snfoundryup binary if it doesn't exist.
-mkdir -p ${LOCAL_BIN}
-curl -# -L ${SNFOUNDRYUP_URL} -o ${SNFOUNDRYUP_PATH}
-chmod +x ${SNFOUNDRYUP_PATH}
+mkdir -p "${LOCAL_BIN}"
+curl -# -L "${SNFOUNDRYUP_URL}" -o "${SNFOUNDRYUP_PATH}"
+chmod +x "${SNFOUNDRYUP_PATH}"
 
 
 # Store the correct profile file (i.e. .profile for bash or .zshenv for ZSH).
@@ -54,5 +51,6 @@ if [ ":$PATH:" != *":${LOCAL_BIN}:"* ]; then
 fi
 
 
-echo && echo "Detected your preferred shell is ${PREF_SHELL} and added snfoundryup to PATH. Run 'source ${PROFILE}' or start a new terminal session to use snfoundryup."
-echo "Then, simply run 'snfoundryup' to install Starknet-Foundry."
+printf "\nDetected your preferred shell is %s and added snfoundryup to PATH. Run 'source %s' or start a new terminal session to use snfoundryup.\n" "$PREF_SHELL" "$PROFILE"
+printf "Then, simply run 'snfoundryup' to install Starknet-Foundry.\n"
+
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -12,10 +12,11 @@ SNFOUNDRYUP_URL="https://raw.githubusercontent.com/foundry-rs/starknet-foundry/m
 SNFOUNDRYUP_PATH="${LOCAL_BIN}/snfoundryup"
 
 # Check for curl
-if ! command -v curl &> /dev/null; then
+if ! command -v curl > /dev/null 2>&1; then
     echo "curl could not be found, please install it first."
     exit 1
 fi
+
 
 # Create the ${HOME}/.local/bin bin directory and snfoundryup binary if it doesn't exist.
 mkdir -p ${LOCAL_BIN}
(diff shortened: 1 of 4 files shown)


Arcticae (commented):
Just tested it locally and it seems to work for me, stellar work!
Please add it to our docs as well 🙏

kontrol-apa:
@Arcticae updated the gitbook as well

scripts/install.sh
```diff
@@ -1,490 +1,57 @@
...
 
-# This is just for indicating that commands' results are being intentionally ignored.
-# Usually, because it's being executed as part of error handling.
-ignore() {
-  "$@"
-}
+# Create the ${HOME}/.local/bin bin directory and snfoundryup binary if it doesn't exist.
+mkdir -p ${LOCAL_BIN}
+curl -# -L ${SNFOUNDRYUP_URL} -o ${SNFOUNDRYUP_PATH}
+chmod +x ${SNFOUNDRYUP_PATH}
```
THenry14, line 23: I know these are quite safe, but let's quote the variable names just to preserve good practices :+1:

scripts/install.sh
```diff
@@ -1,490 +1,57 @@
...
 
-  echo >>"$_profile" && echo "export PATH=\"\$PATH:${LOCAL_BIN_ESCAPED}\"" >>"$_profile"
-  echo \
-    "Detected your preferred shell is ${_pref_shell} and added '${LOCAL_BIN_ESCAPED}' to PATH." \
-    "Run 'source ${_profile}' or start a new terminal session to use Starknet Foundry."
-}
 
-main "$@" || exit 1
+echo && echo "Detected your preferred shell is ${PREF_SHELL} and added snfoundryup to PATH. Run 'source ${PROFILE}' or start a new terminal session to use snfoundryup."
+echo "Then, simply run 'snfoundryup' to install Starknet-Foundry."
```
THenry14, line 57: You could use printf instead, to avoid multiple "echo"es

scripts/snfoundryup
```diff
@@ -0,0 +1,559 @@
...
+# Most of this code is based on/copy-pasted from rustup and protostar installers.
+
+set -u
+
+REPO="https://github.com/foundry-rs/starknet-foundry"
+XDG_DATA_HOME="${XDG_DATA_HOME:-"${HOME}/.local/share"}"
+INSTALL_ROOT="${XDG_DATA_HOME}/starknet-foundry-install"
+LOCAL_BIN="${HOME}/.local/bin"
+LOCAL_BIN_ESCAPED="\$HOME/.local/bin"
+BINARIES="snforge sncast" # Array synthax for shells. List can be expanded for new binaries
```
THenry14, line 21:
```suggestion
BINARIES="snforge sncast" # Array syntax for shells. List can be expanded for new binaries
```

scripts/snfoundryup
```diff
@@ -0,0 +1,559 @@
...
+    v)
+      _requested_ref="tag/v${OPTARG}"
+      _requested_version="$OPTARG"
+      ;;
+    c)
+      need_cmd cargo
+      banner
+      _commit_hash="$OPTARG"
+       clone_and_build "$_commit_hash"
+       exit 0
```
THenry14, line 83:
```suggestion
      need_cmd cargo
      banner
      _commit_hash="$OPTARG"
      clone_and_build "$_commit_hash"
      exit 0
```

scripts/snfoundryup
```diff
@@ -0,0 +1,559 @@
...
+clone_and_build() {
+  local _commit_hash=$1
+  local _tempdir
+
+  if ! _tempdir="$(ensure mktemp -d)"; then
+    exit 1
+  fi
+
+  say "Cloning repository at commit $_commit_hash..."
+  ensure git clone "$REPO" $_tempdir
```
THenry14, line 510:
```suggestion
  ensure git clone "$REPO" "$_tempdir"
```

scripts/snfoundryup
```diff
@@ -0,0 +1,559 @@
...
+  assert_nz "$_installdir" "installdir"
+
+  ensure cp "$_build_output_dir/snforge" "$_installdir/bin/snforge"
+  ensure cp "$_build_output_dir/sncast" "$_installdir/bin/sncast"
+
+  create_symlinks "$_installdir"
+
+}
+
+banner() {
```
THenry14, line 531: That's cool :D but I'm not sure I'm sold on the idea of having this, probably not. @Arcticae thoughts?

THenry14 (commented):
Great contribution 💪

scripts/snfoundryup
```diff
@@ -0,0 +1,559 @@
...
+  local _commit_hash=$1
+  local _tempdir
+
+  if ! _tempdir="$(ensure mktemp -d)"; then
+    exit 1
+  fi
+
+  say "Cloning repository at commit $_commit_hash..."
+  ensure git clone "$REPO" $_tempdir
+  ensure cd "$_tempdir"
```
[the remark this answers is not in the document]
partychad: pushd and popd are sadly not part of sh shell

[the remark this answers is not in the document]
THenry14: Ahh, you're right, I forgot about it. Sorry for the confusion then.

Still maybe it would be a good idea to return to the previous directory in this function (eg `cd -` ?) I know that this function is used only at the end of `c` case for building from commit hash so it does not impact anything the moment, but I'm worried it might in the future when someone wants to use this function for something else (and the directories are mixed up). wdyt?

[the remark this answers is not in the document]
partychad: I think your description is the right way to handle it, there maybe unwanted side effects. Adding a `cd -` to the end now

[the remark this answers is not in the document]
partychad: updated [here](https://github.com/foundry-rs/starknet-foundry/pull/1041/commits/355011f6986699701b300e380a779c8e6f424389)

THenry14:
Hi @partychad , it looks good, but can you address my previous, unresolved comments first? 🙏 after this I will do the next round of review. Thanks!

partychad:
> Hi @partychad , it looks good, but can you address my previous, unresolved comments first? 🙏 after this I will do the next round of review. Thanks!

I have no idea how i missed those comments, working on it 🫡

scripts/snfoundryup
Arcticae: I asked the team, and we're generally against since its' not unix-like behavior

partychad: removed the banner from the script

Arcticae (approved):
LGTM
cc @THenry14

scripts/install.sh
```diff
@@ -1,490 +1,56 @@
...
-    ensure ln -fs "$_binary" "$_symlink"
-    ensure chmod u+x "$_symlink"
-    say "created symlink ${_symlink} -> ${_binary}"
-  done
-}
+# Only add snfoundryup if it isn't already in PATH.
+if [ ":$PATH:" != *":${LOCAL_BIN}:"* ]; then
+    # Add the snfoundryup directory to the path and ensure the old PATH variables remain.
+    echo >> "$PROFILE" && echo "export PATH=\"\$PATH:$LOCAL_BIN\"" >> "$PROFILE"
+fi
```
THenry14, line 51: I think this might get executed each time this script is being called, as (if I remember correctly) test command (`[ ... ]`) cannot match globs. 

edit: yes, it looks that way:
```
wojciech@mbp ~/Tests $ ➜ echo $PATH | grep .local
(...):/Users/wojciech/.local/bin

wojciech@mbp ~/Tests $ ➜ echo $0
-zsh

wojciech@mbp ~/Tests $ ➜ cat ~/.zshenv
. "$HOME/.cargo/env"

export PATH="$PATH:/Users/wojciech/.local/bin"

wojciech@mbp ~/Tests $ ➜ ./install.sh
Installing snfoundryup...
################################################################################################################################################################################################################################################### 100.0%

Detected your preferred shell is zsh and added snfoundryup to PATH. Run 'source /Users/wojciech/.zshenv' or start a new terminal session to use snfoundryup.
Then, simply run 'snfoundryup' to install Starknet-Foundry.

wojciech@mbp ~/Tests $ ➜ cat ~/.zshenv
. "$HOME/.cargo/env"

export PATH="$PATH:/Users/wojciech/.local/bin"

export PATH="$PATH:/Users/wojciech/.local/bin"
```

Can you fix please? 🙏

THenry14 (changes requested):

scripts/install.sh
partychad: good catch, should be fixed now.

THenry14 (approved):
looks good, thanks for the contribution 💪

Issue #436, opened 2023-08-08
Improve cast tests (creating directories)


war-in:
In some cast tests, we are creating directories, but often we are not using tempdirs. As a result, junk folders remain after a failed test run. 
https://github.com/foundry-rs/starknet-foundry/blob/c04585092890f3cd351b36b411a4dae651c35feb/crates/cast/tests/helpers/fixtures.rs#L173
https://github.com/foundry-rs/starknet-foundry/blob/c04585092890f3cd351b36b411a4dae651c35feb/crates/cast/tests/e2e/account/create.rs#L134
https://github.com/foundry-rs/starknet-foundry/blob/c04585092890f3cd351b36b411a4dae651c35feb/crates/cast/tests/e2e/account/deploy.rs#L103

We should always use tempdirs when dynamically creating directories. `tempfile` or `tempdir` crates can be used for this.

partychad:
If this issue is still available, i would love to work on it :)

war-in:
@THenry14 ? 🙏

THenry14:
Sure, assigned :+1:

PR #1064, opened 2023-11-07
Improve cast tests 

partychad:
Resolves: [Issue #436](https://github.com/foundry-rs/starknet-foundry/issues/436)

## Introduced changes

Started using TempDir crate to handle directory creation. Due to it's RAI nature the folder lifetime is tied to file object. To keep files persistent through the function calls, `duplicate_directory_with_salt` function signature is changed to return `TempDir` rather than `String`.

Two additional helper functions have been implemented to obtain the paths for the accounts and keystores directories. Previously, since these directories were created on the fly, relative paths sufficed. However, we have now transitioned to utilizing two new functions that provide paths relative to the cargo.toml file within the sncast crate.

* `pub fn get_accounts_path(relative_path_from_cargo_toml: &str) -> String `
* `pub fn get_keystores_path(relative_path_from_cargo_toml: &str) -> String`

The modifications ensure that the lifespan of the folders is linked to specific objects, which are automatically cleaned up even if tests do not pass. This approach helps to maintain a clean source tree by avoiding the accumulation of temporary files. As the temporary folders are placed in the /tmp directory, (for Unix-like systems) the operating system can handle any necessary cleanup as a secondary measure, should the primary cleanup process fail.

### DIFF
--- a/crates/cast/tests/e2e/account/add.rs
+++ b/crates/cast/tests/e2e/account/add.rs
@@ -86,7 +86,7 @@ pub async fn test_happy_case_add_profile() {
     ];
 
     let snapbox = Command::new(cargo_bin!("sncast"))
-        .current_dir(&current_dir.path())
+        .current_dir(current_dir.path())
         .args(args);
 
     snapbox.assert().stdout_matches(indoc! {r#"
--- a/crates/cast/tests/e2e/account/add.rs
+++ b/crates/cast/tests/e2e/account/add.rs
@@ -58,11 +58,11 @@ pub async fn test_happy_case() {
 
 #[tokio::test]
 pub async fn test_happy_case_add_profile() {
-    let current_dir = Utf8PathBuf::from_path_buf(duplicate_directory_with_salt(
+    let current_dir = duplicate_directory_with_salt(
         CONTRACTS_DIR.to_string() + "/map",
         "put",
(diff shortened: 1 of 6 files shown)


crates/cast/tests/e2e/account/add.rs
```diff
@@ -88,15 +86,16 @@ pub async fn test_happy_case_add_profile() {
...
         .args(args);
 
     snapbox.assert().stdout_matches(indoc! {r#"
         command: account add
         add_profile: Profile successfully added to Scarb.toml
     "#});
-
-    let mut file = current_dir.clone();
+    let current_dir_utf8 =
+        Utf8PathBuf::from_path_buf(current_dir.into_path()).expect("Path contains invalid UTF-8"); // current_dir is consumed but the tmp folder is not deleted, it will persist until a reboot
```
THenry14, line 97: I appreciate it, but let's maybe obstain from unnecessary comments in the code :) 
```suggestion
        Utf8PathBuf::from_path_buf(current_dir.into_path()).expect("Path contains invalid UTF-8");
```

THenry14 (commented):
Great job 💪 Let's fix linter problems reported by CI, and this will be mergeable :+1:

THenry14 (approved):

PR #1067, opened 2023-11-07
Deployment scripts: add `sncast_std` with the `call` command

### LINKED ISSUE #820: Importable call subcommand
Make calling contracts using cast subcomand possible from deployment scripts. 

Consult cairo deployment scripts [design doc](https://github.com/foundry-rs/starknet-foundry/blob/master/design_documents/cairo_deployment_scripts.md) for more context.

abulenok:
Closes #820 

## Introduced changes

- Add deployment scripts with `call` subcommand callable from script.
- Add `sncast script` command.
- Add `sncast_std` scarb package.

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/cast/tests/data/scripts/hello_world/hello_world.cairo
```diff
@@ -0,0 +1,65 @@
...
+};
+
+#[derive(Drop, Clone)]
+struct CallResult {
+    data: Array::<felt252>,
+}
+
+fn call(
+    contract_address: ContractAddress, function_name: felt252, calldata: Array::<felt252>
+) -> CallResult {
```
abulenok, line 23: In the next iteration, the `call` function will become importable from a separate module + more tests for deployment scripts with `call` will be added.

crates/cast/tests/data/scripts/hello_world/hello_world.cairo
```diff
@@ -0,0 +1,65 @@
...
+    let mut i = 0;
+    loop {
+        if i == calldata_len {
+            break;
+        }
+        inputs.append(*calldata[i]);
+        i += 1;
+    };
+
+    let buf = cheatcode::<'call'>(inputs.span());
```
THenry14, line 39: do you mind explaining to me how will this look like after call is moved from separate module? 

ideally we would not have to call `cheatcode::<'call'>` but rather (at least) hide it behind some abstraction like just `let buf = call(...)`, but I'm still new to all those cheatcodes logic so perhaps it's just not possible and I just don't know something.

THenry14:
@cptartur @Radinyn as people with some experience with chetcodes in forge, do you mind checkout out this PR as well? 🙏

crates/cast/tests/data/scripts/hello_world/hello_world.cairo
abulenok: It will work like this: https://github.com/foundry-rs/starknet-foundry/pull/1111/files#diff-6e5eba2cddaafbe74f218b4df832f8f280665ccfe3352ff14a1362f188fb575eR1-R13

THenry14: Great, thanks! Will it be problematic to have multiple scarb packages in one repo? Maybe we should think about moving this elsewhere?

crates/cast/src/helpers/scarb_utils.rs
```diff
@@ -69,18 +70,47 @@ pub fn get_scarb_manifest() -> Result<Utf8PathBuf> {
     Ok(path)
 }
 
-pub fn get_scarb_metadata(manifest_path: &Utf8PathBuf) -> Result<scarb_metadata::Metadata> {
+pub fn get_scarb_metadata(
+    manifest_path: &Utf8PathBuf,
+    with_deps: bool,
```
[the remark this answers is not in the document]
abulenok: like `Option<bool>` ?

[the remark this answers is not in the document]
THenry14: yes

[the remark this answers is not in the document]
Radinyn: not sure what's the difference between `None` and `false` in that case

crates/cast/src/helpers/scarb_utils.rs
```diff
@@ -69,18 +70,47 @@ pub fn get_scarb_manifest() -> Result<Utf8PathBuf> {
...
+        "Failed to read Scarb.toml manifest file, not found in current nor parent directories, {}",
+        env::current_dir()
+            .unwrap()
+            .into_os_string()
+            .into_string()
+            .unwrap()
+    ))
+}
+
+pub fn parse_scarb_metadata(path: &Option<Utf8PathBuf>) -> Result<Metadata> {
```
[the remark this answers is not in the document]
Radinyn: I'm pretty sure my Scarb PR will remove or at least rework this function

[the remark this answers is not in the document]
THenry14: 1. I think this function was added in this PR
2. this link ponts to Bartek's PR :D

crates/cast/src/main.rs
```diff
@@ -321,6 +350,7 @@ async fn main() -> Result<()> {
             print_command_result("show-config", &mut result, value_format, cli.json)?;
             Ok(())
         }
+        Commands::Script(_) => Err(anyhow!("All cases should already be handled")),
```
Radinyn, line 353: Is this what you mean?
```suggestion
        Commands::Script(_) => unreachable!(),
```

crates/cast/src/script.rs
Radinyn, line 1: I wonder if one day we could somehow export this logic to some separate crate, it was problematic in Forge as well. Obviously not in the scope of this PR

crates/cast/src/helpers/scarb_utils.rs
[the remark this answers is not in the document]
Radinyn: 1. I see, it just worked similarly to something else I saw somewhere. Either way in the new approach I'm planning to get the metadata once at the top and then pass it down to everywhere it is needed.
2. Fixed xD

crates/cast/tests/data/scripts/hello_world/hello_world.cairo
abulenok: As per @maciektr, it shouldn't be a problem.

abulenok: Scarb searches for a package that matches the name provided in Scarb.toml

crates/cast/src/helpers/scarb_utils.rs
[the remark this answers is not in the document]
cptartur: I'd assume we'd treat `None` as `false` in the code anyways so no point in changing it in my opinion.

crates/cast/src/helpers/scarb_utils.rs
```diff
@@ -69,18 +70,47 @@ pub fn get_scarb_manifest() -> Result<Utf8PathBuf> {
...
+    if let Some(path) = path {
+        if !path.exists() {
+            bail!("{path} file does not exist!");
+        }
+    }
+
+    let manifest_path = match path.clone() {
+        Some(path) => path,
+        None => get_scarb_manifest().context("Failed to obtain manifest path from scarb")?,
+    };
```
cptartur, line 107: Could these be combined? Quite weird to split this handling to me

crates/cast/src/lib.rs
```diff
@@ -350,6 +350,26 @@ pub fn print_formatted(output: Vec<(&str, String)>, json: bool, error: bool) ->
     Ok(())
 }
 
+fn json_value_to_string(value: &Value, value_format: ValueFormat) -> Option<String> {
```
cptartur, line 353: Are we sure there's no some handling for it built-in in serde?

crates/cast/tests/data/scripts/hello_world/hello_world.cairo
THenry14: perfect, thanks guys!

crates/cast/src/main.rs
cptartur, line 353: Why don't we call `script::run` here?

crates/cast/src/script.rs
```diff
@@ -0,0 +1,360 @@
...
+
+#[derive(Args)]
+#[command(about = "")]
+pub struct Script {
+    /// Path to the script
+    pub script_path: Utf8PathBuf,
+}
+
+pub struct CairoHintProcessor<'a> {
+    pub blockifier_syscall_handler: SyscallHintProcessor<'a>,
```
cptartur, line 66: Do we need blockfier for these scripts at all?

crates/cast/src/script.rs
```diff
@@ -0,0 +1,360 @@
...
+                exec_scopes,
+                selector,
+                input_start,
+                input_end,
+                output_start,
+                output_end,
+            );
+        }
+        self.blockifier_syscall_handler
+            .execute_hint(vm, exec_scopes, hint_data, constants)
```
cptartur, line 131: I'm not sure about how this should be handled exactly, but I'm not sure if whole dependency on blockifier is needed for just running deployment scripts.

Perhaps https://github.com/starkware-libs/cairo/blob/1c70f6459d1f38018da4397466ba2dccf6d48f81/crates/cairo-lang-runner/src/casm_run/mod.rs#L406 from casm-run would be enough?

cptartur: cc @MaksymilianDemitraszek @Arcticae

crates/cast/src/script.rs
```diff
@@ -0,0 +1,360 @@
...
+        input_start: &ResOperand,
+        input_end: &ResOperand,
+        output_start: &CellRef,
+        output_end: &CellRef,
+    ) -> Result<(), HintError> {
+        // Parse the selector.
+        let selector = &selector.value.to_bytes_be().1;
+        let selector = std::str::from_utf8(selector).map_err(|_| {
+            HintError::CustomHint(Box::from(
+                "Failed to parse the  cheatcode selector".to_string(),
```
cptartur, line 162:
```suggestion
                "Failed to parse the cheatcode selector".to_string(),
```

crates/cast/src/script.rs
```diff
@@ -0,0 +1,360 @@
...
+            ))
+        })?;
+
+        // Extract the inputs.
+        let input_start = extract_relocatable(vm, input_start)?;
+        let input_end = extract_relocatable(vm, input_end)?;
+        let inputs = vm_get_range(vm, input_start, input_end).map_err(|_| {
+            HintError::CustomHint(Box::from("Failed to read input data".to_string()))
+        })?;
```
cptartur, line 171: This code for error handling is duplicated in couple of places. A helper or some refactor to not do that would be useful.

crates/cast/src/script.rs
cptartur, line 162: In some cases we return hinterror and in some other we use `expect/unwrap`. This is a bit inconsistent.

crates/cast/src/script.rs
```diff
@@ -0,0 +1,360 @@
...
+        .to_option()
+        .with_context(|| "Compilation failed without any diagnostics.")?;
+    let replacer = DebugReplacer { db };
+
+    let runner = SierraCasmRunner::new(
+        replacer.apply(&sierra_program),
+        Some(MetadataComputationConfig::default()),
+        OrderedHashMap::default(),
+    )
+    .with_context(|| "Failed setting up runner.")?;
```
cptartur, line 273: Why is it needed

crates/cast/src/script.rs
```diff
@@ -0,0 +1,360 @@
...
+    script_path: &Utf8PathBuf,
+    provider: &JsonRpcClient<HttpTransport>,
+    runtime: Runtime,
+    path_to_scarb_toml: &Option<Utf8PathBuf>,
+) -> Result<ScriptResponse> {
+    check_compiler_path(true, Path::new(&script_path))?;
+
+    let metadata = parse_scarb_metadata(path_to_scarb_toml)?;
+
+    let corelib = corelib_for_package(&metadata, &metadata.workspace.members[0])?;
```
cptartur, line 252: To me, this function has too many responsibilities. These should be resolved in the higher level and injected as arguments.

crates/cast/src/script.rs
```diff
@@ -0,0 +1,360 @@
...
+    let replacer = DebugReplacer { db };
+
+    let runner = SierraCasmRunner::new(
+        replacer.apply(&sierra_program),
+        Some(MetadataComputationConfig::default()),
+        OrderedHashMap::default(),
+    )
+    .with_context(|| "Failed setting up runner.")?;
+
+    let func = runner.find_function("::main")?;
```
cptartur, line 275: Why `::` before function name?

crates/cast/src/script.rs
```diff
@@ -0,0 +1,360 @@
...
+
+    let mut cairo_hint_processor = CairoHintProcessor {
+        blockifier_syscall_handler: syscall_handler,
+        hints: &string_to_hint,
+        provider,
+        runtime,
+    };
+
+    match runner.run_function(
+        runner.find_function("::main")?,
```
cptartur, line 322: We already have this function in `func` variable.

crates/cast/src/script.rs
```diff
@@ -0,0 +1,360 @@
...
+                msg: build_readable_text(&data),
+            }),
+            RunResultValue::Panic(panic_data) => Ok(ScriptResponse {
+                status: "script panicked".to_string(),
+                msg: build_readable_text(&panic_data),
+            }),
+        },
+        Err(err) => Err(err.into()),
+    }
+}
```
cptartur, line 340: This could be multiple smaller functions.

crates/cast/src/script.rs
```diff
@@ -0,0 +1,360 @@
...
+        }
+    }
+
+    if readable_text.is_empty() {
+        None
+    } else {
+        readable_text.push('\n');
+        Some(readable_text)
+    }
+}
```
cptartur, line 360: Instead of duplicating it, we could consider moving it to a separate common crate in another PR.

crates/cast/src/script.rs
maciektr: Cairo-run does the same - ensure it's not some_long_name_ending_with_main I guess

crates/cast/src/helpers/scarb_utils.rs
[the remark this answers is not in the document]
THenry14: My reasoning was that we could get rid of writing additional argument `false` everywhere in our current codebase (at the time of writing this comment at least, every place we used this had `false` as argument). 

This is not a crucial change, but I suppose it would be nice to just be able to get scarb metadata with one argument (manifest) and be non breaking in regards to the previous interface.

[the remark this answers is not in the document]
abulenok: But we would still have to pass `None` instead of `false`.  Like `get_scarb_metadata(metadata, None)`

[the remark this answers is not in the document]
abulenok: What we can do though, is create two additional functions `get_scarb_metadata_with_deps` and `get_scarb_metadata_without_deps`, which would call `get_scarb_metadata(metadata, with_deps)` under the hood. This should improve readability.

CHANGELOG.md
```diff
@@ -13,6 +13,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - `available_gas` attribute, it didn't compute correctly gas usage. Contract functions execution cost would not be included.
 
+### Cast
+
+#### Added
+
+- added deployment scripts with `call` subcommand. The script can be launched with `sncast script path/to/script.cairo`
+
```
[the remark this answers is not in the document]
abulenok: agree

crates/cast/src/helpers/scarb_utils.rs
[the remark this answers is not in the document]
THenry14: I leave it up to you to decide, you can change it the way you've described or as it currently is then

crates/cast/src/main.rs
```diff
@@ -110,7 +116,30 @@ async fn main() -> Result<()> {
...
+    }
+}
 
+#[allow(clippy::too_many_lines)]
+async fn run_async_command(
+    cli: Cli,
+    mut config: CastConfig,
+    provider: JsonRpcClient<HttpTransport>,
+    value_format: ValueFormat,
+) -> Result<()> {
```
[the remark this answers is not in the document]
abulenok: To make custom cheatcodes work we need to implement the `HintProcessorLogic` trait with its method `execute_hint`. This method needs to call async `Cast` commands, while not being async by itself. Hence, we cannot do `await` in it and need to utilize the `runtime.block_on` method, which cannot be used inside async tasks. 

And if we call a non-async method from async code, that non-async method is still inside the asynchronous context [(source)](https://docs.rs/tokio/latest/tokio/task/index.html#blocking-and-yielding)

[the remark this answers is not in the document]
abulenok: So changing `main` back to async, will execute non-async functions `script::run() -> runner.run_function() -> execute_hint()` within the same async context

crates/cast/src/lib.rs
```diff
@@ -350,6 +350,26 @@ pub fn print_formatted(output: Vec<(&str, String)>, json: bool, error: bool) ->
...
+                .filter_map(|item| json_value_to_string(item, value_format))
+                .collect::<Vec<String>>()
+                .join(", ");
+            format!("[{arr_as_string}]")
+        }
+        _ => return None,
+    };
+    Some(res)
+}
+
```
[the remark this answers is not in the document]
abulenok: I removed the `res` variable.
As for the other points:

- there are no ownership issues with `value_format` since it implements the `Copy` trait
- `n.as_u64()` we may want this to `panic`, so we know that invalid values are encountered and we have a problem somewhere in our code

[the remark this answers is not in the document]
THenry14: cool, thanks! as for:

> n.as_u64() we may want this to panic

we definitely want avoid panic-ing, we should rather display informative and concise error messages

crates/cast/Cargo.toml
```diff
@@ -25,6 +25,57 @@ promptly = "0.3.1"
 scarb-artifacts = { path = "../scarb-artifacts" }
 # Freeze reqwest dependency, because newer versions cause cast tests to fail on macs
 reqwest = "=0.11.18"
+assert_fs.workspace = true
```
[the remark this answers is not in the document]
abulenok: Removed unused deps. (used `cargo machete` to find them)

crates/cast/src/lib.rs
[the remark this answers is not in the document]
abulenok: If we use `map`, it will not print anything
Do you suggest using a different error message in `expect`?

crates/cast/Cargo.toml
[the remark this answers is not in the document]
abulenok: And refactored others to use those from workspace

crates/cast/src/main.rs
abulenok: Discussed here https://github.com/foundry-rs/starknet-foundry/pull/1067#discussion_r1392461269

crates/cast/src/lib.rs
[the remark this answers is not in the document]
THenry14: yes, expect should be enough here

crates/cast/src/script.rs
abulenok: All function names are of the form `module_name::function_name`

And `find_function` returns the first function that matches the provided suffix. So passing just "main" will result in `find_function` returning any function that ends with "main"

crates/cast/src/script.rs
abulenok: fixed

crates/cast/src/script.rs
abulenok: added an issue https://github.com/foundry-rs/starknet-foundry/issues/1127

crates/cast/src/lib.rs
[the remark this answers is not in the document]
abulenok: improved the error msg

crates/cast/src/script.rs
abulenok: I guess since deployment scripts are not contracts we can do only this https://github.com/starkware-libs/cairo/blob/1c70f6459d1f38018da4397466ba2dccf6d48f81/crates/cairo-lang-runner/src/casm_run/mod.rs#L417-L419 ?

crates/cast/src/script.rs
abulenok: Do you mean the whole deployment scripts logic?

crates/cast/src/script.rs
abulenok: If you mean this https://github.com/foundry-rs/starknet-foundry/pull/1067/files#diff-8355c5a63f2b3a250d96b5722de9db86497c0946b175a45c11506f9626d14724R196-R227 piece of logic, it is consistent with the way `TestExecutionSyscallHandler` handles it.

crates/cast/src/script.rs
abulenok: you mean something like this? https://github.com/foundry-rs/starknet-foundry/blob/6722260dbddf56e309e3c549317a296ac9a017cd/crates/cheatnet/src/execution/contract_execution_syscall_handler.rs#L20-L29

abulenok: could be placed in some common crate along with https://github.com/foundry-rs/starknet-foundry/pull/1067#discussion_r1392818816

crates/cast/src/script.rs
abulenok: You mean why a replacer is needed?

crates/cast/src/helpers/scarb_utils.rs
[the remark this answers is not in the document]
THenry14: Sure. So my proposition is to get rid of this function and use what's in main, then pass it to `run` function. Then it should be easier for whoever gets to merge the changes second to rebase onto master branch containing the other guys changes. Sounds good?

crates/cast/src/script.rs
THenry14: I have discussed such idea with Tomek if I remember correctly. This is something we might be interested in in the future :+1: 

I have created an issue for this, so it does not get lost https://github.com/foundry-rs/starknet-foundry/issues/1134

crates/cast/src/script.rs
THenry14: that would be cool, let's try to asses this. If this is doable, and rather quick change, let's do it straight away. If not, let's do this in separate PR. saounds good?

crates/cast/src/script.rs
THenry14: good idea, let's do this under said issue, in separate PR

crates/cast/tests/data/scripts/hello_world/hello_world.cairo
THenry14: I guess this is this: https://github.com/foundry-rs/starknet-foundry/pull/1111

crates/cast/src/helpers/response_structs.rs
```diff
@@ -19,9 +19,9 @@ pub struct DeclareResponse {
     pub transaction_hash: FieldElement,
 }
 
-#[derive(Serialize)]
+#[derive(Serialize, Clone, Debug)]
 pub struct CallResponse {
-    pub response: String,
+    pub data: Vec<FieldElement>,
```
THenry14, line 24:
```suggestion
    pub response: Vec<FieldElement>,
```

Let's maybe keep the previous name? I guess it is more self-explanatory

crates/cast/tests/data/scripts/hello_world/Scarb.toml
```diff
@@ -0,0 +1,6 @@
+[package]
+name = "hello_world"
+version = "0.1.0"
+
+[dependencies]
+starknet = "2.3.0"
```
THenry14, line 6:
```suggestion
starknet = ">=2.3.0"
```

to avoid having to bump this each time we bump scarb version?

crates/cast/src/script.rs
```diff
@@ -0,0 +1,360 @@
...
+
+        Ok(())
+    }
+}
+
+pub fn run(
+    script_path: &Utf8PathBuf,
+    provider: &JsonRpcClient<HttpTransport>,
+    runtime: Runtime,
+    path_to_scarb_toml: &Option<Utf8PathBuf>,
```
THenry14, line 246: not sure if we should allow that, I think we should just take the script path and try to search for its scarb toml based on it. Allowing to pass this could potentially lead to some confusion and maybe even some errors. wdyt?

crates/cast/src/script.rs
```diff
@@ -0,0 +1,360 @@
...
+    check_compiler_path(true, Path::new(&script_path))?;
+
+    let metadata = parse_scarb_metadata(path_to_scarb_toml)?;
+
+    let corelib = corelib_for_package(&metadata, &metadata.workspace.members[0])?;
+    let db = &mut RootDatabase::builder().build()?;
+
+    init_dev_corelib(db, PathBuf::from(corelib));
+
+    let main_crate_ids = setup_project(db, Path::new(&script_path))?;
```
maciektr, line 257: I know this will not be relevant after our discussion, but **never** use `setup_project` with scarb based project, as it only reads `cairo_project.toml` files.

crates/cast/src/script.rs
```diff
@@ -0,0 +1,365 @@
...
+    let path = metadata
+        .target_dir
+        .unwrap_or(metadata.workspace.root.join("target"))
+        .join(metadata.current_profile)
+        .join(filename.clone());
+
+    ensure!(
+        path.exists(),
+        formatdoc! {r#"
+            package has not been compiled, file does not exist: {filename}
```
maciektr, line 258: Consider showing full path instead of filename only.

crates/cast/src/lib.rs
```diff
@@ -350,6 +350,27 @@ pub fn print_formatted(output: Vec<(&str, String)>, json: bool, error: bool) ->
     Ok(())
 }
 
+fn json_value_to_string(value: &Value, value_format: ValueFormat) -> Option<String> {
+    match value {
+        Value::Number(n) => {
+            let n = n
+                .as_u64()
+                .unwrap_or_else(|| panic!("failed to convert {n} to u64"));
```
THenry14, line 358: let's tru to avoid panics :+1:

crates/cast/tests/data/scripts/hello_world/hello_world.cairo
abulenok: #1111 merged to simplify rebasing

crates/cast/src/lib.rs
abulenok: it's essentially the same as `.expect`
https://stackoverflow.com/questions/55513198/is-there-a-more-concise-way-to-format-expect-message

crates/cast/src/script.rs
abulenok: done

crates/cast/src/script.rs
abulenok: yeah, it was supposed to be changed in #1111 to allow imports

crates/cast/src/helpers/response_structs.rs
abulenok: changed

crates/cast/src/script.rs
abulenok: not relevant anymore, since the compilation is now performed by scarb

crates/cast/src/helpers/scarb_utils.rs
```diff
@@ -1,6 +1,7 @@
 use anyhow::{anyhow, bail, Context, Result};
 use camino::Utf8PathBuf;
 use scarb_metadata;
+use scarb_metadata::Metadata;
```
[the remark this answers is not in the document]
abulenok: done

crates/cast/src/script.rs
```diff
@@ -0,0 +1,360 @@
+use std::any::Any;
```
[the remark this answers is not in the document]
abulenok: done

crates/cast/src/script.rs
abulenok: good point!

crates/cast/src/helpers/scarb_utils.rs
```diff
@@ -69,18 +70,47 @@ pub fn get_scarb_manifest() -> Result<Utf8PathBuf> {
     Ok(path)
 }
 
-pub fn get_scarb_metadata(manifest_path: &Utf8PathBuf) -> Result<scarb_metadata::Metadata> {
+pub fn get_scarb_metadata(
+    manifest_path: &Utf8PathBuf,
+    with_deps: bool,
+) -> Result<scarb_metadata::Metadata> {
```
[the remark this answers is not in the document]
abulenok: Since we have only the `with_deps` flag for now, I think it's redundant to add the `ScarbOpts` struct. 
If the number of additional arguments increases, we may introduce `ScarbOpts`.

crates/cast/src/script.rs
abulenok: done

crates/cast/src/lib.rs
abulenok: This one has a somewhat unique behaviour

crates/cast/src/helpers/scarb_utils.rs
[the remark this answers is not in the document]
maciektr: I wouldn't say it's redundant - you would get `default()`, easier to understand syntax (no positional flags) and easier path to extend args.

crates/cast/src/helpers/scarb_utils.rs
abulenok: updated

crates/cast/src/script.rs
abulenok: It's rather a quick change

abulenok: removed blockifier in https://github.com/foundry-rs/starknet-foundry/pull/1067/commits/9a974fcc0f2ed97df5f152605ac914a5d0a3971d

crates/cast/src/script.rs
abulenok: extracted some logic to `compile_script` & we use scarb build now for compilation

crates/cast/src/script.rs
THenry14: Perfect! Can we clean up the deps in Cargo.toml as well? I believe blockifier is still there

sncast_std/src/lib.cairo
```diff
@@ -0,0 +1,41 @@
+use starknet::{testing::cheatcode, ContractAddress};
+
+#[derive(Drop, Clone)]
+struct CallResult {
+    data: Array::<felt252>,
+}
+
+fn call(
```
MaksymilianDemitraszek, line 8: If we did it whith syscall::call_contract it would work with dispatchers out of the box

crates/cast/src/script.rs
abulenok: good point! 
done

crates/cast/src/lib.rs
MaksymilianDemitraszek: What do you mean by `unique behaviour`, it is not captured by the name. Please rename it to be accurate, current name suggest it just dumps json.

crates/cast/src/starknet_commands/script.rs
```diff
@@ -0,0 +1,340 @@
...
+use starknet::core::types::{BlockId, BlockTag::Pending, FieldElement};
+use starknet::providers::jsonrpc::HttpTransport;
+use starknet::providers::JsonRpcClient;
+use tokio::runtime::Runtime;
+
+#[derive(Args)]
+#[command(about = "")]
+pub struct Script {
+    /// Path to the script
+    pub script_path: Utf8PathBuf,
```
MaksymilianDemitraszek, line 49: I don't like this being based on paths, imo it should be workspace focused to be consistent with scarb/forge philosophy. 

Maybe better would be to have scripts package with fixed name `scripts` and call it by a funciton name?

crates/cast/src/helpers/scarb_utils.rs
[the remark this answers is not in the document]
abulenok: ok, makes sense
done

crates/cast/src/lib.rs
MaksymilianDemitraszek: Also could we unit test it?

MaksymilianDemitraszek (changes requested):

sncast_std/src/lib.cairo
abulenok: Yes, that is the plan.

crates/cast/src/lib.rs
abulenok: [239a777](https://github.com/foundry-rs/starknet-foundry/pull/1067/commits/239a777c680437e95eb5102e11c70a05f116e823)

crates/cast/src/starknet_commands/script.rs
abulenok: https://github.com/foundry-rs/starknet-foundry/pull/1067/commits/a429f89e743c6f364899f541655e5857e8c6ec37 
updated as discussed offline 

Now scripts have to be put into `scripts` folder.  To execute a script, the user will have to pass the script module name.

THenry14 (approved):
lgtm, we can iron out some smaller issues afterwards. Great job 💪

crates/cast/src/starknet_commands/script.rs
abulenok: Ok, now it has been changed. Scripts can exist as a separate Scarb package. We use `path_to_scarb_toml` to launch scripts from an arbitrary directory. `path_to_scarb_toml` is expected to point to the `Scarb.toml` within the package with scripts

THenry14: Context: after discussion with Maciek we decided that if we do this that way, we can still run script based on package and script name, but we can also add extrernal contracts as dependencies which we would want to do either way in the near future

MaksymilianDemitraszek (approved):

MaksymilianDemitraszek (commented):
I am ok with being merged but tests are a bit lacking, also please refactor the json method before merging I still don't know what it does

MaksymilianDemitraszek:
I assume docs will be included in a separate PR?

crates/cast/src/lib.rs
```diff
@@ -75,6 +75,28 @@ impl ValueFormat {
 
         input.to_string()
     }
+
+    #[must_use]
+    pub fn format_json_value(&self, value: &Value) -> Option<String> {
```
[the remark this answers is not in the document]
cptartur: I'm assuming this converts numeric values to strings? And leaves other strings as is? This should be renamed and/or docs should be added

crates/cast/src/starknet_commands/script.rs
```diff
@@ -0,0 +1,341 @@
...
+    ) {
+        Ok(result) => match result.value {
+            RunResultValue::Success(data) => Ok(ScriptResponse {
+                status: "success".to_string(),
+                msg: build_readable_text(&data),
+            }),
+            RunResultValue::Panic(panic_data) => Ok(ScriptResponse {
+                status: "script panicked".to_string(),
+                msg: build_readable_text(&panic_data),
+            }),
```
cptartur, line 281: This seems like `status` could be an enum instead of a string since there are only two values available.

crates/cast/src/starknet_commands/script.rs
```diff
@@ -0,0 +1,341 @@
...
+    }
+}
+
+fn compile_script(path_to_scarb_toml: Option<Utf8PathBuf>) -> Result<Utf8PathBuf> {
+    let scripts_manifest_path = match path_to_scarb_toml {
+        Some(path) => path,
+        None => get_scarb_manifest()
+            .context("Failed to obtain manifest path from scarb")
+            .unwrap(),
+    };
```
cptartur, line 293: Not sure if this logic should be here. It would make sense to accept `scripts_manifest_path` as `Utf8PathBuf` here and do the optional handling at caller site

crates/cast/src/starknet_commands/script.rs
```diff
@@ -0,0 +1,341 @@
...
+    ensure!(
+        scripts_manifest_path.exists(),
+        "Path {scripts_manifest_path} does not exist"
+    );
```
cptartur, line 297: Same goes for this verification

crates/cast/src/starknet_commands/script.rs
```diff
@@ -0,0 +1,341 @@
...
+        path.exists(),
+        "package has not been compiled, file does not exist: {path}"
+    );
+
+    Ok(path)
+}
+
+// taken from starknet-foundry/crates/forge/src/test_case_summary.rs
+/// Helper function to build `readable_text` from a run data.
+fn build_readable_text(data: &Vec<Felt252>) -> Option<String> {
```
cptartur, line 325: Let's create an issue to move this code to a common crate

crates/cast/tests/data/scripts/hello_world/src/using_starknet_syscall.cairo
```diff
@@ -0,0 +1,7 @@
+use starknet::info::get_execution_info;
+use box::BoxTrait;
+
+fn main() {
+    let exec_info = get_execution_info().unbox();
+    assert(1 == 2, 'unreachable');
+}
```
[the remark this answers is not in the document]
cptartur: I think it's just testing that starknet related syscalls do not work

sncast_std/src/lib.cairo
cptartur: How exactly would you integrate dispatchers with these scripts? Dispatcher generation requires the existence of contract interfaces.

crates/cast/src/starknet_commands/script.rs
THenry14: I will make a separate issue for this

crates/cast/src/starknet_commands/script.rs
THenry14: we'll think about it after release, since this is non blocking. thanks.

crates/cast/src/starknet_commands/script.rs
THenry14: https://github.com/foundry-rs/starknet-foundry/issues/1127

sncast_std/src/lib.cairo
THenry14: for generation, Alex mentioned some tool being developed (abigen?), and we will probably want to use that. Apart from that, no specific plans at the moment, only vague ideas.

crates/cast/src/lib.rs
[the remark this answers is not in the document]
abulenok: https://github.com/foundry-rs/starknet-foundry/issues/1163

crates/cast/src/starknet_commands/script.rs
```diff
@@ -0,0 +1,341 @@
...
+use starknet::providers::jsonrpc::HttpTransport;
+use starknet::providers::JsonRpcClient;
+use tokio::runtime::Runtime;
+
+#[derive(Args)]
+#[command(about = "Execute a deployment script")]
+pub struct Script {
+    /// Module name that contains the `main` function, which will be executed
+    pub script_module_name: String,
+}
```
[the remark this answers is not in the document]
abulenok: This is the module name of the script.
To launch the script from an arbitrary directory, the user may use `path-to-scarb-toml` CLI argument.

crates/cast/tests/data/scripts/hello_world/src/using_starknet_syscall.cairo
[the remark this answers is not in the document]
abulenok: exactly, this is expected to fail

crates/cast/tests/data/scripts/hello_world/src/verify_import.cairo
```diff
@@ -0,0 +1,6 @@
+use hello_world_script::some_code::calc_fib;
+
+fn main() {
+    let res = calc_fib();
+    assert(res == 5, res);
+}
```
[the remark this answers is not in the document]
abulenok: removed

crates/cast/src/script.rs
cptartur: Sure but the logic there is not ideal, so we can consider not copying it in the exact manner.

crates/cast/src/helpers/scarb_utils.rs
```diff
@@ -1,12 +1,19 @@
...
+use std::default::Default;
 use std::env;
+use std::fs::canonicalize;
 use std::process::{Command, Stdio};
 use std::str::FromStr;
 
+#[derive(Default, Clone, Copy)]
+pub struct ScarbOpts {
+    pub with_deps: bool,
+}
```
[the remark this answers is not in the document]
abulenok: refactored

crates/cast/src/lib.rs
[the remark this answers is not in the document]
cptartur: The issue is not really refactoring it, but more that it's not immediately obvious what's this function's purpose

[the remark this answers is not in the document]
abulenok: apply ValueFormat formatting to json_values

sncast_std/src/lib.cairo
abulenok: yep, it is yet to be researched.

crates/cast/src/script.rs
abulenok: ok, let's consider reworking it in a separate PR

https://github.com/foundry-rs/starknet-foundry/issues/1168

PR #1071, opened 2023-11-08
999 remove available gas support

tkumor3:
Closes #

## Introduced changes

-

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/test-collector/src/lib.rs
```diff
@@ -556,16 +557,31 @@ pub fn collect_tests(
...
-            available_gas: config.available_gas,
-            ignored: config.ignored,
-            expected_result: config.expected_result,
-            fork_config: config.fork_config,
-            fuzzer_config: config.fuzzer_config,
+        .map(|(test_name, config)| {
+            if config.available_gas.is_some() {
+                gas_error = Some(Err(
+                    anyhow!
+                        ("{} - Gas calculation used for available_gas is incorrect, because it doesn't include the cost of contract execution.",
```
piotmag769, line 564: This is a bad error message, we should just state that it is not supported without bothering the user with unnecessary details

piotmag769, line 564: Cannot we use `bail!` here instead?

cptartur: Agreed

crates/test-collector/src/lib.rs
```diff
@@ -556,16 +557,31 @@ pub fn collect_tests(
...
+                expected_result: config.expected_result,
+                fork_config: config.fork_config,
+                fuzzer_config: config.fuzzer_config,
+            }
         })
         .collect();
 
+    if let Some(err) = gas_error {
+        return err;
+    }
```
cptartur, line 583: Instead of using approach like this, lets refactor the use of `map` so it returns a Result which we can then bubble.

Change `collect()` to something along the lines of this `collect::<Result<Vec<_>>>()?` and update the code inside the map so it users `Result` there.

cptartur (changes requested):

crates/test-collector/src/lib.rs
piotmag769: Doesn't `bail!` solve all of this and is also more readable?

tkumor3: bail! is only shortcut of `return new Err(anyhow!(..))`. @cptartur  I have no idea how to achive, collect::<Result<Vec<_>>>()?. If I use bail! I will get `Vec<Result<_>>`

piotmag769: Yep, thus returning from the function and bubbling the error. If we do it above where I suggested it all this mapping won't be necessary

piotmag769: Oh nvm, it is inside a `map` 🤦 Then I guess this approach is fine

crates/test-collector/src/lib.rs
```diff
@@ -558,7 +558,7 @@ pub fn collect_tests(
         .into_iter()
         .map(|(test_name, config)| {
             if config.available_gas.is_some() {
-                bail!("{} - `gas_calculation` attribute is not support", test_name)
+                bail!("{} - Available gas attribute is not supported: Contract functions execution cost would not be included in the gas calculation.", test_name)
```
cptartur, line 561:
```suggestion
                bail!("{} - `Attribute `available_gas` is not supported: Contract functions execution cost would not be included in the gas calculation.", test_name)
```

cptartur (approved):

piotmag769 (approved):

PR #1084, opened 2023-11-09
add flag rerun-failed, to run only failed tests in prev run

tkumor3:
Closes #

## Introduced changes

-

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/forge/src/init.rs
```diff
@@ -60,6 +61,13 @@ fn add_target_to_toml(path: &Path) -> Result<()> {
     Ok(())
 }
 
+fn extend_gitignore(path: &Path) -> Result<()> {
+    let mut file = OpenOptions::new().write(true).append(true).open(path)?;
+    writeln!(file, ".prev_tests_failed")?;
```
cptartur, line 66: If we already have `.snfoundry_cache` directory, wouldn't it make sense to store this file there?

crates/forge/src/main.rs
```diff
@@ -94,6 +95,10 @@ struct TestArgs {
     /// Control when colored output is used
     #[arg(value_enum, long, default_value_t = ColorOption::Auto, value_name="WHEN")]
     color: ColorOption,
+
+    /// Run only failures
+    #[arg(long = "only_failed")]
+    only_failed: bool,
```
cptartur, line 101: I'd name it `rerun_failed`

crates/forge/src/main.rs
```diff
@@ -152,6 +157,24 @@ fn combine_configs(
     )
 }
 
+fn write_failed_tests(all_failed_tests: &[TestCaseSummary]) -> Result<()> {
+    let test_fails = std::env::current_dir()?.join(".prev_tests_failed");
```
cptartur, line 161: This doesn't seem safe. We should be using a workspace root from scarb to resolve directory for storing `.prev_tests_failed`. You can take a look at this: https://github.com/foundry-rs/starknet-foundry/blob/fb18cdd9ae335677ea12a2448dc86533d3df2052/crates/forge/src/running.rs#L353

cptartur: After that, we should write/extend tests using scarb workspaces so we can verify that this approach works with workspaces.

crates/forge/src/test_filter.rs
```diff
@@ -69,6 +78,13 @@ impl TestsFilter {
             NameFilter::ExactMatch(name) => {
                 cases.into_iter().filter(|tc| tc.name == *name).collect()
             }
+            NameFilter::OnlyFailed => match Self::read_test_fails_file() {
+                Ok(result) => cases
+                    .into_iter()
+                    .filter(|tc| result.iter().any(|name| name == &tc.name))
+                    .collect(),
+                Err(_) => cases,
```
cptartur, line 86: I wonder if we should output some information like: "Unable to find previously failed tests, re-running all.". Because in the way you wrote it it will always fail silently with no feedback for the user.

crates/forge/src/test_filter.rs
```diff
@@ -114,13 +140,13 @@ mod tests {
     #[test]
```
cptartur, line 140: Let's add a test where only_failed is `true`.

crates/forge/tests/e2e/running.rs
```diff
@@ -414,6 +414,52 @@ fn with_include_ignored_flag_and_filter() {
...
+        .current_dir(&temp)
+        .arg("--only_failed")
+        .assert()
+        .code(1);
+    let snapbox = test_runner();
+    let output = snapbox
+        .current_dir(&temp)
+        .arg("--only_failed")
+        .assert()
+        .code(1);
```
cptartur, line 432: `--only_failed` is only needed in the second call to snapbox.

cptartur (changes requested):

crates/forge/src/init.rs
tkumor3: It's clean after each run of snforge

crates/forge/src/test_filter.rs
tkumor3: I'm not sure, what do you mean 🤔

cptartur: You've added a new attribute to tests filter, but there's no unit test covering it.

crates/forge/src/test_filter.rs
tkumor3: In the way I wrote in the case when in some cases the filter doesn't work it will run all cases, and generate the new failed file.  I'm wondering what is the best solution in the case of error 🤔

crates/forge/src/test_filter.rs
tkumor3: should we test it in the unit? We have already covered it in e2e tests. The problem with this test is that it's impossible to write it without ingeration in production code

crates/forge/src/shared_cache.rs
```diff
@@ -0,0 +1,61 @@
...
+
+        for line in all_failed_tests {
+            if let TestCaseSummary::Failed { name, .. } = line {
+                file.write_all((name.clone() + "\n").as_bytes())
+                    .expect("Can not write to file");
+            }
+        }
+
+        Ok(())
+    }
```
cptartur, line 50: I think we could do https://doc.rust-lang.org/std/io/struct.BufWriter.html here since we are calling `write_all` multiple times in a loop.

crates/forge/src/test_filter.rs
cptartur: The amount of code required for mocking doesn't seem to be worth it. Let's make sure we have comprehensive e2e tests instead.

crates/forge/tests/e2e/running.rs
```diff
@@ -414,6 +414,48 @@ fn with_include_ignored_flag_and_filter() {
     );
 }
 
+#[test]
+fn with_rerun_failed_flag() {
```
cptartur, line 418: We should also have a test where there's no `.prev_tests_failed` initially and we call runner two times, first normally and second with `--rerun-failed` to see if it's created correctly.

This is relying on manually prepared cache file which is good, but not exclusively.

tkumor3: sure, but it's related to this [comment]( https://github.com/foundry-rs/starknet-foundry/pull/1084#discussion_r1392209904 ). After making a decision, I will add a new tests

crates/forge/src/shared_cache.rs
```diff
@@ -0,0 +1,53 @@
...
+    let tests_failed_path = cache_dir()?.join(".prev_tests_failed");
+    let file = File::open(tests_failed_path)?;
+    let buf = BufReader::new(file);
+    Ok(buf
+        .lines()
+        .map(|l| l.expect("Could not parse line"))
+        .collect())
+}
+
+pub fn cache_dir() -> Result<Utf8PathBuf> {
```
piotmag769, line 21: `create_cache_dir`, current name suggests that it only returns a path

crates/forge/src/shared_cache.rs
```diff
@@ -0,0 +1,53 @@
...
+            file.write_all((name.clone() + "\n").as_bytes())
+                .expect("Can not write to file");
+        }
+    }
+
+    Ok(())
+}
+
+pub fn clean_cache() -> Result<()> {
+    let cache_dir = cache_dir()?;
```
piotmag769, line 48: It shouldn't call a function that can potentially create a directory for no reason

crates/forge/src/test_filter.rs
piotmag769: It shouldn't fail silently in any scenario, we should just bubble the error from `read_tests_failed_file`

crates/forge/src/test_filter.rs
```diff
@@ -59,7 +66,6 @@ impl TestsFilter {
 
     pub(crate) fn filter_tests(&self, test_crate: CompiledTestCrateRaw) -> CompiledTestCrateRaw {
         let mut cases = test_crate.test_cases;
-
```
piotmag769, line 62: unrelated

crates/forge/src/test_filter.rs
```diff
@@ -43,7 +48,9 @@ impl TestsFilter {
             IgnoredFilter::NotIgnored
         };
 
-        let name_filter = if exact_match {
+        let name_filter = if rerun_failed {
+            NameFilter::RerunFailed
+        } else if exact_match {
```
piotmag769, line 53: It shouldn't override the filtering with names, those should be independent. I. e., I would expect `snforge test --rerun-failed xyz` to only run tests that failed the last time with `xyz` in their logical path

crates/forge/src/main.rs
```diff
@@ -94,6 +93,10 @@ struct TestArgs {
     /// Control when colored output is used
     #[arg(value_enum, long, default_value_t = ColorOption::Auto, value_name="WHEN")]
     color: ColorOption,
+
+    /// Run only failures
+    #[arg(long = "rerun_failed")]
```
piotmag769, line 98: just use `#[arg(long)]` so the name becomes `--rerun-failed`

crates/forge/src/main.rs
```diff
@@ -94,6 +93,10 @@ struct TestArgs {
     /// Control when colored output is used
     #[arg(value_enum, long, default_value_t = ColorOption::Auto, value_name="WHEN")]
     color: ColorOption,
+
+    /// Run only failures
```
piotmag769, line 97: `/// Run tests that failed during the last run`

crates/forge/src/shared_cache.rs
```diff
@@ -0,0 +1,53 @@
...
+    let cache_dir_path = workspace_root.join(CACHE_DIR);
+    std::fs::create_dir_all(&cache_dir_path)?;
+    Ok(cache_dir_path)
+}
+
+pub fn write_failed_tests(all_failed_tests: &[TestCaseSummary]) -> Result<()> {
+    let tests_failed_path = cache_dir()?.join(".prev_tests_failed");
+    let mut file = std::fs::OpenOptions::new()
+        .write(true)
+        .create(true)
```
piotmag769, line 34: Shouldn't we use `truncate` here too? 
https://stackoverflow.com/questions/65782872/how-to-delete-the-contents-of-a-file-before-writing-to-it

crates/forge/src/shared_cache.rs
```diff
@@ -0,0 +1,53 @@
...
+pub fn cache_dir() -> Result<Utf8PathBuf> {
+    let scarb_metadata = MetadataCommand::new().inherit_stderr().exec();
+    let scarb_metadata = scarb_metadata?;
+    let workspace_root = scarb_metadata.workspace.root.clone();
+    let cache_dir_path = workspace_root.join(CACHE_DIR);
+    std::fs::create_dir_all(&cache_dir_path)?;
+    Ok(cache_dir_path)
+}
+
+pub fn write_failed_tests(all_failed_tests: &[TestCaseSummary]) -> Result<()> {
```
piotmag769, line 30: This name is not descriptive enough imo, what about sth like `cache_failed_tests_names`

docs/src/appendix/forge/test.md
```diff
@@ -41,6 +41,10 @@ Run only tests marked with `#[ignore]` attribute.
 
 Run all tests regardless of `#[ignore]` attribute.
 
+## `--rerun_failed`
+
+Run only failed tests from previous run
```
piotmag769, line 46:
```suggestion
Run tests that failed during the last run
```

crates/forge/src/shared_cache.rs
```diff
@@ -0,0 +1,53 @@
...
+use anyhow::Result;
+use camino::Utf8PathBuf;
+use scarb_metadata::MetadataCommand;
+use std::fs::{self, File};
+use std::io::{BufRead, BufReader, Write};
+
+use crate::test_case_summary::TestCaseSummary;
+
+pub fn read_tests_failed_file() -> Result<Vec<String>> {
+    let tests_failed_path = cache_dir()?.join(".prev_tests_failed");
```
piotmag769, line 12: Preferebly we can have `".prev_tests_failed"` as a global `const &str`, then we can reuse it below too

crates/forge/src/shared_cache.rs
```diff
@@ -0,0 +1,53 @@
...
+
+use anyhow::Result;
+use camino::Utf8PathBuf;
+use scarb_metadata::MetadataCommand;
+use std::fs::{self, File};
+use std::io::{BufRead, BufReader, Write};
+
+use crate::test_case_summary::TestCaseSummary;
+
+pub fn read_tests_failed_file() -> Result<Vec<String>> {
```
piotmag769, line 11: This function should check if the last failed file exists:
- if it does, attempt to read it
- if it doesn't, return an empty string

This way we only bubble actual errors and we don't just dismiss them while also handling the case where the file doesn't exist

piotmag769 (changes requested):

crates/forge/src/test_filter.rs
tkumor3: IMO it's a separate feature, it demands merging filters

crates/forge/src/shared_cache.rs
tkumor3: Probably I can use `https://doc.rust-lang.org/std/fs/struct.File.html#method.create` instead

crates/forge/src/shared_cache.rs
tkumor3: I change also `read_tests_failed_file ` to `get_cached_failed_tests_names` to be more consistent

crates/forge/src/shared_cache.rs
tkumor3: I used `Option` instead of empty string

crates/forge/src/test_filter.rs
piotmag769: Yea, those are separate features, that's why they shouldn't be merged. Filtering tests based on `--rerun-failed`should be independent from filtering based on `[TEST_FILTER]` and `--exact`. It would be really confusing if  we would change the behaviour of filtering by name when `--rerun-failed` flag is specified

crates/forge/src/shared_cache.rs
```diff
@@ -0,0 +1,62 @@
...
+pub fn get_cache_dir() -> Result<Utf8PathBuf> {
+    let scarb_metadata = MetadataCommand::new().inherit_stderr().exec();
+    let scarb_metadata = scarb_metadata?;
+    let workspace_root = scarb_metadata.workspace.root.clone();
+    Ok(workspace_root.join(CACHE_DIR))
+}
+
+pub fn get_or_create_cache_dir() -> Result<Utf8PathBuf> {
+    let cache_dir_path = get_cache_dir()?;
+    std::fs::create_dir_all(&cache_dir_path)?;
```
piotmag769, line 39: I think we could just use `create_dir` and rename the func to `create_cache_dir` so we only use it when we actually want to create the directory and get an error otherwise

crates/forge/src/test_filter.rs
```diff
@@ -2,11 +2,13 @@ use crate::collecting::{CompiledTestCrate, CompiledTestCrateRaw, ValidatedForkCo
...
 
+use crate::shared_cache::get_cached_failed_tests_names;
+
 #[derive(Debug, PartialEq)]
 // Specifies what tests should be included
 pub struct TestsFilter {
-    // based on name
-    name_filter: NameFilter,
+    // based on names
+    name_filters: Vec<NameFilter>,
```
piotmag769, line 11: Cannot we just have a separate bool/enum with two variansts like `last_failed`? Much less confusing imo

crates/forge/src/shared_cache.rs
```diff
@@ -0,0 +1,62 @@
...
+}
+
+pub fn get_cache_dir() -> Result<Utf8PathBuf> {
+    let scarb_metadata = MetadataCommand::new().inherit_stderr().exec();
+    let scarb_metadata = scarb_metadata?;
+    let workspace_root = scarb_metadata.workspace.root.clone();
+    Ok(workspace_root.join(CACHE_DIR))
+}
+
+pub fn get_or_create_cache_dir() -> Result<Utf8PathBuf> {
```
piotmag769, line 37: Doesn't need to be `pub`

crates/forge/src/shared_cache.rs
```diff
@@ -0,0 +1,62 @@
...
+        .lines()
+        .map(|l| l.expect("Could not parse line"))
+        .collect();
+    if tests.is_empty() {
+        return Ok(None);
+    }
+    Ok(Some(tests))
+}
+
+pub fn get_cache_dir() -> Result<Utf8PathBuf> {
```
piotmag769, line 30: Doesn't need to be `pub`

crates/forge/src/test_filter.rs
```diff
@@ -43,34 +48,49 @@ impl TestsFilter {
...
+                NameFilter::All => cases,
+                NameFilter::Match(filter) => cases
+                    .into_iter()
+                    .filter(|tc| tc.name.contains(filter))
+                    .collect(),
+                NameFilter::ExactMatch(name) => {
+                    cases.into_iter().filter(|tc| tc.name == *name).collect()
+                }
+                NameFilter::RerunFailed => {
+                    match get_cached_failed_tests_names().expect("Can't read cached failed tests") {
```
piotmag769, line 84: We shouldn't panic here, there can be a legitimate reason that we cannot read this file that is not an hole in our logic e.g. insufficient permissions. Just bubble an error wrapped with `context`

crates/forge/src/shared_cache.rs
```diff
@@ -0,0 +1,62 @@
...
+    let scarb_metadata = MetadataCommand::new().inherit_stderr().exec();
```
piotmag769, line 31: I really dislike this hidden call to metadata. We should make sure to just use paths from the metadata read in `main_execution` (e.g. store path cache dir in `TestFilter`, prob not ideal solution though). The only case when this call is actually needed is for `clean_cache`

crates/forge/src/shared_cache.rs
piotmag769: If we want to just run nothing in both cases I think the empty vector is better than `Option` since on the type level nothing prevents you from having `Some(empty_vector)`. Without wrapping it with `Option` we don't need to handle any cases separately

In your implementation we just effectively ignore the `--rerun-failed` in those cases, right?

tkumor3: currently will run all tests. The reason is if you fix a few tests you have to rerun all because you could break sth while fixing

tkumor3: IMO, it's better to run all tests in case the file is empty because when we return that all tests are filtered it suggests that all tests work and it might not be true. Is possible to break one test while fixing another

MaksymilianDemitraszek: If set of existing tests changed in any way (added a new one, one removed) we should flush the cache, thus run everything.

tkumor3: what do you suggest, compute control sum (md5) based on the names of all tests?

crates/forge/src/shared_cache.rs
```diff
@@ -0,0 +1,64 @@
...
+        .collect();
+    if tests.is_empty() {
+        return Ok(None);
+    }
+    Ok(Some(tests))
+}
+
+fn get_cache_dir(workspace_root: &Utf8PathBuf) -> Result<Utf8PathBuf> {
+    Ok(workspace_root.join(CACHE_DIR))
+}
```
piotmag769, line 32: Isn't it better to just pass the `workspace_root.join(CACHE_DIR)` to every function?

crates/forge/src/test_filter.rs
```diff
@@ -1,14 +1,22 @@
...
 // Specifies what tests should be included
 pub struct TestsFilter {
     // based on name
     name_filter: NameFilter,
     // based on `#[ignore]` attribute
     ignored_filter: IgnoredFilter,
+    // based on rerun_failed flag
+    last_failed_filter: bool,
+
+    workspace_root: Utf8PathBuf,
```
piotmag769, line 19: Better to store `cache_dir_path` here imo

crates/forge/src/lib.rs
```diff
@@ -236,7 +236,7 @@ pub async fn run(
     let test_crates = test_crates
         .into_iter()
         .map(|tc| tests_filter.filter_tests(tc))
-        .collect_vec();
+        .collect::<Result<Vec<CompiledTestCrate<RawForkConfig>>>>()?;
```
piotmag769, line 239: We can use our alias
```suggestion
        .collect::<Result<Vec<CompiledTestCrateRaw>>>()?;
```

crates/forge/src/shared_cache.rs
cptartur: >If set of existing tests changed in any way (added a new one, one removed) we should flush the cache, thus run everything.

This could be done in a separate PR.

But what about the case in which tests didn't change? This is more relevant to the ongoing discussion.

tkumor3: I've already created ticket, and proper PoC

piotmag769: @tkumor3 I think we should merge the PR that changes the logic into this branch and then merge this branch to master

piotmag769 (approved):

crates/forge/src/lib.rs
```diff
@@ -110,7 +109,7 @@ pub async fn run(
     let test_crates = test_crates
         .into_iter()
         .map(|tc| tests_filter.filter_tests(tc))
-        .collect_vec();
+        .collect::<Result<Vec<CompiledTestCrateRaw>>>()?;
```
cptartur, line 112: Do we need this specialization?

crates/forge/src/shared_cache.rs
```diff
@@ -0,0 +1,61 @@
...
+pub const PREV_TESTS_FAILED: &str = ".prev_tests_failed";
+
+use anyhow::{Ok, Result};
+use camino::Utf8PathBuf;
+use forge_runner::test_case_summary::TestCaseSummary;
+use scarb_metadata::MetadataCommand;
+use std::fs::{self, File};
+use std::io::{BufRead, BufReader, BufWriter, Write};
+
+pub fn get_cached_failed_tests_names(cache_dir_path: &Utf8PathBuf) -> Result<Option<Vec<String>>> {
```
cptartur, line 11: Nit, but we could do without `get` in function names

```suggestion
pub fn cached_failed_tests_names(cache_dir_path: &Utf8PathBuf) -> Result<Option<Vec<String>>> {
```

crates/forge/src/shared_cache.rs
cptartur, line 1: We could have some unit tests for these methods, using tempdir

crates/forge/src/test_filter.rs
```diff
@@ -76,16 +91,26 @@ impl TestsFilter {
...
 
+        if self.last_failed_filter {
+            cases = match get_cached_failed_tests_names(&self.cache_dir_path)? {
+                Some(result) => cases
+                    .into_iter()
+                    .filter(|tc| result.iter().any(|name| name == &tc.name))
+                    .collect(),
+                None => cases,
+            }
+        }
```
cptartur, line 102: This will allow `--last-failed` filtering to be combined with other filters. I'm not sure if this is a good behavior. Perhaps we should return early here? Or have it as an entirely separate filter.

cptartur: cc @piotmag769

crates/forge/src/lib.rs
tkumor3: Zrzut ekranu 2023-11-21 o 16 07 10
I changed the `tests_filter` type

crates/forge/src/shared_cache.rs
tkumor3: IMO, in the context of cache the best way is to use prefix set/get so it could be `get_cached_failed_tests_names`
and setter should be named `set_cached_failed_tests_names`

crates/forge/src/shared_cache.rs
tkumor3: IMO, we have converted all these functions in our e2e tests, so it's enough. Those functions only read and write to the file. I can add these tests, but it doesn't make our code safer.

crates/forge/src/test_filter.rs
tkumor3: but what is wrong with combining with other filters?

cptartur (approved):

Issue #417, opened 2023-08-07
Decouple snforge from scarb version


MaksymilianDemitraszek:
Generally we should introduce `test` target in scarb and move our test collection logic there. Scarb should produce casm binaries for each test target. forge should load them and execute. Scarb has to introduce artifacts file to include necessary metadata.

Should be synced with scarb team beforehand.

piotmag769:
Related scarb PR: https://github.com/software-mansion/scarb/pull/886

PR #1090, opened 2023-11-09
Use compiler structs in test-collector

piotmag769:
Test collector refactors related to #417 (we need to clean up the test collector before moving it outside of our repo) to ensure the same structs will be always used in `snforge-test-collector` and `forge` (serialisation and deserialisation has to be stable)

## Introduced changes

- Use `TestExpectation` and `PanicExpectation` from the compiler repository

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


cptartur (commented):
Does this address the issue of serialized and deserialized forms being stable? I think it's only valid under assumption that we are using exactly the same version of `cairo-lang-test-plugin` as scarb does, because there's no guarantee that definition or serialization or deserialization of this enum won't change between versions.

What should be considered instead is manually ensuring that keys, etc. stay the same for the outputs generated by scarb as we are expecting here.

PR #1093, opened 2023-11-10
Added Scarb workspaces for Cast

### LINKED ISSUE #17: Scarb artifact name change
In new scarb version the artifact name will change. Right now it is:

[package name].starknet_artifacts.json

and it will change to

[target name].starknet_artifacts.json

In most cases target name is the same as package name, but it could be different. We should take a target name from scarb metadata, and use the information to search for starknet_artifacts.json file
https://github.com/foundry-rs/starknet-foundry/blob/c04585092890f3cd351b36b411a4dae651c35feb/crates/cast/src/starknet_commands/declare.rs#L109

Docs for scarb metadata can be found here:
https://docs.rs/scarb-metadata/latest/scarb_metadata/struct.TargetMetadata.html#structfield.name

An example on how we retrieve some metadata can be found here:
https://github.com/foundry-rs/starknet-foundry/blob/c04585092890f3cd351b36b411a4dae651c35feb/crates/cast/src/starknet_commands/declare.rs#L91

### LINKED ISSUE #449: Look into scarb workspaces support

Cast

### Feature Request

Scarb will soon support workspaces. We should take a look at this, and figure out if this is something we can/should support.

Be sure to document how these workspaces interact with Cast profiles.

Radinyn:
## Checklist

- [ ] Linked relevant issue
- [ ] Added relevant tests


crates/cast/tests/e2e/main_tests.rs
```diff
@@ -119,7 +120,20 @@ async fn test_happy_case_mixed() {
...
+    let temp_dir = TempDir::new().expect("Unable to create a temporary directory");
+    let manifest_path = temp_dir.path().join("./Scarb.toml");
+    let manifest_path = manifest_path.to_str().unwrap();
+
+    fs_extra::file::copy(
+        "tests/data/files/noconfig_Scarb.toml",
+        manifest_path,
+        &fs_extra::file::CopyOptions::new().overwrite(true),
+    )
+    .unwrap();
```
THenry14, line 132: maybe we could use `crate::helpers::fixtures::duplicate_directory_with_salt` instead? 
if not suitable let's add another helper function that does this, as the code is repeated in multiple places.

crates/cast/src/helpers/build.rs
```diff
@@ -0,0 +1,61 @@
+use anyhow::{Context, Result};
+use scarb_artifacts::{get_contracts_map, target_dir_for_package, StarknetContractArtifacts};
+use scarb_metadata::{Metadata, PackageMetadata};
+use scarb_ui::args::PackagesFilter;
+use std::collections::HashMap;
+use std::fs;
+use std::process::{Command, Stdio};
+
+pub const LIB_CONTRACT_ARTIFACTS_NAME: &str = "__sncast_lib_contract";
```
THenry14, line 9: Let's move this to `helpers::constants`

benchmarks/benchmarks copy.csv
```diff
@@ -0,0 +1,4 @@
+,n_files,n_unit,n_integration,protostar,forge,cairo_test
```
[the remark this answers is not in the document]
Radinyn: Not sure how did this happen, will remove it

crates/cast/src/main.rs
```diff
@@ -372,3 +407,59 @@ fn update_cast_config(config: &mut CastConfig, cli: &Cli) {
...
+    pub metadata: Result<Metadata>,
+    pub package: Result<PackageMetadata>,
+}
+
+// Lazily get all the required package information from CLI
+fn get_package_data(cli: &Cli) -> PackageData {
+    let mut result = PackageData {
+        manifest_path: Err(anyhow!("Could not retrieve the manifest path")),
+        metadata: Err(anyhow!("Could not retrieve metadata")),
+        package: Err(anyhow!("Could not retrieve package metadata")),
```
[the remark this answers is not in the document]
Radinyn: The "no Scarb" error didn't change. The code uses those default messages only when the required previous field wasn't found (eg. there is no metadata when there's no manifest path) in practice these should barely (or never) show up

crates/cast/src/helpers/build.rs
```diff
@@ -0,0 +1,61 @@
...
+            .join(&metadata.current_profile)
+            .join(sierra_filename);
+
+        let lib_artifacts = StarknetContractArtifacts {
+            sierra: fs::read_to_string(sierra_path)?,
+            casm: String::new(), // There seems to be no need for casm in lib
+        };
+
+        contracts.insert(LIB_CONTRACT_ARTIFACTS_NAME.to_owned(), lib_artifacts);
+    }
```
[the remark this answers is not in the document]
Radinyn: This section only inserts lib's content into out contract mapping, the lib section's behavior in Scarb.toml didn't change

crates/cast/src/main.rs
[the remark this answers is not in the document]
THenry14: So I tried to check this, and it seems the behaviour is as you described when we are in the directory with Scarb.toml. However this changes when we pass the path to it:

```
# happy case
$ which scarb
/Users/wojciech/.asdf/shims/scarb

$ cargo run -- --path-to-scarb-toml scripts/Scarb.toml --url http:/127.0.0.1:5055/rpc --account user1 script map_script
command: script
status: success

# remove scarb from PATH
$ export PATH= (...)
$ which scarb
scarb not found

$ cargo run -- --path-to-scarb-toml scripts/Scarb.toml --url http:/127.0.0.1:5055/rpc --account user1 script map_script
Error: Could not retrieve package metadata

# now go into the dir
$ cd scripts
$ cargo run -- --url http:/127.0.0.1:5055/rpc --account user1 script map_script
Error: Failed to obtain manifest path from scarb

Caused by:
    0: Cannot find `scarb` binary in PATH. Make sure you have Scarb installed https://github.com/software-mansion/scarb
    1: cannot find binary path
```

so that's a bit odd

crates/cast/src/helpers/scarb_utils.rs
```diff
@@ -155,39 +156,21 @@ pub fn get_package_metadata<'a>(
 
 pub fn parse_scarb_config(
     profile: &Option<String>,
-    path: &Option<Utf8PathBuf>,
+    package_metadata: Option<&PackageMetadata>,
 ) -> Result<CastConfig> {
```
abulenok, line 160: let's rename this function, maybe `get_cast_config`?

crates/cast/src/main.rs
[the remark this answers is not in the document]
Radinyn: I've moved the scarb check higher - we require it anyway and it looks like it was the exception to this rule.

Cargo.toml
```diff
@@ -75,3 +75,4 @@ which = "4.4.0"
 conversions = { path = "./crates/conversions" }
 test-case = "3.1.0"
 scarb-metadata = "1.8.0"
+scarb-ui = "0.1.2"
```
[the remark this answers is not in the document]
Radinyn: It should stay here, it's used by forge as well

crates/cast/tests/e2e/main_tests.rs
Radinyn: Moved to a new utility function

crates/cast/src/helpers/build.rs
Radinyn: Done

crates/cast/src/helpers/scarb_utils.rs
Radinyn: Done

crates/cast/src/main.rs
```diff
@@ -111,25 +121,44 @@ fn main() -> Result<()> {
         ValueFormat::Default
     };
 
-    let mut config = parse_scarb_config(&cli.profile, &cli.path_to_scarb_toml)?;
+    which::which("scarb")
+        .context("Cannot find `scarb` binary in PATH. Make sure you have Scarb installed https://github.com/software-mansion/scarb")?;
```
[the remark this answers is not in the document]
abulenok: Since `scarb` is not required for certain commands, I don't think we should put it here, this will add an unnecessary limitation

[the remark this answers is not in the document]
abulenok: Maybe extract to the separate function and call whenever we want to use scarb ?

crates/cast/src/helpers/build.rs
```diff
@@ -0,0 +1,61 @@
...
+            "Scarb build returned non-zero exit code: {} - error message: {}",
+            result_code,
+            result_msg
+        );
+    }
+
+    let mut contracts = get_contracts_map(metadata, &package.id)?;
+
+    if for_scripts {
+        // Insert lib "contract"
```
abulenok, line 45: I think it's pretty clear from the code what's going on here. We may consider extracting this to a separate function to improve readability

abulenok, line 45: Maybe let's remove the comment? the code seems to be clear

crates/cast/src/helpers/scarb_utils.rs
```diff
@@ -153,41 +146,23 @@ pub fn get_package_metadata<'a>(
     Ok(package)
 }
 
-pub fn parse_scarb_config(
+pub fn get_scarb_config(
```
abulenok, line 149: Let's put this to `impl CastConfig`, rename it to `from_package_metadata` and make it return `Option`. Then we could do `unwrap_or_default` on it. 
WDYT?

crates/cast/src/main.rs
```diff
@@ -372,3 +410,59 @@ fn update_cast_config(config: &mut CastConfig, cli: &Cli) {
...
+    }
+
+    let package = match cli
+        .packages_filter
+        .clone()
+        .into_packages_filter()
+        .match_one(result.metadata.as_ref().unwrap())
+    {
+        Ok(data) => Ok(data),
+        Err(_) => get_first_package_from_metadata(result.metadata.as_ref().unwrap()),
```
abulenok, line 463: Is this guaranteed by scarb for the main package to be the first in the metadata list?

crates/cast/src/main.rs
```diff
@@ -111,25 +121,44 @@ fn main() -> Result<()> {
...
+        Err(_) => get_scarb_config(&cli.profile, None)?,
+    };
+
     update_cast_config(&mut config, &cli);
 
     let provider = get_provider(&config.rpc_url)?;
     let runtime = Runtime::new().expect("Could not instantiate Runtime");
 
     if let Commands::Script(script) = cli.command {
+        package_data.manifest_path?;
```
abulenok, line 139: I think this line is confusing

crates/cast/src/main.rs
```diff
@@ -372,3 +410,59 @@ fn update_cast_config(config: &mut CastConfig, cli: &Cli) {
 
     config.accounts_file = Utf8PathBuf::from(shellexpand::tilde(&new_accounts_file).to_string());
 }
+
+struct PackageData {
+    pub manifest_path: Result<Utf8PathBuf>,
+    pub metadata: Result<Metadata>,
+    pub package: Result<PackageMetadata>,
+}
```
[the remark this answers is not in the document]
Radinyn: In order to propagate the errors from deeper parts of the code so the messages are more descriptive

crates/cast/src/main.rs
Radinyn: This is the way it worked to this date, so it's not breaking anything, I think it's a sensible strategy

crates/cast/src/main.rs
Radinyn: It pops the first error it encounters, not sure how to do it better I think it makes sense as a missing path is definitely a root cause of missing metadata

crates/cast/src/main.rs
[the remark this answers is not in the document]
MaksymilianDemitraszek: If there is an error you probably should either merge them to one or propagate only the first one and use standard approach with returning Result<x>.

crates/cast/src/helpers/scarb_utils.rs
Radinyn: We would lose our error propagating code in that case as well

crates/cast/src/main.rs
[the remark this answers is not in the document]
MaksymilianDemitraszek: I also don't think such check is necessary

crates/cast/src/helpers/build.rs
```diff
@@ -0,0 +1,61 @@
...
+use scarb_metadata::{Metadata, PackageMetadata};
+use scarb_ui::args::PackagesFilter;
+use std::collections::HashMap;
+use std::fs;
+use std::process::{Command, Stdio};
+
+use super::constants::LIB_CONTRACT_ARTIFACTS_NAME;
+
+#[allow(clippy::too_many_lines)]
+pub fn build(
```
MaksymilianDemitraszek, line 12: It not only builds but also loads them to memory, please split it into two separate functions.

crates/cast/src/main.rs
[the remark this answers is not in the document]
Radinyn: The problem is that those are all needed in their "maybe" form at the beginning but needing just one doesn't necessarily mean that you need the others

crates/cast/src/main.rs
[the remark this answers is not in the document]
MaksymilianDemitraszek: Discussed it with @piotmag769 and I got persuaded we should include that.

MaksymilianDemitraszek (commented):
Partial review, I will be revieing it further

crates/cast/src/main.rs
piotmag769: Is it agreed that we should always take the config from the first workspace member if there is no package specified? What if `--workspace` was specified (I guess `match_one` fails, right?)? 
cc @THenry14

crates/cast/src/starknet_commands/script.rs
```diff
@@ -350,20 +348,19 @@ impl CairoHintProcessor<'_> {
 
 pub fn run(
     module_name: &str,
-    path_to_scarb_toml: &Option<Utf8PathBuf>,
     provider: &JsonRpcClient<HttpTransport>,
     runtime: Runtime,
     config: &CastConfig,
+    contracts: &HashMap<String, StarknetContractArtifacts>,
```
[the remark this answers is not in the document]
piotmag769: Looking at the type it would suggest that this hashmap stores the mapping to contracts. I assume it `StarknetContractArtifacts` was reused to store the compiled script code too? If that's the case it should be refactored

crates/cast/src/helpers/build.rs
```diff
@@ -0,0 +1,61 @@
...
+        let sierra_filename = format!("{}.sierra.json", package.name);
+
+        let sierra_path = target_dir_for_package(metadata)
+            .join(&metadata.current_profile)
+            .join(sierra_filename);
+
+        let lib_artifacts = StarknetContractArtifacts {
+            sierra: fs::read_to_string(sierra_path)?,
+            casm: String::new(), // There seems to be no need for casm in lib
+        };
```
piotmag769, line 55: Please just define a separate structure for non-starknet atrifacts

crates/cast/src/helpers/scarb_utils.rs
```diff
@@ -200,21 +175,38 @@ pub fn get_package_tool_sncast(metadata: &scarb_metadata::Metadata) -> Result<&V
     Ok(tool_sncast)
 }
 
+pub fn get_first_package_from_metadata(metadata: &Metadata) -> Result<PackageMetadata> {
```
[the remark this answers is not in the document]
Radinyn: It omits those packages

crates/cast/src/main.rs
```diff
@@ -372,3 +410,59 @@ fn update_cast_config(config: &mut CastConfig, cli: &Cli) {
...
+
+// Lazily get all the required package information from CLI
+fn get_package_data(cli: &Cli) -> PackageData {
+    let mut result = PackageData {
```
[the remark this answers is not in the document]
Radinyn: The problem if it fails for example on metadata step then we have no real way to get the package metadata, those early returns help to avoid nested matching

crates/cast/src/main.rs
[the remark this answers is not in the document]
MaksymilianDemitraszek: If there is a scenario that you need only one then those should be a separate functions

crates/cast/src/helpers/build.rs
Radinyn: I'm not entirely sure about that. I think it makes sense for a single function to return the artifacts that it built

crates/cast/src/main.rs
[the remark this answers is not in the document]
piotmag769: :c kinda right

crates/cast/src/main.rs
piotmag769: It doesn't look like a good design decision to me. It is basically taking a config of a random workspace member and then e.g. sending transactions using it. In the following scenario:

```
crates/
  A/
    Scarb.toml
  B/
    Scarb.toml
Scarb.toml
```

if I use `sncast` from `crates` directory I basically don't know, as the user, from what crate my config will be taken from. Imo we just shouldn't read the config if it is not precisely specified. By "precisely specified I mean:
- `sncast -p A`
- `sncast --manifest-path crates/A/Scarb.toml` 
- `sncast` called from A/ or A/src etc.

All of the cases should be tested too (prob they are, didn't look into this PR that carefully)

THenry14: the only scenarios when we need to think about scarb and its directory layout are scripts and declare.

For scripts we need to pass package name, which (afaik) should be unique in the setup you've described (please, correct me if I'm wrong).
For declare, we need to provide the contract name - now, I'm not sure about it's uniqueness, this indeed may be problematic in some cases.

Cast allow to pass manifest path to it (`--path-to-scarb-toml`), you can of course always call it from specific dirs.

Given all this, I understand your concern about taking just first package - if I remember correctly this was something suggested to us by scarb team way back when there was no workspace support in scarb. Now that things changed, we should probably indeed take different approach.

I would suggest that (for declare) we accept an argument 'package_name' that is required if there are more packages in workspace than 1. That way we should ensure contracts from proper packages are always taken. Does that sound good?

Radinyn: Discussed on the meeting

crates/cast/src/main.rs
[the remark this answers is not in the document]
THenry14: My understanding is that there is a consensus that we should have this check, although I still don't think this is the right place - we should perform this only in scenarios where scarb is required - and for most subcommands, it isn't

crates/cast/src/helpers/build.rs
THenry14: I guess we could extract scripts part to separate function and only call it in scripts, that would be probably more optimal

crates/cast/tests/helpers/fixtures.rs
```diff
@@ -226,6 +226,23 @@ pub fn duplicate_directory_with_salt(src_path: String, to_be_salted: &str, salt:
     temp_dir
 }
 
+#[must_use]
+pub fn duplicate_directory(src_path: String) -> TempDir {
```
MaksymilianDemitraszek, line 230: probably copy_directory_to_tempdir?

MaksymilianDemitraszek (changes requested):
I generally don't understand `--package` filter from the perspective of cast. What purpose does it serve exactly?

MaksymilianDemitraszek:
Also we probably shouldn't take the first match, we should throw an error when it is ambigous

MaksymilianDemitraszek:
We discussed it with @THenry14 , filter mechanism does not make much sense in the context of scripts.
Instead we should be able to pass module path to script instead of a name:
`script package_name::path::to::script_module`

If only name is passed and it is ambigous in the context of workspace we should print something like this:
```
package_namea::path::to::script_module
package_nameb::path::to::script_module
package_nameb::path2::to::script_module
```

Passing `to::script_module ` should be also possible

crates/cast/src/main.rs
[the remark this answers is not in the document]
piotmag769: @MaksymilianDemitraszek we need it for every command basically

crates/cast/src/main.rs
```diff
@@ -27,13 +33,16 @@ mod starknet_commands;
...
+    #[clap(short = 'p', long)]
     profile: Option<String>,
 
-    /// Path to Scarb.toml that is to be used; overrides default behaviour of searching for scarb.toml in current or parent directories
+    /// Path to Scarb.toml that is to be used; overrides default behaviour of searching for Scarb.toml in current or parent directories
     #[clap(short = 's', long)]
     path_to_scarb_toml: Option<Utf8PathBuf>,
 
+    #[command(flatten)]
+    packages_filter: PackagesFilterLong,
```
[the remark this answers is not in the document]
piotmag769: This is a filter from scarb-ui package, it makes the underlying flags optional

crates/cast/src/main.rs
```diff
@@ -27,13 +33,16 @@ mod starknet_commands;
 #[allow(clippy::struct_excessive_bools)]
 struct Cli {
     /// Profile name in Scarb.toml config file
-    #[clap(short, long)]
+    #[clap(short = 'p', long)]
```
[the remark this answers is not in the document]
piotmag769: Is it? I think it is to override the `-p` option as a short option for `--package` (not sure how clap handles 2 flags with the same short form)

piotmag769:
> I generally don't understand --package filter from the perspective of cast. What purpose does it serve exactly?

@MaksymilianDemitraszek `--package` makes sense, you can choose an explicit package that config will be taken from etc.. I don't understand what `--workspace` does. It was decided that `--workspace` will be disabled, right? I don't see it explicitly disabled, looks like you can still pass it for a workspace with a single package (`match_one` won't fail).

Also what config from Scarb.toml takes priority when both `--manifest-path` and `--package` are both passed? Should it be even allowed? (currently it is from the CLI point of view)

crates/sncast/src/helpers/build.rs
```diff
@@ -0,0 +1,57 @@
...
+    let mut artifacts = get_contracts_map(metadata, &package.id)?;
+
+    if for_scripts {
+        let sierra_filename = format!("{}.sierra.json", package.name);
+
+        let sierra_path = target_dir_for_package(metadata)
+            .join(&metadata.current_profile)
+            .join(sierra_filename);
+
+        let lib_artifacts = StarknetContractArtifacts {
```
piotmag769, line 48: Storing it together with `StarknetContractArtifacts` is kinda a hack, should be stored in a separate struct prob

crates/sncast/src/helpers/scarb_utils.rs
```diff
@@ -276,23 +259,40 @@ pub fn get_package_tool_sncast(metadata: &scarb_metadata::Metadata) -> Result<&V
     Ok(tool_sncast)
 }
 
+pub fn get_first_package_from_metadata(metadata: &Metadata) -> Result<PackageMetadata> {
```
piotmag769, line 262: It terrifies me (I know this logic has been there before) xD Is it only used in tests? If so can we move it to the tests mod?

piotmag769 (commented):
Good bye 👋 👽 😢 🌹 
image

crates/sncast/src/helpers/scarb_utils.rs
Radinyn: It's used in tests in multiple modules

crates/sncast/src/helpers/build.rs
abulenok: maybe introduce a type alias?

PR #1095, opened 2023-11-10
Mock sequencer address

### LINKED ISSUE #609: Cheatcode for modifying `BlockInfo::sequencer_address`
It can be mocked via spoof (not obligatory)

drknzz:
Closes #609 

## Introduced changes

- add cheatcode for mocking the sequencer address
- the `elect` codename is up for renaming

## Checklist

- [x] Linked relevant issue
- [x] Updated relevant documentation - #1120
- [x] Added relevant tests - #1118


piotmag769 (commented):
lgtm, but please add tests and docs

drknzz:
> lgtm, but please add tests and docs

Tets and docs will be 2 seperate PRs

piotmag769:
> Tests and docs will be 2 seperate PRs to keep them reviewable

Usually we do everything in the same PR 🤔

crates/cheatnet/src/cheatcodes/elect.rs
```diff
@@ -0,0 +1,17 @@
+use crate::CheatnetState;
+use starknet_api::core::ContractAddress;
+
+impl CheatnetState {
```
war-in, line 4: Should we maybe add a target for this cheatcode too? 🤔

war-in (approved):

crates/cheatnet/src/cheatcodes/elect.rs
drknzz: Sure, but let's do it in another PR 👍

Arcticae (commented):
Looks good, but please do include docs & tests as well. 
I appreciate the effort to keep those reviewable though ❤️

docs/src/appendix/cheatcodes/start_elect.md
```diff
@@ -0,0 +1,47 @@
+# `start_elect`
+
+> `fn start_elect(contract_address: ContractAddress, sequencer_address: ContractAddress)`
+
+Changes the sequencer address for a contract at the given address.
+The change can be canceled with [`stop_elect`](./stop_elect.md).
+
+- `contract_address` - target contract address
```
piotmag769, line 8: Actually does it make sense to implement it this way if we already have global implementations of syscalls? Shouldn't it take `CheatTarget` as well as the other cheatcodes?

docs/src/testing/using-cheatcodes.md
```diff
@@ -268,3 +268,24 @@ fn mock_constructor_with_warp() {
     assert(block_timestamp == 1000, 'Wrong block timestamp');
 }
 ```
+
+### Mocking the constructor with `elect`
```
piotmag769, line 272: I know it's out of scope, but do we need that many examples? I think prank one is enough.
Also the examples use the outdated (non-global) interface of cheatcodes

drknzz: +1, I don't think we need that many, I'm down to leave just prank

docs/src/appendix/cheatcodes/start_elect.md
drknzz: There's #1099, since the work on this has been started before the first PR with `CheatTarget` was merged. I'll probably update the whole PR to use `CheatTarget` once the tests for this are merged (#1118).

piotmag769: Wouldn't it make additional tests needed?

PR #1096, opened 2023-11-10
Remove predeployed contracts

### LINKED ISSUE #974: Remove predeployed contracts
I think we don't use them anymore as we are on the execution layer only

war-in:
Closes #974 

## Checklist

- [ ] Linked relevant issue
- [ ] Added relevant tests


drknzz (approved):
➕ 1️⃣ 4️⃣ ➖ 9️⃣ 1️⃣ 8️⃣ 0️⃣ 👍

crates/forge/src/main.rs
```diff
@@ -296,6 +269,7 @@ fn main() {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use camino::Utf8PathBuf;
```
cptartur, line 272: Is this needed?

cptartur (approved):
😎

crates/forge/src/main.rs
war-in: yup, it was moved from the section above since Utf8PathBuf is now used only in the test

Arcticae (approved):

PR #1112, opened 2023-11-14
adds action to check cairo code

tkumor3:
Closes #

## Introduced changes

-

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


cptartur (commented):
Looks fine to me, but I'd let @THenry14 review the script

.github/workflows/ci.yml
```diff
@@ -91,6 +91,21 @@ jobs:
...
+      - uses: software-mansion/setup-scarb@v1.3.2
+      - name: Check cairo files format
+        run: |
+          find . -type f -name "Scarb.toml" -print0 | while IFS= read -r -d '' file; do
+            dir="$(dirname "$file")"
+            cd "$dir"
+            echo "Running 'scarb fmt' in directory: $PWD"
+            scarb fmt --check
+            cd - > /dev/null
+          done
```
THenry14, line 107: This looks promising, but have some issues:
- `cd` is pretty error prone, a good practice woule be to add `|| exit 1`
- not sure if this is by design or not, but it does not indicate errors with format - which I suppose it should. what I mean by that is that when there are issues with code formating, we are not failing the test but rather let it pass.

how about we do something like this instead?

```suggestion
        output=$(find . -type f -name "Scarb.toml" -execdir sh -c '
          echo "Running \"scarb fmt\" in directory: $PWD"
          scarb fmt --check
        ' \;)
        echo "$output"
        if grep -iq "Diff" <<< "$output"; then
          exit 1
        fi
        exit 0
```

This fixes all the problems I mentioned + gets rid of while loop, for a price of an additional "echo" to show the diff, but I don't think it's a high price. wdyt?

```

Comment (tkumor3) [reply .github/workflows/ci.yml]: @THenry14 Ok, I replace my script with yours ;)

File: .github/workflows/ci.yml [hunk 2]
```diff
@@ -91,6 +91,23 @@ jobs:
...
+        run: |
+          output=$(find . -type f -name "Scarb.toml" -execdir sh -c '
+              echo "Running \"scarb fmt\" in directory: $PWD"
+              scarb fmt --check
+          ' \;)
+          echo "$output"
+          if grep -iq "Diff" <<< "$output"; then
+              exit 1
+          fi
+              exit 0
```
THenry14, line 109:
```suggestion
          exit 0
```
nitpick, but let's keep it properly formatted :+1:

THenry14 (approved):

PR #1104, opened 2023-11-10
Automate class hash computation in tests

drknzz:
Closes #929 

## Introduced changes

-

## Checklist

- [x] Linked relevant issue
- [x] Updated relevant documentation
- [x] Added relevant tests
- [x] Performed self-review of the code
- [x] Added changes to `CHANGELOG.md`

PR #1121, opened 2023-11-14
Decouple from Scarb

piotmag769:
Related scarb PR: https://github.com/software-mansion/scarb/pull/886

## Introduced changes

- moved test collection and compilation to Scarb

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


piotmag769:
Tests won't pass bcs they require Scarb with `snforge-test-collector` xD

crates/cheatnet/tests/cheatcodes/declare.rs
```diff
@@ -7,9 +7,9 @@ use starknet_api::hash::StarkFelt;
 use starknet_api::stark_felt;
 
 static HELLO_STARKNET_CLASS_HASH: &str =
-    "0x029c29c28d7eed24e3e57e6189cda4fc5a7a5f1d097f02c6692cff91816a5fbc";
+    "0x04ac1b5bcf427bef9eb31691de4104268179e26df6e81eb93492870c3e7ca9cb";
 static CONSTRUCTOR_SIMPLE_CLASS_HASH: &str =
-    "0x045b3e15b898999ca953f197065fb4b4f98af6e92fb012bb26917a8020c3314d";
+    "0x06aa43ff1d4a13360a7ec9ab34924a735adee7fd42042b8d4f910d8386dce556";
```
cptartur, line 12: Why did these change?

crates/forge/src/collecting.rs
cptartur, line 1: This file should probably be renamed

crates/forge/src/lib.rs
```diff
@@ -216,21 +209,22 @@ fn to_runnable(
...
 /// * `contracts` - Map with names of contract used in tests and corresponding sierra and casm artifacts
-/// * `predeployed_contracts` - Absolute path to predeployed contracts used by starknet state e.g. account contracts
 ///
 
 #[allow(clippy::implicit_hasher)]
 pub async fn run(
-    package_path: &Utf8Path,
     package_name: &str,
-    package_source_dir_path: &Utf8Path,
+    snforge_target_dir_path: &Utf8Path,
```
cptartur, line 217: Why not just `target_dir_path`

cptartur: It's in the context of this crate anyways.

crates/forge/src/main.rs
```diff
@@ -166,6 +167,17 @@ fn test_workspace(args: TestArgs) -> Result<bool> {
...
+    let test_build_output = Command::new("scarb")
+        .arg("snforge-test-collector")
+        .env("SCARB_PACKAGES_FILTER", filter_env)
+        .stdout(Stdio::inherit())
+        .stderr(Stdio::inherit())
+        .output()
+        .context("Failed to build test artifacts with Scarb")?;
+    if !test_build_output.status.success() {
+        bail!("scarb snforge-test-collector did not succeed")
+    }
```
Arcticae, line 179: this deserves own function

crates/forge/src/test_filter.rs
```diff
@@ -99,8 +98,9 @@ mod tests {
     use crate::collecting::CompiledTestCrate;
     use crate::test_filter::TestsFilter;
     use crate::CrateLocation;
+    use crate::TestCase;
     use cairo_lang_sierra::program::Program;
-    use test_collector::{ExpectedTestResult, TestCase};
+    use cairo_lang_test_plugin::test_config::TestExpectation;
```
Arcticae, line 103: How is this change related exactly?

crates/forge/src/lib.rs
```diff
@@ -216,21 +209,22 @@ fn to_runnable(
...
     runner_params: Arc<RunnerParams>,
 ) -> Result<Vec<TestCrateSummary>> {
-    let compilation_targets =
-        collect_test_compilation_targets(package_path, package_name, package_source_dir_path)?;
-    let test_crates = compile_tests(&compilation_targets, &runner_params)?;
+    let snforge_test_artifact_path =
+        snforge_target_dir_path.join(format!("{package_name}.sierra.json"));
+    let test_crates = serde_json::from_str::<Vec<CompiledTestCrateRaw>>(&std::fs::read_to_string(
+        snforge_test_artifact_path,
+    )?)?;
```
cptartur, line 226: If the artefacts compilation is going on in scarb now, perhaps this function should already accept them in the deserialized (as structs/vec/whatever rust object) instead of taking the path. 

This way we can still use it in integration tests in an easier way.

crates/forge/test_utils/src/runner.rs
```diff
@@ -99,6 +105,33 @@ impl<'a> TestCase {
 
         dir.child("src/lib.cairo").touch().unwrap();
 
+        let snforge_std_path = Utf8PathBuf::from_str("../../snforge_std")
+            .unwrap()
+            .canonicalize_utf8()
+            .unwrap()
+            .to_string()
+            .replace('\\', "/");
```
cptartur, line 113: Why are we replacing this?

crates/forge/tests/integration/dispatchers.rs
```diff
@@ -576,7 +576,7 @@ fn nonexistent_method_call() {
     assert_case_output_contains!(
         result,
         "test_nonexistent_method_call",
-        "Entry point selector 0x01fdb214e1495025fa4baf660d34f03c0d8b5037cf10311d2a3202a806aa9485 not found in contract 0x00c0727f111a1f9fd7543cf36ae25feaa8184a9188daa4c9c148605445dcba9f"
+        "Entry point selector 0x01fdb214e1495025fa4baf660d34f03c0d8b5037cf10311d2a3202a806aa9485 not found in contract 0x0612c9ea6cdc79914ba9eae055225ee9df90368b02c9e463247d6c2d313000bb"
```
Arcticae, line 579: What did cause that bump?

crates/forge/tests/e2e/running.rs
```diff
@@ -614,6 +614,7 @@ fn init_new_project_test() {
             [package]
             name = "test_name"
             version = "0.1.0"
+            edition = "2023_10"
```
Arcticae, line 617: Is this now a required parameter?

crates/forge/test_utils/src/runner.rs
```diff
@@ -99,6 +105,33 @@ impl<'a> TestCase {
...
+                [package]
+                name = "test_package"
+                version = "0.1.0"
+    
+                [[target.starknet-contract]]
+                sierra = true
+                casm = true
+    
+                [dependencies]
+                starknet = "2.3.1"
```
cptartur, line 128: This will have to be manually updated with every cairo bump which we'll probably forget about. Perhaps this value should be taken from top-level `.tool-versions` file, assuming scarb will have the same versions as starknet.

crates/forge/tests/integration/spy_events.rs
```diff
@@ -428,7 +428,7 @@ fn event_emitted_wrong_data_asserted() {
     assert_case_output_contains!(
         result,
         "test_assert_wrong_data",
-        "2814149079029271315825228546613746536844456515452335425622822705028162471071"
+        "385731233638135702203933925049152883559461865803221708830049339625617856128"
```
cptartur, line 431: Why did it change?

crates/scarb-artifacts/src/lib.rs
```diff
@@ -164,24 +171,20 @@ pub fn corelib_for_package(metadata: &Metadata, package: &PackageId) -> Result<U
...
-
     let package = metadata
         .get_package(package)
         .ok_or_else(|| anyhow!("Failed to find metadata for package = {package}"))?;
 
-    let package_path = package.root.clone();
-    let package_source_dir_path = compilation_unit.target.source_root();
-
-    Ok((package_path, Utf8PathBuf::from(package_source_dir_path)))
+    Ok((package.root.clone(), package.manifest_path.clone()))
```
cptartur, line 183: Was this only used in forge? So we don't accidentally break some behavior in cast.

crates/forge/tests/e2e/running.rs
```diff
@@ -759,12 +760,10 @@ fn available_gas_error() {
     let snapbox = test_runner();
 
     let output = snapbox.current_dir(&temp).assert().failure();
-    assert_stdout_contains!(
+    assert_stderr_contains!(
```
cptartur, line 763: Why did it only change here while other diagnostics are still printed to stdout?

crates/cheatnet/tests/cheatcodes/declare.rs
drknzz: +1, although these are computed automatically on master ([#1104](https://github.com/foundry-rs/starknet-foundry/pull/1104))

crates/forge/src/collecting.rs
```diff
@@ -36,169 +35,34 @@ pub(crate) struct TestCompilationTarget {
...
-    let tests_dir_path = package_path.join("tests");
-    if tests_dir_path.exists() {
-        compilation_targets.push(TestCompilationTarget {
-            crate_name: "tests".to_string(),
-            crate_location: CrateLocation::Tests,
-            lib_content: get_or_create_test_lib_content(tests_dir_path.as_path())?,
-            crate_root: tests_dir_path,
-        });
-    }
+pub trait ForkConfig {}
```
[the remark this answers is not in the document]
piotmag769: Why? It was a conscious decision, it is copied from somewhere else. It is to parametrisation of `TestCase<T>` with random type - the type has to implement this empty trait

crates/forge/src/collecting.rs
```diff
@@ -36,169 +35,34 @@ pub(crate) struct TestCompilationTarget {
...
 
-    Ok(compilation_targets)
-}
+impl ForkConfig for RawForkConfig {}
```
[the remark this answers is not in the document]
piotmag769: Above^

crates/forge/src/lib.rs
```diff
@@ -216,21 +209,22 @@ fn to_runnable(
 /// * `runner_config` - A configuration of the test runner
 /// * `corelib_path` - Absolute path to the Cairo corelib
 /// * `contracts` - Map with names of contract used in tests and corresponding sierra and casm artifacts
-/// * `predeployed_contracts` - Absolute path to predeployed contracts used by starknet state e.g. account contracts
 ///
 
 #[allow(clippy::implicit_hasher)]
 pub async fn run(
```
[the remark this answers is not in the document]
piotmag769: You mean the comment? Agree, will do

crates/forge/src/main.rs
piotmag769: Then we want `scarb build` in its own function too?

crates/forge/src/lib.rs
piotmag769: Disagree, with `target_dir_path` you don't really know if it is a path to `target` or `target/dev/snforge`

crates/forge/test_utils/src/runner.rs
maciektr: `\\` are used on windows, this unifies paths across systems

crates/forge/tests/e2e/running.rs
maciektr: It's optional, if not filled in it defaults to an edition set as default on compiler side - currently `2023_01`

crates/forge/test_utils/src/runner.rs
piotmag769: Out of scope + we should stop updating it manually in all the other places anyways, it should work until a new Cairo (and thus Scarb) major is released.

Imagine if we did not bump it in `snforge_std` for 2.3.1. You will have `starknet=2.3.0` specified, which is the same as `starknet=^2.3.0`, so it would still work with both Scarb 2.3.1 and 2.3.0. Now we basically limited our users to use Scarb 2.3.1 only.

Summary: until a new major of Scarb/Cairo is released, updating starknet dependency in Scarb.tomls for tests is not harmful, but useless. Updating it in `snforge_std` is limiting the range of compatible Scarb versions (we should stop doing that).

crates/forge/src/lib.rs
piotmag769: This would put even more logic into `main.rs`, I don't really like that. Also it doesn't simplify integration tests imo, we don't want to generate them manually for tests anyways

crates/forge/tests/e2e/running.rs
piotmag769: If not specified it uses the default, but this is the expected toml content. We use `scarb new` which generates it in the Scarb.toml of the new package

crates/forge/tests/integration/dispatchers.rs
piotmag769: I use the Scarb built locally for testing, it has a different compiler version

crates/cheatnet/tests/cheatcodes/declare.rs
piotmag769: https://github.com/foundry-rs/starknet-foundry/pull/1121#discussion_r1394072646

piotmag769: Will rebase on master

crates/forge/tests/integration/spy_events.rs
piotmag769: https://github.com/foundry-rs/starknet-foundry/pull/1121#discussion_r1394072646

crates/scarb-artifacts/src/lib.rs
piotmag769: Yup, it was only used in forge

crates/forge/tests/e2e/running.rs
piotmag769: It is not a diagnostic, it is a custom error of ours. It comes from stderr of `scarb-snforge-test-collector` which we inherit.

In case of erros caught directly in forge we print them `fn main` in `forge/src/main.rs` to the stdout (don't really know why, I wanted to change it someday but remember we decided not to do it).

crates/forge/test_utils/src/runner.rs
maciektr: Yes, https://docs.swmansion.com/scarb/docs/reference/specifying-dependencies.html#version-requirements

crates/forge/test_utils/src/runner.rs
maciektr: A comment would be nice though

crates/forge/src/main.rs
```diff
@@ -166,6 +167,17 @@ fn test_workspace(args: TestArgs) -> Result<bool> {
         bail!("Scarb build did not succeed")
     }
 
+    let test_build_output = Command::new("scarb")
```
maciektr, line 170: If you want to rely on scarb profile chose by user (line 149), you should also pass it to scarb on build cmd via `SCARB_PROFILE` env var. https://docs.swmansion.com/scarb/docs/writing-extensions/subcommands.html#environment-variables

crates/forge/src/test_filter.rs
piotmag769: Oh wait, those shouldn't be here, left from the first version xD Thx

crates/forge/src/lib.rs
cptartur: This doesn't necessarily have to be in main, it can be a separate function in lib that would be called before `run`. Similarly to how collecting and compilation was split into multiple functions called in a sequence

crates/forge/test_utils/src/runner.rs
cptartur: Ideally we'd test it across some range of values, because this approach only ensures forge works with a specific version. Even if, in theory, it should work with multiple.

crates/forge/src/lib.rs
cptartur: It's not immediately obvious that it is meant to be a path to `target/dev/snforge` with this name either. Target dir could also suggest a directory for which you want to run tests i.e. scarb package.

crates/forge/tests/e2e/running.rs
cptartur: It could be a weird from the perspective of use to show these errors like this, without an `[ERROR]` tag before the message or whatever. 

It may be fine if it's just `available_gas` but we should consider some approach of displaying the collector errors in manner more consistent with the rest of the forge.

crates/forge/src/lib.rs
piotmag769: I think we just want to replace the collection + compilation logic here, no? So it would make sense to read it here I think, I can extract it to a separate function that with a docstring explaining what it does

cptartur: This seems like a good middle ground for now, we can do that.

crates/forge/src/main.rs
piotmag769: I guess every if every scarb process uses this env var it should be fine

maciektr: Yes, we do

crates/forge/tests/e2e/running.rs
piotmag769: Minor thing imo, you can add an issue if you want

cptartur: Quite a change in UX in my opinion, and annotations and errors emitted by them are already limited.

crates/forge/src/lib.rs
piotmag769: I think it's obvious what target directory is considering we are in Rust-like environment

crates/forge/tests/e2e/diagnostics_and_plugins.rs
```diff
@@ -80,11 +80,15 @@ fn print_error_if_attributes_incorrect() {
         #[fuzzer()]
                 ^^
         
-        error: Plugin diagnostic: Expected panic must be of the form `expected: <tuple of felt252s>`.
+        error: Plugin diagnostic: Expected panic must be of the form `expected: <tuple of felt252s and strings>` or `expected: "some string"` or `expected: <some felt252>`.
+         --> {mock_tests_dir_path_str}/tests/test_should_panic.cairo:2:15
+        #[should_panic(url: "https://test.com")]
+                      ^***********************^
+        
+        error: Plugin diagnostic: Expected panic must be of the form `expected: <tuple of felt252s and strings>` or `expected: "some string"` or `expected: <some felt252>`.
```
piotmag769, line 88: #1128

CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [0.11.0-rc.0] - 2023-11-17
+
```
piotmag769, line 11: Will be reverted, made release for testing purposes

Cargo.toml
```diff
@@ -2,15 +2,14 @@
...
-    "crates/test-collector",
     "crates/cast",
     "crates/cheatnet",
     "crates/conversions",
     "crates/scarb-artifacts",
 ]
 
 [workspace.package]
-version = "0.10.2"
+version = "0.11.0-rc.0"
```
piotmag769, line 12: Same here ^

crates/forge/src/lib.rs
```diff
@@ -206,31 +199,38 @@ fn to_runnable(
...
+
 /// Run the tests in the package at the given path
 ///
 /// # Arguments
 ///
-/// * `package_path` - Absolute path to the top-level of the Cairo package
-/// * `lib_path` - Absolute path to the main file in the package (usually `src/lib.cairo`)
-/// * `linked_libraries` - Dependencies needed to run the package at `package_path`
+/// * `package_name` - Name of the package specified in Scarb.toml
+/// * `snforge_target_dir_path` - Absolute path to the directory with snforge test artifacts (`{package_path}/target/{profile_name}/snforge`)
```
cptartur, line 220:
```suggestion
/// * `snforge_target_dir_path` - Absolute path to the directory with snforge test artifacts (Usually `{package_path}/target/{profile_name}/snforge`)
```

crates/forge/test_utils/src/runner.rs
```diff
@@ -99,6 +112,33 @@ impl<'a> TestCase {
 
         dir.child("src/lib.cairo").touch().unwrap();
 
+        let snforge_std_path = Utf8PathBuf::from_str("../../snforge_std")
+            .unwrap()
+            .canonicalize_utf8()
+            .unwrap()
+            .to_string()
+            .replace('\\', "/");
```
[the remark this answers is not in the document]
piotmag769: It is in other 8 places in our codebase too, do we really need that comment?

crates/forge/src/scarb.rs
```diff
@@ -29,10 +31,43 @@ pub fn config_from_scarb_for_package(
...
+        .env("SCARB_PACKAGES_FILTER", filter.to_env())
+        .stderr(Stdio::inherit())
+        .stdout(Stdio::inherit())
+        .output()
+        .context("Failed to build test artifacts with Scarb")?;
+
+    if build_output.status.success() {
+        Ok(())
+    } else {
+        Err(anyhow!("scarb snforge-test-collector did not succeed"))
```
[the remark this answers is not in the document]
piotmag769: It is a binary call name i.e. `scarb snofrge-test-collector`

crates/forge/test_utils/src/runner.rs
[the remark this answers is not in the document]
maciektr: It doesn't sound like something that should be copied to 8 independent places, but if it is - yes, you need comments, it's beyond confusing.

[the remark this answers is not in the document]
piotmag769: This replace happens in other places only in tests to make them work for Windows contributors

PR #1129, opened 2023-11-15
Fix CheatTarget behaviour

### LINKED ISSUE #1122: Fix the behavior of mock targets
Desired behavior, is that a cancel (on single or a set) when `global` is in effect, will leave rest mocked and only those unmocked that were specified. This set of unmocks can be further enlargened, by later unmocks, and smallened by same or different mock values.

Examples:

```
start(all, x)    // (123 == x)
start(123, y)    // (123 == y)
stop(all)        // (123 == default)
```

```
start(123, x)   // (123 == x)
start(all, y)   // (123 == y)
stop(123)       // (123 == default)
```

```
start(123, x)   // (123 == x) (321 == default)
start(all, y)   // (123 == y) (321 == y)
stop([123, 321])       // (123/321 == default)
start(321, z)   // (123 == default) (321 == z)
```

drknzz:
Closes #1122

### Description of the bug:
- All 3 of `address_is_warped` / `address_is_rolled` / `address_is_pranked` had a flaw in logic, where they'd falsely return `true` after `start_cheat(CheatTarget::All)` and `stop_cheat(CheatTarget::One)`
- `start_cheat(CheatTarget::All)` would set the (in the `prank` case for example) `global_prank` field to Some(value)
- `stop_cheat(CheatTarget::One(address))` would insert `(address, CheatStatus::Uncheated)` into `pranked_contracts` hashmap
- `is_address_pranked` would now return `true`, since it'd check first if [`global_prank.is_some()`](https://github.com/foundry-rs/starknet-foundry/blob/9987ecdd88b82f1547c999aa4e8d918dc86c4579/crates/cheatnet/src/state.rs#L248)

Tests for that case have been added for all 3 of the cheats, however only the behaviour of `prank` turned out to be wrong.

This is because unlike for the `prank` cheat, entering the if statement for [`roll` and `warp`](https://github.com/foundry-rs/starknet-foundry/blob/9987ecdd88b82f1547c999aa4e8d918dc86c4579/crates/cheatnet/src/execution/execution_info.rs#L108) doesn't change the outcome, since what `get_cheated_block_info_ptr` does later inside the if, is [override `original_block_info`](https://github.com/foundry-rs/starknet-foundry/blob/9987ecdd88b82f1547c999aa4e8d918dc86c4579/crates/cheatnet/src/execution/execution_info.rs#L118C18-L118C37) with the cheated values.

[For `prank` however](https://github.com/foundry-rs/starknet-foundry/blob/9987ecdd88b82f1547c999aa4e8d918dc86c4579/crates/cheatnet/src/execution/execution_info.rs#L138) it's not the case, and an [expect](https://github.com/foundry-rs/starknet-foundry/blob/9987ecdd88b82f1547c999aa4e8d918dc86c4579/crates/cheatnet/src/execution/execution_info.rs#L142) provokes the panic.

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/cheatnet/src/state.rs
```diff
@@ -270,11 +273,11 @@ impl CheatnetState {
 
     #[must_use]
     pub fn address_is_rolled(&self, contract_address: &ContractAddress) -> bool {
-        self.global_roll.is_some()
-            || matches!(
-                self.rolled_contracts.get(contract_address),
-                Some(CheatStatus::Cheated(_))
-            )
+        match self.rolled_contracts.get(contract_address) {
```
Arcticae, line 276: can we generalise this into a function?

Arcticae (approved):

PR #1145, opened 2023-11-17
remove send_shut_down

tkumor3:
Closes #

## Introduced changes

-

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


piotmag769 (approved):

MaksymilianDemitraszek:
Why it is not needed anymore, what has changed?

piotmag769:
@MaksymilianDemitraszek I guess before the last major refactors we had a `break` in this loop if we encountered `Failed` when using `--exit-first`

https://github.com/foundry-rs/starknet-foundry/blob/master/crates/forge/src/lib.rs#L366

Without this break we don't need to wait for any task to complete, as they are all guaranteed to complete anyways (we always loop through every task)

cptartur (approved):

PR #1146, opened 2023-11-17
[draft][scripts] Add declare

THenry14:
Closes #

## Introduced changes

-

## Checklist

- [ ] Linked relevant issue
- [ ] Added relevant tests


sncast_std/src/lib.cairo
```diff
@@ -39,3 +39,28 @@ fn call(
...
+    transaction_hash: felt252,
+}
+
+fn declare(contract_name: felt252, max_fee: Option<felt252>) -> DeclareResult {
+    let mut inputs = array![contract_name];
+
+    match max_fee {
+        Option::Some(val) => {
+            inputs.append(0);
+            inputs.append(max_fee.unwrap());
```
abulenok, line 55: can we do just
```suggestion
        Option::Some(val) => {
            inputs.append(0);
            inputs.append(val);
```

crates/cast/src/lib.rs
```diff
@@ -268,7 +268,6 @@ pub async fn wait_for_tx(
     tx_hash: FieldElement,
     retries: u8,
 ) -> Result<&str> {
-    println!("Transaction hash: {tx_hash:#x}");
```
abulenok, line 271: What was the reason to remove this?

crates/cast/tests/e2e/script/declare.rs
```diff
@@ -0,0 +1,50 @@
+use crate::helpers::constants::{SCRIPTS_DIR, URL};
```
abulenok, line 1: Let's add a test that will declare the same contract twice and do `stderr_matches` the error msg.

abulenok: My idea is to test somehow that the first declare was successful and the contract is declared on the chain.

crates/cast/src/lib.rs
THenry14: This would be outputted each time we do any kind of transaction, so after running the script there was a possibility that we would have multiple "transaction hash" messages, that say nothing. Now the user will have a control over it in deployment scripts, and for regular cast commands we output this info nonetheless

crates/cast/tests/e2e/script/declare.rs
THenry14: Agreed. This scenario is covered by regular cast tests either way, so we should be good to go. For now, as this is not essential to have it at the moment, let me just make it in different PR, after release - is that ok?

PR #1153, opened 2023-11-20
POC of second version of rerun-filed

tkumor3:
Closes #

## Introduced changes

-

## Checklist

- [ ] Linked relevant issue
- [ ] Added relevant tests


crates/forge/src/main.rs
```diff
@@ -221,10 +234,25 @@ fn test_workspace(args: TestArgs) -> Result<bool> {
                 )
                 .await?;
 
+                let mut passed_tests_names = extract_passed_tests_names(&tests_file_summaries);
```
piotmag769, line 237: I think we should either change `all_failed_tests` to only store names or store whole `passed_tests` instead of `passed_tests_names`

PR #1156, opened 2023-11-20
Add a design doc for transactional testing

### LINKED ISSUE #1033: Create a design doc for `transactional tests`
It should include:
- Use cases
- Alternative approaches
- (pseudo)Code examples
- Possibility of reusing logic (visual representation is welcome)

Arcticae:
Closes #1033

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


design_documents/transactional_testing/transactional_testing.md
```diff
@@ -0,0 +1,75 @@
...
+## Proposed solution
+
+We could consider extending `cast script`, since the functionalities would overlap.
+This could be considered as a variant of `cast script`, in a way. 
+
+The extending/differentiating factors would be:
+
+- Extra environment-specific functions (for katana/devnet)
+- Additional RPC functions support (receipts, etc.)
+- Test-like behavior (fail/pass)
```
cptartur, line 30: For me, it's lacking some concrete examples: What exactly would have to be added as environment-specific functions? 

I'm assuming we'd need some kind of asserts, but what other things are needed?

design_documents/transactional_testing/transactional_testing.md
```diff
@@ -0,0 +1,75 @@
...
+the contracts in conditions as close to real networks conditions as possible.
+
+Also, the goal is to make the deployments testable and validation 
+of the deployments possible to the user. 
+
+The tests should be also runnable in a variety of environments, depending on the
+use case (namely katana, devnet, or testnet). This would increase user's confidence
+in the release process.
+
+## Proposed solution
```
cptartur, line 21: Would cheatcodes be possible with this solution? I'm assuming not.

MaksymilianDemitraszek: It could be possible if node implements them but those shouldn't be our focus atm

design_documents/transactional_testing/transactional_testing.md
```diff
@@ -0,0 +1,75 @@
...
+
+We could consider extending `cast script`, since the functionalities would overlap.
+This could be considered as a variant of `cast script`, in a way.
```
MaksymilianDemitraszek, line 24: Mix of forge test and cast script

MaksymilianDemitraszek: I wonder if this should be a part of forge test or we should have a specilized tool for it 🤔

MaksymilianDemitraszek: I think this desing doc should decide on something

MaksymilianDemitraszek (commented):
Questions:
- What would be file organization for such tests in terms of scarb workspace
- What is the scope for MVP
- What entity would be responsibe for compilation

MaksymilianDemitraszek:
Also would like to see a full example of such test testing a decently-complicated scenario

design_documents/transactional_testing/transactional_testing.md
```diff
@@ -0,0 +1,75 @@
...
+### Pros
+
+- Support of external tech (devnet, katana) via specialized libraries
+- Flow close to real starknet transactions, fees, testing how the protocol changes affect contracts
+- Ability to test accounts (__validate__, __execute__)
+- Ability to test real deployment/interaction scripts in environment close to real one
+
+### Cons
+- Keeping up the specialized libraries updated as tech adds new features/changes
+- Overall higher maintenance cost
```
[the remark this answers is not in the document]
Arcticae: That's why i added it in "Cons"

design_documents/transactional_testing/transactional_testing.md
Arcticae: Those are extra endpoints provided by the node

design_documents/transactional_testing/transactional_testing.md
Arcticae: Some functions like that are supported (in devnet for example)

design_documents/transactional_testing/transactional_testing.md
Arcticae: I think it already does, since the alternative approaches are integrating it into either one of the existing ones.
I am proposing a new tool for that.

design_documents/transactional_testing/transactional_testing.md
```diff
@@ -0,0 +1,75 @@
...
+from the point of user to use the deployment scripts in that way.
+
+
+### 2. Providing utilities to emulate transactional testing
+It's similar to alternative 1, but rather than integrating those concepts into deployment scripts,
+we could implement them into snforge itself.
+This approach would include trying to include this kind of flow into the current tests.
+We could provide utilities to test validation + execution in accounts, and/or add more cheatcodes which you'd use in the 
+said scenarios only (submit_txn, call, etc.). This would make for more confusing flow, if not used correctly
+(lack of intuitive separation of concerns for std funcs).
```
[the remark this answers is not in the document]
Arcticae: And test and so on... I don't really like the approach but i do think it's valid, that's why i put it in here.

design_documents/transactional_testing/transactional_testing.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Proposed solution
+
+We could consider extending `cast script`, since the functionalities would overlap.
+This could be considered as a variant of `cast script`, in a way. 
+
+The extending/differentiating factors would be:
+
+- Extra environment-specific functions (for katana/devnet)
+- Additional RPC functions support (receipts, etc.)
```
drknzz, line 29: Shouldn't this be a part of `sncast_std` (so pure cast deployment scripts)?

war-in (commented):
Would those tests be useful on non-local networks like testnet where the state is persistent? Let's say we would like to run a test which declares and deploys the contract and then run it again. It will fail because of double declare

design_documents/transactional_testing/transactional_testing.md
piotmag769: @THenry14 idea is to merge cast_std and forge_std, just provide different backends for them. Isn't it a good idea to do that first and then implement deployment scripts as another backend?

design_documents/transactional_testing/transactional_testing.md
```diff
@@ -0,0 +1,75 @@
...
+- Test-like behavior (fail/pass)
+
+
+## Solution architecture
+
+This is pretty straightforward. We ought to use the existing architecture for hint interception,
+provide extra functionality via libraries (one for forge, second one for devnet), and collect the cases from
```
piotmag769, line 36: >one for forge, second one for devnet

Wdym by that?

design_documents/transactional_testing/transactional_testing.md
```diff
@@ -0,0 +1,75 @@
...
+The extending/differentiating factors would be:
+
+- Extra environment-specific functions (for katana/devnet)
+- Additional RPC functions support (receipts, etc.)
+- Test-like behavior (fail/pass)
+
+
+## Solution architecture
+
+This is pretty straightforward. We ought to use the existing architecture for hint interception,
```
piotmag769, line 35: I wonder if we shouldn't unify the backend interface for deployment scripts and forge first, and then try to implement e2e tests. This amount of code can become unmaintainable very quickly and migrating will be really painful

piotmag769 (commented):
Do we plan on implementing the compilation of those tests on Scarb's side? If we want to do it in any time in the future I suggest doing it this way from the beggining, probably incorporating it into our scarb-snforge-test-collector

design_documents/transactional_testing/transactional_testing.md
Arcticae: [add docs links for specific katana/devnet functions](https://github.com/foundry-rs/starknet-foundry/pull/1156/commits/38e44121eecb5edb6364ee410e87db89d1708341)

design_documents/transactional_testing/transactional_testing.md
Arcticae: [add docs links for specific katana/devnet functions](https://github.com/foundry-rs/starknet-foundry/pull/1156/commits/38e44121eecb5edb6364ee410e87db89d1708341)
see this commit for docs links (supported extra functionalities)

design_documents/transactional_testing/transactional_testing.md
Arcticae: I am not aware about specific things that will be implemented there but might be needed to test things, quite possibly they will overlap but it's hard to tell in what degree.

It's feasible though to make some common utilities.

Arcticae:
> Would those tests be useful on non-local networks like testnet where the state is persistent? Let's say we would like to run a test which declares and deploys the contract and then run it again. It will fail because of double declare

That's a good take. We could actually make a helper function which would query if the class hash is declared and then either declare or just return (code the idempotency into it)

design_documents/transactional_testing/transactional_testing.md
Arcticae: I think there is an effort by @MaksymilianDemitraszek already which is going to unify this, so we'll be basing off that. I will include it in the doc.

design_documents/transactional_testing/transactional_testing.md
Arcticae: Ah brain fart, i meant katana and devnet

Arcticae:
I added a usage example, included your suggestions, and also added a third alternative to this kind of tool

design_documents/transactional_testing/transactional_testing.md
```diff
@@ -0,0 +1,172 @@
...
+    ./e2e/test_integration.cairo
+    ./e2e/test_integration2.cairo
+  ./pkg1
+    ./src
+    ./test
+    ./e2e
+      ./test_contract_1.cairo
+  ./pkg2
+  ...
+```
```
cptartur, line 71: I wonder if theses shouldn't go under `test/e2e`, though it would be harder to maintain.

design_documents/transactional_testing/transactional_testing.md
```diff
@@ -0,0 +1,172 @@
...
+    let account_balance = call(
+        address=mock_contract.address, 
+        function="get_balance",
+        calldata=Calldata::from_calls(array![Call(...), Call(...)])
+    );
+    
+    assert(account_balance > ..., "not correct balance");
+    dump(); // Dump can be used for pre-populating your test environment
+}
+```
```
cptartur, line 128: Looking at this example usage I wonder if it actually would make sense to just follow the alternative approach 1. and extend the deployment scripts.

I see that syntax is a bit different, but it seems to me like the backed for this will be very, very similar to cast: Not only the runtime part, but also just logic for making rpc calls to the whatever devnet we're using.

war-in (approved):

design_documents/transactional_testing/transactional_testing.md
piotmag769: I would prefer not to change the collection logic on forge's side once again :/ In theory tests under `tests/e2e` won't be collected if there is no `tests/e2e.cairo` file but we there is nothing preventing someone from having forge tests structured like this:

```
tests/
  e2e.cairo   <- # mod my_forge_test; #
  e2e/
     my_forge_test.cairo
```

design_documents/transactional_testing/transactional_testing.md
```diff
@@ -0,0 +1,172 @@
...
+- Keeping up the specialized libraries updated as tech adds new features/changes
+- Overall higher maintenance cost
+- Test execution speed significantly lower (depending on the used endpoints' throughput)
+- Dry-run mode from cast scripts might be sufficient for some use cases (testing deployments)
+- Might be confusing to users if not communicated correctly
+
+
+## Usage examples
+1. Test transaction signing
+```
```
abulenok, line 91:
```suggestion
```rust
```
for syntax highlighting

design_documents/transactional_testing/transactional_testing.md
```diff
@@ -0,0 +1,172 @@
...
+  ./e2e
+    ./e2e/test_integration.cairo
+    ./e2e/test_integration2.cairo
+  ./pkg1
+    ./src
+    ./test
+    ./e2e
+      ./test_contract_1.cairo
+  ./pkg2
+  ...
```
drknzz, line 70:
```suggestion
project/
├── Scarb.toml
├── e2e/
│   ├── test_integration.cairo
│   └── test_integration2.cairo
├── pkg1/
│   ├── src
│   ├── test
│   └── e2e/
│       └── test_contract_1.cairo
├── pkg2
└── ...
```

Arcticae:
Takeaway is that we want to do this, just not right now
After migrating cast to runtime extensions we might want to proceed it further and see, 
how it fits in into it.

Issue #1151, opened 2023-11-19
Test case filtering no longer works - Attribute `available_gas` is not supported error


parketh:
### Which component is your bug related to?

Forge

### Foundry Version

0.10.2

### What operating system are you using?

MacOS

### What system architecture are you using?

arm

### What happened

I have a package that uses both the native `cairo-test` and `snforge` for its testing suite, organised as follows:

```
package
  - tests
    - cairo_test
    - snforge
```

In v0.9.1, it was possible to selectively run `snforge` tests with `snforge test snforge`.

However, since upgrading to v0.10.2 this is now breaking: `snforge test <filter>` no longer excludes `cairo-test` cases and throws the following error:

```
Attribute `available_gas` is not supported: Contract functions execution cost would not be included in the gas calculation.
```

### Trace

```
snforge test snforge                                                                                                                                                                                                                                                
   Compiling lib(amm) amm v0.1.0 (/Users/parkyeung/Documents/Programming/sphinx/amm/Scarb.toml)
   Compiling starknet-contract(amm) amm v0.1.0 (/Users/parkyeung/Documents/Programming/sphinx/amm/Scarb.toml)
    Finished release target(s) in 16 seconds
[ERROR] amm::tests::cairo_test::math::test_math::test_pow - Attribute `available_gas` is not supported: Contract functions execution cost would not be included in the gas calculation.
```

### Is there an existing issue for this?

- [x] I have searched the existing issues and verified no issue exits for this problem.

piotmag769:
Hmmm, it was a desired behaviour to throw this error, but I think we did it in a wrong place - will change it so it only fails the test (therefore compilation will pass and filtering will be possible). Thanks for the report, we missed that one can have both sets of tests in one package!

parketh:
Hey thanks for the fast turnaround on this, however the issue still seems to be breaking after the fix.

I have created a minimal repo here to demonstrate this, running off of commit `df736e0`: https://github.com/parketh/failing-snforge-test

Running `snforge test snforge` throws the same error as above:
```
snforge test snforge                                                                                                                                                                                                                                                                                           ─╯
   Compiling snforgetest v0.1.0 (/Users/parkyeung/Documents/Programming/failing-snforge-test/Scarb.toml)
    Finished release target(s) in 1 second
[ERROR] snforgetest::test::cairo_test::test_add_cairo_test - Attribute `available_gas` is not supported: Contract functions execution cost would not be included in the gas calculation.
```

[...]

PR #1158, opened 2023-11-20
Panic if gas is set

piotmag769:


piotmag769:
Closing, we need to handle the gas error differently: #1151

PR #1159, opened 2023-11-20
Running benchmarks as PR checklist step

piotmag769:


cptartur (changes requested):
Cannot merge this until we restore benchmarks (they were temporarily reverted).

PR #1160, opened 2023-11-20
Fail a test instead of whole runner when using `available_gas`

piotmag769:
We prevented users having both snforge and cairo-test tests in their package from filtering and running only snforge tests (check issue description)

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


cptartur (approved):
Okay for a temporary solution

Issue #1135, opened 2023-11-16
update docs and changelog regarding deployment scripts


THenry14:
After everything is ready for release

PR #1165, opened 2023-11-21
Add deployment scripts docs

THenry14:
Closes https://github.com/foundry-rs/starknet-foundry/issues/1135

## Introduced changes

- add deployment scripts docs

## Checklist

- [X] Linked relevant issue
- [X] Added relevant tests


docs/src/starknet/script.md
```diff
@@ -0,0 +1,176 @@
+# Cairo Deployment Scripts
```
abulenok, line 1: Let's add something about using `path_to_scarb_toml` when running `sncast script` from arbitrary directory

THenry14: We mention this in several places in the docs, but never in sections dedicated to specific subcommands. Do you think we should change that? I guess this is rather top-level stuff

docs/src/starknet/script.md
```diff
@@ -0,0 +1,176 @@
...
+[dependencies]
+starknet = ">=2.3.0"
+sncast_std = { git = "https://github.com/foundry-rs/starknet-foundry.git", tag = "v0.11.0" }
+```
+
+To run the script, do:
+
+```shell
+$ sncast \
+  --rpc-url http://127.0.0.1:5050 \
```
abulenok, line 76:
```suggestion
  --url http://127.0.0.1:5050 \
```
https://foundry-rs.github.io/starknet-foundry/appendix/cast/common.html#--url--u-rpc_url

THenry14: great catch! chnged in newest commit

docs/src/starknet/script.md
```diff
@@ -0,0 +1,176 @@
...
+```
+
+Please note that `map` contract was specified as the dependency. In our example, it resides in the filesystem. It has to
+be compiled in `build-external-contracts`.
+
+To run the script, do:
+
+```shell
+$ sncast \
+  --rpc-url http://127.0.0.1:5050 \
```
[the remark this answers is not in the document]
THenry14: done :+1:

docs/src/starknet/script.md
```diff
@@ -0,0 +1,176 @@
...
+[[target.starknet-contract]]
+sierra = true
+casm = true
+build-external-contracts = [
+    "map::Map"
+]
+```
+
+Please note that `map` contract was specified as the dependency. In our example, it resides in the filesystem. It has to
+be compiled in `build-external-contracts`.
```
abulenok, line 159:
```suggestion
Please note that `map` contract was specified as the dependency. In our example, it resides in the filesystem. To generate the artifacts for it that will be accessible from the script you need to use the `build-external-contracts` property.
```

docs/src/starknet/script.md
abulenok: okay, it can stay as is then

abulenok (approved):

PR #1166, opened 2023-11-21
Extensible Starknet Runtime type boilerplate

MaksymilianDemitraszek:
Related #916 

## Introduced changes

- Introduces type boilerplate for the extension system, future PRs will migrate existing codebase to use this boilerplate.

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


war-in (commented):
Would be nice to see it in action (some migrated code). Currently, I can't see what it simplifies

piotmag769 (commented):
I really like this, but I think it would be better to do a stack of PRs implementing this trait for both forge and cast scripts so we are not left with a dead code that will be forgotten

MaksymilianDemitraszek:
@piotmag769 yea I have another PR wip wich migrates stuff partially

crates/forge-runner/src/runtime.rs
```diff
@@ -0,0 +1,135 @@
...
+}
+
+impl<'a> HintProcessorLogic for StarknetRuntime<'a> {
+    fn execute_hint(
+        &mut self,
+        vm: &mut VirtualMachine,
+        exec_scopes: &mut ExecutionScopes,
+        hint_data: &Box<dyn Any>,
+        constants: &HashMap<String, Felt252>,
+    ) -> Result<(), HintError> {
```
Arcticae, line 51:
```suggestion
    ) -> HintExecutionResult {
```
is that correct?
would make for clearer code

crates/forge-runner/src/runtime.rs
```diff
@@ -0,0 +1,135 @@
...
+}
+
+#[allow(dead_code)]
+enum HintHandlingResult {
+    Forward,
+    Result(HintExecutionResult),
+}
+
+#[allow(dead_code)]
+enum CheatcodeHadlingResult {
```
Arcticae, line 122:
```suggestion
enum CheatcodeHandlingResult {
```

crates/forge-runner/src/runtime.rs
```diff
@@ -0,0 +1,135 @@
...
+}
+
+trait ExtensionLogic {
+    fn override_system_call(_syscall: SyscallSelector) -> HintHandlingResult {
+        HintHandlingResult::Forward
+    }
+
+    fn handle_cheatcode(_selector: &str, _inputs: Vec<Felt252>) -> CheatcodeHadlingResult {
+        CheatcodeHadlingResult::Forward
+    }
```
Arcticae, line 134: Hard to say what will happen after this PR but you might be in a similar situation that i was, where separate Extensions would require cross-extension state access. Did you take that into account?

crates/forge-runner/src/runtime.rs
```diff
@@ -0,0 +1,135 @@
...
+        reference_ids: &HashMap<String, usize>,
+        references: &[HintReference],
+    ) -> Result<Box<dyn Any>, VirtualMachineError> {
+        self.hint_handler
+            .compile_hint(hint_code, ap_tracking_data, reference_ids, references)
+    }
+}
+
+pub struct ExtendedRuntime<Handler, Runtime: HintProcessor> {
+    pub extension_handler: Handler,
```
piotmag769, line 69: Actually what is the purpose of this one? It is not enforce to implement any trait for it, I think it needs a constraint

PR #1167, opened 2023-11-21
Run `cargo fmt --check` on whole workspace

cptartur:
Closes #

## Introduced changes

- Instead of specifying `cargo fmt --check -p ...` for every crate, run it for the whole workspace, the same way as we do for lints.

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -116,14 +116,8 @@ jobs:
         with:
           toolchain: stable
       - uses: Swatinem/rust-cache@3cf7f8cc28d1b4e7d01e3783be10a97d55d483c8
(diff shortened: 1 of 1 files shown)


war-in (approved):

THenry14:
out of curiousity - what is the cause of this change? were there any problems, or is this some kind of optimization?

cptartur:
> out of curiousity - what is the cause of this change? were there any problems, or is this some kind of optimization?

I've realized that we wasn't checking the formatting of the `forge-runner` crate. I've changed it to the whole workspace, so we don't have to remember to add every new crate separately.

PR #1169, opened 2023-11-21
Add declare for deployment scripts

THenry14:
Closes https://github.com/foundry-rs/starknet-foundry/issues/816

## Introduced changes

- add declare subcommand to deployment scripts

## Checklist

- [X] Linked relevant issue
- [X] Added relevant tests


crates/cast/tests/data/scripts/declare/no_contract/Scarb.toml
```diff
@@ -0,0 +1,12 @@
...
+version = "0.1.0"
+
+[dependencies]
+starknet = ">=2.3.0"
+sncast_std = { path = "../../../../../../../sncast_std" }
+
+[lib]
+sierra = true
+casm = true
+
```
abulenok, line 12: I guess those are not needed here, works for me locally when removed.
Possibly, the same for other Scarb.toml files in `tests/data/scripts`

abulenok: we can do the cleanup in a separate PR though.

THenry14: removed in newest commit :+1:
left it for map_script, as this one needs it

PR #1182, opened 2023-11-23
skip build/publish docs on prerelase

tkumor3:
Closes #

## Introduced changes

-

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -173,9 +173,10 @@ jobs:
 
   deploy-docs:
     name: Deploy documentation
-    needs: create-release
+    needs: [ create-release, verify-version ]
     permissions:
       contents: read
       pages: write
       id-token: write
+    if: ${{ !contains(needs.verify-version.outputs.version, '-')  }}
     uses: ./.github/workflows/docs.yml


.github/workflows/release.yml
```diff
@@ -173,9 +173,10 @@ jobs:
 
   deploy-docs:
     name: Deploy documentation
-    needs: create-release
+    needs: [ create-release, verify-version ]
     permissions:
       contents: read
       pages: write
       id-token: write
+    if: ${{ !contains(needs.verify-version.outputs.version, '-')  }}
```
THenry14, line 181: is the prerelease name somehow defined? I can see that there were prereleases with `-alpha*`, but also with just regular semver version (eg https://github.com/foundry-rs/starknet-foundry/releases/tag/v0.8.0)

if it is well defined, than I'd suggest we replaced `-` with `-alpha`.
if it is not, maybe we should get the release info from gh and check if it is a pre-release? (we could do it with curl probably, or I bet there are some gh actions for this as well).

wdyt?

tkumor3:
https://github.com/orgs/foundry-rs/projects/3/views/5?pane=issue&itemId=45623877 - improve release flow

.github/workflows/release.yml
THenry14: https://github.com/foundry-rs/starknet-foundry/issues/1189

THenry14 (approved):

PR #1205, opened 2023-11-27
Add option to get private key from file in `account add`

### LINKED ISSUE #696: Account add --private-key from file
I wonder if we should allow for getting this from file as well. Potentially it can be a security issue to have private keys in plain text in terminal, so some people may want to be able to provide it in some more secure way

_Originally posted by @THenry14 in https://github.com/foundry-rs/starknet-foundry/pull/660#discussion_r1328883312_

ddoktorski:
Closes #696 

## Introduced changes

- Allow user to provide account private key from a file in `account add` command

## Checklist

- [X] Linked relevant issue
- [X] Added relevant tests

### DIFF
--- a/crates/cast/src/starknet_commands/account/add.rs
+++ b/crates/cast/src/starknet_commands/account/add.rs
@@ -1,11 +1,11 @@
 use crate::starknet_commands::account::{
     add_created_profile_to_configuration, prepare_account_json, write_account_to_accounts_file,
 };
-use anyhow::{ensure, Result};
+use anyhow::{ensure, Context, Result};
 use camino::Utf8PathBuf;
-use cast::get_chain_id;
 use cast::helpers::response_structs::AccountAddResponse;
 use cast::helpers::scarb_utils::CastConfig;
+use cast::{get_chain_id, parse_number};
 use clap::Args;
 use starknet::core::types::BlockTag::Pending;
 use starknet::core::types::{BlockId, FieldElement};
@@ -23,7 +23,7 @@ pub struct Add {
     pub name: String,
 
     /// Address of the account
-    #[clap(short, long)]
+    #[clap(short, long, requires = "private_key_input")]
     pub address: FieldElement,
 
     /// Class hash of the account
@@ -36,8 +36,12 @@ pub struct Add {
     pub deployed: bool,
 
     /// Account private key
-    #[clap(long)]
-    pub private_key: FieldElement,
+    #[clap(long, group = "private_key_input")]
+    pub private_key: Option<FieldElement>,
+
+    /// Path to the file holding account private key
+    #[clap(long = "private-key-file", group = "private_key_input")]
+    pub private_key_file_path: Option<Utf8PathBuf>,
 
     /// Account public key
     #[clap(long)]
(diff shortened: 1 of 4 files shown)


crates/cast/src/starknet_commands/account/add.rs
```diff
@@ -23,7 +23,7 @@ pub struct Add {
     pub name: String,
 
     /// Address of the account
-    #[clap(short, long)]
+    #[clap(short, long, requires = "private_key_input")]
```
ddoktorski, line 26: It's a bit of a hack in order to display in the help message information after "usage" that it is required to provide either `--private-key` or `--private-key-file`. This applies as well to the error message when neither of these options is provided. I'm open to suggestions for alternative ways to handle this.

docs/src/appendix/cast/account/add.md
```diff
@@ -30,9 +30,16 @@ Specify account deployment status as deployed.
...
+Optional.
 
 Account private key.
+Required if `--private-key-file` is not passed.
+
+## `--private-key-file <PRIVATE_KEY_FILE_PATH>`
+Optional.
+
+Path to the file holding account private key.
+Required if `--private-key-file` is not passed.
```
abulenok, line 42: I think it's more appropriate to keep it near the information mentioning if the argument is optional or not

```suggestion
Optional. Required if `--private-key-file` is not passed.

Account private key.

## `--private-key-file <PRIVATE_KEY_FILE_PATH>`
Optional. Required if `--private-key-file` is not passed.

Path to the file holding account private key.
```

abulenok (approved):

crates/cast/src/starknet_commands/account/add.rs
```diff
@@ -60,7 +64,12 @@ pub async fn add(
     provider: &JsonRpcClient<HttpTransport>,
     add: &Add,
 ) -> Result<AccountAddResponse> {
-    let private_key = &SigningKey::from_secret_scalar(add.private_key);
+    let private_key = match &add.private_key_file_path {
+        Some(file_path) => get_private_key_from_file(file_path)
+            .with_context(|| format!("Failed to obtain private key from the file {file_path}"))?,
+        None => add.private_key.unwrap(),
```
THenry14, line 70: Let's generally obstain from unwrap-ing in prod code. The issue here is that we do not know if private_key is not None, and if it is this will panic - which ewill abruptly terminate the program.

Unwraping may be ok if you are sure about the Result type (or in tests).

crates/cast/tests/e2e/account/add.rs
```diff
@@ -214,7 +215,106 @@ pub async fn test_missing_arguments() {
...
+        "--deployed",
+    ];
+
+    let snapbox = runner(&args);
+
+    snapbox.assert().stderr_matches(indoc! {r"
+        command: account add
+        error: Failed to obtain private key from the file [..]
+    "});
+}
```
THenry14, line 320: We could also test when:
- private key is invalid
- is in int format

THenry14 (commented):
looks good 🔥
you can apply requested changes in separate PR as this one is already merged.

crates/cast/src/starknet_commands/account/add.rs
ddoktorski: Thanks! It this case it actually cannot be None as clap would throw an error earlier if both `private-key` and `private-key-file` are not provided, but I get the point.

PR #1206, opened 2023-11-27
Prototype running tests from casm

### LINKED ISSUE #1140: Migrate to CASM artifacts and new `SierraCasmRunner` (2.5.0)
Create a prototype of changes required to run test from casm code either by making necessary changes to `SierraCasmRunner` (which would eventually be upstreamed), or changing the code to use `casm_run` module directly.

cptartur:
Closes #1140 

## Introduced changes

-

## Checklist

- [ ] Linked relevant issue
- [ ] Added relevant tests


war-in:
Ideally, we should be able to create `CasmRunner` on our own. It is needed because VM has `pub(crate)` fields which `CasmRunner` can read.
This is the reason why `SierraCasmRunner` is copied now (we need its internal `CasmRunner` to extract used resources from the VM)

Arcticae:
Agree with @war-in, injecting the runner would be a nice way to do it 
Did we consider moving away from the `SierraCasmRunner` altogether or breaking it up into parts, so we can reuse the ones that we need?

crates/forge-runner/Cargo.toml
```diff
@@ -44,3 +44,4 @@ cheatnet = { path = "../cheatnet" }
 conversions = { path = "../conversions" }
 scarb-artifacts = { path = "../scarb-artifacts" }
 test-collector = { path = "../test-collector" }
+indoc = "2.0.4"
```
war-in, line 47: I think we have it in the top-level Cargo

crates/forge-runner/src/lib.rs
```diff
@@ -188,20 +190,71 @@ pub enum TestCrateRunResult {
...
-        SierraCasmRunner::new(
-            tests.sierra_program.clone(),
-            Some(MetadataComputationConfig::default()),
-            OrderedHashMap::default(),
-        )
-        .context("Failed setting up runner.")?,
-    );
+    let sierra_program = tests.sierra_program.clone();
+    let metadata_config = Some(MetadataComputationConfig::default());
+    let gas_usage_check = metadata_config.is_some();
```
war-in, line 250: I can't see the purpose of this value

crates/forge-runner/src/running.rs
```diff
@@ -265,15 +293,50 @@ pub fn run_test_case(
         None
     };
 
+    // copied from casm_run
```
war-in, line 296: Why do we need to copy this from casm_run?
I thought it would be possible just to put CairoRunner inside some function and then call finalize on it and vm

cptartur: I've tried to minimize changes to the casm_run interface, so we have to "copy" these lines:

```rust
    let mut vm = VirtualMachine::new(true);
    let data: Vec<MaybeRelocatable> = instructions
        .flat_map(|inst| inst.assemble().encode())
        .map(Felt252::from)
        .map(MaybeRelocatable::from)
        .collect();
    let data_len = data.len();
```

that is preparation of data to be passed to the runner

crates/forge-runner/src/running.rs
```diff
@@ -187,34 +208,42 @@ pub struct RunResultWithInfo {
...
+    pub return_types: Vec<(GenericTypeId, i16)>,
+}
+
 #[allow(clippy::too_many_arguments)]
 #[allow(clippy::too_many_lines)]
 pub fn run_test_case(
     args: Vec<Felt252>,
     case: &TestCaseRunnable,
-    runner: &SierraCasmRunner,
+    casm_program: &CairoProgram,
```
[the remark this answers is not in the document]
cptartur: Agreed, but we can do that in a separate PR. I want this one to only have changes necessary for migrating to casm.

crates/forge-runner/src/running.rs
```diff
@@ -78,7 +83,8 @@ fn build_hints_dict<'b>(
 
 pub fn run_test(
     case: Arc<TestCaseRunnable>,
-    runner: Arc<SierraCasmRunner>,
+    casm_program: Arc<CairoProgram>,
+    test_details: Arc<TestDetails>,
     runner_config: Arc<RunnerConfig>,
     runner_params: Arc<RunnerParams>,
     send: Sender<()>,
```
cptartur, line 90: The exact signature of this needs refactoring, but that can be done once we're aligned on the changes to SierraCasmRunner.

crates/forge-runner/src/cairo_runner.rs
cptartur, line 1: This replicates the structure of the runner in cairo repo so it will be easier to compare/upstream.

crates/forge-runner/src/running.rs
```diff
@@ -134,13 +155,13 @@ pub(crate) fn run_fuzz_test(
...
 
-fn build_context(block_info: CheatnetBlockInfo) -> EntryPointExecutionContext {
+pub fn build_context(block_info: CheatnetBlockInfo) -> EntryPointExecutionContext {
     let block_context = cheatnet_constants::build_block_context(block_info);
     let account_context = cheatnet_constants::build_transaction_context();
     EntryPointExecutionContext::new(&block_context, &account_context, ExecutionMode::Execute)
 }
 
-fn build_syscall_handler<'a>(
+pub fn build_syscall_handler<'a>(
```
war-in, line 164: are those pubs still relevant?

crates/forge-runner/src/cairo_runner/casm_run.rs
```diff
@@ -0,0 +1,98 @@
...
+    .map_err(CairoRunError::from)?;
+    CairoRunner::new(&program, "all_cairo", false)
+        .map_err(CairoRunError::from)
+        .map_err(Box::new)
+}
+
+#[allow(dead_code)]
+/// Runs `program` on layout with prime, and returns the memory layout and ap value.
+/// Allows injecting custom `HintProcessor`.
+pub fn run_function<'a, 'b: 'a, Instructions>(
```
war-in, line 65: Do we still need that?

crates/forge-runner/src/running.rs
```diff
@@ -277,15 +308,60 @@ pub fn run_test_case(
...
+                &mut vm,
+                &runner,
+                &mut forge_runtime
+                    .extended_runtime
+                    .extended_runtime
+                    .extended_runtime
+                    .hint_handler,
+                0,
+                2,
+            );
```
war-in, line 340: I'm not sure if we need `finalize` anymore. I think we can simply call `get_execution_resources` on the `runner` now

crates/forge-runner/src/running.rs
cptartur: Probably not all of them

crates/forge-runner/src/cairo_runner/casm_run.rs
cptartur: We do, it exist for the original casm_run in the cairo repo and I wanted to indicate how we've changed the internals of this function while still maintaining the original interface.

crates/forge-runner/src/lib.rs
```diff
@@ -202,20 +204,71 @@ pub enum TestCrateRunResult {
...
```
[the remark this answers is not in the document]
war-in: Ok, so we could get rid of that

crates/forge-runner/src/running.rs
```diff
@@ -277,15 +308,60 @@ pub fn run_test_case(
...
+
+            let cells = runner.relocated_memory;
+            let ap = vm.get_relocated_trace().unwrap().last().unwrap().ap;
```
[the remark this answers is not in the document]
cptartur: runner is injected into `run_function_with_runner`, so for it to return it'd have to also clone this value. Or we'd have to do some magic with ownerships.

crates/forge-runner/src/cairo_runner/sierra_casm_runner.rs
```diff
@@ -164,25 +200,43 @@ impl SierraCasmRunner {
...
+            hint_processor,
+            hints_dict,
+            instructions,
+            builtins,
+        )
+    }
+
+    #[allow(clippy::cast_sign_loss)]
+    /// Handling the main return value to create a `RunResultValue`.
+    pub fn handle_main_return_value(
```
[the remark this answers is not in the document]
cptartur: I'm unsure how to handle these. We could move them to be a separate functions in this module, but they don't make much sense outside of `SierraCasmRunner` context, that's why I left them in this impl.

crates/forge-runner/src/cairo_runner/casm_run.rs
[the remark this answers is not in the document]
cptartur: Basically  some parts of `run_function` were extracted to `run_function_with_runner` and `build_runner`, so we can use them directly. Minimal amount of entirely new code was added.

crates/forge-runner/src/cairo_runner/sierra_casm_runner.rs
```diff
@@ -81,80 +92,105 @@ impl SierraCasmRunner {
...
+        let (results_data, gas_counter) = Self::get_results_data(&return_types, &cells, ap);
         assert!(results_data.len() <= 1);
+
         let value = if results_data.is_empty() {
             // No result type - no panic.
             RunResultValue::Success(vec![])
         } else {
-            let [(ty, values)] = <[_; 1]>::try_from(results_data).ok().unwrap();
-            self.handle_main_return_value(&ty, values, &cells)
+            let (ty, values) = results_data[0].clone();
```
MaksymilianDemitraszek, line 167: Why you moved it outside from `handle_main_return_value`? Hard to say what this code does, maybe move it to a separate function and name it.

crates/forge-runner/src/cairo_runner/sierra_casm_runner.rs
```diff
@@ -50,9 +53,17 @@ pub struct SierraCasmRunner {
     /// The casm program matching the Sierra code.
     pub casm_program: CairoProgram,
     #[allow(dead_code)]
-    // Mapping from class_hash to contract info.
+    /// Mapping from class_hash to contract info.
     pub starknet_contracts_info: OrderedHashMap<Felt252, ContractInfo>,
 }
+
+#[allow(dead_code)]
```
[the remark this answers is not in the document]
MaksymilianDemitraszek: IMO it would be better to use Option instead.

crates/forge-runner/src/cairo_runner/sierra_casm_runner.rs
MaksymilianDemitraszek: Also it should be called from inside of `handle_main_return_value`, there is not point in leaking it.

MaksymilianDemitraszek (changes requested):
Generally a soild piece of work great job! Left some comments, overall we should leak less logic to upper layers.

crates/forge-runner/src/lib.rs
```diff
@@ -202,20 +204,70 @@ pub enum TestCrateRunResult {
...
-    let runner = Arc::new(
-        SierraCasmRunner::new(
-            tests.sierra_program.clone(),
-            Some(MetadataComputationConfig::default()),
-            OrderedHashMap::default(),
-        )
-        .context("Failed setting up runner.")?,
-    );
+    let sierra_program = tests.sierra_program.clone();
+    let metadata_config = Some(MetadataComputationConfig::default());
```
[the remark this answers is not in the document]
cptartur: I will extract it to a separate function for now, I don't want to touch compilation beyond what's necessary in this PR.

crates/forge-runner/src/lib.rs
```diff
@@ -236,17 +288,22 @@ pub async fn run_tests_from_crate(
...
         let case = Arc::new(case.clone());
         let args: Vec<ConcreteTypeId> = args.into_iter().cloned().collect();
-        let runner = runner.clone();
+        let test_details = Arc::new(build_test_details(&case.name, &sierra_program));
 
         tasks.push(choose_test_strategy_and_run(
             args,
             case.clone(),
-            runner,
+            casm_program.clone(),
```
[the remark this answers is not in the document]
cptartur: That's just cloning of an `Arc` so it's not a problem.

crates/forge-runner/src/cairo_runner/sierra_casm_runner.rs
```diff
@@ -81,80 +92,105 @@ impl SierraCasmRunner {
         })
     }
 
+    /// Runs the vm starting from a function in the context of a given starknet state.
```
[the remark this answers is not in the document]
cptartur: Cairo repo has the same comment, it was copied

crates/forge-runner/src/cairo_runner/sierra_casm_runner.rs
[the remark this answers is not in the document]
cptartur: If function can panic, the result value is wrapped in the panic wrapper, so we have to know the size of an inner type. I could just take this size as `Option<i16>`, but I though it would be less obvious to the caller what happens if it's passed or not.

crates/forge-runner/src/cairo_runner/casm_run.rs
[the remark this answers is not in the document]
MaksymilianDemitraszek: I don't think it is worth the effort, we will upstream it soon.

crates/forge-runner/src/running.rs
cptartur: Tried to use `finalize_execution` from blockifier directly instead, but it takes all arguments by value instead of reference and we still need access to `forge_runtime` after finalization.

crates/forge-runner/src/lib.rs
```diff
@@ -202,20 +204,71 @@ pub enum TestCrateRunResult {
...
+    let sierra_program = tests.sierra_program.clone();
+    let metadata_config = Some(MetadataComputationConfig::default());
+    let gas_usage_check = metadata_config.is_some();
+    let metadata = create_metadata(&sierra_program, metadata_config).unwrap();
+    let casm_program =
+        cairo_lang_sierra_to_casm::compiler::compile(&sierra_program, &metadata, gas_usage_check)
+            .unwrap();
+
+    // let runner = SierraCasmRunner::new(casm_program).context("Failed setting up runner.")?;
+    let casm_program = Arc::new(casm_program);
```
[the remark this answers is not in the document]
cptartur: tasks.push is inside a loop, so we'd have to clone the casm_program.

crates/forge-runner/src/running.rs
Arcticae: @cptartur can't we do it last then?

crates/forge-runner/src/lib.rs
```diff
@@ -236,17 +283,22 @@ pub async fn run_tests_from_crate(
             continue;
         };
 
-        let function = runner.find_function(&case_name)?;
+        let function = sierra_program
+            .funcs
+            .iter()
+            .find(|f| f.id.debug_name.clone().unwrap().ends_with(&case_name))
+            .unwrap();
```
Arcticae, line 290: So we still kind of rely on sierra here don't we?

Arcticae (commented):
I think this probably should be followed up with upstreaming it so the code does not change a lot between versions, also i think this should be based off 2.4.0 since it's just around the corner of being merged

crates/forge-runner/src/lib.rs
cptartur: We do, but this could be replaced. What we need is offsets to the function and it's argument and return types, and this information cannot be obtained from the casm code. My idea was that whatever module will be responsible for the compilation from sierra to casm, it will also output these information in some format.

crates/forge-runner/src/running.rs
cptartur: We'd have to clone `cells` we get from the casm runner even if we did it last. Because otherwise the compiler complains about it being partially moved.

I'll try thinking of the way of using `finalize_execution` but it's not trivial.

cptartur: Checked again, and the biggest problem is that `finalize_execution` needs to take `SyscallHintProcessor` by value, and we still need access to `forge_runtime` after finalization, so we can extract execution resources from it.

cptartur:
> I think this probably should be followed up with upstreaming it so the code does not change a lot between versions, also i think this should be based off 2.4.0 since it's just around the corner of being merged

Will update this PR for 2.4.0 once relevant PRs are merged. And the changes will be upstreamed once we agree on some version.

crates/forge-runner/src/lib.rs
Arcticae: Let's make a TODO for that perhaps, and discuss on the friday's meeting about compilation from sierra to casm

crates/forge-runner/src/running.rs
cptartur: Removed some code from finalize and inlined the rest of it (adding resources). It should be simpler now.

crates/forge-runner/src/lib.rs
cptartur: Added TODOs in code. Not creating issues for these for now

crates/forge-runner/src/cairo_runner/sierra_casm_runner.rs
cptartur: Moved to a separate method, maybe it's a bit cleaner now.

crates/forge-runner/src/cairo_runner/sierra_casm_runner.rs
```diff
@@ -378,6 +457,23 @@ impl SierraCasmRunner {
         Ok((ctx.instructions, builtins))
     }
 
+    /// Returns the instructions to add to the beginning of the code to successfully call the main
+    /// function, as well as the builtins required to execute the program.
+    pub fn create_entry_code(
```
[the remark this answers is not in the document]
cptartur: We want to retain original interface of SierraCasmRunner and it it was there

crates/forge-runner/src/cairo_runner/sierra_casm_runner.rs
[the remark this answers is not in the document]
Arcticae: TBH depends on the convention in the upstream, i would leave the review of the style to them on this one

Arcticae (approved):
LGTM

war-in (approved):

crates/forge-runner/src/cairo_runner/casm_run.rs
```diff
@@ -0,0 +1,105 @@
...
+use cairo_vm::serde::deserialize_program::{BuiltinName, HintParams, ReferenceManager};
+use cairo_vm::types::program::Program;
+use cairo_vm::types::relocatable::MaybeRelocatable;
+use cairo_vm::vm::errors::cairo_run_errors::CairoRunError;
+use cairo_vm::vm::runners::cairo_runner::CairoRunner;
+use cairo_vm::vm::vm_core::VirtualMachine;
+use std::collections::HashMap;
+
+#[allow(dead_code)]
+type RunFunctionRes = (Vec<Option<Felt252>>, usize);
```
MaksymilianDemitraszek, line 14:
```suggestion
type RunFunctionResult = (Vec<Option<Felt252>>, usize);
```

MaksymilianDemitraszek: easier to read, not worth the 3 letters

crates/forge-runner/src/cairo_runner/casm_run.rs
```diff
@@ -0,0 +1,105 @@
...
+        .run_until_pc(end, vm, hint_processor)
+        .map_err(CairoRunError::from)?;
+    runner
+        .end_run(true, false, vm, hint_processor)
+        .map_err(CairoRunError::from)?;
+    runner.relocate(vm, true).map_err(CairoRunError::from)?;
+    Ok(())
+}
+
+pub fn build_runner(
```
MaksymilianDemitraszek, line 39:
```suggestion
pub fn build_cairo_runner(
```
Also I would split it into build program and build runner. as this one builds program more than runner.

crates/forge-runner/src/cairo_runner/sierra_casm_runner.rs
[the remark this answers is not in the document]
MaksymilianDemitraszek: You mean this whole commented out code? Please remove it.

crates/forge-runner/src/cairo_runner/sierra_casm_runner.rs
```diff
@@ -164,25 +174,68 @@ impl SierraCasmRunner {
         })
     }
 
-    /// Handling the main return value to create a `RunResultValue`.
-    #[allow(clippy::cast_sign_loss)]
-    fn handle_main_return_value(
+    /// Extract inner type if `ty` is a panic wrapper
+    fn maybe_inner_type(&self, ty: &GenericTypeId, func: &Function) -> Option<ConcreteTypeId> {
```
MaksymilianDemitraszek, line 178:
```suggestion
    fn maybe_inner_type(&self, ty: &GenericTypeId, func: &Function) -> Option<ConcreteTypeId> {
```
Please don't use maybe prefix, it is already encoded in type info. Name doesn't suggest what this function does. What is `inner_type`?

crates/forge-runner/src/cairo_runner/sierra_casm_runner.rs
```diff
@@ -81,80 +86,85 @@ impl SierraCasmRunner {
...
         } else {
-            let [(ty, values)] = <[_; 1]>::try_from(results_data).ok().unwrap();
-            self.handle_main_return_value(&ty, values, &cells)
+            let (ty, values) = results_data[0].clone();
+            if let Some(inner_ty) = self.maybe_inner_type(&ty, func) {
+                let inner_ty_size = self.type_sizes[&inner_ty];
+                Self::handle_main_return_value(Some(inner_ty_size), values, &cells)
+            } else {
+                Self::handle_main_return_value(None, values, &cells)
+            }
```
MaksymilianDemitraszek, line 167:
```suggestion
            let inner_type_size = self.maybe_inner_type(&ty, func)
                .map(|x| x.type_sizes[&inner_ty]);
            Self::handle_main_return_value(inner_type_size, values, &cells)
```
This is a nitpick/kownledge sharing, no need to change.

crates/forge-runner/src/cairo_runner/sierra_casm_runner.rs
```diff
@@ -81,80 +86,85 @@ impl SierraCasmRunner {
...
         assert!(results_data.len() <= 1);
+
         let value = if results_data.is_empty() {
             // No result type - no panic.
             RunResultValue::Success(vec![])
         } else {
-            let [(ty, values)] = <[_; 1]>::try_from(results_data).ok().unwrap();
-            self.handle_main_return_value(&ty, values, &cells)
+            let (ty, values) = results_data[0].clone();
+            if let Some(inner_ty) = self.maybe_inner_type(&ty, func) {
```
MaksymilianDemitraszek, line 162: I still don't understand what is the motivation of running `maybe_inner_type` outside of `handle_main_return_value`. What's the point of leaking this implementation detail?

crates/forge-runner/src/running.rs
```diff
@@ -183,39 +202,48 @@ fn build_syscall_handler<'a>(
     )
 }
 
+#[derive(Debug)]
 pub struct RunResultWithInfo {
     pub(crate) run_result: Result<RunResult, RunnerError>,
     pub(crate) gas_used: u128,
 }
 
+// TODO merge this into test-collector's `TestCase`
```
MaksymilianDemitraszek, line 211: Please create an issue, this one is important.

MaksymilianDemitraszek (approved):
LGTM

crates/forge-runner/src/cairo_runner/casm_run.rs
cptartur: That's copied from Cairo repo

crates/forge-runner/src/cairo_runner/sierra_casm_runner.rs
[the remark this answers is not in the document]
cptartur: This code is not commented in the cairo repo, I did it here, because changing it is not relevant to us.

crates/forge-runner/src/cairo_runner/sierra_casm_runner.rs
cptartur: If return type is a panic wrapper, it wraps some actual return type. It was already named `inner_type` in cairo repo code.

cptartur: Renamed

crates/forge-runner/src/cairo_runner/sierra_casm_runner.rs
cptartur: Wa want to be able to use `handle_main_return_value` directly in our code. And for it to handle getting inner type it'd have to accept both `self` and `func` as arguments. By having it just accept extracted inner type size, it's reusable for us.

cptartur:
Cairo PR: https://github.com/starkware-libs/cairo/pull/4644

crates/forge-runner/src/running.rs
cptartur: #1433

drknzz:
Closed in favor of #1626

PR #1217, opened 2023-11-27
Add smoke tests on release

### LINKED ISSUE #198: Smoke tests on release
Integrate it into new automated workflow introduced in #859 

After building binaries we should try to create new forge project with snforge and see if tests run in it. Additionally, we should try to do some basic chain interactions with sncast.

If either of these fails, the release workflow should fail. 

We could consider still uploading artifcats to a draft even if tests fail, so they can be replicated though it's not necessary - verison built directly from master branch (or specific commit/tag if that's necessary/applicable) should match this one from the release workflow.

See scarb's release workflow for reference

cptartur:
Closes #198

## Introduced changes

- Added very basic tests that are triggered before before github release is created. Tests will use binaries built by release workflow.

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


.github/workflows/release.yml
```diff
@@ -137,10 +137,51 @@ jobs:
...
+          BINARY_PATH="${BINARY_PATH%.tar.gz}"
+          BINARY_PATH="${BINARY_PATH#artifacts/}"
+          
+          SNFORGE_PATH=$(realpath $BINARY_PATH/bin/snforge)
+          SNCAST_PATH=$(realpath $BINARY_PATH/bin/sncast)
+          
+          REPO_URL=${{ github.repositoryUrl }}
+          REVISION=${{ github.sha }}
+          
+          ./scripts/smoke_test.sh $SNFORGE_PATH $SNCAST_PATH $REPO_URL $REVISION
```
THenry14, line 179: Let's double quote variables in this snippet to prevent globbing/splitting

.github/workflows/release.yml
```diff
@@ -137,10 +137,51 @@ jobs:
...
+
+      - name: Unpack artifact
+        run: |
+          tar xzvf ${{ env.BINARY_PATH }}
+
+      - name: Smoke test
+        run: |
+          BINARY_PATH="${{ env.BINARY_PATH }}"
+          BINARY_PATH="${BINARY_PATH%.tar.gz}"
+          BINARY_PATH="${BINARY_PATH#artifacts/}"
```
THenry14, line 171: Can this be combined?

scripts/smoke_test.sh
```diff
@@ -0,0 +1,25 @@
...
+$SNFORGE_PATH init my_project
+pushd my_project || exit
+scarb remove snforge_std
+scarb add snforge_std --git "$REPO_URL" --rev "$REVISION"
+$SNFORGE_PATH test || exit
+popd || exit
+
+# Check cast
+
+OUTPUT=$($SNCAST_PATH --url http://188.34.188.184:9545/rpc/v0.4 call --contract-address 0x071c8d74edc89330f314f3b1109059d68ebfa68874aa91e9c425a6378ffde00e --function "get_balance")
```
THenry14, line 19: Let's obfuscate the url by placing it in secrets. The slug and port can be defined here.

scripts/smoke_test.sh
```diff
@@ -0,0 +1,25 @@
...
+
+EXPECTED=$'command: call\nresponse: [0x2]'
+
+if [[ "$OUTPUT" != "$EXPECTED" ]]; then
+    exit 1
+fi
```
THenry14, line 25: this is fine, but you could just directly grep instead
```suggestion
if ! sncast --url http://188.34.188.184:9545/rpc/v0.4 call \
    --contract-address 0x071c8d74edc89330f314f3b1109059d68ebfa68874aa91e9c425a6378ffde00e \
    --function "get_balance" | grep -q "response: \[0x2\]"; then 
  exit 1
fi

```

.github/workflows/release.yml
cptartur: I don't think so, or at least I cannot get it to work.

cptartur:
Testing if it works on my fork: https://github.com/cptartur/starknet-foundry/actions/runs/7016971212

Arcticae (approved):

.github/workflows/release.yml
THenry14: we could do this with awk to improve readability, or try to improve this some other way like temp variables, but I guess this is not essential atm. I leave the decision up to you.

THenry14 (approved):
a nitpick, nothing blocking

.github/workflows/release.yml
```diff
@@ -137,10 +137,85 @@ jobs:
...
+
+      - name: Get binary path
+        shell: bash
+        run: |
+          if [[ ${{ matrix.target }} == *windows* ]]; then
+            BINARY_PATH="artifacts/starknet-foundry-${{ needs.verify-version.outputs.version }}-${{ matrix.target }}.zip"
+          else
+            BINARY_PATH="artifacts/starknet-foundry-${{ needs.verify-version.outputs.version }}-${{ matrix.target }}.tar.gz"
+          fi
+          echo "BINARY_PATH=$BINARY_PATH" >> $GITHUB_ENV
```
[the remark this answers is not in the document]
cptartur: I'll merge this version and follow up in another PR

PR #1218, opened 2023-11-27
Add type boilerplate

MaksymilianDemitraszek:


crates/forge-runner/src/runtime.rs
```diff
@@ -0,0 +1,135 @@
...
+
+#[allow(dead_code)]
+enum CheatcodeHadlingResult {
+    Forward,
+    Result(Vec<Felt252>),
+}
+
+trait ExtensionLogic {
+    fn override_system_call(_syscall: SyscallSelector) -> HintHandlingResult {
+        HintHandlingResult::Forward
```
piotmag769, line 129: What is the distinction between `Hint` and `Cheatcode`, do we mean `Hint` as Call etc.?

crates/forge-runner/src/runtime.rs
```diff
@@ -0,0 +1,135 @@
...
+        ap_tracking_data: &ApTracking,
+        reference_ids: &HashMap<String, usize>,
+        references: &[HintReference],
+    ) -> Result<Box<dyn Any>, VirtualMachineError> {
+        self.hint_handler
+            .compile_hint(hint_code, ap_tracking_data, reference_ids, references)
+    }
+}
+
+pub struct ExtendedRuntime<Handler, Runtime: HintProcessor> {
```
piotmag769, line 68: No trait enforced for `Handler`, it could be more strict I think. From type definition you won't know what is it used for

crates/forge-runner/src/runtime.rs
```diff
@@ -0,0 +1,135 @@
...
+    pub extension_handler: Handler,
+    pub extended_runtime: Runtime,
+}
+
+impl<Handler, Runtime: HintProcessor> ResourceTracker for ExtendedRuntime<Handler, Runtime> {
```
piotmag769, line 73: It seems that rust doesn't allow `impl ResourceTracker for ExtendedRuntime` even though it could infer the types when needed 🤔

crates/forge-runner/src/runtime.rs
```diff
@@ -0,0 +1,135 @@
...
+}
+
+#[allow(dead_code)]
+enum CheatcodeHadlingResult {
+    Forward,
+    Result(Vec<Felt252>),
+}
+
+trait ExtensionLogic {
+    fn override_system_call(_syscall: SyscallSelector) -> HintHandlingResult {
```
piotmag769, line 128: I know the `_` is for linter, but it forces the arg to be named precisely `_syscall` when implementing the trait, no?

MaksymilianDemitraszek: This boilerplate evolves a bit in the next PR, i suggest looking there

crates/forge-runner/src/runtime.rs
MaksymilianDemitraszek: I updated it, to match the following PR signatures

crates/forge-runner/src/runtime.rs
MaksymilianDemitraszek: It has changed to state

crates/forge-runner/src/runtime.rs
```diff
@@ -0,0 +1,204 @@
...
+            input_end,
+            output_start,
+            output_end,
+        })) = maybe_extended_hint
+        {
+            // Parse the selector.
+            let selector = &selector.value.to_bytes_be().1;
+            let selector = std::str::from_utf8(selector).map_err(|_| {
+                CustomHint(Box::from(
+                    "Failed to parse the  cheatcode selector".to_string(),
```
piotmag769, line 102: double space

crates/forge-runner/src/runtime.rs
```diff
@@ -0,0 +1,204 @@
...
+#[allow(dead_code)]
+#[derive(Debug)]
+pub enum SyscallHandlingResult {
+    Forward,
+    Result(()),
+}
+
+#[allow(dead_code)]
+#[derive(Debug)]
+pub enum CheatcodeHadlingResult {
```
piotmag769, line 176: typo

crates/forge-runner/src/runtime.rs
```diff
@@ -0,0 +1,204 @@
...
+    Forward,
+    Result(()), // TODO now use buffer later rewrite to return vector
+}
+
+pub trait ExtensionLogic {
+    type Runtime: HintProcessorLogic;
+
+    fn get_extended_runtime_mut(&mut self) -> &mut Self::Runtime;
+
+    fn get_extended_runtime(&self) -> &Self::Runtime;
```
piotmag769, line 186: I don't get the point of these

crates/forge-runner/src/runtime.rs
MaksymilianDemitraszek: fixed

crates/forge-runner/src/runtime.rs
MaksymilianDemitraszek: Those are required to ensure type of Runtime has `HintProcessorLogic` implemented

piotmag769: It looks more like a hack, I don't like it tbh

piotmag769 (approved):

crates/forge-runner/src/runtime.rs
MaksymilianDemitraszek: I will take a look if this can be simplified, but keep in mind

> It looks more like a hack, I don't like it tbh
is not very high quality feedback on somebodys work. Quality feedback includes the understanding of a tradeoffs and offers a suggestion how to improve things.

Arcticae (approved):

PR #1241, opened 2023-11-29
IO Runtime as POC of extension composition

MaksymilianDemitraszek:
commit-id:dccb78de

---

**Stack**:
- #1248
- #1241 ⬅

### DIFF
--- a/crates/forge-runner/src/running.rs
+++ b/crates/forge-runner/src/running.rs
@@ -15,6 +15,7 @@ use cairo_vm::serde::deserialize_program::HintParams;
 use cairo_vm::types::relocatable::Relocatable;
 use cheatnet::execution::cheatable_syscall_handler::CheatableSyscallHandler;
 use itertools::chain;
+use runtime::io_runtime_extension::IORuntimeState;
 
 use crate::gas::gas_from_execution_resources;
 use crate::sierra_casm_runner::SierraCasmRunner;
@@ -27,11 +28,10 @@ use cairo_lang_runner::{Arg, RunResult, RunnerError};
 use cairo_vm::vm::vm_core::VirtualMachine;
 use camino::Utf8Path;
 use cheatnet::constants as cheatnet_constants;
-use cheatnet::execution::contract_execution_syscall_handler::ContractExecutionSyscallHandler;
 use cheatnet::forking::state::ForkStateReader;
 use cheatnet::state::{BlockInfoReader, CheatnetBlockInfo, CheatnetState, ExtendedStateReader};
 use runtime::forge_runtime_extension::{ForgeRuntime, TestExecutionState};
-use runtime::{ExtendedRuntime, RuntimeExtension};
+use runtime::{ExtendedRuntime, ExtensionLogic, RuntimeExtension};
 use starknet::core::types::BlockTag::Latest;
 use starknet::core::types::{BlockId, MaybePendingBlockWithTxHashes};
 use starknet::core::utils::get_selector_from_name;
@@ -242,8 +242,10 @@ pub fn run_test_case(
     let cheatable_syscall_handler =
         CheatableSyscallHandler::wrap(syscall_handler, &mut cheatnet_state);
 
-    let contract_execution_syscall_handler =
-        ContractExecutionSyscallHandler::wrap(cheatable_syscall_handler);
+    let io_runtime = ExtendedRuntime(RuntimeExtension {
+        extended_runtime: cheatable_syscall_handler,
+        extension_state: IORuntimeState {},
+    });
 
     let test_execution_state = TestExecutionState {
         environment_variables: &runner_params.environment_variables,
@@ -251,7 +253,7 @@ pub fn run_test_case(
     };
 
     let mut forge_runtime = ExtendedRuntime(RuntimeExtension {
-        extended_runtime: contract_execution_syscall_handler,
+        extended_runtime: io_runtime,
         extension_state: test_execution_state,
     });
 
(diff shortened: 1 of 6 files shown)


crates/runtime/src/lib.rs
```diff
@@ -207,6 +205,15 @@ impl<Handler, Runtime: SyscallPtrAccess + HintProcessor> SyscallPtrAccess
         self.extended_runtime.verify_syscall_ptr(ptr)
     }
 }
+impl<Runtime: SyscallPtrAccess> SyscallPtrAccess for ExtendedRuntime<Runtime> {
```
Arcticae, line 208: Isn't that doable with a macro btw?

crates/runtime/src/lib.rs
```diff
@@ -38,13 +38,13 @@ pub trait SyscallPtrAccess {
 }
 
 // TODO this is only temporary, after we migrate everything to extension it will be auto-derived
-impl<'a> SyscallPtrAccess for ContractExecutionSyscallHandler<'a> {
+impl<'a> SyscallPtrAccess for CheatableSyscallHandler<'a> {
```
Arcticae, line 41: This can also be done with a macro
I don't think you have to migrate everything to do it

crates/forge-runner/src/running.rs
```diff
@@ -361,11 +363,18 @@ fn get_latest_block_number(url: &Url) -> Result<BlockId> {
...
         .0
         .extended_runtime
+        .0
+        .get_extended_runtime()
         .child
+        .resources;
+    let cheatnet_used_resources = &runtime
+        .0
+        .extended_runtime
+        .0
```
Arcticae, line 376: With the nesting this can quickly get tedious - can we implement `Deref` for the runtime tuples?

crates/forge-runner/src/running.rs
```diff
@@ -361,11 +363,18 @@ fn get_latest_block_number(url: &Url) -> Result<BlockId> {
...
+    let test_used_resources = &runtime
         .0
         .extended_runtime
+        .0
+        .get_extended_runtime()
         .child
+        .resources;
+    let cheatnet_used_resources = &runtime
+        .0
+        .extended_runtime
```
Arcticae, line 375: Why are we mixing accessing it directly with method-based access?

crates/runtime/src/forge_runtime_extension.rs
```diff
@@ -38,32 +39,28 @@ use cairo_lang_starknet::contract::starknet_keccak;
...
 // This runtime extension provides an implementation logic for functions from snforge_std library.
-impl<'a> ExtensionLogic
-    for RuntimeExtension<TestExecutionState<'a>, ContractExecutionSyscallHandler<'a>>
-{
-    type Runtime = ContractExecutionSyscallHandler<'a>;
+impl<'a> ExtensionLogic for RuntimeExtension<TestExecutionState<'a>, IORuntime<'a>> {
+    type Runtime = IORuntime<'a>;
 
-    fn get_extended_runtime_mut(&mut self) -> &mut ContractExecutionSyscallHandler<'a> {
+    fn get_extended_runtime_mut(&mut self) -> &mut IORuntime<'a> {
```
Arcticae, line 59: I think this is doable with a macro as well

crates/runtime/src/io_runtime_extension.rs
```diff
@@ -0,0 +1,72 @@
...
+                print(inputs);
+                Ok(CheatcodeHandlingResult::Handled(vec![]))
+            }
+            _ => Err(EnhancedHintError::from(HintError::CustomHint(
+                "Only `print` cheatcode is available in contracts.".into(),
+            ))),
+        }
+    }
+
+    fn override_system_call(
```
Arcticae, line 46: Isn't that doable with a default impl for a trait?

crates/runtime/src/lib.rs
MaksymilianDemitraszek: I need only the last one that is not migrated yet to have this implemented. After I migrate it fully it will be auto derived

crates/runtime/src/lib.rs
MaksymilianDemitraszek: It is generic, so it will work for all structs

crates/runtime/src/io_runtime_extension.rs
piotmag769: Not sure if you can provide default impl for just one function that the trait requires :c

piotmag769 (approved):
LGTM overall

crates/forge-runner/src/running.rs
MaksymilianDemitraszek: This way of accessing it will be dropped in a following PR

crates/runtime/src/forge_runtime_extension.rs
MaksymilianDemitraszek: Not sure about that, maybe but it will be changing in the next PR so we can look into this later, it is not very heavy on boilerplate

crates/runtime/src/io_runtime_extension.rs
MaksymilianDemitraszek: DONE

crates/runtime/src/forge_runtime_extension.rs
MaksymilianDemitraszek: It is removed at all in the next PR, resolving

crates/forge-runner/src/running.rs
MaksymilianDemitraszek: Resoliving, removed in the next PR

crates/forge-runner/src/running.rs
MaksymilianDemitraszek: Removed in the next PR, resolving

crates/runtime/src/lib.rs
MaksymilianDemitraszek: There will be no more implementations of it, next PR simplifies it grately

PR #1242, opened 2023-11-29
Add Secp256 Curves

drknzz:
Closes #

## Introduced changes

-

## Checklist

- [ ] Linked relevant issue
- [ ] Added relevant tests


crates/forge/tests/integration/signing.rs
drknzz, line 1: I'll gladly take in any ideas for tests

crates/forge/tests/integration/signing.rs
```diff
@@ -3,10 +3,10 @@ use test_utils::running_tests::run_test_case;
 use test_utils::{assert_passed, test_case};
 
 #[test]
-fn simple_signing_flow() {
+fn simple_signing_flow_stark_curve() {
     let test = test_case!(indoc!(
         r"
-            use snforge_std::signature::{ StarkCurveKeyPair, StarkCurveKeyPairTrait, Signer, Verifier };
+            use snforge_std::signature::stark_curve::{ StarkCurveKeyPair, StarkCurveKeyPairTrait, Signer, Verifier };
```
drknzz, line 9: This is a breaking change

crates/runtime/src/forge_runtime_extension.rs
```diff
@@ -448,6 +439,140 @@ impl<'a> ExtensionLogic
...
+                };
+
+                Ok(CheatcodeHandlingResult::Handled(vec![
+                    Felt252::from_bytes_be(&signing_key_bytes[16..32]),
+                    Felt252::from_bytes_be(&signing_key_bytes[0..16]),
+                    Felt252::from_bytes_be(&verifying_key_bytes[17..33]),
+                    Felt252::from_bytes_be(&verifying_key_bytes[1..17]),
+                    Felt252::from_bytes_be(&verifying_key_bytes[49..65]),
+                    Felt252::from_bytes_be(&verifying_key_bytes[33..49]),
+                ]))
```
drknzz, line 487: These come from the different order of halves in cairo u256

drknzz: The order of bytes could be changed here to "normal", but then it would be reversed in cairo.

crates/runtime/src/forge_runtime_extension.rs
```diff
@@ -448,6 +439,140 @@ impl<'a> ExtensionLogic
...
+
+                            let signature: p256::ecdsa::Signature =
+                                p256::ecdsa::signature::hazmat::PrehashSigner::sign_prehash(
+                                    &signing_key,
+                                    &msg_hash,
+                                )
+                                .unwrap();
+
+                            signature.split_bytes()
+                        }
```
drknzz, line 526: Should some of these operations be `.expect()`'ed or a Result passed to cairo and panic used there?

crates/forge/tests/integration/signing.rs
```diff
@@ -3,10 +3,10 @@ use test_utils::running_tests::run_test_case;
 use test_utils::{assert_passed, test_case};
 
 #[test]
-fn simple_signing_flow() {
+fn simple_signing_flow_stark_curve() {
```
MaksymilianDemitraszek, line 6: Can we make the name of this test and behaviour it tests consistent with other curve tests?

MaksymilianDemitraszek (changes requested):
Logic-wise LGTM overall, major issue I see here is a lack of unification in terms of cario interfaces.

I understand that you probably wanted to avoid breaking changes but you moved the module, causing breaking changes either way.

Please unify the interfaces. If you want to prevent breaking changes, you can leave stark cure module as it was as a duplicate and mark it deprecated (but I would just break it, we are not stable yet). 

Do we have an issue for docs so we don't forget about them?

crates/runtime/src/forge_runtime_extension.rs
drknzz: So currently there is no way to make this `sign_prehash` return `Result::Err`, as it will sign any `u256` correctly using the `signing_key` above. What could be bubbled is the error from `SigningKey::from_slice` above.

snforge_std/src/signature/elliptic_curve/key_pair.cairo
```diff
@@ -0,0 +1,122 @@
...
+use starknet::secp256_trait::{Secp256Trait, Secp256PointTrait, is_valid_signature};
+use starknet::secp256k1::{Secp256k1Point, Secp256k1Impl};
+use starknet::secp256r1::{Secp256r1Point, Secp256r1Impl};
+use starknet::{SyscallResult, SyscallResultTrait};
+
+#[derive(Copy, Drop)]
+struct KeyPair<T> {
+    private_key: u256,
+    public_key: T,
+}
```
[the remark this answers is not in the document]
drknzz: I feel like this should be done in another PR, together with the unification of stark curve and the secp curves.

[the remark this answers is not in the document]
MaksymilianDemitraszek: Works for me, can you create an issue?

snforge_std/src/signature/elliptic_curve/interface.cairo
```diff
@@ -0,0 +1,7 @@
+trait Signer<T> {
+    fn sign(self: T, message_hash: u256) -> (u256, u256);
+}
+
+trait Verifier<T> {
+    fn verify(self: T, message_hash: u256, signature: (u256, u256)) -> bool;
+}
```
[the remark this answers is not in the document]
drknzz: #1257

snforge_std/src/signature/elliptic_curve/key_pair.cairo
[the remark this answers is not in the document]
drknzz: #1257

crates/forge/tests/integration/signing.rs
drknzz: #1257

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
```diff
@@ -443,6 +434,163 @@ impl<'a> ExtensionLogic for ForgeExtension<'a> {
...
+
+                let verifying_key_bytes = {
+                    match curve.as_deref() {
+                        Some("Secp256k1") => {
+                            let Ok(signing_key) = k256::ecdsa::SigningKey::from_slice(&secret_key)
+                            else {
+                                return Ok(CheatcodeHandlingResult::Handled(vec![
+                                    Felt252::from(1),
+                                    Felt252::from_("invalid secret_key".to_string()),
+                                ]));
```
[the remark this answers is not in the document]
drknzz: I agree, but I think I'd rather keep passing these mapped Errors explicit

snforge_std/src/signature/elliptic_curve/key_pair.cairo
```diff
@@ -0,0 +1,130 @@
...
+use starknet::secp256_trait::{Secp256Trait, Secp256PointTrait, is_valid_signature};
+use starknet::secp256k1::{Secp256k1Point, Secp256k1Impl};
+use starknet::secp256r1::{Secp256r1Point, Secp256r1Impl};
+use starknet::{SyscallResult, SyscallResultTrait};
+
+#[derive(Copy, Drop)]
+struct KeyPair<T> {
+    secret_key: u256,
+    public_key: T,
+}
```
MaksymilianDemitraszek, line 20: If you merge it before tomorrows release you will make breaking changes twice.

MaksymilianDemitraszek (approved):

PR #1248, opened 2023-11-29
Make extension a struct

MaksymilianDemitraszek:
**Stack**:
- #1248 ⬅
- #1241


MaksymilianDemitraszek:
This PR makes extension a struct to:
- allow implementing extension outside of runtime module
- simplifies type indirection
- removes some getters

crates/runtime/src/forge_runtime_extension.rs
MaksymilianDemitraszek, line 1: This is generally worth reviewing, to see how extension interface has changed

war-in:
This one is so nice! I lost hope on the previous one but now I'm saved.
I wonder why we have two separate PRs because it would be much easier to review it as one 🤔

drknzz (approved):
It's getting nicer with each pr 👍

crates/runtime/src/lib.rs
```diff
@@ -104,17 +104,12 @@ impl<'a> HintProcessorLogic for StarknetRuntime<'a> {
     }
 }
 
-pub struct RuntimeExtension<ExtensionState, Runtime> {
-    pub extension_state: ExtensionState,
-    pub extended_runtime: Runtime,
+pub struct ExtendedRuntime<Extension: ExtensionLogic> {
+    pub extension: Extension,
+    pub extended_runtime: <Extension as ExtensionLogic>::Runtime,
```
piotmag769, line 109: :o based

PR #1249, opened 2023-11-30
Use Scarb version from .tool-versions in CI

cptartur:
Closes #

## Introduced changes

- https://github.com/software-mansion/setup-scarb#inputs action will by default use Scarb version from `.tool-versions` file if it is not specified in the workflow file. This PR removes the scarb version from CI workflow, so version defined in `.tool-versions` will be used.

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


THenry14 (changes requested):
let's remove from cast as well (sadly I cannot suggest change)

THenry14 (approved):

PR #1274, opened 2023-12-04
Update starknet-rs and devnet

THenry14:
This PR bumps starknet-rs and devnet to support rpc 0.5.1. This is needed for (among other issues) https://github.com/foundry-rs/starknet-foundry/issues/1327

## Introduced changes

- Bump starknet-rs and starknet-devnet deps

## Checklist

- [X] Linked relevant issue
- [X] Added relevant tests


crates/sncast/tests/e2e/call.rs
```diff
@@ -85,7 +85,7 @@ fn test_wrong_function_name() {
 
     snapbox.assert().stderr_matches(indoc! {r"
         command: call
-        error: Entry point [..] not found in contract.
+        error: Contract error
```
war-in, line 88: I assume this reflects changes from the rpc update, right? Those old errors were nice. `Contract error` is not meaningful at all.

war-in (approved):
Approved, however previous errors were better 😢

crates/sncast/tests/e2e/call.rs
THenry14: that's true, I liked them better as well. I have created an issue (for now on our part) to talk to devnet guys and see if something's could be done with this.

DelevoXDG: `Contract error` should also have `data` with `revert_error` that has more specific details, maybe it's just not currently handled?

Issue #1230, opened 2023-11-28
Create a mechanism for tx status retrieval and wait


THenry14:
instead of --wait, let's have a mechanism to allow users to handle tx exec times.

Issue #1201, opened 2023-11-27
Add more tests for scripts


THenry14:
Let's try to add more tests to scripts, to make sure more scenarios are being covered (eg, more scarb workspaces layouts, more in-script scenarios etc)

PR #1283, opened 2023-12-04
Allow to modify wait timeout and retry interval + set nonce

THenry14:
https://github.com/foundry-rs/starknet-foundry/issues/1230

Docs will be added in different PR, aiming to improve scripts docs

## Introduced changes

- add `--wait-timeout` and `--wait-retry-interval` flags to allow adjusting those values 
- allow setting nonce for declare/deploy/invoke using `--nonce` flag, and in deployment scripts 

## Checklist

- [X] Linked relevant issue
- [X] Added relevant tests


crates/cast/src/helpers/constants.rs
```diff
@@ -16,9 +16,11 @@ pub const UDC_ADDRESS: &str = "0x041a78e741e5af2fec34b695679bc6891742439f7afb848
 pub const OZ_CLASS_HASH: &str =
     "0x058d97f7d76e78f44905cc30cb65b91ea49a4b908a76703c54197bca90f81773";
 
-// used in wait_for_tx. Txs are fetched every 5 seconds
+// used in wait_for_tx. Txs will be fetched every 5s with timeout of 60s - so 12 attempts
 #[allow(dead_code)]
-pub const DEFAULT_RETRIES: u8 = 12;
+pub const WAIT_TIMEOUT: u16 = 60;
```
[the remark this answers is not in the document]
DelevoXDG: Yes, should probably increase to 5 mins, often it takes longer to generate the block.

DelevoXDG (commented):
Looking good overall. 
I don't think there are currently any tests that test cli directly, like most of the e2e devnet tests. It would be hard to make reasonable tests on devnet since there is not support for delay between block generation, but maybe we could at least add these to check if the parameters passed to cli are parsed correctly, wdyt?

THenry14:
> Looking good overall. I don't think there are currently any tests that test cli directly, like most of the e2e devnet tests. It would be hard to make reasonable tests on devnet since there is not support for delay between block generation, but maybe we could at least add these to check if the parameters passed to cli are parsed correctly, wdyt?

Yeah, this is not testable sadly :( args parsing is already tested :+1:
I will probably add some integration tests for scripts in later stages, so we are able to test such things.

crates/cast/src/lib.rs
```diff
@@ -282,16 +295,17 @@ pub async fn wait_for_tx(
                 code: MaybeUnknownErrorCode::Known(StarknetError::TransactionHashNotFound),
                 message: _,
             })) => {
-                println!("Waiting for transaction to be received ({i} retries left)");
+                let remaining_time = i * u16::from(retry_interval);
```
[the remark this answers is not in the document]
THenry14: It's going down - we aer iterating on `.rev()` values here :)

crates/cast/src/main.rs
```diff
@@ -65,9 +65,17 @@ struct Cli {
...
+    #[clap(short = 'w', long)]
     wait: bool,
 
+    /// Adjusts the time after which --wait assumes transaction was not received or rejected
+    #[clap(short = 't', long)]
+    wait_timeout: Option<u16>,
+
+    /// Adjusts the time between consecutive attempts to fetch transaction by --wait flag
+    #[clap(long)]
+    wait_retry_interval: Option<u8>,
```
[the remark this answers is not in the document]
THenry14: not really, I just figured this is probably not something users will tinker with either way (at least not using cli, maybe rather cast config).

crates/cast/src/starknet_commands/script.rs
```diff
@@ -208,8 +208,17 @@ impl CairoHintProcessor<'_> {
...
+                } else {
+                    None
+                };
+                offset += 1;
+                let nonce = if inputs[offset] == 0.into() {
+                    offset += 1;
+                    Some(inputs[offset].clone().into_())
                 } else {
                     None
                 };
```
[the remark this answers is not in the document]
THenry14: we will soon move to use runtime extensions, this will probably be a part of a change as well

crates/cast/src/main.rs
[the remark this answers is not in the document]
DelevoXDG: I thought it might be reasonable to have it just to be consistent with `wait_timeout`, but on the other thought, it's probably gonna be used significantly less frequently

crates/cast/src/starknet_commands/script.rs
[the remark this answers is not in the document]
DelevoXDG: This change or some other?

DelevoXDG (approved):

crates/cast/src/lib.rs
```diff
@@ -344,10 +358,17 @@ pub async fn handle_wait_for_tx<T>(
     provider: &JsonRpcClient<HttpTransport>,
     transaction_hash: FieldElement,
     return_value: T,
-    wait: bool,
+    wait_config: WaitForTx,
 ) -> Result<T> {
-    if wait {
-        return match wait_for_tx(provider, transaction_hash, DEFAULT_RETRIES).await {
+    if wait_config.wait {
```
ddoktorski, line 363: If I understand correctly in the current solution when user passes `--wait-timeout` or `--wait-retry-interval` in a command without `--wait` flag, then both flags are ignored. Imo `--wait` flag should be used only to load wait config from Scarb.toml or to use default values.

crates/cast/src/starknet_commands/declare.rs
```diff
@@ -86,11 +91,17 @@ pub async fn declare(
     let casm_class_hash = casm_contract_definition.class_hash()?;
 
     let declaration = account.declare(Arc::new(contract_definition.flatten()?), casm_class_hash);
-    let execution = if let Some(max_fee) = max_fee {
+    // todo: refactor setting max_fee and nonce
```
abulenok, line 94: Let's add an issue for this

crates/cast/src/lib.rs
THenry14: > If I understand correctly in the current solution when user passes --wait-timeout or --wait-retry-interval in a command without --wait flag, then both flags are ignored. 

yes, because we're not waiting for the txs

> Imo --wait flag should be used only to load wait config from Scarb.toml or to use default values.

not sure I follow - atm you can use just `--wait`, and there are default values for both timeout and retry (or are taken from Scarb.toml, if defined). You do not need to pass --wait, if you don't want to wait 👍

crates/cast/src/starknet_commands/declare.rs
THenry14: https://github.com/foundry-rs/starknet-foundry/issues/1303

crates/cast/src/main.rs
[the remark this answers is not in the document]
ddoktorski: I think we should have either short forms for both flags or for neither of them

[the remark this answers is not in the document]
THenry14: k, changed

crates/cast/src/lib.rs
DelevoXDG: I think we should accept flags like these at all times. We don't need rpc url when using `account create`, but are still able to provide it, just like with any other config values set in cast config.

crates/cast/tests/e2e/script/declare.rs
```diff
@@ -21,7 +21,7 @@ async fn test_missing_field() {
         .args(args);
```
abulenok, line 21: The `e2e` tests with `--wait-timeout`, `--wait-retry-interval` flags (passed from CLI and parsed from Scarb.toml) are needed.

THenry14: yes, will be added in https://github.com/foundry-rs/starknet-foundry/issues/1201 , after we manage to add tests with integration network.

for now I have tested this manually, since this is virtaully not testable with devnet atm

abulenok: No proof the `PropertyFromCastConfig` approach works correctly when parsing stuff from Scarb.toml

crates/cast/src/helpers/scarb_utils.rs
```diff
@@ -42,14 +117,13 @@ pub fn get_profile<'a>(tool_sncast: &'a Value, profile: &Option<String>) -> Resu
...
+pub fn get_property<T>(tool: &Value, field: &str) -> T
 where
-    T: From<&'a str> + Default,
+    T: PropertyFromCastConfig + Default,
 {
     tool.get(field)
-        .and_then(Value::as_str)
-        .map(T::from)
-        .unwrap_or_default()
+        .and_then(T::from_toml_value)
```
abulenok, line 125: I wonder how this will distinguish between u8 and u16 and apply the correct default value?

THenry14: it infers the type, and the type must have implemented PropertyFromCastConfig for this to work, hence the lines 79-97

PR #1290, opened 2023-12-05
Scarb test collector

### LINKED ISSUE #417: Decouple snforge from scarb version

cptartur:
Closes #417

## Introduced changes

- Added support for scarb test collector. Needs https://github.com/software-mansion/scarb-nightlies/releases/tag/nightly-2023-12-04 of scarb to use new collector.

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


cptartur:
Changes to tests here: https://github.com/foundry-rs/starknet-foundry/pull/1292

crates/forge/src/lib.rs
```diff
@@ -89,8 +93,6 @@ fn to_runnable(
 /// * `corelib_path` - Absolute path to the Cairo corelib
 /// * `contracts` - Map with names of contract used in tests and corresponding sierra and casm artifacts
 /// * `predeployed_contracts` - Absolute path to predeployed contracts used by starknet state e.g. account contracts
```
piotmag769, line 95: Update the comment pls (was updated in my PR)

crates/forge-runner/src/lib.rs
```diff
@@ -151,10 +155,21 @@ impl ValidatedForkConfig {
...
+        let block_id = match value.block_id_type.to_lowercase().as_str() {
+            "number" => BlockId::Number(value.block_id_value.parse().unwrap()),
+            "hash" => BlockId::Hash(
+                Felt252::from(value.block_id_value.parse::<BigInt>().unwrap()).into_(),
+            ),
+            "tag" => {
+                assert_eq!(value.block_id_value, "Latest");
+                BlockId::Tag(Latest)
+            }
+            _ => unreachable!(),
```
MaksymilianDemitraszek, line 168: Ok with merging it for now but this defnitely is reachable and should be handled properly.

crates/forge/src/main.rs
```diff
@@ -177,52 +181,95 @@ fn test_workspace(args: TestArgs) -> Result<bool> {
...
-                    ),
-                    runner_config,
-                    runner_params,
-                    &forge_config.fork,
-                )
-                .await?;
-
-                let mut failed_tests = extract_failed_tests(tests_file_summaries);
-                all_failed_tests.append(&mut failed_tests);
+                if args.use_scarb_collector {
```
piotmag769, line 184: Some of the stuff in repeats itself, I would draw them out of the conditional statement

crates/forge/src/scarb_test_collector.rs
```diff
@@ -0,0 +1,16 @@
+use crate::collecting::CompiledTestCrateRaw;
```
MaksymilianDemitraszek, line 1: Either move `build_test_artifacts_with_scarb` here or move this to the `scarb` module

crates/forge/src/scarb_test_collector.rs
```diff
@@ -0,0 +1,16 @@
...
+use anyhow::Result;
+use camino::Utf8Path;
+
+/// Load test artifacts (deserialized compiled test crates) generated by `scarb snforge-test-collector`
```
MaksymilianDemitraszek, line 5: Please drop this comment, it does not provide any value.

crates/forge/src/scarb/config.rs
```diff
@@ -63,7 +61,7 @@ pub(crate) struct RawForkTarget {
     pub block_id: HashMap<String, String>,
 }
 
-pub(super) fn validate_raw_fork_config(raw_config: &RawForgeConfig) -> Result<()> {
+fn validate_raw_fork_config(raw_config: &RawForgeConfig) -> Result<()> {
```
MaksymilianDemitraszek, line 64: Shouldn't this return ForgeConfig? we can omit this by mistake right now

piotmag769 (approved):

crates/test-collector/src/lib.rs
```diff
@@ -338,59 +338,58 @@ fn extract_fork_config_from_args(db: &dyn SyntaxGroup, attr: &Attribute) -> Opti
...
+
+    None
+}
+
+// Only valid options are `BlockTag::Latest` and `Latest`
+fn tag_path_is_valid(tag_elements: &[PathSegment], db: &dyn SyntaxGroup) -> bool {
+    (tag_elements.len() == 1
+        || (tag_elements.len() == 2 && tag_elements[0].identifier(db).as_str() == "BlockTag"))
+        && tag_elements.last().unwrap().identifier(db).as_str() == "Latest"
 }
```
MaksymilianDemitraszek, line 393: Why there are two valid options here?

crates/test-collector/src/lib.rs
```diff
@@ -338,59 +338,58 @@ fn extract_fork_config_from_args(db: &dyn SyntaxGroup, attr: &Attribute) -> Opti
...
-                    .identifier(db)
-                    .to_string();
-                return match tag.as_str() {
-                    "Latest" => Some(BlockId::Tag(BlockTag::Latest)),
-                    _ => None,
-                };
+                let tag_elements = block_tag.elements(db);
+                if tag_path_is_valid(&tag_elements, db) {
+                    return Some("Latest".to_string());
+                }
```
MaksymilianDemitraszek, line 379: Probably should throw an error otherwise?

crates/test-collector/src/lib.rs
MaksymilianDemitraszek, line 1: Why any changes to this file are introduced here? If for consistency with scarb collector then we should drop them and do them in a separate PR to avoid any unwanted bugs.

crates/forge/src/scarb/config.rs
piotmag769: It shouldn't, it could take the ownership of `RawForgeConfig` and return the same instance of it though

crates/test-collector/src/lib.rs
piotmag769: We pretend that the stuff enclosed in attributes are type instances e.g. `BlockId::Tag(BlockTag::Latest)` but in fact we never even check if the types are present in the scope, if they are valid etc., we just treat path parts as strings. We should prob fix that someday, but when it was introduces no one really reviewed it carefully I guess (myself included)
cc @war-in

crates/test-collector/src/lib.rs
piotmag769: Yup, that's why they are here

crates/test-collector/src/lib.rs
piotmag769: We do, we exit the match statement resulting in returning `None` at the bottom of the function, which then later appends stuff to diagnostic errors

crates/forge-runner/src/lib.rs
piotmag769: It is not reachable, if you have `RawForkParams` it has to be valid

crates/forge/src/main.rs
cptartur: Most of them repeat, it was just the simples thing to do since we're removing it soon anyway.

crates/test-collector/src/lib.rs
cptartur: Assume mostly backwards compatibilty. Even our codebase uses `BlockId::Tag(Latest)` instead of `BlockId::Tag(BlockTag::Latest)` which I'd assume is the correct syntax.

crates/test-collector/src/lib.rs
MaksymilianDemitraszek: This is very convoluted, error should be returned by the validating function.

crates/test-collector/src/lib.rs
MaksymilianDemitraszek: @piotmag769 I assume those should be done before this one then.

crates/forge-runner/src/lib.rs
cptartur: Replaced this with `bail!` with proper error message

crates/test-collector/src/lib.rs
piotmag769: It has been written this way since the test collector existence, it is basically a copied approach from cairo test

crates/test-collector/src/lib.rs
piotmag769: @MaksymilianDemitraszek It made sense for the stack of PRs that existed before

crates/test-collector/src/lib.rs
cptartur: We allow using both `Latest` and `BlockTag::Latest` syntax for `BlockId::Tag`. Technically only `BlockTag::Latest` should be allowed type wise, but just `Latest` is already used extensively in multiple places.

crates/test-collector/src/lib.rs
cptartur: This can be changed later, but I'd do that only on the new collector side. No point in upgrading it here.

crates/test-collector/src/lib.rs
MaksymilianDemitraszek: Aren't those values produces by our code? Why we produce two different values?

cptartur: These are not produced by use. User provides them when creating a forking test with `#[fork(...)]` attribute. We just validate them here.

crates/forge-runner/src/lib.rs
cptartur: #1304

crates/test-collector/src/lib.rs
cptartur: #1305

PR #1292, opened 2023-12-05
Update cairo code for 2.4.0

cptartur:
Closes #

## Introduced changes

- Updated the cairo code to use `abi(embed_v0)` where possible instead of `external`, and removed unused variables.

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


piotmag769 (approved):

crates/cast/tests/data/contracts/no_casm/src/lib.cairo
```diff
@@ -1,7 +1,15 @@
...
     #[storage]
     struct Storage {}
-    #[external(v0)]
-    fn empty(ref self: ContractState) {}
+
+    #[abi(embed_v0)]
+    impl minimal_contractImpl of super::Iminimal_contract<ContractState> {
+        fn empty(ref self: ContractState) {}
+    }
 }
```
war-in, line 15: Soo we do have to define traits for all methods, right?

crates/cheatnet/tests/contracts/src/common/constructor_simple2.cairo
```diff
@@ -10,15 +16,17 @@ mod ConstructorSimple2 {
...
+        fn add_to_number(ref self: ContractState, number: felt252) -> felt252 {
+            let new_number = self.number.read() + number;
+            self.number.write(new_number);
+            new_number
+        }
 
-    #[external(v0)]
-    fn get_number(self: @ContractState) -> felt252 {
-        self.number.read()
+        #[abi(embed_v0)]
```
war-in, line 27: That's strange 🤔 I thought abi should be defined only once, above the trait

crates/forge/tests/integration/elect.rs
```diff
@@ -154,7 +154,6 @@ fn elect_complex() {
                 let elect_checker1 = IElectCheckerDispatcher { contract_address: contract.deploy(@ArrayTrait::new()).unwrap() };
                 let elect_checker2 = IElectCheckerDispatcher { contract_address: contract.deploy(@ArrayTrait::new()).unwrap() };
 
-                let old_seq_addr1 = elect_checker1.get_sequencer_address();
```
war-in, line 157: @drknzz This variable was never used. Should it be removed or maybe the test should be updated?

crates/forge/tests/integration/roll.rs
```diff
@@ -158,7 +158,6 @@ fn roll_complex() {
                 let roll_checker1 = IRollCheckerDispatcher { contract_address: contract.deploy(@ArrayTrait::new()).unwrap() };
                 let roll_checker2 = IRollCheckerDispatcher { contract_address: contract.deploy(@ArrayTrait::new()).unwrap() };
 
-                let old_block_number1 = roll_checker1.get_block_number();
```
war-in, line 161: Same as above. Maybe we should update the test rather than removing the value 🤔

crates/forge/tests/integration/syscalls.rs
```diff
@@ -104,15 +121,17 @@ fn library_call_syscall() {
...
-                    fn get_thing(self: @ContractState) -> felt252 {
-                        self.thing.read()
+                    #[abi(embed_v0)]
+                    impl ExecutorImpl of super::IExecutor<ContractState> {
+                        fn add_two(ref self: ContractState, number: felt252) -> felt252 {
+                            self.thing.write(10);
+                            number + 2
+                        }
+
+                        #[abi(embed_v0)]
```
war-in, line 131: 👀

crates/forge/tests/integration/test_state.rs
```diff
@@ -239,19 +250,21 @@ fn test_library_calls() {
...
-                       self.value.write(number);
+                    #[abi(embed_v0)]
+                    impl LibraryContractImpl of super::ILibraryContract<ContractState> {
+                        fn get_value(
+                            self: @ContractState,
+                        ) -> felt252 {
+                           self.value.read()
+                        }
+
+                        #[abi(embed_v0)]
```
war-in, line 261: 👀 👀

crates/forge/tests/integration/test_state.rs
```diff
@@ -362,7 +375,7 @@ fn test_cant_call_test_contract() {
                     }
         
 
-                    #[external(v0)]
+                    #[abi(embed_v0)]
```
war-in, line 378: So it is possible to write abi just above one function 😮 Maybe we shouldn't make traits with just one function then

war-in (changes requested):

crates/cast/tests/data/contracts/no_casm/src/lib.cairo
cptartur: I think technically you could still use `#[external(v0)]` because I don't think it got removed entirely (actually I've left it in one or two places because cast tests weren't passing). But generally from what I understand, yes. ` #[abi(embed_v0)]` will only work in context of impl of `starknet::interface` so we have to define them.

crates/cheatnet/tests/contracts/src/common/constructor_simple2.cairo
cptartur: Left over by accident probably, will recheck it

crates/cast/tests/data/contracts/no_casm/src/lib.cairo
war-in: I think this comment proves that it works without an interface
https://github.com/foundry-rs/starknet-foundry/pull/1292#discussion_r1420088407

war-in: Oh, so dispatchers would not have any methods, right?

war-in (approved):
Please wait for the @drknzz 🙏 
Otherwise approved ✅

THenry14 (approved):
looks good :+1:

crates/forge/tests/integration/elect.rs
drknzz: The test is okay, it should be removed 👍

crates/forge/tests/integration/roll.rs
drknzz: The elect test was copied from this one; so it's just an unused variable, the test does test what it's supposed to 👍

drknzz (approved):

piotmag769:
Do we plan to merge it? @cptartur

cptartur:
> Do we plan to merge it? @cptartur

We do. I didn't want to cairo bump to 2.4.0 with merging it.

PR #913, opened 2023-10-18
Benchmark performance

cptartur:
Closes #682

## Introduced changes

- Added benchmarks for collecting, compiling and running tests using Critierion.

## Checklist

- [x] Linked relevant issue
- [x] Updated relevant documentation
- [ ] Added relevant tests
- [x] Performed self-review of the code
- [x] Added changes to `CHANGELOG.md`

PR #1296, opened 2023-12-05
Revert "Revert "Benchmark performance (#913)" (#1075)"

cptartur:
Closes #

## Introduced changes

-

## Checklist

- [ ] Linked relevant issue
- [ ] Added relevant tests


crates/sncast/Cargo.toml
```diff
@@ -1,5 +1,5 @@
 [package]
-name = "cast"
```
piotmag769, line 2: [[bin]] section at the bottom could be removed now

piotmag769 (commented):
Don't we want to rename forge -> snforge too now?

THenry14:
I will review this more thoroughly later today. Would greatly appreciate it if this wasn't a part of todays release, as this looks like a can of worms.

> Don't we want to rename forge -> snforge too now?
> by piotmag769

I second that

cptartur:
> Can we run smoke tests on a build from the branch? (we should probably consider if it doesn't affect anything in the release process)

Locally running smoke tests seems to be working. But I'll try running a release from our repo fork to see if it works correctly.

cptartur:
@Arcticae releases seem to be working with these changes https://github.com/cptartur/starknet-foundry/actions/runs/7129508207

crates/sncast/Cargo.toml
```diff
@@ -1,5 +1,5 @@
...
+name = "sncast"
 description = "Cast - a Starknet Foundry CLI client"
```
THenry14, line 3:
```suggestion
description = "sncast - a Starknet Foundry CLI client"
```
I guess

crates/sncast/README.md
THenry14, line 1: This file contains `Cast` string, but I'm unable to suggest changes to it

crates/sncast/src/lib.rs
THenry14, line 1: This file contains `cast` string, but I'm unable to suggest change

crates/sncast/src/main.rs
THenry14, line 1: This file contains `cast` string, but I'm unable to suggest change

crates/sncast/tests/helpers/fixtures.rs
```diff
@@ -1,9 +1,9 @@
 use crate::helpers::constants::{ACCOUNT_FILE_PATH, CONTRACTS_DIR, DEVNET_ENV_FILE, URL};
```
THenry14, line 1: This file contains two `cast` string, but I'm unable to suggest change (although this is a test file, we should probably be consistent)

crates/sncast/tests/integration/wait_for_tx.rs
```diff
@@ -3,11 +3,11 @@ use crate::helpers::{
     fixtures::{create_test_provider, from_env},
```
THenry14, line 3: This file contains `cast` string, but I'm unable to suggest change

CONTRIBUTING.md
```diff
@@ -41,7 +41,7 @@ $ cargo test integration  # runs all integration tests
 $ cargo test e2e          # runs all e2e tests
 ```
 
-Similarly, to run cast tests make sure you are in `crates/cast` directory:
+Similarly, to run cast tests make sure you are in `crates/sncast` directory:
```
THenry14, line 44:
```suggestion
Similarly, to run sncast tests make sure you are in `crates/sncast` directory:
```
Not sure, but I guess we should change all recurences? I'm afraid it might be a bit confusing for users if we use cast/sncast interchangeably. @Arcticae ?

THenry14: Ok this actually is recurring thing in this PR, so I won't add comment per every file that contains this discrepancy, as this does not have much sense. Let's decide on naming convention first, I will continue the review then.

THenry14:
Let's also check github/CI files, as I think codeowners won't work atm

crates/sncast/tests/helpers/fixtures.rs
cptartur: These are not in the context of the tool but type casting

.github/CODEOWNERS
```diff
@@ -1,5 +1,5 @@
-# Cast
-/crates/cast/               @foundry-rs/starknet-foundry-cast
+# chain-interactions
+/crates/sncast/             @foundry-rs/starknet-foundry-cast
```
cptartur, line 2: The team could be renamed to `chain-interactions` as well, but that's for a separate PR.

README.md
```diff
@@ -27,7 +27,7 @@ Starknet Foundry, like its [Ethereum counterpart](https://github.com/foundry-rs/
 
 - [Forge](https://github.com/foundry-rs/starknet-foundry/tree/master/crates/forge): Starknet testing
   framework (like Truffle, Hardhat and DappTools but for Starknet).
-- [Cast](https://github.com/foundry-rs/starknet-foundry/tree/master/crates/cast): All-in-one tool for
+- [Cast](https://github.com/foundry-rs/starknet-foundry/tree/master/crates/sncast): All-in-one tool for
```
cptartur, line 30: Didn't change the link to `Sncast` because it would look weird/be inconsistent with Forge.

cptartur: Let me know if you think we should change it though.

Arcticae (approved):

README.md
THenry14: imo we should probably use one naming convention everywhere to avoid cast vs sncast confusion (and the same would go for forge). my understanding is that forge will be changed to snforge as well, in different PR, right? so do you plan to change the cast naming then?

@MaksymilianDemitraszek please take a look and have a say

THenry14 (commented):
looks good, after the issue raised here https://github.com/foundry-rs/starknet-foundry/pull/1296#discussion_r1422092676 is resolved.

if we go down the name changing route, we should probably also change naming in CHANGELOG.md, bug_report template, feature_request template, ci workflow and docs. Not sure about existing design docs, probably does not make much difference.

cptartur:
> looks good, after the issue raised here [#1296 (comment)](https://github.com/foundry-rs/starknet-foundry/pull/1296#discussion_r1422092676) is resolved.
> 
> if we go down the name changing route, we should probably also change naming in CHANGELOG.md, bug_report template, feature_request template, ci workflow and docs. Not sure about existing design docs, probably does not make much difference.

Discussed offline, for now, we want to rename crates to have `sn` prefix and leave names without prefix in docs etc. Can be changed if this turns out to cause confusion.

crates/sncast/Cargo.toml
cptartur: I don't see much benefit in that. I know that there are some default values for `[[bin]]` target, but we don't have to rely on them.

PR #1298, opened 2023-12-05
Add `script init` command

### LINKED ISSUE #1196: Add script init
Let add an `init` subcommand to `script`, that would create a script template, Scarb.toml and a directory structure for script.
Proposed way of invoking it is:

```
$ sncast script <myscript> --init
```

It then should create a direcoty structure:

```
$ ls .
scripts

$ tree
.
├── Scarb.toml
└── src
    ├── lib.cairo
    └── myscript.cairo
```

where the files could be:

`Scarb.toml`
```
[package]
name = "myscript"
version = "0.0.1"

[dependencies]
starknet = ">=2.3.1" # This should default to whatever scarb version we are using in repo atm
sncast_std = { git = "https://github.com/foundry-rs/starknet-foundry.git", tag = "v0.11.0" } # This should default to whatever te current version of snfoundry is

[lib]
sierra = true
casm = true
```

`src/lib.cairo`
```
mod myscript;
```

`src/myscript.cairo`
```
use sncast_std;
use debug::PrintTrait;

fn main() {
    'Put your code here!'.print();
}
```

There should be tests written for added code, the docs should be updated as well.

ddoktorski:
Closes #1196 

## Introduced changes

- Command to initialize template for deployment scripts

Please note that change in the command to run script, namely to have `script run <MODULE_NAME>` instead of `script <MODULE_NAME>` will be addressed in a separate PR.

## Checklist

- [X] Linked relevant issue
- [X] Added relevant tests


crates/cast/src/starknet_commands/script/init.rs
```diff
@@ -0,0 +1,141 @@
...
+        get_cairo_version(&scarb_manifest_path).context("Failed to obtain cairo version")?;
+    let starknet_dependency = format!("starknet@>={cairo_version}");
+
+    ScarbCommand::new()
+        .current_dir(script_root_dir)
+        .args(["--offline", "add", &starknet_dependency])
+        .run()?;
+
+    Ok(())
+}
```
THenry14, line 103: is there is there a reason not to combine those two functions?

crates/cast/src/starknet_commands/script/init.rs
```diff
@@ -0,0 +1,141 @@
...
+        .map_or_else(
+            || current_dir.join(SCRIPTS_DIR).join(script_name),
+            |_| current_dir.join(script_name),
+        );
+
+    Utf8PathBuf::from_path_buf(script_root_dir_path)
+        .map_err(|_| anyhow!("Failed to obtain script root dir path"))
+}
+
+fn init_scarb_project(script_name: &str, script_root_dir: &Utf8PathBuf) -> Result<()> {
```
THenry14, line 46: does this need to be a separate function?

crates/cast/src/starknet_commands/script/init.rs
```diff
@@ -0,0 +1,141 @@
...
+        .context("Failed to add sncast_std dependency to Scarb.toml")?;
+    add_starknet_dependency(script_root_dir)
+        .context("Failed to add starknet dependency to Scarb.toml")?;
+
+    Ok(())
+}
+
+fn add_sncast_std_dependency(script_root_dir: &Utf8PathBuf) -> Result<()> {
+    let cast_version = env!("CARGO_PKG_VERSION");
+    let cast_version = format!("v{cast_version}");
```
THenry14, line 73:
```suggestion
    let cast_version = format!("v{}", env!("CARGO_PKG_VERSION"));
```
I guess this should work?

crates/cast/src/starknet_commands/script/init.rs
```diff
@@ -0,0 +1,141 @@
...
+            fn main() {
+                'Put your code here!'.print();
+            }
+        "},
+    )?;
+
+    Ok(())
+}
+
+fn overwrite_lib_file(script_name: &str, script_root_dir: &Utf8PathBuf) -> Result<()> {
```
THenry14, line 130: why `overwrite` ?

crates/cast/src/starknet_commands/script/init.rs
```diff
@@ -0,0 +1,141 @@
...
+
+    ScarbCommand::new()
+        .current_dir(script_root_dir)
+        .args(["--offline", "add", &starknet_dependency])
+        .run()?;
+
+    Ok(())
+}
+
+fn modify_files_in_src_dir(script_name: &str, script_root_dir: &Utf8PathBuf) -> Result<()> {
```
THenry14, line 105: we should probably check if such files/packages does not exist yet, and only then call proceed with making files.

btw, I guess you want to bring some structure, but not sure making a function that just calls two other functions is the best approach - maybe we could move all the logic here instead? let's discuss.

crates/cast/src/starknet_commands/script/mod.rs
```diff
@@ -0,0 +1,19 @@
+use crate::starknet_commands::script::init::Init;
+use clap::{Args, Subcommand};
+
+pub mod init;
+pub mod run;
+
+#[derive(Args)]
+pub struct Script {
+    /// Module name that contains the `main` function, which will be executed
+    pub script_module_name: Option<String>,
```
THenry14, line 10: this should probably not be an Option

docs/src/appendix/cast/script.md
```diff
@@ -1,7 +0,0 @@
-# `script`
```
THenry14, line 1: This is temporary, until we have `run` command, right? I guess we could leave this, have it to act like a kind of overview (like for example in `account`), link to init + store `script` in here as well. wdyt?

THenry14:
let's also merge/rebase to master before the next round of review. Overall, great work mate 💪

crates/cast/src/main.rs
```diff
@@ -354,6 +345,35 @@ async fn run_async_command(
...
+    config: &CastConfig,
+    script: &Script,
+    value_format: ValueFormat,
+) -> Result<()> {
+    if let Some(starknet_commands::script::Commands::Init(init)) = &script.command {
+        let mut result = starknet_commands::script::init::init(init);
+        print_command_result("script init", &mut result, value_format, cli.json)?;
+    } else {
+        let provider = get_provider(&config.rpc_url)?;
+        let script_module_name = script.script_module_name.as_ref().ok_or_else(|| {
```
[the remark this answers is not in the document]
ddoktorski: If you mean `script_module_name` error, once there is `script run` added it will be handled by clap. I will add PR for this shortly

crates/cast/src/starknet_commands/script/init.rs
ddoktorski: Same comment as [here](https://github.com/foundry-rs/starknet-foundry/pull/1298/files#r1417573712)

docs/src/appendix/cast/script/script.md
```diff
@@ -0,0 +1,7 @@
+# `script`
+Compile and run a cairo deployment script or initialize script template by using subcommand [`init`](./init.md)
+
+## `<MODULE_NAME>`
+Required.
+
+Module name that contains the 'main' function that will be executed.
```
abulenok, line 7: Let's the argument name consistent with `init`

```suggestion
## `<SCRIPT_NAME>`
Required.

Script module name that contains the 'main' function that will be executed.
```

crates/cast/src/starknet_commands/script/init.rs
```diff
@@ -0,0 +1,140 @@
...
+
+pub fn init(init_args: &Init) -> Result<ScriptInitResponse> {
+    let script_root_dir_path = get_script_root_dir_path(&init_args.script_name)?;
+
+    init_scarb_project(&init_args.script_name, &script_root_dir_path)?;
+    add_dependencies(&script_root_dir_path)?;
+    modify_files_in_src_dir(&init_args.script_name, &script_root_dir_path)?;
+
+    Ok(ScriptInitResponse {
+        status: format!("Successfully initialized `{}`", init_args.script_name),
```
abulenok, line 26: We should consider unifying status messages between run and init (introduce a `ScriptStatus` enum for example). Not necessary in this PR.

crates/cast/src/starknet_commands/script/init.rs
```diff
@@ -0,0 +1,140 @@
...
+
+fn get_script_root_dir_path(script_name: &str) -> Result<Utf8PathBuf> {
+    let current_dir = std::env::current_dir()?;
+
+    let script_root_dir_path = current_dir
+        .file_name()
+        .and_then(|dir_name| dir_name.to_str())
+        .filter(|&dir_name| dir_name == SCRIPTS_DIR)
+        .map_or_else(
+            || current_dir.join(SCRIPTS_DIR).join(script_name),
```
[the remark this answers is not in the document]
abulenok: This looks counterintuitive to me

crates/cast/src/starknet_commands/script/init.rs
ddoktorski: > we should probably check if such files/packages does not exist yet, and only then call proceed with making files.

I am not sure about this and open to further discussion
- if files don't exist they will be created unless there was an issue with `init_scarb_project` (there must be correct directories structure which assume is true after this function is successful)
- if we try to create package that already exists then `scarb new` will fail with
```
error: Failed to init Scarb project: `scarb metadata` exited with error
```

docs/src/appendix/cast/script/script.md
ddoktorski: Apologies these docs are confusing now. It is correct as it is now, because this is `MODULE_NAME` used to run script, not to init. I will fix it in next PR with `script run`.

crates/cast/src/starknet_commands/script/init.rs
THenry14: I'm still not convinced, as this function does nothing more than wrapping ScarbCommand::new. This is not reused anywhere nor does it modify the arguments in any way. I get it's a matter of preference, I just have my doubts about splitting every single thing into its own function - sometimes it makes things too overwhelming and decreases the readability.

Of course all things we've mentioned here are probably more of a subjective matter. I am rather against such shredding into smaller functions, but am more than happy to discuss if you guys prefer it the other way.

crates/cast/src/starknet_commands/script/init.rs
THenry14: I have mixed feelings, as right now we actually use 3 functions to just add 2 dependencies (add_starknet_dependency, add_sncast_std_dependency and add_dependencies). I'm not sure we need all three to be honest. I

crates/cast/src/starknet_commands/script/init.rs
THenry14: just improving on this error message (eg adding context that package already exists) should be enough :+1:

crates/cast/src/starknet_commands/script/init.rs
THenry14: if not in this PR, let's create an issue and link it here :+1:

crates/cast/src/starknet_commands/script/init.rs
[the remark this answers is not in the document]
THenry14: > It's potentially useful when the user inits scripts project inside scarb package with contracts, but may be annoying when trying to init a scripts project in an arbitrary location.

that's true. I'm afraid that verifying this will add too much of a complexity - after all there might be some edge scenarios where we are (for example) in some workspace where do just want to have scripts, and the contracts are in differen workspace? (not sure, I'm just guessing here).

crates/cast/tests/e2e/script/init.rs
```diff
@@ -0,0 +1,141 @@
...
+        command: script init
+        status: Successfully initialized `{script_name}`
+    "});
+
+    assert!(scripts_dir_path.join(script_name).exists());
+    assert!(!scripts_dir_path
+        .join(SCRIPTS_DIR)
+        .join(script_name)
+        .exists());
+}
```
THenry14, line 141: let's also add test case for failed creation

crates/cast/src/starknet_commands/script/init.rs
ddoktorski: > sometimes it makes things too overwhelming and decreases the readability.

I would say it is exactly the opposite

crates/cast/src/starknet_commands/script/init.rs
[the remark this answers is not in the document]
THenry14: On a second though - what do you guys think about such idea: we check if there is a `scripts` directory in cwd, if it is - we create a package dir in there, if it isn't we create pkg dir in cwd. After init we print out an info, something like "New package {xyz} created in {path}". So - we do not create scripts directory at all. This should simplify a few things, and we do not have this problem with removing empty scripts dir.

crates/cast/src/starknet_commands/script/init.rs
THenry14: In order to move things forward let's resolve this discussion and leave your changes, as this is not something inherently wrong. I would like to take more time on defining one approach for the whole project, because I don't like that we have different approaches for different subcommands. This may, most probably, become a part of Cast Refactor milestone. Sounds good?

crates/cast/src/starknet_commands/script/init.rs
THenry14: as per https://github.com/foundry-rs/starknet-foundry/pull/1298#discussion_r1417477652

crates/cast/src/starknet_commands/script/init.rs
```diff
@@ -0,0 +1,158 @@
...
+    add_sncast_std_dependency(script_root_dir)
+        .context("Failed to add sncast_std dependency to Scarb.toml")?;
+    add_starknet_dependency(script_root_dir)
+        .context("Failed to add starknet dependency to Scarb.toml")?;
+
+    Ok(())
+}
+
+fn add_sncast_std_dependency(script_root_dir: &Utf8PathBuf) -> Result<()> {
+    let cast_version = format!("v{}", env!("CARGO_PKG_VERSION"));
```
THenry14, line 84: Is this variable available in built package?

crates/cast/src/starknet_commands/script/init.rs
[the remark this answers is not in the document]
abulenok: From the user's perspective, this approach might be inconsistent. 
Whenever I use some CLI tool init command, I expect init to be executed in cwd.

Maybe a flag? Smth like `--path=<path-to-folder-where-script-project-will-be-initialized>`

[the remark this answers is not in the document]
ddoktorski: It also makes sense to me. What do you think @THenry14?

crates/cast/src/starknet_commands/script/init.rs
ddoktorski: If I understand correctly this variable is captured at compile time and is embedded into the binary as a static string.

crates/cast/src/starknet_commands/script/init.rs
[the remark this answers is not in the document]
THenry14: I'd like to avoid adding a new flag to be honest. I think the behaviour where we create a package dir in cwd OR if there is scripts dir in cwd we create package dir in cwd/scripts makes the most sense here, as it's reasonable to expect we are in some workspace then. 

That would be probably also more consistent with cargo and scarb. It is reasonable to expect that tests are in tests directory, I guess the same should go for scripts. 

Of source, that would mean we should print out to the user the path where the script pkg was created. thoughts?

crates/cast/src/starknet_commands/script/init.rs
THenry14: Cool, just wanted to make sure. I guess we can just verify by building the binary, and trying to run the init, to be double-sure.

ddoktorski: I can confirm that it works as expected.

crates/cast/src/starknet_commands/script/init.rs
ddoktorski: https://github.com/foundry-rs/starknet-foundry/issues/1408

crates/cast/src/starknet_commands/script/init.rs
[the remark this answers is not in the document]
abulenok: hm, ok we can do that, but maybe let's impose stronger conditions that need to be met to create a script package in the `scripts` dir. Like the existence `Scarb.toml` file and `src` folder in cwd. WDYT?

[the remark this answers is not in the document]
THenry14: Going back to this:
> I think the behaviour where we create a package dir in cwd OR if there is scripts dir in cwd we create package dir in cwd/scripts makes the most sense here

now that I think about it, let's be stricter. let's require `scripts` directory, and drop the 'or create pkg dir in cwd'. This may become too complex to support, and as I said - it's not unreasonable to expect the scripts be in scripts directory. If someone wants to do it any other way, init command isn't for this person and it can always be made "manually"

> hm, ok we can do that, but maybe let's impose stronger conditions that need to be met to create a script package in the scripts dir. Like the existence Scarb.toml file and src folder in cwd. WDYT?

this may make sense. If toml and src are present, I'd suggest we just add a new script file + include it in mod. 
if they are not present (but the dir exists), we could probably just error out?
if the dir is not present, let's create it (along with toml, src, script file).

Does that make sense to you? or have I forgotten about something?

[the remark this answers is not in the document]
ddoktorski: > let's require scripts directory, and drop the 'or create pkg dir in cwd'.

Is there a plan to include the creation of a `scripts` folder in the `snforge init` command? If not, users will have to create this manually every time.

> If toml and src are present, I'd suggest we just add a new script file + include it in mod.

Should we do this only when we're in the `scripts` directory? Is there another way to verify that a Scarb project in the cwd is associated with a script? Generally I'm not sure that extending `script init` to add a new script file and include it in the `mod` is in the scope of `script init`.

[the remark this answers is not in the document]
THenry14: > Is there a plan to include the creation of a scripts folder in the snforge init command? If not, users will have to create this manually every time.

snforge has nothing to do with this, sncast should do this (in init subcommand)

> Should we do this only when we're in the scripts directory? Is there another way to verify that a Scarb project in the cwd is associated with a script? Generally I'm not sure that extending script init to add a new script file and include it in the mod is in the scope of script init.

this one is very tricky indeed. I think this _could_ be in scope, but it depends on us it it _will be_. If this it too complicated I'd just fall back to init just be able to create a simple layout (eg scripts directory with one example script inside, and nothing more)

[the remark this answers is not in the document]
ddoktorski: As we discussed previously, let's keep things straightforward:
1. The `script init` command should be used once per project to set up a template for a deployment script.
2. Always create a `scripts` directory and nest a script package directory within it. If there's already a `scripts` directory in the cwd, it should throw an error.

Please let me know if someone disagrees with this approach.

crates/sncast/src/main.rs
```diff
@@ -402,6 +400,36 @@ async fn run_async_command(
...
+    numbers_format: NumbersFormat,
+    output_format: &OutputFormat,
+) -> Result<()> {
+    if let Some(starknet_commands::script::Commands::Init(init)) = &script.command {
+        let mut result = starknet_commands::script::init::init(init);
+        print_command_result("script init", &mut result, numbers_format, output_format)?;
+    } else {
+        let provider = get_provider(&config.rpc_url)?;
+        let script_module_name = script.script_module_name.as_ref().ok_or_else(|| {
+            anyhow!("required positional argument SCRIPT_MODULE_NAME not provided")
```
ddoktorski, line 417: This line is a temporary solution and it will be removed in https://github.com/foundry-rs/starknet-foundry/pull/1320

crates/sncast/src/starknet_commands/script/init.rs
```diff
@@ -0,0 +1,154 @@
+use anyhow::{ensure, Context, Ok, Result};
+use camino::Utf8PathBuf;
+use std::fs;
+
+use clap::Args;
+use indoc::{formatdoc, indoc};
+use scarb_metadata::ScarbCommand;
```
THenry14, line 7: Let's use `ScarbCommand` from scarb-api crate in our repo instead

crates/sncast/src/starknet_commands/script/init.rs
```diff
@@ -0,0 +1,154 @@
...
+    let script_main_file_path = script_root_dir.join("src").join(script_main_file_name);
+
+    fs::write(
+        script_main_file_path,
+        indoc! {r"
+            use sncast_std;
+            use debug::PrintTrait;
+
+            fn main() {
+                'Put your code here!'.print();
```
THenry14, line 129: [nitpick]: maybe we could put something more usable here as well? Having a simple `call` to some simple contract displaying "hello world" on testnet would be soo cool, and would actually show at least some of the scripts' functionality

crates/sncast/tests/e2e/script/init.rs
```diff
@@ -0,0 +1,135 @@
+use camino::Utf8PathBuf;
+use indoc::{formatdoc, indoc};
+use snapbox::cmd::{cargo_bin, Command};
+use sncast::helpers::constants::SCRIPTS_DIR;
+use sncast::helpers::scarb_utils::get_cairo_version;
+use tempfile::TempDir;
+
+#[test]
+fn test_script_init_files_created() {
```
THenry14, line 9:
```suggestion
fn test_script_happy_case() {
```

nitpick, to conform to conventions already present in repo

crates/sncast/tests/e2e/script/init.rs
```diff
@@ -0,0 +1,135 @@
...
+        status: Successfully initialized `{script_name}` at [..]/scripts/{script_name}
+    "});
+
+    let script_dir_path = temp_dir.path().join(SCRIPTS_DIR).join(script_name);
+    let scarb_toml_path = script_dir_path.join("Scarb.toml");
+
+    let scarb_toml_content = std::fs::read_to_string(&scarb_toml_path).unwrap();
+    let lib_cairo_content = std::fs::read_to_string(script_dir_path.join("src/lib.cairo")).unwrap();
+    let main_file_content =
+        std::fs::read_to_string(script_dir_path.join(format!("src/{script_name}.cairo"))).unwrap();
```
THenry14, line 53: This basically tests the same thing as previous test, right? so maybe we do not need the previous test at all?

docs/src/appendix/cast/script/init.md
```diff
@@ -0,0 +1,17 @@
+# `init`
+Create script template and directories structure.
+
+The command creates the following structure inside `scripts` directory:
```
THenry14, line 4: It also creates the scripts directory itself, so perhaps we should reprase this a bit?

crates/sncast/tests/e2e/script/init.rs
ddoktorski: The initial idea was to have one test that checks files structure and another that checks files content. However if the structure is wrong, the second test will throw an error anyway, hence I am fine with removing the first test.

crates/cast/src/starknet_commands/script/init.rs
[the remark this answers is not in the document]
abulenok: I don't understand why we cannot run `init` multiple times in the same cwd with distinct script names?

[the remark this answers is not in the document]
THenry14: This increases the complexity of init unnecessarily. It ought to be a simple, straightforward way to create a script package with the directory layout etc, no need to try to cover all the possibilities. After all this is _init_, I don't think it's unreasonable to leave it up to the users to further modify the structure if needed (add more packages/scripts - of course we will need to improve our docs).

Further down the line, if there is a demand, we can think about expanding it. At the moment, in my opinion, it is sufficient as is.

docs/src/appendix/cast/script/init.md
```diff
@@ -0,0 +1,18 @@
...
+        ├── Scarb.toml
+        └── src
+            ├── lib.cairo
+            └── my_script.cairo
+```
+
+## `<SCRIPT_NAME>`
+Required.
+
+Name of a script to create.
```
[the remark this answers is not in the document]
ddoktorski: There's no need to run `script init` within existing scarb project. Theoretically, it can be executed from any location.

crates/sncast/src/starknet_commands/script/init.rs
```diff
@@ -0,0 +1,159 @@
...
+            use sncast_std::{call, CallResult};
+
+            // The example below uses a contract deployed to the Goerli testnet
+            fn main() {
+                let contract_address = 0x7ad10abd2cc24c2e066a2fee1e435cd5fa60a37f9268bfbaf2e98ce5ca3c436;
+                let call_result = call(contract_address.try_into().unwrap(), 'get_greeting', array![]);
+                let call_result = *call_result.data[0];
+                let mut greeting: ByteArray = "";
+                greeting.append_word(call_result, 16);
+                println!("{greeting:?}");
```
ddoktorski, line 134: As I mentioned before this not an ideal way to print a shortstring. In the [ByteArrayImpl](https://github.com/starkware-libs/cairo/blob/main/corelib/src/byte_array.cairo#L49-L50) we can read that `append_word` should be a private function.

ddoktorski, line 134: As I mentioned before this is not an ideal way to print a shortstring. In the [ByteArrayImpl](https://github.com/starkware-libs/cairo/blob/133ee0e5b2f1ef12696fce781212a201e3018b2b/corelib/src/byte_array.cairo#L49-L50) we can read that `append_word` should be a private function.

THenry14: https://github.com/foundry-rs/starknet-foundry/pull/1622

I believe this PR might mitigate your problem; I guess you can rebase onto it and just print call_result directly.

abulenok: CallResult contains an array of felts, which will be printed as felts, not as short strings in my PR

THenry14: but we can use the traits you defined to just print call results, which is our main objective. this PR should not try to resolve any problems with printing short strings, this is a separate problem.

ddoktorski: I guess if we make direct use of this trait to print call results in this case, it will result in printing something like `[96231036770510887841935600920878740513]`

abulenok: alternatively, you can add smth like `assert(call_result == 'blabla')`

crates/sncast/src/helpers/constants.rs
```diff
@@ -30,3 +30,5 @@ pub const CREATE_KEYSTORE_PASSWORD_ENV_VAR: &str = "CREATE_KEYSTORE_PASSWORD";
 
 pub const SCRIPT_LIB_ARTIFACT_NAME: &str = "__sncast_script_lib";
 pub const CONFIG_FILENAME: &str = "snfoundry.toml";
+
+pub const SCRIPTS_DIR: &str = "scripts";
```
THenry14, line 34:
```suggestion
pub const INIT_SCRIPTS_DIR: &str = "scripts";
```
maybe, to avoid confusion? technically scripts dir can be anything, and this is only used for init

crates/sncast/src/starknet_commands/script/init.rs
```diff
@@ -0,0 +1,157 @@
...
+        script_main_file_path,
+        indoc! {r"
+            use sncast_std::{call, CallResult};
+
+            // The example below uses a contract deployed to the Goerli testnet
+            fn main() {
+                let contract_address = 0x7ad10abd2cc24c2e066a2fee1e435cd5fa60a37f9268bfbaf2e98ce5ca3c436;
+                let call_result = call(contract_address.try_into().unwrap(), 'get_greeting', array![]);
+                let call_result = *call_result.data[0];
+                assert(call_result=='Hello, Starknet!', '');
```
THenry14, line 132: we can now also add `println!("call_result: {}", call_result);` here as well, just to output something, right?

crates/sncast/src/starknet_commands/script/mod.rs
```diff
@@ -0,0 +1,23 @@
...
+pub mod run;
+
+#[derive(Args)]
+pub struct Script {
+    /// Module name that contains the `main` function, which will be executed
+    pub module_name: Option<String>,
+
+    /// Specifies scarb package to be used
+    #[clap(long)]
+    pub package: Option<String>,
```
THenry14, line 14: Since we've added packages support, it now may make sense to use it for init as well (if passed) right?
we could create multiple script packages inside scripts directory + now we probably also should make sure this directory is included in a workspace (since we now support them). Let's try to assess, we might want to consult scarb team on how we should approach this (maybe scarb has a mechanism that fits our needs?)

THenry14 (commented):
Let's also update the `script` docs with an example of init (+ there is a statement that init will be added, so we must remove it)

crates/sncast/src/starknet_commands/script/mod.rs
ddoktorski: The initial plan was for #1320 to follow immediately after this one, so there shouldn't be any problems because the `package` flag will be used solely for the `run` subcommand.

We can change the order and modify the #1320 allowing it to be merged prior to this PR.

crates/sncast/src/starknet_commands/script/init.rs
ddoktorski: Are we sure that it makes sense to print this?
```
call_result: [96231036770510887841935600920878740513]
```

THenry14: I'd say so, yeah. I know it may look weird, but it shows off a potentially important feature - being able to directly print result structs

crates/sncast/src/starknet_commands/script/mod.rs
THenry14: as you prefer, it does not make much difference to me tbh

ddoktorski: Let's stick to the current order then.

crates/sncast/src/starknet_commands/script/init.rs
```diff
@@ -0,0 +1,157 @@
...
+    fs::write(
+        script_main_file_path,
+        indoc! {r#"
+            use sncast_std::{call, CallResult};
+
+            // The example below uses a contract deployed to the Goerli testnet
+            fn main() {
+                let contract_address = 0x7ad10abd2cc24c2e066a2fee1e435cd5fa60a37f9268bfbaf2e98ce5ca3c436;
+                let call_result = call(contract_address.try_into().unwrap(), 'get_greeting', array![]);
+                assert(*call_result.data[0]=='Hello, Starknet!', '');
```
abulenok, line 131:
```suggestion
                assert(*call_result.data[0]=='Hello, Starknet!', *call_result.data[0]);
```

crates/sncast/tests/e2e/script/init.rs
```diff
@@ -0,0 +1,137 @@
...
+    assert_eq!(
+        main_file_content,
+        indoc! {r#"
+            use sncast_std::{call, CallResult};
+
+            // The example below uses a contract deployed to the Goerli testnet
+            fn main() {
+                let contract_address = 0x7ad10abd2cc24c2e066a2fee1e435cd5fa60a37f9268bfbaf2e98ce5ca3c436;
+                let call_result = call(contract_address.try_into().unwrap(), 'get_greeting', array![]);
+                assert(*call_result.data[0]=='Hello, Starknet!', '');
```
abulenok, line 68:
```suggestion
                assert(*call_result.data[0]=='Hello, Starknet!', *call_result.data[0]);
```

crates/sncast/src/starknet_commands/script/init.rs
```diff
@@ -0,0 +1,162 @@
...
+pub fn init(init_args: &Init) -> Result<ScriptInitResponse> {
+    let script_root_dir_path = get_script_root_dir_path(&init_args.script_name)?;
+
+    init_scarb_project(&init_args.script_name, &script_root_dir_path)?;
+
+    let modify_files_result = add_dependencies(&script_root_dir_path)
+        .and_then(|()| modify_files_in_src_dir(&init_args.script_name, &script_root_dir_path));
+
+    print_as_warning(
+        &anyhow!("The newly created script isn't auto-added to the workspace. You may need to manually add it to your workspace configuration in Scarb.toml")
```
THenry14, line 28: nit: Let's link to the docs that explain what to do exactly.

docs/src/starknet/script.md
```diff
@@ -35,12 +35,21 @@ Some of the planned features that will be included in future versions are:
...
+### Initialize a script
+
+To get started, a deployment script with all required elements can be initialized using the following command:
+
+```shell
+$ sncast script init my_script
+```
+
+For more details, see [init command](../appendix/sncast/script/init.md).
+
```
THenry14, line 52: nit: In a `Note` field, let's add more specific information about aht to add (and where) if we are working in a workspace. Scarb docs for workspaces could be linked in here to add more context, as well.

THenry14 (approved):
a nitpicky comments about improving docs and messaging. apart from that, lgtm :+1:

ddoktorski:
https://github.com/foundry-rs/starknet-foundry/issues/1733

THenry14 (approved):

PR #1301, opened 2023-12-06
Add get_nonce to cast_std

THenry14:
Closes https://github.com/foundry-rs/starknet-foundry/issues/1230

## Introduced changes

- add `get_nonce` function to cast_std, allowing to manually retrieve nonce to be passed to other functions

Docs in different PR.

## Checklist

- [X] Linked relevant issue
- [X] Added relevant tests


crates/cast/src/lib.rs
```diff
@@ -163,6 +163,20 @@ pub fn decode_chain_id(chain_id: FieldElement) -> String {
...
+    block_id: &str,
+    address: FieldElement,
+) -> Result<FieldElement> {
+    Ok(provider
+        .get_nonce(
+            get_block_id(block_id).expect("Could not obtain block id"),
+            address,
+        )
+        .await
+        .expect("Could not get current nonce"))
```
DelevoXDG, line 177: Nitpick, but this function is not necessarily for fetching a "current" nonce, but for any given `block_id`, right?

THenry14: For now only for block tags (pending, latest). Removed word "current" in newest commit.

DelevoXDG (approved):
lgtm 🤩

crates/cast/src/lib.rs
THenry14: no use for it currently, that's the reason for limitation (and it is limited on scripts side, so actually make sense to remove "current" from this function)

PR #1333, opened 2023-12-08
remove test collector from forge

tkumor3:
Closes #

## Introduced changes

-

## Checklist

- [ ] Linked relevant issue
- [ ] Added relevant tests


crates/forge/src/lib.rs
```diff
@@ -129,7 +84,7 @@ pub async fn run(
 /// * `runner_params` - A struct with parameters required to run tests e.g. map with contracts
 /// * `fork_target` - A configuration of forks used in tests
 #[allow(clippy::implicit_hasher)]
-pub async fn run_with_scarb_collector(
+pub async fn run(
```
cptartur, line 87: Run internal could be inlined into this function and removed. There's no point in splitting them

crates/forge/src/compiled_raw.rs
cptartur, line 1: Do structs in this module need to be `pub`?

cptartur, line 1: I think since these are only needed for deserializing artifacts from scarb, we could move them into a module inside `scarb`.

And `to_runneable` could be moved there as well so we can hide this raw artifacts from the developers in other part of the code. 

I'd introduce a function that loads artifacts from scarb, parses fork configs there and outputs runneable structs. This way logic in `lib.rs` could be changed so it already handes only runneable code.

crates/forge/src/scarb_test_collector.rs
cptartur, line 1: Code from here could probably be moved to `scarb.rs` since there's just one collector now.

crates/forge/tests/data/file_reading/.snfoundry_cache/.prev_tests_failed
cptartur, line 1: Should this be commited?

cptartur (changes requested):

crates/forge/src/compiled_raw.rs
cptartur: Discussed offline: will be done in a separate PR.

cptartur (approved):

snforge_std/Scarb.toml
```diff
@@ -3,4 +3,4 @@ name = "snforge_std"
 version = "0.1.0"
 
 [dependencies]
-starknet = "2.3.1"
+starknet = "2.4.0"
```
piotmag769, line 6: It is fine but please remember to not bump it in later bumps (it should stay set to the lowest compatible scarb version, otherwise the test collector thing is useless)

piotmag769 (approved):
Looks like my code, so I approve xD

PR #1337, opened 2023-12-11
Universal sierra -> casm compilation library

### LINKED ISSUE #1211: Implement rust library for unversal compilation of `Sierra` -> `CASM`
It should be a separate package, which allows the user to provide any sierra,
and it will automatically detect & compile it into CASM

war-in:
Closes #1211

## Introduced changes

- new sierra-casm crate
- sierra-casm exposes the `compile` function which takes `FlattenedSierraClass` from starknet-rs and returns `CasmContractClass`

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/cheatnet/src/forking/state.rs
MaksymilianDemitraszek, line 1: Can we do migration in a separate PR? We need to make sure we removed all related structures and code.

crates/sierra-casm/src/lib.rs
```diff
@@ -0,0 +1,63 @@
...
+use cairo_lang_starknet::contract_class::{ContractClass, ContractEntryPoints};
+use cairo_lang_starknet_v1_0_0_alpha6::casm_contract_class::CasmContractClass as CasmContractClassV1Alpha;
+use cairo_lang_starknet_v1_0_0_alpha6::contract_class::ContractClass as ContractClassV1Alpha;
+use cairo_lang_starknet_v1_0_0_rc0::casm_contract_class::CasmContractClass as CasmContractClassV1Rc0;
+use cairo_lang_starknet_v1_0_0_rc0::contract_class::ContractClass as ContractClassV1Rc0;
+use cairo_lang_starknet_v1_1_1::casm_contract_class::CasmContractClass as CasmContractClassV1_1;
+use cairo_lang_starknet_v1_1_1::contract_class::ContractClass as ContractClassV1_1;
+
+pub fn compile(definition: &FlattenedSierraClass) -> anyhow::Result<CasmContractClass> {
+    macro_rules! compile_contract {
```
MaksymilianDemitraszek, line 18: This macro is not very readable, this is imo not a good use case, can we just write code instead?

crates/sierra-casm/src/lib.rs
```diff
@@ -0,0 +1,63 @@
...
+
+    let json = serde_json::json!({
+        "sierra_program": converted_sierra_program,
+        "contract_class_version": value.contract_class_version,
+        "entry_points_by_type": converted_entry_points,
+    });
+    serde_json::from_value::<T>(json)
+}
+
+fn casm_to_main_casm<T>(value: &T) -> anyhow::Result<CasmContractClass>
```
MaksymilianDemitraszek, line 57: I don't undestand what this function does.

MaksymilianDemitraszek (changes requested):
Overall LGTM, but please migrate forge in a separate PR, we need to make sure we remove all related code.

crates/sierra-casm/src/lib.rs
```diff
@@ -0,0 +1,63 @@
...
+            }
+        }};
+    }
+
+    compile_contract!(ContractClass, CasmContractClass);
+    compile_contract!(ContractClassV1_1, CasmContractClassV1_1);
+    compile_contract!(ContractClassV1Rc0, CasmContractClassV1Rc0);
+    compile_contract!(ContractClassV1Alpha, CasmContractClassV1Alpha);
+
+    unreachable!()
```
Arcticae, line 33: I think we settled on not using the unreachable macro where it can be a panic (i think this case can even be a handleable error since we are extracting it out to a crate)

crates/sierra-casm/src/lib.rs
Arcticae: Perhaps a regular function would be better

crates/sierra-casm/src/lib.rs
```diff
@@ -0,0 +1,63 @@
...
+use cairo_lang_starknet::casm_contract_class::CasmContractClass;
+use cairo_lang_starknet::contract_class::{ContractClass, ContractEntryPoints};
+use cairo_lang_starknet_v1_0_0_alpha6::casm_contract_class::CasmContractClass as CasmContractClassV1Alpha;
+use cairo_lang_starknet_v1_0_0_alpha6::contract_class::ContractClass as ContractClassV1Alpha;
+use cairo_lang_starknet_v1_0_0_rc0::casm_contract_class::CasmContractClass as CasmContractClassV1Rc0;
+use cairo_lang_starknet_v1_0_0_rc0::contract_class::ContractClass as ContractClassV1Rc0;
+use cairo_lang_starknet_v1_1_1::casm_contract_class::CasmContractClass as CasmContractClassV1_1;
+use cairo_lang_starknet_v1_1_1::contract_class::ContractClass as ContractClassV1_1;
+
+pub fn compile(definition: &FlattenedSierraClass) -> anyhow::Result<CasmContractClass> {
```
Arcticae, line 17: I think the interface should be more universal, so that the user does not have to use `starknet::core` in order to compile

crates/sierra-casm/src/lib.rs
Arcticae: Netiher do i 😬

Arcticae (changes requested):

crates/sierra-casm/src/lib.rs
war-in: We do have to convert `CasmContractClass` which we got from the compilation to the latest CasmContractClas. Will add a description and maybe rename this function

crates/cheatnet/src/forking/state.rs
Arcticae, line 1: Changes from this file should be reverted

crates/sierra-casm/src/lib.rs
```diff
@@ -0,0 +1,46 @@
...
+            }
+        }};
+    }
+
+    compile_contract!(ContractClass, CasmContractClass);
+    compile_contract!(ContractClassV1_1, CasmContractClassV1_1);
+    compile_contract!(ContractClassV1Rc0, CasmContractClassV1Rc0);
+    compile_contract!(ContractClassV1Alpha, CasmContractClassV1Alpha);
+
+    Err("Unable to compile Sierra to Casm".to_string())
```
Arcticae, line 35: A reason would be nice here

crates/sierra-casm/src/lib.rs
```diff
@@ -0,0 +1,46 @@
...
+}
+
+/// converts `CasmContractClass` from the old `cairo_lang_starknet` library version
+/// to the `CasmContractClass` from the newest version
+fn old_casm_to_newest_casm<T>(value: &T) -> CasmContractClass
+where
+    T: Serialize + DeserializeOwned,
+{
+    let serialized = serde_json::to_value(value).unwrap();
+    serde_json::from_value::<CasmContractClass>(serialized).unwrap()
```
Arcticae, line 45: We should probably chain those errors

Arcticae, line 45: Shouldn't this be of type `T`?

Arcticae (changes requested):

crates/sierra-casm/src/lib.rs
war-in: I think it is not necessary because Casm should not change. This serialization is only needed because of the types

war-in: Nope, we want to deserialize it to the latest CasmContractClass

crates/sierra-casm/src/lib.rs
war-in: 🔒

crates/sierra-casm/src/lib.rs
Arcticae: Ah okay

Arcticae (approved):

PR #1351, opened 2023-12-12
Parametrized tests design document

### LINKED ISSUE #1307: Design parametric tests support
Like fuzzing but hardcoded

Complex argument values need exploring and designing

Please look up quickcheck (rust fuzzer) or other tools, on how it's done currently

cptartur:
Closes #1307 

## Introduced changes

- Added a design document for parametrized tests

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests

### DIFF
--- a/design_documents/parametrizing_tests_with_fixed_values.md
+++ b/design_documents/parametrizing_tests_with_fixed_values.md
@@ -78,8 +78,8 @@ Provided names must be unique and must not be the same as automatically generate
 For example:
 
 ```cairo
-#[test_case(1, 3, name: 'my_name')]
-#[test_case(3, 5, name: 'a_test')]
+#[test_case(1, 3, name: "my_name")]
+#[test_case(3, 5, name: "a_test")]
 fn my_test(a: felt252, b: u32) {
     // ...
 }
--- a/design_documents/parametrizing_tests_with_fixed_values.md
+++ b/design_documents/parametrizing_tests_with_fixed_values.md
@@ -195,3 +195,11 @@ This is problematic because `snforge` would treat these generated tests as compl
 
 Test collector should be modified, so that cases of the same test as a single bundle.
 This way we can handle execution of them more easily.
+
+## Implementation Plan
+
+1. Prototype creating a plugin for test code generation.
+   Initially, it can support just simple arguments and unnamed test cases.
+2. If/Once code generation approach is proven viable, add support for complex types arguments and introduce necessary
+   test collection logic according to the document.
+3. Named test cases can be added after MVP is created.
--- a/design_documents/parametrizing_tests_with_fixed_values.md
+++ b/design_documents/parametrizing_tests_with_fixed_values.md
@@ -158,6 +158,8 @@ Adding `#[ignore]` attribute to parametrized test should ignore all of it's test
 
 In case `--rerun-failed` flag is used, only failed cases of a parametrized test should be rerun.
 
+Rerun cases should still be displayed together: The behavior should be exactly the same as if just running test cases.
+
 ## Required Changes to Test Collector
 
 To support parametric test, we need to introduce some changes to the test collector in Scarb as outlined below.
--- a/design_documents/parametrizing_tests_with_fixed_values.md
+++ b/design_documents/parametrizing_tests_with_fixed_values.md
@@ -134,7 +134,7 @@ This change is mostly straightforward: It can be done in the same manner as all 
 For each test that has cases containing fixtures, we need to generate a new test that has calls to these fixtures
 inline.
 Additionally, if we follow the alternative approach of handling [parametrized fixtures](#parametrized-fixtures),
-we will also need to generate the neccessary argument definitions.
+we will also need to generate the necessary argument definitions.
 
 ### Argument Injection
 
--- a/design_documents/parametrizing_tests_with_fixed_values.md
+++ b/design_documents/parametrizing_tests_with_fixed_values.md
@@ -132,7 +132,7 @@ This change is mostly straightforward: It can be done in the same manner as all 
 ### Code Generation
 
 If the parameterized test case uses different fixtures, we will have to generate several versions of the test case.
-Additionally, if we follow the alternative approach of handling [parametrized fixtures](#parametrized-fixtures),
+Additionally, if we follow the alternative approach of handling [parameterized fixtures](#parametrized-fixtures),
 we will also need to generate the necessary argument definitions.
 
 ### Argument Injection
--- a/design_documents/parametrizing_tests_with_fixed_values.md
+++ b/design_documents/parametrizing_tests_with_fixed_values.md
@@ -128,7 +128,14 @@ fn my_test_generated(a: felt252) {
 While it is technically possible to use complex types in attributes instead of defining fixtures, handling of them in
 code would be quite problematic based on my limited research.
 
-At least initially, we do not want to support syntax with complex types. For example, this would not be allowed:
+Annotations do not seem to be type-checked at all.
+Arguments passed to annotations are returned as AST node in the compiler code.
+Handling and generating the necessary Cairo code from them is not trivial.
+
+Based on these conclusions, it is logical we limit our support only simple types or calls to functions, without support
+for nested calls, etc.
+
+For example, this syntax would not be allowed:
 
 ```cairo
 #[parametrize]
--- a/design_documents/parametrizing_tests_with_fixed_values.md
+++ b/design_documents/parametrizing_tests_with_fixed_values.md
@@ -89,14 +89,14 @@ An example output could look similarly to this:
 
 ```shell
 $ snforge test
-[PASS] tests::parametrized(a = 1, b = 2, c = my_fixture(1, 2))
-[PASS] tests::parametrized(a = 3, b = 5, c = my_fixture(4, 5))
-[FAIL] tests::parametrized(a = 4, b = 5, c = my_fixture(3, 2))
+[PASS] tests::parametrized(a = 1, b = 2)
+[PASS] tests::parametrized(a = 3, b = 5)
+[FAIL] tests::parametrized(a = 4, b = 5)
 
 Failure data:
     original value: [344693033283], converted to a string: [PANIC]
     
-[PASS] tests::parametrized(a = 5, b = 7, c = my_fixture(7, 8))
+[PASS] tests::parametrized(a = 5, b = 7)
 # ...
 ```
 
@@ -118,8 +118,6 @@ To support parametric test, we need to introduce some changes to the test collec
 This change is mostly straightforward: It can be done in the same manner as all other attributes.
 
 ### Code Generation
-
-If the parameterized test case uses different fixtures, we will have to generate several versions of the test case.
 
 For this approach, my recommendation is to use plugins for code generation.
 We could create a separate plugin for just generating test cases from our attribute and make sure it is executed before
--- a/design_documents/parametrizing_tests_with_fixed_values.md
+++ b/design_documents/parametrizing_tests_with_fixed_values.md
@@ -1,0 +1,106 @@
+# Parametrizing Tests With Fixed Values
+
+## Context
+
+Various test runners for other programming languages provide ways of parametrizing test cases with fixed values.
+This way, multiple test cases can be executed for the same test source code.
+
+## Goal
+
+Propose a way of adding support for parametrized tests to Starknet Foundry's `snforge`.
+
+## Proposed Solution
+
+Introduce three new attributes: `parametrize`, `case` and.
+
+- `parametrize` - defines a parametrized test.
+- `case` - defines a specific test case
+
+### Parametrization With Simple Arguments
+
+```cairo
+#[parametrize]
+#[case(1, 3)]
+#[case(3, 5)]
+#[test]
+fn my_test(a: felt252, b: u32) {
+    // case a = 1, b = 3
+    // case a = 3, b = 5
+    // ...
+}
+```
+
+For test cases like this, values could simply be injected into a runner for each `case`.
+
+### Parametrization With Fixtures
+
+```cairo
+struct MyStruct {
+    a: felt252, 
+    b: Array<felt252>
+}
+
+fn my_fixture() -> MyStruct {
+    MyStruct { a: 1, b: array![2, 3] }
+}
+
+#[parametrize]
+#[case(1, my_fixture())]
+#[case(3, my_fixture())]
+#[test]
+fn my_test(a: felt252, b: MyStruct) {
+    // case a = 1, b = MyStruct { a: 1, b: [2, 3] }
+    // case a = 3, b = MyStruct { a: 1, b: [2, 3] }
+    // ...
+}
+
+// This would require some code generation
+```
+
+To handle parametrization with function calls, we would have to generate a necessary code.
+An example generated code for the snippet above could look like this
+
+```cairo
+// ...
+#[test]
+fn my_test_generated(a: felt252) {
+    let b = my_fixture();
+    // ...
+}
+```
+
+### Parametrized Fixtures
+
+```cairo
+struct MyStruct {
+    a: felt252, 
+    b: Array<felt252>
+}
+
+fn my_fixture(a: felt252, b: felt252, c: felt252) -> MyStruct {
+    MyStruct { a: a, b: array![b, c] }
+}
+
+#[parametrize]
+#[case(1, my_fixture(2, 3, 4))]
+#[case(3, my_fixture(5, 6, 7))]
+#[test]
+fn my_test(a: felt252, b: MyStruct) {
+    // case a = 1, b = MyStruct { a: 2, b: [3, 4] }
+    // case a = 3, b = MyStruct { a: 5, b: [6, 7] }
+    // ...
+}
+
+// This would require some code generation
+```
+
+This could be handled in a very similar manner as the [non-parametrized fixture case](#parametrization-with-fixtures):
+
+```cairo
+// ...
+#[test]
+fn my_test_generated(a: felt252) {
+    let b = my_fixture(2, 3, 4);
+    // ...
+}
+```
--- a/design_documents/parametrizing_tests_with_fixed_values.md
+++ b/design_documents/parametrizing_tests_with_fixed_values.md
@@ -123,6 +123,23 @@ fn my_test_generated(a: felt252) {
 }
 ```
 
+### Parametrizing With Complex Types
+
+While it is technically possible to use complex types in attributes instead of defining fixtures, handling of them in
+code would be quite problematic based on my limited research.
+
+At least initially, we do not want to support syntax with complex types. For example, this would not be allowed:
+
+```cairo
+#[parametrize]
+#[case(MyStruct { a: 1, b: array![2, 3] })]
+#[test]
+fn my_test(a: MyStruct) {
+    // ...
+}
+```
+### Problems With Code Generation
+
 The code generation solution has some problems that will need to be addressed before it is implemented.
 See [section below](#possible-problems-with-code-generation) for details.
 
--- a/design_documents/parametrizing_tests_with_fixed_values.md
+++ b/design_documents/parametrizing_tests_with_fixed_values.md
@@ -140,7 +140,7 @@ we will also need to generate the necessary argument definitions.
 The structure representing a test case (currently named `TestCaseRaw`), should gain additional field `cases` containing
 arguments for each of the parametrized cases.
 
-```cairo
+```rust
 struct TestCaseRaw {
     // ...
     cases: Vec<Vec<String>>
--- a/design_documents/parametrizing_tests_with_fixed_values.md
+++ b/design_documents/parametrizing_tests_with_fixed_values.md
@@ -106,7 +106,7 @@ Parallel test execution implementation in forge makes deterministic test executi
 For good user experience, we should aim to implement parametrized tests in such a way that cases results are displayed
 in deterministic order and grouped together.
 
-Test execution itself can be performed in any order as long as we display the results in grouped manner.
+Test execution itself can be performed in any order as long as we display the results in a grouped manner.
 We can follow the fuzzed tests implementation for this.
 
 ### Behavior of `--exit-first` Flag
--- a/design_documents/parametrizing_tests_with_fixed_values.md
+++ b/design_documents/parametrizing_tests_with_fixed_values.md
@@ -109,6 +109,16 @@ in deterministic order and grouped together.
 Test execution itself can be performed in any order as long as we display the results in grouped manner.
 We can follow the fuzzed tests implementation for this.
 
+### Behavior of `--exit-first` Flag
+
+In case `--exit-first` flag is used we should try to have consistent behavior of parametrized tests.
+
+If some other test fails and there are cases of parametrized tests that did not finish execution,
+we should not display any results of parametrized test.
+
+In case a parametrized test case fails, we should display it.
+Other cases may not be displayed in this case.
+
 ## Required Changes to Test Collector
 
 To support parametric test, we need to introduce some changes to the test collector in Scarb as outlined below.
--- a/design_documents/parametrizing_tests_with_fixed_values.md
+++ b/design_documents/parametrizing_tests_with_fixed_values.md
@@ -104,3 +104,65 @@ fn my_test_generated(a: felt252) {
     // ...
 }
 ```
+
+Alternatively, we could generate code where arguments of `my_fixture` are injected:
+
+```cairo
+// ...
+#[test]
+fn my_test_generated(a: felt252, my_fixture_a: felt252, my_fixture_b: felt252, my_fixture_c: felt252) {
+    let b = my_fixture(my_fixture_a, my_fixture_b, my_fixture_c);
+    // ...
+}
+```
+
+This has the benefit of only needing to generate one version of the test. Necessary arguments just need to be injected.
+
+The code generation solution has some problems that will need to be addressed before it is implemented.
+See [section below](#possible-problems-with-code-generation) for details.
+
+## Required Changes to Test Collector
+
+To support parametric test, we need to introduce some changes to the test collector in Scarb.
+
+### Define the Necessary Attributes
+
+This change is mostly straightforward: It can be done in the same manner as all other attributes.
+
+### Code Generation
+
+For each test that has cases containing fixtures, we need to generate a new test that has calls to these fixtures
+inline.
+Additionally, if we follow the alternative approach of handling [parametrized fixtures](#parametrized-fixtures),
+we will also need to generate the neccessary argument definitions.
+
+### Argument Injection
+
+The structure representing a test case (currently named `TestCaseRaw`), should gain additional field `cases` containing
+arguments for each of the parametrized cases.
+
+```cairo
+struct TestCaseRaw {
+    // ...
+    cases: Vec<Vec<String>>
+}
+```
+
+Where each `Vec<String>` represents arguments for a single case.
+Arguments should be only defined using primitive types.
+
+These arguments could be then handled in `snforge` and injected in the same manner as fuzzer arguments are.
+
+## Possible Problems With Code Generation
+
+In case the parametrized test case uses different fixtures in different cases, we will have to generate multiple
+versions of the test case.
+
+This is problematic because `snforge` would treat these generated tests as completely separate entities.
+
+Some possible solutions for this problem would be:
+
+- Modifying the `TestCaseRaw` so it contains some additional field that allow tests cases to reference each other: This
+  way would know which tests are just different cases of the same sour code.
+- Modify `TestCaseRaw` so it can contain references to multiple "implementations".
+  `snforge` would then find these implementations and run them in a bundle.
--- a/design_documents/parametrizing_tests_with_fixed_values.md
+++ b/design_documents/parametrizing_tests_with_fixed_values.md
@@ -62,6 +62,9 @@ the same base test.
 Test in `snforge` can be filtered by name.
 It is important that generated test names do not break the test filtering logic.
 
+For consistent UX, we should not allow filtering specific test cases.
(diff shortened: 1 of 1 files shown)


design_documents/parametrizing_tests_with_fixed_values.md
```diff
@@ -0,0 +1,106 @@
...
+Various test runners for other programming languages provide ways of parametrizing test cases with fixed values.
+This way, multiple test cases can be executed for the same test source code.
+
+## Goal
+
+Propose a way of adding support for parametrized tests to Starknet Foundry's `snforge`.
+
+## Proposed Solution
+
+Introduce three new attributes: `parametrize`, `case` and.
```
drknzz, line 14: Is there 3rd attribute or?

design_documents/parametrizing_tests_with_fixed_values.md
```diff
@@ -0,0 +1,106 @@
...
+
+- `parametrize` - defines a parametrized test.
+- `case` - defines a specific test case
+
+### Parametrization With Simple Arguments
+
+```cairo
+#[parametrize]
```
drknzz, line 22: Would this be necessary?

drknzz (commented):
Seems pretty straight forward

design_documents/parametrizing_tests_with_fixed_values.md
cptartur: Leftover by accident

design_documents/parametrizing_tests_with_fixed_values.md
cptartur: Probably not, we can remove it. Perhaps having this attribute could simplify some handling in the actual implementation, though it's hard to say.

Arcticae (commented):
Can we also do a case study on what needs to be done on test collector side in order to support this?

design_documents/parametrizing_tests_with_fixed_values.md
```diff
@@ -0,0 +1,168 @@
...
+
+Where each `Vec<String>` represents arguments for a single case.
+Arguments should be only defined using primitive types.
+
+These arguments could be then handled in `snforge` and injected in the same manner as fuzzer arguments are.
+
+## Possible Problems With Code Generation
+
+In case the parametrized test case uses different fixtures in different cases, we will have to generate multiple
+versions of the test case.
```
drknzz, line 159: Please reword this sentence 🙏

Arcticae (approved):

design_documents/parametrizing_tests_with_fixed_values.md
war-in: how about 

```
#[parametrize(
    case(...),
    case(...),
]
```

would that be possible?

design_documents/parametrizing_tests_with_fixed_values.md
```diff
@@ -0,0 +1,168 @@
...
+    b: Array<felt252>
+}
+
+fn my_fixture() -> MyStruct {
+    MyStruct { a: 1, b: array![2, 3] }
+}
+
+#[parametrize]
+#[case(1, my_fixture())]
+#[case(3, my_fixture())]
```
war-in, line 49: Would it be possible to create a struct directly in the `case`?

design_documents/parametrizing_tests_with_fixed_values.md
```diff
@@ -0,0 +1,168 @@
...
+An example generated code for the snippet above could look like this
+
+```cairo
+// ...
+#[test]
+fn my_test_generated(a: felt252) {
+    let b = my_fixture();
+    // ...
+}
+```
```
war-in, line 70: So if I have this test
```
#[case(2, 1, my_fixture(), 7]
fn my_test(a: u8, b: u8, c: MyStruct, d: u8)
```

my generated code would look like
```
fn my_test_generated(a: u8, b: u8, d: u8) {
    let c = my_fixture();
```
right?

war-in (commented):
The interface seems nice, but I'm afraid it would require many lines of spaghetti code

design_documents/parametrizing_tests_with_fixed_values.md
cptartur: Syntax like 

```cairo
#[case(MyStruct { a: 1, b: array![2, 3] })]`
```

should be valid in cairo. However from what I understand, attributes don't get type-checked, and we'd have to manually handle this struct creation.

I wonder now if this could be just handled with code-gen as well.

design_documents/parametrizing_tests_with_fixed_values.md
cptartur: Correct. There's some ambiguity in how "removing" arguments from function signature would be handled, so details of this may vary. But that was my idea.

war-in: Yeah, that was my point. It seems complicated

cptartur: Some complication could maybe be removed if we decide to rely entirely on code generation: Each case would generate a new cairo code with all arguments inlined. 

It would cost us some extra code, but I'm assuming users would most often define only a couple of cases.

The problem with this approach would be ensuring some determinism in how these cases are executed: I'd like parametrized tests cases to run in order they were defined, so you always get deterministic failures from them.

cptartur: Updated the code so we only rely on code generation.

design_documents/parametrizing_tests_with_fixed_values.md
```diff
@@ -0,0 +1,192 @@
...
+}
+
+#[test]
+fn my_test_case_1() {
+  let a: felt252 = 3;
+  let b: u32 = 5;
+}
+```
+
+The explicit type annotations for each argument are necessary.
```
MaksymilianDemitraszek, line 53: I think those are necessary by language semantics, nothing this doc-specific

design_documents/parametrizing_tests_with_fixed_values.md
MaksymilianDemitraszek: Imo this is too verbose. We should drop `test` and `parametrize`

`#[test_case(1,2,)]` is sufficient an consistent with rust. 

The main goal of this syntax is to write succinct tests, so it is better to keep the syntax succinct.

design_documents/parametrizing_tests_with_fixed_values.md
```diff
@@ -0,0 +1,192 @@
...
+
+```cairo
+struct MyStruct {
+    a: felt252, 
+    b: Array<felt252>
+}
+
+fn my_fixture() -> MyStruct {
+    MyStruct { a: 1, b: array![2, 3] }
+}
```
MaksymilianDemitraszek, line 66: Missing details on:
- more advanced use-cases and how this solution plays along with them. What are common use cases, how this solution handles them? (imo not very well)
- I have to write a function for every case even if it is fixed, this is a load of boilderplate and kinda kills the purpose, it is easier to just write a normal test.
- How this will be type checked? What are the constraints of fixture functions?

design_documents/parametrizing_tests_with_fixed_values.md
```diff
@@ -0,0 +1,192 @@
...
+    let a = 1;
+    let b = my_fixture(2, 3, 4);
+    // ...
+}
+```
+
+### Parametrizing With Complex Types
+
+While it is technically possible to use complex types in attributes instead of defining fixtures, handling of them in
+code would be quite problematic based on my limited research.
```
MaksymilianDemitraszek, line 129: I don't understand why this would be more problematic than fixture composing, constructors are not much different from functions. 
Imo this syntax makes more sense and we should focus on.

MaksymilianDemitraszek: We could look into using cairo parser to parse this syntax. Imo that should be a goal of this design doc.

design_documents/parametrizing_tests_with_fixed_values.md
```diff
@@ -0,0 +1,192 @@
...
+Failure data:
+    original value: [344693033283], converted to a string: [PANIC]
+    
+[PASS] tests::parametrized(a = 5, b = 7, c = my_fixture(7, 8))
+# ...
+```
+
+## Required Changes to Test Collector
+
+To support parametric test, we need to introduce some changes to the test collector in Scarb.
```
MaksymilianDemitraszek, line 169: What changes?

design_documents/parametrizing_tests_with_fixed_values.md
MaksymilianDemitraszek: Imo generally you underestimate the complexity of this solution, if you have provided more details you would run into more problems. It might just look plausible.

MaksymilianDemitraszek (changes requested):

design_documents/parametrizing_tests_with_fixed_values.md
cptartur: I think language semantics are quite important in the case of this document.

design_documents/parametrizing_tests_with_fixed_values.md
cptartur: From my limited research, annotations aren't really type-checked, at least not by default. You get an reference to a node in AST with whatever you have in there and have to handle it. I don't think it is easy or obvious to handle complex structures in this so that's why I'm not proposing a solution like this.

Only allowing simple values or calls to functions (no support for nested calls etc.) seems easy enough to implement however.

design_documents/parametrizing_tests_with_fixed_values.md
cptartur: There are subheaders below this outlining the changes

design_documents/parametrizing_tests_with_fixed_values.md
cptartur: Can be done I think

design_documents/parametrizing_tests_with_fixed_values.md
cptartur: > more advanced use-cases and how this solution plays along with them. What are common use cases, how this solution handles them? (imo not very well)

What do you mean by more advanced use cases? 

> I have to write a function for every case even if it is fixed, this is a load of boilderplate and kinda kills the purpose, it is easier to just write a normal test.

If your function takes simple values as arguments then you don't need a fixture. Requiring them for complex types is mostly so it's easier to implement. 

> How this will be type checked? What are the constraints of fixture functions?

I don't think annotations are type checked at all. We'd have to have our own handling (can be discussed in this document).

design_documents/parametrizing_tests_with_fixed_values.md
MaksymilianDemitraszek: @cptartur then you should add a section as well that functions should have names and are prefixed with `fn` keyword and functions can be called from other functions etc. You write a document under specific assumptions and don't mix existing mechanisms with your design otherwise it is hard to read.

design_documents/parametrizing_tests_with_fixed_values.md
MaksymilianDemitraszek: > What do you mean by more advanced use cases?

Real life use-cases, example of a test which could exists/exist in some real codebase and its boilerplate would be reduced

> We'd have to have our own handling (can be discussed in this document).

If you propose something like this you should discuss them, I assume those are not type checked by the compiler.

design_documents/parametrizing_tests_with_fixed_values.md
MaksymilianDemitraszek: Sorry, hard to read in without markdown formatting 😄

design_documents/parametrizing_tests_with_fixed_values.md
MaksymilianDemitraszek: > Only allowing simple values or calls to functions (no support for nested calls etc.) seems easy enough to implement however.

I don't think this is based on anything other than your intuition. Please explain in the document what you have tried in regards of structs and where you ran into problems. Explain how this checking will work, what exact restrictions will be in place.

cptartur: I've added an explanation to the document, though I'm not sure if we should be including things like that.

design_documents/parametrizing_tests_with_fixed_values.md
cptartur: The point of this section is to ensure that generated code has this type annotations included, because it's important for its behavior.

I can remove this section if you really think it's not necessary, but it seems at list a bit useful.

design_documents/parametrizing_tests_with_fixed_values.md
```diff
@@ -128,7 +128,14 @@ fn my_test_generated(a: felt252) {
 While it is technically possible to use complex types in attributes instead of defining fixtures, handling of them in
 code would be quite problematic based on my limited research.
 
-At least initially, we do not want to support syntax with complex types. For example, this would not be allowed:
+Annotations do not seem to be type-checked at all.
```
MaksymilianDemitraszek, line 131: Why this is the problem, i don't understand. Can't we typecheck them?

piotmag769: We can I think, requires changing the test collector and digging into the compiler internals, but should be possible

design_documents/parametrizing_tests_with_fixed_values.md
```diff
@@ -128,7 +128,14 @@ fn my_test_generated(a: felt252) {
...
+Arguments passed to annotations are returned as AST node in the compiler code.
+Handling and generating the necessary Cairo code from them is not trivial.
+
+Based on these conclusions, it is logical we limit our support only simple types or calls to functions, without support
```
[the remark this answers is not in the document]
piotmag769: I think the only valid research here will be trying to implement it, otherwise it is still an intuition based thing

design_documents/parametrizing_tests_with_fixed_values.md
```diff
@@ -0,0 +1,199 @@
...
+fn my_test_case_1() {
+  let a: felt252 = 3;
+  let b: u32 = 1;
+}
+
+#[test]
+fn my_test_case_1() {
+  let a: felt252 = 3;
+  let b: u32 = 5;
+}
```
drknzz, line 50:
```suggestion
#[test]
fn my_test_case_1() {
  let a: felt252 = 1;
  let b: u32 = 3;
}

#[test]
fn my_test_case_2() {
  let a: felt252 = 3;
  let b: u32 = 5;
}
```

Probably should be like this, unless I don't understand the mechanism

design_documents/parametrizing_tests_with_fixed_values.md
piotmag769: @MaksymilianDemitraszek +10

design_documents/parametrizing_tests_with_fixed_values.md
piotmag769: Agree with Maks, this section is just describing Cairo spec

design_documents/parametrizing_tests_with_fixed_values.md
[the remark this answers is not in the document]
MaksymilianDemitraszek: > will be trying to implement it

I think it should be done then, there is huge difference it time commitment between doing som Poc for local experiment and implementing something properly.

design_documents/parametrizing_tests_with_fixed_values.md
```diff
@@ -128,7 +128,14 @@ fn my_test_generated(a: felt252) {
 While it is technically possible to use complex types in attributes instead of defining fixtures, handling of them in
 code would be quite problematic based on my limited research.
 
-At least initially, we do not want to support syntax with complex types. For example, this would not be allowed:
+Annotations do not seem to be type-checked at all.
+Arguments passed to annotations are returned as AST node in the compiler code.
+Handling and generating the necessary Cairo code from them is not trivial.
```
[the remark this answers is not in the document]
cptartur: I tried. I'm not saying it's not possible, most likely it is doable. But I think it's not obvious.

design_documents/parametrizing_tests_with_fixed_values.md
piotmag769: I think such simple fixtures don't provide much benefit and cloud the scope of this task, I think it is another feature that could be considered, but I wouldn't do it there. 
Fixtures start to be beneficial if they are supported on wider range like in pytest imo

piotmag769 (commented):
.

MaksymilianDemitraszek:
I would explore such direction
```
#[test_case(Type {})]
fn(x: Type) -> {

}
```

resolves to 

```
fn() -> {
    let x: Type = Type {};
}
```

It should trivialy cover all cases.

design_documents/parametrizing_tests_with_fixed_values.md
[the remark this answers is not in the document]
cptartur: Not relevant anymore

design_documents/parametrizing_tests_with_fixed_values.md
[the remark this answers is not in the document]
cptartur: Not relevant anymore

design_documents/parametrizing_tests_with_fixed_values.md
cptartur: Not relevant anymore

design_documents/parametrizing_tests_with_fixed_values.md
cptartur: Not relevant anymore

design_documents/parametrizing_tests_with_fixed_values.md
cptartur: Not relevant anymore

design_documents/parametrizing_tests_with_fixed_values.md
```diff
@@ -0,0 +1,143 @@
...
+Generate code for each of the test cases so no handling of values injection, etc. is necessary.
+The compiler would then validate if the values for the generated code are actually valid.
+Thanks to that, we could have failures for invalid arguments early in the execution.
+
+### Parametrization With Arguments
+
+```cairo
+#[test_case(1, 3)]
+#[test_case(3, 5)]
+fn my_test(a: felt252, b: u32) {
```
war-in, line 25: Rusts' `test_case` also has a `name` parameter for each case. I think it would be good to have it here because errors would be more descriptive (the user would know which case failed)

design_documents/parametrizing_tests_with_fixed_values.md
```diff
@@ -0,0 +1,143 @@
...
+## User Experience
+
+Unlike fuzz-tests, we should run and indicate the result of all parametrized cases.
+If some test cases fail, others should still be executed and their results should be displayed to the user.
+
+An example output could look similarly to this:
+
+```shell
+$ snforge test
+[PASS] tests::parametrized(a = 1, b = 2, c = my_fixture(1, 2))
```
war-in, line 92: just noting that there is no information about fixtures up to this moment. Idk if it is ok or not

design_documents/parametrizing_tests_with_fixed_values.md
```diff
@@ -0,0 +1,143 @@
...
+[PASS] tests::parametrized(a = 3, b = 5, c = my_fixture(4, 5))
+[FAIL] tests::parametrized(a = 4, b = 5, c = my_fixture(3, 2))
```
war-in, line 94: If cases had a name we could print them like that
```
[PASS] tests::some_parametrized_test::some_case_name
[PASS] tests::some_parametrized_test::some_other_case_name
```

design_documents/parametrizing_tests_with_fixed_values.md
```diff
@@ -0,0 +1,143 @@
...
+```
+
+## Deterministic Test Execution Order
+
+Parallel test execution implementation in forge makes deterministic test execution order non-trivial.
+For good user experience, we should aim to implement parametrized tests in such a way that cases are executed in
+deterministic order directly after each other.
+
+This is not a technical requirement for implementing parametrized tests but will make for a much better UX, especially
+in case of displaying failures of some parametrized cases.
```
war-in, line 110: I think the problem here is not in the order but in output collection.
Ideally, we could be able to execute cases in parallel, collect the output and print it after all cases are finished

design_documents/parametrizing_tests_with_fixed_values.md
cptartur: It'd have to be investigated in code, but we could support some variant of these:

1. `#[test_case(1, 2)]` and `#[test_case('my_name', 1, 2)]`
2. `#[test_case(1, 2)]` and `#[test_case('my_name', (1, 2))]`
3. `#[test_case(1, 2)]`and `#[test_case(case_name: 'my_name', 1, 2)]`
4. `#[test_case(1, 2)]`and `#[test_case(case_name: 'my_name', (1, 2))]`
5. `#[test_case(1, 2)]`and `#[test_case(case_name: 'my_name', arguments: (1, 2))]`

I think option 3. or 4. seem clearest and easiest to handle. What do you think

design_documents/parametrizing_tests_with_fixed_values.md
cptartur: Leftover from previous version, removed.

design_documents/parametrizing_tests_with_fixed_values.md
cptartur: Correct, they could probably run in any order as long as we output them in deterministic manner. Handling of this would not be obvious in case `--exit-first` flag is used however.

design_documents/parametrizing_tests_with_fixed_values.md
war-in: I like the way the `test_case` crate handles it
`#[test_case(1, 2; 'my_name')]`

Arguments are the most important ones here and we do not provide unnecessary code

design_documents/parametrizing_tests_with_fixed_values.md
war-in: Probably, but we should investigate it because we want those tests to be fast

design_documents/parametrizing_tests_with_fixed_values.md
cptartur: I'm not sure it can be exactly replicated in cairo, but if we used `,` to separate all arguments it could be done I think. So `#[test_case(1, 2, 'my_name')]`

design_documents/parametrizing_tests_with_fixed_values.md
cptartur: I'll update this section so it explains we only want results displayed in deterministic order.

piotmag769: I think if we enforce naming the tests in the attribute (as you described above) we can just treat each case as a separate test and everything will be fine

design_documents/parametrizing_tests_with_fixed_values.md
war-in: That will be misleading and could be taken as a parameter. Can't we do anything in the attributes?

cptartur: Tried using `;` and it doesn't seem to be allowed, though it may require further investigation.

design_documents/parametrizing_tests_with_fixed_values.md
cptartur: I don't want to enforce naming things, this is inflexible. And for me parametric tests only make sense if there are treated as a single "bundle" in output.

piotmag769: `test_case` crate treats them as a separate entities, example output when testing https://github.com/foundry-rs/starknet-foundry/blob/93b2a2a976073411d3ec694dcc330b67f1b8ab3b/crates/cheatnet/tests/starknet/cheat_fork.rs#L135-L137
```console
running 2 tests
test starknet::cheat_fork::warp_cairo0_contract::when_common_call ... ok
test starknet::cheat_fork::warp_cairo0_contract::when_library_call ... ok
```
I think we could stick to that, I don't see any major UX benefit of the [pytest-like approach](https://docs.pytest.org/en/7.1.x/how-to/parametrize.html#pytest-mark-parametrize-parametrizing-test-functions), I guess it's a matter of taste

cptartur: It's a matter of preference, but for me only argument for not bundling tests together is easier implementation. From the UX side there are no benefits.

piotmag769: I agree, but if it is a matter of preference then easier implementation becomes an argument I guess (even though it really shouldn't, but since there are not other arguments...)

design_documents/parametrizing_tests_with_fixed_values.md
```diff
@@ -0,0 +1,151 @@
...
+### Generated Cases Names
+
+Generated test cases should be named in such a way that we can automatically detect which tests are generated cases of
+the same base test.
+
+Test in `snforge` can be filtered by name.
+It is important that generated test names do not break the test filtering logic.
+
+This can be resolved by either using names for generated tests that still work with filters as the base test would do,
+or changing the filtering logic, so it can recognize generated test cases and treat them accordingly.
```
piotmag769, line 57: Enforcing a name would solve this issue as well

design_documents/parametrizing_tests_with_fixed_values.md
```diff
@@ -0,0 +1,151 @@
...
+
+Cairo repository contains multiple examples of plugins that generate Cairo code which we could follow:
+
+- https://github.com/starkware-libs/cairo/blob/main/crates/cairo-lang-plugins/src/plugins/generate_trait.rs
+- https://github.com/starkware-libs/cairo/blob/main/crates/cairo-lang-plugins/src/plugins/panicable.rs
+- https://github.com/starkware-libs/cairo/blob/main/crates/cairo-lang-semantic/src/inline_macros/array.rs
+
+Details of this are for the implementer to investigate, but my recommendation would be to not attempt any approaches
+where we try to modify the test code in the compiler itself but generate the necessary test cases.
+
```
Arcticae, line 144: Can we have at the end some takeaways and a stage-based plan for implementation/mvp?

piotmag769: Wait, actually we want to generate the code in the compiler and then inject it into the db, right? Isn't it what the plugins do? I may be wrong tho

cptartur: What I meant by this is we want to just generate cairo code and inject it. Avoid any kind of magic where we try to modify already parsed code (if that's even possible).

design_documents/parametrizing_tests_with_fixed_values.md
cptartur: Enforcing names is quite cumbersome from the UX standpoint

design_documents/parametrizing_tests_with_fixed_values.md
piotmag769: Could we make it more clear then? I understood it as the opposite

design_documents/parametrizing_tests_with_fixed_values.md
piotmag769: True, we can make a default name if there is no name provided

drknzz (commented):
How would attributes like `#[ignore]` work with the cases? Would it be possible to ignore just a single test case or only the whole suite?

Also let's mark that `--rerun-failed` should rerun only the failed cases.

piotmag769:
@drknzz `#[ignore]` should ignore the all the cases imo (`test_case` works this way too). Agree with `--rerun-failed`, but this should be easy imo

drknzz (commented):
Also, probably an extreme use case, how would fuzzing / parametrizing interact? If at all?

piotmag769:
Fuzzing shouldn't be allowed to be mixed with parametrized test cases imo

design_documents/parametrizing_tests_with_fixed_values.md
cptartur: From code generation standpoint, both versions are pretty much te same. It's just a case of collecting and displaying them together in some logical manner.

design_documents/parametrizing_tests_with_fixed_values.md
```diff
@@ -0,0 +1,205 @@
...
+fn my_test(a: felt252, b: u32) {
+    // ...
+}
+```
+
+For test cases like this, we would generate two actual test cases:
+
+```cairo
+#[test]
+fn my_test_case_1() {
```
drknzz, line 43: How will named test cases be generated? They shouldn't break unnamed test cases, if they are named "1" for example

drknzz (approved):

drknzz (commented):
Should the test cases support `max_steps` argument, after #1384 is merged?

design_documents/parametrizing_tests_with_fixed_values.md
war-in: maybe
`#[test_case::some_name(...)]` would be possible? Wdyt?

Arcticae:
> Should the test cases support `max_steps` ...

I think it should automatically apply for all the cases

design_documents/parametrizing_tests_with_fixed_values.md
Arcticae: I think it's an implementation detail

Arcticae (approved):
I think everything beyond that is excessive research and we should get to implementation instead.

cptartur:
> > Should the test cases support `max_steps` ...
> 
> I think it should automatically apply for all the cases

In general, attributes should apply to all cases of a parametrized test.

design_documents/parametrizing_tests_with_fixed_values.md
cptartur: I've considered adding some kind of a prefix to the test name to mark it's user generated, but that's for an implementer to handle.

MaksymilianDemitraszek (approved):

PR #1354, opened 2023-12-12
Test benchmarks in CI

cptartur:
Closes #

## Introduced changes

- Added a CI step that runs benchmarks using `cargo run --benches` command. This will run each benchmark only once, verifying that it passes, without time measurements etc.

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests

### DIFF
--- a/.github/workflows/check_benchmarks.yml
+++ b/.github/workflows/check_benchmarks.yml
@@ -1,0 +1,21 @@
+name: Check benchmarks
+
+on:
+  push:
+    branches:
+      - master
+  workflow_dispatch:
+
+jobs:
+  benchmarks:
+    name: Verify benchmarks are up to date
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@439cf607258077187679211f12aa6f19af4a0af7
+        with:
+          toolchain: stable
+      - uses: Swatinem/rust-cache@3cf7f8cc28d1b4e7d01e3783be10a97d55d483c8
+      - uses: software-mansion/setup-scarb@v1.3.2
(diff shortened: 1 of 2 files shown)


Arcticae (commented):
@cptartur does this have to be run on every commit? Wouldn't like a on-merge to main work better in this case, since this is not something that we necessarily need each time for the app to be releasable?

Arcticae:
> Though if we don't add this step as required to the master branch protection I think merging PRs with benchmarks broken would still be possible. And we'll immediately know that a PR broke the benchmarks.

Yes, but we'll be taking up 6m of the time possibly blocking the merging while it's not necessary to block the PR because of that. And we'll still know after merging which one broke it

cptartur:
> > Though if we don't add this step as required to the master branch protection I think merging PRs with benchmarks broken would still be possible. And we'll immediately know that a PR broke the benchmarks.
> 
> Yes, but we'll be taking up 6m of the time possibly blocking the merging while it's not necessary to block the PR because of that. And we'll still know after merging which one broke it



> > Though if we don't add this step as required to the master branch protection I think merging PRs with benchmarks broken would still be possible. And we'll immediately know that a PR broke the benchmarks.
> 
> Yes, but we'll be taking up 6m of the time possibly blocking the merging while it's not necessary to block the PR because of that. And we'll still know after merging which one broke it

Updated. Benchmarks tests will now only trigger on push (to master since it's the only push type that causes CI workflow to trigger).

Arcticae (changes requested):
@cptartur i know this is easier but i would just run it with a different workflow, i don't need to know that it was skipped each time under a PR...

Arcticae (approved):

PR #1363, opened 2023-12-13
Switch to universal sierra compilation

### LINKED ISSUE #1175: Incompatible ABIs cause the `contract_class` code parsing to crash

Forge

### Foundry Version

0.10.2

MacOS

x86

### What happened

We don't need the ABIs to be parsed for the code to execute, so we should consider omitting the parsing overall (or parse selectively), since the following example is crashing the newer parser (no backwards compatibility).

### Trace

Reproduction:

```
#[derive(Serde)]
struct Propdetails {
    payload: felt252,
}

#[starknet::interface]
trait IGovernance<State> {
    fn get_proposal_details(self: @State, param: felt252) -> Propdetails;
}

#[test]
#[fork(url: "https://rpc.starknet-testnet.lava.build", block_id: BlockId::Number(904597))]
fn test_forking_functionality() {
    let gov_contract_addr: ContractAddress = 0x7ba1d4836a1142c09dde23cb39b2885fe350912591461b5764454a255bdbac6.try_into().unwrap();
    let dispatcher = IGovernanceDispatcher { contract_address: gov_contract_addr };
    let propdetails = dispatcher.get_proposal_details(1);
    assert(propdetails.payload==0x78b4ccacdc1c902281f6f13d94b6d17b1f4c44ff811c01dea504d43a264f611, 'payload not match');
}

```
Error:
```
thread 'tokio-runtime-worker' panicked at crates/cheatnet/src/forking/state.rs:213:90:
called `Result::unwrap()` on an `Err` value: Error("missing field `kind`", line: 1, column: 2356)
[ERROR] task 10 panicked
```

- [X] I have searched the existing issues and verified no issue exits for this problem.

### LINKED ISSUE #1212: Switch forking sierra compilation to a dedicated package
Blocked by #1211

war-in:
Closes #1212
Closes #1175

## Introduced changes

- `sierra_casm::compile` is used to compile Cairo 1 contracts while forking
- `sierra_casm::compile` is used to compile sierra when it is not present in scarb-api
- abi is unnecessary so it was omitted

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/cheatnet/src/forking/state.rs
```diff
@@ -201,21 +199,14 @@ impl StateReader for ForkStateReader {
...
-                    entry_points_by_type: converted_entry_points,
-                    abi: Some(converted_abi),
-                };
-                let casm_contract_class: CasmContractClass =
-                    CasmContractClass::from_contract_class(sierra_contract_class, false).unwrap();
+                let sierra_contract_class = serde_json::json!({
+                    "sierra_program": converted_sierra_program,
+                    "contract_class_version": "",
+                    "entry_points_by_type": flattened_class.entry_points_by_type
+                });
```
piotmag769, line 207: I think it would be more readable if we created the `ContractClass` struct and then deserialized it to json, now it looks a bit like magic. If field names change it will be hard to find the issue

piotmag769 (approved):
Do we plan to migrate contract sierra -> casm compilation too? Do we have a task for that?

crates/cheatnet/src/forking/state.rs
war-in: Yeah, but here's why it is handled like that.
We do not know which Sierra version it is so we would have to perform some weird conversion

piotmag769: Okay, I trust you

Arcticae (commented):
Can we uncomment the test which was commented before to see that the fix indeed works?

war-in:
> Can we uncomment the test which was commented before to see that the fix indeed works?

Which one was commented? I can't remember 😅

war-in:
> Do we plan to migrate contract sierra -> casm compilation too? Do we have a task for that?

Migrated 🫡

crates/sncast/src/starknet_commands/declare.rs
```diff
@@ -75,8 +77,14 @@ pub async fn declare(
 
     let contract_definition: SierraClass = serde_json::from_str(&contract_artifacts.sierra)
         .context("Failed to parse sierra artifact")?;
-    let casm_contract_definition: CompiledClass =
-        serde_json::from_str(&contract_artifacts.casm).context("Failed to parse casm artifact")?;
+    let casm_contract_definition: CompiledClass = if contract_artifacts.casm.is_empty() {
```
THenry14, line 80: That is cool! do you mind giving me some more context on why we should do this/why we can do it now and we couldn't earlier?

crates/sncast/src/starknet_commands/declare.rs
```diff
@@ -75,8 +77,14 @@ pub async fn declare(
...
+        let sierra: Value = serde_json::from_str(&contract_artifacts.sierra).unwrap();
+        let casm = serde_json::to_string(&compile(sierra).unwrap()).unwrap();
```
THenry14, line 82: Let's maybe avoid unwraping, and try to handle possible error(s) with expect/context ?

piotmag769 (commented):
LGTM. Can we remove `casm = true` from the packages in forge e2e tests? This way we test our compilation even further. I would leave it in one of them to make sure we can read the casm generated by Scarb. WDYT?

crates/sncast/src/starknet_commands/declare.rs
piotmag769: Before we relied on Scarb to do the compilation to casm, now we have a package (will be a binary in a near future) to do sierra->casm compilation

crates/sncast/src/starknet_commands/declare.rs
piotmag769, line 82: I think this should be abstracted from cast too (same as in the forge case)

crates/sncast/src/starknet_commands/declare.rs
THenry14: Yeah, we discussed with Marcin offline, but thanks!

crates/forge/tests/integration/setup_fork.rs
```diff
@@ -272,3 +273,33 @@ fn fork_get_block_info_fails() {
         "Unable to get block with tx hashes from fork"
     );
 }
+
+#[test]
+// found in: https://github.com/foundry-rs/starknet-foundry/issues/1175
+fn problematic_contract_fork() {
```
piotmag769, line 279: `fn incompatible_abi` maybe?

piotmag769: Also do we have a test against a fork with stored contract code in another sierra version? To test the `get_compiled_contract_class` from `ForkStateReader` (it was the first reason that caused us to switch to universal sierra compilation, right?)

war-in: This one is in the 0.1.0

piotmag769 (approved):

Arcticae (approved):

Issue #1102, opened 2023-11-10
Always use block hash or block number in fork cache


cptartur:
Currently, when using forking, separate cache entries will be created for the same block if it's identified by both block hash and number in different tests. We could always use one of these values as cache keys so we don't duplicate these entries.

Issue #901, opened 2023-10-17
Investigate how to remove `spawn_blocking` from forge code


cptartur:

piotmag769:
https://stackoverflow.com/a/62536772
In short: there is no way to spawn a runtime from a runtime, tokio seems to check that. That's why we have to use spawn blocking which uses another thread pool. However we can try to use `Handle` to avoid creating additional runtimes
https://docs.rs/tokio/latest/tokio/runtime/struct.Handle.html#method.spawn

MaksymilianDemitraszek:
We could consider either implementing some sophisticated strategy of collecting the required data and then batching requests together. Though in my opinion this is not worth the effort atm. Otherwise we would need an async cairo vm which is not feasible. 

@Draggu @Arcticae  closing it for now as I don't think it will worth the effort.

PR #1383, opened 2023-12-14
Use consistent latest block number across whole run, use block numbers instead of block hashes

### LINKED ISSUE #1049: Use consistent latest block number across whole test suite
I think this is the kind of information we should have before even running the test, so:
1. It's consistent all across the board (whole suite)
2. We don't have to return it from the test itself

Can we please address this?

_Originally posted by @Arcticae in https://github.com/foundry-rs/starknet-foundry/pull/1007#discussion_r1383032262_

piotmag769:
## Introduced changes

- Use consistent latest block number for each url across whole run - closes #1049 
- Use block numbers instead of hashes to enable reusing cache (same block can be referenced by both hash and number, but the cache from number cannot be reused for hash and the other way round) - partially solves #1102 (will open another PR to enforce proper types etc.)
- Latest block numbers for each url are now printed at the end (if `BlockId::Tag(Latest)` was specified for such url)

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/cheatnet/tests/contracts/Scarb.lock
```diff
@@ -1,6 +0,0 @@
-# Code generated by scarb DO NOT EDIT.
-version = 1
-
-[[package]]
-name = "cheatnet_testing_contracts"
-version = "0.1.0"
```
piotmag769, line 6: It got removed when clearing git cache and readding everything, shouldn't have been there in the first place

crates/forge/src/block_number_map.rs
```diff
@@ -0,0 +1,118 @@
...
+    pub fn get_url_to_latest_block_number(&self) -> &HashMap<String, BlockNumber> {
+        &self.url_to_latest_block_number
+    }
+}
+
+async fn get_latest_block_number(url: &Url) -> Result<BlockNumber> {
+    let client = JsonRpcClient::new(HttpTransport::new(url.clone()));
+
+    match Handle::current()
+        .spawn(async move { client.get_block_with_tx_hashes(BlockId::Tag(Latest)).await })
```
piotmag769, line 60: Btw this is the solution that should be used to solve https://github.com/foundry-rs/starknet-foundry/issues/901
cc @cptartur @tkumor3 @Arcticae @MaksymilianDemitraszek

crates/forge/src/block_number_map.rs
```diff
@@ -0,0 +1,118 @@
...
+            assert_eq!(fork_params_string.block_id_value, "Latest");
+            if let Some(block_number) = block_number_map.get_latest_block_number(&url_str) {
+                *block_number
+            } else {
+                let latest_block_number = get_latest_block_number(&url).await?;
+                block_number_map.add_latest_block_number(url_str, latest_block_number);
+                latest_block_number
+            }
+        }
+        _ => unreachable!(),
```
Arcticae, line 115: Can we program this so it does not use `unreachable!` macro?

crates/forge/src/block_number_map.rs
```diff
@@ -0,0 +1,118 @@
...
+        }
+        "tag" => {
+            assert_eq!(fork_params_string.block_id_value, "Latest");
+            if let Some(block_number) = block_number_map.get_latest_block_number(&url_str) {
+                *block_number
+            } else {
+                let latest_block_number = get_latest_block_number(&url).await?;
+                block_number_map.add_latest_block_number(url_str, latest_block_number);
+                latest_block_number
+            }
```
Arcticae, line 113: Shouldn't fetching this be responsibility of the class that caches the results? I find it confusing to have a standalone function do this

crates/forge/src/block_number_map.rs
piotmag769: I can throw a panic here or add an error message, but idk why it changes anything. Those are verified in scarb test collector or when parsing config from Scarb.toml

crates/forge/src/block_number_map.rs
piotmag769: Can do that

Arcticae (approved):

PR #1384, opened 2023-12-14
add max steps

tkumor3:
Closes #

## Introduced changes

-

## Checklist

- [x] Linked relevant issue
- [ ] Added relevant tests


crates/cheatnet/src/constants.rs
```diff
@@ -46,7 +46,7 @@ pub const TEST_CONTRACT_CLASS_HASH: &str = "0x117";
 pub const STEP_RESOURCE_COST: f64 = 0.01_f64;
 // snforge_std/src/cheatcodes.cairo::test_address
 pub const TEST_ADDRESS: &str = "0x01724987234973219347210837402";
-
+pub const DEFAULT_MAX_TX_STEPS: u32 = 3_000_000;
```
Arcticae, line 49: What's the justification of this value?

Arcticae (changes requested):
There's no e2e tests which should be ran here and docs

tkumor3:
I need the new scarb version to write e2e tests. I asked Maciek to build nightly. Docs will be created in a separate PR. I assume that we want to inform people about it when the scarb 2.5.0 is released.

crates/cheatnet/src/constants.rs
tkumor3: https://github.com/orgs/foundry-rs/projects/3/views/4?pane=issue&itemId=40320964

crates/forge/src/main.rs
```diff
@@ -79,6 +79,9 @@ struct TestArgs {
     /// Seed for the fuzzer
     #[arg(short = 's', long)]
     fuzzer_seed: Option<u64>,
+    ///Max steps
```
piotmag769, line 82: Space

crates/forge/src/main.rs
```diff
@@ -79,6 +79,9 @@ struct TestArgs {
...
+    #[arg(short = 'm', long)]
+    max_steps: Option<u32>,
```
piotmag769, line 84: I wonder if we want to have it configurable from the cli. Tbh I would even remove fuzzer config from the cli in the future, it's weird to have tests changing their internal behaviour depending on what cli args were used
cc @Arcticae

piotmag769:
Tbh I think we can add docs in this PR and just add a temporary warning that the attribute in the cairo code needs scarb nightly-xyz to work

crates/cheatnet/src/constants.rs
MaksymilianDemitraszek: Could you link the docs here?

.tool-versions
```diff
@@ -1 +1 @@
-scarb 2.4.0
+scarb nightly-2023-12-18
```
MaksymilianDemitraszek, line 1: I assume we merge this one when 2.5.0 is released?

MaksymilianDemitraszek: If not, we shouldn't merge this change to `.tool-versions`

crates/forge-runner/src/running.rs
```diff
@@ -286,6 +288,12 @@ pub fn run_test_case(
     let block_context = get_context(&forge_runtime).block_context.clone();
     let execution_resources = get_all_execution_resources(forge_runtime);
 
+    let max_steps = if case.max_steps.is_some() {
+        case.max_steps
+    } else {
+        runner_config.max_steps
+    };
```
MaksymilianDemitraszek, line 295: Please use match, it is safer beasue it checks for exhaustiveness.

crates/forge-runner/src/running.rs
```diff
@@ -286,6 +288,12 @@ pub fn run_test_case(
...
+    check_max_steps_limit(&execution_resources, max_steps)?;
```
MaksymilianDemitraszek, line 296: Does this function check if the value of max steps is correct? name suggests it.

crates/forge/src/main.rs
MaksymilianDemitraszek: I assume @piotmag769 that this is usefull when all your tests have a lot of steps.

crates/forge-runner/src/running.rs
MaksymilianDemitraszek: Also please don't resolve it here. Please set the default value inside case in cli layer and make it not optional. This leaks cli logic into runner.

MaksymilianDemitraszek (changes requested):
I don't think it works as supposed, can you use it to increase the maximum step limit inside vm/starknet 🤔 ? Also please include integration test checking if this works.

Please create an issue for docs
Please add integration test verifying the correctness when contract calls occur.

.tool-versions
piotmag769: Why? setup scarb action detects .tool-versions and takes scarb version from it (yea, I got to know that yesterday, didn't expect it too xD)

MaksymilianDemitraszek: Because it is unstable and we shouldn't run our tests on it.

piotmag769: If we want to merge this one we have to run tests on this version. The other option is waiting with merging this PR until new stable scarb comes out

MaksymilianDemitraszek: >If we want to merge this one we have to run tests on this version.

we should either ignore those tests or wait, we should not compromise existing features because of the new ones.

tkumor3:
https://github.com/foundry-rs/starknet-foundry/pull/1384#pullrequestreview-1788587082

@MaksymilianDemitraszek I'm wondering, isn't it more useable to set VM max steps to starknet max value and return info about exceeding the limit?  Another thing, do we create only one instance of VM during test running?  we create `BlockContext` in 3 separate places. So I decided to get steps in the same way we get gas.

piotmag769:
@tkumor3 yup, we create the VM in a lot of places. The only improvement I can see is keeping the current number of steps somewhere and passing the `steps_left` to each new VM, so it is handled automatically by blockifier. This way test gets executed to the point of exceeding max steps instead of executing the whole tests and asserting it wasn't exceeded at the end. I guess it is a more general question since we would want to handle available gas in a similar way
cc @war-in @Arcticae

cptartur:
> @tkumor3 yup, we create the VM in a lot of places. The only improvement I can see is keeping the current number of steps somewhere and passing the `steps_left` to each new VM, so it is handled automatically by blockifier. This way test gets executed to the point of exceeding max steps instead of executing the whole tests and asserting it wasn't exceeded at the end. I guess it is a more general question since we would want to handle available gas in a similar way cc @war-in @Arcticae

Unsure about this one. Actually enforcing steps limit could be useful in cases where the users has possibly a long running test and wants it fail early. Though this argument is kind of a stretch.

Ultimately what matters is that we explicitly communicate to users that the steps limit was exceeded, probably by failing a test case.

piotmag769:
@cptartur keep in mind we might want to include the current left resources info in the stack trace in the future, it will limit our ability to do so. This is actually the case with gas, check #1418

cptartur:
> @cptartur keep in mind we might want to include the current left resources info in the stack trace in the future, it will limit our ability to do so. This is actually the case with gas, check #1418

If that's the case, and we need to do this properly for gas, we should handle steps in the same manner. So actually count and enforce it.

cc @MaksymilianDemitraszek @drknzz @Arcticae

crates/forge/tests/data/max_steps/tests/with_attr.cairo
```diff
@@ -0,0 +1,15 @@
+use max_steps::fib;
+
+#[test]
+#[max_steps(120)]
+fn hard_test() {
+    fib(0, 1, 30344);
+    assert(2 == 2, 'simple check');
```
Arcticae, line 7: That's kind of a bummer that we have to make it panicable by adding this assertion, since `max_steps` is kind of an assertion

crates/forge-runner/src/running.rs
```diff
@@ -356,3 +364,20 @@ fn get_context<'a>(runtime: &'a ForgeRuntime) -> &'a EntryPointExecutionContext
...
+    max_steps: Option<u32>,
+) -> Result<()> {
+    let executed_steps = execution_resources.execution_resources.vm_resources.n_steps;
+    if let Some(max_steps) = max_steps {
+        if executed_steps
+            > max_steps
+                .to_usize()
+                .context("Converting max_steps to usize")?
+        {
+            bail!(" Max steps limit exceeded. Limit: {max_steps}. Executed: {executed_steps}")
```
Arcticae, line 379:
```suggestion
            bail!("Max steps limit exceeded. Limit: {max_steps}. Executed: {executed_steps}")
```

crates/forge-runner/src/running.rs
Arcticae: yeah the name seems off

crates/forge/tests/data/max_steps/tests/with_attr.cairo
Arcticae: Can we perhaps make a task to make it work like an assertion (so we don't require the panicable return type of test function)?

Arcticae:
> If that's the case, and we need to do this properly for gas, we should handle steps in the same manner. So actually count and enforce it.

If it's possible to do via blockifier, then we should do it. Although i think gas is different since it has built in compilation mechanisms to generate code that "consumes" the gas before the function is run, so doing that might prove difficult. 

If it's not possible to be done automatically i think the current solution is reasonable.

crates/forge/tests/data/max_steps/tests/with_attr.cairo
piotmag769: Idk if it is worth doing and if it is doable, I believe it has to be panicable to get the `RunResult` in `run_test_case`, but maybe I'm wrong
cc @MaksymilianDemitraszek

piotmag769:
When it comes to bumping `max_steps` above the starknet allowed maximum I think we can bump it to the value the user requested if it is greater than 3M, otherwise stay at 3M. We should check what happens if we exceed the `max_steps` limit from the blockifier context, I guess we would still fail but with different error message

crates/forge/tests/data/max_steps/tests/with_attr.cairo
MaksymilianDemitraszek: I believe it could be relaxed 🤔 not sure if this is worth the effort atm though. Could we create an issue?

MaksymilianDemitraszek:
> I guess we would still fail but with different error message

@piotmag769 why with a different error message?

Issue #1247, opened 2023-11-29
Execute Cast tests in tempdir 


Radinyn:
Some of the Cast tests are executed directly inside the tests folder, this may induce some bugs when running tests locally as Scarb.lock and target files are generated.

PR #1386, opened 2023-12-14
Allow passing optional --yes flag to delete

### LINKED ISSUE #935: Add optional `--yes` flag to subcommands that require user input
We should add an optional `--yes` flag to `sncast account delete` (and other subcommands that require user input) to make them usable in non-interactive scenarios.

DelevoXDG:
Closes #935 

## Introduced changes

- Support optional `--yes` flag to be passed when running `sncast delete`

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/sncast/src/starknet_commands/account/delete.rs
```diff
@@ -48,12 +53,14 @@ pub fn delete(
         serde_json::from_str(&contents).expect("failed to read file { path }");
```
ddoktorski, line 53: Not added in this PR, but does `{ path }` work correctly here?

THenry14: I think so, bo the message is terrible tbh; it says very little to nothing.

crates/sncast/tests/e2e/account/delete.rs
```diff
@@ -253,6 +253,44 @@ pub async fn test_happy_case_without_network_args() {
     let _ = tokio::fs::remove_file("temp_scarb6.toml").await;
 }
 
+#[tokio::test]
+pub async fn test_happy_case_with_yes_flag() {
+    // Creating dummy accounts test file
+    create_dummy_accounts_file("temp_accounts7.json").await;
+    create_dummy_scarb_file("temp_scarb7.toml").await;
```
ddoktorski, line 260: Not in the scope of this PR, but what do you think about changing these functions to use `TempDir` and to return paths to created files? Then we wouldn't need to remove files at the end.

THenry14: https://github.com/foundry-rs/starknet-foundry/issues/1247

crates/sncast/tests/e2e/account/delete.rs
```diff
@@ -253,6 +253,44 @@ pub async fn test_happy_case_without_network_args() {
...
+
+    // Run test with no additional user input
+    let snapbox = Command::new(cargo_bin!("sncast")).args(args);
+    let bdg = snapbox.assert();
+    let out = bdg.get_output();
+    let stdout_str =
+        std::str::from_utf8(&out.stdout).expect("failed to convert command output to string");
+
+    assert!(out.stderr.is_empty());
+    assert!(stdout_str.contains("Account successfully removed"));
```
ddoktorski, line 288: nit: 

We can also use
```
    snapbox.assert().stdout_matches(indoc! {r"
        command: account delete
        Account successfully removed
    "});
```

crates/sncast/src/starknet_commands/account/delete.rs
```diff
@@ -23,6 +23,10 @@ pub struct Delete {
     /// Network where the account exists; defaults to network of rpc node
     #[clap(long)]
     pub network: Option<String>,
+
+    /// Assume "yes" as answer to confirmation prompt and run non-interactively
+    #[clap(long, num_args = 1, default_value = "false")]
+    pub yes: Option<bool>,
```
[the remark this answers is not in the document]
ddoktorski: What about short flag `-y`?

crates/sncast/tests/e2e/account/delete.rs
DelevoXDG: A change we deserve, but don't necessarily need right now? 
Short way is used in all delete tests, and I think it's reasonable to avoid longer test cases, since we don't expect any extra info here, wdyt?

ddoktorski: Sure, it can stay as it is now

THenry14 (approved):

PR #1391, opened 2023-12-14
Use `BlockNumber` instead of `BlockId` in cache

piotmag769:
## Introduced changes

- use `BlockNumber` instead of `BlockId` since after #1383 we always use `BlockId::Number` in cache

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/cheatnet/src/forking/state.rs
```diff
@@ -177,10 +177,10 @@ impl StateReader for ForkStateReader {
             if let Some(cache_hit) = self.cache.get_compiled_contract_class(class_hash) {
                 Ok(cache_hit)
             } else {
-                match self.runtime.block_on(
-                    self.client
-                        .get_class(self.block_id, FieldElement::from_(*class_hash)),
-                ) {
+                match self.runtime.block_on(self.client.get_class(
+                    BlockId::Number(self.block_number.0),
```
war-in, line 181: Would it be possible to create a function/ other suitable solution for this line? It is repeated few times

crates/cheatnet/src/forking/state.rs
```diff
@@ -33,18 +33,18 @@ use url::Url;
 #[derive(Debug)]
 pub struct ForkStateReader {
     client: JsonRpcClient<HttpTransport>,
-    block_id: BlockId,
+    block_number: BlockNumber,
```
war-in, line 36: How about keeping `BlockId::BlockNumber` here? Would it be possible? I just thought that because we always convert it to BlockId

crates/cheatnet/src/forking/state.rs
piotmag769: Will change it to a function, just didn't want to have `self.block_id` since it allows other variants than `BlockId::Number`

crates/cheatnet/src/forking/state.rs
piotmag769: Yes and no - `BlockId` allows other variants as well on the type level :c I will write a function that creates `BlockId` from it, it is just an enum so it is not costly

crates/cheatnet/src/forking/cache.rs
```diff
@@ -79,35 +80,19 @@ impl ToString for ForkCacheContent {
...
-
-fn block_id_to_string(block_id: BlockId) -> String {
-    match block_id {
-        BlockId::Hash(x) => Felt252::from_(x).to_str_radix(16),
-        BlockId::Number(x) => x.to_string(),
-        BlockId::Tag(x) => match x {
-            BlockTag::Latest => "latest".to_string(),
-            BlockTag::Pending => unreachable!(),
-        },
+        self.save();
```
cptartur, line 87: I wonder if that won't cause the cache to grow very fast in case someone uses `latest` in their tests. We'd be caching a new block every couple runs basically.

piotmag769: That's true, but it that really an issue? Cache can always be cleared

cptartur: But we don't have any automatic mechanism of doing so. I don't know how large can a single block cache be, but it should be investigated if an issue like this concerns us.

cc @war-in

Arcticae: Meh i think the cache files are extremely small and shouldn't really worry us

piotmag769: And tbh latest never got there before anyways, we always got the block number before the test, so we already always cached it xD

war-in: In the beginning, it was there. Some changes were made and we forgor about it

cptartur (commented):
Code looks okay. I'm unsure about if all of these tests need to fetch the true latest block (maybe they do).

crates/cheatnet/tests/starknet/cheat_fork.rs
```diff
@@ -9,15 +9,21 @@ use conversions::IntoConv;
...
 
 #[test_case("return_caller_address"; "when common call")]
 #[test_case("return_proxied_caller_address"; "when library call")]
 fn prank_cairo0_contract(selector: &str) {
-    let mut cached_fork_state = create_fork_cached_state();
+    let rt = Runtime::new().unwrap();
+    let cache_dir = TempDir::new().unwrap();
+
+    let mut cached_fork_state =
+        rt.block_on(create_fork_cached_state(cache_dir.path().to_str().unwrap()));
```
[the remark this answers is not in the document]
piotmag769: Wasn't sure either, that's why I did it this way. I guess the only benefit is that we test against the latest network state so we are sure sth unexpected did not change? Idk tho, I would let @war-in decide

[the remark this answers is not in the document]
war-in: Honestly, not sure either. I think we can stick to some predetermined number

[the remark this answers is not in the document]
piotmag769: Okay, will change it then

cptartur (approved):

PR #1397, opened 2023-12-15
Refactor structure dumping

### LINKED ISSUE #1163: Refactor `format_json_value`
... and probably print_command_result

https://github.com/foundry-rs/starknet-foundry/pull/1067#discussion_r1400336692

MaksymilianDemitraszek:
Closes #1163

## Introduced changes

- Improved the code, making it cleaner and more secure
- Constraints the serde json generated froms structs to limited format.
- Introduced `Decimal` and `Hex` types.
- splitted process into atomic steps

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


MaksymilianDemitraszek:
Not finished please don't review it for now

MaksymilianDemitraszek:
@THenry14 @abulenok It is finished, feel free to review

crates/sncast/src/response/print.rs
```diff
@@ -0,0 +1,242 @@
...
+            .collect(),
+        _ => panic!("Expected an object"),
+    }
+}
+
+fn value_to_output_value(value: Value) -> OutputValue {
+    match value {
+        Value::Array(a) => OutputValue::Array(a.into_iter().map(value_to_output_value).collect()),
+        Value::String(s) => OutputValue::String(s.to_string()),
+        s => panic!("{s:?} cannot be auto-serialized to output"),
```
THenry14, line 140: philosophical question - we avoid panicing in cast an any cost basically, I know generally it is ok to do that in binaries, but I'm not sure we should break prog execution that way. Is there any special rationale to do that instead of returning Err in a manner we did up until this point? This is something I wanted to discuss on interfaces meeting, but I'm happy to briefly discuss here as well.

MaksymilianDemitraszek: Rule of thumb is:
- return `Result` when error can be caused by user input/behaviour
- panic when error should never happen and if it happend then you have wrong assumptions while writing the function so you want to know you did something incorrectly. 

Here the generated values are purely based on `CommandResonse` structures which are defined in our codebase in a way that should never parse to something including for example numbers.

MaksymilianDemitraszek: For example you should be ok using panic in cases like `"test".parse().unwrap()` because it is hardcoded. If it breaks you probably should panic because that will clearly signalise to the user that soft is broken and that is not their fault.

THenry14: sure, makes sense. we panic in tests, just avoided doing so in prod code :+1: thanks for an explanation

THenry14 (approved):
lgtm 💪

crates/sncast/src/lib.rs
```diff
@@ -46,55 +45,22 @@ struct Account {
...
-                let arr_as_string = arr
-                    .iter()
-                    .filter_map(|item| self.format_json_value(item))
-                    .collect::<Vec<String>>()
-                    .join(", ");
-                Some(format!("[{arr_as_string}]"))
-            }
-            _ => None,
+    pub fn from_flags(hex_format: bool, dec_format: bool) -> Self {
+        // Clap validates that both are not passed at same time
```
abulenok, line 57: Let's replace this comment with an assert

crates/sncast/src/response/print.rs
```diff
@@ -0,0 +1,242 @@
...
+}
+
+/// Constrained subset of `serde::json`
+#[derive(PartialEq, Eq, Debug)]
+pub enum OutputValue {
+    String(String),
+    Array(Vec<OutputValue>),
+}
+
+type OutputData = Vec<(String, OutputValue)>;
```
abulenok, line 35: Why not add `OrderedMap(IndexMap<String, OutputValue>)`? Then `OutputData` will not be needed

crates/sncast/src/response/print.rs
```diff
@@ -0,0 +1,242 @@
...
+    Ok(())
+}
+
+fn pretty_output(output: OutputData, output_format: &OutputFormat) -> Result<Vec<String>> {
+    match output_format {
+        OutputFormat::Json => {
+            let json_output: HashMap<String, OutputValue> = output.into_iter().collect();
+            let pretty_string = serde_json::to_string(&json_output)?;
+            Ok(vec![pretty_string])
+        }
```
abulenok, line 97:
```suggestion
            let json_output = serde_json::to_string(&json_output)?;
            Ok(vec![json_output])
        }
```

since `serde_json::to_string` is used instead of `serde_json::to_string_pretty`

crates/sncast/src/response/print.rs
MaksymilianDemitraszek: Beacuse we allow maps only on the top level.

MaksymilianDemitraszek: Added comment about this

PR #1400, opened 2023-12-15
Handle sierra compilation in the `smart` way

### LINKED ISSUE #1395: Handle sierra compilation in the `smart` way
Sierra keeps an information about its version so we should use it to choose suitable compiler.
We don't want to try each

war-in:
Closes #1395 

## Introduced changes

- sierra version is extracted from `sierra_program`
- we choose a compiler based on the extracted sierra version
- tests are improved (new test cases were added)

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/sierra-casm/src/lib.rs
```diff
@@ -19,18 +17,19 @@ pub fn compile(mut sierra_json: Value) -> Result<CasmContractClass, String> {
...
-    compile_contract!(ContractClass, CasmContractClass);
-    compile_contract!(ContractClassV1_1, CasmContractClassV1_1);
-    compile_contract!(ContractClassV1Rc0, CasmContractClassV1Rc0);
-    compile_contract!(ContractClassV1Alpha, CasmContractClassV1Alpha);
+    let sierra_version = parse_sierra_version(&sierra_json)?;
+    match sierra_version.as_slice() {
+        [1, 2..=4, 0] => compile_contract!(ContractClass, CasmContractClass),
+        [1, 0..=1, 0] => compile_contract!(ContractClassSierraV1, CasmContractClassSierraV1),
+        [0, ..] => compile_contract!(ContractClassSierraV0, CasmContractClassSierraV0),
+        _ => {}
```
piotmag769, line 31: We can just move the return error here I guess

crates/sierra-casm/tests/compilation.rs
```diff
@@ -1,50 +1,30 @@
...
+    use sierra_casm::compile;
+    use std::fs::File;
+    use test_case::test_case;
+
+    #[test_case("1_4_0"; "sierra 1.4.0")]
+    #[test_case("1_3_0"; "sierra 1.3.0")]
+    #[test_case("1_2_0"; "sierra 1.2.0")]
+    #[test_case("1_1_0"; "sierra 1.1.0")]
+    #[test_case("1_0_0"; "sierra 1.0.0")]
+    #[test_case("0_1_0"; "sierra 0.1.0")]
```
piotmag769, line 12: Indeed mądre

piotmag769 (approved):

crates/sierra-casm/src/lib.rs
```diff
@@ -19,23 +17,23 @@ pub fn compile(mut sierra_json: Value) -> Result<CasmContractClass, String> {
...
     macro_rules! compile_contract {
         ($sierra_type:ty, $casm_type:ty) => {{
-            if let Ok(sierra_class) = serde_json::from_value::<$sierra_type>(sierra_json.clone()) {
-                if let Ok(casm_class) = <$casm_type>::from_contract_class(sierra_class, true) {
-                    return Ok(old_casm_to_newest_casm::<$casm_type>(&casm_class));
-                }
-            }
+            let sierra_class = serde_json::from_value::<$sierra_type>(sierra_json.clone()).unwrap();
+            let casm_class = <$casm_type>::from_contract_class(sierra_class, true).unwrap();
+            return Ok(old_casm_to_newest_casm::<$casm_type>(&casm_class));
```
cptartur, line 22: Couldn't this just be a (generic) function or something like that? I like macros, but I'm unsure if this one provides any benefit.

crates/sierra-casm/src/lib.rs
```diff
@@ -19,23 +17,23 @@ pub fn compile(mut sierra_json: Value) -> Result<CasmContractClass, String> {
...
-    )
+    let sierra_version = parse_sierra_version(&sierra_json)?;
+    return match sierra_version.as_slice() {
+        [1, 2..=4, 0] => compile_contract!(ContractClass, CasmContractClass),
+        [1, 0..=1, 0] => compile_contract!(ContractClassSierraV1, CasmContractClassSierraV1),
+        [0, ..] => compile_contract!(ContractClassSierraV0, CasmContractClassSierraV0),
+        _ =>
+            Err(
+                "Unable to compile Sierra to Casm. No matching ContractClass or CasmContractClass found"
+                    .to_string(),
```
cptartur, line 34: We could also output the version: `No matching ... found for version X.Y.Z`.

crates/sierra-casm/src/lib.rs
```diff
@@ -47,3 +45,15 @@ where
...
     serde_json::from_value::<CasmContractClass>(serialized).unwrap()
 }
+
+fn parse_sierra_version(sierra_json: &Value) -> Result<Vec<u8>, String> {
+    let parsed_values: Vec<u8> = sierra_json["sierra_program"]
+        .as_array()
+        .ok_or("Unable to read sierra_program. Make sure it is an array of felts")?
+        .iter()
+        .take(3)
+        .map(|x| u8::from_str_radix(&x.as_str().unwrap()[2..], 16).unwrap_or_default())
```
cptartur, line 55: Should we fallback to default and not panic?

crates/sierra-casm/src/lib.rs
war-in: Unfortunately, it provides :/ It wouldn't be possible to use `T::from_contract_class` in the function body, because that function is not in the the trait

crates/sierra-casm/src/lib.rs
piotmag769: We can add a comment explaining that I think

cptartur (approved):

crates/sierra-casm/src/lib.rs
```diff
@@ -19,26 +17,27 @@ pub fn compile(mut sierra_json: Value) -> Result<CasmContractClass, String> {
...
-
-    Err(
-        "Unable to compile Sierra to Casm. No matching ContractClass or CasmContractClass found"
-            .to_string(),
-    )
+    let sierra_version = parse_sierra_version(&sierra_json)?;
+    return match sierra_version.as_slice() {
+        [1, 2..=4, 0] => compile_contract!(ContractClass, CasmContractClass),
+        [1, 0..=1, 0] => compile_contract!(ContractClassSierraV1, CasmContractClassSierraV1),
+        [0, ..] => compile_contract!(ContractClassSierraV0, CasmContractClassSierraV0),
```
Arcticae, line 30: What about the minor versions? Will those require updating the range as well?

Arcticae (approved):

PR #1401, opened 2023-12-15
Deployment scripts with ExtendedRuntime

abulenok:
Closes #

## Introduced changes

-

## Checklist

- [ ] Linked relevant issue
- [ ] Added relevant tests


crates/sncast/src/starknet_commands/script.rs
```diff
@@ -1,38 +1,40 @@
...
-use cairo_vm::serde::deserialize_program::ApTracking;
-use cairo_vm::types::exec_scope::ExecutionScopes;
+use cairo_vm::types::relocatable::Relocatable;
 use cairo_vm::vm::errors::hint_errors::HintError;
-use cairo_vm::vm::errors::vm_errors::VirtualMachineError;
-use cairo_vm::vm::runners::cairo_runner::{ResourceTracker, RunResources};
 use cairo_vm::vm::vm_core::VirtualMachine;
 use camino::Utf8PathBuf;
+use cheatnet::constants::{build_block_context, build_transaction_context};
+use cheatnet::state::{CheatnetBlockInfo, DictStateReader};
```
abulenok, line 28: https://github.com/foundry-rs/starknet-foundry/issues/1419

abulenok: This will be refactored in a separate PR

cptartur (approved):

crates/sncast/src/starknet_commands/script.rs
```diff
@@ -51,131 +53,24 @@ pub struct Script {
     pub script_module_name: String,
 }
 
-pub struct CairoHintProcessor<'a> {
+pub struct CastExtension<'a> {
```
piotmag769, line 56: I think moving `CastExtension` to another file and just implementing `run` is this one would be good, but prob out of scope of this PR

crates/sncast/src/starknet_commands/script.rs
```diff
@@ -452,17 +337,53 @@ pub fn run(
...
-    match runner.run_function(
+    let mut cast_runtime = ExtendedRuntime {
+        extension: cast_extension,
+        extended_runtime: StarknetRuntime {
+            hint_handler: syscall_handler,
+        },
+    };
+
+    let mut vm = VirtualMachine::new(true);
+    match runner.run_function_with_vm(
```
piotmag769, line 383: I believe in the future we may want to have the `execute` function that is generalised too (traits etc.) to make sure we don't do the same work twice. For example now we change this place to use casm runner in #1206
cc @cptartur @MaksymilianDemitraszek

piotmag769 (approved):

crates/sncast/src/starknet_commands/script.rs
THenry14: how much effort is this? perhaps this should be done as an immediate follow up to this one?

crates/sncast/src/starknet_commands/script.rs
```diff
@@ -375,49 +256,53 @@ impl CairoHintProcessor<'_> {
...
-                    .write(Felt252::from_(nonce))
-                    .expect("Failed to insert nonce");
 
-                Ok(())
+                let res: Vec<Felt252> = vec![Felt252::from_(nonce)];
+                Ok(CheatcodeHandlingResult::Handled(res))
             }
-            _ => Err(anyhow!("Unknown cheatcode selector: {selector}")),
+            _ => Err(EnhancedHintError::from(anyhow!(
+                "Cheatcode {selector} not implemented"
```
THenry14, line 282: I'm not sure we should use `cheatcode` naming in errors and messages to users in cast tbh. I find it a bit misleading, as this is something that immidiately makes me think of forge - in cast we have subcommands (or just functions in cast_std) instead. Perhaps we should stick to this naming? I'm not 100% sure, so perhaps let's discuss.

crates/sncast/src/starknet_commands/script.rs
```diff
@@ -375,49 +256,53 @@ impl CairoHintProcessor<'_> {
...
         }?;
 
-        let result_end = buffer.ptr;
-        insert_value_to_cellref!(vm, output_start, result_start)?;
-        insert_value_to_cellref!(vm, output_end, result_end)?;
+        Ok(res)
+    }
 
-        Ok(())
+    fn override_system_call(
```
THenry14, line 289: why do we need this?

crates/sncast/src/starknet_commands/script.rs
abulenok: almost no effort, so yeah, I will make a PR immediately after this one

crates/sncast/src/starknet_commands/script.rs
abulenok: To turn off Starknet syscalls. 
When we decide which ones we want to support, we will forward the call to blockifier for them and block others

THenry14: sure, thanks for clarification. Do you mind adding a comment about this?

abulenok: We could do that, but I'm afraid this comment can get outdated at some point.

I feel like it's deductible from the code since the name of the function suggests that we override syscall behaviour and for each syscall, the error is thrown (which tells that Starknet syscalls are not supported)

abulenok: I think the `override_system_call` may be renamed to `override_starknet_system_call` or `override_contract_system_call`

@MaksymilianDemitraszek WDYT?

crates/sncast/src/starknet_commands/script.rs
abulenok: I think we can stick to using `subcommands` in deployment scripts.

abulenok: Other options I'm thinking of are:
- "Selector {selector} not implemented"
- "{selector} not implemented"

abulenok: But we can use `subcommands` for now

crates/sncast/src/starknet_commands/script.rs
piotmag769: I think in case of plain cairo you just say "hints", I have seen word "syscall" used only in the context of starknet

crates/sncast/src/starknet_commands/script.rs
MaksymilianDemitraszek: yes sir

crates/sncast/src/starknet_commands/script.rs
MaksymilianDemitraszek: It is probably CastScriptExtension

crates/sncast/src/starknet_commands/script.rs
THenry14: Cool, let's make it a stacked PR and merge those two PRs at once, sounds ok?

crates/sncast/src/starknet_commands/script.rs
THenry14: this error originates in scripts, so I think `functions` may have more sense. this is just a nitpick though, if you like subcommands more I won't oppose.

crates/sncast/src/starknet_commands/script.rs
abulenok: renamed

crates/sncast/src/starknet_commands/script.rs
abulenok: https://github.com/foundry-rs/starknet-foundry/pull/1506

piotmag769:
Do we want to merge it? @MaksymilianDemitraszek @THenry14

crates/sncast/src/starknet_commands/script.rs
MaksymilianDemitraszek: I think cheatcode is the name of the cairo function, nothing cast/forge related, you shouldn't handle it at all here, some extensions under this one might be able to handle it.

crates/sncast/src/starknet_commands/script.rs
MaksymilianDemitraszek: override_starknet_system_call i am ok with changing this if you think it will be more intuitive but please do it in a separate PR.

MaksymilianDemitraszek (approved):
LGTM but please not handle missing cheatcodes here.

THenry14 (approved):

PR #1416, opened 2023-12-19
replace transaction receipt with transaction status to get transaction status

tkumor3:
Closes #

## Introduced changes

-

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/sncast/src/lib.rs
```diff
@@ -278,15 +279,27 @@ pub async fn wait_for_tx(
...
-            Ok(receipt) => {
-                return match receipt.execution_result() {
-                    ExecutionResult::Succeeded => Ok("Transaction accepted"),
-                    ExecutionResult::Reverted { reason } => {
-                        Err(anyhow!("Transaction has been reverted = {reason}"))
+        match provider.get_transaction_status(tx_hash).await {
+            Ok(status) => match status {
+                starknet::core::types::TransactionStatus::Received => {
+                    return Ok("Transaction has been received");
+                }
```
ddoktorski, line 286: I think once transaction is received it still can be rejected or reverted so I guess we shouldn't return here?

crates/sncast/src/lib.rs
```diff
@@ -278,15 +279,27 @@ pub async fn wait_for_tx(
...
+                starknet::core::types::TransactionStatus::AcceptedOnL2(execution_status)
+                | starknet::core::types::TransactionStatus::AcceptedOnL1(execution_status) => {
+                    match execution_status {
+                        starknet::core::types::TransactionExecutionStatus::Succeeded => {
+                            return Ok("Transaction accepted")
+                        }
+
+                        starknet::core::types::TransactionExecutionStatus::Reverted => {
+                            return Err(anyhow!("Transaction has been reverted"))
+                        }
```
ddoktorski, line 299: Are you sure this case is possible? I mean `AcceptedOnL2`/`AcceptedOnL1` + `Reverted`

crates/sncast/tests/e2e/invoke.rs
```diff
@@ -34,6 +38,80 @@ async fn test_happy_case() {
...
+        "99999999999999999",
+        "--nonce",
+        &nonce_str,
+    ]);
+
+    let snapbox = runner(&args);
+
+    snapbox.assert().stderr_matches(indoc! {r"
+        command: invoke
+        error: Transaction has been reverted
```
ddoktorski, line 112: Wasn't the purpose of this test to check if the transaction was rejected? (the name is `test_tx_rejected`)

crates/sncast/src/lib.rs
THenry14: yup, we should sleep and wait a bit longer, until it changes status.

crates/sncast/tests/e2e/invoke.rs
```diff
@@ -34,6 +38,80 @@ async fn test_happy_case() {
 
     assert!(matches!(receipt, Invoke(_)));
 }
+#[tokio::test]
+async fn test_tx_reverted() {
```
THenry14, line 42: I guess we should rather have an integration test on this change (we already test wait_for_tx transactions in tests/integration/wait_for_tx.rs). 

That way we could keep a distinction between our test files (eg this file should probably test just invoke subcommand, and not waiting for txs). I'm not opposed to have a test that uses --wait in here, but I guess that we shouldn't mix different tests (with different purposes) with each other, it might get too messy and harder to maintain.

crates/sncast/src/lib.rs
tkumor3: So I will print it only and continue waiting for acceptation.

crates/sncast/src/lib.rs
tkumor3:
```
pub enum TransactionStatus {
    Received,
    Rejected,
    AcceptedOnL2(TransactionExecutionStatus),
    AcceptedOnL1(TransactionExecutionStatus),
}
```
I based on this starknet-rs enum

crates/sncast/tests/e2e/invoke.rs
tkumor3: good catch! Tomek🏺

crates/sncast/src/lib.rs
tkumor3: Zrzut ekranu 2023-12-22 o 14 39 47

crates/sncast/src/lib.rs
```diff
@@ -278,15 +279,27 @@ pub async fn wait_for_tx(
...
     for i in (1..retries).rev() {
-        match provider.get_transaction_receipt(tx_hash).await {
-            Ok(receipt) => {
-                return match receipt.execution_result() {
-                    ExecutionResult::Succeeded => Ok("Transaction accepted"),
-                    ExecutionResult::Reverted { reason } => {
-                        Err(anyhow!("Transaction has been reverted = {reason}"))
+        match provider.get_transaction_status(tx_hash).await {
+            Ok(status) => match status {
+                starknet::core::types::TransactionStatus::Received => {
```
THenry14, line 284: I think we should probably combine this with ` Err(StarknetError(TransactionHashNotFound))`, otherwise we loose an info about retries/timeouts + in both cases we wait for tx acceptance, which at the higher level is very similar scenario

crates/sncast/tests/helpers/fixtures.rs
```diff
@@ -122,6 +123,51 @@ pub async fn invoke_map_contract(key: &str, value: &str, account: &str, contract
     execution.send().await.unwrap();
 }
 
+pub async fn invoke_udc_contract(
```
THenry14, line 126: I'd rather this could be used as general `invoke_contract` helper function, that takes an address, salt, uniqueness etc as arguments. That way this will be more reusable and won't have hardcoded values.

crates/sncast/src/lib.rs
tkumor3: I left both information, about the reason for retry and number of retries

crates/sncast/src/lib.rs
```diff
@@ -278,22 +279,36 @@ pub async fn wait_for_tx(
...
+                            return Err(anyhow!("Transaction has been reverted"))
+                        }
                     }
                 }
-            }
+            },
             Err(StarknetError(TransactionHashNotFound)) => {
-                let remaining_time = i * u16::from(retry_interval);
-                println!("Waiting for transaction to be accepted ({i} retries / {remaining_time}s left until timeout)");
+                println!("Transaction has not be found");
```
THenry14, line 304:
```suggestion
                println!("Transaction not received yet - transaction hash not found");
```
nitpick

crates/sncast/tests/e2e/invoke.rs
```diff
@@ -3,7 +3,9 @@ use crate::helpers::fixtures::{
     default_cli_args, from_env, get_transaction_hash, get_transaction_receipt,
 };
 use crate::helpers::runner::runner;
+
```
THenry14, line 6: were those added by formatter?

THenry14 (commented):
looks good, a couple of nitpicks though

crates/sncast/src/lib.rs
```diff
@@ -278,22 +279,36 @@ pub async fn wait_for_tx(
...
+                }
+                starknet::core::types::TransactionStatus::AcceptedOnL2(execution_status)
+                | starknet::core::types::TransactionStatus::AcceptedOnL1(execution_status) => {
+                    match execution_status {
+                        starknet::core::types::TransactionExecutionStatus::Succeeded => {
+                            return Ok("Transaction accepted")
+                        }
+
+                        starknet::core::types::TransactionExecutionStatus::Reverted => {
+                            return Err(anyhow!("Transaction has been reverted"))
```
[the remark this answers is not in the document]
tkumor3: It could be done by calling rpc for a receipt. Status endpoint returns only status

crates/sncast/tests/e2e/invoke.rs
tkumor3: probably yes, should I remove it?

crates/sncast/src/lib.rs
```diff
@@ -278,22 +279,36 @@ pub async fn wait_for_tx(
...
-        match provider.get_transaction_receipt(tx_hash).await {
-            Ok(receipt) => {
-                return match receipt.execution_result() {
-                    ExecutionResult::Succeeded => Ok("Transaction accepted"),
-                    ExecutionResult::Reverted { reason } => {
-                        Err(anyhow!("Transaction has been reverted = {reason}"))
+        match provider.get_transaction_status(tx_hash).await {
+            Ok(status) => match status {
+                starknet::core::types::TransactionStatus::Received => {
+                    println!("Transaction has been received");
```
ddoktorski, line 285: If there is a delay between the receipt and acceptance of a transaction I guess we may end up printing this message multiple times, which I think might not be an ideal outcome. What are your thoughts?

crates/sncast/src/lib.rs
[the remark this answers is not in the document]
THenry14: then maybe we should call for receipt in this place and give the user revert reason? we do not have a command to allow users to fetch it, that's why I think it might make sense. Could be removed after we add such functionality.

crates/sncast/tests/e2e/invoke.rs
THenry14: if formatter adds it, I guess removing this will end up with failed CI. Let's make sure that was the case, and if not - remove it. If yes, let's leave this.

crates/sncast/src/lib.rs
THenry14: hmm, that's true actually. Maybe we should combine Received with HashNotFound, as they are very similar? I think i raised this somewhere in this PR. That would make for simpler messaging probably.

ddoktorski: We can also include some logic to print the received message just once. As for `HashNotFound`, in my view we probably don't need to print a message at all. We could treat it as an expected error, which is not relevant for a user and simply retry after it.

THenry14: yup, that's what I had in mind. Basically we could just combine both matches into one, and only print the wait/timeout message

THenry14: relevant discussion https://github.com/foundry-rs/starknet-foundry/pull/1416#discussion_r1440442633

tkumor3: There are nested matches, any idea how to do it better than I did?

THenry14: Rough idea would be to simply get rid of nested matches (for example match status can be removed). 

To be more precise:
- remove Ok(status) => match status altogether
- pack TransactionStatuses in Result (Ok())
- combine Received with TxHashNotFound
- profit

Let me know if everything is clear, and if you need any further help :+1:

crates/sncast/src/lib.rs
```diff
@@ -278,22 +279,29 @@ pub async fn wait_for_tx(
...
+                        return get_revert_reason(provider, tx_hash).await
                     }
                 }
             }
-            Err(StarknetError(TransactionHashNotFound)) => {
-                let remaining_time = i * u16::from(retry_interval);
-                println!("Waiting for transaction to be accepted ({i} retries / {remaining_time}s left until timeout)");
-            }
+            Ok(starknet::core::types::TransactionStatus::Received)
+            | Err(StarknetError(TransactionHashNotFound)) => {}
```
THenry14, line 298: Any reason why the remaining time calculation and printing cannot be included in case where received | notfound?

crates/sncast/src/lib.rs
```diff
@@ -302,6 +310,23 @@ pub async fn wait_for_tx(
...
 }
 
+async fn get_revert_reason(
+    provider: &JsonRpcClient<HttpTransport>,
+    tx_hash: FieldElement,
+) -> Result<&str> {
+    match provider.get_transaction_receipt(tx_hash).await {
+        Ok(receipt) => match receipt.execution_result() {
+            starknet::core::types::ExecutionResult::Succeeded => {
+                unreachable!()
```
THenry14, line 320: why `unreachable`?

crates/sncast/tests/integration/wait_for_tx.rs
```diff
@@ -50,6 +52,29 @@ async fn test_rejected_transaction() {
     assert!(resp.to_string().contains("InsufficientMaxFee"));
 }
 
+#[tokio::test]
+#[should_panic(expected = "Transaction has been reverted = Insufficient max")]
```
THenry14, line 56: Are you sure the revert message is `Insufficient max` ? seems a bit odd

crates/sncast/src/lib.rs
tkumor3: We use this function to get the reason for the revert, so currently, it's executed only when we know that tx was reverted.

crates/sncast/tests/integration/wait_for_tx.rs
tkumor3: In this example yes! The max_fee is not enough, so that's the reason for the revert

THenry14: I just checked and it seems the message is actually longer :P I guess should_panic matches pattern somehow. Let's inlude a bit longer message to avoid confusion (`(...) max fee:` should be enough)

```
---- integration::wait_for_tx::test_wait_for_reverted_transaction stdout ----
Transaction hash = 0x2e7a07b5273b142f9657bac3f642eb5f9bb38512545f6b564791d2cbbc91203
thread 'integration::wait_for_tx::test_wait_for_reverted_transaction' panicked at crates/sncast/tests/integration/wait_for_tx.rs:74:10:
called `Result::unwrap()` on an `Err` value: Transaction has been reverted = Insufficient max fee: max fee: Fee(332323232324342), actual fee: Fee(801700000000000)
```

crates/sncast/src/lib.rs
tkumor3: but what we should do in case when the status doesn't match `Reverted`? should we panic?

THenry14: yes, this is a error in our codebase, so we should panic

THenry14: > And what about the case when a call response with an error?
throw this error, but do not panic

tkumor3: done

THenry14 (approved):

PR #1417, opened 2023-12-19
`store` cheatnet function

Arcticae:
- [x] Linked relevant issue
- [x] Added relevant tests


crates/cheatnet/src/runtime_extensions/forge_runtime_extension/cheatcodes/load.rs
```diff
@@ -43,7 +44,21 @@ pub fn load(
...
 /// <https://docs.starknet.io/documentation/architecture_and_concepts/Smart_Contracts/contract-storage>
 #[must_use]
-pub fn map_storage_address(address: FieldElement) -> FieldElement {
+fn normalize_storage_address(address: FieldElement) -> FieldElement {
     let modulus = Felt252::from(2).pow(251) - Felt252::from(256);
     address % modulus.into_()
 }
+
+#[must_use]
+pub fn calculate_variable_address(selector: Felt252, key: Option<&[Felt252]>) -> Felt252 {
```
drknzz, line 53: Could this function take in variable name instead of selector? It seems every use needs to be preceded by calculating the selector

crates/cheatnet/tests/cheatcodes/store.rs
```diff
@@ -0,0 +1,208 @@
...
+    let class_hash = blockifier_state.declare(&contract, &contracts).unwrap();
+
+    let contract_address = deploy(&mut blockifier_state, &mut cheatnet_state, &class_hash, &[])
+        .unwrap()
+        .contract_address;
+
+    store(
+        &mut blockifier_state,
+        contract_address,
+        &starknet_keccak("balance".as_ref()),
```
drknzz, line 29: Could this be `felt_selector_from_name`?

crates/cheatnet/tests/cheatcodes/load.rs
```diff
@@ -130,15 +126,13 @@ fn load_state_map_complex_value() {
     )
     .unwrap();
 
-    // This can be abstracted in cairo code itself, without compromising load's interface simplicity
```
drknzz, line 133: Is this comment no longer valid?

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/cheatcodes/load.rs
Arcticae: This is going to be calculated in cairo via the `selector!` macro so that's why i decided to converge towards that interface

crates/cheatnet/tests/cheatcodes/store.rs
Arcticae: good catch :)

crates/cheatnet/tests/cheatcodes/load.rs
Arcticae: Yes i think it's been abstracted pretty well in the `calculate_address` so that in cairo you just provide the `selector!` (which you provide yourself as a string) and the `Key`s will be taken from the appropriate hash-serializable key value (if the value you want to fetch is in a mapping you have to implement hash)

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/cheatcodes/load.rs
drknzz: Maybe another function then? 👀

crates/cheatnet/tests/cheatcodes/load.rs
drknzz: Okay, so I guess a few lines down there is the same comment that is still present

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/cheatcodes/load.rs
Arcticae: it would be written purely for tests, so i can perhaps put it in there if you'd like that solution more

crates/cheatnet/tests/cheatcodes/load.rs
Arcticae: Deleted

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/cheatcodes/load.rs
drknzz: I would like that 👍

crates/cheatnet/tests/cheatcodes/store.rs
war-in, line 1: Should we maybe test errors? Would that be possible?

drknzz (approved):

crates/cheatnet/tests/cheatcodes/store.rs
Arcticae: There are none that are expected TBH (that i know of)
Any specifc ones in mind?

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/cheatcodes/load.rs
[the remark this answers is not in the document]
Arcticae: I can put it in a separate file but i thought keeping it in either one or other file would make more sense.
I can as well merge those 2 files into one (store + load + those utils)
Thoughts?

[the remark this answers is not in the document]
war-in: I think a merge would be perfect since those functions are strongly connected

crates/cheatnet/tests/cheatcodes/store.rs
war-in: There is one `?` in the `store` function. We could check that however I'm not sure if its necessary

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/cheatcodes/load.rs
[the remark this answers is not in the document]
Arcticae: Done

crates/cheatnet/tests/cheatcodes/store.rs
Arcticae: I checked the `StateError` and it seems that those mostly unexpected so i don't think we should really test for that

drknzz (approved):

war-in (approved):

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/cheatcodes/storage.rs
```diff
@@ -43,7 +72,21 @@ pub fn load(
...
 /// <https://docs.starknet.io/documentation/architecture_and_concepts/Smart_Contracts/contract-storage>
 #[must_use]
-pub fn map_storage_address(address: FieldElement) -> FieldElement {
+fn normalize_storage_address(address: FieldElement) -> FieldElement {
     let modulus = Felt252::from(2).pow(251) - Felt252::from(256);
     address % modulus.into_()
 }
+
+#[must_use]
+pub fn calculate_variable_address(selector: Felt252, key: Option<&[Felt252]>) -> Felt252 {
```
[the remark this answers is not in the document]
Arcticae: The cheatcode is meant to be used in tandem with `selector!` macro

crates/cheatnet/tests/cheatcodes/mod.rs
```diff
@@ -9,4 +13,13 @@ mod precalculate_address;
...
+mod store;
 mod warp;
+
+pub fn map_entry_address(var_name: &str, key: &[Felt252]) -> Felt252 {
+    calculate_variable_address(felt_selector_from_name(var_name), Some(key))
+}
+
+pub fn variable_address(var_name: &str) -> Felt252 {
+    calculate_variable_address(felt_selector_from_name(var_name), None)
+}
```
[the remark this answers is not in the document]
Arcticae: We won't be calculating the selector under the hood since we need it from the user. Right now there isn't a reliable option covering all cases which will work for this interface (long strings only supported in macros)

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/cheatcodes/storage.rs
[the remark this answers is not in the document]
piotmag769: Makes sense

PR #1426, opened 2023-12-21
Add .tool-versions to tempdirs

### LINKED ISSUE #1392: Set `asdf local scarb...` for all temdir tests

Forge

### Foundry Version

latest

MacOS

arm

### What happened

We should set local scarb when running tests because they are running in the temdirs so `.tool-versions` from our main directory is not enough

### Trace

_No response_

drknzz:
Closes #1392 

## Introduced changes

-

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/forge/test_utils/src/lib.rs
```diff
@@ -1,3 +1,12 @@
-pub mod corelib;
 pub mod runner;
 pub mod running_tests;
+
+use anyhow::Result;
+use assert_fs::fixture::PathCopy;
+
+pub fn tempdir_with_tool_versions() -> Result<assert_fs::TempDir> {
```
drknzz, line 7: Might not be the place for it, will gladly take better ideas

crates/forge/src/scarb.rs
```diff
@@ -75,9 +75,10 @@ mod tests {
     use indoc::{formatdoc, indoc};
     use scarb_metadata::MetadataCommand;
     use std::str::FromStr;
+    use test_utils::tempdir_with_tool_versions;
 
     fn setup_package(package_name: &str) -> TempDir {
-        let temp = TempDir::new().unwrap();
+        let temp = tempdir_with_tool_versions().unwrap();
```
war-in, line 81: I don't think we should use functions from the `test_utils` outside the tests and `benchmarks`
cc @cptartur 

maybe it's time to consider a better place for test_utils dir?

crates/forge/test_utils/src/lib.rs
war-in: Let's put it somewhere in the `tests` dir

crates/forge/src/scarb.rs
cptartur: If `tempdir_with_tool_versions` is used in normal code, it should be added to the `forge` crate somewhere. And then we can still use it in tests.

Arcticae (approved):
Apart from @war-in comments it looks good to me

crates/forge/src/scarb.rs
war-in: but it shouldn't appear in the `normal` code. it is used inside some scarb tests

cptartur: Ah, right. That's test code. I don't think it's inherently wrong to use code from `test_utils` in normal package. I'd just make sure we only use `test_utils` as a dev-dependency.

crates/forge/tests/e2e/collection.rs
```diff
@@ -1,12 +1,13 @@
...
 use assert_fs::fixture::PathCopy;
 use indoc::indoc;
+use test_utils::tempdir_with_tool_versions;
 
 use crate::e2e::common::runner::test_runner;
 
 #[test]
 fn collection_with_lib() {
-    let temp = assert_fs::TempDir::new().unwrap();
+    let temp = tempdir_with_tool_versions().unwrap();
```
piotmag769, line 10: We could replace those in this file with `setup_package` I think

piotmag769 (approved):

crates/forge/test_utils/src/lib.rs
drknzz: Kept it here, as it is used in `forge/src/scarb.rs` tests and `test_utils` is a dev-dependency anyway

PR #1440, opened 2023-12-27
Refactor values extraction from inputs in HintProcessor / ExtensionLogic

abulenok:
Closes #

## Introduced changes

- Refactor values extraction from inputs in `HintProcessor::execute_hint` / `ExtenstionLogic::handle_cheatcode`

## Checklist

- [ ] Linked relevant issue
- [ ] Added relevant tests

### DIFF
--- a/crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
+++ b/crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
@@ -26,7 +26,7 @@ use starknet_api::core::ContractAddress;
 use crate::runtime_extensions::forge_runtime_extension::cheatcodes::spy_events::SpyTarget;
 use crate::runtime_extensions::forge_runtime_extension::file_operations::string_into_felt;
 use cairo_lang_starknet::contract::starknet_keccak;
-use runtime::utils::{read_option_felt, read_option_vec, read_vec};
+use runtime::utils::Reader;
 use runtime::{
     CheatcodeHandlingResult, EnhancedHintError, ExtendedRuntime, ExtensionLogic,
     SyscallHandlingResult,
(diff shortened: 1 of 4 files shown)


crates/runtime/src/utils.rs
```diff
@@ -0,0 +1,30 @@
+use cairo_felt::Felt252;
+use num_traits::cast::ToPrimitive;
+use num_traits::identities::One;
+
+pub fn read_felt_and_increase_idx(buffer: &[Felt252], idx: &mut usize) -> Felt252 {
```
piotmag769, line 5: I would stick with the previous names, the fact that idx changes is signalised by the signature

piotmag769 (approved):
LGTM, please address the comment

crates/runtime/src/utils.rs
MaksymilianDemitraszek: I would create `Reader` struct which holds an index and buffer instead, would hide this implementation detail and be more concise. (Like Buffer in cario but not relying on pointers etc).

MaksymilianDemitraszek: We could go one step further and pass it to the function instead of `buffer: &[Felt252]`

MaksymilianDemitraszek:
Introduced changes section is empty, would be nice to have is briefly described what this PR introduces.

Arcticae (changes requested):
Agree with @MaksymilianDemitraszek comment about wrapping the functionality into a struct

crates/runtime/src/utils.rs
MaksymilianDemitraszek: Also those functions problably shouldn't panic when parsing fails, input data could be anything as user can put anything as cheatcode argument.

Arcticae: #1469

abulenok: the only place where it may panic is [here](https://github.com/foundry-rs/starknet-foundry/pull/1440/files#diff-4658fbded6e024a7e21c213e4b278f37d12083f980631a8b91d6f9a09f54685cR21), which is OK, since it indicates that this is not a vector

abulenok: I'll improve error msg

piotmag769: @abulenok they were prob thinking of places like this too
https://github.com/foundry-rs/starknet-foundry/blob/e9daf1227c6589a4fc0c26aa22851a4e78a58c55/crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs#L549

MaksymilianDemitraszek: @piotmag769 this actually is probably out of scope of this work (as you mentioned), could we create a separate issue for stuff like this?

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
```diff
@@ -45,6 +46,29 @@ pub struct ForgeExtension<'a> {
...
+            Some(1) => CheatTarget::One(self.read_felt().into_()),
+            Some(2) => {
+                let contract_addresses: Vec<_> = self
+                    .read_vec()
+                    .iter()
+                    .map(|el| ContractAddress::from_(el.clone()))
+                    .collect();
+                CheatTarget::Multiple(contract_addresses)
+            }
+            _ => unreachable!("Invalid CheatTarget variant"),
```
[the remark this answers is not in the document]
Arcticae: The PR doesn't include the issue being closed so i think that one will be still relevant

crates/runtime/src/utils.rs
Arcticae, line 1: I think those should not panic, so that we're able to handle the error appropriately down the line

Arcticae (changes requested):

Arcticae (approved):
Disregard my last review, just saw the comment that we want to handle this separately

war-in (approved):
🔥

PR #1466, opened 2024-01-02
Better print for trace info

piotmag769:
## Introduced changes

Self explanatory

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
```diff
@@ -508,6 +508,36 @@ impl<'a> ExtensionLogic for ForgeExtension<'a> {
...
+                    Some(2) => EntryPointType::L1Handler,
+                    _ => unreachable!("Invalid EntryPointType variant"),
+                };
+                let entry_point_selector = &inputs[1];
+
+                let calldata_length = inputs[2].to_usize().unwrap();
+                let calldata = &inputs[3..(3 + calldata_length)];
+
+                let storage_address = &inputs[3 + calldata_length];
+                let caller_address = &inputs[4 + calldata_length];
```
drknzz, line 524: There are `read_` functions in this file that could be used here

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
```diff
@@ -508,6 +508,37 @@ impl<'a> ExtensionLogic for ForgeExtension<'a> {
                 }
                 Ok(CheatcodeHandlingResult::Handled(output))
             }
+            "print_call_entry_point" => {
+                let mut idx = 0;
+                let entry_point_type = match read_felt(&inputs, &mut idx).to_u8() {
+                    Some(0) => EntryPointType::Constructor,
+                    Some(1) => EntryPointType::External,
+                    Some(2) => EntryPointType::L1Handler,
+                    _ => unreachable!("Invalid EntryPointType variant"),
```
piotmag769, line 517: This is bad, but I assume it will be fixed when we have a more generic solution (custom error being wrapped with an error message about version of snforge_std) in #1440

war-in (commented):
Can't we just use `println` macro? I feel kinda worried about using cheatcodes just for printing stuff

piotmag769:
@war-in this was requested by @Arcticae, does cairo have `println!` macro already?

war-in:
> @war-in this was requested by @Arcticae, does cairo have `println!` macro already?

Yup, it was introduced in the 2.4.0
https://arc.net/l/quote/qdbvuotj

crates/forge/tests/e2e/trace_info.rs
```diff
@@ -19,19 +19,15 @@ fn trace_info_print() {
...
-        original value: [1242951957743815716965], converted to a string: [Call Type]
-        original value: [1130458220], converted to a string: [Call]
-        [PASS] tests::available_gas::test_trace_info_print, gas: ~1848
+
+        Entry Point Type: External
+        Selector: 0x24417f6c16e7636bef3061ce6a66f3a0bcaed4d62b43635d086869a32c204db
+        Calldata: [10]
+        Storage Address: 0x5df528bb5565afbd82e8edf5dc454f94ad15d4c1140f0661059dc33a451aa3e
+        Caller Address: 0x1724987234973219347210837402
+        Call Type: Call
```
MaksymilianDemitraszek, line 28: For debugging purposes probably would be worth to store the contract name (in case we have it) and print it too.

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
MaksymilianDemitraszek: Canno we return unrecoverable error from here?

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
```diff
@@ -508,6 +508,37 @@ impl<'a> ExtensionLogic for ForgeExtension<'a> {
                 }
                 Ok(CheatcodeHandlingResult::Handled(output))
             }
+            "print_call_entry_point" => {
```
MaksymilianDemitraszek, line 511: Probably we should move it to IO extension, it would be useful to make it reusable in cast.

MaksymilianDemitraszek: Also we could make it more generic and call it flexible_print or something like that so we can reuse it for other structures.

crates/forge/tests/e2e/trace_info.rs
piotmag769: Sadly it is not there anymore, I doubt it is even in sierra code of the contract

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
piotmag769: Not really since it is later cast to `HintError`, this error will result in `CairoRunError` and be evaluated to failed test result

piotmag769:
Will open another PR, changed the approach and rebased

PR #1484, opened 2024-01-04
Set edition in template

### LINKED ISSUE #1176: Set edition in template & snforge_std and update imports
Set an edition in forge template to an explicit value (preferably the newest edition), and update code and imports so it works.

drknzz:
Closes #1176 

## Introduced changes

- Set `edition = "2023_10"` in template and snforge_std
- Update snforge_std imports

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests

### DIFF
--- a/crates/forge/src/init.rs
+++ b/crates/forge/src/init.rs
@@ -8,7 +8,9 @@ use std::fs::{self, OpenOptions};
 use std::io::Write;
 use std::path::Path;
 use std::str::from_utf8;
-use toml_edit::{ArrayOfTables, Document, Item, Table};
+use toml_edit::{value, ArrayOfTables, Document, Item, Table};
+
+const CAIRO_EDITION: &str = "2023_10";
 
 static TEMPLATE: Dir = include_dir!("starknet_forge_template");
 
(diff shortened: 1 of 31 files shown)


snforge_std/Scarb.toml
```diff
@@ -1,6 +1,7 @@
 [package]
 name = "snforge_std"
 version = "0.1.0"
+edition = "2023_10"
```
drknzz, line 4: Should it be added to changelog?

crates/forge/src/init.rs
```diff
@@ -61,6 +63,17 @@ fn add_target_to_toml(path: &Path) -> Result<()> {
     Ok(())
 }
 
+fn set_cairo_edition(path: &Path, cairo_edition: &str) -> Result<()> {
+    let config_file = fs::read_to_string(path)?;
+    let mut doc = config_file.parse::<Document>().expect("invalid document");
```
piotmag769, line 68: I think this parsing should be moved to `run` or a separate function so we don't parse Scarb.toml twice (here and in `add_target_to_toml`)

snforge_std/Scarb.toml
piotmag769: We can I guess

piotmag769 (approved):

Arcticae (approved):

piotmag769:
Actually shouldn't we set it explicitly in `sncast_std` too? cc @THenry14

THenry14:
> Actually shouldn't we set it explicitly in `sncast_std` too? cc @THenry14

It makes sense to do this, yes.

PR #1485, opened 2024-01-04
Add design doc for new declare/script interface

THenry14:
## Introduced changes

- Opens up a discussion about sncast declare/script interfaces, first raised under this PR https://github.com/foundry-rs/starknet-foundry/pull/1093

## Checklist

- [X] Linked relevant issue
- [X] Added relevant tests

### DIFF
--- a/design_documents/sncast_interface_overhaul.md
+++ b/design_documents/sncast_interface_overhaul.md
@@ -25,19 +25,22 @@ to look like this:
 
 ```bash
 $ # this is the same module
-$ sncast script package_namea::path::to::script_module
+$ sncast script path::to::script_module
 $ sncast script to::script_module
 $ sncast script script_module
 
 $ # this is the same contract
-$ sncast declare package_namea::path::to::contract
+$ sncast declare path::to::contract
 $ sncast declare to::contract
-$ sncast declare script_module
+$ sncast declare contract
+
+$ # contract is ambiguous - there are multiple contracts named 'contract' in multiple packages
+$ sncast --package packagea declare contract
 ```
 
 In case there is only one script module/contract with specified name in workspace, only this name can be used to target it.
-Otherwise, if it is ambiguous in the context of workspace, we should show all matches with full paths, to let user decide
-which one to use.
+Otherwise, if it is ambiguous in the context of workspace, we should show all matches with full paths (along with package names),
+to let user decide which one to use. The `--package` flag would then be required, and would indicate which package to use.
 
 This change would mean that from now on we will have unified interface for declare/script, always assume we're in some workspace/package, 
 the `--path-to-scarb-toml` flag can be removed.
--- a/design_documents/sncast_interface_overhaul.md
+++ b/design_documents/sncast_interface_overhaul.md
@@ -31,8 +31,8 @@ $ sncast script script_module
 
 $ # this is the same contract
 $ sncast declare package_namea::path::to::contract
-$ sncast to::contract
-$ sncast script script_module
+$ sncast declare to::contract
+$ sncast declare script_module
 ```
 
 In case there is only one script module/contract with specified name in workspace, only this name can be used to target it.
--- a/design_documents/sncast_interface_overhaul.md
+++ b/design_documents/sncast_interface_overhaul.md
@@ -1,0 +1,47 @@
+# sncast interface overhaul
+
+## Context
+In light of needed changes on how the scripts and contract declaration work (ie adding support for scarb workspaces and various directory layouts),
+we need to re-think how we call the respective subcommands, and try to improve it. This design doc covers only a portion
+(although significant one) of sncast interface - it was decided recently we should make the interfaces as 'clap-native'
+as possible, but this design doc won't cover those changes in detail, as they are pretty straight forward and already have
+an opened issue.
+
+## The 'now'
+Currently, when declaring a contract we require user to pass the contract name using `--contract-name` flag. This flag only
+takes the contract's name, so for example it is not possible to specify the package this contract was defined in (only
+achievable when passing `--path-to-scarb-toml`). 
+
+When running a script we require a positional argument, that is the script package name. Similarly to declare, we could
+pass `--path-to-scarb-toml` if we wanted to pick the manifest we wanted to use.
+
+These approaches are a bit different to each other, and have a common flaw in the shape of having to specify a manifest 
+file one wants to work with, when the package needs to be specified. This may seem quite quirky, as we probably could 
+assume that we're always in some workspace/package.
+
+## Proposed changes
+We could streamline the way we interact with scarb-related subcommands (declare, script at the moment of writing this)
+to look like this:
+
+```bash
+$ # this is the same module
+$ sncast script package_namea::path::to::script_module
+$ sncast script to::script_module
+$ sncast script script_module
+
+$ # this is the same contract
+$ sncast declare package_namea::path::to::contract
+$ sncast to::contract
+$ sncast script script_module
+```
+
+In case there is only one script module/contract with specified name in workspace, only this name can be used to target it.
+Otherwise, if it is ambiguous in the context of workspace, we should show all matches with full paths, to let user decide
+which one to use.
+
+This change would mean that from now on we will have unified interface for declare/script, always assume we're in some workspace/package, 
+the `--path-to-scarb-toml` flag can be removed.
+
+## Implementation
+As much things as possible should be imported from scarb-api crate. The mechanism for path matching could use scarb's 
+package filter mechanism underneath, but probably shouldn't be exposed to the cast interface.
--- a/design_documents/sncast_interface_overhaul.md
+++ b/design_documents/sncast_interface_overhaul.md
@@ -35,16 +35,16 @@ $ sncast declare to::contract
 $ sncast declare contract
 
 $ # contract is ambiguous - there are multiple contracts named 'contract' in multiple packages
-$ sncast --package packagea declare contract
+$ sncast declare --package packagea contract
 ```
 
 In case there is only one script module/contract with specified name in workspace, only this name can be used to target it.
 Otherwise, if it is ambiguous in the context of workspace, we should show all matches with full paths (along with package names),
-to let user decide which one to use. The `--package` flag would then be required, and would indicate which package to use.
+to let user decide which one to use.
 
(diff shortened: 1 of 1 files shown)


design_documents/sncast_interface_overhaul.md
```diff
@@ -0,0 +1,47 @@
...
+Currently, when declaring a contract we require user to pass the contract name using `--contract-name` flag. This flag only
+takes the contract's name, so for example it is not possible to specify the package this contract was defined in (only
+achievable when passing `--path-to-scarb-toml`). 
+
+When running a script we require a positional argument, that is the script package name. Similarly to declare, we could
+pass `--path-to-scarb-toml` if we wanted to pick the manifest we wanted to use.
+
+These approaches are a bit different to each other, and have a common flaw in the shape of having to specify a manifest 
+file one wants to work with, when the package needs to be specified. This may seem quite quirky, as we probably could 
+assume that we're always in some workspace/package.
```
piotmag769, line 20: > as we probably could assume that we're always in some workspace/package.

I think we MUST assume that we are in a workspace (single package is a workspace too) for both`declare` and `script`

design_documents/sncast_interface_overhaul.md
```diff
@@ -0,0 +1,47 @@
...
+$ sncast script script_module
+
+$ # this is the same contract
+$ sncast declare package_namea::path::to::contract
+$ sncast to::contract
+$ sncast script script_module
+```
+
+In case there is only one script module/contract with specified name in workspace, only this name can be used to target it.
+Otherwise, if it is ambiguous in the context of workspace, we should show all matches with full paths, to let user decide
```
piotmag769, line 39: Isn't it better to explicitly provide the `--package` flag and then specify the path as if it were from the source root of the package?
```
$ sncast script --package my_package first_mod::second_mod::script_mod
```
This way we always know where to look for, there is no ambiguity allowed in the first place

piotmag769 (changes requested):
1. I disagree with ditching the `--package` flag for declare and script, I think it should be required. A user should always explicitly specify a package to look for a script/contract in. The only case when it could be omitted is for a workspace with a single member.
2. The approach with paths is impossible to implement with `sncast declare` (see the comment below). We should instead pass the name of the contract.
```
$ sncast declare --package my_package MyContract
``` 
The only case this is ambiguous is when there are 2 contracts with the same name in one package. We need to sync with Scarb team to somehow resolve it (would be good for `snforge` too). cc @maciektr

THenry14:
> 1. I disagree with ditching the `--package` flag for declare and script, I think it should be required. A user should always explicitly specify a package to look for a script/contract in to avoid ambiguity. The only case when it could be omitted is for a workspace with a single member.

We do not support it now, so there's no ditching. At the moment we require users to pass a path to correct manifest. Not sure users should always specify the package, I don't think this is necessary to be honest. If we have only one contract called "whatever" in all the packages in all workspaces, why shouldn't we be able to just do `sncast declare whatever`? And if the contract name isn't unique across the packages, why shouldn't we only then require users to specify (eg using path I proposed in this doc)?

>     2. The approach with paths is impossible to implement with `sncast declare` (see the comment below). We should instead pass the name of the contract.

why?

design_documents/sncast_interface_overhaul.md
```diff
@@ -0,0 +1,47 @@
...
+```bash
+$ # this is the same module
+$ sncast script package_namea::path::to::script_module
+$ sncast script to::script_module
+$ sncast script script_module
+
+$ # this is the same contract
+$ sncast declare package_namea::path::to::contract
+$ sncast to::contract
+$ sncast script script_module
```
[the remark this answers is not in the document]
THenry14: yes, sorry - sloppy copy, fixed.

Yeah, I understand the `path::to` part might be confusing in the context of declare - in this case it will probably always be `package_name::contract_name` or just `contract_name`.

design_documents/sncast_interface_overhaul.md
```diff
@@ -0,0 +1,47 @@
...
+```
+
+In case there is only one script module/contract with specified name in workspace, only this name can be used to target it.
+Otherwise, if it is ambiguous in the context of workspace, we should show all matches with full paths, to let user decide
+which one to use.
+
+This change would mean that from now on we will have unified interface for declare/script, always assume we're in some workspace/package, 
+the `--path-to-scarb-toml` flag can be removed.
```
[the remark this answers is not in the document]
THenry14: I just don't think it will be required to have it. We can still support the env var, but I would be inclined to remove `--path-to-scarb-toml` as we'll require to always be in some workspace. Main reason to add this flag was that that was not always the case.

design_documents/sncast_interface_overhaul.md
THenry14: What you suggested is basically the same idea to what's in this design doc, only that I move the package to the script path and make it non required (if fiven mod is unique). I'm not sure how is it better to have additional, required flag tbh.

design_documents/sncast_interface_overhaul.md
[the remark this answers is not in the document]
piotmag769: Okay, didn't know we want to remove it from all commands

design_documents/sncast_interface_overhaul.md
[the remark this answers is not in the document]
piotmag769: I thin it's confusing to have `package_name::contract_name` when it is not an actual cairo path, at least it is for me. I think it's clearer with `--package` flag

design_documents/sncast_interface_overhaul.md
piotmag769: Okay, I had a wrong assumption that we want to support workspaces for other commands, mb. Tbh I prefer this `--package` arg to unify the interface with my vision of `sncast declare` (see above)

piotmag769:
>We do not support it now, so there's no ditching

Mb, wrong assumption.

> If we have only one contract called "whatever" in all the packages in all workspaces, why shouldn't we be able to just do sncast declare whatever

Okay, you convinced me. I think the `--package` flag should be optional

>>The approach with paths is impossible to implement with sncast declare (see the comment below). We should instead pass the name of the contract.

>why?

Didn't understand your idea with paths for `declare`, I think it is confusing to have sth looking like a cairo path that may actually not be a valid cairo path :/ That's why I would rather have the optional `--package` flag

piotmag769:
Not sure if compilation optimisations are that relevant for `sncast`, but the optional `--package` flag would make it easy to compile only what we need. If user passes it, it is obvious which package we should build with `scarb build --package $PKG`, in case `--package` is not passed we build for the entire workspace with `scarb build --workspace`. As I said though, it may be not that important for `sncast`

design_documents/sncast_interface_overhaul.md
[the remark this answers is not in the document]
THenry14: I also like `--package`, but only if it were to be optional. @MaksymilianDemitraszek ? I remember you had stronger opinion on thsi flag

design_documents/sncast_interface_overhaul.md
```diff
@@ -0,0 +1,47 @@
...
+```bash
+$ # this is the same module
+$ sncast script package_namea::path::to::script_module
+$ sncast script to::script_module
+$ sncast script script_module
+
+$ # this is the same contract
+$ sncast declare package_namea::path::to::contract
+$ sncast declare to::contract
+$ sncast declare script_module
```
MaksymilianDemitraszek, line 35: This one is probably not the same contract, right?

design_documents/sncast_interface_overhaul.md
MaksymilianDemitraszek: I think @piotmag769 is right here, it would be better to move package name under the package flag. It would result in interface consistent with scarb, cargo and snforge.

In rust when you want to run one test you specify
```
 cargo test --package forge --test main -- integration::env::read_short_string --exact --nocapture 
```

So if there is no flag we should search for path in all packages. If there is a flag we should search for it in specified packages. If there are two results we should throw an ambiguity error.

design_documents/sncast_interface_overhaul.md
```diff
@@ -0,0 +1,50 @@
...
+In case there is only one script module/contract with specified name in workspace, only this name can be used to target it.
+Otherwise, if it is ambiguous in the context of workspace, we should show all matches with full paths (along with package names),
+to let user decide which one to use. The `--package` flag would then be required, and would indicate which package to use.
+
+This change would mean that from now on we will have unified interface for declare/script, always assume we're in some workspace/package, 
+the `--path-to-scarb-toml` flag can be removed.
+
+## Implementation
+As much things as possible should be imported from scarb-api crate. The mechanism for path matching could use scarb's 
+package filter mechanism underneath, but probably shouldn't be exposed to the cast interface.
```
piotmag769, line 50: Path matching has nothing to do with scarb's filtering really, more with the format of future scarb artifacts

design_documents/sncast_interface_overhaul.md
```diff
@@ -0,0 +1,50 @@
...
+$ sncast declare to::contract
+$ sncast declare contract
+
+$ # contract is ambiguous - there are multiple contracts named 'contract' in multiple packages
+$ sncast --package packagea declare contract
+```
+
+In case there is only one script module/contract with specified name in workspace, only this name can be used to target it.
+Otherwise, if it is ambiguous in the context of workspace, we should show all matches with full paths (along with package names),
+to let user decide which one to use. The `--package` flag would then be required, and would indicate which package to use.
```
piotmag769, line 43: Kinda complicated way to describe it, I would just say that in case of ambiguity we should show all matches with full paths etc.. Ambiguity doesn't imply that --package is required e.g. you can have two modules with the same name in the same package

piotmag769 (approved):

MaksymilianDemitraszek:
@Arcticae can you please skim it too see if there are any inconsistencies with snforge interface?

MaksymilianDemitraszek (approved):
Overall LGTM.

One problem I see is changing the interface in `sncast declare` will also change the the way it behaves in `script->declare` subcommand. The problem is the fact that this interface probably should be consistent with `snforge` declare. Also if forge investigates some native solution would be nice to also investigate that in `sncast`. @Arcticae @THenry14 can you please sync on this and make sure it is taken under account 🙏 .

MaksymilianDemitraszek:
Another question what is the matching algorithm of the module path? IMO should be specified.

PR #1496, opened 2024-01-05
`Display` for trace

piotmag769:
We have to wait with merging for 2.5.0, some of the `Debug` impls are not present in corelib in 2.4.0

## Introduced changes

- implemented `Display` trait for `CallEntryPoint` and `Array<CallEntryPoint>`

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests

### DIFF
--- a/crates/forge/tests/data/trace/tests/test_trace.cairo
+++ b/crates/forge/tests/data/trace/tests/test_trace.cairo
@@ -1,0 +1,13 @@
+use snforge_std::{declare, ContractClassTrait, PrintTrait};
+use snforge_std::trace::{get_last_call_trace, DisplayArrayCallEntryPoint};
+
+use trace_info::{ISimpleContractDispatcherTrait, ISimpleContractDispatcher};
+
+#[test]
+fn test_trace_print() {
+    let contract_address = declare('SimpleContract').deploy(@array![]).unwrap();
+
+    ISimpleContractDispatcher { contract_address }.simple_call(10);
+
+    println!("{}", get_last_call_trace());
+}
--- a/crates/forge/tests/e2e/trace.rs
+++ b/crates/forge/tests/e2e/trace.rs
@@ -27,7 +27,7 @@ fn trace_info_print() {
         Caller address: 469394814521890341860918960550914
         Call type: Call
         
-        [PASS] tests::test_trace::test_trace_print (gas: ~2461)
+        [PASS] tests::test_trace::test_trace_print (gas: ~[..])
         Tests: 1 passed, 0 failed, 0 skipped, 0 ignored, 0 filtered out
         "}
     );
--- a/crates/forge/tests/e2e/trace.rs
+++ b/crates/forge/tests/e2e/trace.rs
@@ -1,0 +1,34 @@
+use crate::assert_stdout_contains;
+use crate::e2e::common::runner::{setup_package, test_runner};
+use indoc::indoc;
+
+#[test]
+fn trace_info_print() {
+    let temp = setup_package("trace");
+    let snapbox = test_runner();
+
+    let output = snapbox.current_dir(&temp).assert().success();
+
+    assert_stdout_contains!(
+        output,
+        indoc! {r"
+        [..]Compiling[..]
+        [..]Finished[..]
+        
+        
+        Collected 1 test(s) from trace_info package
+        Running 0 test(s) from src/
+        Running 1 test(s) from tests/
+        
+        Entry point type: External
+        Selector: 1024936690255032842875463015856695803908824911639649964560344728797058303195
+        Calldata: [10]
+        Storage address: 2455907345917140204161489130051965975039431451193458942118883935209731300862
+        Caller address: 469394814521890341860918960550914
+        Call type: Call
+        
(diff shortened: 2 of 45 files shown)


crates/forge/tests/data/trace/tests/test_trace.cairo
```diff
@@ -0,0 +1,13 @@
+use snforge_std::{declare, ContractClassTrait, PrintTrait};
+use snforge_std::trace::{get_last_call_trace, DisplayArrayCallEntryPoint};
```
piotmag769, line 2: This trait impl import is needed bcs sth in cairo is not right when it comes to some Display imports and print macros, see https://github.com/starkware-libs/cairo/pull/4745#pullrequestreview-1805963713

snforge_std/src/trace.cairo
```diff
@@ -28,3 +29,59 @@ fn get_last_call_trace() -> Array<CallEntryPoint> {
...
+}
+
+impl DisplayCallEntryPoint of Display<CallEntryPoint> {
+    fn fmt(self: @CallEntryPoint, ref f: Formatter) -> Result<(), Error> {
+        write!(f, "Entry point type: ")?;
+        Display::fmt(self.entry_point_type, ref f)?;
+        write!(f, "\nSelector: ")?;
+        Display::fmt(self.entry_point_selector, ref f)?;
+        write!(f, "\nCalldata: ")?;
+        Debug::fmt(self.calldata, ref f)?;
```
war-in, line 63: Why are some of them `Debug` and others `Display`? 🤔

snforge_std/src/trace.cairo
```diff
@@ -28,3 +29,59 @@ fn get_last_call_trace() -> Array<CallEntryPoint> {
...
+impl DisplayArrayCallEntryPoint of Display<Array<CallEntryPoint>> {
+    fn fmt(self: @Array<CallEntryPoint>, ref f: Formatter) -> Result<(), Error> {
+        let mut i = 0;
+        loop {
+            if i == self.len() {
+                break;
+            }
+            write!(f, "\n").unwrap();
+            Display::fmt(self[i], ref f).unwrap();
+            write!(f, "\n").unwrap();
```
war-in, line 82: Do we want to have two blank lines between displayed calls?

crates/forge/tests/data/trace/tests/test_trace.cairo
```diff
@@ -0,0 +1,13 @@
...
+use trace_info::{ISimpleContractDispatcherTrait, ISimpleContractDispatcher};
+
+#[test]
+fn test_trace_print() {
+    let contract_address = declare('SimpleContract').deploy(@array![]).unwrap();
+
+    ISimpleContractDispatcher { contract_address }.simple_call(10);
+
+    println!("{}", get_last_call_trace());
+}
```
war-in, line 13: Would be nice to have array printing test

crates/forge/tests/e2e/trace.rs
```diff
@@ -0,0 +1,34 @@
...
+        Running 1 test(s) from tests/
+        
+        Entry point type: External
+        Selector: 1024936690255032842875463015856695803908824911639649964560344728797058303195
+        Calldata: [10]
+        Storage address: 2455907345917140204161489130051965975039431451193458942118883935209731300862
+        Caller address: 469394814521890341860918960550914
+        Call type: Call
+        
+        [PASS] tests::test_trace::test_trace_print (gas: ~2461)
```
drknzz, line 30: I think gas probably should be left as `[..]`, since it is out of scope of this test to assert it (and if gas calculation changes, this test will fail)

crates/forge/tests/data/trace/tests/test_trace.cairo
piotmag769: It is an array printing test, `get_last_call_trace` returns an array

drknzz: @war-in probably meant printing multiple calls from the latest trace

snforge_std/src/trace.cairo
piotmag769: `Debug` is already implemented in corelib for `ContractAddress` and `Array<T, +Debug<T>>` (so for `Array<felt252>` too), those prints look pretty nice so I reused them

crates/forge/tests/data/trace/tests/test_trace.cairo
piotmag769: I don't see how this would test anything more than this tbh

war-in: Because if something is broken when printing an array we will catch that

snforge_std/src/trace.cairo
war-in: Will Display be implemented for those with Debug? Would be nice to use only one trait

crates/forge/tests/data/trace/tests/test_trace.cairo
piotmag769: But this prints an `Array`, it's just that it has one element. But okay, will change it xD

piotmag769:
Irrelevant, the struct changed

PR #1500, opened 2024-01-08
1336 set proper nonce

tkumor3:
Closes #

## Introduced changes

Change the way of getting nonce. Instead of obtaining it from the latest block, it gets it from the pending. 
map_script test from `crates/sncast/tests/e2e/script/general.rs` cover this changes
## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/sncast/tests/data/scripts/script_manual_tests/src/test_declare_after_invoke.cairo
```diff
@@ -0,0 +1,45 @@
+use sncast_std::{
+    declare, deploy, invoke, call, DeclareResult, DeployResult, InvokeResult, CallResult, get_nonce,
+};
+use debug::PrintTrait;
+use starknet::{ContractAddress};
+fn main() {
+    let max_fee = 99999999999999999;
+
+    let contract_address: ContractAddress =
+        0x18abd0648d901555975edde1330cffcf37c1da1edbf20b889ba1101e70a6c22
```
THenry14, line 10: if possible, let's avoid hardcoding addresses in scripts, as they might change after cairo update and this can become very troublesome to adjust

crates/sncast/tests/data/scripts/script_manual_tests/Scarb.toml
```diff
@@ -0,0 +1,24 @@
...
+name = "test_script"
+version = "0.1.0"
+
+# See more keys and their definitions at https://docs.swmansion.com/scarb/docs/reference/manifest.html
+
+[dependencies]
+starknet = ">=2.3.0"
+sncast_std = { git = "https://github.com/foundry-rs/starknet-foundry.git", tag = "v0.13.1" }
+map = { path = "../../contracts/map" }
+constructor_with_params = { path = "../../contracts/constructor_with_params" }
```
THenry14, line 11: if you intend to run it on devnet in some automated tests (which I strongly suggest), please do not use those contracts directly as it may break things. We declare those in out test codebase, and one contract cannot be declared multiple times. It is the best practice to add custom contracts for your new test script, that only this script uses.

crates/sncast/tests/data/scripts/script_manual_tests/Scarb.toml
```diff
@@ -0,0 +1,24 @@
+[package]
```
THenry14, line 1: not sure we should put `manual tests` in the repo at this stage. It would be better to just make it a regular test that runs on devnet, that you happen to be able to easily run on real network (as it is with map_script at the moment). 

I'm just not sure anyone will remember that something like this exists somewhere in our tests files, as it is not mentioned nor used aywhere.

crates/sncast/tests/data/scripts/script_manual_tests/src/lib.cairo
```diff
@@ -0,0 +1,3 @@
+mod test_deploy;
+mod test_declare_after_invoke;
+mod test_two_undeployed_declare;
```
THenry14, line 3: can you share specific scenarios where flows shouldn/shouldn't work? What I mean is, we should think about how to write tests that show us that something goes wrong (if possible on devnet), these all seems like happy cases to me (and most of them are probably already covered in our existing test cases, right?)

crates/sncast/tests/data/scripts/script_manual_tests/src/test_declare_after_invoke.cairo
```diff
@@ -0,0 +1,45 @@
...
+        .try_into()
+        .expect('Invalid class hash value');
+
+    let invoke_result = invoke(
+        contract_address, 'put', array![0x11, 0x2], Option::Some(max_fee), Option::None
+    );
+
+    let invoke_nonce = get_nonce('pending');
+    invoke_nonce.print();
```
THenry14, line 19: I get the intent, though if we were to run those also on devnet I'd remove unnecessary prints in favor of some descriptive comments maybe? For someone who wants to run those on integration/testnet good comment will be a better entry to understand what is happening and what is expected. wdyt?

crates/sncast/tests/data/scripts/script_manual_tests/src/test_declare_after_invoke.cairo
tkumor3: I'm wondering if should we leave these scripts in repo? It was added to show what I did.

crates/sncast/tests/data/scripts/script_manual_tests/src/lib.cairo
tkumor3: yep, all of them are heppy cases. No idea what could failure

crates/sncast/src/starknet_commands/deploy.rs
```diff
@@ -55,20 +55,20 @@ pub async fn deploy(
     let salt = extract_or_generate_salt(salt);
 
     let factory = ContractFactory::new(class_hash, account);
-    let deployment = factory.deploy(constructor_calldata.clone(), salt, unique);
+    let execution = factory.deploy(constructor_calldata.clone(), salt, unique);
```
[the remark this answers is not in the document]
tkumor3: IMO it's cleaner if this var doesn't change name after extending with a new attribute

THenry14 (changes requested):

crates/sncast/src/lib.rs
```diff
@@ -168,11 +168,12 @@ pub async fn get_account<'a>(
     keystore: Option<Utf8PathBuf>,
 ) -> Result<SingleOwnerAccount<&'a JsonRpcClient<HttpTransport>, LocalWallet>> {
     let chain_id = get_chain_id(provider).await?;
-    let account = if let Some(keystore) = keystore {
+    let mut account = if let Some(keystore) = keystore {
         get_account_from_keystore(provider, chain_id, &keystore, account)?
     } else {
         get_account_from_accounts_file(account, accounts_file, provider, chain_id)?
     };
+    let account = account.set_block_id(get_block_id("pending")?).clone();
```
ddoktorski, line 176: Can you clarify the reasoning behind setting the block id here, rather than using a default value for `"pending"` within the `get_nonce` function?

tkumor3: We want to use a pending block as default. There is also an issue that will add a flag that allows one to choose which block will be used. IMO is a better way of using starknet-rs which is under sncast. For example, if you set nonce manually (different than what is on the block), sth might stop working. e.g. estimated fee doesn't work.

crates/sncast/tests/e2e/main_tests.rs
```diff
@@ -247,6 +247,7 @@ async fn test_keystore_undeployed_account() {
 }
 
 #[tokio::test]
+#[ignore = "Account should be deployed before run this test"]
```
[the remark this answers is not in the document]
tkumor3: These PRs seem to be independent. I have no idea which one will be merged first, I will resolve the conflict after the merge first, is it ok?

CHANGELOG.md
```diff
@@ -29,6 +35,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Cairo edition in `snforge_std` and `sncast_std` set to `2023_10`
 - `snforge_std::signature` module with `stark_curve`, `secp256k1_curve` and `secp256r1_curve` submodules
 
+### Cast
+
+#### Changed
+
+- get nonce for transaction from pending block instead of latest
+
```
THenry14, line 43: This should be removed

crates/sncast/src/lib.rs
THenry14, line 176: isn't it possible to just 
```
account.set_block_id(...)
```
without reassignment?

crates/sncast/tests/e2e/main_tests.rs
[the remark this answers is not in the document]
THenry14: sure, just let's remember to do this

docs/src/starknet/script.md
```diff
@@ -13,6 +13,18 @@ have a `main` function in the module you want to run. `cast_std` docs can be fou
 Please note that **`sncast script` is in development**. While it is already possible to declare, deploy, invoke and call
 contracts from within Cairo, its interface, internals and feature set can change rapidly each version.
 
+> ⚠️⚠️ By default, the nonce for each transaction is being taken from the pending block ⚠️⚠️
+>
+> Some RPC nodes can be configured with higher poll itervals, which means they may return "older" nonces
+> in pending blocks, or even not be able to obtain pending blocks at all. When running scripts you see
+> error, you may need to manually set nonce and max_fee for transactions.
```
THenry14, line 20:
```suggestion
> in pending blocks, or even not be able to obtain pending blocks at all. When running a script you see an
> error, you may need to manually set both nonce and max_fee for transactions.
```

Let's also state what's the error that might indicate this problem (wrong nonce afaik, right?)

crates/sncast/src/lib.rs
tkumor3: Zrzut ekranu 2024-01-19 o 11 25 39

ddoktorski: Thanks for the clarification. What do you think about modifying the `get_nonce` function to make `block_id` an optional parameter? If it's not provided we can default to use `pending`. I suppose in the majority of cases we want to get nonce for the pending block. This way calling `get_nonce()` instead of `get_nonce('pending')` would be more convenient.

THenry14: This just means you need to return an account, and not a refference to it. you can derefference or clone, should be fine.

docs/src/starknet/script.md
THenry14: let's state the error that might suggest this is the case

crates/sncast/tests/e2e/main_tests.rs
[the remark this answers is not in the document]
tkumor3: DONE ✅

crates/sncast/src/lib.rs
tkumor3: In Rust optional works a little bit differently, if I refactor it to use it will be `get_nonce(Opion::None) | get_nonce(Option::Some('latest')).

tkumor3: I can't derefference because Account does not implement the trait `Copy`

docs/src/starknet/script.md
tkumor3: So I need a node without the pending block, to check what error will be returned

crates/sncast/src/lib.rs
THenry14: I see you've managed to do this :+1:

docs/src/starknet/script.md
THenry14: That would be ideal, but I know it may be hard to do + it does not mean that this will be the error users see. That's why I'd suggest to generalise - state that if you see errors like XYZ, this might be the case that you are facing this problem. sounds good?

crates/sncast/src/lib.rs
ddoktorski: I see, if finding some kind of a workaround requires too much effort or is not possible, let's leave it as it is.

THenry14 (approved):

PR #1501, opened 2024-01-08
Added check_lib_version utility to scarb api

MaksymilianDemitraszek:
**Stack**:
- #1510
- #1508
- #1501 ⬅

### DIFF
--- a/crates/scarb-api/Cargo.toml
+++ b/crates/scarb-api/Cargo.toml
@@ -14,4 +14,5 @@ serde.workspace = true
 serde_json.workspace = true
 thiserror.workspace = true
 which.workspace = true
+semver = "1.0.0"
 sierra-casm = { path="../sierra-casm" }
--- a/crates/scarb-api/src/lib.rs
+++ b/crates/scarb-api/src/lib.rs
@@ -1,6 +1,7 @@
 use anyhow::{anyhow, Context, Result};
 use camino::{Utf8Path, Utf8PathBuf};
 use scarb_metadata::{CompilationUnitMetadata, Metadata, PackageId};
+use semver::VersionReq;
 use serde::Deserialize;
 use serde_json::Value;
 use std::collections::HashMap;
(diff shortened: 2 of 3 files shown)


MaksymilianDemitraszek:
Will be later used in forge & cast

crates/scarb-api/src/lib.rs
```diff
@@ -191,6 +192,24 @@ pub fn name_for_package(metadata: &Metadata, package: &PackageId) -> Result<Stri
     Ok(package.name.clone())
 }
 
+/// Checks if all packages have a version compatible with the specified requirement
```
piotmag769, line 195:
```suggestion
/// Checks if a package have a version compatible with the specified requirement
```

crates/scarb-api/src/lib.rs
```diff
@@ -191,6 +192,24 @@ pub fn name_for_package(metadata: &Metadata, package: &PackageId) -> Result<Stri
...
+pub fn is_required_package_version(
```
piotmag769, line 196:
```suggestion
pub fn is_package_compatible(
```

crates/scarb-api/src/lib.rs
```diff
@@ -280,6 +299,48 @@ mod tests {
...
+        manifest_path
+            .write_str(&formatdoc!(
+                r#"
+                [package]
+                name = "version_checker"
+                version = "0.1.0"
+
+                [[target.starknet-contract]]
+                sierra = true
+                casm = true
```
piotmag769, line 316: [nit] - don't need that since we use our sierra-casm compiler

piotmag769 (approved):

crates/scarb-api/Cargo.toml
```diff
@@ -14,4 +14,5 @@ serde.workspace = true
 serde_json.workspace = true
 thiserror.workspace = true
 which.workspace = true
+semver = "1.0.0"
```
piotmag769, line 17: We should add that to a top lvl Cargo.toml and inherit it here instead

MaksymilianDemitraszek: It is moved there in next pr, here we use it only here.

crates/scarb-api/src/lib.rs
MaksymilianDemitraszek: Done

crates/scarb-api/src/lib.rs
MaksymilianDemitraszek: Done

PR #1502, opened 2024-01-08
Migrate to `universal-sierra-compiler` binary

### LINKED ISSUE #1434: Plug in `universal-sierra-compiler` binary
We should use [universal-sierra-compiler](https://github.com/software-mansion/universal-sierra-compiler/) binary to compile sierra exclusively (this should include removal of the internal crate).
The binary should be upgradable from the command line, pulled from either crates.io or github and swappable without switching the version of starknet-foundry. 
Starknet-Foundry should be shipped/installed with one of those binaries as well (the most recent one).

war-in:
Closes #1434

## Introduced changes

- `sierra-casm` crate was deleted
- `scarb-api` and `forking` module are using `universal-sierra-compiler` to compile sierra 

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/cheatnet/src/forking/state.rs
```diff
@@ -248,3 +252,25 @@ impl StateReader for ForkStateReader {
...
+        .args(vec![
+            "--sierra-input-path",
+            temp_sierra_file.path().to_str().unwrap(),
+        ])
+        .output()
+        .context("Error while compiling Sierra of forked contract")
+        .unwrap()
+        .stdout;
+    serde_json::from_slice(&casm).unwrap()
+}
```
Arcticae, line 276: Isn't this something that we'll want to reuse later on?

Arcticae (approved):
LGTM

crates/cheatnet/src/forking/state.rs
war-in: Do you mean that we should cache that?

crates/scarb-api/src/lib.rs
```diff
@@ -51,7 +51,19 @@ impl StarknetContractArtifacts {
...
-            None => String::new(),
+            None => {
+                let casm = Command::new("universal-sierra-compiler")
+                    .current_dir(base_path)
+                    .args(vec![
+                        "--sierra-input-path",
+                        starknet_contract.artifacts.sierra.as_str(),
+                    ])
+                    .output()
+                    .unwrap()
```
piotmag769, line 62: Maybe `context` informing what has happened and encouraging to:
1. Make sure you have the latest universal compiler binary installed (e.g. sierra 1.5 is coming out soon, it won't work with previous compiler binaries).
2. Contact us if it doesn't help (Unexpected Protostar error vibes xD), since this mean for 99% that we have fucked up and need to fix it asap

piotmag769 (approved):
LGTM, please address the comment

crates/cheatnet/src/forking/state.rs
Arcticae: I mean as for the code, we will want to reuse it so it should be somewhere else

crates/cheatnet/src/forking/state.rs
```diff
@@ -248,3 +252,25 @@ impl StateReader for ForkStateReader {
...
+
+fn generate_casm(sierra_contract_class: &Value) -> CasmContractClass {
+    let mut temp_sierra_file = Builder::new().tempfile().unwrap();
+    let _ = temp_sierra_file
+        .write(
+            serde_json::to_vec(sierra_contract_class)
+                .unwrap()
+                .as_slice(),
+        )
+        .unwrap();
```
drknzz, line 264: At this point, sierra_contract_class is a valid sierra code, right?

drknzz (approved):

crates/cheatnet/src/forking/state.rs
war-in: You're right. We do use the code running `universal-sierra-compiler` here and in the scarb-api.
I wonder what is the best place to put it in? Should I create a crate `universal-sierra-compiler-api`? Or do you maybe have a better idea?

crates/cheatnet/src/forking/state.rs
war-in: Hmm, what do you mean by valid sierra code? 🤔

crates/cheatnet/src/forking/state.rs
war-in: I'm currently working on the `universal-sierra-compiler-api` crate. Will add it in the separate PR

crates/cheatnet/src/forking/state.rs
drknzz: Basically that none of those `.unwraps()` is going to fail

war-in: The sierra_contract_class is just a `Value` here, so yes. None of those `.unwraps()` should fail

drknzz: It's serialized, so that's why I was concerned 👍

piotmag769 (approved):

Issue #1481, opened 2024-01-03
Validate `snforge_std` & `sncast_std` compatibility when running tests


Arcticae:
We should add some validation and versioning for `snforge_std`, so that we're able to determine which version the project is running on, and which the backend is compatible with. 

Also a warning should be issued whenever the versions of backend and stdlib are different.
The scope also includes bumping the version for `snforge_std` for releases

Related to #1469

PR #1510, opened 2024-01-09
Added sncast_std recommended version check

MaksymilianDemitraszek:
commit-id:5b848aa5


MaksymilianDemitraszek:
Closes: https://github.com/foundry-rs/starknet-foundry/issues/1481

Adds warning to sncast about incompatible sncast_std version when running tests. Does not fail the execution.

crates/sncast/tests/data/scripts/old_sncast_std/scripts/src/map_script.cairo
```diff
@@ -0,0 +1,59 @@
+use sncast_std::{
+    declare, deploy, invoke, call, DeclareResult, DeployResult, InvokeResult, CallResult, get_nonce
+};
+
+fn second_contract() {
```
THenry14, line 5: I'm afraid this can lead to race condition with the other `general` script that does exactly the same, with the same contracts (and two identical contracts cannot be declared - it works fine when your test executes second as you do not check the result, but if your script will ever get executed first I guess the other test will fail, because we check its execution result).

I think contract declaration (and maybe even deployment) is not essential to the test that uses this script, so let's maybe just remove it?

crates/sncast/src/response/print.rs
```diff
@@ -163,6 +163,10 @@ fn apply_numbers_formatting(value: OutputValue, formatting: NumbersFormat) -> Ou
     }
 }
 
+pub fn print_warning(error: &Error) {
+    println!("Warning: {error}");
```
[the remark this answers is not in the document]
MaksymilianDemitraszek: We do this in forge, we can add this here as well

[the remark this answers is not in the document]
MaksymilianDemitraszek: Done

crates/sncast/tests/data/scripts/old_sncast_std/scripts/src/map_script.cairo
MaksymilianDemitraszek: Done

crates/sncast/tests/data/scripts/old_sncast_std/scripts/Scarb.toml
```diff
@@ -0,0 +1,15 @@
...
+starknet = ">=2.3.0"
+sncast_std = { git = "https://github.com/foundry-rs/starknet-foundry.git", tag = "v0.13.1" }
+
+[lib]
+sierra = true
+casm = true
+
+[[target.starknet-contract]]
+sierra = true
+casm = true
```
piotmag769, line 15: newline xD

piotmag769 (approved):

crates/sncast/tests/data/scripts/old_sncast_std/contracts/src/lib.cairo
```diff
@@ -0,0 +1 @@
+fn some_func() {}
```
THenry14, line 1: This is not imported in your script, and thus not used anywhere. I suppose we could get rid of this whole (contract) directory, as it is non essential to the test.

THenry14 (changes requested):

crates/sncast/tests/data/scripts/old_sncast_std/contracts/src/lib.cairo
MaksymilianDemitraszek: Done

THenry14 (approved):

crates/sncast/src/response/print.rs
```diff
@@ -163,6 +164,11 @@ fn apply_numbers_formatting(value: OutputValue, formatting: NumbersFormat) -> Ou
     }
 }
 
+pub fn print_as_warning(error: &Error) {
+    let warning_tag = style("Warning:").color256(11);
+    println!("{warning_tag} {error}");
```
drknzz, line 169:
```suggestion
    let warning_tag = style("Warning").color256(11);
    println!("{warning_tag}: {error}");
```
just a suggestion, the colon wouldn't be colored

drknzz (approved):

PR #1514, opened 2024-01-10
`available_gas` support

### LINKED ISSUE #618: `available_gas` support in forge
We should support the attribute in test collector and enforce it in the runner

war-in:
Closes #618

## Introduced changes

- `available_gas` attribute can be used with snforge now

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/forge-runner/src/running.rs
```diff
@@ -198,20 +198,17 @@ pub fn run_test_case(
     runner_config: &Arc<RunnerConfig>,
     runner_params: &Arc<RunnerParams>,
 ) -> Result<RunResultWithInfo> {
+    // we need to do this "trick" to be compatible with the older scarb versions
     ensure!(
-        case.available_gas.is_none(),
-        "\n    Attribute `available_gas` is not supported\n"
+        case.available_gas != Some(0),
+        "\n    Make sure you use the scarb version where `available_gas` attribute is supported\n"
```
piotmag769, line 204: I would specify the version, it will be supported from 2.5.0 onwards (I assume we want to merge both related PRs before 2.5.0 which will be released next week, right?)

crates/forge-runner/src/running.rs
```diff
@@ -198,20 +198,17 @@ pub fn run_test_case(
...
     );
-    let available_gas = Some(usize::MAX);
 
     let func = runner.find_function(case.name.as_str()).unwrap();
-    let initial_gas = runner
-        .get_initial_available_gas(func, available_gas)
-        .unwrap();
```
piotmag769, line 210: This one is sierra gas, right? Do we need to remove this? It prob changes nothing, but it looks like an automatic calculation of the gas required for the top lvl VM to execute the test (it doesn't care about gas used by blockifier anyways)

crates/forge-runner/src/test_case_summary.rs
```diff
@@ -195,13 +195,26 @@ impl TestCaseSummary<Single> {
...
-                    arguments,
-                    test_statistics: (),
-                    gas_info: gas,
-                },
+                ExpectedTestResult::Success => {
+                    if let Some(available_gas) = test_case.available_gas {
+                        if gas > available_gas as u128 {
+                            return TestCaseSummary::Failed {
+                                name,
+                                msg: Some("\n\tavailable_gas exceeded".to_string()),
```
piotmag769, line 203: "Test cost exceeded the available gas"

crates/forge-runner/src/running.rs
```diff
@@ -324,7 +321,7 @@ fn extract_test_case_summary(
             }
         }
         // `ForkStateReader.get_block_info`, `get_fork_state_reader` may return an error
-        // unsupported `available_gas` attribute may be specified
+        // older scarb version where `available_gas` attribute was not supported may be used
```
piotmag769, line 324:
```suggestion
        // `available_gas` may be specified with Scarb ~2.4
```

piotmag769 (commented):
Do we want to merge before scarb 2.5.0? cc @Arcticae

crates/forge-runner/src/test_case_summary.rs
Arcticae, line 203: I don't think that this constructor is a right place to put this logic

Arcticae: It should be placed in own function and ran elsewhere, after `run_test_case` result is retrieved

crates/forge-runner/src/running.rs
war-in: Yes, this is a sierra gas. Maybe I will revert this change because we may need this one in the future

crates/forge-runner/src/gas.rs
```diff
@@ -106,3 +108,28 @@ fn get_onchain_data_segment_length(state_changes_count: StateChangesCount) -> us
...
+    case: &TestCaseRunnable,
+    summary: TestCaseSummary<Single>,
+) -> TestCaseSummary<Single> {
+    match summary {
+        TestCaseSummary::Passed {
+            name,
+            arguments,
+            gas_info,
+            ..
+        } if case
```
Arcticae, line 122: you could just pass `available_gas` instead of `case`

Arcticae (commented):
Can we add a test with fuzzing so we safeguard ourselves from breaking it?

crates/forge/src/lib.rs
```diff
@@ -105,6 +107,8 @@ pub async fn run(
     let not_filtered: usize = test_crates.iter().map(|tc| tc.test_cases.len()).sum();
     let filtered = all_tests - not_filtered;
 
+    verify_scarb_version_if_available_gas_used(&test_crates)?;
```
piotmag769, line 110: It's too early to fail, we should only fail the specific test case imo (in `run_test_case`)

crates/forge-runner/src/running.rs
```diff
@@ -304,12 +295,15 @@ fn extract_test_case_summary(
...
-                    result_with_info.gas_used,
-                )),
+                Ok(run_result) => {
+                    let summary = TestCaseSummary::from_run_result_and_info(
+                        run_result,
+                        case,
+                        args,
+                        result_with_info.gas_used,
+                    );
+                    Ok(check_available_gas(&case.available_gas, summary))
```
piotmag769, line 305: Imo `check_available_gas` this should be used inside `from_run_result_and_info`, this way we don't leak the logic

crates/forge/src/lib.rs
war-in: Why? Let's assume someone has 100 tests with `available_gas` attribute. He will get 200 error messages with info about the scarb version

crates/forge-runner/src/running.rs
war-in: Summoning @Arcticae

piotmag769: Okay, I see @Arcticae's comment and I just disagree xD Even the name `from_run_result_and_info` suggests that it retrieves the result from run result and *additional info* like gas used. Otherwise it is just confusing - why would you construct the result and the change it randomly?

crates/forge/src/lib.rs
piotmag769: Still there is no reason why the rest of the tests couldn't be run. Also that's how we did it before - we only failed the specific test when the attribute was specified, not everything

crates/forge-runner/src/running.rs
Arcticae: I could go both ways but i always tend to leave business logic out of constructors (just a habit that i have, and it produces clean code most of the times). We could go with a builder as well. But i really do have vietnam flashbacks when it comes to builders.

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `mean` and `std deviation` fields are displayed for gas usage while running fuzzing tests 
 - Cairo edition in `snforge_std` and `sncast_std` set to `2023_10`
 - `snforge_std::signature` module with `stark_curve`, `secp256k1_curve` and `secp256r1_curve` submodules
+- `available_gas` attribute is now supported
```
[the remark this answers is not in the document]
Arcticae: I think this will come out in 2.4.4

crates/forge-runner/src/running.rs
Arcticae: @war-in leaving the decision up to you.

crates/forge/src/lib.rs
```diff
@@ -164,6 +169,21 @@ pub async fn run(
...
 
+fn warn_if_available_gas_used_with_incompatible_scarb_version(
+    test_crates: &Vec<CompiledTestCrateRaw>,
+) {
+    for test_crate in test_crates {
+        for case in &test_crate.test_cases {
+            if case.available_gas == Some(0) && scarb_version() <= Version::new(2, 5, 0) {
+                print_warning(&anyhow!(
+                    "`available_gas` attribute was probably specified when using Scarb ~2.4. \
+                    Make sure to use Scarb >=2.5.0"
```
Arcticae, line 180: This will be doable in 2.4.4 already

crates/forge/src/lib.rs
```diff
@@ -164,6 +169,21 @@ pub async fn run(
     Ok(summaries)
 }
 
+fn warn_if_available_gas_used_with_incompatible_scarb_version(
```
Arcticae, line 172: I don't think it's a suitable place to have this function

crates/forge/src/init.rs
```diff
@@ -106,30 +104,13 @@ pub fn run(project_name: &str) -> Result<()> {
...
-        .expect("Could not find cairo version")
-        .as_str();
-
+    // We can assume Cairo and Scarb versions are the same
+    let scarb_version = scarb_version();
     ScarbCommand::new_with_stdio()
         .current_dir(&project_path)
         .offline()
         .arg("add")
-        .arg(format!("starknet@{cairo_version}"))
```
Arcticae, line 132: `cairo_version`/`scarb_version` - These won't always be equal

crates/forge-runner/src/sierra_casm_runner.rs
```diff
@@ -380,6 +380,7 @@ impl SierraCasmRunner {
 
     /// Returns the initial value for the gas counter.
     /// If `available_gas` is None returns 0.
+    #[allow(dead_code)]
```
Arcticae, line 383: Why is it needed here?

Arcticae (changes requested):

crates/forge/src/init.rs
war-in: Why :o I thought they were connected

Arcticae: they are but there are cases when scarb is released without new cairo version

crates/forge-runner/src/sierra_casm_runner.rs
war-in: I've removed the initial_gas calculation because the current way was not appropriate.
It mixed old and new gas logic

crates/forge/src/lib.rs
war-in: We already have the `warn_if..` function in the `main.rs` because it is used only there. I don't think there is a need to create separate file just for this function

crates/scarb-api/src/version.rs
```diff
@@ -0,0 +1,42 @@
...
+use regex::Regex;
+use semver::Version;
+use std::str::from_utf8;
+
+pub struct ScarbVersionOutput {
+    pub scarb: Version,
+    pub cairo: Version,
+}
+
+pub fn scarb_version_command() -> Result<ScarbVersionOutput> {
```
Arcticae, line 12: Can we add a follow-up issue to refactor this, after the `Metadata` is integrated into the crate as well? 
Namely: i think this should be accessible via the `ScarbCommand` struct from the crate (through a new constructor method or something like that)

Arcticae (approved):

crates/scarb-api/src/version.rs
war-in: I'm not sure because we already can get `scarb --version` output from the command. It's just a utility function and we are not providing any in the command itself

Arcticae: But we probably will soon enough

crates/forge-runner/src/gas.rs
```diff
@@ -106,3 +107,25 @@ fn get_onchain_data_segment_length(state_changes_count: StateChangesCount) -> us
...
+    match summary {
+        TestCaseSummary::Passed {
+            name,
+            arguments,
+            gas_info,
+            ..
+        } if available_gas.map_or(false, |available_gas| gas_info > available_gas as u128) => {
+            TestCaseSummary::Failed {
+                name,
+                msg: Some("\n\tTest cost exceeded the available gas".to_string()),
```
drknzz, line 124: Will the actual gas be still displayed somewhere?

crates/sncast/tests/data/multicall_configs/deploy_invoke_calldata_ids.toml
```diff
@@ -1,19 +1,19 @@
 [[call]]
 call_type = "deploy"
-class_hash = "0x7644be7f58307726aa836e945edede13e9e08c38eaf2186d4d48eca7dd435ac"
+class_hash = "0x51ce0a6fdc37672013380682f1d39871daea3acef92bc0891464aa55c858a45"
```
drknzz, line 3: How were these obtained?

war-in: from some file called `.env... devnet...`
It was consulted with @THenry14

crates/forge-runner/src/gas.rs
war-in: Nope 🤔 I think I should add this

crates/forge/tests/integration/available_gas.rs
```diff
@@ -37,7 +37,7 @@ fn available_gas_exceeded() {
     assert_case_output_contains!(
         result,
         "keccak_cost",
-        "Test cost exceeded the available gas"
+        "Test cost exceeded the available gas. Gas consumed: ~11"
```
drknzz, line 40: [total nit] But I liked **Consumed gas: ~11** more 😄 something about the word `gas` being closer to the value, like in passed tests

drknzz (approved):
:shipit:

piotmag769 (commented):
🆗

PR #1529, opened 2024-01-15
Fix false positive test

tkumor3:
Closes #

## Introduced changes

-

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/sncast/tests/helpers/devnet.rs
```diff
@@ -66,9 +70,48 @@ fn start_devnet() {
         "/constructor_with_params/target/dev/constructor_with_params_ConstructorWithParams",
         "CAST_WITH_CONSTRUCTOR",
     ));
+
+    rt.block_on(deploy_keystore_account());
+
     dotenv::from_filename(DEVNET_ENV_FILE).unwrap();
 }
 
+async fn deploy_keystore_account() {
```
THenry14, line 79: this should be moved to fixtures and imported here

crates/sncast/tests/helpers/devnet.rs
```diff
@@ -66,9 +70,48 @@ fn start_devnet() {
...
+    let keystore_path = "tests/data/keystore/deployed_key.json";
+    let account_path = "tests/data/keystore/deployed_account_copy.json";
+
+    fs::copy("tests/data/keystore/deployed_account.json", account_path).unwrap();
```
THenry14, line 83: could we use tempdir for this?

crates/sncast/tests/helpers/devnet.rs
```diff
@@ -66,9 +70,48 @@ fn start_devnet() {
...
+        account_path,
+        "account",
+        "deploy",
+        "--max-fee",
+        "99999999999999999",
+        "--class-hash",
+        DEVNET_OZ_CLASS_HASH,
+    ];
+
+    Command::new(cargo_bin!("sncast"))
```
THenry14, line 109: I don't really like testing cast with `sncast` binary, we should probably avoid it.
We could just deploy an account using `OpenZeppelinAccountFactory` from the sdk, maybe we could even hold deployed account file in repo to not have to create it from scratch (which would require us to write additional logic just for tests). wdyt?

THenry14 (changes requested):

crates/sncast/tests/helpers/devnet.rs
tkumor3: IMO it's overkill! this method is as regular before test action. Probably we could execute it in this particular test. Using OpenZeppelinAccountFactory will duplicate code from cast. I fought to use existing method but I had problem with imports

crates/sncast/tests/data/keystore/deployed_key.json
```diff
@@ -0,0 +1 @@
+{"crypto":{"cipher":"aes-128-ctr","cipherparams":{"iv":"969dc1d196eccd96518b4945820a7ac0"},"ciphertext":"fd20c0dfa110976bffa70b54321487c4546d0c26816c8383ffcfa1ba099f73b1","kdf":"scrypt","kdfparams":{"dklen":32,"n":8192,"p":1,"r":8,"salt":"177919634a458573cf15ee537f658cca630d32aaae1a3b7cecf86a427ef01ea9"},"mac":"adada9c6d157999ba340dc99c6b6596698904026e3f619d025962d1cbcd2a50e"},"id":"caf9c371-4e46-45c7-b682-a44667a83725","version":3}
```
ddoktorski, line 1: nit: Could you format this file?

crates/sncast/tests/helpers/fixtures.rs
```diff
@@ -69,6 +72,44 @@ pub async fn declare_contract(account: &str, path: &str, shortname: &str) -> Fie
     class_hash
 }
 
+pub async fn deploy_keystore_account() {
+    let keystore_path = "tests/data/keystore/deployed_key.json";
+    let account_path = "tests/data/keystore/deployed_account.json";
```
ddoktorski, line 77: nit: For consistency with other file names in the `keystore` directory, how about renaming:
- `deployed_key.json` -> `my_key_deployed.json`
- `deployed_account` -> `my_account_deployed.json`?

THenry14: Actually, these are two different accounts, so perhaps we should change the naming altogether? `(deployed | undeployed)_(account | key)` should be descriptive enough. wdyt?

crates/sncast/tests/helpers/fixtures.rs
```diff
@@ -69,6 +72,44 @@ pub async fn declare_contract(account: &str, path: &str, shortname: &str) -> Fie
...
+    let private_key =
+        SigningKey::from_keystore(keystore_path, "123").expect("Could not get the private_key");
+
+    let provider = get_provider(URL).expect("Could not get the provider");
+    let chain_id = get_chain_id(&provider).await.expect("msg");
```
THenry14, line 82:
```suggestion
    let chain_id = get_chain_id(&provider).await.expect("Could not get chain_id from provider");
```
the `msg` is not very descriptive

crates/sncast/tests/helpers/fixtures.rs
```diff
@@ -69,6 +72,44 @@ pub async fn declare_contract(account: &str, path: &str, shortname: &str) -> Fie
...
+
+    let contents = std::fs::read_to_string(account_path).expect("Failed to read accounts file");
```
THenry14, line 84:
```suggestion
    let contents = std::fs::read_to_string(account_path).expect("Failed to read keystore account file");
```

keystore account file != accounts file

crates/sncast/tests/helpers/fixtures.rs
```diff
@@ -69,6 +72,44 @@ pub async fn declare_contract(account: &str, path: &str, shortname: &str) -> Fie
...
+    let items: serde_json::Value = serde_json::from_str(&contents)
+        .with_context(|| format!("Failed to parse accounts file at = {account_path}"))
```
THenry14, line 86:
```suggestion
        .with_context(|| format!("Failed to parse keystore account file at = {account_path}"))
```

crates/sncast/tests/helpers/fixtures.rs
```diff
@@ -69,6 +72,44 @@ pub async fn declare_contract(account: &str, path: &str, shortname: &str) -> Fie
...
+        .expect("msg");
```
THenry14, line 87: is this necessary?
+ not very descriptive

crates/sncast/tests/helpers/fixtures.rs
```diff
@@ -69,6 +72,44 @@ pub async fn declare_contract(account: &str, path: &str, shortname: &str) -> Fie
...
+
+    let factory = OpenZeppelinAccountFactory::new(
+        parse_number(DEVNET_OZ_CLASS_HASH).expect("Could not parse DEVNET_OZ_CLASS_HASH"),
+        chain_id,
+        LocalWallet::from_signing_key(private_key),
+        provider,
+    )
+    .await
+    .expect("");
```
THenry14, line 96: Missing expect message

crates/sncast/tests/helpers/fixtures.rs
```diff
@@ -69,6 +72,44 @@ pub async fn declare_contract(account: &str, path: &str, shortname: &str) -> Fie
...
+            .expect("Could not get address"),
+        9_999_999_999_999_999_999,
+    )
+    .await;
+
+    factory
+        .deploy(parse_number("0xa5d90c65b1b1339").expect("Could not parse salt"))
+        .send()
+        .await
+        .expect("Could not deploy");
```
THenry14, line 110:
```suggestion
        .expect("Could not deploy keystore account");
```

crates/sncast/tests/e2e/main_tests.rs
```diff
@@ -269,7 +269,7 @@ async fn test_keystore_declare() {
         .current_dir(contract_path.path())
         .args(args);
 
-    snapbox.assert().success().get_output().stderr.is_empty();
+    assert!(snapbox.assert().success().get_output().stderr.is_empty());
 
     fs::remove_dir_all(contract_path).unwrap();
```
ddoktorski, line 274: Not directly related to this PR, but I think `remove_dir_all` might not be necessary here. Since `contract_path` is a `TempDir`, it should be automatically deleted when it goes out of scope. What do you think?

THenry14: Definitely, good catch

tkumor3: yep, probably someone copied from another test, which write to current_dir

THenry14 (approved):
🔥

PR #1530, opened 2024-01-15
`usc-api` crate

war-in:
## Introduced changes

- `usc-api` was created to help using `universal-sierra-compiler` command

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/usc-api/src/lib.rs
```diff
@@ -0,0 +1,30 @@
...
+        .inherit_stderr()
+        .arg("--sierra-input-path")
+        .arg(sierra_file_name)
+        .command()
+        .output()
+        .context(
+            "Error while compiling Sierra of the contract. \
+            Make sure you have the latest universal-sierra-binary installed. \
+            Contact us if it doesn't help",
+        )
```
drknzz, line 26:
```suggestion
        .context(
            "Error while compiling Sierra of the contract. \
            Make sure you have the latest universal-sierra-binary installed.",
        )
```
It's natural, that if it someone doesn't manage to make it work after the proposed solution, they'd come for support.

war-in: @piotmag769 wdyt?

crates/universal-sierra-compiler-api/src/lib.rs
```diff
@@ -0,0 +1,30 @@
...
+        .arg("--sierra-input-path")
+        .arg(sierra_file_name)
+        .command()
+        .output()
+        .context(
+            "Error while compiling Sierra of the contract. \
+            Make sure you have the latest universal-sierra-binary installed. \
+            Contact us if it doesn't help",
+        )
+        .unwrap();
```
piotmag769, line 27: I don't think we should unwrap, it may be users fault, we should pretty print that - `?` and return `Result` from this function

crates/usc-api/src/lib.rs
piotmag769: Both are fine

Arcticae: I'd leave it out as well

crates/universal-sierra-compiler-api/src/lib.rs
war-in: Who is the `user`?

piotmag769: Someone using snforge, having older version of our compiler installed

piotmag769 (approved):

MaksymilianDemitraszek:
Please rename it to something without acronym, it is hard to understand, `sierra_compiler_api` for example.

war-in:
already done 😅

PR #1549, opened 2024-01-16
Add flag to save call traces to files

piotmag769:
Closes https://github.com/software-mansion/cairo-profiler/issues/3

## Introduced changes

- call traces are saved in `.snfoundry_trace` directory (maybe we should add a flag to clean this directory too?)

## Checklist

- [x] Linked relevant issue
- [ ] Added relevant tests


crates/cheatnet/src/blockifier_structs.rs
```diff
@@ -0,0 +1,62 @@
+// structs copied from blockifier to make CallEntryPoint serializable
```
piotmag769, line 1: #1554

crates/forge-runner/src/running.rs
```diff
@@ -335,3 +341,16 @@ fn get_context<'a>(runtime: &'a ForgeRuntime) -> &'a EntryPointExecutionContext
...
+        serde_json::to_string(&cheatnet_state.call_trace).expect("Failed to serialize call trace");
+
+    let dir_to_save_trace = PathBuf::from(TRACE_DIR);
+    fs::create_dir_all(&dir_to_save_trace).expect("Failed to create a file to save call trace to");
+    fs::write(
+        dir_to_save_trace.join(test_case_name),
+        call_trace_serialized,
+    )
+    .expect("Failed to write call trace to a file");
+}
```
Arcticae, line 356: This should not be in this file

crates/forge/src/main.rs
```diff
@@ -140,6 +145,7 @@ fn combine_configs(
         fuzzer_seed
             .or(forge_config.fuzzer_seed)
             .unwrap_or_else(|| thread_rng().next_u64()),
+        save_call_trace || forge_config.save_call_trace,
```
Arcticae, line 148: I don't think there should be this logical operator here, it might lead to confusion.
Why is the `save_call_trace` a mandatory parameter as well?

crates/cheatnet/src/state.rs
```diff
@@ -140,8 +140,9 @@ pub enum CheatStatus<T> {
 }
 
 /// Tree structure representing trace of a call.
+#[derive(Serialize, Deserialize)]
 pub struct CallTrace {
-    pub entry_point: CallEntryPoint,
+    pub entry_point: crate::blockifier_structs::CallEntryPoint,
```
Arcticae, line 145: Does this have to be in other type? I think it could be converted only on dump request

crates/cheatnet/src/blockifier_structs.rs
```diff
@@ -0,0 +1,62 @@
...
+use serde::{Deserialize, Serialize};
+use starknet_api::core::{ClassHash, ContractAddress, EntryPointSelector};
+use starknet_api::deprecated_contract_class::EntryPointType;
+use starknet_api::transaction::Calldata;
+
+#[derive(Clone, Debug, Default, Eq, PartialEq, Deserialize, Serialize)]
```
Arcticae, line 7: Did you try alternative approaches instead of copying? Wouldn't manual serialization be better instead? Even as a makeshift solution

MaksymilianDemitraszek:
TODO use `impl CallTrace { serialize }`

MaksymilianDemitraszek:
I will be working on this one further

crates/cheatnet/src/blockifier_structs.rs
MaksymilianDemitraszek: I think we will just provide such structures in profiler as a library and forge will be able to use it to meet the standard. It will be much easier than realying on blockifier internals.

crates/forge/src/main.rs
MaksymilianDemitraszek: It is a boolean (flag), defaulted to false.

crates/forge-runner/src/trace_data.rs
```diff
@@ -0,0 +1,107 @@
...
+
+use crate::test_case_summary::{Single, TestCaseSummary};
+
+pub const TRACE_DIR: &str = ".snfoundry_trace";
+
+/// Tree structure representing trace of a call.
+#[derive(Debug, Clone, Serialize)]
+pub struct CallTrace {
+    pub entry_point: CallEntryPoint,
+    pub nested_calls: Vec<CallTrace>,
```
MaksymilianDemitraszek, line 20: It might be memory consuing, later we want to move to id-based graph. There is an issue for it already.

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 #### Added
 
 - `store` and `load` cheatcodes
+- `--save-trace--data` flag to `snforge test` command
```
drknzz, line 15:
```suggestion
- `--save-trace-data` flag to `snforge test` command
```

crates/forge/src/main.rs
```diff
@@ -98,6 +98,10 @@ struct TestArgs {
     /// Run tests that failed during the last run
     #[arg(long)]
     rerun_failed: bool,
+
+    /// Save call traces of all tests
```
drknzz, line 102: All tests or only non-fuzzing passed ones?

MaksymilianDemitraszek: I forgot to update it 👍

crates/forge-runner/src/lib.rs
```diff
@@ -190,6 +191,12 @@ pub async fn run_tests_from_crate(
 
         print_test_result(&result);
 
+        if runner_config.save_trace_data {
+            if let AnyTestCaseSummary::Single(result) = &result {
+                save_trace_data(result);
```
Arcticae, line 196: If user has only fuzzing tests this will fail silently

crates/forge/tests/e2e/trace_data.rs
```diff
@@ -0,0 +1,39 @@
...
+        .code(1);
+
+    assert!(temp
+        .join(TRACE_DIR)
+        .join("simple_package::tests::test_fib.json")
+        .exists());
+    assert!(!temp
+        .join(TRACE_DIR)
+        .join("tests::test_simple::test_failing.json")
+        .exists());
```
Arcticae, line 22: you aren't actually testing any contents, are you?

Arcticae (approved):
Mostly OK, some nitpicks, you can as well correct some things after merging since i don't see a point in blocking this

crates/forge-runner/src/lib.rs
MaksymilianDemitraszek: Specification mentions it is not for fuzz tests

crates/forge/tests/e2e/trace_data.rs
MaksymilianDemitraszek: I feel like semantics of the trace content are integration tested by @piotmag769. I don't feel like explicitly asserting the content here would make a very strong test above what we have. If you have an idea how we can test it i am open to add something here.

crates/forge/src/main.rs
piotmag769: Yup, it is the same as for `exit_first`

crates/forge/tests/e2e/trace_data.rs
piotmag769: I guess we can test content of one of the files - it doesn't test the trace collection mechanism, but we will be sure someone doesn't mess up sth in the saving to file section of the code in the future

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 #### Added
 
 - `store` and `load` cheatcodes
+- `--save-trace-data` flag to `snforge test` command
```
piotmag769, line 15: Would be nice to mention the purpose

crates/forge-runner/src/trace_data.rs
piotmag769: Do we need this struct? We could just derive `Serialize` on the `cheatnet::state::CallTrace` 🤔

MaksymilianDemitraszek: We need this as I simply copied the module to thoth, later we will use it from thoth lib as it will be responsible for standarizing the format.

CHANGELOG.md
MaksymilianDemitraszek: Done

crates/forge/tests/e2e/trace_data.rs
MaksymilianDemitraszek: I've added the file parsing to the test.

PR #1562, opened 2024-01-19
Revert migration to USC binary

Arcticae:
## Introduced changes

- Reverts changes in order to make release without it (breaking changes in USC incoming)


war-in (approved):
😭

Issue #1474, opened 2024-01-03
Extract scarb build from declare subcommand


THenry14:
Declare should not build contracts, everything should be built beforehand.

PR #1565, opened 2024-01-19
Extract build from declare

THenry14:
-> * Extract build from declare
-- * Add support for --package flag (https://github.com/foundry-rs/starknet-foundry/pull/1569)
-- * Add workspace tests (https://github.com/foundry-rs/starknet-foundry/pull/1572)

todo:

## Introduced changes

Closes https://github.com/foundry-rs/starknet-foundry/issues/1474

- Extracts build logic from declare to avoid multiple artifact builds

## Checklist

- [X] Linked relevant issue
- [X] Added relevant tests

### DIFF
--- a/crates/sncast/src/helpers/scarb_utils.rs
+++ b/crates/sncast/src/helpers/scarb_utils.rs
@@ -1,5 +1,4 @@
 use super::constants::{WAIT_RETRY_INTERVAL, WAIT_TIMEOUT};
-use crate::BuildConfig;
 use anyhow::{anyhow, bail, Context, Result};
 use camino::{Utf8Path, Utf8PathBuf};
 use scarb_api::{get_contracts_map, ScarbCommand, StarknetContractArtifacts};
@@ -51,6 +50,11 @@ impl Default for CastConfig {
             wait_retry_interval: WAIT_RETRY_INTERVAL,
         }
     }
+}
+
+pub struct BuildConfig {
+    pub scarb_toml_path: Utf8PathBuf,
+    pub json: bool,
 }
 
 pub trait PropertyFromCastConfig: Sized {
@@ -192,7 +196,7 @@ pub fn get_scarb_metadata_with_deps(
     execute_scarb_metadata_command(&command)
 }
 
-#[must_use]
+#[must_use] //todo :remove
 pub fn verify_or_determine_scarb_manifest_path(
     path_to_scarb_toml: &Option<Utf8PathBuf>,
 ) -> Option<Utf8PathBuf> {
--- a/crates/sncast/src/helpers/scarb_utils.rs
+++ b/crates/sncast/src/helpers/scarb_utils.rs
@@ -1,10 +1,12 @@
 use super::constants::{WAIT_RETRY_INTERVAL, WAIT_TIMEOUT};
+use crate::BuildConfig;
 use anyhow::{anyhow, bail, Context, Result};
 use camino::{Utf8Path, Utf8PathBuf};
-use scarb_api::ScarbCommand;
+use scarb_api::{get_contracts_map, ScarbCommand, StarknetContractArtifacts};
 use scarb_metadata;
 use serde::{Deserialize, Serialize};
 use serde_json::Value;
+use std::collections::HashMap;
 use std::default::Default;
 use std::env;
 use std::fs::canonicalize;
@@ -274,6 +276,23 @@ pub fn get_package_tool_sncast(metadata: &scarb_metadata::Metadata) -> Result<&V
         .ok_or_else(|| anyhow!("No field [tool.sncast] found in package"))?;
 
     Ok(tool_sncast)
+}
+
+pub fn build(config: &BuildConfig) -> Result<HashMap<String, StarknetContractArtifacts>> {
+    let mut cmd = ScarbCommand::new_with_stdio();
+    cmd.arg("build").manifest_path(&config.scarb_toml_path);
+    if config.json {
+        cmd.json();
+    }
+    cmd.run()
+        .map_err(|e| anyhow!(format!("Failed to build using scarb; {e}")))?;
+
+    let metadata = get_scarb_metadata_command(&config.scarb_toml_path)?
+        .exec()
+        .expect("Failed to obtain metadata");
+    let package = get_package_metadata(&metadata, &config.scarb_toml_path)
+        .with_context(|| anyhow!("Failed to find package"))?;
+    get_contracts_map(&metadata, &package.id)
 }
 
 #[cfg(test)]
--- a/crates/sncast/src/helpers/scarb_utils.rs
+++ b/crates/sncast/src/helpers/scarb_utils.rs
@@ -196,7 +196,7 @@ pub fn get_scarb_metadata_with_deps(
     execute_scarb_metadata_command(&command)
 }
 
-#[must_use] //todo :remove
+#[must_use]
 pub fn verify_or_determine_scarb_manifest_path(
     path_to_scarb_toml: &Option<Utf8PathBuf>,
 ) -> Option<Utf8PathBuf> {
--- a/crates/sncast/src/helpers/scarb_utils.rs
+++ b/crates/sncast/src/helpers/scarb_utils.rs
@@ -52,11 +52,6 @@ impl Default for CastConfig {
     }
 }
 
-pub struct BuildConfig {
-    pub scarb_toml_path: Utf8PathBuf,
-    pub json: bool,
-}
-
 pub trait PropertyFromCastConfig: Sized {
     fn from_toml_value(value: &Value) -> Option<Self>;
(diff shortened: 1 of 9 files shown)


crates/sncast/src/helpers/scarb_utils.rs
```diff
@@ -51,6 +52,11 @@ impl Default for CastConfig {
     }
 }
 
+pub struct BuildConfig {
+    pub scarb_toml_path: Utf8PathBuf,
+    pub json: bool,
+}
+
```
MaksymilianDemitraszek, line 59: [[Nitpick]] can we move this struct above the only function that uses it?

crates/sncast/src/main.rs
```diff
@@ -122,9 +124,22 @@ fn main() -> Result<()> {
     let runtime = Runtime::new().expect("Failed to instantiate Runtime");
 
     if let Commands::Script(script) = cli.command {
+        let manifest_path = match cli.path_to_scarb_toml {
+            Some(path) => path,
+            None => get_scarb_manifest().context("Failed to obtain manifest path from Scarb")?,
+        };
```
MaksymilianDemitraszek, line 130: Could we create a function `load_manifest`? it would ensure error is always the same.

crates/sncast/src/helpers/scarb_utils.rs
```diff
@@ -276,6 +282,23 @@ pub fn get_package_tool_sncast(metadata: &scarb_metadata::Metadata) -> Result<&V
...
+    }
+    cmd.run()
+        .map_err(|e| anyhow!(format!("Failed to build using scarb; {e}")))?;
+
+    let metadata = get_scarb_metadata_command(&config.scarb_toml_path)?
+        .exec()
+        .expect("Failed to obtain metadata");
+    let package = get_package_metadata(&metadata, &config.scarb_toml_path)
+        .with_context(|| anyhow!("Failed to find package"))?;
+    get_contracts_map(&metadata, &package.id)
```
MaksymilianDemitraszek, line 299: I feel like loading contracts from files is a bit of side-effect in build function.

crates/sncast/src/main.rs
```diff
@@ -122,9 +124,22 @@ fn main() -> Result<()> {
...
+        let metadata = get_scarb_metadata_with_deps(&manifest_path)?;
+        let package_metadata = get_package_metadata(&metadata, &manifest_path)?;
+        let mut artifacts = build(&BuildConfig {
+            scarb_toml_path: manifest_path.clone(),
+            json: cli.json,
```
MaksymilianDemitraszek, line 135: What passing cli.json to scarb implies?

crates/sncast/src/helpers/scarb_utils.rs
MaksymilianDemitraszek: If you want succinctness I would one more function `build_and_load_artifacts`

crates/sncast/src/starknet_commands/script.rs
```diff
@@ -49,6 +46,8 @@ use starknet::providers::jsonrpc::HttpTransport;
 use starknet::providers::JsonRpcClient;
 use tokio::runtime::Runtime;
 
+type ScriptStarknetContractArtifacts = StarknetContractArtifacts;
```
MaksymilianDemitraszek, line 49: Why do we need this alias?

MaksymilianDemitraszek: Generally aliases are not very good as those do not provide any type safety. If you want such you should use `struct ScriptStarknetContractArtifacts(StarknetContractArtifacts)`, though I don't feel this provides any useful information here.

crates/sncast/src/starknet_commands/script.rs
```diff
@@ -381,42 +383,28 @@ fn warn_if_sncast_std_not_compatible(scarb_metadata: &Metadata) -> Result<()> {
...
-        scripts_manifest_path.exists(),
-        "The path = {scripts_manifest_path} does not exist"
-    );
-
-    ScarbCommand::new_with_stdio()
-        .arg("build")
-        .manifest_path(&scripts_manifest_path)
-        .run()
-        .context("failed to compile script with scarb")?;
+fn ensure_lib_artifact(
```
MaksymilianDemitraszek, line 386: I don't know what this function does looking at the name.

MaksymilianDemitraszek (commented):
Overall, very positive change, good job! Some comments included.

crates/sncast/src/helpers/scarb_utils.rs
THenry14: Sure, done in newest commit :+1:

crates/sncast/src/main.rs
THenry14: This does not load it, but rather searches for it. I have added such function in https://github.com/foundry-rs/starknet-foundry/pull/1569 (`ensure_scarb_manifest_path`, and this function additionally verifies it exists. as suggested in this PR, the name is a subject to change). Is that enough, or did you have something else in mind as well?

crates/sncast/src/helpers/scarb_utils.rs
THenry14: Sure, makes sense. Can I make this change in https://github.com/foundry-rs/starknet-foundry/pull/1569, as in this PR I modify build function further, and it will be much easier to do it there?

crates/sncast/src/main.rs
THenry14: scarb output will be in json format, just like sncast. this is for consistency reasons (it's reasonable to assume that users might want the whole output in json, not just some parts of it)

crates/sncast/src/starknet_commands/script.rs
THenry14: I have added it just to not use StarknetContractArtifacts for lib paths injection directly. We need this to make sure script artifacts can be found (and this also removes requirement for users to add [lib] parts in their Scarb.toml)

crates/sncast/src/starknet_commands/script.rs
THenry14: It's supposed to ensure that script artifacts are present in artifacts map, basically what having `[lib] (...)` in Scarb.toml would do. Should I rename to `ensure_script_artifact` or something similar?

abulenok: this looks like a case for an assert

maybe `assert!(script_lib_artifact_exists(metadata, package_metadata, artifacts)?)`

abulenok: From what I can see, the hash map this function returns is not used anywhere

THenry14: > this looks like a case for an assert
maybe assert!(script_lib_artifact_exists(metadata, package_metadata, artifacts)?)

Not 100% sure, as this function is supposed to add this, not check if this exists. Maybe just `inject_lib_artifact` would be better name then?

> From what I can see, the hash map this function returns is not used anywhere

Added assignment in newest commit, to properly indicate that the run is using it

crates/sncast/src/helpers/scarb_utils.rs
THenry14: Did just that, here's the change https://github.com/foundry-rs/starknet-foundry/pull/1569/commits/f00233d509b86ea6f9d16ab1edbdd8d67278ae4a

crates/sncast/src/starknet_commands/script.rs
MaksymilianDemitraszek: `ensure` suggests it is an assert, but it seems like this function returns something. That's where I am lost.

THenry14: English is nnot my native language, but I'm pretty sure ensure != assert. 

Ensure means to make certain something will happen (ensure_lib_artifact = make sure lib artifact will be in Map, make it happen)

Assert means to state a fact of belief confidently and forcefully (assert_lib_artefact = lib artifact has to be there, otherwise the program needs to fail)

So, it makes sense to me that `ensure_lib_artifact` returns something, as it does not asserts its existence but rather makes sure it's there.

THenry14: As discussed offline, renamed it to `inject_lib_artifact` in https://github.com/foundry-rs/starknet-foundry/pull/1565/commits/ba406e7a45dc22a5ba3f4f276a023957b31f6fb6

crates/sncast/src/starknet_commands/script.rs
MaksymilianDemitraszek: I would drop the alias completely and rename `StarknetContractArtifacts` to `CairoArtifact` or something like this (represent artifacts hierarchy properly). Can be done in a separate PR.

MaksymilianDemitraszek (approved):

abulenok (approved):

Issue #1329, opened 2023-12-07
Refactor main function


THenry14:
It smells. We should reduce the line number there. We should also be more consistent in passing cli structs to every command (if possible - need to check with scripts).

PR #1569, opened 2024-01-22
Add support for --package flag

THenry14:
-- * Extract build from declare (https://github.com/foundry-rs/starknet-foundry/pull/1565)
-> * Add support for --package flag
-- * Add workspace tests (https://github.com/foundry-rs/starknet-foundry/pull/1572)

todo:

## Introduced changes

Closes https://github.com/foundry-rs/starknet-foundry/issues/17
Closes https://github.com/foundry-rs/starknet-foundry/issues/449

- Adds support for `--package` flag that can be used to determine the scarb package to work with

## Checklist

- [X] Linked relevant issue
- [X] Added relevant tests


CHANGELOG.md
```diff
@@ -21,6 +21,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Error message for tests that should panic but pass
 
+### Cast
+
+#### Added
+
+- `--package` flag to specify scarb package to work with
```
MaksymilianDemitraszek, line 28: To which command? What does it mean "to work with"?

crates/sncast/src/helpers/scarb_utils.rs
```diff
@@ -196,48 +197,63 @@ pub fn get_scarb_metadata_with_deps(
     execute_scarb_metadata_command(&command)
 }
 
-#[must_use]
-pub fn verify_or_determine_scarb_manifest_path(
-    path_to_scarb_toml: &Option<Utf8PathBuf>,
-) -> Option<Utf8PathBuf> {
+pub fn ensure_scarb_manifest_path(path_to_scarb_toml: &Option<Utf8PathBuf>) -> Result<Utf8PathBuf> {
```
MaksymilianDemitraszek, line 200: I don't understand what this function does from the name.

THenry14: Carry over from https://github.com/foundry-rs/starknet-foundry/pull/1565#discussion_r1464663169

This function ensures valid path to manifest file. I agree that name may be a bit unfortunate, maybe I should change it to something like `get_scarb_manifest`?

docs/src/appendix/sncast/common.md
```diff
@@ -12,6 +12,13 @@ Path to `Scarb.toml` file.
 
 If supplied, `sncast` will not look for `Scarb.toml` file in current (or parent) directory, but will use this path instead.
 
+## `--package <NAME>`
+Optional.
+
+Name of the package that should be used.
+
+If supplied, this package will be used for all operations. Required if more than one package exists in a workspace.
```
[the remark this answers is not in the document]
THenry14: For now - yes in my opinion. We don't have cairo paths in scarb artifacts as of now, so specifying a package is our only choice to pick specific module if we're in a workspace with multiple packages.

CHANGELOG.md
THenry14: Hmm, perhaps you're right. I have added it as global for now, since other things use it indirectly (basically everything that touches cast config, eg accounts). This was because I thought that peraps packages should be able to have their own configs, but I'm not sure now. wydt?

MaksymilianDemitraszek: I think we shouldn't include it when command does not operate on packages. 

Commands having their own configs is not a problem of foundry, scarb should resolve it based on where command is run from.

THenry14: Wait, I'm talking about cast config ([tools.sncast.proflie])

THenry14: Ok, to move things forward I'll remove such possibility (to define configs per package), as perhaps it indeed does not make much sense for some commands. One, top-level config per workspace should be en ough, especially so that we allow for multiple profiles

MaksymilianDemitraszek: Agreed this should be investigated, we can look into later and discuss it with scarb team.

crates/sncast/src/helpers/scarb_utils.rs
MaksymilianDemitraszek: If it returns scarb manifest i think it would be a better name.

CHANGELOG.md
MaksymilianDemitraszek: Can we create an issue for investigating how per package configs work?

MaksymilianDemitraszek:
So the final conclusion is we make `--package` a per command flag, yes?

THenry14:
> So the final conclusion is we make `--package` a per command flag, yes?

Yes, that is correct. I'm working on the changes, but had to jump in to some reviews etc. Will commit later today.

MaksymilianDemitraszek (commented):
Submitting so you can re-request when its ready.

crates/sncast/src/helpers/scarb_utils.rs
THenry14: it turned out there is a funtion named `get_scarb_manifest` (it uses scarb to get the manifest), so I have renamed `ensure_scarb_manifest_path` to `assert_manifest_path_exists`, as after some changes this function actually asserts its existence. (https://github.com/foundry-rs/starknet-foundry/pull/1569/commits/bcc6b0fac2cb70a8deddb51e6ba380034d1574d2)

THenry14: Updated docs + changelog (please check the docs especially, as I'm afraid I didn't stress enough the fact that only most top-level Scarb.toml holds the config https://github.com/foundry-rs/starknet-foundry/pull/1569/commits/40b52d3fb1ea6c6a83570824c676f582aa1d383e)

Moved flags to respective subcommands https://github.com/foundry-rs/starknet-foundry/pull/1569/commits/4791a4d94c28f71cd8e82dc90c9a358a283d8581

THenry14:
> > So the final conclusion is we make `--package` a per command flag, yes?
> 
> Yes, that is correct. I'm working on the changes, but had to jump in to some reviews etc. Will commit later today.

Updated docs + changelog (please check the docs especially, as I'm afraid I didn't stress enough the fact that only most top-level Scarb.toml holds the config https://github.com/foundry-rs/starknet-foundry/pull/1569/commits/40b52d3fb1ea6c6a83570824c676f582aa1d383e)

Moved flags to respective subcommands https://github.com/foundry-rs/starknet-foundry/pull/1569/commits/4791a4d94c28f71cd8e82dc90c9a358a283d8581

CHANGELOG.md
piotmag769: Wdym by top-level config?

crates/sncast/src/helpers/scarb_utils.rs
```diff
@@ -191,43 +192,69 @@ pub fn get_scarb_metadata_with_deps(
...
+    }
+}
+
+pub fn get_main_package_metadata(manifest_path: &Utf8PathBuf) -> Result<PackageMetadata> {
+    let metadata = get_scarb_metadata(manifest_path)?;
+    let main_package = metadata
+        .packages
+        .into_iter()
+        .find(|package| package.manifest_path == metadata.workspace.manifest_path)
+        .expect("Failed to find main package in metadata");
```
piotmag769, line 256: This will fail for virtual workspaces I believe
https://docs.swmansion.com/scarb/docs/reference/workspaces.html#virtual-workspace

THenry14: yup, you are right. I have rebased https://github.com/foundry-rs/starknet-foundry/pull/1572 on friday but did not checked its status, and its failing. trying to figure out a fix now.

CHANGELOG.md
THenry14: yes, this is indeed an issue. So when we have per subcommands --package flag, we'll probably need to check members first, take every tool.sncast config from those packages and then check for default config (without profile name) - at least that's the only sensible solution I can think of. Main drawback of this is one can only have one default (tool.scast) profile per workspace. thoughts?

THenry14: Discussed with @piotmag769 possible solutions to this pickle, and we came up with this flow:

```
- we try to declare/script
    - for workspaces:
        - if there is only one package -> return its castconfig (if it exists, else empty)
        - more than one package -> require `--package` flag, return its castconfig (if it exists, else empty)
    - for single packages:
        - return its castconfig (if it exists, else empty)

- we try to invoke/call/account/whatever else:
    - for workspaces:
        - if there is only one package -> return its castconfig (if it exists, else empty)
        - more than one package -> print out warning that package couldn not be determined, and return an empty castconfig
    - for single packages:
        - return its castconfig (if it exists, else empty)
```

It's not ideal, but allows for using configs from packages for subcommands supporting it, and makes other cases more explicit.

**An alternative approach** would be to revert the `--package only to declare/script` altogether, and make it global again which would allow to pick workspaces (and thus, ccastconfigs) for each subcommand. That would be a cleaner and easier solution, but would mean that (in some cases) we must specify a package to invoke some random contract, which is far from perfect. 

Thoughts? Is the solution 1) perfectly acceptable, should we lean to solution 2) with the alternative approach, or maybe something different altogether?

crates/sncast/src/main.rs
```diff
@@ -149,6 +151,9 @@ fn main() -> Result<()> {
         print_command_result("script", &mut result, numbers_format, &output_format)?;
         Ok(())
     } else {
+        let mut config = load_config(&cli.profile, &None)?;
```
THenry14, line 154: improving this will be a part of https://github.com/foundry-rs/starknet-foundry/issues/1329

MaksymilianDemitraszek (approved):

crates/sncast/src/helpers/scarb_utils.rs
```diff
@@ -63,43 +63,59 @@ pub fn get_scarb_metadata_with_deps(manifest_path: &Utf8PathBuf) -> Result<Metad
...
+            "Path to Scarb.toml manifest does not exist = {manifest_path}"
+        ));
     }
 
-    Some(manifest_path)
+    Ok(manifest_path)
 }
 
-pub fn get_package_metadata<'a>(
+fn get_package_metadata_by_name<'a>(
```
piotmag769, line 86: This is prob `get_workspace_member_metadata_by_name`, right? It should use `metadata.workspace. members`

crates/sncast/src/helpers/scarb_utils.rs
```diff
@@ -63,43 +63,59 @@ pub fn get_scarb_metadata_with_deps(manifest_path: &Utf8PathBuf) -> Result<Metad
...
+            &package_name
+        ))
+}
+
+fn get_default_package_metadata(metadata: &Metadata) -> Result<&PackageMetadata> {
+    match metadata.packages.iter().collect::<Vec<_>>().as_slice() {
+        [package] => Ok(package),
+        [] => Err(anyhow!("No package found in metadata")),
+        _ => Err(anyhow!(
+            "More than one package found in metadata - specify package using --package flag"
```
abulenok, line 105:
```suggestion
        [] => Err(anyhow!("No package found in scarb metadata")),
        _ => Err(anyhow!(
            "More than one package found in scarb metadata - specify package using --package flag"
```

crates/sncast/src/starknet_commands/declare.rs
```diff
@@ -45,7 +49,7 @@ pub async fn declare(
     let contract_name: String = contract_name.to_string();
     let contract_artifacts = artifacts
         .get(&contract_name)
-        .ok_or(anyhow!("Failed to find artifacts in starknet_artifacts.json file. Please ensure you have enabled sierra and casm code generation in Scarb.toml"))?;
+        .ok_or(anyhow!("Failed to find artifacts in starknet_artifacts.json file. Please make sure you have specified correct package using `--package` flag and that you have enabled sierra and casm code generation in Scarb.toml."))?;
```
abulenok, line 52: It would be nice to include the contract name in the error message.

Smth like `Failed to find artifacts in starknet_artifacts.json file for {contract_name} contract`

crates/sncast/src/helpers/scarb_utils.rs
THenry14: yup, as we discussed offline - this will be a part of scarb_utils cleanup mentioned in pr description :+1:

crates/sncast/src/starknet_commands/declare.rs
THenry14: the diff showed, because I merged a few PRs and did not rebase - I have now done this, and contract name is included in the message.

Issue #1564, opened 2024-01-19
Refactor `scarb_version_command` function from the scarb-api/src/version.rs


war-in:
Can we add a follow-up issue to refactor this, after the `Metadata` is integrated into the crate as well? 
Namely: i think this should be accessible via the `ScarbCommand` struct from the crate (through a new constructor method or something like that)

_Originally posted by @Arcticae in https://github.com/foundry-rs/starknet-foundry/pull/1514#discussion_r1458945470_

PR #1570, opened 2024-01-22
1509 scarb metadata

### LINKED ISSUE #1509: Scarb metadata error should be nicely printed
```
+ stdout:
+ {"type":"error","message":"failed to parse manifest at: /private/var/folders/0_/yp337wzx1ss_qgrd9kpfx9_00000gn/T/.tmpqpHqq8/Scarb.toml\n\nCaused by:\n    dependency (snforge_std) specification is ambiguous, only one of `git` or `path` is allowed"}
+ 
```
Currently scarb metadata in forge just redirects json to stdout. if we use json we should interpret and pretty print the error.

Draggu:
Closes #1509, #1564

## Introduced changes

- Included `MetadataCommand` and `scarb_version_command ` into `ScarbCommand`. Added pretty print of `scarb` errors.

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


CHANGELOG.md
```diff
@@ -16,6 +16,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 #### Changed
 
 - `available_gas` attribute is now supported (Scarb >= 2.4.4 is required)
+- Included `MetadataCommand` into `ScarbCommand`
```
MaksymilianDemitraszek, line 19: I think it shouldn't be mentioned in changelog, we try to keep in changelog only user facing changes.

MaksymilianDemitraszek (changes requested):
I think we shouldn't solve two separate issues in one PR unless those are very tightly coupled. Better idea would be to make two separate PRs addressing each of those.

Could you please split it into two separate PRs each addressing one problem? I know it is extra work but it is really important when working in a large team to introduce changes gradually. This way everybody can keep up with them 😞.

crates/scarb-api/src/output_transform.rs
```diff
@@ -0,0 +1,35 @@
+use std::borrow::Cow;
+
+pub trait OutputTransform: Clone + Default {
```
MaksymilianDemitraszek, line 3: I really like the idea of having two different stream transformations available!
Though I think it would be much simpler instead of encoding those as generics to just provide
methods/structs to transform the output. I don't see any benefit of having output transformation being encoded as a part of command definition type.

crates/scarb-api/src/metadata_command.rs
```diff
@@ -0,0 +1,247 @@
...
+    fn transform_stderr(stderr: &[u8]) -> Cow<'_, [u8]> {
+        pretty_print(stderr)
+    }
+    fn transform_stdout(stdout: &[u8]) -> Cow<'_, [u8]> {
+        pretty_print(stdout)
+    }
+}
+
+pub struct ScarbMetadataCommand<Stdout = PassByPrint, Stderr = PassByPrint> {
+    inner: ScarbCommand<Stdout, Stderr>,
```
MaksymilianDemitraszek, line 57: I think this is not a good idea to try to emulate subtyping here. Better idea would be to represent `ScarbCommand` as enum (nicely documents available options) and name a strcut containing shared arguments `ScarbCommandSharedArgs` or something similar.

MaksymilianDemitraszek (changes requested):
Some comments regarding the code itself, i think it is very good direction you taking trying to generalize it 🔥 !
We need to achieve two goals here:
- generalization of commands hierarchy (Metadata, Version)
- generalization of output transformation (Identity, PrettyPrint)

I think it would be much easier to discuss those in separation as I mentioned in the previous comment. After you split the work I am happy to assist both discussions!

I am leaving some additional comments and general feelings about specific parts of the code.

PR #1571, opened 2024-01-23
Fix fork error handling

### LINKED ISSUE #1497: Non-deterministic missing contracts in forking
At times there are some false-positives when trying to retrieve/interact with contracts, namely:
```
Got an exception while executing a hint: Hint Error: Requested contract address ContractAddress(PatriciaKey(StarkFelt("0x00000005dd3d2f4429af886cd1a3b08289dbcea99a294197e9eb43b0e0325b4b"))) is not deployed.
```
(mainnet)

Hint: might be related to a caching issue

Arcticae:
Closes #1497 

## Introduced changes

- Fixes incosistencies in fork state behavior when errors should be raised

TODO: 
- [x] Fix gas calculation (is crashing right now in forking)

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/cheatnet/src/state.rs
```diff
@@ -317,13 +318,12 @@ impl TraceData {
...
 
-fn match_node_response<T: Default>(result: StateResult<T>) -> StateResult<T> {
+fn match_node_response<T>(result: StateResult<T>) -> StateResult<T> {
     match result {
         Ok(class_hash) => Ok(class_hash),
-        Err(StateError::StateReadError(msg)) if msg.contains("node") => {
-            Err(StateError::StateReadError(msg))
-        }
-        _ => Ok(Default::default()),
+        Err(StateReadError(msg)) if msg.contains("node") => Err(StateReadError(msg)),
```
drknzz, line 324: This arm is not needed, since the last arm would do the same anyway?

crates/cheatnet/src/state.rs
```diff
@@ -317,13 +318,12 @@ impl TraceData {
     }
 }
 
-fn match_node_response<T: Default>(result: StateResult<T>) -> StateResult<T> {
+fn match_node_response<T>(result: StateResult<T>) -> StateResult<T> {
```
drknzz, line 321: convert_node_response? but it'd still not be very clear to me what's the purpose of this function through that name

drknzz: This is basically `map_err` if the error is other than `UndeclaredClassHash`

crates/cheatnet/src/forking/state.rs
```diff
@@ -165,9 +167,12 @@ impl StateReader for ForkStateReader {
                     .cache_get_class_hash_at(contract_address, class_hash);
                 Ok(class_hash)
             }
+            Err(ProviderError::StarknetError(StarknetError::ContractNotFound)) => {
+                Ok(Default::default())
+            }
```
drknzz, line 172: Is this okay to do?

Arcticae: Yes - we have to do it because it's what blockifier expects (during i.e. deploy, we check for this and if the hash is default, we allow deployment)

crates/cheatnet/src/state.rs
Arcticae: That's right

crates/cheatnet/src/state.rs
Arcticae: I will convert this into `map_err`

Arcticae: i actually removed it overall, not needed

drknzz: awesome

drknzz (approved):
fire emoji

PR #1572, opened 2024-01-23
Add workspaces tests

THenry14:
-- * Extract build from declare (https://github.com/foundry-rs/starknet-foundry/pull/1565)
-- * Add support for --package flag (https://github.com/foundry-rs/starknet-foundry/pull/1569)
-> * Add workspace tests

todo: 

## Introduced changes

- Add a few more tests for scarb workspaces

## Checklist

- [X] Linked relevant issue
- [X] Added relevant tests


crates/sncast/tests/e2e/declare.rs
```diff
@@ -286,12 +288,75 @@ async fn test_many_packages_default() {
...
+        "99999999999999999",
+    ];
+
+    let snapbox = Command::new(cargo_bin!("sncast"))
+        .current_dir(tempdir.path())
+        .args(args);
+    snapbox.assert().success().stderr_matches(indoc! {r"
+        ...
+        command: declare
+        error: Failed to find artifacts in starknet_artifacts.json file[..]
```
abulenok, line 360: It would be nice to include the contract name in the error message.

Smth like `Failed to find artifacts in starknet_artifacts.json file for {contract_name} contract`

crates/sncast/tests/e2e/script/general.rs
```diff
@@ -124,3 +124,54 @@ async fn test_incompatible_sncast_std_version() {
         ...
```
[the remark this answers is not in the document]
THenry14: This is basically each of our pre-existing tests, as every package must ezist in a workspace, right? So for example our `map` contract is a single package, that resides in a workspace:

```
starknet-foundry/crates/sncast/tests/data/contracts/map $ ➜ scarb metadata --format-version 1 | grep workspace -A 5
  "workspace": {
    "manifest_path": "/Users/wojciech/git-repos/starkware/starknet-foundry/crates/sncast/tests/data/contracts/map/Scarb.toml",
    "root": "/Users/wojciech/git-repos/starkware/starknet-foundry/crates/sncast/tests/data/contracts/map",
    "members": [
      "map 0.1.0 (path+file:///Users/wojciech/git-repos/starkware/starknet-foundry/crates/sncast/tests/data/contracts/map/Scarb.toml)"
    ]
```

do you think something more should be added?

crates/sncast/tests/e2e/declare.rs
```diff
@@ -286,12 +288,75 @@ async fn test_many_packages_default() {
...
+        "--account",
+        "user8",
+        "--int-format",
+        "--json",
+        "declare",
+        "--package",
+        "cast_fibonacci",
+        "--contract-name",
+        "FibonacciContract",
+        "--max-fee",
```
tkumor3, line 318: [nitpick]: in tests, we can use auto_estimated `max-fee`. IMO it will be cleaner.

tkumor3 (approved):

crates/sncast/tests/e2e/declare.rs
THenry14: most of the time it won't work sadly :( the problem is that we execute most of the tests concurrently, and the order of them is dufferent from run to runu - that is why the fee price fluctuates, big time. This may lead to non deterministic tests, failing from time to time because the estimated fee that was valid a second ago, is now too little.

PR #1605, opened 2024-01-26
Added signal mechanism to Extensions

MaksymilianDemitraszek:


MaksymilianDemitraszek:
This PR extends extension framework with signal handlers. Those are executed after handling a cheatcode/syscall. Don't have access to cairo memory and cannot raise an error to user.

It will be used for incremental learning extension which will learn the names associated with contract addresses etc. required for profiler.

MaksymilianDemitraszek:
Future refactoring work:  #1610

crates/runtime/src/lib.rs
```diff
@@ -291,6 +385,26 @@ pub trait ExtensionLogic {
     ) -> Result<CheatcodeHandlingResult, EnhancedHintError> {
         Ok(CheatcodeHandlingResult::Forwarded)
     }
+
+    /// Different from `override_system_call` because it cannot be overridden,
+    /// always receives a signal and cannot return an error
+    fn listen_system_call(
```
Arcticae, line 391: those names are somewhat wrong since it does not listen, but triggers a handler

Arcticae (approved):

crates/runtime/src/lib.rs
```diff
@@ -220,6 +203,117 @@ impl<Extension: ExtensionLogic> HintProcessorLogic for ExtendedRuntime<Extension
...
 }
 
+struct VmIoPointers<'a> {
+    input_start: &'a ResOperand,
+    input_end: &'a ResOperand,
+    output_start: &'a CellRef,
+    output_end: &'a CellRef,
+}
+
+impl<Extension: ExtensionLogic> ExtendedRuntime<Extension> {
```
piotmag769, line 213: I would add a comment above that it is here just to make `impl<Extension: ExtensionLogic> HintProcessorLogic for ExtendedRuntime<Extension>` more readable, it won't be immediately obvious for someone reading it in the future

piotmag769 (approved):

crates/runtime/src/lib.rs
```diff
@@ -220,6 +203,117 @@ impl<Extension: ExtensionLogic> HintProcessorLogic for ExtendedRuntime<Extension
...
+            .listen_system_call(selector, vm, &mut self.extended_runtime);
+    }
+
+    fn propagate_cheatcode_signal(&mut self, selector: &str, inputs: Vec<Felt252>) {
+        self.extended_runtime
+            .propagate_cheatcode_signal(selector, inputs.clone());
+        self.extension
+            .listen_cheatcode(selector, inputs, &mut self.extended_runtime);
+    }
+}
```
piotmag769, line 315: Really smart, gj

crates/runtime/src/lib.rs
MaksymilianDemitraszek: Done

crates/runtime/src/lib.rs
MaksymilianDemitraszek: Not sure about that 🤔 Isn't this purpose of splitting into functions to make the code more readable? I don't understand why this is different.

crates/runtime/src/lib.rs
MaksymilianDemitraszek: I extended it a little bit

PR #1639, opened 2024-01-31
`syscall_ptr` with `SegmentArena` fix 

### LINKED ISSUE #1614: bug: Bad syscall_ptr

snforge

### Foundry Version

0.14.0

MacOS

arm

### What happened

# Bug Report

**Cairo version:**
2.4.4

**Current behavior:**
Running `snforge test unit_tests` on https://github.com/keep-starknet-strange/unruggable.meme/pull/168 fails after introducing the unique_count function.

```rust
fn unique_count<T, +Copy<T>, +Drop<T>, +PartialEq<T>, +Into<T, felt252>>(mut self: Span<T>) -> u32 {
    let mut dict: Felt252Dict<felt252> = Default::default();
    let mut counter = 0;
    loop {
        match self.pop_front() {
            Option::Some(value) => {
                let value: felt252 = (*value).into();
                if dict.get(value).is_one() {
                    continue;
                }
                dict.insert(value, One::one());
                counter += 1;
            },
            Option::None => { break; }
        }
    };
    counter
}
```

This function is called here: https://github.com/keep-starknet-strange/unruggable.meme/pull/168/commits/3c9650d2c31935348f4e544bffaae19969fe176f#diff-6efe688b97fc8a048f099b061edb855bc8e01aeaf909e9695614946336fe8a5aR412

interestingly, the unit test associated to this function doesn't fail.
https://github.com/keep-starknet-strange/unruggable.meme/pull/168/commits/3c9650d2c31935348f4e544bffaae19969fe176f#diff-5eac007cd88b52b4d4889af203d613ca1aa11bdfdf555a9024599265b04735acR3

### Trace

```shell
Failure data:
    Got an exception while executing a hint:
    Bad syscall_ptr; expected: Relocatable { segment_index: 10, offset: 0 }, got: Relocatable { segment_index: 12, offset: 0 }.

```

Arcticae:
Closes #1614 

## Introduced changes

- Fixes inconsistency which is generated because of inconsistent syscall pointer when syscall_arena builtin is allocated in the segments (done only if it is required, on CASM generation level)

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests

### DIFF
--- a/crates/forge-runner/src/running.rs
+++ b/crates/forge-runner/src/running.rs
@@ -22,6 +22,8 @@ use cairo_lang_runner::casm_run::{build_cairo_runner, run_function_with_runner};
 use cairo_lang_runner::{
     build_hints_dict, initialize_vm, Arg, RunResult, RunnerError, SierraCasmRunner,
 };
+use cairo_lang_sierra::extensions::segment_arena::SegmentArenaType;
+use cairo_lang_sierra::extensions::NoGenericArgsGenericType;
 use cairo_lang_sierra::ids::GenericTypeId;
 use cairo_lang_sierra_to_casm::compiler::CairoProgram;
 use cairo_vm::types::relocatable::{MaybeRelocatable, Relocatable};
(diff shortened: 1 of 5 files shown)


drknzz (commented):
Did you manage to provoke the error from the issue? I don't see it in CI

piotmag769 (approved):
Gj

Arcticae:
@drknzz  https://github.com/foundry-rs/starknet-foundry/pull/1639/files#diff-60aaeb3ffd3b68fa6b48b3bbc8549c296853497cdf6dfbf956353393a6d88bc6R799
This one was failing before, you can try locally pass segment index as 10 everywhere (as it was before) and then you would get an error there.

drknzz (approved):
> @drknzz https://github.com/foundry-rs/starknet-foundry/pull/1639/files#diff-60aaeb3ffd3b68fa6b48b3bbc8549c296853497cdf6dfbf956353393a6d88bc6R799 This one was failing before, you can try locally pass segment index as 10 everywhere (as it was before) and then you would get an error there.

Okay, it's just good to push a failing reproduction like this to CI at the beginning imo, to have a proof that the changes introduced actually fixed the problem

Other than that, well done 💯💯💯

crates/forge-runner/src/running.rs
```diff
@@ -137,16 +139,26 @@ fn build_syscall_handler<'a>(
...
 
+    // Segment arena is allocated conditionally, so segment index is automatically moved (+2 segments)
+    let segment_index = if test_param_types
+        .iter()
+        .any(|(ty, _)| ty == &SegmentArenaType::ID)
+    {
+        12
+    } else {
+        10
+    };
```
MaksymilianDemitraszek, line 154: maybe move it to separate funciton `get_syscall_segment_index`.

PR #1645, opened 2024-02-01
Decode strings in panic data

### LINKED ISSUE #1612: Revamp test collector failure output

snforge

### Description

We should try to improve the error message so it's easier for the user to make sense of it

Image

drknzz:
Closes #1612 

## Introduced changes

- Detect strings / shortstrings / felts in panic data and display accordingly

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests

### DIFF
--- a/crates/forge-runner/src/test_case_summary.rs
+++ b/crates/forge-runner/src/test_case_summary.rs
@@ -281,7 +281,7 @@ fn build_readable_text(data: &[Felt252]) -> Option<String> {
     let string = if let [item] = &items[..] {
         item.clone()
     } else {
-        items.join(", ")
+        format!("({})", items.join(", "))
     };
 
     let mut result = indent_string(&format!("\n{string}"));
--- a/crates/forge-runner/src/test_case_summary.rs
+++ b/crates/forge-runner/src/test_case_summary.rs
@@ -2,14 +2,13 @@ use crate::compiled_runnable::TestCaseRunnable;
 use crate::expected_result::{ExpectedPanicValue, ExpectedTestResult};
 use crate::gas::check_available_gas;
 use crate::trace_data::CallTrace;
-use cairo_felt::{felt_str, Felt252};
-use cairo_lang_runner::short_string::{as_cairo_short_string, as_cairo_short_string_ex};
+use cairo_felt::Felt252;
+use cairo_lang_runner::casm_run::format_next_item;
+use cairo_lang_runner::short_string::as_cairo_short_string;
 use cairo_lang_runner::{RunResult, RunResultValue};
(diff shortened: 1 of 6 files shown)


crates/forge-runner/src/test_case_summary.rs
```diff
@@ -264,23 +268,78 @@ impl TestCaseSummary<Single> {
...
     }
+    Some(format!("original value: [{first_felt}]"))
+}
+
+// https://github.com/starkware-libs/cairo/blob/1a4ccb01f6fb0a74bfc23238ca1088fe2659d087/crates/cairo-lang-runner/src/casm_run/mod.rs#L2227
+
+/// Tries to format a string, represented as a sequence of `Felt252`s.
+/// If the sequence is not a valid serialization of a ByteArray, returns None and doesn't change the
+/// given iterator (`values`).
+fn try_format_string<T>(values: &mut T) -> Option<String>
```
drknzz, line 312: Unfortunately it's a private function

piotmag769: Can we have an issue to make it public in the compiler?

crates/forge-runner/src/test_case_summary.rs
```diff
@@ -264,23 +266,77 @@ impl TestCaseSummary<Single> {
...
     } else {
-        readable_text.push('\n');
-        Some(readable_text)
+        let mut result = items.join("").to_string().replace('\n', "\n    ");
+        result.push('\n');
+        Some(result)
+    }
+}
+
+fn format_next_felt<T>(values: &mut T) -> Option<String>
```
piotmag769, line 287: another private one, right?

crates/forge-runner/src/test_case_summary.rs
```diff
@@ -264,23 +266,77 @@ impl TestCaseSummary<Single> {
...
+    if first_felt == felt_str!(BYTE_ARRAY_MAGIC, 16) {
+        if let Some(string) = try_format_string(values) {
+            return Some(string);
+        }
+    }
+
+    if let Some(short_string) = as_cairo_short_string(&first_felt) {
+        return Some(format!(
+            "original value: [{first_felt}], converted to a string: [{short_string}]"
+        ));
```
piotmag769, line 302: Can we revamp this one too? Rn only strings are printed nicely, shortstrings could be too. Example from cairo-test:
image

war-in (approved):
Let's get rid of `original_value: ...` and it's 🔥

crates/forge-runner/src/test_case_summary.rs
Arcticae: Also i think a panic array output should be output as a whole instead of `n` formatted lines like that

Arcticae (approved):
:shipit:

PR #1644, opened 2024-02-01
Replace Scarb.toml tool config with snfoundry.toml 

THenry14:
https://github.com/foundry-rs/starknet-foundry/issues/1620
This PR is a part of stack, that aims to replace Scarb.toml tool configuration with snfoundry.toml:

-- Add mechanism for snfoundry.toml discovery (https://github.com/foundry-rs/starknet-foundry/pull/1643)
-> Replace Scarb.toml based configuration with snfoundry.toml based one
-- Use the same profile for sncast and scarb
-- Update docs

## Introduced changes

- use snfoundry.toml for tool configuration

## Checklist

- [X] Linked relevant issue
- [X] Updated relevant documentation
- [X] Added relevant tests
- [X] Performed self-review of the code
- [X] Added changes to `CHANGELOG.md`

PR #1646, opened 2024-02-01
Use the same profile for sncast and scarb

THenry14:
https://github.com/foundry-rs/starknet-foundry/issues/1620
This PR is a part of stack, that aims to replace Scarb.toml tool configuration with snfoundry.toml:

-- Add mechanism for snfoundry.toml discovery (https://github.com/foundry-rs/starknet-foundry/pull/1643)
-- Replace Scarb.toml based configuration with snfoundry.toml based one (https://github.com/foundry-rs/starknet-foundry/pull/1644)
-> Use the same profile for sncast and scarb
-- Update docs (https://github.com/foundry-rs/starknet-foundry/pull/1647)

## Introduced changes

- Use the same profile for sncast and scarb

## Checklist

- [X] Linked relevant issue
- [X] Added relevant tests


MaksymilianDemitraszek (approved):
How does it behave when profile is not present in one of the configs?

Somebody from chain interactions team should review it before merging.

THenry14:
> How does it behave when profile is not present in one of the configs?

When someone passes the profile with `--profile` flag, and it does not exist in snfoundry.toml, we throw out an error that there is no such profile. When no `--profile` is passed we default to requiring inputs from cli.

This does not affect scarb - if there is a profile with specified name in scarb, we use this profile. Otherwise we fall back to scarb's default (which is `dev` profile).

MaksymilianDemitraszek:
> Otherwise we fall back to scarb's default (which is dev profile).

I am not sure if it is good as it is inconsistent with snfoundry toml behaviour.

THenry14:
> > Otherwise we fall back to scarb's default (which is dev profile).
> 
> I am not sure if it is good as it is inconsistent with snfoundry toml behaviour. Also it may lead to unexpected implicit behaviour. Example: somebody wants to compile something with profile y but forgot the name, he doesn't know that he uses wrong profile name but gets stuff compiled not the way they wanted.
> 
> On the other hand it does not require dev to specify corresponding profiles in each scarb config.

The last part was my main reasoning behind doing this that way - it would be unreasonable to require users to create scarb profile per each snfoundry profile. That implicit behaviour is scarb's default, so I wouldn't bother too much from our point of view, although a nice idea would be to print a warning when this occurs - I mean, if there is no scarb profile with given name, and that we default to dev. wdyt?

MaksymilianDemitraszek:
>That implicit behaviour is scarb's default

If it is the case I think it is ok 👍

ddoktorski:
To be honest I'm a bit unsure about the idea of using a common `profile` flag for sncast and scarb:
- It presents a risk that a user may unintentionally give the scarb and cast profiles the same name
- It demands that users understand that the profile flag is used for scarb solely in the context of `script` and `declare` commands. In all other instances, it's used only as a cast profile name
- I suspect that sncast profiles are frequently used compared to custom scarb profiles (please correct me if I'm wrong). If this is indeed the case, it might not be worth introducing potentially confusing logic that would only serve a small number of users

THenry14:
> To be honest I'm a bit unsure about the idea of using a common `profile` flag for sncast and scarb:
> 
>     * It presents a risk that a user may unintentionally give the scarb and cast profiles the same name
> 
>     * It demands that users understand that the profile flag is used for scarb solely in the context of `script` and `declare` commands. In all other instances, it's used only as a cast profile name
> 
>     * I suspect that sncast profiles are frequently used compared to custom scarb profiles (please correct me if I'm wrong). If this is indeed the case, it might not be worth introducing potentially confusing logic that would only serve a small number of users

Thanks for pointing these things out, here's how I see them:

ad 1. that is true, but not sure there is anything we can do about it (+ if we should do anything about it), except maybe informing users every time that specified profile is used. We cannot do everything for the users, I'd say everything is in docs, go get familiar with them. 

ad 2. yup, scarb is used for those two commands only so it makes sense that scarb's profile won't be used, when scarb is not used. Technically speaking, each time scarb is involved the profile will be taken, and in all other cases it does not matter because nothing happens with scarb. User has a one interface, and that's it.

ad 3. custom profiles - perhaps, but my guess is quite a lot of users can actually use `dev` and `release` (so built in profiles). up until now we did not actually support it, meaning everything was with the default `dev` profile.
to be honest I see nothing confusing about this - from cast user pov you are using a profile in sncast that is passed to scarb as well, and that's it.

piotmag769 (approved):

abulenok:
What about the case when the user wants to use profile `x` for sncast and profile `y` for Scarb, for example in deployment scripts?

crates/sncast/src/helpers/scarb_utils.rs
```diff
@@ -108,23 +108,31 @@ pub fn get_package_metadata<'a>(
...
 pub fn build(config: &BuildConfig) -> Result<HashMap<String, StarknetContractArtifacts>> {
     let mut cmd = ScarbCommand::new_with_stdio();
-    cmd.arg("build").manifest_path(&config.scarb_toml_path);
+    let metadata = get_scarb_metadata_command(&config.scarb_toml_path)?
+        .exec()
+        .expect("Failed to obtain metadata");
+    let profile = if metadata.profiles.contains(&config.profile) {
+        &config.profile
+    } else {
+        "dev"
```
abulenok, line 122: I suggest we introduce `const SCARB_DEFAULT_PROFILE`

THenry14: why?

THenry14:
> What about the case when the user wants to use profile `x` for sncast and profile `y` for Scarb, for example in deployment scripts?

we do not support such cases.

crates/sncast/src/helpers/scarb_utils.rs
THenry14: moved discussion offline, will make any changes in separate PR if necessary

PR #1647, opened 2024-02-01
Update docs

THenry14:
https://github.com/foundry-rs/starknet-foundry/issues/1620
This PR is a part of stack, that aims to replace Scarb.toml tool configuration with snfoundry.toml:

-- Add mechanism for snfoundry.toml discovery (https://github.com/foundry-rs/starknet-foundry/pull/1643)
-- Replace Scarb.toml based configuration with snfoundry.toml based one (https://github.com/foundry-rs/starknet-foundry/pull/1644)
-- Use the same profile for sncast and scarb (https://github.com/foundry-rs/starknet-foundry/pull/1646)
-> Update docs

## Introduced changes

- Update docs and changelog to include info about snfoundry.toml

## Checklist

- [X] Linked relevant issue
- [X] Added relevant tests


docs/src/appendix/sncast/account/add.md
```diff
@@ -50,7 +50,7 @@ Optional.
 
 Salt for the account address.
 
-## `--add-profile`
+## `--add-profile <NAME>`
 Optional.
 
-If passed, a profile with corresponding data will be added to Scarb.toml.
+If passed, a profile with corresponding name will be added to Scarb.toml.
```
MaksymilianDemitraszek, line 56:
```suggestion
If passed, a profile with corresponding name will be added to snfoundry.toml.
```

docs/src/appendix/sncast/account/create.md
```diff
@@ -18,10 +18,10 @@ Optional.
 
 Salt for the account address. If omitted random one will be generated.
 
-## `--add-profile`
+## `--add-profile <NAME>`
 Optional.
 
-If passed, a profile with corresponding data will be added to Scarb.toml.
+If passed, a profile with corresponding name will be added to Scarb.toml.
```
MaksymilianDemitraszek, line 24:
```suggestion
If passed, a profile with corresponding name will be added to snfoundry.toml.
```

docs/src/appendix/sncast/common.md
```diff
@@ -3,7 +3,8 @@
 ## `--profile, -p <PROFILE_NAME>`
 Optional.
 
-Profile name in `Scarb.toml` config file.
+Used for both `snfoundry.toml` and `Scarb.toml` if specified (defaulting to `dev` in `Scarb.toml` if absent).
+Defaults to `default`.
```
MaksymilianDemitraszek, line 7: Not sure if i understand. First sentence is about the being specified case, second about the default, but you put parentheses with additional information there.

docs/src/getting-started/scarb.md
```diff
@@ -8,7 +8,7 @@ Starknet Foundry uses [Scarb](https://docs.swmansion.com/scarb) to:
 - [build contracts](https://docs.swmansion.com/scarb/docs/extensions/starknet/contract-target.html)
 
 One of the core concepts of Scarb is its [manifest file](https://docs.swmansion.com/scarb/docs/reference/manifest.html) - `Scarb.toml`.
-It can be also used to provide [configuration](../projects/configuration.md) for Starknet Foundry.
+It can be also used to provide [configuration](../projects/configuration.md) for Starknet Foundry Forge.
```
MaksymilianDemitraszek, line 11:
```suggestion
It can be also used to provide [configuration](../projects/configuration.md) for Starknet Foundry.
```
I believe we need to make more effort describing the difference between foundry and scarb tomls.

Good example is where configuration should be taken from (which package) for `sncast get-nonce`

MaksymilianDemitraszek (changes requested):
I don't like the narration you put here that
- foundry toml is for cast
- scarb is for forge

The problem is much more nuanced and we should put more effort to explain the difference between the configs.

docs/src/projects/configuration.md
```diff
@@ -18,30 +18,33 @@ exit_first = true
...
-With `Scarb.toml` configured this way, we can just pass `--profile myprofile` argument to make sure `sncast` uses parameters
+With `snfoundry.toml` configured this way, we can just pass `--profile myprofile` argument to make sure `sncast` uses parameters
 defined in the profile.
 
 > 📝 **Note**
 > `Scarb.toml` file has to be present in current or any of the parent directories.
 > Alternatively, you can also point to `Scarb.toml` path with `--path-to-scarb-toml <PATH>` flag.
 
+> 📝 **Note**
+> If there is a profile with the same name in Scarb.toml, scarb will use this profile. If not, scarb will default to using the dev profile.
```
ddoktorski, line 46: Can we highlight that this only affects the `declare` and `script` commands?

docs/src/projects/configuration.md
```diff
@@ -18,30 +18,33 @@ exit_first = true
...
 # ...
 ```
 
-With `Scarb.toml` configured this way, we can just pass `--profile myprofile` argument to make sure `sncast` uses parameters
+With `snfoundry.toml` configured this way, we can just pass `--profile myprofile` argument to make sure `sncast` uses parameters
 defined in the profile.
 
 > 📝 **Note**
 > `Scarb.toml` file has to be present in current or any of the parent directories.
 > Alternatively, you can also point to `Scarb.toml` path with `--path-to-scarb-toml <PATH>` flag.
```
ddoktorski, line 43:
```suggestion
> `snfoundry.toml` file has to be present in current or any of the parent directories.
```

docs/src/appendix/sncast/common.md
THenry14: this is not trivial to explain in one sentence, so maybe I did too big somersault here.

What I wanted to include in this iformation is:

```
if profile is passed, use it in sncast and scarb, unless there is no such profile in scarb, then `dev` will be used.
If profile is not passed, use `default` in cast and `dev` in scarb
```

I feel somethings like this:

```
Used for both `snfoundry.toml` and `Scarb.toml` if specified.
Defaults to `default` (`snfoundry.toml`) and `dev` (`Scarb.toml`).
```

Does not fully explains it. any ideas?

docs/src/projects/configuration.md
THenry14: done in https://github.com/foundry-rs/starknet-foundry/pull/1647/commits/239fee695100fdb98bd9d0dabd8d0a24ede91e3d , please check if the working etc is correct

docs/src/appendix/sncast/common.md
MaksymilianDemitraszek: Yes

docs/src/getting-started/scarb.md
MaksymilianDemitraszek: > Not sure what should be done to make it more explicit to be honest. In configuration part it is described that forge uses Scarb.toml and sncast uses snfoundry.toml, I guess it is a good place to describe such things.

I believe sncast may use Scarb.toml in the future, and forge snfoundry.toml . I believe the difference is not in the tool but between semantics of those two configs. Scarb.toml is a package definition. snfoundry.toml holds configuration that doesn't have to be tied to a specific package.

docs/src/appendix/sncast/common.md
THenry14: That's cool, do you mind sharing? :D

docs/src/getting-started/scarb.md
THenry14: > Scarb.toml is a package definition. snfoundry.toml holds configuration that doesn't have to be tied to a specific package.

That is true, but only for sncast. Correct me if I'm wrong, but forge can technically hold (at least at the moment) some forge-specific things in Scarb.toml as well, right?

docs/src/appendix/sncast/common.md
MaksymilianDemitraszek: Sorry didn't sleep well today :D , i read it as 

> Does it fully explain the idea?

I think what you suggested is sufficient.

docs/src/getting-started/scarb.md
MaksymilianDemitraszek: Yes, because everything in forge is package specific, as every test belongs to specific package, so it is possible to always resolve which config should be the source. 

get_nonce is a good example why snfoundry.toml is necessary in cast, because from which package you would take the config otherwise.

Cast still should use scarb toml in case we have things that can always be resolved in the context of specific package. I think for example that script configuration should be held (or inherited) by scarb toml from snfoundry toml as scripts are always in the context of specific package.

docs/src/appendix/sncast/common.md
MaksymilianDemitraszek: But you can also make it as you specified above. I don't think there is any hurt in verbosity.

docs/src/getting-started/scarb.md
THenry14: Ok, so my understanding is that I should:
- drop `Forge` from the description about holding config in Scarb.toml
- emphasize more than snfoundry is used for cast, because we are not always in some package (as opposed to forge)

is that right?

> Cast still should use scarb toml in case we have things that can always be resolved in the context of specific package. I think for example that script configuration should be held (or inherited) by scarb toml from snfoundry toml as scripts are always in the context of specific package.

I think this might just make things more messy. I'd stick with using snfoundry for that, as technically we do not configure a package, but rather a script execution (so cast). + actually support for reading configs from Scarb.toml is dropped for cast, so it might be misleading to suggest something like this is possible

docs/src/appendix/sncast/common.md
THenry14: done https://github.com/foundry-rs/starknet-foundry/pull/1647/commits/173cdcf3660f62ed54621c4f138a0caf0c0e3f70

docs/src/getting-started/scarb.md
MaksymilianDemitraszek: I am ok with it but there is no explanation (or at least I don't see it) why snfoundry toml is used instead of Scarb toml for cast.

MaksymilianDemitraszek (approved):
I am ok with merging but please provide an explanation why snfoundry.toml is necessary and cast cannot use Scarb toml. Imo this will be very confusing for the users

docs/src/getting-started/scarb.md
THenry14: sure, done https://github.com/foundry-rs/starknet-foundry/pull/1647/commits/1cd4ef0544e5c279ec7419d08dc4f6485e446605

PR #1636, opened 2024-01-31
Use scarb metadata error pretty print

Draggu:
Closes #1509

## Introduced changes

- Use `MetadataCommand` output, see: https://github.com/software-mansion/scarb/pull/1098

## Checklist

- [x] Linked relevant issue
- [x] Updated relevant documentation
- [x] Added relevant tests
- [x] Performed self-review of the code
- [x] Added changes to `CHANGELOG.md`

PR #1661, opened 2024-02-05
Upgrade scarb-metadata

THenry14:
## Introduced changes

- upgrade scarb-metadata to avoid problems with `missing field `experimental_features` at line 1 column 3126`

## Checklist

- [X] Linked relevant issue
- [X] Added relevant tests

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4857,9 +4857,9 @@ dependencies = [
 
 [[package]]
 name = "scarb-metadata"
-version = "1.10.0"
+version = "1.11.1"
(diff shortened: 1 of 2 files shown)


piotmag769:
We need metadata >= 1.11.0 for #1636, but metadata fix the fix will be released today. Maybe we can bump it up instead of downgrading it?

THenry14:
> We need metadata >= 1.11.0 for #1636, but metadata fix the fix will be released today. Maybe we can bump it up instead of downgrading it?

Indeed, bumped

tkumor3 (approved):

PR #1667, opened 2024-02-07
Use correct number types

### LINKED ISSUE #1477: Treat hexes as decimals, not strings
hexes are also decimals, so we should change the types to decimal in sncast (currently these are strings)

DelevoXDG:
Closes #1477 

## Introduced changes

- Rename `Hex` to `Felt`
- Change `max_fee` type from `Decimal` to `Felt`
- Change type of `Create.class_hash`: `Option<String>` -> `Option<FieldElement>`
- Change type of `Deploy.class_hash`: `Option<String>` -> `Option<FieldElement>`

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/sncast/src/starknet_commands/account/create.rs
```diff
@@ -92,8 +88,8 @@ pub async fn create(
     }
 
     Ok(AccountCreateResponse {
-        address: Hex(address),
-        max_fee: Decimal(max_fee),
+        address: Felt(address),
+        max_fee: Felt(max_fee),
```
THenry14, line 92: Does this change the default returned value format? (I mean does it print out as 0x1 instead of 1?)
I think we should print out max_fee as numbers, not hexes.

DelevoXDG: Makes sense.

We currently apply number formatting to all of the output values. The way I see it, we can create a separate class e.g. `FeltAsDecimal`, or a allow creating a `Felt` while specifying it's default string format. I don't think we should use current `Decimal`, since u64 is not guaranteed to be enough for `max_fee`.

THenry14: I meant just by default. If someone passes --int, everything should be int, and likewise when someone uses --hex everything should be hex. But, if no explicit flag was passed, we should print out addresses/hashes as hex, and max_fee (potentially others, if they make sense) as numbers.

DelevoXDG: I understand, I was rather asking about the approach that should be used 😅 
To be more memory efficient, decided to go for a separate `FeltAsDecimal` (not completely sure about the name).

THenry14: ok, I must've misunderstand you. 
I'm not sure having a separate struct, that is exaxctly the same as Felt struct, is a way to go. I think we should rather have multiple serializers instead, and use one or the other depeding on needs

crates/sncast/tests/e2e/account/create.rs
```diff
@@ -36,7 +36,8 @@ pub async fn test_happy_case() {
         std::str::from_utf8(&out.stdout).expect("failed to convert command output to string");
     assert!(stdout_str.contains("command: account create"));
     assert!(stdout_str.contains("max_fee: "));
-    assert!(stdout_str.contains("address: "));
+    assert!(!stdout_str.contains("max_fee: 0x"));
```
THenry14, line 39: Max fee seems to be duplicated (there is an assert for it in 38)
This test should assert for the default representation, right? so it probably should not start with 0x, as max_fee should be in dec representation by default

DelevoXDG: The first assert checks that `max_fee` is present, the second assert checks that that there's **no hex value** after `max_fee`

THenry14: obraz

THenry14 (approved):

PR #1669, opened 2024-02-07
Simplify resources calculation

piotmag769:
## Introduced changes

- mutable reference to used resources from `SyscallHintProcessor` used in `StarknetRuntime` for test execution are now passed to `call_entry_point` function, providing automatic resource calculation (lower layers handle it for us, we don't have to manually append resources for: call, library call and l1 handler)

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests

### DIFF
--- a/crates/cheatnet/tests/common/mod.rs
+++ b/crates/cheatnet/tests/common/mod.rs
@@ -1,10 +1,10 @@
-use blockifier::execution::common_hints::ExecutionMode;
 use blockifier::execution::entry_point::{
     CallEntryPoint, CallType, EntryPointExecutionContext, ExecutionResources,
 };
 use blockifier::execution::execution_utils::ReadOnlySegments;
 use blockifier::execution::syscalls::hint_processor::SyscallHintProcessor;
 use cairo_felt::Felt252;
+use cairo_lang_casm::hints::Hint;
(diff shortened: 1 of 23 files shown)


Arcticae (approved):

drknzz (approved):
i like 👍

crates/cheatnet/tests/common/mod.rs
```diff
@@ -103,8 +110,31 @@ pub fn call_contract(
...
+        false,
+    )
+    .unwrap();
+    let hints = HashMap::new();
+
+    let mut syscall_hint_processor = SyscallHintProcessor::new(
+        blockifier_state.blockifier_state,
+        &mut execution_resources,
+        &mut entry_point_execution_context,
+        Relocatable {
```
MaksymilianDemitraszek, line 127: Why have you selected this pointer?

piotmag769: This pointer doesn't do anything here, syscall handler is passed there only to share mut refs to resources, context and state. I can write a comment explaining that

piotmag769: I also think that this solution is not ideal, maybe another struct passing only the 3 things we actually use would be good, but idk if it wouldn't just introduce a bunch of unnecessary mess

PR #1683, opened 2024-02-08
USC raw sierra compilation demo

war-in:
# This is just a demo, it will not be merged

- USC is used for tests sierra compilation
- footer and header needs to be added manually


Arcticae (approved):
Looks mint, i think we have it  🙌

PR #1685, opened 2024-02-08
Add an instruction how to update expected version to development docs

### LINKED ISSUE #1660: Add an instruction how to update expected version to development docs.
> I would also add an instruction how to update it somewhere in our development docs.
> Also, a check before releasing in our release script, 
> "Did you update `starknet.rs` package during this release cycle? (y/n) -> What is the spec version it supports? (input...)" 
> 
> Can be added as a separate issue since it's a bit extra.

_Originally posted by @Arcticae in https://github.com/foundry-rs/starknet-foundry/pull/1635#discussion_r1473135253_

Draggu:
Closes #1660

## Introduced changes

- Add an instruction how to update expected version to development docs
- Add check in release script

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


piotmag769 (approved):
LGTM, but let's wait for @THenry14 to check bash stuff

docs/src/development/rpc-version.md
```diff
@@ -0,0 +1,5 @@
+# RPC Version (starknet.rs)
+
+If `starknet.rs` version changed, check if `rpc` version supported by it also changed.
+If so update it in `./crates/forge/expected-version`.
```
Arcticae, line 4:
```suggestion
If so update it in `./crates/forge/expected-rpc-version`.  
```

crates/forge/expected-rpc-version
```diff
@@ -0,0 +1 @@
+0.6.0
```
piotmag769, line 1: newline xD

crates/forge/expected-rpc-version
piotmag769, line 1: Actually just realised - cast team may want to reuse this solution, so prob having it in the repo root is better? Can use `project_root` to get the path

THenry14: yup, we're working on this atm, I believe the plan on our end was to hold it in a variable
@ddoktorski please take a look, maybe it is worth to sync with Piotr

scripts/release.sh
```diff
@@ -2,6 +2,15 @@
...
 
+parent_path=$( cd "$(dirname "${BASH_SOURCE[0]}")" ; pwd -P )
+echo "Was starknet.rs package updated during this release cycle?"
+select yn in "Yes" "No"; do
+    case $yn in
+        Yes ) echo "What is the spec version it supports?";read spec_version;echo $spec_version > "$parent_path/../crates/forge/expected-rpc-version";break;;
+        No ) break;;
+    esac
+done
+
```
THenry14, line 13: I don't think this is needed at all; if someone makes a change to update starkner-rs with a new spec version, and this very same person does not update said file straight away - this is a bug. Moreover, I'm pretty positive at least some of currently existing sncast tests would fail if that's the case + @ddoktorski is actually adding new tests (based on devnet) to specifically test the spec versions.

crates/forge/src/warn.rs
```diff
@@ -186,9 +187,9 @@ mod tests {
 
         assert_stdout_contains(
             stdout,
-            indoc!(
+            formatdoc!(
                 r"
-                    [WARNING] The RPC node with url = http://127.0.0.1:3030/rpc has unsupported version = (0.5.0), use node supporting RPC version 0.6.0
+                    [WARNING] The RPC node with url = http://127.0.0.1:3030/rpc has unsupported version = (0.5.0), use node supporting RPC version {EXPECTED_RPC_VERSION}
```
THenry14, line 192: Depending on where the discussion on company's slack regarding our nodes' urls will go, I'd actually suggest this test was not mocked. Since you already use real nodes for some of the tests, it would be trivial for forge/chetnet to test the functionality based on real-life output (especially so that the nodes support more than one spec version). This is something that is not very easy in scnast (devnet only supports one fixed version), so it may be worth to add this here.

crates/forge/expected-rpc-version
ddoktorski: I think it makes sense to move it to the repo root, so we will reuse it on the cast side

scripts/release.sh
ddoktorski: For further clarity, we will add a simple test to ensure the devnet spec version corresponds with our currently supported version. As a result, overlooking an update to the supported RPC version will become highly unlikely + as @THenry14 said some sncast tests probably would fail anyway.

Arcticae: If this will be clear enough (commented preferably alongside the test?) for the dev to deduce, then i am happy. That is the main concern - not to forget it and make it obsolete in the process

crates/forge/src/warn.rs
Draggu: Changed

crates/forge/expected-rpc-version
Draggu: Placed it in `shared` crate

Arcticae (changes requested):

Arcticae (approved):

PR #1689, opened 2024-02-09
Fix some of the script tests

abulenok:
Closes #

## Introduced changes

- Use the correct contract address in deployment script invoke tests
- Improve some asserts (those asserts were only checking the beginning of the error msg, which is the same for different errors)

## Checklist

- [ ] Linked relevant issue
- [x] Added relevant tests

### DIFF
--- a/crates/sncast/tests/data/scripts/invoke/src/max_fee_too_low.cairo
+++ b/crates/sncast/tests/data/scripts/invoke/src/max_fee_too_low.cairo
@@ -3,7 +3,7 @@ use starknet::{ContractAddress, Felt252TryIntoContractAddress};
 use traits::Into;
 
 fn main() {
-    let map_contract_address = 0x1e52f6ebc3e594d2a6dc2a0d7d193cb50144cfdfb7fdd9519135c29b67e427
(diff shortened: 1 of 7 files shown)


THenry14 (changes requested):
I am missing a few key pieces of information here:
- what is this PR about?
- why it was opened? (there's no issue)
- what it tries to fix, and why (did someone forgot to update hashes in tests? did the error messages change?)
- is this a fix to the problem (point above), or just a fix to the tests?

abulenok:
> I am missing a few key pieces of information here:
> 
> * what is this PR about?
> * why it was opened? (there's no issue)
> * what it tries to fix, and why (did someone forgot to update hashes in tests? did the error messages change?)
> * is this a fix to the problem (point above), or just a fix to the tests?

The problem was that the contract addresses in some tests were not valid (this is mentioned in the "introduced changes" section) and the test asserts were not able to discover this, since they were checking only the beginning of the error msg 

There is no issue because I discovered this today and made a fix instead of creating an issue

THenry14 (approved):
cool, thanks for more context, and for fixing this 👍 
hopefully this will make places where hashes need to be changed more discoverable

Issue #1693, opened 2024-02-12
Fix wrong profile being passed to scarb


THenry14:
There is a bug, where we pass a cast profile to scarb to load artifact and since the profile might not exist in scarb - we get an error. Happens for declare.

PR #1695, opened 2024-02-12
Fix passing non-existent profile to scarb

THenry14:
Closes https://github.com/foundry-rs/starknet-foundry/issues/1693

## Introduced changes

- fixes a bug, where we allowed passing non existent profile to scarb, which resulted with missing artifacts error

## Checklist

- [X] Linked relevant issue
- [X] Added relevant tests


piotmag769 (approved):
Do we want to release a hotfix with that? It is kinda an annoying bug tbh that makes declare fail for all profiles expect dev and default

crates/sncast/src/helpers/scarb_utils.rs
```diff
@@ -154,7 +154,10 @@ pub fn build_and_load_artifacts(
     build(package, config).map_err(|e| anyhow!(format!("Failed to build using scarb; {e}")))?;
 
     let metadata = get_scarb_metadata_with_deps(&config.scarb_toml_path)?;
-    get_contracts_map(&metadata, &package.id, Some(&config.profile))
+    match &metadata.profiles.contains(&config.profile) {
```
ddoktorski, line 157: How about displaying a warning message regarding the Scarb profile being used in the following scenarios?
1. No profile has been provided 
2. The provided profile does not exist

piotmag769: I think a message informing what scarb profile is used is a good thing in general

THenry14:
> Do we want to release a hotfix with that? It is kinda an annoying bug tbh that makes declare fail for all profiles expect dev and default

yeah, I think so. release would work as well, but I think most people do not have this tightly coupled

crates/sncast/src/helpers/scarb_utils.rs
THenry14: Yeah, makes sense, I'd do it only for profiles that do not exist in scarb (as I don't think we should inform when profile is not set, scarb does not inform about its default and neither should we imo).

Something like: "Profile <name> does not exist in scarb, using the default "dev" profile" ?

ddoktorski: Sounds good

THenry14: added warn message in https://github.com/foundry-rs/starknet-foundry/pull/1695/commits/446fbb3b942438cfbcfe4c972ba27074a85c5905

piotmag769 (approved):

Issue #1016, opened 2023-10-31
Design the single-call spanning cheatcode variants


Arcticae:
This should include:
- warp
- roll
- prank
- spoof
We can thing of an alternative solution to foundry since it's not that obvious and might be confusing for users

piotmag769:
Consider implementing n_times versions instead

PR #1703, opened 2024-02-13
Cheat n times

drknzz:
#1016

## Introduced changes

- Change signature of `start_cheat` to include **CheatSpan**
```rust
pub enum CheatSpan {
    Indefinite,
    Number(usize),
}

pub fn start_cheat<T: Clone, S: BuildHasher>(
    global_cheat: &mut Option<(T, CheatSpan)>,
    contract_cheats: &mut HashMap<ContractAddress, CheatStatus<T>, S>,
    target: CheatTarget,
    cheat_value: T,
+   span: CheatSpan,
)
```

- Add `CheatedData` to call stack
```rust
pub struct CheatedData {
    pub block_number: Option<Felt252>,
    pub block_timestamp: Option<Felt252>,
    pub caller_address: Option<ContractAddress>,
    pub sequencer_address: Option<ContractAddress>,
    pub tx_info: Option<TxInfoMock>,
}

struct CallStackElement {
    // when we exit the call we use it to calculate resources used by the call
    resources_used_before_call: ExecutionResources,
    call_trace: Rc<RefCell<CallTrace>>,
+   cheated_data: CheatedData,
}
```

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/cheatnet/src/state.rs
```diff
@@ -121,10 +140,21 @@ impl StateReader for ExtendedStateReader {
...
 
 pub enum CheatStatus<T> {
-    Cheated(T),
+    Cheated(T, CheatSpan),
     Uncheated,
 }
 
+impl<T> CheatStatus<T> {
+    pub fn decrement_cheat_span(&mut self) {
+        if let CheatStatus::Cheated(_, CheatSpan::Number(n)) = self {
```
Arcticae, line 149: This should probably panic otherwise, or return result

crates/cheatnet/src/state.rs
```diff
@@ -36,6 +37,24 @@ pub enum CheatTarget {
...
 }
 
+// Specifies the duration of the cheat
+#[derive(Clone)]
+pub enum CheatSpan {
+    Indefinite,
+    Number(usize),
+}
+
+impl Add<usize> for CheatSpan {
```
Arcticae, line 47: i don't think i see a use-case for this

Arcticae (approved):
LGTM

crates/cheatnet/src/state.rs
drknzz: Good catch, I've removed all of them and forgot about it 👍

crates/cheatnet/src/state.rs
drknzz: That's actually the logic on this - `CheatStatus` becomes `Uncheated`, once its cheat span goes down to 0.

piotmag769 (commented):
Request again if needed after introducing proposed changes

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/cheated_syscalls.rs
```diff
@@ -47,15 +48,18 @@ pub fn get_execution_info_syscall(
...
 ) -> SyscallResult<GetExecutionInfoResponse> {
     let execution_info_ptr = syscall_handler.get_or_allocate_execution_info_segment(vm)?;
+    let current_call_stack = &mut runtime_state.cheatnet_state.trace_data.current_call_stack;
 
-    let contract_address = syscall_handler.storage_address();
+    let cheated_data = if current_call_stack.size() == 1 {
+        CheatedData::new(
+            runtime_state.cheatnet_state,
+            &syscall_handler.storage_address(),
+        )
```
piotmag769, line 57: This way every `get_caller_address` etc. done directly in the test code will count as one call for the purpose of our cheatcodes (the constructor modifies the cheatnet state, commented below), I don't think we want this. In fact there is no way to call the test code contract from any other contract: 

https://github.com/foundry-rs/starknet-foundry/blob/9a1b4bc0c85c99f2ec21eb62c0ec99b8ea310e54/crates/cheatnet/src/constants.rs#L63)

I think we should prevent cheating `test_address` with any span but `Indefinite`. Maybe we can do it on Cairo level and make assertions in our Rust code - I don't really like coupling Cairo and Rust that much, but I have no better idea, feel free to explore possible solutions.
cc @Arcticae

piotmag769 (changes requested):

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/cheated_syscalls.rs
drknzz: Why is this not what we want? 

Maybe it's not directly visible without tests, but the current behaviour of `CheatTarget::All`, is cheating each address n times separately. It does not share `CheatSpan`'s between addresses.

// This has been refactored via the rest of the comments, but the logic stayed the same

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/entry_point.rs
```diff
@@ -33,12 +33,29 @@ pub fn execute_call_entry_point(
...
     context: &mut EntryPointExecutionContext,
 ) -> EntryPointExecutionResult<CallInfo> {
+    let cheated_data = if let CallType::Delegate = entry_point.call_type {
+        runtime_state
+            .cheatnet_state
+            .trace_data
+            .current_call_stack
+            .top_cheated_data()
+            .clone()
+    } else {
```
piotmag769, line 43: So we just don't want to treat the library call as calling another entrypoint? It goes through the Starknet OS layer, but I guess it makes sense? cc @Arcticae

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/entry_point.rs
```diff
@@ -33,12 +33,29 @@ pub fn execute_call_entry_point(
...
+        let contract_address = &entry_point.storage_address;
+        let cheated_data_ = runtime_state
```
piotmag769, line 45: Weird underscore xD

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/entry_point.rs
```diff
@@ -33,12 +33,24 @@ pub fn execute_call_entry_point(
...
+    let cheated_data = if let Some(contract_address) = &entry_point.code_address {
+        CheatedData::new(runtime_state.cheatnet_state, contract_address)
+    } else {
+        runtime_state
+            .cheatnet_state
+            .trace_data
+            .current_call_stack
+            .top_cheated_data()
+            .clone()
+    };
```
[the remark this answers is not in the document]
piotmag769: Yup, it should still be pranked

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/cheated_syscalls.rs
piotmag769: > Maybe it's not directly visible without tests, but the current behaviour of CheatTarget::All, is cheating each address n times separately. It does not share CheatSpan's between addresses.

I don't understand what you mean, don't we set a separate var with global cheat? Also I don't see how is it relevant tbh

Nonetheless I still don't think we should decrease the counter if it will be `Indefinite` every time, it is just an operation that does nothing

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/entry_point.rs
drknzz: How is it weird xD

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/cheated_syscalls.rs
drknzz: I'm not a huge fan of constraining the domain in specific cases, I'd rather allow for any CheatSpan on test_address, just don't update the cheats

drknzz: [Do not update cheats for test address](https://github.com/foundry-rs/starknet-foundry/pull/1703/commits/1105f407b309c86ee8182aceae4819a84d9fe0d1)

piotmag769: >I'm not a huge fan of constraining the domain in specific cases, I'd rather allow for any CheatSpan on test_address, just don't update the cheats

Agree, that's a better idea, we can still constraint them on the Cairo level later on, right?

piotmag769 (approved):
LGTM

Issue #1690, opened 2024-02-09
Implement `.env` based fetching of node URL for snfoundry CI


Arcticae:
### Which components does the task require to be changed? (think hard pls)

snforge

### Description

It should be stored locally in .env file and on CI in secrets

We need to make sure that it's also accessible to forking repos, when we approve the run ourselves

THenry14:
cargo configs could potentially be used for this (instead of env files) as well https://doc.rust-lang.org/cargo/reference/config.html

Arcticae:
@MaksymilianDemitraszek please describe the proposed solution here

MaksymilianDemitraszek:
Foundry should be able to resolve environmental variables from both `Scarb.toml` and `snfoundry.toml`. It means that
a value `field="$ENVVAR"`, is read from scarb config and `snfoundry.toml` as a string `"$ENVVAR"`. Later foundry should resolve all environmental variables to its values by reading and replacing them.

tomek0123456789:
Closed in #2003, #2004 and #2006

PR #1635, opened 2024-01-31
Add rpc version assertion

Draggu:
Closes #1561

## Introduced changes

- Added warning when rpc version is incompatible

## Checklist

- [x] Linked relevant issue
- [x] Updated relevant documentation
- [x] Added relevant tests
- [x] Performed self-review of the code
- [x] Added changes to `CHANGELOG.md`

PR #1704, opened 2024-02-14
Add incompatible RPC version warning to sncast

### LINKED ISSUE #1560: Spec version assertion
A lot of users report problems trying to old rpc endpoint. Cast should handle such case gently and print a warning in such case.

ddoktorski:
Closes #1560 

## Introduced changes

- Print a warning message about the incompatible RPC version when executing `sncast` commands

Note: The logic is partially tested on the forge side. For comprehensive testing in sncast, we would need to either use one of our nodes or use mocks. We may also consider moving more code to the shared crate, please share your thoughts on any improvements in this regard.

## Checklist

- [X] Linked relevant issue
- [ ] Added relevant tests


crates/forge/src/warn.rs
```diff
@@ -57,28 +56,15 @@ pub(crate) async fn warn_if_incompatible_rpc_version(
...
     }
 
     // assert version
     for handle in handles {
         let (version, url) = handle.await;
-        let version = version??;
+        let version = version?;
 
-        if !expected_version.matches(&version) {
+        if !is_supported_version(&version) {
```
DelevoXDG, line 67: I wonder what happens when `--json` flag is passed? Since the warnings aren't intended to abort the command, I think storing those in some `warnings` sequence would be better.

crates/sncast/src/lib.rs
```diff
@@ -86,6 +88,20 @@ pub fn get_provider(url: &str) -> Result<JsonRpcClient<HttpTransport>> {
...
 }
 
+pub async fn warn_if_incompatible_rpc_version(
+    provider: &JsonRpcClient<HttpTransport>,
+    url: &str,
+) -> Result<()> {
+    let node_spec_version = get_and_parse_spec_version(provider).await?;
+    if !is_supported_version(&node_spec_version) {
+        print_as_warning(&anyhow!(
+            "The RPC node with the url {url} is using an unsupported version {node_spec_version}. The current version supported by sncast is {EXPECTED_RPC_VERSION}"
```
DelevoXDG, line 98: I wonder if we should print that warning if the major version is correct. Especially considering there might be just minors difference between versions returned by the node e.g. `0.5.0rc4`, `0.5.0` and `0.5.1`

DelevoXDG (commented):
Both mocks and nodes seem like a good idea for sncast tests. I do lean towards node tests, with only concern that we might not have an available node with an unsupported version, but I suppose this is highly unlikely.

crates/sncast/src/lib.rs
ddoktorski: I guess that nodes would return `0.5.0` even for cases like `0.5.0rc4`, but I am not sure. Outputting that warning seems reasonable even when the difference is only in the minor version, as it can still result in unpredictable behavior.

DelevoXDG: > I guess that nodes would return `0.5.0` even for cases like `0.5.0rc4`, but I am not sure. 

Not really, they sometimes do output rc version.

> Outputting that warning seems reasonable even when the difference is only in the minor version, as it can still result in unpredictable behavior.

Maybe you're right, just that warning is not stating that version "might be incompatible", but that it's "unsupported", which I'm not sure is often correct for minor versions

crates/forge/src/warn.rs
ddoktorski: When the `--json` flag is passed, it is ignored by this warning. If we decide to modify this behavior, we also need to adjust this for the other [warning](https://github.com/foundry-rs/starknet-foundry/blob/master/crates/sncast/src/starknet_commands/script.rs#L261) about the `sncast_std` version that appears when executing the script command.

crates/shared/src/lib.rs
```diff
@@ -1 +1,22 @@
...
 pub mod consts;
+
+#[must_use]
+pub fn is_supported_version(version: &Version) -> bool {
+    let expected_version =
+        Version::parse(EXPECTED_RPC_VERSION).expect("Failed to parse the expected RPC version");
+    *version == expected_version
+}
+
+pub async fn get_and_parse_spec_version(client: &JsonRpcClient<HttpTransport>) -> Result<Version> {
```
piotmag769, line 15:
```suggestion
pub async fn get_rpc_version(client: &JsonRpcClient<HttpTransport>) -> Result<Version> {
```
[nit] The signature already tells it's parsed

piotmag769, line 15: Would be cleaner to put it in `mod rpc` or sth

crates/sncast/tests/integration/lib_tests.rs
```diff
@@ -136,3 +137,10 @@ async fn test_get_account_failed_to_convert_field_elements() {
         .to_string()
         .contains("Failed to convert account address = address to FieldElement"));
 }
+
+#[tokio::test]
+async fn test_supported_rpc_version_matches_devnet_version() {
+    let provider = create_test_provider();
+    let devnet_spec_version = get_and_parse_spec_version(&provider).await.unwrap();
+    assert!(is_supported_version(&devnet_spec_version));
+}
```
piotmag769, line 146: It doesn't really test anything from sncast, it should be a test for `shared` crate tbh

crates/sncast/src/lib.rs
piotmag769: Agree with Maksim here

crates/sncast/src/lib.rs
```diff
@@ -86,6 +88,20 @@ pub fn get_provider(url: &str) -> Result<JsonRpcClient<HttpTransport>> {
     Ok(provider)
 }
 
+pub async fn warn_if_incompatible_rpc_version(
+    provider: &JsonRpcClient<HttpTransport>,
+    url: &str,
+) -> Result<()> {
```
piotmag769, line 94: I think this function could be moved to shared crate and used both in sncast and snforge

crates/forge/src/warn.rs
THenry14: I agree, ideally warnings should be printed to stderr, to allow users to filter them. Let's do this (all of them use the same function, so this should be rather easy).

I believe, however, that the issue is may not be isolated to these particular cases. Now this should be verified with scarb team, but it seems some of the commands (like declare) may also break json, because scarb returns multiple objects (only last json is from sncast):

```
{"status":"compiling","message":"lib(map) map v0.1.0 (/Users/wojciech/git-repos/starkware/starknet-foundry/crates/sncast/tests/data/contracts/map/Scarb.toml)"}
{"type":"warn","message":"Sierra, textual Sierra and CASM lib targets have been disabled, Scarb will not produce anything"}
{"status":"compiling","message":"starknet-contract(map) map v0.1.0 (/Users/wojciech/git-repos/starkware/starknet-foundry/crates/sncast/tests/data/contracts/map/Scarb.toml)"}
{"status":"finished","message":"release target(s) in 1 second"}
{"class_hash":"0x43a9554442216b8c5b551c6cae6357ce69fd9838e9bf40fb27b455cac5ba9f4","transaction_hash":"0x21d5dfe1854c294b371632ea9f2435017d0373f8adc6d4194c88a16c87eb0c","command":"declare"}
```

we should take a closer look at this under different issue. We must also decide what to do with our `--wait` function (under the same issue), as it also uses `println` to inform about tx hash, so it can also potentially break json. Let me know if you will make one, or if you want me to do this.

crates/sncast/src/lib.rs
THenry14: I'd change the wording from `unsupported` to `incompatible` as well, tbh. 
But I'd still display it for minor, patch and rc versions, because (and I assume you meant `minor` and not `major`, as major has always been `0`, and will be just that for the forseeable future):

- generally rc versions should not make their way to nodes used by "normal" people (these are rcs, so should rather be used for test purposes only ad bugs are expected)
- sw are rather relaxed about their semver, and they also introduce braking changes to patch versions. so our check would be worthless in that case - I'd rather be more explicit.

crates/sncast/src/main.rs
```diff
@@ -138,6 +138,9 @@ fn main() -> Result<()> {
         )
         .expect("Failed to build script");
         let metadata_with_deps = get_scarb_metadata_with_deps(&manifest_path)?;
+
+        runtime.block_on(warn_if_incompatible_rpc_version(&provider, &config.rpc_url))?;
```
THenry14, line 142: let's move this higher up, right after config & provider, becase if there are incompatible versions they may result in errors, so we may never even see this warning.

crates/sncast/tests/integration/lib_tests.rs
THenry14: I guess the idea was that this is the only check we can do in sncast (as we only use devnet for tests), so if we move the function to shared crate - I agree.

crates/sncast/src/lib.rs
ddoktorski: Does it make sense to create `mod print` int the `shared` crate and move `print_as_warning` there, so that it can be used in both forge and cast?

crates/sncast/tests/integration/lib_tests.rs
ddoktorski: The main rationale behind this decision was to avoid setting up devnet in the shared crate solely for this test. Although we could potentially run a similar test using a real node, it would currently require hardcoding node URLs in an additional location, so it's rather better to wait until #1690. 

For now we can either keep the test here or move it somewhere to the forge tests (given that we already have hardcoded urls there). Once we have envs in place we can add this test in the shared crate. Wdyt?

crates/forge/src/warn.rs
ddoktorski: Discussed offline with @THenry14 that he will create an issue for this and link it here.

crates/sncast/tests/integration/lib_tests.rs
piotmag769: Tbh I don't see an issue with hardcoing it in one more place and then using secrets when #1690 when it's introduced, it's easier this way then to remember to move the test imo

piotmag769 (approved):

crates/shared/src/rpc.rs
```diff
@@ -0,0 +1,28 @@
...
+    let expected_version =
+        Version::parse(EXPECTED_RPC_VERSION).expect("Failed to parse the expected RPC version");
+    *version == expected_version
+}
+
+pub async fn get_rpc_version(client: &JsonRpcClient<HttpTransport>) -> Result<Version> {
+    client
+        .spec_version()
+        .await
+        .context("Error while calling RPC node")?
```
DelevoXDG, line 25: Nit: not sure this is the best error message

DelevoXDG (approved):

crates/forge/src/pretty_printing.rs
```diff
@@ -5,11 +5,6 @@ use forge_runner::{test_case_summary::AnyTestCaseSummary, test_crate_summary::Te
 use starknet_api::block::BlockNumber;
 use std::collections::HashMap;
 
-pub fn print_warning(error: &Error) {
-    let warning_tag = style("WARNING").color256(11);
-    println!("[{warning_tag}] {error}");
-}
-
 pub fn print_error_message(error: &Error) {
```
THenry14, line 8: [nitpick]: Let's open an issue to move other print methods to shared create as well :+1: 
Depending on if we might want to use it i sncast, it should be assigned to either one of forge or chain interactions team.

crates/forge/tests/e2e/fork_warning.rs
```diff
@@ -66,7 +66,7 @@ fn should_print_warning() {
             r"
                 [..]Compiling[..]
                 [..]Finished[..]
-                [WARNING] The RPC node with url = http://188.34.188.184:9545/rpc/v0_5 has unsupported version = (0.5.1), use node supporting RPC version ({EXPECTED_RPC_VERSION})
+                [WARNING] RPC node with the url http://188.34.188.184:9545/rpc/v0_5 uses incompatible version 0.5.1. Expected version: {EXPECTED_RPC_VERSION}
```
THenry14, line 69: [nitpick]: I believe this is forge's convention to use `=` extensively in their messages, so we might want to leave those here. If I remember correctly, as a part of consistency between forge and cast, they also tried to incorporate those to our messages some time ago.

crates/forge/src/warn.rs
THenry14: https://github.com/foundry-rs/starknet-foundry/issues/1723

crates/sncast/tests/integration/lib_tests.rs
THenry14: > Tbh I don't see an issue with hardcoing it in one more place

Noo, let's not do this, please :( I think the best at the moment would be to move all those tests altogether to forge, and deal with extracting them to shared crate when the variable is taken out to env. I would much prefer this to leaving this test (with devnet) here.

THenry14 (commented):
no explicit approval from me, as I think that the best would be to move the devnet test out to forge (with a use of real node rpc) so everything is in one place. Apart from that, lgtm

crates/sncast/tests/integration/lib_tests.rs
piotmag769: Up to u, but please make sure this test is moved to the shared crate after #1690

THenry14: yup, just as I said :+1:

Draggu:
Realized that there was a missing `tokio::spawn()` call in the `warn_if_incompatible_rpc_version()` function.
It was like this from the beginning https://github.com/foundry-rs/starknet-foundry/pull/1635.

crates/sncast/tests/integration/lib_tests.rs
ddoktorski: As discussed offline, we've agreed to leave this test as is for now. I've put a TODO note for it, and will make sure that we move it to the shared crate as soon as #1690 is done.

crates/forge/src/pretty_printing.rs
ddoktorski: https://github.com/foundry-rs/starknet-foundry/issues/1740 - assigned to `chain-interactions` for now

crates/forge/tests/e2e/fork_warning.rs
ddoktorski: After taking a brief look, it doesn't seem to be very consistent throughout forge.

@piotmag769 can you confirm whether you'd prefer using `=` in this warning? Personally I don't like it, but if this is a convention I can change it.

THenry14: I'm fine with whatever, if noone from forge objects, this can be as is.

THenry14 (approved):
great job mate

PR #1709, opened 2024-02-14
Execute sncast tests in tempdir

### LINKED ISSUE #1247: Execute Cast tests in tempdir

DelevoXDG:
Closes #1247 

## Introduced changes

- Use temp dirs in sncast create tests
  - Update related fixtures and helper functions
- Minor refactor
  - Replace `TempDir::new` calls with `tempdir()`
  - Replace Command::new calls with runner calls

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/sncast/src/helpers/configuration.rs
```diff
@@ -4,7 +4,7 @@ use anyhow::{anyhow, Result};
 use camino::Utf8PathBuf;
 use serde::{Deserialize, Serialize};
 use std::fs;
-use tempfile::TempDir;
+use tempfile::{tempdir, TempDir};
```
THenry14, line 7:
```suggestion
use tempfile::tempdir;
```

This is no longer used, right?

crates/sncast/tests/e2e/account/add.rs
```diff
@@ -3,14 +3,13 @@ use crate::helpers::runner::runner;
...
 use std::fs;
-use tempfile::TempDir;
+use tempfile::tempdir;
 
 #[tokio::test]
 pub async fn test_happy_case() {
-    let accounts_file = "./tmp-a1/accounts.json";
-    _ = fs::remove_file(accounts_file);
+    let tempdir = tempdir().expect("Unable to create a temporary directory");
+    let accounts_file = "./accounts.json";
```
THenry14, line 12: nit: later on you join tempdir with accounts_file, yielding "/temp/path/to/./accounts.json", which is fine, but not very elegant.

THenry14: applicable to all the places you do this

crates/sncast/tests/e2e/account/create.rs
```diff
@@ -337,33 +329,52 @@ pub async fn test_keystore_without_account() {
...
 
-#[test_case("tests/data/keystore/my_key.json", "tests/data/keystore/my_account_new.json", "error: Keystore file tests/data/keystore/my_key.json already exists" ; "when keystore exists")]
-#[test_case("tests/data/keystore/my_key_new.json", "tests/data/keystore/my_account.json", "error: Account file tests/data/keystore/my_account.json already exists" ; "when account exists")]
+#[test_case("./tests/data/keystore/my_key.json", "./tests/data/keystore/my_account_new.json", "error: Keystore file my_key.json already exists" ; "when keystore exists")]
+#[test_case("./tests/data/keystore/my_key_new.json", "./tests/data/keystore/my_account.json", "error: Account file my_account.json already exists" ; "when account exists")]
 pub fn test_keystore_already_exists(keystore_path: &str, account_path: &str, error: &str) {
+    let temp_dir = tempdir().expect("Unable to create a temporary directory");
+
+    let keystore_path = Path::new(keystore_path);
+    if keystore_path.exists() {
```
THenry14, line 346: why do we need to check if this exists?

crates/sncast/tests/e2e/account/create.rs
```diff
@@ -337,33 +329,52 @@ pub async fn test_keystore_without_account() {
...
+        )
+        .expect("Unable to copy account file");
+    }
+
     let args = vec![
         "--url",
         URL,
         "--keystore",
-        keystore_path,
+        keystore_path.file_name().unwrap().to_str().unwrap(),
```
THenry14, line 368: we could just construct a str representation of the path straight away, since we do not need Path representation, right? this could simplify such lines, I guess

crates/sncast/tests/e2e/multicall/run.rs
```diff
@@ -1,22 +1,39 @@
...
-    let path = project_root::get_project_root().expect("failed to get project root path");
-    let path = Path::new(&path)
-        .join(MULTICALL_CONFIGS_DIR)
-        .join("deploy_invoke.toml");
-    let path_str = path.to_str().expect("failed converting path to str");
-
-    args.append(&mut vec!["multicall", "run", "--path", path_str]);
-
-    let snapbox = runner(&args);
+    let temp_dir = tempdir().expect("Unable to create temporary directory");
```
THenry14, line 9: is there any gain from using tempdirs for multicalls? I was under impression that they do not tent to change fs in any way, so maybe it is wasteful to do this?

crates/sncast/tests/e2e/main_tests.rs
```diff
@@ -33,6 +32,8 @@ async fn test_happy_case_from_sncast_config() {
 
 #[tokio::test]
 async fn test_happy_case_from_cli_no_scarb() {
+    let tempdir = tempdir().expect("Unable to create temporary directory");
```
THenry14, line 35: this is kid of pointless, as call does not modify fs, and does not depend in it in any meaningful way (eg reading hashes/addresses etc)

crates/sncast/tests/e2e/script/call.rs
```diff
@@ -1,15 +1,17 @@
 use crate::helpers::constants::{SCRIPTS_DIR, URL};
```
THenry14, line 1: the same comment as above - let's maybe avoid creating tempdirs in unnecessary places like call etc. We should only use tempdirs, when we need to alter fs or depend on some salted contracts/data

crates/sncast/tests/helpers/fixtures.rs
```diff
@@ -359,6 +355,41 @@ pub fn duplicate_script_directory(
     let mut file = File::create(manifest_path).expect("Failed to create file");
     file.write_all(modified_toml.as_bytes())
         .expect("Failed to write to file");
+}
+
+pub fn duplicate_script_directory(
```
THenry14, line 360: why do we need copy script directory function, and duplicate script directory function? they sound very redundant

crates/sncast/tests/helpers/fixtures.rs
```diff
@@ -359,6 +355,41 @@ pub fn duplicate_script_directory(
...
+
+    let dest_dir = Utf8PathBuf::from_path_buf(temp_dir.path().to_path_buf())
+        .expect("Failed to create Utf8PathBuf from PathBuf");
+
+    copy_script_directory(&src_dir, dest_dir, deps);
+
+    temp_dir
+}
+
+pub fn duplicate_workspace_directory(
```
THenry14, line 374: why do we need this? by the name I assume you just want to copy a directory, so why not use `fs` or some other crate?

THenry14 (changes requested):

crates/sncast/src/helpers/configuration.rs
DelevoXDG: Used in:
```rs
pub fn copy_config_to_tempdir(src_path: &str, additional_path: Option<&str>) -> TempDir {
```

crates/sncast/tests/e2e/account/add.rs
DelevoXDG: This is the way it was done before (e.g. in `test_happy_case_add_profile`), so I assumed it was done for readability purposes. Should I change this everywhere as a part of this PR?

crates/sncast/tests/e2e/account/create.rs
DelevoXDG: To copy only those files that exist, as the test cases are the following:
1. Keystore file already exists; Account files does not exist
2. Keystore file does not exist; Account file already exists

crates/sncast/tests/e2e/account/create.rs
DelevoXDG: We use a Path representation for `.exists()` method

crates/sncast/tests/e2e/multicall/run.rs
DelevoXDG: There were always some test cases that relied on reading the fs, even though they didn't modify it, and these tests still utilized tempdirs. 

I thought that for consistency we should then proceed and use temp dirs everywhere we interact with fs in the tests, aside from for example `--accounts-file` that is tested separately.

crates/sncast/tests/helpers/fixtures.rs
DelevoXDG: We want to copy the specified members and their deps, including `sncast_std` (use `duplicate_script_directory` for every member)

crates/sncast/tests/helpers/fixtures.rs
DelevoXDG: In `duplicate_workspace_directory`, I needed to reuse part of the code from `duplicate_script_directory`, but without creating a new temp dir, so that part was moved to `copy_script_directory`

crates/sncast/tests/e2e/account/create.rs
THenry14: Let's split this test into two then, to keep test cases as simple as possible (withouth ifs etc)

crates/sncast/tests/e2e/account/create.rs
THenry14: we can use absolute path though, right? it should not matter for the test case, until the file exists

crates/sncast/tests/e2e/multicall/run.rs
THenry14: reading fs is fine (unless we need salted contracts), it's the writing that may be troublesome and should use tempdirs. Creating a tempdir just to read a file (and that read does not conflict with anything) is in my opinion excessive, potentially it can slow down our tests a bit

crates/sncast/tests/e2e/main_tests.rs
THenry14: But since the test is being executed by default in `repo/crates/sncast/` there is very little chance there is a config file there, right? (it shouldn't be, and it won't ever be in repo).
To properly make sure about the assumption you've mentioned, we could rather assert for arguments that were used in cli, but I don't think this is necessary tbh

crates/sncast/tests/e2e/script/call.rs
THenry14: ahh shit, this is `script call` test, and my dumb ass thought this was `subcommand call` test. Sorry, you're right. This should be left as is.

crates/sncast/tests/helpers/fixtures.rs
THenry14: Let's then rename the functions to properly indicate what they do - as `duplicate_directory` and `copy_directory` on the face of it does mean exactly the same thing

crates/sncast/tests/helpers/fixtures.rs
THenry14: I'm not sure why we need a distinction between `script` and `workspacez directory, it seems to me this could be a one function tbh

crates/sncast/tests/e2e/account/deploy.rs
```diff
@@ -330,22 +346,34 @@ pub async fn test_happy_case_keystore() {
...
+
+    let keystore_path = "./my_key.json";
+    let account_path = "./account.json";
+
+    fs_extra::file::copy(
+        "tests/data/keystore/my_key.json",
+        tempdir.path().join(keystore_path),
+        &fs_extra::file::CopyOptions::new().overwrite(true),
+    )
+    .expect("Unable to copy keystore file");
```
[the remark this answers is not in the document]
THenry14: sure, especially if we can reuse this in other places as well

crates/sncast/tests/e2e/multicall/run.rs
DelevoXDG: Ok, that makes sense. I wonder if we should then remove temp dirs in tests that had them prior to my changes, but didn't alter fs.

crates/sncast/tests/helpers/fixtures.rs
DelevoXDG: How about `copy_script_directory` and `copy_script_directory_to_tempdir`?

crates/sncast/tests/helpers/fixtures.rs
DelevoXDG: It doesn't seem to me we can merge it into one function, they have different structures: for workspace we need to specify `members`, where each member is a script, which is something we don't do for scripts.

crates/sncast/tests/e2e/multicall/run.rs
THenry14: Do we have such tests? It may make sense, but I would be inclined to maybe do this in different PR, if the change is big

crates/sncast/tests/helpers/fixtures.rs
THenry14: makes sense

crates/sncast/tests/helpers/fixtures.rs
THenry14: hmm, ok I'm convinced. :+1:

crates/sncast/tests/e2e/multicall/run.rs
DelevoXDG: > Do we have such tests?

Yes, for example `main_tests::test_happy_case_from_sncast_config`

> It may make sense, but I would be inclined to maybe do this in different PR, if the change is big

That makes sense, I'm just a bit unsure about what should be our decision process for picking which tests to execute in tempdir and which not to, both for my changes and tests that already used temp dirs. How should we handle:
1. Tests which are supposed to change fs, but since we expect the error, they never do (e.g. `account::deploy::test_deploy_keystore_no_status`)
2. Same kind of tests, but specifically those that use copied contract dirs (e.g. `declare::scarb_build_fails_manifest_does_not_exist`)

THenry14: > Tests which are supposed to change fs, but since we expect the error, they never do (e.g. account::deploy::test_deploy_keystore_no_status)

But they might if test fails (so tx for some reason goes through), so this should be in tempdir imo

> Same kind of tests, but specifically those that use copied contract dirs (e.g. declare::scarb_build_fails_manifest_does_not_exist)

temdpris as well, basically the same scenario as 1. I suppose all declare test should use tempdirs, as they also build contracts/scripts. For other tests - it depends on what they actually do/can do

THenry14 (approved):

PR #1715, opened 2024-02-15
Fix cairo0 cheatcode cancelling behavior

### LINKED ISSUE #1632: Fix cairo0 cheatcode-cancelling behavior

snforge

### Description

Scenario:
Given contract A (cairo 1), and contract B (cairo 0), we have a following interaction:
`A -> B -> A`
after B interacts with A, there is no cheating capabilities in A itself

Draggu:
Closes #1632

## Introduced changes

- Fixed cairo0 cheatcode cancelling behavior for `call_contract` and `deploy`

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Forge
+
+#### Fixed
+
+- Calling Cairo 0 contract does not longer cancel cheatcodes
```
piotmag769, line 14:
```suggestion
- Calling Cairo 0 contract does not longer cancel cheatcodes in further calls
```

crates/cheatnet/tests/contracts/src/prank/prank_checker_cairo0.cairo
```diff
@@ -0,0 +1,49 @@
...
+
+    #[storage]
+    struct Storage {}
+
+    #[abi(embed_v0)]
+    impl ICairo1ContractImpl of super::ICairo1Contract<ContractState> {
+        fn start(
+            ref self: ContractState,
+            cairo0_address: ContractAddress,
+            pranked_address: ContractAddress
```
piotmag769, line 30:
```suggestion
            expected_caller_address: ContractAddress
```
pranked address is the address of the contract that was pranked i.e. address of this contract

crates/cheatnet/tests/contracts/src/prank/prank_checker_cairo0.cairo
```diff
@@ -0,0 +1,49 @@
...
+
+            assert(pranked_address == get_caller_address(), 'address should be pranked');
+
+            cairo0_contract.callback(contract_address.into());
+
+            assert(pranked_address == get_caller_address(), 'address should be pranked');
+        }
+
+        fn end(self: @ContractState) {
+            assert(123.try_into().unwrap() == get_caller_address(), 'should be same');
```
piotmag769, line 46: It leaks the logic. I think the best idea (since the cairo 0 contract is already implemented) would be do write the expected caller address to the storage in `start` function. Then here you could read it and assert it is correct.

crates/cheatnet/tests/contracts/src/prank/prank_checker_cairo0.cairo
```diff
@@ -0,0 +1,49 @@
...
+            };
+
+            assert(pranked_address == get_caller_address(), 'address should be pranked');
+
+            cairo0_contract.callback(contract_address.into());
+
+            assert(pranked_address == get_caller_address(), 'address should be pranked');
+        }
+
+        fn end(self: @ContractState) {
```
piotmag769, line 45: [nit] To make sure this entrypoint was called by the cairo 0 contract (in theory we don't know if it executed) we could write an arbitrary value to a storage in `end` and then read it in `start` and assert it. But maybe that's just me being picky xD

piotmag769 (commented):
Very solid work, good job! 🚀 
Should we add a test for the `deploy` flow too e.g., cairo 0 contract doing a deploy preferably of another cairo 0 (test code -> cairo 1 -> cairo 0 -> cairo 1 flow is already tested) contract and checking if the constructor is pranked? Or is it excessive? cc @Arcticae

Arcticae:
@piotmag769 i think this is good enough

Arcticae (approved):

piotmag769 (approved):

PR #1720, opened 2024-02-15
Introduce `ValidWaitParams`

abulenok:
Closes #

## Introduced changes

- ValidWaitParams::new ensures that wait_timeout and wait_retries are valid
- Created as a result of this discussion https://github.com/foundry-rs/starknet-foundry/pull/1710#discussion_r1490813951 

## Checklist

- [ ] Linked relevant issue
- [ ] Added relevant tests

### DIFF
--- a/crates/sncast/src/helpers/configuration.rs
+++ b/crates/sncast/src/helpers/configuration.rs
@@ -1,5 +1,5 @@
 use super::constants::{CONFIG_FILENAME, WAIT_RETRY_INTERVAL, WAIT_TIMEOUT};
-use crate::ValidWaitParams;
+use crate::ValidatedWaitParams;
 use anyhow::{anyhow, Result};
 use camino::Utf8PathBuf;
 use serde::{Deserialize, Serialize};
@@ -13,7 +13,7 @@ pub struct CastConfig {
     pub account: String,
     pub accounts_file: Utf8PathBuf,
     pub keystore: Option<Utf8PathBuf>,
-    pub wait_params: ValidWaitParams,
+    pub wait_params: ValidatedWaitParams,
 }
 
 impl CastConfig {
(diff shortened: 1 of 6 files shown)


crates/sncast/src/helpers/configuration.rs
```diff
@@ -1,19 +1,19 @@
...
+use crate::ValidWaitParams;
 use anyhow::{anyhow, Result};
 use camino::Utf8PathBuf;
 use serde::{Deserialize, Serialize};
 use std::fs;
 use tempfile::TempDir;
 use toml::Value;
 
-#[derive(Deserialize, Serialize, Clone, Debug, PartialEq)]
+#[derive(Default, Deserialize, Serialize, Clone, Debug, PartialEq)]
```
THenry14, line 10: just to be sure you've checked - a default value for an Option is `None`, right? (I'm thinking about keystore param)

crates/sncast/src/lib.rs
```diff
@@ -75,8 +76,54 @@ impl NumbersFormat {
 
 pub struct WaitForTx {
     pub wait: bool,
-    pub timeout: u16,
-    pub retry_interval: u8,
+    pub wait_params: ValidWaitParams,
+}
+
+#[derive(Deserialize, Serialize, Clone, Debug, Copy, PartialEq)]
+pub struct ValidWaitParams {
```
THenry14, line 83: Let's just rename it to WaitParams, as it does not make sense to pass invalid params :)

THenry14 (approved):
Two comments that are nitpicks, so after addressing them this is mergeable :+1:

crates/sncast/src/helpers/configuration.rs
abulenok: yeah

and `#[derive(Default)]` was suggested by clippy

crates/sncast/src/lib.rs
abulenok: maybe `ValidatedWaitParams`? :)

THenry14: Still a bit redundant, but good enough for me :+1:

abulenok: I just think that ValidatedWaitParams gives a signal that it's not just a struct collecting `retry_interval` and `timeout`, but a struct that also assures that these params are valid

Issue #1231, opened 2023-11-28
Write tx info to state file


THenry14:
Refer to design doc for details

PR #1722, opened 2024-02-16
Add scripts state file with write functionality

THenry14:
Closes https://github.com/foundry-rs/starknet-foundry/issues/1231

## Introduced changes

- Add structs along with necessary functions for writing to script state file

## Checklist

- [X] Linked relevant issue
- [X] Added relevant tests


crates/sncast/src/state.rs
```diff
@@ -0,0 +1,324 @@
...
+#[derive(Clone, Deserialize, Serialize, Debug, PartialEq)]
+pub struct ErrorResponse {
+    pub message: String,
+}
+
+#[derive(Clone, Deserialize, Serialize, Debug, PartialEq)]
+pub enum ScriptTransactionStatus {
+    // script executed successfully, transaction accepted/succeeded
+    Success,
+    // script executed successfully, transaction not rejected/reverted
```
abulenok, line 70:
```suggestion
    // script executed successfully, transaction rejected/reverted
```

crates/sncast/src/state.rs
```diff
@@ -0,0 +1,324 @@
...
+        transactions: transaction,
+    }
+}
+
+pub fn write_txs_to_state_file(
+    state_file_path: &Utf8PathBuf,
+    tx_entries: ScriptTransactionEntries,
+) -> Result<()> {
+    let mut state_file = load_or_create_state_file(state_file_path)
+        .unwrap_or_else(|_| panic!("Failed to write to state file {state_file_path}"));
```
abulenok, line 119:
```suggestion
        .unwrap_or_else(|_| panic!("Failed to load state file {state_file_path}"));
```

crates/sncast/src/state.rs
```diff
@@ -0,0 +1,324 @@
...
+    fs::write(
+        state_file_path,
+        serde_json::to_string_pretty(&state_file)
+            .expect("Failed to convert ScriptTransactionsSchema to json"),
+    )
+    .unwrap_or_else(|_| panic!("Failed to write new transactions to state file {state_file_path}"));
+    Ok(())
+}
+
+// todo: add support for multiple state file versions (when needed)
```
abulenok, line 130: let's create an issue or remove this comment for now (as we will probably not remember about this comment).

crates/sncast/src/state.rs
```diff
@@ -0,0 +1,324 @@
...
+fn verify_version(version: u8) -> Result<()> {
+    match version {
+        STATE_FILE_VERSION => Ok(()),
+        _ => Err(anyhow!(format!("Unsupported state file version {version}"))),
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
abulenok, line 139: let's add a test writing multiple transaction entries at once

crates/sncast/src/state.rs
THenry14: :D 

https://github.com/foundry-rs/starknet-foundry/pull/1722/commits/035373f68cfe58f80fefb8b1ac4c7a8493621d07

crates/sncast/src/state.rs
THenry14: https://github.com/foundry-rs/starknet-foundry/commit/035373f68cfe58f80fefb8b1ac4c7a8493621d07

crates/sncast/src/state.rs
THenry14: not sure we will have such scenario in the code, but sure - added such test in https://github.com/foundry-rs/starknet-foundry/commit/035373f68cfe58f80fefb8b1ac4c7a8493621d07

crates/sncast/src/state.rs
THenry14: Hmm, not sure about this tbh. Maybe I should use `with_context` instead, to include the message from load function? This way we will have a cleaner error message, pointing to the exact problem. wdyt?

abulenok: I like it

crates/sncast/src/state.rs
```diff
@@ -0,0 +1,392 @@
...
+pub struct ErrorResponse {
+    pub message: String,
+}
+
+#[derive(Clone, Deserialize, Serialize, Debug, PartialEq)]
+pub enum ScriptTransactionStatus {
+    // script executed successfully, transaction accepted/succeeded
+    Success,
+    // script executed successfully, transaction rejected/reverted
+    Fail,
```
abulenok, line 71: Do we intend for users to read state file? If so it may be beneficial to add info about whether the transaction was reverted or rejected, since reverted transactions alter the state of the blockchain and cost money

THenry14: > Do we intend for users to read state file? 

Not directly probably, but it will be possible

> If so it may be beneficial to add info about whether the transaction was reverted or rejected, since reverted transactions alter the state of the blockchain and cost money

This information is intended to be held in `ScriptTransactionOutput` -> `ErrorResponse`. There is an example test showing this off `test_load_or_create_state_file_exists_with_txs`, I thought having this there would be a pretty sane choice as this is actually a transaction output. The `ScriptTransactionStatus` on the other hand was meant to be a simple, quick way for us to determine the tx status (without going into its output) - and wheter such tx should be replayed later on (or not).

Does that make sense?

crates/sncast/src/state.rs
THenry14: done https://github.com/foundry-rs/starknet-foundry/pull/1722/commits/66f6ffd83e44492fa7ed0afccd78f9103f30f4fd

crates/sncast/src/state.rs
abulenok: yep, makes sense

PR #1735, opened 2024-02-19
Raise steps limit

### LINKED ISSUE #1652: Raising the limit of steps

Forge

### Feature Request

During one test suite run, we could raise the step limit on demand

If a user overrides the defaults, we should warn if the limit is raised beyond the maximum acceptable in normal conditions (on starknet).

Draggu:
Closes #1652

## Introduced changes

- Added option `--max-n-steps` that allow changing steps limit (3_000_000, now default)

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/runtime/src/starknet/context.rs
```diff
@@ -140,6 +143,13 @@ pub fn build_context(block_info: BlockInfo) -> EntryPointExecutionContext {
     .unwrap()
 }
 
+pub fn set_max_steps(entry_point_ctx: &mut EntryPointExecutionContext, max_n_steps: u32) {
+    entry_point_ctx.block_context.invoke_tx_max_n_steps = max_n_steps;
```
Arcticae, line 147: Why is this one needed? isn't `vm_run_resources` sufficient?

Draggu: I followed the logic of `build_context()` -> `build_block_context()` to get same state that with `BlockContext` direct construction.
But yes, in fact it's not necessary.

Arcticae (approved):

docs/src/appendix/snforge/test.md
```diff
@@ -58,6 +58,10 @@ Display additional info about used resources for passed tests.
 
 ## `--save-trace-data`
 
+Number of maximum steps during single test
```
piotmag769, line 61:
```suggestion
Number of maximum steps during a single test. For fuzz tests this value is applied to each subtest separately.
```

crates/forge/tests/data/steps/src/lib.cairo
```diff
@@ -0,0 +1,60 @@
+#[cfg(test)]
+mod tests {
+    #[test]
+    // require 570031 steps
```
piotmag769, line 4:
```suggestion
    // requires 570031 steps
```

crates/forge/src/main.rs
```diff
@@ -108,6 +108,10 @@ struct TestArgs {
     /// Save execution traces of all test which have passed and are not fuzz tests
     #[arg(long)]
     save_trace_data: bool,
+
+    /// Max steps per test
+    #[arg(long)]
+    max_n_steps: Option<u32>,
```
piotmag769, line 114: Can we have the same description as in docs? Clap parses this comment and it is shown in CLI if you type `snforge --help`

crates/forge-runner/src/lib.rs
```diff
@@ -46,18 +45,16 @@ mod running;
...
-        "RangeCheck",
-        "Bitwise",
-        "EcOp",
-        "Poseidon",
-        "SegmentArena",
-        "GasBuiltin",
-        "System",
-    ]
-});
+pub const BUILTINS: [&str; 8] = [
```
piotmag769, line 48: Good catch xD Was used as vector before prob

crates/runtime/src/starknet/context.rs
piotmag769: I would leave the line though, I think it's better to be consistent xD

piotmag769 (approved):

crates/forge/tests/data/steps/src/lib.cairo
```diff
@@ -0,0 +1,60 @@
...
+    fn steps_570031() {
+        let mut i = 0;
+
+        loop {
+            if i == 30_000 {
+                break;
+            }
+
+            i = i + 1;
```
war-in, line 13: [nit] We could probably perform some more complicated operations to reach the expected steps faster

war-in (approved):

PR #1741, opened 2024-02-21
CallFailure::Panic for unknown entry point

### LINKED ISSUE #1739: Unknown Entry Point Error fails `should_panic` tests

snforge

### Description

https://github.com/foundry-rs/starknet-foundry/blob/dc55ebf4f20de60e7d743d24286174e013328033/crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/rpc.rs#L147

drknzz:
Closes #1739

## Introduced changes

- [Change CallFailure::Error to CallFailure::Panic](https://github.com/foundry-rs/starknet-foundry/commit/4cef7f838992f41804e4f8e79a2787e3f4816a67)

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests

### DIFF
--- a/crates/forge/tests/integration/should_panic.rs
+++ b/crates/forge/tests/integration/should_panic.rs
@@ -51,7 +51,7 @@ fn should_panic_unknown_entry_point() {
             use snforge_std::{declare, ContractClass, ContractClassTrait};
 
             #[test]
-            #[should_panic(expected: "Entry point selector 0x00000000000000000000696e6578697374656e745f656e7472795f706f696e74 not found in contract 0x06288882237f586f11e7a1bcdd1b4841708747bec96952dce019cd76ff3d806f")]
+            #[should_panic]
             fn should_panic_with_no_expected_data() {
                 let contract = declare('HelloStarknet');
                 let contract_address = contract.deploy(@ArrayTrait::new()).unwrap();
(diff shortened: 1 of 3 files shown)


crates/forge/tests/integration/should_panic.rs
```diff
@@ -35,3 +38,43 @@ fn should_panic() {
...
+        indoc!(
+            r"
+            use array::ArrayTrait;
+            use starknet::{call_contract_syscall, ContractAddress, Felt252TryIntoContractAddress};
+            use result::ResultTrait;
+
+            use snforge_std::{declare, ContractClass, ContractClassTrait};
+
+            #[test]
+            #[should_panic]
```
drknzz, line 54: In rust, it's possible to have `should_panic` match expected error by a substring, meanwhile we [always assert the full message](https://github.com/foundry-rs/starknet-foundry/blob/07c3ef85a97bfb1ed8eacaa529f6e98569489f94/crates/forge-runner/src/test_case_summary.rs#L251). Maybe it's something we could add an option for?

drknzz: Here the string error cannot be asserted because it contains the address of a contract

piotmag769 (approved):

crates/forge/tests/integration/should_panic.rs
piotmag769: We can add an issue for that, good idea. I think substring should be the default behavior

PR #1748, opened 2024-02-22
Don't use `stderr_matches` in Cast tests

### LINKED ISSUE #1658: Support running tests with short or full `RUST_BACKTRACE`

Cast

### Feature Request

Currently, some tests use `stderr_matches` to check error output. These checks fail when backtrace is enabled (env variable `RUST_BACKTRACE` is not `0`).

We should opt for using `contains` instead, while creating a proper error message in case the check fails, similar to the one created by `stderr_matches`:
```
--- Expected
++++ actual:   stdout
   1    1 | command: account add
   2      - add_profile: Profile successfully added to snfoundry.toml
        2 + add_profile: Profile my_account_add successfully added to snfoundry.toml    
```

DelevoXDG:
Closes #1658 

## Introduced changes

- Move `output_assert.rs` from forge `test_utils` to shared crate
  - Modify behaviour of `assert_stderr_contains` and `assert_stdout_contains`; Actually validate content inclusion, not strict matching
- Replaced `output.stderr_matches()` calls with `assert_stderr_contains()` in Cast tests

## Checklist

- [x] Linked relevant issue
- [ ] Added relevant tests


crates/sncast/tests/e2e/account/create.rs
```diff
@@ -242,14 +247,18 @@ pub async fn test_happy_case_keystore() {
...
+        output,
+        indoc! {r"
         command: account create
         add_profile: --add-profile flag was not set. No profile added to snfoundry.toml
         address: 0x[..]
         max_fee: [..]
         message: Account successfully created[..]
-    "});
+        "},
+    );
```
DelevoXDG, line 261: I'm not sure whether we want to use `assert_stdout_contains` for consistency (same logic that is applied to `stderr`) in these cases, or keep `stdout_matches`, since we don't expect any stack trace and hereby can strictly match a snippet.

DelevoXDG:
Should we also replace other cases where `.contains()` on `stderr` or `stdout` is used prior to my changes for consistency, or deal with this separately?  

https://github.com/foundry-rs/starknet-foundry/blob/68e8dc2b9c27b1ccc94a4b1d470a18eb68eb755e/crates/sncast/tests/e2e/account/create.rs#L37

https://github.com/foundry-rs/starknet-foundry/blob/68e8dc2b9c27b1ccc94a4b1d470a18eb68eb755e/crates/sncast/tests/e2e/account/create.rs#L368

war-in (approved):
I think we should unify the `contains` case too. Can be done in separate PR merged into this one

crates/shared/src/test_utils/output_assert.rs
```diff
@@ -51,27 +51,29 @@ fn assert_output_contains(output: &str, lines: &str) {
...
-        out.push_str(&remaining_line);
-        out.push('\n');
+
+    if !contains {
+        for remaining_line in actual_lines {
+            out.push_str("+ ");
+            out.push_str(&remaining_line);
+            out.push('\n');
+        }
     }
```
THenry14, line 74: nit: could probably use an iter().foreach() with a closure instead

```suggestion
    if !contains {
      actual_lines.iter().for_each(|line| out.push_str(&format!("+ {}\n", line)));
    }
```

THenry14 (approved):

crates/shared/src/test_utils/output_assert.rs
DelevoXDG: Updated this, but I'm not sure it's more readable this way 😅

PR #1749, opened 2024-02-22
Fix config file naming, add script to appendix

THenry14:
## Introduced changes

- Adds script to appendix
- fixes wrong config names

## Checklist

- [X] Linked relevant issue
- [X] Added relevant tests

### DIFF
--- a/docs/src/appendix/sncast/account/add.md
+++ b/docs/src/appendix/sncast/account/add.md
@@ -4,7 +4,7 @@ Import an account to accounts file.
 Account information will be saved to the file specified by `--accounts-file` argument,
 which is `~/.starknet_accounts/starknet_open_zeppelin_accounts.json` by default.
 
-## Required Common Arguments — Passed By CLI or Specified in `Scarb.toml`
+## Required Common Arguments — Passed By CLI or Specified in `snfoundry.toml`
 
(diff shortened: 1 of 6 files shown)


docs/src/appendix/sncast/account/add.md
```diff
@@ -4,7 +4,7 @@ Import an account to accounts file.
 Account information will be saved to the file specified by `--accounts-file` argument,
 which is `~/.starknet_accounts/starknet_open_zeppelin_accounts.json` by default.
 
-## Required Common Arguments — Passed By CLI or Specified in `Scarb.toml`
+## Required Common Arguments — Passed By CLI or Specified in `snfoundry.toml`
 
 * [`url`](../common.md#--url--u-rpc_url)
```
ddoktorski, line 9: Can we consider including this information also for other commands or maybe even in `common.md`? As far as I know `script init` is the only command that currently doesn't require url.

abulenok (approved):

docs/src/appendix/sncast/account/add.md
THenry14: sure, created an issue and will handle in separate pr
https://github.com/foundry-rs/starknet-foundry/issues/1759

PR #1754, opened 2024-02-22
Add hashing function for scripts idempotency

### LINKED ISSUE #1545: Implement hashing function for scripts idempotency

Cast

### Feature Request

The objective of this task is to write a function (along with necessary unit tests) that will generate a unique id for a function, based on its name, arguments etc. This function will be later used to generate ids for scripts state file.

When starting this issue we should discuss the approach. General idea can be found in design doc https://github.com/foundry-rs/starknet-foundry/blob/master/design_documents/cairo_deployment_scripts.md#idempotency

abulenok:
Closes #1545

## Introduced changes

- Implemented function for hashing calls to script subcommand (transforms input to bytes in hex format)

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests

### DIFF
--- a/crates/sncast/src/state/hashing.rs
+++ b/crates/sncast/src/state/hashing.rs
@@ -2,7 +2,6 @@ use cairo_felt::Felt252;
 use itertools::Itertools;
 
 // if we change API this might have collisions with old API hashes
-#[allow(dead_code)]
 pub fn generate_id(selector: &str, inputs: &[Felt252]) -> String {
     let mut res = selector.as_bytes().to_owned();
     let mut inputs_bytes: Vec<u8> = inputs.iter().flat_map(Felt252::to_bytes_be).collect_vec();
@@ -12,7 +11,7 @@ pub fn generate_id(selector: &str, inputs: &[Felt252]) -> String {
 
 #[cfg(test)]
 mod tests {
-    use crate::state::hashing::hash_script_subcommand_call;
+    use crate::state::hashing::generate_id;
     use cairo_felt::Felt252;
     use num_traits::Num;
 
@@ -22,7 +21,7 @@ mod tests {
 
     #[test]
     fn basic_case() {
-        let hash = hash_script_subcommand_call("aaa", &[Felt252::from(b'a')]);
+        let hash = generate_id("aaa", &[Felt252::from(b'a')]);
         assert_eq!(hash, "61616161");
     }
 
@@ -33,7 +32,7 @@ mod tests {
             Felt252::from(1),
             Felt252::from(1),
         ];
-        let hash = hash_script_subcommand_call("declare", inputs.as_ref());
+        let hash = generate_id("declare", inputs.as_ref());
         assert_eq!(hash, DECLARE_HEX.to_owned() + "4d6170613211");
     }
 
@@ -51,7 +50,7 @@ mod tests {
             Felt252::from(1),
             Felt252::from(1),
         ];
-        let hash = hash_script_subcommand_call("deploy", inputs.as_ref());
+        let hash = generate_id("deploy", inputs.as_ref());
         assert_eq!(
             hash,
             DEPLOY_HEX.to_owned()
@@ -74,7 +73,7 @@ mod tests {
             Felt252::from(1),
             Felt252::from(1),
         ];
-        let hash = hash_script_subcommand_call("invoke", inputs.as_ref());
+        let hash = generate_id("invoke", inputs.as_ref());
         assert_eq!(
             hash,
             INVOKE_HEX.to_owned()
--- a/crates/sncast/src/state/hashing.rs
+++ b/crates/sncast/src/state/hashing.rs
@@ -3,7 +3,7 @@ use itertools::Itertools;
 
 // if we change API this might have collisions with old API hashes
 #[allow(dead_code)]
-pub fn hash_script_subcommand_call(selector: &str, inputs: &[Felt252]) -> String {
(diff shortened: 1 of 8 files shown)


crates/sncast/src/state/hashing.rs
```diff
@@ -0,0 +1,84 @@
+use cairo_felt::Felt252;
+use itertools::Itertools;
+
+// if we change API this might have collisions with old API hashes
+#[allow(dead_code)]
+pub fn hash_script_subcommand_call(selector: &str, inputs: &[Felt252]) -> String {
+    let mut res = selector.as_bytes().to_owned();
+    let mut inputs_bytes: Vec<u8> = inputs.iter().flat_map(Felt252::to_bytes_be).collect_vec();
+    res.append(&mut inputs_bytes);
+    res.iter().map(|item| format!("{item:x?}")).join("").clone()
```
THenry14, line 10: Technically speaking - is there a max length to this? what if someone tries to pass a very long calldata, wouldn't we end up with a comically long ids?

crates/sncast/src/state/hashing.rs
```diff
@@ -0,0 +1,84 @@
+use cairo_felt::Felt252;
+use itertools::Itertools;
+
+// if we change API this might have collisions with old API hashes
+#[allow(dead_code)]
+pub fn hash_script_subcommand_call(selector: &str, inputs: &[Felt252]) -> String {
```
THenry14, line 6:
```suggestion
pub fn generate_id(selector: &str, inputs: &[Felt252]) -> String {
```
?

crates/sncast/src/state/state_file.rs
```diff
@@ -1,3 +1,4 @@
+#![allow(dead_code)]
```
THenry14, line 1: did clippy complain about this?

THenry14, line 1: We should use this new function in generate_transaction_entry_with_id probably

THenry14 (changes requested):

crates/sncast/src/state/state_file.rs
abulenok: yes

crates/sncast/src/state/hashing.rs
abulenok: Nope, no max length for this one. The longer the input, the longer the output

abulenok: but, will this be a problem?

THenry14: maybe not directly, but this may inflate the file size, and make the file less human-readable.

crates/sncast/src/state/state_file.rs
abulenok: https://github.com/foundry-rs/starknet-foundry/pull/1754/commits/4e0d449c0cf21cf524b71ba710d604e2108f2e6b

crates/sncast/src/state/hashing.rs
abulenok: I'll take a look into hashing functions with a fixed-size output.

crates/sncast/src/state/hashing.rs
```diff
@@ -0,0 +1,77 @@
...
+
+// if we change API this might have collisions with old API hashes
+pub fn generate_id(selector: &str, inputs: &[Felt252]) -> String {
+    let mut res = selector.as_bytes().to_owned();
+    let mut inputs_bytes: Vec<u8> = inputs.iter().flat_map(Felt252::to_bytes_be).collect_vec();
+    res.append(&mut inputs_bytes);
+    let res = res.iter().map(|item| format!("{item:x?}")).join("").clone();
+
+    let hash = Sha3_256::new().chain_update(res).finalize();
+    Base64::encode_string(&hash)
```
THenry14, line 15: is base64 necessary, after we already did sha? I'm a bit worried about special characters in ids (but maybe I shouldn't be?)

crates/sncast/src/state/state_file.rs
```diff
@@ -102,9 +106,10 @@ pub fn load_or_create_state_file(path: &Utf8PathBuf) -> Result<ScriptTransaction
 #[must_use]
 pub fn generate_transaction_entry_with_id(
     tx_entry: ScriptTransactionEntry,
+    inputs: &[Felt252],
 ) -> ScriptTransactionEntries {
     // todo (1545): Implement hashing function for unique ids
```
THenry14, line 111:
```suggestion
```

crates/sncast/src/state/hashing.rs
abulenok: base64 is needed to serialise hash value into string

https://github.com/RustCrypto/hashes#:~:text=If%20you%20need%20to%20serialize%20hash%20value%20into%20string%2C%20you%20can%20use%20crates%20like%20base16ct%20and%20base64ct%3A

abulenok: we can use base16 to serialise the hash value into a hex string, but it will be longer

abulenok: > I'm a bit worried about special characters in ids (but maybe I shouldn't be?)

hm, what are the possible drawbacks?

abulenok: > the string will be longer
but it's not a big difference, so we can do a safer option

abulenok: updated

THenry14 (approved):

PR #1756, opened 2024-02-22
improvements (mostly allocation related)

Draggu:
## Introduced changes

- Some improvements like removing unnecessary `.clone()` or creating `Vec` when not needed, using `String` in place where `&str` is completely fine etc.

### DIFF
--- a/crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
+++ b/crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
@@ -83,12 +83,10 @@ impl<'a> ExtensionLogic for ForgeExtension<'a> {
     fn handle_cheatcode(
         &mut self,
         selector: &str,
-        inputs: Vec<Felt252>,
+        mut reader: BufferReader<'_>,
         extended_runtime: &mut Self::Runtime,
     ) -> Result<CheatcodeHandlingResult, EnhancedHintError> {
-        let mut reader = BufferReader::new(&inputs);
-
-        let res = match selector {
+        match selector {
             "start_roll" => {
                 let target = reader.read_cheat_target();
                 let block_number = reader.read_felt();
@@ -607,9 +605,7 @@ impl<'a> ExtensionLogic for ForgeExtension<'a> {
                 Ok(CheatcodeHandlingResult::Handled(vec![map_entry_address]))
             }
             _ => Ok(CheatcodeHandlingResult::Forwarded),
-        }?;
-
-        Ok(res)
+        }
     }
 
     fn override_system_call(
@@ -620,7 +616,7 @@ impl<'a> ExtensionLogic for ForgeExtension<'a> {
     ) -> Result<SyscallHandlingResult, HintError> {
         match selector {
             DeprecatedSyscallSelector::ReplaceClass => Err(HintError::CustomHint(Box::from(
-                "Replace class can't be used in tests".to_string(),
+                "Replace class can't be used in tests",
             ))),
             _ => Ok(SyscallHandlingResult::Forwarded),
         }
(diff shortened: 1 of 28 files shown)


crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
```diff
@@ -645,18 +641,24 @@ fn handle_deploy_result(
...
-    output.append(&mut serialize_call_entry_point(&call_trace.entry_point));
+
+    serialize_call_trace_inner(call_trace, &mut output);
+
+    output
+}
+
+// append all to one output Vec instead of allocating new one for each nested call
+fn serialize_call_trace_inner(call_trace: &CallTrace, output: &mut Vec<Felt252>) {
+    serialize_call_entry_point(&call_trace.entry_point, output);
```
war-in, line 652: Do we really need this `inner` function? We could make the main one take output and write into it

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
```diff
@@ -676,21 +678,22 @@ fn serialize_call_entry_point(call_entry_point: &CallEntryPoint) -> Vec<Felt252>
...
     output.push(call_entry_point.caller_address.into_());
     output.push(Felt252::from(call_type));
-
-    output
 }
 
 #[must_use]
 pub fn cheatcode_panic_result(panic_data: Vec<Felt252>) -> Vec<Felt252> {
-    let mut result = vec![Felt252::from(1), Felt252::from(panic_data.len())];
+    let mut result = Vec::with_capacity(panic_data.len() + 2);
```
war-in, line 692: Do we care for such optimisations? I so, are all possible `vec!` replaced?

crates/cheatnet/src/state.rs
```diff
@@ -164,10 +164,8 @@ impl NotEmptyCallStack {
     }
 
     pub fn top(&mut self) -> Rc<RefCell<CallTrace>> {
-        let top_val = self.0.pop().unwrap();
-        let borrowed_ref = top_val.call_trace.clone();
-        self.0.push(top_val);
-        borrowed_ref
+        let top_val = self.0.last().unwrap();
+        top_val.call_trace.clone()
```
war-in, line 168: Not sure if this is correct. I thought we wanted to have a reference to the top call

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
Draggu: In fact we don't, good point.

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
Draggu: We may not in this case (i don't know), but doing it this way won't hurt us (both speed and readability). If I found a place where `Vec::with_capacity()` can be  easily used, then used it. Of course I could miss something.

crates/scarb-api/src/lib.rs
```diff
@@ -204,15 +204,20 @@ pub fn package_matches_version_requirement(
...
-        [] => Ok(true),
-        [package] => Ok(version_req.matches(&package.version)),
-        _ => Err(anyhow!("Package {name} is duplicated in dependencies")),
+        .filter(|package| package.name == name);
+
+    let package = packages.next();
+
+    if packages.next().is_some() {
+        return Err(anyhow!("Package {name} is duplicated in dependencies"));
+    }
```
piotmag769, line 216: This can get error prone pretty fast, I don't think collecting to vector hurts here

crates/runtime/src/lib.rs
```diff
@@ -380,7 +378,7 @@ pub trait ExtensionLogic {
     fn handle_cheatcode(
         &mut self,
         _selector: &str,
-        _inputs: Vec<Felt252>,
+        _reader: BufferReader,
```
piotmag769, line 381: `_input_reader` imo

crates/cheatnet/tests/common/cache.rs
```diff
@@ -5,10 +5,14 @@ use std::path::PathBuf;
...
 pub fn read_cache(file_pattern: &str) -> Map<String, Value> {
-    let cache_files: Vec<PathBuf> = glob(file_pattern).unwrap().filter_map(Result::ok).collect();
-    assert!(cache_files.len() < 2, "Multiple matching cache files found");
+    let mut cache_files = glob(file_pattern).unwrap().filter_map(Result::ok);
+    let cache_file = cache_files.next().expect("Cache file not found");
+
+    assert!(
+        cache_files.next().is_none(),
+        "Multiple matching cache files found"
+    );
```
piotmag769, line 14: Again, not convinced to this pattern, error prone imo

crates/runtime/src/lib.rs
Draggu: Good point. FIxed.

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
```diff
@@ -83,15 +83,13 @@ impl<'a> ExtensionLogic for ForgeExtension<'a> {
     fn handle_cheatcode(
         &mut self,
         selector: &str,
-        inputs: Vec<Felt252>,
+        mut inputreader: BufferReader<'_>,
```
piotmag769, line 86:
```suggestion
        mut input_reader: BufferReader<'_>,
```

piotmag769 (approved):

PR #1708, opened 2024-02-14
Error handling for deployment scripts `invoke` subcommand

abulenok:
Partially solves #1361

## Introduced changes

- `invoke` from `sncast_std` now returns `Result<InvokeResult, ScriptCommandError>`
- Added tests for `invoke` error handling in deployment scripts

## Checklist

- [x] Linked relevant issue
- [ ] Updated relevant documentation
- [ ] Added relevant tests
- [ ] Performed self-review of the code
- [ ] Added changes to `CHANGELOG.md`

PR #1763, opened 2024-02-23
Adjust assertion logic in Cast tests

DelevoXDG:
## Introduced changes

Refactor test cases to unify logic of output assertions, following changes from #1748  

- Replace asserts that used `stdout_str.contains` with `assert_stdout_contains()` or `.stdout_matches()`
- Replace asserts that used `stderr_str.contains` with `assert_stderr_contains()`
- Use `.as_stderr()` and `.as_stdout()` instead of manually converting to strings

## Checklist

- [ ] Linked relevant issue
- [ ] Added relevant tests


crates/sncast/tests/e2e/account/create.rs
```diff
@@ -30,11 +30,9 @@ pub async fn test_happy_case() {
     ];
 
     let snapbox = runner(&args).current_dir(temp_dir.path());
-    let bdg = snapbox.assert();
-    let out = bdg.get_output();
+    let output = snapbox.assert();
 
-    let stdout_str =
-        std::str::from_utf8(&out.stdout).expect("failed to convert command output to string");
+    let stdout_str = output.as_stdout();
```
abulenok, line 35: why not use `stdout_matches` or `assert_stdout_contains`

crates/sncast/tests/e2e/account/create.rs
```diff
@@ -73,14 +71,14 @@ pub async fn test_happy_case_generate_salt() {
...
-    assert!(stdout_str.contains("command: account create"));
-    assert!(stdout_str.contains("max_fee: "));
-    assert!(stdout_str.contains("address: "));
+    snapbox.assert().success().stdout_matches(indoc! {r"
+        command: account create
+        add_profile: --add-profile flag was not set. No profile added to snfoundry.toml
+        address: 0x[..]
+        max_fee: [..]
+        message: Account successfully created[..]
+        "});
```
abulenok, line 81: What is the difference between using `assert_stdout_contains` and `.stdout_matches`?

crates/sncast/tests/e2e/account/create.rs
DelevoXDG: Because we also use `!stdout_str.contains` there, hence we need to convert to string anyway. Thought it would be reasonable to not use multiple approaches.

crates/sncast/tests/e2e/account/create.rs
DelevoXDG: `assert_stdout_contains` verifies if the segment is within `stdout`, while `.stdout_matches` ensures that `stdout` contains only the specified segment and nothing else.

abulenok: but in `. stdout_matches` we can use `...`, which skips multiple lines. That allows us to check if the output contains only a segment

DelevoXDG: The following above will fail of only the segment is present:
```rs
snapbox.assert().success().stdout_matches(indoc! {r"
     command: account create
     address: 0x[..]
     max_fee: [..]
     message: Account successfully created[..]
     ...
     "});
```

abulenok: yeah, my point is that we can do both. We can assert the full output, we can assert a segment with `...`.
but we can say that we assert the full output, but skip some lines with `...`. Although these lines need to exist

abulenok (approved):

crates/sncast/tests/e2e/account/create.rs
abulenok: will also be useful here https://github.com/foundry-rs/starknet-foundry/pull/1708#discussion_r1507356235

DelevoXDG: Sure, will try to have this and #1748 merged to master ASAP

Issue #1542, opened 2024-01-16
Improve ambigous error messages


THenry14:
Up until now we returned error messages as they were from node/sdk. We should improve on that, and make sure some meaningful errors (or at least messages) are being returned. Example of error that should be improved:

```
command: account create
error: TransactionExecutionError
```

this is when there is no class hash of account contract to be used

PR #1774, opened 2024-02-27
Fix error message when creating account without correct class_hash

abulenok:
As a part of #1542

## Introduced changes

- Return proper error message when undeclared account class hash is used
- Return proper error message when fee estimation fails in account create

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests

### DIFF
--- a/crates/sncast/src/starknet_commands/account/create.rs
+++ b/crates/sncast/src/starknet_commands/account/create.rs
@@ -1,19 +1,22 @@
 use crate::starknet_commands::account::{
     add_created_profile_to_configuration, prepare_account_json, write_account_to_accounts_file,
 };
-use anyhow::{bail, Context, Result};
+use anyhow::{anyhow, bail, Context, Result};
 use camino::Utf8PathBuf;
 use clap::Args;
 use serde_json::json;
 use sncast::helpers::configuration::CastConfig;
 use sncast::helpers::constants::{CREATE_KEYSTORE_PASSWORD_ENV_VAR, OZ_CLASS_HASH};
 use sncast::response::structs::{AccountCreateResponse, Felt};
-use sncast::{extract_or_generate_salt, get_chain_id, get_keystore_password, parse_number};
+use sncast::{
+    extract_or_generate_salt, get_chain_id, get_keystore_password, handle_account_factory_error,
+    handle_rpc_error, parse_number,
+};
 use starknet::accounts::{AccountFactory, OpenZeppelinAccountFactory};
-use starknet::core::types::{FeeEstimate, FieldElement};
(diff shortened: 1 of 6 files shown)


crates/sncast/src/starknet_commands/account/create.rs
```diff
@@ -52,6 +55,7 @@ pub async fn create(
     let class_hash = class_hash.unwrap_or_else(|| {
         FieldElement::from_hex_be(OZ_CLASS_HASH).expect("Failed to parse OZ class hash")
     });
+    check_class_hash_exists(provider, class_hash).await?;
```
ddoktorski, line 58: Can we make the same check in the `account add` command?

crates/sncast/tests/e2e/account/create.rs
```diff
@@ -53,6 +53,32 @@ pub async fn test_happy_case() {
...
+        "my_account_create_happy",
+        "--salt",
+        "0x1",
+    ];
+
+    let snapbox = runner(&args).current_dir(temp_dir.path());
+
+    snapbox.assert().stderr_matches(indoc! {r"
+        command: account create
+        error: The class 0x58d97f7d76e78f44905cc30cb65b91ea49a4b908a76703c54197bca90f81773 is undeclared, try using --class-hash with a class hash that is already declared
```
ddoktorski, line 78: nit: Maybe instead of using the default value here, we can pass a class hash that is more clearly incorrect, something like `0x123`?

crates/sncast/src/starknet_commands/account/create.rs
abulenok: As I see, we don't perform any validations when doing `account add`
Introducing class hash validation we should also validate account address if the `--deployed` flag is passed
WDYT?

crates/sncast/tests/e2e/account/create.rs
abulenok: https://github.com/foundry-rs/starknet-foundry/pull/1774/commits/2d5d3b6e1662824c5632dc72a48e7eb6fdf179bd

crates/sncast/src/starknet_commands/account/create.rs
ddoktorski: Agree, it makes sense

abulenok: Added in https://github.com/foundry-rs/starknet-foundry/pull/1774/commits/ea8356603bcee52c75eb5dd7c77bcf2ced8a93be

And created an issue for account address verification https://github.com/foundry-rs/starknet-foundry/issues/1780

ddoktorski (approved):
LGTM

Issue #1751, opened 2024-02-22
Add L1 storage to trace data


war-in:
### Which component is your feature related to?

Forge

### Feature Request

We should add storage diff to trace.

MaksymilianDemitraszek:
Specific case we should handle correctly is the case when the storage update is negative (entry point updates state that was previously changed to default value). We probably need to track each storage cell individually and blame the entry point that set the storage last.

Issue #1725, opened 2024-02-16
MVP function level profiling


war-in:
### Which component is your feature related to?

Forge

### Feature Request

@MaksymilianDemitraszek please elaborate on that 🙏

piotmag769:
Blocked by #846

MaksymilianDemitraszek:
We want the graph to be a call graph of all functions not only entry-points.

Relevant mappins are already built in the cairo compiler. We need to extend scarb to be able to produce artifacts from those which profiler will be able to consume as optional.

PR #1823, opened 2024-03-05
Storage diff in trace

war-in:
Closes (on forge side) #1751 

## Introduced changes

-

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/cheatnet/src/state.rs
```diff
@@ -446,6 +465,25 @@ impl TraceData {
     }
 }
 
+fn get_storage_values_updated(after: &CommitmentStateDiff, before: &CommitmentStateDiff) -> isize {
```
Arcticae, line 468: This will be misleading and could be inaccurate when profiling for few reasons:
1. The diff is calculated in scope of the whole transaction, so that you don't pay for series of ops like `a: 1 -> a: 2`  and then later `a: 2 -> a: 1` (`a` being the storage address). This profiling metric makes sense only for a top-level call in the test IMO.
2. This doesn't seem to take into account the forking scenario, where we calculate the diff based on the chain state (above scenario applies)

Arcticae (changes requested):

MaksymilianDemitraszek:
Closing this one for now as it only accumulates conflicts. we will work on it after #1725

PR #1840, opened 2024-03-07
Add vm trace

piotmag769:
Closes #

## Introduced changes

-

## Checklist

- [ ] Linked relevant issue
- [ ] Added relevant tests


crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/mod.rs
```diff
@@ -177,7 +177,6 @@ fn execute_syscall<Request: ExecuteCall + SyscallRequest>(
     Ok(())
 }
 
-#[derive(Debug)]
```
piotmag769, line 180: All `Clone` and `Debug` derives are due to `TraceEntry` not implementing any of these. Maybe I should use the `ProfilerTraceEntry` from here instead of the vm type? It leaks the abstraction, but saves type conversion on potentially thousands of struct entities during `ProfilerCallTrace` building

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/deprecated/cairo0_execution.rs
```diff
@@ -66,24 +67,25 @@ pub fn execute_entry_point_call_cairo0(
...
+        vm,
+        runner,
+        cheatable_syscall_handler.extended_runtime.hint_handler,
+        call,
+        previous_vm_resources,
+        implicit_args,
+        n_total_args,
+    )?;
+
+    Ok((execution_result, syscall_counter, vec![]))
```
piotmag769, line 87: Just to keep type the same with `execute_entry_point_call_cairo1`

crates/forge-runner/src/build_trace_data.rs
```diff
@@ -49,6 +50,11 @@ pub fn build_profiler_call_trace(
             .iter()
             .map(|c| build_profiler_call_trace(c, contracts_data))
             .collect(),
+        vm_trace: value
+            .vm_trace
+            .iter()
+            .map(build_profiler_trace_entry)
+            .collect(),
```
piotmag769, line 57: This can be quite expensive (millions of elements potentially)

crates/cheatnet/src/runtime_extensions/common.rs
```diff
@@ -23,3 +25,16 @@ pub fn sum_syscall_counters(mut a: SyscallCounter, b: &SyscallCounter) -> Syscal
...
+pub fn get_relocated_vm_trace(vm: &VirtualMachine) -> Vec<TraceEntry> {
+    vm.get_relocated_trace()
+        .unwrap()
+        .iter()
+        .map(|x| TraceEntry {
+            pc: x.pc,
+            ap: x.ap,
+            fp: x.fp,
+        })
+        .collect()
```
piotmag769, line 39: This can be quite expensive too (millions of elements potentially), maybe we should collect it conditionally? Sadly `get_relocated_trace` returns a reference to a vector :/

crates/forge-runner/src/build_trace_data.rs
```diff
@@ -192,6 +198,14 @@ fn build_profiler_call_type(value: CallType) -> ProfilerCallType {
     }
 }
 
+fn build_profiler_trace_entry(value: &TraceEntry) -> ProfilerTraceEntry {
```
Arcticae, line 201: unidiomatic construction (use `From` instead)

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/entry_point.rs
```diff
@@ -172,18 +174,23 @@ pub fn execute_call_entry_point(
...
-        Ok((call_info, syscall_counter)) => remove_syscall_resources_and_exit_success_call(
-            call_info,
-            syscall_counter,
-            context,
-            resources,
-            runtime_state,
-        ),
-        Err(err) => exit_error_call(err, runtime_state, resources, entry_point),
+    match result {
+        Ok((call_info, syscall_counter, vm_trace)) => {
```
Arcticae, line 178: If the trace is heavy, won't this affect our memory footprint/perf significantly? 
Have you checked this?

Arcticae: I think this is worth benchmarking

crates/forge-runner/src/build_trace_data.rs
MaksymilianDemitraszek: I think both types here do not belong to this package so unfortunately you cannot. It is consistent with other methods in this module though.

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/entry_point.rs
piotmag769: This line or this PR in general? The PR prob will. I can collect the traces conditionally (only when `--save-trace-data` or `--build-profile` flag is used), WDYT?

Arcticae: That's reasonable for me. Although i would check the performance anyway.

piotmag769: They changed nothing for our current benchmarks, it prob will for larger tests, but my approach will be the same anyways

cc @cptartur - first benchmark usage 🎉

crates/cheatnet/src/state.rs
```diff
@@ -423,7 +433,12 @@ impl TraceData {
         used_syscalls: SyscallCounter,
         result: CallResult,
         l2_to_l1_messages: &[OrderedL2ToL1Message],
+        vm_trace: Option<Vec<TraceEntry>>,
     ) {
+        assert!(
+            (vm_trace.is_some() && self.collect_vm_trace)
+                || (vm_trace.is_none() && !self.collect_vm_trace)
+        );
```
piotmag769, line 441: To make optional collection I would have to either use `Option` (which I did) or use dynamic dispatching (overkill). This is the best type guarantee I can make for the way I did it

crates/cheatnet/src/state.rs
```diff
@@ -387,6 +388,14 @@ impl CheatnetState {
...
 
 impl TraceData {
+    #[must_use]
+    pub fn default_vm_trace(&self) -> Option<Vec<TraceEntry>> {
+        if self.collect_vm_trace {
+            Some(vec![])
+        } else {
+            None
+        }
+    }
```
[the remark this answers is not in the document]
MaksymilianDemitraszek: We probably need to change the signature on the side of the profiler though.

MaksymilianDemitraszek (approved):
Please address the comments first, expecially the one with default value

crates/forge-runner/src/running.rs
```diff
@@ -219,6 +218,8 @@ pub fn run_test_case(
         block_info,
         ..Default::default()
     };
+    cheatnet_state.trace_data.collect_vm_trace =
+        runner_config.execution_data_to_save.is_vm_trace_needed();
```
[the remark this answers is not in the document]
piotmag769: I wanted to avoid using verbs in function name (since it suggests it does sth), I can change it for consistency

crates/forge-runner/src/build_trace_data.rs
```diff
@@ -32,14 +32,20 @@ pub const TRACE_DIR: &str = ".snfoundry_trace";
 pub const TEST_CODE_CONTRACT_NAME: &str = "SNFORGE_TEST_CODE";
 pub const TEST_CODE_FUNCTION_NAME: &str = "SNFORGE_TEST_CODE_FUNCTION";
 
-#[must_use]
 pub fn build_profiler_call_trace(
     value: &Rc<RefCell<CallTrace>>,
     contracts_data: &ContractsData,
-) -> ProfilerCallTrace {
+) -> Option<ProfilerCallTrace> {
```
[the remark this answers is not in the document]
piotmag769: We only need the `ProfilerCallTrace` when want to save the call trace data - that's how we determine if we want to collect the vm trace (since collecting the vm trace can impact the memory usage). We decide based on the `ExecutionDataToSave` if we need the vm trace or not. If we don't then all the `CallTrace` structs will contain `vm_trace: None`. Since profiler struct requires vm_trace we cannot build the profiler struct in that case (we don't need it anyways) - that's why I made the trace field in test result in `TestCaseSummary::<Single>::Passed` optional - so this function will return `None`.

I thought about it for a bit and imo it is the best option without introducing type families, which would change 1241241 signatures (really, tried it and gave up).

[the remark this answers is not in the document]
piotmag769: The other idea was to pass `ExecutionData` to this function and assert that the `vm_trace` is indeed `Some` or `None` as implied by `ExecutionData`, but it adds even more params to functions and I didn't see a clear benefit of that other asserting that we for sure don't collect the trace when not needed - if we don't collect the trace when required the code will panic in `maybe_save_execution_data`

crates/cheatnet/src/state.rs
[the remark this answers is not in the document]
piotmag769: >Why we need a default 🤔 It seems like a possible source of bugs. Wouldn't be better to make it Optional?

I don't understand wdym

THenry14 (commented):
looks good from sncast team's point of view.
I don't give an explicit approval though, since I don't quite understand all the blockifier-rlated changes (I'm afraid I do not have enough context, sorry)

crates/cheatnet/src/state.rs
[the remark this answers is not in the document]
piotmag769: Done

crates/forge-runner/src/build_trace_data.rs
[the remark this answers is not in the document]
piotmag769: Changed, no longer an option

Arcticae (approved):

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/cairo1_execution.rs
```diff
@@ -31,7 +34,7 @@ pub fn execute_entry_point_call_cairo1(
     runtime_state: &mut RuntimeState, // Added parameter
     resources: &mut ExecutionResources,
     context: &mut EntryPointExecutionContext,
-) -> EntryPointExecutionResult<(CallInfo, SyscallCounter)> {
+) -> EntryPointExecutionResult<(CallInfo, SyscallCounter, Option<Vec<TraceEntry>>)> {
```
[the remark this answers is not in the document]
piotmag769: If it gets to 4 elements I will consider it xD

PR #1842, opened 2024-03-07
Hoist all deps to workspace

### LINKED ISSUE #1527: Hoist dependencies, extract dev-dependencies
Every non-local dep in package specific Cargo.tomls should be inherited from the workspace Cargo.toml. We should also extract dev-depencencies like `indoc`

MaksymilianDemitraszek:
Closes #1527 

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


Cargo.toml
```diff
@@ -85,3 +85,28 @@ num-integer = "0.1.46"
 scarb-ui = "0.1.3"
 semver = "1.0.22"
 bimap = "0.6.3"
+primitive-types = "0.12.1"
+shellexpand = "3.1.0"
+toml = "0.8.10"
+rpassword = "7.3.1"
+promptly = "0.3.1"
+# Freeze reqwest dependency, because newer versions cause cast tests to fail on macs
+reqwest = "0.11.18"
```
piotmag769, line 94: Missing `=`

crates/sncast/Cargo.toml
```diff
@@ -42,19 +41,19 @@ cairo-vm.workspace = true
...
+base16ct.workspace = true
 
 [dev-dependencies]
 ctor.workspace = true
 snapbox.workspace = true
 indoc.workspace = true
 project-root.workspace = true
-reqwest = "0.11.18"
-tempfile = "3.10.1"
+reqwest.workspace = true
```
piotmag769, line 52: It is both in dev-deps and in dependencies, is it on purpose? @THenry14

piotmag769 (approved):
Looks good, we could also omit patch versions where possible just like Scarb does (e.g. 1.9 instead of 1.9.1)

MaksymilianDemitraszek:
@piotmag769 are there any actual benefits of doing so? I think result is the same, what we have now is only slightly more cluttered, no?

crates/sncast/Cargo.toml
MaksymilianDemitraszek: I don't think this is necessary, removed from dev deps

piotmag769:
@MaksymilianDemitraszek true, up 2 u

crates/forge/Cargo.toml
```diff
@@ -45,15 +43,15 @@ rand.workspace = true
...
 cairo-vm.workspace = true
-ark-ff = "0.4.0-alpha.7"
-ark-std = "0.4.0"
-ark-secp256k1 = "0.4.0"
-ark-secp256r1 = "0.4.0"
+ark-ff.workspace = true
+ark-std.workspace = true
+ark-secp256k1.workspace = true
+ark-secp256r1.workspace = true
 # openssl is being used, please do not remove it!
```
war-in, line 52: This comment can be moved to the top Cargo.toml

MaksymilianDemitraszek: I don't think so (I did so at first) I think it should be here because it is used invisibly by **this** package. In workspace we should only keep comments regarding the versions etc.

PR #1843, opened 2024-03-07
Parsing arguments from memory now fail single test only

### LINKED ISSUE #1469: Parsing arguments from memory panics
We should instead fail the test and suggest upgrading the tool/using higher specific version

This was the case especially for the new Cheatcode interface which can be used as a case study

Draggu:
Closes #1469

## Introduced changes

- Parsing arguments from memory now fail single test with message, instead of stoping execution

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests

### DIFF
--- a/crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
+++ b/crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
@@ -32,7 +32,7 @@ use crate::runtime_extensions::forge_runtime_extension::cheatcodes::storage::{
 use crate::runtime_extensions::forge_runtime_extension::file_operations::string_into_felt;
 use cairo_lang_starknet::contract::starknet_keccak;
 use conversions::byte_array::ByteArray;
-use runtime::utils::BufferReader;
+use runtime::utils::{BufferReadResult, BufferReader};
 use runtime::{
     CheatcodeHandlingResult, EnhancedHintError, ExtendedRuntime, ExtensionLogic,
     SyscallHandlingResult,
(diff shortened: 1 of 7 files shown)


crates/cheatnet/tests/cheatcodes/spoof.rs
```diff
@@ -113,22 +111,24 @@ impl TxInfo {
...
-        let resource_bounds_len = reader.read_felt();
-        let resource_bounds = reader.read_vec_body(
-            3 * resource_bounds_len.to_usize().unwrap(), // ResourceBounds struct has 3 fields
-        );
-        let tip = reader.read_felt();
-        let paymaster_data = reader.read_vec();
-        let nonce_data_availability_mode = reader.read_felt();
-        let fee_data_availability_mode = reader.read_felt();
-        let account_deployment_data = reader.read_vec();
+        let version = reader.read_felt().unwrap();
```
Arcticae, line 114: Shouldn't this also return the `Result`? Isn't it possible that the failure will crash the runner here?

Arcticae (approved):
LGTM, one nitpick

crates/cheatnet/tests/cheatcodes/spoof.rs
piotmag769: Why do we unwrap here? Would be nice to leave a comment before resolving

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
```diff
@@ -54,25 +54,26 @@ pub struct ForgeExtension<'a> {
...
             Some(2) => {
                 let contract_addresses: Vec<_> = self
-                    .read_vec()
+                    .read_vec()?
                     .iter()
                     .map(|el| ContractAddress::from_(el.clone()))
                     .collect();
                 CheatTarget::Multiple(contract_addresses)
             }
             _ => unreachable!("Invalid CheatTarget variant"),
```
piotmag769, line 75: I know I approved it but I noticed that this can also panic (e.g. if we add additional enum variants and someone doesn't swtich the backend etc.). Can we fix this?

crates/cheatnet/tests/cheatcodes/spoof.rs
Draggu: Because it's in our tests only

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
Draggu: https://github.com/foundry-rs/starknet-foundry/pull/1880

PR #1844, opened 2024-03-08
Skip successful transactions based on state file

### LINKED ISSUE #1233: Skip txs based on state file
after #1231 and #1232. refer to design doc for more context.

abulenok:
## Introduced changes

- Introduced `StateManager` which handles cached state-related actions
- Skip successful transactions based on state file

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/sncast/tests/data/scripts/state_file/src/rerun_failed_tx.cairo
```diff
@@ -0,0 +1,17 @@
+use sncast_std::{
+    declare, deploy, invoke, call, DeclareResult, DeployResult, InvokeResult, CallResult, get_nonce
+};
+
+fn main() {
+    let max_fee = 99999999999999999;
+    let salt = 0x3;
```
ddoktorski, line 7: nit: Is it necessary?

crates/sncast/tests/e2e/script/general.rs
```diff
@@ -318,3 +322,177 @@ async fn test_missing_field() {
...
+        status: success
+    "});
+
+    let state_file_path = Utf8PathBuf::from_path_buf(
+        script_dir
+            .path()
+            .join(get_default_state_file_name(script_name, "alpha-goerli")),
+    )
+    .unwrap();
+    let tx_entries_after_first_run = read_txs_from_state_file(&state_file_path).unwrap().unwrap();
```
ddoktorski, line 491: Let's read the state file before executing the script and use `assert_tx_entry_failed` to explicitly show that the state file has been overwritten.

crates/sncast/tests/e2e/script/general.rs
```diff
@@ -318,3 +322,177 @@ async fn test_missing_field() {
         ...
     "});
 }
+
+#[tokio::test]
+async fn test_same_script_twice_with_state_file_enabled() {
```
ddoktorski, line 327: nit:
```suggestion
async fn test_run_script_twice_with_state_file_enabled() {
```

crates/sncast/src/state/state_file.rs
```diff
@@ -1,13 +1,82 @@
...
+            }
+        } else {
+            Self::default()
+        };
+
+        Ok(res)
+    }
+
+    #[must_use]
+    pub fn maybe_get_success_output(&self, tx_id: &str) -> Option<ScriptTransactionOutput> {
```
ddoktorski, line 50: I would remove the prefix `maybe` since we have a similar method named this way in `ScriptTransactionEntries`. Alternatively we can rename it in both places to something like `get_output_if_success`.

crates/sncast/src/state/state_file.rs
```diff
@@ -1,13 +1,82 @@
...
+        if let Some(state) = &mut self.inner {
+            state.executed_transactions_current_run.insert(
+                tx_id,
+                ScriptTransactionEntry::from(selector.to_string(), result),
+            );
+
+            write_txs_to_state_file(
+                &state.state_file,
+                state.executed_transactions_current_run.clone(),
+            )?;
```
ddoktorski, line 74: I am not sure if this is a good idea to overwrite the state file each time after inserting a transaction entry. Would it be possible to do it only once at the end of the script execution?

crates/sncast/src/state/state_file.rs
```diff
@@ -40,17 +109,71 @@ impl ScriptTransactionEntries {
...
+                return Some(entry.output.clone());
+            }
+        }
+        None
+    }
 }
 
 #[derive(Clone, Deserialize, Serialize, Debug, PartialEq)]
 pub struct ScriptTransactionEntry {
     pub name: String,
```
ddoktorski, line 130: Why this is `name` instead of `type`?

crates/sncast/tests/helpers/fixtures.rs
```diff
@@ -537,3 +540,37 @@ pub fn get_keystores_path(relative_path_from_cargo_toml: &str) -> String {
...
+    assert_eq!(tx_entry.status, status);
+
+    let ScriptTransactionOutput::ErrorResponse(response) = tx_entry.output.clone() else {
+        panic!("Wrong response")
+    };
+    for msg in msg_contains {
+        assert!(response.message.contains(msg));
+    }
+
+    assert!(tx_entry.timestamp > 1_709_853_748);
```
ddoktorski, line 560: Can we define a constant for this number?

crates/sncast/src/state/state_file.rs
```diff
@@ -1,13 +1,82 @@
...
+
+struct InnerStateManager {
+    state_file: Utf8PathBuf,
+    executed_transactions_prev_run: ScriptTransactionEntries,
+    executed_transactions_current_run: ScriptTransactionEntries,
+}
+
+#[derive(Default)]
+pub struct StateManager {
+    inner: Option<InnerStateManager>,
```
ddoktorski, line 25: Why this is `Option`?

ddoktorski (changes requested):

crates/sncast/src/starknet_commands/script/run.rs
```diff
@@ -58,6 +63,10 @@ pub struct Run {
     /// Specifies scarb package to be used
     #[clap(long)]
     pub package: Option<String>,
+
+    /// Do not write to state file
```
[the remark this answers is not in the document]
THenry14: Maybe rather "Do not use the state file" ?

crates/sncast/tests/data/scripts/state_file/src/rerun_failed_tx.cairo
THenry14: removed

crates/sncast/tests/e2e/script/general.rs
THenry14: Nothing should be overwritten, the point is to rerun a script that should have an invoke call that fails, but the script succeeds (from what I understand)

crates/sncast/src/state/state_file.rs
THenry14: it could be problematic, since the script may panic - then we loose every transaction up until this point. It probably could be done, but I'd be inclined to do this in separate PR

crates/sncast/src/state/state_file.rs
THenry14: `type` is a reserved keyword. besides, it marks a function `name`, not its type (in a contexts of script)

crates/sncast/tests/helpers/fixtures.rs
THenry14: sure, named it `SCRIPT_ORIGIN_TIMESTAMP`, but if you have a better idea - please share

crates/sncast/src/state/state_file.rs
ddoktorski: Currently if script panics, we end up losing all transactions from the previous execution, which isn't an ideal situation either. I'm fine with creating an issue and addressing it in a separate PR.

crates/sncast/src/state/state_file.rs
ddoktorski: In the state file, it shows a structure like `transaction` -> `name` -> for instance `declare`. Using `name` makes sense within the context of functions but is less fitting for transactions. Consider this as a nitpick.

crates/sncast/tests/e2e/script/general.rs
ddoktorski: So what's the point of adding the file [rerun_failed_tx_alpha-goerli_state.json](https://github.com/foundry-rs/starknet-foundry/pull/1844/files#diff-8f17219dce3b6579af66d2d4c1c83524e209704104f172ac44e1d9e3798c1808)?

crates/sncast/src/state/state_file.rs
ddoktorski: I would lean towards having `StateManager` that is `None` in such cases, rather than `StateManager` devoid of actual logic. The reasoning behind this design isn't immediately apparent when looking at the code, although I get that it might have been simpler to implement it this way.

crates/sncast/tests/e2e/script/general.rs
THenry14: Not sure tbh, my bet is that the intention was to make sure that failed tx do not overwrite state file when the status did not change, that's how I interpret this test tbh

crates/sncast/src/state/state_file.rs
THenry14: this is why we derive `Default` in here - so it is None i such cases :+1:

crates/sncast/src/state/state_file.rs
THenry14: https://github.com/foundry-rs/starknet-foundry/issues/1967

crates/sncast/src/state/state_file.rs
THenry14: Hmm, I see your point, as this is a nitpick let's consider this under https://github.com/foundry-rs/starknet-foundry/issues/1968 as I'm not sure on the impact it may have

crates/sncast/tests/e2e/script/general.rs
ddoktorski: I understand it differently. The name contains `rerun_failed`, so I expect that this test was supposed to simulate that the previously failed transaction will be repeated and the state file will be updated upon success. In my opinion such a test makes a lot of sense.

crates/sncast/tests/data/scripts/state_file/rerun_failed_tx_alpha-goerli_state.json
```diff
@@ -0,0 +1,17 @@
...
+          "type": "ErrorResponse",
+          "message": "An error occurred in the called contract = ContractErrorData { revert_error: \"Error in the called contract (0x00f6ecd22832b7c3713cfa7826ee309ce96a2769833f093795fafa1b8f20c48b):\\nError at pc=0:4835:\\nGot an exception while executing a hint: Requested contract address ContractAddress(PatriciaKey(StarkFelt(\\\"0x07537a17e169c96cf2b0392508b3a66cbc50c9a811a8a7896529004c5e93fdf6\\\"))) is not deployed.\\nCairo traceback (most recent call last):\\nUnknown location (pc=0:67)\\nUnknown location (pc=0:1835)\\nUnknown location (pc=0:2554)\\nUnknown location (pc=0:3436)\\nUnknown location (pc=0:4040)\\n\" }"
+        },
+        "status": "Error",
+        "timestamp": 1712136454,
+        "misc": null
+      }
+    }
+  }
+}
```
ddoktorski, line 17: nit: Newline

ddoktorski (approved):

PR #1846, opened 2024-03-08
Hotfix `--build-profile`

piotmag769:


war-in (approved):

MaksymilianDemitraszek:
We agreed with @Arcticae that it should be an error.

piotmag769:
@Arcticae now it throws an error instead

Issue #1202, opened 2023-11-27
Allow for easy switching of account cairo version


THenry14:
We do support creating accounts written in cairo 0. We also allow to pass custom class hash for account deployment. 
It seems though that OZ has released an account contract written in cairo1. We should add its class hash to cast, and allow users to create such account with a simple flag. The default should probably stay as cairo 0 for now though.

PR #1849, opened 2024-03-08
Write legacy information to accounts file

ddoktorski:
#1202 

This PR is part of the stack:

-> Write legacy information to accounts file (#1849)
-- Enable using Cairo 1 accounts in sncast (#1850)
-- Use Cairo 1 accounts in sncast tests (#1851)

## Introduced changes

- Write `legacy` information to accounts file in `account create` and `account add` to distinguish between Cairo 0 and Cairo >=1 accounts
- Rename `DEVNET_OZ_CLASS_HASH` -> `DEVNET_OZ_CLASS_HASH_CAIRO_0`

## Checklist

- [X] Linked relevant issue
- [X] Added relevant tests

### DIFF
--- a/crates/sncast/src/lib.rs
+++ b/crates/sncast/src/lib.rs
@@ -8,7 +8,7 @@ use serde_json::Value;
 use starknet::core::types::{
     BlockId, BlockTag,
     BlockTag::{Latest, Pending},
-    ContractErrorData, FieldElement,
+    ContractClass, ContractErrorData, FieldElement,
     StarknetError::{ClassHashNotFound, ContractNotFound, TransactionHashNotFound},
 };
 use starknet::core::utils::UdcUniqueness::{NotUnique, Unique};
@@ -48,6 +48,7 @@ struct Account {
     salt: Option<String>,
     deployed: Option<bool>,
     class_hash: Option<String>,
+    legacy: Option<bool>,
 }
 
 #[derive(Debug, PartialEq, Eq, Copy, Clone)]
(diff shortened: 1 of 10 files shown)


crates/sncast/src/lib.rs
```diff
@@ -48,6 +48,7 @@ struct Account {
     salt: Option<String>,
     deployed: Option<bool>,
     class_hash: Option<String>,
+    legacy: Option<bool>,
```
THenry14, line 51: I'm afraid that the definition of legacy may be too fluid for the future. Maybe we should just hold cairo version in here?

crates/sncast/src/lib.rs
```diff
@@ -327,6 +328,54 @@ fn get_account_from_accounts_file<'a>(
...
+async fn extract_class_hash_or_fetch(
+    class_hash: Option<FieldElement>,
+    address: FieldElement,
+    provider: &JsonRpcClient<HttpTransport>,
+) -> Result<FieldElement> {
+    match class_hash {
+        Some(class_hash) => Ok(class_hash),
+        None => get_class_hash_by_address(provider, address)
+            .await?
+            .ok_or_else(|| anyhow!("No class hash exists for the provided address {address:#x}")),
```
THenry14, line 353: is it possible? :D Maybe we should rephrase the error message to something like "Failed to fetch class hash for address {}` ?

crates/sncast/src/lib.rs
ddoktorski: This way, it aligns with the keystore account files, making it easier to use the same `Account` struct. Moreover, differentiating between Cairo 1 and 2 accounts seems non-trivial (please let me know if there a simple to way to achieve this), and it's not relevant in terms of encoding.

crates/sncast/src/lib.rs
ddoktorski: In this case "No class hash exists" actually means that the provided address does not exist in the given network, so maybe let's rephrase it to `The provided address {} does not exist`?

crates/sncast/src/lib.rs
THenry14: ok, makes sense. guess we will have to wait and see how thing will look like with newer cairo versions then.

crates/sncast/src/lib.rs
ddoktorski: After all I decided to remove the `extract_class_hash_or_fetch` function in order to limit RPC calls. All RPC errors related to getting the contract class are now being handled by the `handle_rpc_error` function.

crates/sncast/src/lib.rs
```diff
@@ -327,6 +328,42 @@ fn get_account_from_accounts_file<'a>(
...
 }
 
+pub async fn check_if_legacy_contract(
+    class_hash: Option<FieldElement>,
+    address: FieldElement,
+    provider: &JsonRpcClient<HttpTransport>,
+) -> Result<bool> {
+    let contract_class = match class_hash {
+        Some(class_hash) => provider.get_class(BlockId::Tag(Pending), class_hash).await,
+        None => provider.get_class_at(BlockId::Tag(Pending), address).await,
```
THenry14, line 338: should this use `get_class_hash_by_address`?

ddoktorski: We want to receive a contract class, not a class hash. I initially moved `get_class_hash_by_address` here since it was utilized in the first solution. If `account/add.rs` is a better location for it, it can certainly be moved back.

THenry14: ahh shit, misread. sorry.

THenry14 (approved):

PR #1850, opened 2024-03-08
Enable using Cairo 1 accounts in sncast

ddoktorski:
#1202 

This PR is part of the stack:

-- Write legacy information to accounts file (#1849)
-> Enable using Cairo 1 accounts in sncast (#1850)
-- Use Cairo 1 accounts in sncast tests (#1851)

## Introduced changes

- Modify logic in `get_account` function to enable using Cairo >=1 accounts

## Checklist

- [X] Linked relevant issue
- [X] Added relevant tests

### DIFF
--- a/crates/sncast/src/lib.rs
+++ b/crates/sncast/src/lib.rs
@@ -225,7 +225,7 @@ async fn build_account(
             &account_data.address
         )
     })?;
-    verify_account_address(address, provider).await?;
+    verify_account_address(address, chain_id, provider).await?;
 
     let class_hash = account_data
         .class_hash
@@ -252,14 +252,16 @@ async fn build_account(
 
 async fn verify_account_address(
     address: FieldElement,
+    chain_id: FieldElement,
     provider: &JsonRpcClient<HttpTransport>,
 ) -> Result<()> {
     match provider.get_nonce(BlockId::Tag(Pending), address).await {
         Ok(_) => Ok(()),
         Err(error) => {
             if let StarknetError(ContractNotFound) = error {
+                let decoded_chain_id = decode_chain_id(chain_id);
                 Err(anyhow!(
-                    "No account exists with the provided address {address:#x}"
+                    "Account with address {address:#x} not found on network {decoded_chain_id}"
                 ))
             } else {
                 Err(handle_rpc_error(error))
--- a/crates/sncast/src/lib.rs
+++ b/crates/sncast/src/lib.rs
@@ -41,7 +41,7 @@ pub mod response;
 pub mod state;
 
 #[derive(Deserialize, Serialize, Clone, Debug)]
-struct Account {
+struct AccountData {
     private_key: String,
     public_key: String,
     address: String,
@@ -199,9 +199,9 @@ pub async fn get_account<'a>(
 ) -> Result<SingleOwnerAccount<&'a JsonRpcClient<HttpTransport>, LocalWallet>> {
     let chain_id = get_chain_id(provider).await?;
     let account_info = if let Some(keystore) = keystore {
-        get_account_info_from_keystore(account, &keystore)?
+        get_account_data_from_keystore(account, &keystore)?
     } else {
-        get_account_info_from_accounts_file(account, chain_id, accounts_file)?
+        get_account_data_from_accounts_file(account, chain_id, accounts_file)?
     };
 
     let account = build_account(account_info, chain_id, provider).await?;
@@ -210,23 +210,23 @@ pub async fn get_account<'a>(
 }
 
 async fn build_account(
-    account_info: Account,
+    account_data: AccountData,
     chain_id: FieldElement,
     provider: &JsonRpcClient<HttpTransport>,
 ) -> Result<SingleOwnerAccount<&JsonRpcClient<HttpTransport>, LocalWallet>> {
     let signer = LocalWallet::from(SigningKey::from_secret_scalar(
-        parse_number(&account_info.private_key)
+        parse_number(&account_data.private_key)
             .context("Failed to convert private key to FieldElement")?,
     ));
-    let address = parse_number(&account_info.address).with_context(|| {
+    let address = parse_number(&account_data.address).with_context(|| {
         format!(
             "Failed to convert account address = {} to FieldElement",
-            &account_info.address
+            &account_data.address
         )
     })?;
     verify_account_address(address, provider).await?;
 
-    let class_hash = account_info
+    let class_hash = account_data
         .class_hash
         .map(|class_hash| {
             parse_number(&class_hash).with_context(|| {
@@ -239,7 +239,7 @@ async fn build_account(
         .transpose()?;
 
     let account_encoding =
-        get_account_encoding(account_info.legacy, class_hash, address, provider).await?;
+        get_account_encoding(account_data.legacy, class_hash, address, provider).await?;
 
     let mut account =
         SingleOwnerAccount::new(provider, signer, address, chain_id, account_encoding);
@@ -280,7 +280,10 @@ pub async fn check_class_hash_exists(
     }
 }
 
-fn get_account_info_from_keystore(account: &str, keystore_path: &Utf8PathBuf) -> Result<Account> {
+fn get_account_data_from_keystore(
+    account: &str,
+    keystore_path: &Utf8PathBuf,
+) -> Result<AccountData> {
     if !keystore_path.exists() {
         bail!("Failed to find keystore file");
     }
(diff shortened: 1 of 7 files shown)


crates/sncast/src/lib.rs
```diff
@@ -269,63 +286,92 @@ fn get_account_from_keystore<'a>(
...
-        ExecutionEncoding::Legacy,
-    ))
+        salt,
+        deployed,
+        class_hash,
+        legacy,
+    })
 }
 
-fn get_account_from_accounts_file<'a>(
```
ddoktorski, line 298: Note: `get_account_info_from_accounts_file` and `get_account_info_from_keystore` can also be utilized within `starknet_commands/account/deploy.rs` to refactor and simplify the code.

crates/sncast/src/lib.rs
```diff
@@ -269,63 +286,92 @@ fn get_account_from_keystore<'a>(
...
+async fn get_account_encoding(
+    legacy: Option<bool>,
+    class_hash: Option<FieldElement>,
+    address: FieldElement,
+    provider: &JsonRpcClient<HttpTransport>,
+) -> Result<ExecutionEncoding> {
+    if let Some(legacy) = legacy {
+        Ok(map_encoding(legacy))
+    } else {
+        let legacy = check_if_legacy_contract(class_hash, address, provider).await?;
```
THenry14, line 372: I'm struggling a bit to understand what problem this function (get_account_encoding) tries to solve, sorry. do you mind sharing a rationale behind this? will we call the network each time to check if an account is legacy, if there is no such info in account file?

obraz

crates/sncast/src/lib.rs
```diff
@@ -252,12 +274,7 @@ pub async fn check_class_hash_exists(
     }
 }
 
-fn get_account_from_keystore<'a>(
-    provider: &'a JsonRpcClient<HttpTransport>,
-    chain_id: FieldElement,
-    keystore_path: &Utf8PathBuf,
-    account: &str,
-) -> Result<SingleOwnerAccount<&'a JsonRpcClient<HttpTransport>, LocalWallet>> {
+fn get_account_info_from_keystore(account: &str, keystore_path: &Utf8PathBuf) -> Result<Account> {
```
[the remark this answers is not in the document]
ddoktorski: It depends what we mean by "account" 😅 Currently, It returns an `Account` struct, where it previously returned `SingleOwnerAccount` from starknet-rs. Instead of renaming the function, I would prefer to rename the `Account` struct to something like `AccountInfo`/`AccountData`. Wdyt?

crates/sncast/src/lib.rs
ddoktorski: When we create a `SingleOwnerAccount` we need to provide an encoding parameter that decides how calls will be encoded. There are 2 options: `ExecutionEncoding::Legacy` and `ExecutionEncoding::New`. Until now, we had hardcoded `ExecutionEncoding::Legacy`, but to support Cairo >=1 accounts, we need some logic to choose the correct encoding.

In the account file we can store information for `address`, `class_hash` and `legacy`, but only `address` is required. The function to choose the correct account encoding operates as follows:
- If `legacy` is not `None`, then it determines the encoding solely based on this boolean value. In this case we don't need to execute any RPC calls
- Otherwise it calls `check_if_legacy_contract`, this function first checks the encoding based on `class_hash`, and if `class_hash` is `None`, it checks based on `address`

As I think now the function `check_if_legacy_contract` can be slightly simplified, given that we can always get `ContractClass` only based on `address` using the `getClassAt` starknet method.

crates/sncast/src/lib.rs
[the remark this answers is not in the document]
THenry14: AccountData makes sense

crates/sncast/src/lib.rs
THenry14: would it make sense to, if there is no info in the file, do the call to network and store legacy info in the file, so it can be later retrieved? (as opposed to call to network everytime)

ddoktorski: Generally I think it would make sense. I am not sure though if this should be done in get account logic, it doesn't seem to be an appropriate place in the code to overwrite the accounts file. Currently we make changes in the accounts file in commands `account add/create/deploy`. Logically this makes sense and is expected. I'm leaning towards not implementing logic to store `legacy` data here. If there is any latency issue arise, it's very easy for users to manually add `"legacy": true` to their accounts file.

ddoktorski: Correct me if I'm wrong, but in script execution we only need to get account once? If this the case then we need to make only one additional RPC call (after making the change that I mentioned in the first message).

THenry14: as discussed yesterday offline, we need to write this info to the file, as there are more requests.

Main idea is to just do it silently in get_account.
The other idea would be to print a warning that legacy info is missing from accounts file, and have something like `account info-update` (name to be determined) that users would be able to call in order to update the file. That way we would by default make eccessive calls, but it would be more fair to users since we will give them an explicit way to update account info and get rid of these calls.

Both approaches have their flaws, but imo both are perfectly acceptable. wdyt?

crates/sncast/src/lib.rs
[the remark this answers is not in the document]
ddoktorski: https://github.com/foundry-rs/starknet-foundry/pull/1850/commits/f3f21ed2ec862b8bf4e6f23dee6941cfa6c04644

crates/sncast/src/lib.rs
```diff
@@ -269,63 +286,92 @@ fn get_account_from_keystore<'a>(
...
+    let class_hash = parse_to_string(deployment_info, "class_hash");
+    let salt = parse_to_string(deployment_info, "salt");
+    let deployed = parse_to_string(deployment_info, "status").map(|status| &status == "deployed");
+
+    let variant_info = account_info
+        .get("variant")
+        .expect("Failed to find variant key in the account JSON file");
+    let public_key = parse_to_string(variant_info, "public_key")
+        .ok_or_else(|| anyhow::anyhow!("Failed to get public key from account JSON file"))?;
+    let legacy = variant_info.get("legacy").and_then(Value::as_bool);
```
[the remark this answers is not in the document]
ddoktorski: https://github.com/foundry-rs/starknet-foundry/pull/1850/commits/64574ab0e709dfa0b61131d2e95cd2320c0a58b8

crates/sncast/src/lib.rs
ddoktorski: Adding a command to update account information seems reasonable to me. Created an issue to track this https://github.com/foundry-rs/starknet-foundry/issues/1901

crates/sncast/src/lib.rs
```diff
@@ -252,80 +275,104 @@ pub async fn check_class_hash_exists(
...
+    raise_if_empty(name, "Account name")?;
+    check_account_file_exists(path)?;
+
+    let accounts: HashMap<String, HashMap<String, AccountData>> = read_and_parse_json_file(path)?;
+    let network_name = chain_id_to_network_name(chain_id);
+
+    let account = accounts
+        .get(&network_name)
+        .and_then(|accounts_map| accounts_map.get(name))
+        .cloned();
```
THenry14, line 346: it would be cool to get rid of this clone, we could probably just merge this + ok_or_else (below) and return directly

ddoktorski: I don't see an easy way to get rid of `cloned` at the moment, but I simplified the code slightly in https://github.com/foundry-rs/starknet-foundry/pull/1850/commits/a5edc902094e1d45f6f0c68ee4db5e9becadb6d3

crates/sncast/src/lib.rs
```diff
@@ -220,24 +199,68 @@ pub async fn get_account<'a>(
...
+    address: FieldElement,
+    provider: &JsonRpcClient<HttpTransport>,
+) -> Result<()> {
+    match provider.get_nonce(BlockId::Tag(Pending), address).await {
         Ok(_) => Ok(()),
         Err(error) => {
             if let StarknetError(ContractNotFound) = error {
-                Err(anyhow!("Invalid account address"))
+                Err(anyhow!(
+                    "No account exists with the provided address {address:#x}"
```
THenry14, line 262: nitpick: let's add chain_id to this message as well; just because an address does not exist in network X, it does not mean it doesn't for network Y as well

THenry14 (approved):

crates/sncast/src/lib.rs
ddoktorski: https://github.com/foundry-rs/starknet-foundry/pull/1850/commits/6ac076e32698fd18a88cec63ece2c4766ff9f3ee

PR #1851, opened 2024-03-08
Use Cairo 1 accounts in sncast tests

ddoktorski:
#1202 

This PR is part of the stack:

-- Write legacy information to accounts file (#1849)
-- Enable using Cairo 1 accounts in sncast (#1850)
-> Use Cairo 1 accounts in sncast tests (#1851)

## Introduced changes

- Use Cairo 1 accounts in sncast tests by default
- Run basic tests for both Cairo 0 and 1 accounts

## Checklist

- [X] Linked relevant issue
- [X] Added relevant tests


crates/sncast/tests/helpers/devnet.rs
```diff
@@ -34,7 +34,14 @@ fn start_devnet() {
...
 
     Command::new("tests/utils/devnet/bin/starknet-devnet")
-        .args(["--port", &port, "--seed", &SEED.to_string()])
+        .args([
+            "--port",
+            &port,
+            "--seed",
+            &SEED.to_string(),
+            "--account-class",
+            "cairo1",
```
ddoktorski, line 43: Note: In the current version of `starknet-devnet-rs`, which supports RPC `0.7.0`, it has become the default setting. Therefore, it can be potentially removed later.

crates/sncast/tests/helpers/constants.rs
```diff
@@ -15,4 +15,4 @@ pub const DEVNET_OZ_CLASS_HASH_CAIRO_0: &str =
 pub const DEVNET_PREDEPLOYED_ACCOUNT_ADDRESS: &str =
     "0x691a61b12a7105b1372cc377f135213c11e8400a546f6b0e7ea0296046690ce";
 pub const DEVNET_OZ_CLASS_HASH_CAIRO_1: &str =
-    "0x61dac032f228abef9c6626f995015233097ae253a7f72d68552db02f2971b8f";
+    "0x4c6d6cf894f8bc96bb9c525e6853e5483177841f7388f74a46cfda6f028c755";
```
ddoktorski, line 18: Note: It will be necessary to update it after devnet updates.

crates/sncast/tests/data/accounts/accounts.json
```diff
@@ -47,10 +47,17 @@
             "public_key": "0x4fdaeb60e75399877bd5d14d3dac2e805f3c2a6732851a19db63fee0cfae00d",
             "address": "0x68d6e92b8e2310c550178b2fe484d5cb89dd6aa93ae804afd4d124e9d6c3cd3"
         },
-        "user9": {
+        "cairo1": {
```
THenry14, line 50: these are cairo0 accounts, right?

crates/sncast/tests/helpers/constants.rs
THenry14: let's add in code comment with a path where it is defined in devnet, maybe?

crates/sncast/tests/helpers/devnet.rs
THenry14: https://github.com/foundry-rs/starknet-foundry/issues/1862

crates/sncast/tests/data/accounts/accounts.json
ddoktorski: After setting `--account-class cairo1` in the `starknet-devnet` command, all predeployed accounts are Cairo 1.

THenry14: let me rephrase my question - what are the accounts you have added here?

ddoktorski: I've renamed `user9` -> `cairo1` and included the account that we deploy in the `deploy_keystore_account` fixture to this file (`cairo0`). I haven't added any new account to our devnet network that wasn't already present.

THenry14: > I've renamed user9 -> cairo1

what is the rationale behind this?

> included the account that we deploy in the deploy_keystore_account 

keystore accounts should never be added to accounts file, these are different types altogether and shouldn not be mixed

ddoktorski: > what is the rationale behind this?

This account is used in parameterized tests, that execute the same command for 2 accounts. It's more self explanatory to have `--account cairo1`, rather that `--account user9`.

> keystore accounts should never be added to accounts file, these are different types altogether and should not be mixed

We don't use this account in any other test, by reusing it there is no need to add redundant code. However, if you prefer to use the keystore account only for deployment, I can generate a separate account for Cairo 0 and add a fixture that deploys it.

ddoktorski: https://github.com/foundry-rs/starknet-foundry/pull/1851/commits/7f41085cdea12063307b7d8c854cbec2bbd7468e

crates/sncast/tests/helpers/constants.rs
THenry14: > This needs to be updated whenever we update the devnet

is this true? I think class hashes are idependent of devnet version, and we will need to update this only if devnet updates OZ contract, but let's double-check

crates/sncast/tests/helpers/devnet.rs
THenry14: the support is merged to master, let's conform to that

crates/sncast/tests/helpers/fixtures.rs
```diff
@@ -79,41 +79,83 @@ pub async fn deploy_keystore_account() {
...
+    let private_key = get_from_json_as_str(account_data, "private_key");
+
+    let private_key = SigningKey::from_secret_scalar(
+        parse_number(private_key).expect("Failed to convert private key to FieldElement"),
+    );
+
+    deploy_account_helper(address, DEVNET_OZ_CLASS_HASH_CAIRO_0, salt, private_key).await;
+}
+
+async fn deploy_account_helper(
```
THenry14, line 125:
```suggestion
async fn deploy_test_account(
```
nit: the function name implies we will deploy something that is an account helper, not an account

crates/sncast/tests/helpers/constants.rs
ddoktorski: > we will need to update this only if devnet updates OZ contract

This is correct. However, it looks like class hashes are actually updated with each devnet version. For example compare [v0.0.2](https://github.com/0xSpaceShard/starknet-devnet-rs/blob/v0.0.2/crates/starknet-devnet-core/src/constants.rs#L20) and [v0.0.3](https://github.com/0xSpaceShard/starknet-devnet-rs/blob/v0.0.3/crates/starknet-devnet-core/src/constants.rs#L21)

THenry14: Perhaps, but we should not expect this to happen often

crates/sncast/tests/e2e/script/deploy.rs
```diff
@@ -61,21 +64,9 @@ async fn test_same_salt_and_class_hash_deployed_twice() {
         output,
         indoc! {r#"
         ScriptCommandError::WaitForTransactionError(WaitForTransactionError::TransactionError(TransactionError::Reverted(ErrorData { msg: "Error in the called contract ([..]):
-        Error at pc=0:81:
```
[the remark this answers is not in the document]
ddoktorski: It's necessary to make changes in these outputs after using Cairo 1 accounts. Therefore, I think it makes sense to both fix and enhance them at once.

crates/sncast/tests/helpers/devnet.rs
ddoktorski: https://github.com/foundry-rs/starknet-foundry/pull/1851/commits/ac91f9dc315adf88880ad66d910b2c5c1d43d874

crates/sncast/tests/helpers/constants.rs
ddoktorski: I've removed that comment, left only the relevant link to the devnet

crates/sncast/tests/e2e/script/deploy.rs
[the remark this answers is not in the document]
THenry14: that's weird, but may actually make sense since this may be connected with cairo0 internals somehow. thanks for the info.

THenry14 (approved):

Issue #1852, opened 2024-03-10
`#../appendix/xxx` should be `../appendix/xxx` otherwize jump will not work


bodhi-crypo:
`#../appendix/xxx` should be `../appendix/xxx` otherwize jump will not work

_Originally posted by @bodhi-crypo in https://github.com/foundry-rs/starknet-foundry/pull/1713#discussion_r1518803868_

PR #1860, opened 2024-03-11
Isolate configuration crate from cast

MaksymilianDemitraszek:
commit-id:83f55207

---

**Stack**:
- #1861
- #1860 ⬅

### DIFF
--- a/crates/configuration/src/lib.rs
+++ b/crates/configuration/src/lib.rs
@@ -1,0 +1,227 @@
+use anyhow::{anyhow, Context, Result};
+use std::fs;
+
+use camino::Utf8PathBuf;
+use tempfile::{tempdir, TempDir};
+use toml::Value;
+pub const CONFIG_FILENAME: &str = "snfoundry.toml";
+
+/// Defined in snfoundry.toml
+/// Configuration not associated with any specific package
+pub trait GlobalConfig {
+    #[must_use]
+    fn tool_name() -> &'static str;
+
+    fn from_raw(config: serde_json::Value) -> Result<Self>
+    where
+        Self: Sized;
+}
+
+/// Defined in scarb manifest
+/// Configuration associated with a specific package
+pub trait PackageConfig {}
+
+fn get_with_ownership(config: serde_json::Value, key: &str) -> Option<serde_json::Value> {
+    match config {
+        serde_json::Value::Object(mut map) => map.remove(key),
+        _ => None,
+    }
+}
+
+pub fn get_profile(
+    raw_config: serde_json::Value,
+    tool: &str,
+    profile: &Option<String>,
+) -> Result<serde_json::Value> {
+    let profile_name = profile.as_deref().unwrap_or("default");
+    let tool_config = get_with_ownership(raw_config, tool)
+        .context("Failed to find {tool} config in snfoundry.toml file")?;
+
+    match get_with_ownership(tool_config, profile_name) {
+        Some(profile_value) => Ok(profile_value),
+        None if profile_name == "default" => Ok(serde_json::Value::Object(Default::default())),
+        None => Err(anyhow!("Profile [{}] not found in config", profile_name)),
+    }
+}
+
+pub fn load_global_config<T: GlobalConfig + Default>(
+    path: &Option<Utf8PathBuf>,
+    profile: &Option<String>,
+) -> Result<T> {
+    let config_path = path
+        .as_ref()
+        .and_then(|p| search_config_upwards_relative_to(p).ok())
+        .or_else(|| find_config_file().ok());
+
+    match config_path {
+        Some(path) => {
+            let raw_config_toml = fs::read_to_string(path)
+                .expect("Failed to read snfoundry.toml config file")
+                .parse::<Value>()
+                .context("Failed to parse snfoundry.toml config file")?;
+
+            let raw_config_json = serde_json::to_value(raw_config_toml)
+                .expect("Conversion from TOML value to JSON value should not fail.");
+
+            let profile = get_profile(raw_config_json, T::tool_name(), profile)?;
+            T::from_raw(profile)
+        }
+        None => Ok(T::default()),
+    }
+}
+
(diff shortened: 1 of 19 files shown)


MaksymilianDemitraszek:
Related #1852

## Introduced changes



- This is an intro work to include env variables support in forge and cast configs

Follow up work:
- isolate forge config into `configuration` crate
- add support for resolving the env variables

## Checklist



- [ ] Linked relevant issue
- [ ] Added relevant tests

crates/configuration/src/lib.rs
```diff
@@ -0,0 +1,266 @@
...
+
+/// Defined in scarb manifest
+/// Configuration associated with a specific package
+pub trait PackageConfig {}
+
+pub fn get_profile(raw_config: &Value, tool: &str, profile: &Option<String>) -> Result<Value> {
+    let profile_name = profile.as_deref().unwrap_or("default");
+    let config = raw_config
+        .get(tool)
+        .expect("Failed to find sncast config in snfoundry.toml file");
```
THenry14, line 28: probably a nitpick, since firge does not use snfoundry.toml yet but since we're in a shared crate, shouldn't this be generalised?

crates/configuration/src/lib.rs
```diff
@@ -0,0 +1,266 @@
...
+        })
+}
+
+pub fn find_config_file() -> Result<Utf8PathBuf> {
+    search_config_upwards_relative_to(&Utf8PathBuf::try_from(
+        std::env::current_dir().expect("Failed to get current directory"),
+    )?)
+}
+
+pub trait PropertyFromCastConfig: Sized {
```
THenry14, line 78: Should this + impls names be generalized as well? I suppose these may not be cast specific, and perfectly usable for forge as well

crates/configuration/src/lib.rs
```diff
@@ -0,0 +1,266 @@
...
+    use std::fs;
+
+    use tempfile::tempdir;
+
+    use super::*;
+
+    #[test]
+    fn find_config_in_current_dir() {
+        let tempdir =
+            copy_config_to_tempdir("../sncast/tests/data/files/stubtool_snfoundry.toml", None);
```
THenry14, line 145: I guess it would make sense to hold test files for this crate in this crate

crates/sncast/src/starknet_commands/account/mod.rs
```diff
@@ -162,11 +161,24 @@ pub fn add_created_profile_to_configuration(
...
-    use sncast::helpers::configuration::{copy_config_to_tempdir, CastConfig};
+    use configuration::CONFIG_FILENAME;
+    use sncast::helpers::configuration::CastConfig;
     use sncast::helpers::constants::DEFAULT_ACCOUNTS_FILE;
     use std::fs;
+    use tempfile::{tempdir, TempDir};
 
     use crate::starknet_commands::account::add_created_profile_to_configuration;
+    #[must_use]
+    pub fn copy_config_to_tempdir(src_path: &str, additional_path: Option<&str>) -> TempDir {
```
THenry14, line 172: this is the same function that is defined in configuration crate, let's just import it :+1:

crates/configuration/src/lib.rs
MaksymilianDemitraszek: It should, I missed this one :0

crates/configuration/src/lib.rs
MaksymilianDemitraszek: Not sure what do you mean 🤔

THenry14: we are in a different crate (configuration), so should we use a file (stubtool_snfoundry.toml) that sits in a different crate? I think the file could be moved

MaksymilianDemitraszek: Ah yes sure, I intended to do so, forgot about it

crates/configuration/src/lib.rs
```diff
@@ -0,0 +1,266 @@
...
+
+pub fn get_property<T>(entries: &Value, field: &str) -> Option<T>
+where
+    T: PropertyFromCastConfig + Default,
+{
+    entries.get(field).and_then(T::from_toml_value)
+}
+
+#[must_use]
+pub fn copy_config_to_tempdir(src_path: &str, additional_path: Option<&str>) -> TempDir {
```
Arcticae, line 123: It's only used for tests

MaksymilianDemitraszek: Yes but in different modules :(
I wanted to move it to test module but that actually is not trivial bcs it is used in multiple places. This is a reason behind duplicate I have left.

crates/configuration/src/lib.rs
MaksymilianDemitraszek: Done

crates/configuration/src/lib.rs
MaksymilianDemitraszek: Done

crates/configuration/src/lib.rs
MaksymilianDemitraszek: Done

crates/sncast/src/starknet_commands/account/mod.rs
MaksymilianDemitraszek: Done

crates/configuration/src/lib.rs
```diff
@@ -0,0 +1,227 @@
...
+/// Defined in scarb manifest
+/// Configuration associated with a specific package
+pub trait PackageConfig {}
+
+fn get_with_ownership(config: serde_json::Value, key: &str) -> Option<serde_json::Value> {
+    match config {
+        serde_json::Value::Object(mut map) => map.remove(key),
+        _ => None,
+    }
+}
```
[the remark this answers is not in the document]
MaksymilianDemitraszek: We need to remove it to avoid cloning. Otherwise you could not take the ownership

THenry14 (approved):

PR #1861, opened 2024-03-11
Unified configuration with cast

MaksymilianDemitraszek:
commit-id:6467410e

---

**Stack**:
- #1911
- #1910
- #1861 ⬅

### DIFF
--- a/crates/configuration/src/lib.rs
+++ b/crates/configuration/src/lib.rs
@@ -1,4 +1,5 @@
 use anyhow::{anyhow, Context, Result};
+use scarb_metadata::{Metadata, PackageId};
 use std::fs;
 
 use camino::Utf8PathBuf;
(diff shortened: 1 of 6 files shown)


crates/configuration/src/lib.rs
```diff
@@ -21,7 +22,14 @@ pub trait GlobalConfig {
 
 /// Defined in scarb manifest
 /// Configuration associated with a specific package
-pub trait PackageConfig {}
+pub trait PackageConfig {
+    #[must_use]
+    fn tool_name() -> String;
```
Draggu, line 27: I don't see a case where `&str` would not be enough

Draggu (approved):

crates/configuration/src/lib.rs
MaksymilianDemitraszek: We probably would have to make it static though, no?

Draggu: Yes

Arcticae (approved):

piotmag769 (approved):

PR #1875, opened 2024-03-14
Module import granularity

drknzz:


drknzz:
Currently there is no way to enforce this without using [unstable features](https://rust-lang.github.io/rustfmt/?version=master&search=import#imports_granularity)

PR #1893, opened 2024-03-15
Add utility construction methods to String which clarify which version is used

### LINKED ISSUE #1314: Add utility construction methods to `String` which clarify which version is used
i.e. `from_dec_string`/`from_hex_string` instead of our custom `from_`
The problem is with misuse of the methods on different types of strings (dec string could be interpreted as a hex one)

Also in scope of this issue is removing this function and its' usage:

https://github.com/foundry-rs/starknet-foundry/blob/0c3d2fe4ab31aa4484fa216417408ae65a149efe/crates/cheatnet/src/runtime_extensions/forge_runtime_extension/file_operations.rs#L69

Draggu:
Closes #1314

## Introduced changes

- Title

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/cheatnet/tests/starknet/forking.rs
```diff
@@ -731,12 +704,10 @@ fn test_calling_nonexistent_url() {
...
-    let contract_address = Felt252::from(
-        BigUint::from_str(
-            "3216637956526895219277698311134811322769343974163380838558193911733621219342",
-        )
-        .unwrap(),
+    let contract_address = Felt252::try_from_dec_str(
+        "3216637956526895219277698311134811322769343974163380838558193911733621219342",
     )
+    .unwrap()
     .into_();
```
THenry14, line 711: why does this one need to be `into_()`?

crates/conversions/src/felt252.rs
```diff
@@ -69,6 +83,47 @@ impl FromShortString<Felt252> for Felt252 {
...
+}
+
+impl<T> TryInferFormat for T
+where
+    T: From<Felt252>,
+{
+    fn infer_format_and_parse(mut value: &str) -> Result<Vec<Self>, ParseFeltError> {
+        let mut result = Err(ParseFeltError);
+
+        let is_short_string = value.starts_with('\'') && value.ends_with('\'');
```
THenry14, line 98: is this assumption safe? shouldn't we rather check the actual length/size instead?

crates/conversions/src/felt252.rs
```diff
@@ -69,6 +83,47 @@ impl FromShortString<Felt252> for Felt252 {
...
+
+        if is_short_string {
+            let value_escaped = value.replace("\\n", "\n").replace("\\'", "'");
+
+            result =
+                result.or_else(|_| Felt252::from_short_string(&value_escaped).map(|f| vec![f]));
+        }
+
+        if is_string {
+            //TODO
```
THenry14, line 113: let's improve on todo, preferably add issue number

crates/conversions/src/felt252.rs
THenry14: ok I now see that this is assumed in snforge_std docs, but still should be only depend on that? seems rather risky to me

Draggu: `Felt252::from_short_string` checks it already so no need to duplicate logic here.
```rs
fn from_short_string(short_string: &str) -> Result<Felt252, ParseFeltError> {
    if short_string.len() <= 31 && short_string.is_ascii() {
        Ok(Felt252::from_bytes_be(short_string.as_bytes()))
    } else {
        Err(ParseFeltError)
    }
}
```
This only decides if apply `Felt252::from_short_string` or not

crates/conversions/src/felt252.rs
Draggu: Missed it, outdated

crates/cheatnet/tests/starknet/forking.rs
Draggu: Don't need. Fixed

Arcticae (approved):
LGTM, although the changes to the API (format) of files should be included in changelog

crates/conversions/src/felt252.rs
THenry14: I see your point, but it just seems counter-intuitive to me that `is_short_string` does not check if something is actually a short string, but `from_short_string` that is being called later in the code does.

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/file_operations.rs
```diff
@@ -124,98 +114,32 @@ mod tests {
             "ad": "string",
             "test": ["1",2,"3",4]
         }"#;
-        let result = json_values_sorted_by_keys(string).unwrap();
-        let expected_result = ["\"string\"", "4", "\"1\"", "2", "\"3\"", "4"];
+        let (_temp, file_path) = create_file(string);
+        let result = read_json(file_path).unwrap();
+        let mut expected_result = "string".serialize_as_felt252_vec();
+        expected_result.push(Felt252::from(4)); // .test.len()
```
THenry14, line 120:
```suggestion
        expected_result.push(Felt252::from(4));
```

nitpick, I don't think this is necessary but if you disagree, that's fine. In that case we might want to improve on this comment.

THenry14 (approved):

PR #1906, opened 2024-03-18
Enable blobs for data availability gas calculations

### LINKED ISSUE #1819: Turn on blobs-based gas calculation

Forge

### Feature Request

We should probably consider allowing the users configure the cost estimation based on the commitment scheme (right now `use_kzg_da` flag is one of the possible configurations). The gas estimation will be lower if it's turned on. 

This can also be left out, if we want to consider the gas estimation to be a max possible value (this config could be based later on on the changing gas prices, but right now it's based on system config).

Arcticae:
Closes #1819

## Introduced changes

-

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Default `chain_id` has been changed from `SN_GOERLI` to `SN_SEPOLIA`
 - Supported RPC version is now 0.7.0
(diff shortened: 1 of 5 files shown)


drknzz (approved):
Should there be a statement in docs about kzg being on? From what I understand it could at some point (unlikely) be turned off on the network

piotmag769 (commented):
Is it finally all aligned (starknet docs, blockifier and our code)? :o

PR #1907, opened 2024-03-18
Revamp event mechanics

### LINKED ISSUE #1889: Revamp event mechanics

snforge

### Description

1. We should implement a way of automatic writing the emitted events to the segment where the events are stored, instead of requiring the user to fetch the events manually (which is easy to forget). 
2. We should send the event to each spy that is elligible for receiving the spy (its group captures the address that the event is from)

Draggu:
Closes #1889

## Introduced changes

- Revamp event mechanics

## Checklist

- [x] Linked relevant issue
- [ ] Added relevant tests


piotmag769:
We decided on different approach so I wouldn't break it twice. If you disagree with event emission writing stuff to vm memory implicitly feel free to discuss it on Slack

Arcticae:
^ Yep, we'd like to start this when we start the refactor and use different approach (more compilcated than this, requires keeping segment info in the runtime and writing to that segment dynamically on event). Closing for now, sine we don't want to start this refactor yet.

Issue #1746, opened 2024-02-21
Add profile to empty `snfoundry.toml` file fails


ddoktorski:
### Which components does the task require to be changed? (think hard pls)

sncast

### Description

If `snfoundry.toml` file doesn't contain any sncast profiles, an attempt to add a new profile using the `--add-profile` flag will result in a failure.

To reporoduce:
1. Create an empty `snfoundry.toml` file
2. From the same directory, execute `account create` or `account add` with the flag `--add-profile` passed

PR #1911, opened 2024-03-18
Added empty config by default if tool not found

MaksymilianDemitraszek:
commit-id:d061bd7a

---

**Stack**:
- #1921
- #1911 ⬅

### DIFF
--- a/crates/configuration/src/lib.rs
+++ b/crates/configuration/src/lib.rs
@@ -44,7 +44,7 @@ pub fn get_profile(
 ) -> Result<serde_json::Value> {
     let profile_name = profile.as_deref().unwrap_or("default");
     let tool_config = get_with_ownership(raw_config, tool)
-        .context("Failed to find {tool} config in snfoundry.toml file")?;
+        .unwrap_or(serde_json::Value::Object(serde_json::Map::new()));
 
     match get_with_ownership(tool_config, profile_name) {
         Some(profile_value) => Ok(profile_value),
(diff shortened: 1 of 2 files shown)


MaksymilianDemitraszek:
Closes #1746

## Introduced changes



- Env variables are supported in configuration files

## Checklist



- [x] Linked relevant issue
- [x] Added relevant tests

crates/configuration/src/lib.rs
```diff
@@ -44,7 +44,7 @@ pub fn get_profile(
 ) -> Result<serde_json::Value> {
     let profile_name = profile.as_deref().unwrap_or("default");
     let tool_config = get_with_ownership(raw_config, tool)
-        .context("Failed to find {tool} config in snfoundry.toml file")?;
+        .unwrap_or(serde_json::Value::Object(serde_json::Map::new()));
```
THenry14, line 47: Is this a good idea? This may make the troubleshooting a bit harder, as we silently return an empty object instead of an error when someone specifies a profile that does not exist.

MaksymilianDemitraszek: I did this for consistency with the case of missing config. But I generally agree, I will change both.

crates/sncast/tests/e2e/account/add.rs
```diff
@@ -502,3 +503,29 @@ pub async fn test_private_key_as_int_in_file() {
...
+        "my_account_add",
+        "--address",
+        DEVNET_PREDEPLOYED_ACCOUNT_ADDRESS,
+        "--private-key",
+        "0x456",
+        "--add-profile",
+        "random",
+    ];
+
+    runner(&args).current_dir(tempdir.path()).assert().success();
```
ddoktorski, line 530: I think this line doesn't necessarily confirm that the command was executed successfully. Could we possibly use `assert_stdout_contains`, or even ensure that the config file includes the new profile?

THenry14 (approved):

Arcticae (approved):

PR #1915, opened 2024-03-19
Check for output()?.status.success() when calling a binary where needed

### LINKED ISSUE #1866: Check for `output()?.status.success()` when calling a binary where needed

snforge, sncast

### Description

Rn e.g. when calling USC we don't check the status, we just do `output()?`. It will result in an error appearing in wrong place and misleading users. We also have some occurrences of it in our tests

Didn't find any occurrences of it in sncast but it's worth checking it to, just look for `output()` in the codebase.

Draggu:
Closes #1866

## Introduced changes

- Check for output()?.status.success() when calling a binary where needed

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4147,6 +4147,7 @@ dependencies = [
  "semver",
  "serde",
  "serde_json",
+ "shared",
  "thiserror",
  "universal-sierra-compiler-api",
  "which",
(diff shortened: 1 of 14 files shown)


piotmag769 (approved):

Arcticae (approved):
LGTM

PR #1924, opened 2024-03-20
Test OS costs in resources

Arcticae:
Follow up for #1919

### DIFF
--- a/crates/forge/tests/integration/resources.rs
+++ b/crates/forge/tests/integration/resources.rs
@@ -193,3 +193,42 @@ fn accumulate_syscalls() {
     assert_syscall(&result, "single_write", StorageWrite, 1);
     assert_syscall(&result, "double_write", StorageWrite, 2);
 }
+
+#[test]
+fn steps_count_includes_os_resources() {
+    let test = test_case!(indoc!(
+        "
+            use starknet::{SyscallResultTrait, StorageAddress};
+
+            #[test]
+            fn syscall_storage_write() {
+                let storage_address: StorageAddress = 10.try_into().unwrap();
+                starknet::storage_write_syscall(0, storage_address, 10).unwrap_syscall();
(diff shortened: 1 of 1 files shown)


crates/forge/tests/integration/resources.rs
```diff
@@ -193,3 +193,42 @@ fn accumulate_syscalls() {
...
+                starknet::storage_write_syscall(0, storage_address, 10).unwrap_syscall();
+                assert(1 == 1, 'haha');
+            }
+
+            #[test]
+            fn syscall_storage_write_baseline() {
+                let _storage_address: StorageAddress = 10.try_into().unwrap();
+                // starknet::storage_write_syscall(0, storage_address, 10).unwrap_syscall();
+                // starknet::storage_write_syscall(0, storage_address, 10).unwrap_syscall();
+                // starknet::storage_write_syscall(0, storage_address, 10).unwrap_syscall();
```
drknzz, line 217: Does this need to be here?

crates/forge/tests/integration/resources.rs
```diff
@@ -193,3 +193,42 @@ fn accumulate_syscalls() {
     assert_syscall(&result, "single_write", StorageWrite, 1);
     assert_syscall(&result, "double_write", StorageWrite, 2);
 }
+
+#[test]
+fn steps_count_includes_os_resources() {
```
drknzz, line 198: I guess then this test doesn't have anything to do with steps?

crates/forge/tests/integration/resources.rs
Arcticae: i just wanted to underline that this is exactly 1:1 the same with the syscalls not present

crates/forge/tests/integration/resources.rs
Arcticae: right, corrected

drknzz (approved):

PR #1928, opened 2024-03-21
Add info about possible reasons of `cheatcode xyz is not supported in this runtime`

### LINKED ISSUE #1818: Add info about possible reasons of `cheatcode xyz is not supported in this runtime`

snforge, sncast

### Description

Mainly snforge task, but can be done in sncast too
In snforge the reasons are:
90% - using test functions from cairo-test (`corelib/src/starknet/test.cairo`)
10% - version of `snforge_std` with new cheatcode (that is used by the user), but old `snforge` binary

Draggu:
…this runtime

Closes #1818

## Introduced changes

- Adde info about possible reasons of `cheatcode xyz is not supported in this runtime`

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/runtime/src/lib.rs
```diff
@@ -124,9 +141,18 @@ impl<'a> HintProcessorLogic for StarknetRuntime<'a> {
...
             let selector = parse_selector(selector)?;
-            return Err(HintError::CustomHint(
-                format!("Cheatcode `{selector}` is not supported in this runtime").into(),
-            ));
+
+            let mut base_error =
+                format!("Cheatcode `{selector}` is not supported in this runtime\n").to_string();
+
+            if CAIRO_TEST_CHEATCODES.contains(&selector.as_str()) {
+                base_error.push_str("Check if cheatcodes are imported from snforge_std");
```
THenry14, line 149: runtime should not refer to snforge_std directly, as sncast uses it as well

crates/runtime/src/lib.rs
```diff
@@ -124,9 +141,18 @@ impl<'a> HintProcessorLogic for StarknetRuntime<'a> {
         })) = maybe_extended_hint
         {
             let selector = parse_selector(selector)?;
-            return Err(HintError::CustomHint(
-                format!("Cheatcode `{selector}` is not supported in this runtime").into(),
-            ));
+
+            let mut base_error =
+                format!("Cheatcode `{selector}` is not supported in this runtime\n").to_string();
```
THenry14, line 146: `Cheatcode` is rather unfortunate name in scope of sncast, as we do not have cheatcodes but functions. Let's not use this jargon in messages directed to users in runtime if possible

THenry14 (commented):
we might also use an sncast test checking this

crates/runtime/src/lib.rs
Arcticae: I disagree, the libfunc is even called `starknet::testing::cheatcode` so i don't think the naming is off

crates/forge/tests/integration/runtime.rs
```diff
@@ -24,7 +24,39 @@ fn missing_cheatcode_error() {
     assert_case_output_contains(
         &result,
         "missing_cheatcode_error",
-        "Cheatcode `not_existing123` is not supported in this runtime",
+        indoc!(
+            r"
+            Cheatcode `not_existing123` is not supported in this runtime
+            Check if used snforge_std is compatible with used snforge binary
```
Arcticae, line 30: It might also be the case for things inside `sncast`'s script so it's not the only possible outcome here. We should probably have a broader explanation

Arcticae (changes requested):

piotmag769 (approved):
The idea is good, change the messages pls

crates/runtime/src/lib.rs
Arcticae: Calling it a `function` or something else is less accurate imo but i can live with that

Arcticae (approved):

crates/runtime/src/lib.rs
THenry14: I don't necessary think we should call it a `function` as well, if this message is changed smartly we can avoid both `cheatcode` and `function` I suppose

Draggu: If not `cheatcode` or `function`, any proposition? I don't see any point not to call it `function` (except that `cheatcode` would be more precise). @THenry14

THenry14: why not just `{selector} is not supported in this runtime\n` ?

Draggu: Because our error doesn't tell you the line, only the test case, and without the line, it's cryptic

THenry14: I'm not sure how an additional word "cheatcode" or "function" helps you solve this particular problem tbh. 
But I'm fine with calling this a function then

PR #1939, opened 2024-03-26
Add docstrings to trace info

Arcticae:
## Introduced changes

- Adds docstrings that are displayed later on when using LS

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests

### DIFF
--- a/docs/src/appendix/snforge-library/get_call_trace.md
+++ b/docs/src/appendix/snforge-library/get_call_trace.md
@@ -9,104 +9,17 @@ to [`snforge-std` source](https://github.com/foundry-rs/starknet-foundry/tree/v0
 
 Gets current call trace of the test, up to the last call made to a contract.
 
-## Example call trace
+The whole structure is represented as a tree of calls, in which each contract interaction
+is a new execution scope - thus resulting in a new nested trace.
 
-```cairo
-use snforge_std::trace::{CallTrace, CallEntryPoint, CallType, EntryPointType, get_call_trace};
-...
-
-let ctrace = CallTrace {
-    entry_point: CallEntryPoint {
-        entry_point_type: EntryPointType::External,
-        entry_point_selector: test_selector(),
-        calldata: array![],
-        contract_address: test_address(),
-        caller_address: 0.try_into().unwrap(),
-        call_type: CallType::Call,
-    },
-    nested_calls: array![
-        CallTrace {
-            entry_point: CallEntryPoint {
-                entry_point_type: EntryPointType::Constructor,
-                entry_point_selector: selector!("constructor"),
-                calldata: array![346.into()],
-                contract_address: 123.try_into().unwrap(),
-                caller_address: test_address(),
-                call_type: CallType::Call,
-            },
-            nested_calls: array![
-                CallTrace {
-                    entry_point: CallEntryPoint {
-                        entry_point_type: EntryPointType::External,
-                        entry_point_selector: selector!("other_entrypoint"),
-                        calldata: array![1],
-                        contract_address: 567.try_into().unwrap(),
-                        caller_address: 123.try_into().unwrap(),
-                        call_type: CallType::Call,
-                    },
-                    nested_calls: array![],
-                    result: CallResult::Success(array![101])
-                }
-            ],
-            result: CallResult::Success(array![])
-        },
-        CallTrace {
-            entry_point: CallEntryPoint {
-                entry_point_type: EntryPointType::L1Handler,
-                entry_point_selector: selector!("handle_l1_message"),
-                calldata: array![123, 567.into()],
-                contract_address: 346.try_into().unwrap(),
-                caller_address: 0.try_into().unwrap(),
-                call_type: CallType::Call,
-            },
-            nested_calls: array![],
-            result: CallResult::Failure(CallFailure::Panic(array![482670963043]))
-        }
-    ]
-};
-```
-
-The topmost-call is representing the test call, which will always be present if you're running a test.
-It can have nested `CallTrace` - it's an array of subsequent traces made in scope of the call.
-
-The whole structure is represented as a tree of calls, in which each contract interaction (`constructor`
-call, `l1_handler` call, library or a regular contract `call` via dispatcher) is a new execution scope - thus resulting
-in a new nested trace.
+> 📝 **Note**
+>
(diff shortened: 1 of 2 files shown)


Arcticae:
Examples:
image
image

snforge_std/src/trace.cairo
drknzz, line 1: Could the field comments be above each field?

snforge_std/src/trace.cairo
```diff
@@ -2,48 +2,55 @@ use core::result::ResultTrait;
 use core::starknet::testing::cheatcode;
 use core::starknet::ContractAddress;
 
+/// Tree-like structure which contains all of the calls and sub-calls along with the results
 #[derive(Drop, Serde, PartialEq)]
 struct CallTrace {
-    entry_point: CallEntryPoint,
-    nested_calls: Array<CallTrace>,
+    entry_point: CallEntryPoint, /// The topmost entry point of this trace
```
piotmag769, line 8: Wouldn't it be better to put it above the field definition?

snforge_std/src/trace.cairo
```diff
@@ -2,48 +2,55 @@ use core::result::ResultTrait;
 use core::starknet::testing::cheatcode;
 use core::starknet::ContractAddress;
 
+/// Tree-like structure which contains all of the calls and sub-calls along with the results
```
piotmag769, line 5:
```suggestion
/// Tree-like structure which contains all of starknet calls and sub-calls that happened in a test
```

snforge_std/src/trace.cairo
piotmag769, line 8:
```suggestion
    /// A summary of a call to a contract entry point
    entry_point: CallEntryPoint,
```

snforge_std/src/trace.cairo
```diff
@@ -2,48 +2,55 @@ use core::result::ResultTrait;
...
 #[derive(Drop, Serde, PartialEq)]
 struct CallTrace {
-    entry_point: CallEntryPoint,
-    nested_calls: Array<CallTrace>,
+    entry_point: CallEntryPoint, /// The topmost entry point of this trace
+    nested_calls: Array<CallTrace>, /// All the calls that happened in the context of `entry_point`
```
piotmag769, line 9:
```suggestion
    /// The inner calls that happened as a result of calling the `entry_point`
    nested_calls: Array<CallTrace>,
```

snforge_std/src/trace.cairo
```diff
@@ -2,48 +2,55 @@ use core::result::ResultTrait;
...
     result: CallResult
 }
 
+/// A single function entry point summary
```
piotmag769, line 13:
```suggestion
/// A summary of call to a contract entry point
```

docs/src/appendix/snforge-library/get_call_trace.md
```diff
@@ -9,104 +9,17 @@ to [`snforge-std` source](https://github.com/foundry-rs/starknet-foundry/tree/v0
 
 Gets current call trace of the test, up to the last call made to a contract.
 
-## Example call trace
```
piotmag769, line 12: Do we want to move it to the docs?

Arcticae: Not really, i don't think it's that useful and it's not really maintainable as well.

snforge_std/src/trace.cairo
Arcticae: I don't think there's strict causality there. There could be other calls that happened because of this, but not in the scope of this call, so i think the previous wording is more accurate (i could change "context" to "scope" perhaps?)

snforge_std/src/trace.cairo
Arcticae: I don't see a point in copy-pasting the struct doc in the doc to the field, it doesn't bring any extra information to the table. I'd rather just remove it instead

snforge_std/src/trace.cairo
Arcticae: What if it's a library call, it's not a strictly call to a contract per-se

snforge_std/src/trace.cairo
Arcticae: I don't like this definition because it's not emphasising the recursiveness of the struct, it appears then like it's top-level only. Also, why remove the info about the results?

snforge_std/src/trace.cairo
Arcticae: done

snforge_std/src/trace.cairo
Arcticae: done

snforge_std/src/trace.cairo
piotmag769: My only concern is that we should emphasise that it is a starknet call, not a regular function call

snforge_std/src/trace.cairo
piotmag769: Sound good

snforge_std/src/trace.cairo
piotmag769: call  \in {contract call, library call} imo, but if you think ur version is better feel free to resolve

snforge_std/src/trace.cairo
piotmag769: Yea, scope is better imo

snforge_std/src/trace.cairo
```diff
@@ -2,48 +2,69 @@ use core::result::ResultTrait;
...
+    /// Regular call
     Call,
+    /// Library call
     Delegate,
 }
 
+/// Result of a contract or a library call
 #[derive(Drop, Serde, PartialEq)]
 enum CallResult {
+    /// A successful call with it's result
```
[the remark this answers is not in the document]
Arcticae: It's description of an enum variant as well so it's not accurate

snforge_std/src/trace.cairo
```diff
@@ -2,48 +2,69 @@ use core::result::ResultTrait;
...
     Success: Array<felt252>,
-    Failure: CallFailure
+    /// A failed call along with it's panic data
```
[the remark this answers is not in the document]
Arcticae: Same as above

drknzz (approved):
🇨🇳

PR #1940, opened 2024-03-26
Handle OS errors gracefully

### LINKED ISSUE #1795: Handle OS errors gracefully

snforge

### Description

We often panic when writing/reading a file/creating dir/other OS related operation fail. We should bubble those errors instead (they will be pretty printed in main)

Draggu:
Closes #1795

## Introduced changes

- Handle OS errors gracefully

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


Arcticae (approved):

crates/cheatnet/src/forking/state.rs
```diff
@@ -40,14 +41,13 @@ pub struct ForkStateReader {
...
-    #[must_use]
-    pub fn new(url: Url, block_number: BlockNumber, cache_dir: &str) -> Self {
-        ForkStateReader {
-            cache: ForkCache::load_or_new(&url, block_number, cache_dir),
+    pub fn new(url: Url, block_number: BlockNumber, cache_dir: &str) -> Result<Self> {
+        Ok(ForkStateReader {
+            cache: ForkCache::load_or_new(&url, block_number, cache_dir)?,
             client: JsonRpcClient::new(HttpTransport::new(url)),
             block_number,
             runtime: Runtime::new().expect("Could not instantiate Runtime"),
```
drknzz, line 49: context?

crates/forge-runner/src/running.rs
```diff
@@ -379,6 +379,7 @@ fn get_fork_state_reader(
                 *block_number,
                 workspace_root.join(CACHE_DIR).as_ref(),
             )
+            .unwrap()
```
drknzz, line 382: It's ok right? (just to double check)

drknzz (approved):
👍

crates/cheatnet/src/forking/state.rs
MaksymilianDemitraszek: Can this fail in any way by user input?

crates/forge-runner/src/running.rs
Draggu: No, it's not. Good catch

crates/cheatnet/src/forking/state.rs
Draggu: Done

Draggu:
Panics in `ForkCache::save` are not changed due to the fact it's called on drop and there is no way to propagate errors

Arcticae:
> Panics in `ForkCache::save` are not changed due to the fact it's called on drop and there is no way to propagate errors

We can change the logic there if it's required for errors to bubble. It's not necessary to have it on `drop`

Draggu:
> > Panics in `ForkCache::save` are not changed due to the fact it's called on drop and there is no way to propagate errors
> 
> We can change the logic there if it's required for errors to bubble. It's not necessary to have it on `drop`

But then it's easy to forget calling `save()`. Either propagate or check if saved by compiler

Issue #1078, opened 2023-11-08
Increase max-fee in deploy account docs


drknzz:
### Which component is your feature related to?

Cast

### Feature Request

There is `--max-fee 864600000000` specified in docs when deploying an account, but sending a tx with such results in
```
error: Max fee is smaller than the minimal transaction cost (validation plus fee transfer)
```

[Here](https://foundry-rs.github.io/starknet-foundry/starknet/account.html?highlight=864600000000#general-example) and probably in some other places.

Vap0RSh4wN:
hi, may I ask how can I solve this problem? o(╥﹏╥)o
thanks!

THenry14:
> hi, may I ask how can I solve this problem? o(╥﹏╥)o thanks!

Hi, this issue is aliready assigned, sorry

PR #1946, opened 2024-03-28
[docs] Adjust max fee, fix wrong short flag in invoke

THenry14:
Closes https://github.com/foundry-rs/starknet-foundry/issues/1078

## Introduced changes

- changes max fee suggested in docs
- fixes wrong short flag in invoke subcommand

## Checklist

- [X] Linked relevant issue
- [X] Added relevant tests

### DIFF
--- a/docs/src/starknet/account.md
+++ b/docs/src/starknet/account.md
@@ -48,7 +48,7 @@ Do the following to start interacting with the Starknet:
       --url http://127.0.0.1:5050 \
       account deploy
       --name some-name \
-      --max-fee 864600000000
+      --max-fee 9999999999999
     
     command: account deploy
     transaction_hash: 0x20b20896ce63371ef015d66b4dd89bf18c5510a840b4a85a43a983caa6e2579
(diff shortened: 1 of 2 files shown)


docs/src/starknet/account.md
```diff
@@ -48,7 +48,7 @@ Do the following to start interacting with the Starknet:
       --url http://127.0.0.1:5050 \
       account deploy
       --name some-name \
-      --max-fee 864600000000
+      --max-fee 9999999999999
```
ddoktorski, line 51: Perhaps we could also adjust the output above to match these numbers? Apart from that I think the example below with salt should be updated as well.

THenry14: Thought about it, but these are outputs (not inputs), so I'm rather reluctant. Users will see something different either way

ddoktorski: That's fine. Regarding the second part I meant updating `--max-fee` under [Custom Account Contract](https://foundry-rs.github.io/starknet-foundry/starknet/account.html#custom-account-contract) example.

THenry14: in this case, i think it may make sense to keep it, as the input is directly taken from the output of a previous command; I think it may be confusing if we discredit the output of the first command completely

ddoktorski (approved):

PR #1953, opened 2024-03-29
Reuse calculated class hashes from `ContractsData` in `declare`

### LINKED ISSUE #1719: Reuse calculated class hashes from `ContractsData` in `declare`

snforge

### Description

We don't need to calculate the class hashes in `declare` anymore since they are precalculated after #1718

Draggu:
Closes #1719

## Introduced changes

- Reuse calculated class hashes from `ContractsData` in `declare`

### DIFF
--- a/crates/forge-runner/src/running.rs
+++ b/crates/forge-runner/src/running.rs
@@ -239,6 +239,7 @@ pub fn run_test_case(
     let forge_extension = ForgeExtension {
         environment_variables: &runner_params.environment_variables,
         contracts: &runner_params.contracts_data.contracts,
+        class_hashes: &runner_params.contracts_data.class_hashes,
     };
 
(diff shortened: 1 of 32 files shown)


crates/forge-runner/src/running.rs
```diff
@@ -239,6 +239,7 @@ pub fn run_test_case(
     let forge_extension = ForgeExtension {
         environment_variables: &runner_params.environment_variables,
         contracts: &runner_params.contracts_data.contracts,
+        class_hashes: &runner_params.contracts_data.class_hashes,
```
piotmag769, line 242: [nit] Better to pass the entire struct, no? Not sure though

piotmag769 (approved):

piotmag769 (approved):

Arcticae (approved):

PR #1965, opened 2024-04-02
Add phantom nodes on deploy with no constructor

piotmag769:
For purposes of cairo-profiler

## Introduced changes

- self-explanatory

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests

### DIFF
--- a/crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/rpc.rs
+++ b/crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/rpc.rs
@@ -39,6 +39,12 @@ pub enum CallResult {
     Failure(CallFailure),
 }
 
+impl Default for CallResult {
+    fn default() -> Self {
+        Self::Success { ret_data: vec![] }
+    }
+}
+
 /// Enum representing possible call failure and its' type.
 /// `Panic` - Recoverable, meant to be caught by the user.
 /// `Error` - Unrecoverable, equivalent of panic! in rust.
--- a/crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/rpc.rs
+++ b/crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/rpc.rs
@@ -41,7 +41,7 @@ pub enum CallResult {
 
 impl Default for CallResult {
     fn default() -> Self {
-        Self::Success { ret_data: vec![] }
+        Self::Failure(CallFailure::Error { msg: String::new() })
     }
 }
 
--- a/crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/rpc.rs
+++ b/crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/rpc.rs
@@ -45,7 +45,7 @@ impl Default for CallResult {
     }
 }
 
-/// Enum representing possible call failure and its' type.
+/// Enum representing possible call failure and its type.
 /// `Panic` - Recoverable, meant to be caught by the user.
 /// `Error` - Unrecoverable, equivalent of panic! in rust.
 #[derive(Debug, Clone, Serialize, Deserialize)]
--- a/crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/rpc.rs
+++ b/crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/rpc.rs
@@ -39,12 +39,6 @@ pub enum CallResult {
     Failure(CallFailure),
 }
 
-impl Default for CallResult {
-    fn default() -> Self {
-        Self::Failure(CallFailure::Error { msg: String::new() })
-    }
-}
-
 /// Enum representing possible call failure and its type.
 /// `Panic` - Recoverable, meant to be caught by the user.
 /// `Error` - Unrecoverable, equivalent of panic! in rust.
--- a/crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
+++ b/crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
@@ -690,13 +690,7 @@ fn serialize_call_trace(call_trace: &CallTrace, output: &mut Vec<Felt252>) {
     let visible_calls: Vec<_> = call_trace
         .nested_calls
         .iter()
-        .filter_map(|call_node| {
-            if let CallTraceNode::EntryPointCall(call) = call_node {
-                Some(call)
-            } else {
-                None
-            }
-        })
+        .filter_map(CallTraceNode::extract_entry_point_call)
         .collect();
 
     output.push(Felt252::from(visible_calls.len()));
@@ -833,13 +827,7 @@ pub fn update_top_call_execution_resources(runtime: &mut ForgeRuntime) {
     let nested_calls_syscalls = top_call
         .nested_calls
         .iter()
-        .filter_map(|trace_node| {
-            if let CallTraceNode::EntryPointCall(trace) = trace_node {
-                Some(trace)
-            } else {
-                None
-            }
-        })
+        .filter_map(CallTraceNode::extract_entry_point_call)
         .fold(SyscallCounter::new(), |syscalls, trace| {
             sum_syscall_counters(syscalls, &trace.borrow().used_syscalls)
         });
--- a/crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
+++ b/crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
@@ -684,6 +684,10 @@ fn handle_deploy_result(
 
 // append all to one output Vec instead of allocating new one for each nested call
 fn serialize_call_trace(call_trace: &CallTrace, output: &mut Vec<Felt252>) {
+    if !call_trace.node_type.is_visible_to_user() {
+        return;
+    }
+
     serialize_call_entry_point(&call_trace.entry_point, output);
 
     output.push(Felt252::from(call_trace.nested_calls.len()));
--- a/crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
+++ b/crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
@@ -1,5 +1,5 @@
 use self::contracts_data::ContractsData;
-use crate::state::NodeType;
+use crate::state::CallTraceNode;
 use crate::{
     runtime_extensions::{
         call_to_blockifier_runtime_extension::{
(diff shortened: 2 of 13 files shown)


crates/forge/tests/e2e/trace_resources.rs
piotmag769, line 1: All changes in this file are due to phantom nodes added to trace, so indexes had to change

crates/cheatnet/src/state.rs
```diff
@@ -158,6 +159,24 @@ pub struct CallTrace {
...
+#[derive(Default, Copy, Clone)]
+pub enum NodeType {
+    #[default]
+    Regular,
+    Phantom,
+}
+
+impl NodeType {
+    #[must_use]
+    pub fn is_visible_to_user(&self) -> bool {
```
MaksymilianDemitraszek, line 174: Not sure what this function does, but it seems like mixing presentation layer with data definition layer.

crates/cheatnet/src/state.rs
```diff
@@ -158,6 +159,24 @@ pub struct CallTrace {
     pub nested_calls: Vec<Rc<RefCell<CallTrace>>>,
     pub result: CallResult,
     pub vm_trace: Option<Vec<TraceEntry>>,
+    pub node_type: NodeType,
```
MaksymilianDemitraszek, line 162: It is not the best solution, you do not encode on the type level the information that Phantom node has no interest in the any CallTrace fields. Please make CallTrace an enum instead.

crates/cheatnet/src/state.rs
```diff
@@ -158,6 +159,24 @@ pub struct CallTrace {
...
+}
+
+#[derive(Default, Copy, Clone)]
+pub enum NodeType {
+    #[default]
+    Regular,
+    Phantom,
+}
```
MaksymilianDemitraszek, line 170: Probably should be added to the profiler definition, with some comments why we need those nodes to be present.

crates/cheatnet/src/state.rs
```diff
@@ -446,6 +455,16 @@ impl TraceData {
         last_call.result = result;
         last_call.vm_trace = vm_trace;
     }
+
+    pub fn add_phantom_call(&mut self) {
```
MaksymilianDemitraszek, line 459: Not sure if using `phantom` name is a good idea here:
- `phantom` already has meaning in rust
- we use those nodes only for deploys without constructors, it is better to use a specific name than abstract

MaksymilianDemitraszek (changes requested):

crates/cheatnet/src/state.rs
Arcticae: Definetely shouldn't be in cheatnet

crates/cheatnet/src/state.rs
Arcticae: It will be easier to implement and interpret probably as well

crates/cheatnet/src/state.rs
Arcticae: `add_empty_call` would also work i guess

aren't there any future uses for this? btw - isn't this kind of design flaw, where we should always treat the deploy as a special case, no matter if it has the constructor or not? it'd make for more consistent code i think (and less hacky)

crates/forge/tests/e2e/trace_resources.rs
Arcticae, line 1: that is just horrendous XD

crates/cheatnet/src/state.rs
piotmag769: It is added to the profiler definition

crates/forge/tests/e2e/trace_resources.rs
piotmag769: Unlucky, I cannot do anything about it xD

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/rpc.rs
```diff
@@ -39,6 +39,12 @@ pub enum CallResult {
     Failure(CallFailure),
 }
 
+impl Default for CallResult {
+    fn default() -> Self {
+        Self::Success { ret_data: vec![] }
+    }
```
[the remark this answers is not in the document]
piotmag769: Can be a failure, no difference for me

crates/cheatnet/src/state.rs
piotmag769: >aren't there any future uses for this?

Not really

>isn't this kind of design flaw, where we should always treat the deploy as a special case, no matter if it has the constructor or not

No, we shouldn't we should only treat it as a special case in case of an empty constructor. The non-empty constructor goes through `execute_call_entry_point` so it is collected same as a regular call

crates/cheatnet/src/state.rs
piotmag769: True, no idea why I did it this way, will change to enum

crates/cheatnet/src/state.rs
piotmag769: Will be gone with enum introduction

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/rpc.rs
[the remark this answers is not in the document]
piotmag769: Found a bug thanks to this btw

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
```diff
@@ -744,7 +757,7 @@ fn serialize_call_result(call_result: &CallResult, output: &mut Vec<Felt252>) {
                     serialize_failure_data(0, panic_data.iter().cloned(), panic_data.len(), output);
                 }
                 CallFailure::Error { msg } => {
-                    let data = ByteArray::from(msg.as_str()).serialize_with_magic();
+                    let data = ByteArray::from(msg.as_str()).serialize_no_magic();
```
piotmag769, line 760: https://github.com/foundry-rs/starknet-foundry/issues/1990

crates/forge/tests/e2e/build_profile.rs
```diff
@@ -2,6 +2,7 @@ use super::common::runner::{setup_package, test_runner};
 use forge_runner::profiler_api::PROFILE_DIR;
 
 #[test]
+#[ignore] // TODO(#1991): remove ignore when new profiler is released
```
piotmag769, line 5: #1991

Arcticae (approved):

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/rpc.rs
```diff
@@ -39,7 +39,13 @@ pub enum CallResult {
     Failure(CallFailure),
 }
 
-/// Enum representing possible call failure and its' type.
+impl Default for CallResult {
+    fn default() -> Self {
+        Self::Failure(CallFailure::Error { msg: String::new() })
+    }
+}
```
MaksymilianDemitraszek, line 46: why?

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
```diff
@@ -827,6 +840,13 @@ pub fn update_top_call_execution_resources(runtime: &mut ForgeRuntime) {
     let nested_calls_syscalls = top_call
         .nested_calls
         .iter()
+        .filter_map(|trace_node| {
+            if let CallTraceNode::EntryPointCall(trace) = trace_node {
```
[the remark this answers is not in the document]
MaksymilianDemitraszek: I agree with @Arcticae here, function name is easier to parse than some random if.

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
```diff
@@ -686,10 +687,22 @@ fn handle_deploy_result(
...
+        .nested_calls
+        .iter()
+        .filter_map(|call_node| {
+            if let CallTraceNode::EntryPointCall(call) = call_node {
+                Some(call)
+            } else {
+                None
+            }
+        })
+        .collect();
```
MaksymilianDemitraszek, line 700: I don't understand what `visible` means in this context.

MaksymilianDemitraszek (changes requested):

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/rpc.rs
piotmag769: To avoid boilerplate code when creating `CallTrace`, I can write `..Default::default()` now

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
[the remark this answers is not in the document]
piotmag769: Can change

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
piotmag769: Visible to the user, we don't want to put nodes that has no call (deploy without constructor) to cairo

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/rpc.rs
MaksymilianDemitraszek: Imo this hides some important information.

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
piotmag769: Disagree, it shouldn't appear in cairo at all, I would have to expose this type in snforge_std then

MaksymilianDemitraszek: and why this is a problem? this way you hide this detail deep in the internals

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/rpc.rs
piotmag769: You would literally never leave it default though, I can revert this if you want to

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
piotmag769: I can create an issue for that, but it's out of scope of this PR, I don't want to waste time on this

piotmag769: https://github.com/foundry-rs/starknet-foundry/issues/2000

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/rpc.rs
MaksymilianDemitraszek: discussed offline `build_failed` function should be used instead.

MaksymilianDemitraszek (approved):

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/rpc.rs
piotmag769: Used `default_successful_call` instead

PR #1969, opened 2024-04-03
Use `setup-universal-sierra-compiler` action

### LINKED ISSUE #1858: Add and integrate an action for USC for `setup-snfoundry`

snforge

### Description

Please try using composite actions in order to achieve that

When closing - please unblock https://github.com/foundry-rs/starknet-foundry/issues/1834

tomek0123456789:
Closes #1858

## Introduced changes

- Added [`setup-universal-sierra-compiler`](https://github.com/software-mansion/setup-universal-sierra-compiler) action

Couldn't integrate it using composite actions, sadly.

## Checklist

- [ ] Linked relevant issue
- [ ] Added relevant tests

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -201,7 +201,7 @@ jobs:
           BINARY_PATH="${BINARY_PATH%.tar.gz}"
           BINARY_PATH="${BINARY_PATH%.zip}"
           BINARY_PATH="${BINARY_PATH#artifacts/}"
-          export UNIVERSAL_SIERRA_COMPILER="${{ env.UDC_PATH }}"
+          echo $UNIVERSAL_SIERRA_COMPILER
           
           if [[ ${{ matrix.target }} == *windows* ]]; then
             SNFORGE_PATH=$(readlink -f $BINARY_PATH/bin/snforge.exe)
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -190,16 +190,7 @@ jobs:
           fi
 
       - name: Install universal-sierra-compiler
-        shell: bash
-        run: |
-          if [[ ${{ matrix.target }} == *windows* ]]; then
-            curl -L -o udc.zip https://github.com/software-mansion/universal-sierra-compiler/releases/download/v2.0.0/universal-sierra-compiler-v2.0.0-x86_64-pc-windows-msvc.zip
-            unzip udc.zip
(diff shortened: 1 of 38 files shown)


piotmag769 (approved):

.github/workflows/release.yml
```diff
@@ -190,16 +190,7 @@ jobs:
           fi
 
       - name: Install universal-sierra-compiler
-        shell: bash
```
THenry14, line 193: From what I've seen, a support for windows has been added to the gh action - that is cool. Do you happen to know if there any plans to also include this support to install.sh script?

.github/workflows/release.yml
```diff
@@ -190,16 +190,7 @@ jobs:
...
-            unzip udc.zip
-            echo "UDC_PATH=$(pwd)/universal-sierra-compiler-v2.0.0-x86_64-pc-windows-msvc/bin/universal-sierra-compiler.exe" >> $GITHUB_ENV
-          else
-            curl -L https://raw.githubusercontent.com/software-mansion/universal-sierra-compiler/master/scripts/install.sh | sh
-            echo "UDC_PATH=${HOME}/.local/bin/universal-sierra-compiler" >> $GITHUB_ENV
-          fi
+        uses: software-mansion/setup-universal-sierra-compiler@v1
 
       - name: Smoke test
         shell: bash
```
THenry14, line 196: I can't comment on the lines below, but there we are using `UDC_PATH`, that is used in this Smoke tests job step - I think it should be removed, otherwise it may break smoke tests (UDC_PATH overwrites UNIVERSAL_SIERRA_COMPILER, and after this is merged I think it will set it to an empty string. This variable is being used by snforge/sncast if I recall correctly, so there will be no path to usc)

THenry14 (changes requested):

.github/workflows/release.yml
tomek0123456789: Sure, I could do that later

.github/workflows/release.yml
tomek0123456789: Done!

Arcticae (approved):

THenry14 (approved):

tomek0123456789:
Will merge after making sure that release action works as intended, right now it doesn't success on my fork.

THenry14:
> Will merge after making sure that release action works as intended, right now it doesn't success on my fork.

There may be issues when running on a fork. I suppose you could comment out the logic that deploys docs + creates a release, and set the triggering branch to your own. That way, the release flow (building, testing) will be properly tested on our env, just without the last step (release itself). Makes sense?

tomek0123456789:
Release check on fork ran here: https://github.com/tomek0123456789/starknet-foundry/actions/runs/8554397339/job/23481990666

PR #1993, opened 2024-04-08
Fix init script

ddoktorski:
## Introduced changes

- Fix init script that hasn't been updated after introducing error handling in scripts
- Fix test that checks whether the init script compiles -> overwrite Scarb.toml to use the master branch instead of the latest tag

## Checklist

- [ ] Linked relevant issue
- [X] Added relevant tests

### DIFF
--- a/crates/sncast/tests/e2e/script/init.rs
+++ b/crates/sncast/tests/e2e/script/init.rs
@@ -1,7 +1,8 @@
 use crate::helpers::runner::runner;
 use camino::Utf8PathBuf;
 use indoc::{formatdoc, indoc};
-use shared::test_utils::output_assert::assert_stderr_contains;
+use scarb_api::ScarbCommand;
+use shared::test_utils::output_assert::{assert_stderr_contains, assert_stdout_contains};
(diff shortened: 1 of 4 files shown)


crates/sncast/tests/e2e/script/init.rs
```diff
@@ -116,26 +117,44 @@ fn test_init_twice_fails() {
...
+        .parse::<toml_edit::DocumentMut>()
+        .expect("Failed to parse Scarb.toml");
+
+    let tab = doc["dependencies"]["sncast_std"]
+        .as_inline_table_mut()
+        .unwrap();
+    // Using a tag during the release process will cause the test to fail
+    // as the new tag won't exists in the repository yet
+    tab.remove("tag").expect("Failed to remove the tag");
+    tab.insert("branch", "master".into());
```
THenry14, line 151: could we just modify Scarb.toml directly to use branch instead?

ddoktorski: It seems that we can simply use `scarb add` to overwrite `sncast_std` to use the master branch instead of a tag.

PR #2003, opened 2024-04-11
Add functions to retrieve `SEPOLIA_RPC_URL`

tomek0123456789:
commit-id:918e569e

Fix cargo lint errors

Change implementation of rpc_url()

rpc_url() -> node_url() (to rebase)

Apply review suggestions (to rebase)

Envs from dotenv (to rebase)

Add .env.template file, rename functions once again (to rebase)

---

**Stack**:
- #2006
- #2004
- #2003 ⬅

### DIFF
--- a/crates/shared/src/consts.rs
+++ b/crates/shared/src/consts.rs
@@ -1,1 +1,2 @@
 pub const EXPECTED_RPC_VERSION: &str = "0.7.0";
+pub const RPC_URL_VERSION: &str = "v0_7";
--- a/crates/shared/src/test_utils/mod.rs
+++ b/crates/shared/src/test_utils/mod.rs
@@ -1,1 +1,2 @@
+pub mod node_url;
 pub mod output_assert;
--- a/crates/shared/src/test_utils/node_url.rs
+++ b/crates/shared/src/test_utils/node_url.rs
@@ -1,0 +1,29 @@
+use anyhow::{Context, Result};
+use std::env;
+use url::Url;
+
+/// Loads node RPC URL from `SEPOLIA_RPC_URL` environmental variable set in dotenv.
+///
+/// #### Note:
+/// - `node_rpc_url()` -> <https://example.com/rpc/v0_7>
+/// - `node_url()` -> <https://example.com/>
+pub fn node_rpc_url() -> Result<Url> {
+    dotenv::dotenv().ok();
+    let node_rpc_url = env::var("SEPOLIA_RPC_URL")
+        .context("The required environmental variable `SEPOLIA_RPC_URL` is not set. Please set it manually or in .env file"
+    )?;
+
+    Url::parse(&node_rpc_url).with_context(|| {
+        format!("Failed to parse the URL from the `SEPOLIA_RPC_URL` environmental variable: {node_rpc_url}")
+    })
+}
+
+/// Loads node URL from `SEPOLIA_RPC_URL` environmental variable and parses it,
+/// returning URL with no slug (`rpc/v0_7` suffix).
+pub fn node_url() -> Result<Url> {
+    let mut node_url = node_rpc_url()?;
+    node_url.set_path("");
+    node_url.set_query(None);
+
+    Ok(node_url)
+}
--- a/.env.template
+++ b/.env.template
@@ -1,0 +1,1 @@
+SEPOLIA_RPC_URL=https://example.com/rpc/v0_7
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ Scarb.lock
 .spr.yml
 .snfoundry_cache/
 .snfoundry_trace/
+.env
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4376,6 +4376,7 @@ dependencies = [
  "cairo-felt",
  "cairo-lang-runner",
  "console",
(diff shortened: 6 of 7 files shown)


crates/shared/src/test_utils/mod.rs
```diff
@@ -1 +1,24 @@
...
+        .with_context(|| "The required environmental variable `NODE_URL` is not set. Please set it manually or in .cargo/config.toml file"
+    )?;
+
+    Url::parse(&node_url).with_context(|| {
+        format!("Failed to parse the URL from the `NODE_URL` environmental variable: {node_url}")
+    })
+}
+
+pub fn node_url_with_version() -> Result<Url> {
+    let mut node_url = node_url().unwrap();
```
Draggu, line 19:
```suggestion
    let mut node_url = node_url()?;
```

crates/shared/src/test_utils/mod.rs
```diff
@@ -1 +1,24 @@
+use crate::consts::RPC_URL_VERSION;
+use anyhow::{Context, Result};
+use std::env;
+use url::Url;
+
 pub mod output_assert;
+
+pub fn node_url() -> Result<Url> {
+    let node_url = env::var("NODE_URL")
+        .with_context(|| "The required environmental variable `NODE_URL` is not set. Please set it manually or in .cargo/config.toml file"
```
Draggu, line 10:
```suggestion
        .context("The required environmental variable `NODE_URL` is not set. Please set it manually or in .cargo/config.toml file"
```

Draggu (approved):

crates/shared/src/test_utils/mod.rs
```diff
@@ -1 +1,24 @@
+use crate::consts::RPC_URL_VERSION;
+use anyhow::{Context, Result};
+use std::env;
+use url::Url;
+
 pub mod output_assert;
+
+pub fn node_url() -> Result<Url> {
+    let node_url = env::var("NODE_URL")
```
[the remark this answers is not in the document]
tomek0123456789: Used only `.cargo/config.toml`file for configuration, would `.env` be really necessary since we can do it that way? cc others

crates/shared/src/test_utils/node_url.rs
```diff
@@ -0,0 +1,22 @@
...
+    })
+}
+
+pub fn node_url_with_version() -> anyhow::Result<Url> {
+    let mut node_url = node_url()?;
+    node_url.set_path(format!("rpc/{RPC_URL_VERSION}").as_str());
+    node_url.set_query(None);
+
+    Ok(node_url)
+}
```
piotmag769, line 22: Does it leave the port unchanged?

crates/shared/src/test_utils/mod.rs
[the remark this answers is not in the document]
Arcticae: nah no need i think

Arcticae (approved):

piotmag769 (approved):

crates/shared/src/test_utils/mod.rs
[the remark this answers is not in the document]
THenry14: there is no need, but I really think that if we go this way we must make sure the url is not accidentally commited to the repo at any point. using .env instead is an easy way to make sure that never happens - just a thought

crates/shared/src/consts.rs
```diff
@@ -1 +1,2 @@
 pub const EXPECTED_RPC_VERSION: &str = "0.7.0";
+pub const RPC_URL_VERSION: &str = "v0_7";
```
THenry14, line 2: this may be a nitpick, but I suppose this will prevent users from using older rpcs for tests? are we 100% sure we want to go that way?

tomek0123456789: Don't we update the tests to use the latest possible version in foundry? Like [here](https://github.com/foundry-rs/starknet-foundry/commit/b458b753e9e6e2f69703da0aafc917230c2b237a#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR20)

THenry14: yup, newer versions of snfoundry suoport newer versions of rpc, but I don't think they block the older ones. So technically it should be possible to use newest snforge (say with rpc 0.7 support) and run it against rpc 0.6? I assume that, becase you raise a warning about rpc version mismatch, and not an error. or is that not the case at all?

tomek0123456789: Yes, it emits a warning when an older version is used. I don't want to add another env variable for setting rpc version, though.

THenry14: An alternative would be to just have a NODE_URL variable, that holds complete url right? you wouldn't have to set specific version in code, you could just change it in .env file/in workflow file in CI. But it all probably depends on your (snforge) needs etc. @Arcticae any thoughts on that?

Arcticae: Isn't this only for tests anyway? I guess if we want to run the tests against multiple ones we'd have to parametrise our tests somehow but i think it's out of scope here

THenry14: ok, let's leave this as is for now then :+1:

THenry14 (approved):

PR #2004, opened 2024-04-11
Update CI and tests with `SEPOLIA_RPC_URL` getter function

tomek0123456789:
commit-id:99e99636

Apply review suggestions to tests (to rebase)

Update release workflow

Replacing placeholders in a different test function (to rebase)

Remove env var from forks (to rebase)

Update tests once again

fix invalid ci

Tests, CI updates

---

**Stack**:
- #2006
- #2004 ⬅

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,6 +8,9 @@ on:
       - master
   workflow_dispatch:
 
+env:
+  SEPOLIA_RPC_URL: ${{ secrets.NODE_URL }}:7070/rpc/v0_7
+
 jobs:
   test-forge-unit-and-integration:
     name: Test Forge / Unit and Integration Tests
@@ -20,7 +23,8 @@ jobs:
       - uses: software-mansion/setup-universal-sierra-compiler@v1
       - run: cargo test --release --lib -p forge
       - run: cargo test --release --bin snforge
-      - run: cargo test --release integration -p forge
+      - name: Run Forge unit and integration tests
+        run: cargo test --release integration -p forge
 
   test-forge-e2e:
     name: Test Forge / E2E Tests
@@ -57,7 +61,8 @@ jobs:
       - uses: Swatinem/rust-cache@23bce251a8cd2ffc3c1075eaa2367cf899916d84
       - uses: software-mansion/setup-scarb@v1.3.2
       - uses: software-mansion/setup-universal-sierra-compiler@v1
-      - run: cargo test --release e2e -p forge
+      - name: Run Forge E2E tests
+        run: cargo test --release e2e -p forge
 
   test-forge-runner:
     name: Test Forge Runner
@@ -93,11 +98,8 @@ jobs:
         run: ./scripts/install_devnet.sh
       - uses: software-mansion/setup-scarb@v1.3.2
       - uses: software-mansion/setup-universal-sierra-compiler@v1
-      - name: Run tests
-        env:
-          NODE_URL: ${{ secrets.NODE_URL }}
-        run: |
-          SEPOLIA_RPC_URL="${NODE_URL}:7070/rpc/v0_7" cargo test --release -p sncast
+      - name: Run Cast tests
+        run: cargo test --release -p sncast
 
   test-conversions:
     name: Test Conversions
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -195,7 +195,7 @@ jobs:
       - name: Smoke test
         shell: bash
         env:
-          RPC_URL: ${{ secrets.CHEATNET_RPC_URL }}
+          NODE_URL: ${{ secrets.NODE_URL }}
         run: |
           BINARY_PATH="${{ env.BINARY_PATH }}"
           BINARY_PATH="${BINARY_PATH%.tar.gz}"
@@ -213,7 +213,7 @@ jobs:
           REPO_URL=${{ github.repositoryUrl }}
           REVISION=${{ github.sha }}
           
-          ./scripts/smoke_test.sh "$RPC_URL" "$SNFORGE_PATH" "$SNCAST_PATH" "$REPO_URL" "$REVISION"
+          ./scripts/smoke_test.sh "${NODE_URL}:7070/rpc/v0_7" "$SNFORGE_PATH" "$SNCAST_PATH" "$REPO_URL" "$REVISION"
 
   create-release:
     name: Create release
--- a/crates/forge/tests/e2e/common/runner.rs
+++ b/crates/forge/tests/e2e/common/runner.rs
@@ -3,12 +3,15 @@ use assert_fs::TempDir;
 use camino::Utf8PathBuf;
 use indoc::formatdoc;
 use shared::command::CommandExt;
+use shared::test_utils::node_url::node_rpc_url;
 use snapbox::cmd::{cargo_bin, Command as SnapboxCommand};
+use std::path::Path;
 use std::process::Command;
 use std::str::FromStr;
 use std::{env, fs};
 use test_utils::tempdir_with_tool_versions;
 use toml_edit::{value, DocumentMut};
+use walkdir::WalkDir;
 
 pub(crate) fn runner(temp_dir: &TempDir) -> SnapboxCommand {
     SnapboxCommand::new(cargo_bin!("snforge"))
@@ -49,11 +52,32 @@ pub(crate) fn setup_package_with_file_patterns(
 
     manifest_path.write_str(&scarb_toml.to_string()).unwrap();
 
+    // TODO (#2074): do that on .cairo.template files only
+    replace_node_rpc_url_placeholders(temp.path());
+
     temp
 }
 
 pub(crate) fn setup_package(package_name: &str) -> TempDir {
     setup_package_with_file_patterns(package_name, BASE_FILE_PATTERNS)
+}
+
+fn replace_node_rpc_url_placeholders(dir_path: &Path) {
+    let url = node_rpc_url().unwrap();
+    let temp_dir_files = WalkDir::new(dir_path);
+    for entry in temp_dir_files {
+        let entry = entry.unwrap();
+
+        let path = entry.path();
+
+        if path.is_file() {
+            let content = fs::read_to_string(path).unwrap();
+
+            let modified_content = content.replace("{{ NODE_RPC_URL }}", url.as_str());
+
+            fs::write(path, modified_content).unwrap();
+        }
+    }
 }
 
 pub(crate) fn setup_hello_workspace() -> TempDir {
--- a/crates/forge/tests/e2e/diagnostics_and_plugins.rs
+++ b/crates/forge/tests/e2e/diagnostics_and_plugins.rs
@@ -1,127 +1,128 @@
-use super::common::runner::{setup_package, test_runner};
+use super::common::runner::{setup_package_with_file_patterns, test_runner, BASE_FILE_PATTERNS};
 use indoc::formatdoc;
+use shared::test_utils::node_url::node_rpc_url;
 use shared::test_utils::output_assert::assert_stderr_contains;
 
 #[test]
 fn print_error_if_attributes_incorrect() {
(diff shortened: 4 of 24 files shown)


Draggu (approved):

.github/workflows/release.yml
```diff
@@ -195,7 +195,7 @@ jobs:
       - name: Smoke test
         shell: bash
         env:
-          RPC_URL: ${{ secrets.CHEATNET_RPC_URL }}
+          RPC_URL: ${{ secrets.RPC_URL }}
```
ddoktorski, line 198: Don't `secrets.RPC_URL` and `secrets.NODE_URL` have the same value?

tomek0123456789: Forgot about that, I should have replaced it to `NODE_URL`. Good catch, thanks!

.github/workflows/ci.yml
```diff
@@ -42,6 +50,14 @@ jobs:
         if: github.event_name == 'pull_request'
         run: echo "REPO_NAME=$(echo ${{ github.event.pull_request.head.repo.full_name }}.git)" >> $GITHUB_ENV
 
+      - name: Set NODE_URL for foundry-rs/starknet-foundry
+        if: ${{ github.event.pull_request.head.repo.full_name == 'foundry-rs/starknet-foundry' }}
+        run: echo "NODE_URL=$(echo ${{ secrets.NODE_URL }}:7070)" >> $GITHUB_ENV
+
+      - name: Set NODE_URL for forks
+        if: ${{ github.event.pull_request.head.repo.full_name != 'foundry-rs/starknet-foundry' }}
+        run: echo "NODE_URL=$(echo ${{ secrets.NODE_URL }})" >> $GITHUB_ENV
```
piotmag769, line 59: Is it safe? Why without a port?

.github/workflows/ci.yml
```diff
@@ -72,6 +88,14 @@ jobs:
...
     runs-on: ubuntu-latest
     steps:
+      - name: Set NODE_URL for foundry-rs/starknet-foundry
+        if: ${{ github.event.pull_request.head.repo.full_name == 'foundry-rs/starknet-foundry' }}
+        run: echo "NODE_URL=$(echo ${{ secrets.NODE_URL }}:7070)" >> $GITHUB_ENV
+
+      - name: Set NODE_URL for forks
+        if: ${{ github.event.pull_request.head.repo.full_name != 'foundry-rs/starknet-foundry' }}
+        run: echo "NODE_URL=$(echo ${{ secrets.NODE_URL }})" >> $GITHUB_ENV
+
```
piotmag769, line 98: Not really a fan of that solution, are we sure the last url path segment will be the rpc version in whole ecosystem? Also hardcoing the port here doesn't seem good, imo it would be better to hardcode it in the variable

tomek0123456789: As far as I know, all ([Juno](https://juno.nethermind.io/), [Pathfinder](https://github.com/eqlabs/pathfinder?tab=readme-ov-file#json-rpc-api), [Papyrus](https://github.com/starkware-libs/papyrus/?tab=readme-ov-file#sending-api-requests-to-the-node)) node providers agreed to use the same suffix.

As for hardcoding the port (as we discussed offline), it is for the possibility of using out other node with a different port (so that the `NODE_URL` variable stays the same).

.github/workflows/ci.yml
```diff
@@ -13,6 +13,14 @@ jobs:
     name: Test Forge / Unit and Integration Tests
     runs-on: ubuntu-latest
     steps:
+      - name: Set NODE_URL for foundry-rs/starknet-foundry
+        if: ${{ github.event.pull_request.head.repo.full_name == 'foundry-rs/starknet-foundry' }}
+        run: echo "NODE_URL=$(echo ${{ secrets.NODE_URL }}:7070)" >> $GITHUB_ENV
```
Arcticae, line 18: shouldn't info about port be part of the env var?

crates/forge/tests/e2e/common/runner.rs
```diff
@@ -56,6 +59,33 @@ pub(crate) fn setup_package(package_name: &str) -> TempDir {
...
+    package_name: &str,
+    file_patterns: &[&str],
+) -> TempDir {
+    let temp_dir = setup_package_with_file_patterns(package_name, file_patterns);
+    replace_rpc_url_placeholders(temp_dir.path());
+    temp_dir
+}
+
+fn replace_rpc_url_placeholders(dir_path: &Path) {
+    let url = node_url_with_version().unwrap();
```
Arcticae, line 72: i think this should be smarter and only replace the files with the explicit `.cairo.template` suffix. can be done in later PR

crates/forge/tests/e2e/diagnostics_and_plugins.rs
```diff
@@ -1,127 +1,130 @@
...
 
 #[test]
 fn print_error_if_attributes_incorrect() {
-    let mock_tests_dir = setup_package("diagnostics_and_plugins");
-    let mock_tests_dir_path = mock_tests_dir.path().canonicalize().unwrap();
-    let mock_tests_dir_path_str = mock_tests_dir_path.to_str().unwrap();
-
+    let node_url = node_url_with_version().unwrap();
+    let mock_tests_dir =
+        setup_package_with_rpc_url_placeholders("diagnostics_and_plugins", BASE_FILE_PATTERNS);
```
Arcticae, line 12: i think it should be done automatically, on every `setup_package`

crates/forge/tests/e2e/common/runner.rs
piotmag769: I don't see a need, what it replaces is not a valid syntax for our attributes anyways

.github/workflows/ci.yml
tomek0123456789: `sncast` might want to use this variable in the future, but with a different port (integration node). We could create many secrets/env vars, but then it would require different parsing in tests and higher maintenance. I'm not happy with either choice

cc @THenry14

crates/forge/tests/e2e/common/runner.rs
Arcticae: i don't understand this response @piotmag769, can you explain?

.github/workflows/ci.yml
Arcticae: I mean, if they want to use different port, it's better not to include it here, and i'd rather have 2 different variables for 2 different endpoints than hardcoding this (which is not allowing different ports inside our tests as well)

.github/workflows/ci.yml
Arcticae: This one i'm fine with, but the latter i'd rather have in 2 variables (since it's a different variable essentially)

.github/workflows/ci.yml
THenry14: Honestly I'd leave this as is, it will be easier to maintain than multiple secrets. Our IP won't change, probably ever, ports have changed from time to time.

.github/workflows/ci.yml
THenry14: I'm not sure we should care about setting this for forks. From my experience very few (if any) people actually run CIs on their forks, they just run tests locally and then run tests on our CI when we approve it. If someone is savvy enough to set all the necessary envs and enable actions on their fork, then they are savvy enough to change this port on their fork imo. 

As a side note, I quite like the flow we have in sncast, where we set the whole url in env, but I already asked a question about it in https://github.com/foundry-rs/starknet-foundry/pull/2003/files#r1572097275, since I'm not sure this is applicable to snforge in the same degree it is to sncast.

crates/forge/tests/e2e/diagnostics_and_plugins.rs
tomek0123456789: `setup_package` is called almost 50 times across all tests, wouldn't this be a bit of an unnecessary overhead?

.github/workflows/ci.yml
tomek0123456789: All right, so I guess we remove `NODE_URL` env var setting for forks?

crates/forge/tests/e2e/common/runner.rs
piotmag769: I meant if the stuff in not replaced in `.cairo` files we get an error in compilation anyways, so idk if the function being smarter really makes a lot of sense

crates/forge/tests/e2e/diagnostics_and_plugins.rs
Arcticae: that's why i wanted to have a function which would only look for the "marked" filenames in order to replace anything. but it's a nitpick more than anything, so if you think it's not worth it, you can resolve

.github/workflows/ci.yml
Arcticae: I don't get it. Since the ports will change in the future, why hardcode them? It's easier to update the GH env var than to make a PR and have it merged, isn't it?

THenry14: I find to harder to be honest :D for example, not every developer may be able to do so due to permissions, but every dev will be able to open a PR. Moreover it is very clear what is being used at a glance here (is INTEGRATION_NODE_URL 7070 or 7171?). I don't kow, up to you really since both approaches are fine, but personally Id probably leave this as is.

crates/forge/tests/e2e/diagnostics_and_plugins.rs
```diff
@@ -1,127 +1,128 @@
...
         #[should_panic(url: "https://test.com")]
                       ^***********************^
         Error: Failed to compile test artifact, for detailed information go through the logs above
         
         error: Plugin diagnostic: Expected fork config must be of the form `url: <double quote string>, block_id: <snforge_std::BlockId>`.
-         --> {mock_tests_dir_path_str}/tests/test_fork.cairo:86:7
-        #[fork(
-              ^
+         --> [..]/tests/test_fork.cairo[..]
+        #[fork(url: "http://188.34.188.184:7070/rpc/v0_7", block_id: BlockId::Tag(xddd::d00pa::hehe::BlockTag::Latest))]
```
THenry14, line 120:
```suggestion
        #[fork(url: "{node_rpc_url}", block_id: BlockId::Tag(xddd::d00pa::hehe::BlockTag::Latest))]
```
should this url be here?

.github/workflows/release.yml
```diff
@@ -195,7 +195,7 @@ jobs:
       - name: Smoke test
         shell: bash
         env:
-          RPC_URL: ${{ secrets.CHEATNET_RPC_URL }}
+          NODE_URL: ${{ secrets.NODE_URL }}
```
THenry14, line 198: nitpick: we could just pass the whole url to the script (atm we construct it basen on node ip in smoke_test.sh script), it may potential updates to CIs just a bit easier.

.github/workflows/ci.yml
```diff
@@ -13,14 +13,22 @@ jobs:
...
       - uses: software-mansion/setup-scarb@v1.3.2
       - uses: software-mansion/setup-universal-sierra-compiler@v1
       - run: cargo test --release --lib -p forge
       - run: cargo test --release --bin snforge
-      - run: cargo test --release integration -p forge
+      - name: Run Forge unit and integration tests
+        env:
+          NODE_URL: ${{ secrets.NODE_URL }}
+        run: |
+          SEPOLIA_RPC_URL="${NODE_URL}:7070/rpc/v0_7" cargo test --release integration -p forge
```
THenry14, line 31: nit: it should be possible to set this env variable for the whole workflow at the top of this file https://docs.github.com/en/actions/learn-github-actions/variables#defining-environment-variables-for-a-single-workflow

that way we wouldn't have to set it separately per job

crates/forge/tests/e2e/diagnostics_and_plugins.rs
tomek0123456789: Nope, missed that, thanks for noticing!

THenry14 (approved):
💪

PR #2005, opened 2024-04-11
Add TODO in release.yml

tomek0123456789:
**Stack**:
- #2006
- #2005 ⬅
- #2004
- #2003

⚠️ *Part of a stack created by [spr](https://github.com/ejoffe/spr). Do not merge manually using the UI - doing so may have unexpected results.*

PR #2006, opened 2024-04-11
Update `Running Tests` docs section

tomek0123456789:
**Stack**:
- #2006 ⬅
- #2005
- #2004
- #2003


Arcticae (approved):

docs/src/development/environment-setup.md
```diff
@@ -27,50 +29,70 @@ Please make sure you're using Scarb installed via [asdf](https://asdf-vm.com/) -
...
 
 
+### Environmental variables
+
+Set `NODE_URL` environmental variable to a Sepolia testnet node URL:
+- either manually in your shell
+    ```shell
+    $ export NODE_URL="https://example.com" 
+    ```
+- or inside [`.cargo/config.toml` file](https://doc.rust-lang.org/cargo/reference/config.html#configuration-format)
```
THenry14, line 54: This is an option, but we must make sure our url is not commited at any point to the repo. Maybe we could suggest using dotenv and .env file instead? we alredy do so in sncast cc @ddoktorski

THenry14: also, we use the same variable but in different format in sncast (also set in .env or just in env). if we are putting this info in docs, maybe we should include the same info on sncast as well?

tomek0123456789: Yup, changing that to `.env` file. 
What do you exactly mean by including the same info on sncast?

THenry14: We also use .env and depend on node url declared in it, but up until now we only had this info in readme. Since we are expanding the docs to include this info we should include an info it is needed for snforge and sncast (there will probably be two separate files?)

tomek0123456789: Changed env to be `SEPOLIA_RPC_URL` to match `sncast` env file format

THenry14 (approved):

docs/src/development/environment-setup.md
```diff
@@ -27,53 +29,65 @@ Please make sure you're using Scarb installed via [asdf](https://asdf-vm.com/) -
...
 
 ```shell
 $ cargo fmt
 ```
 
-For linting, it uses [clippy](https://github.com/rust-lang/rust-clippy). You can run it with this command:
+For linting, it uses [clippy](https://github.com/rust-lang/rust-clippy). You can run it with:
 
 ```shell
 $ cargo clippy --all-targets --all-features -- --no-deps -W clippy::pedantic -A clippy::missing_errors_doc -A clippy::missing_panics_doc -A clippy::default_trait_access
```
ddoktorski, line 87: nit: This line can be updated to align with the lint alias defined in the [config.toml](https://github.com/foundry-rs/starknet-foundry/blob/master/.cargo/config.toml).
```suggestion
$ cargo clippy --all-targets --all-features -- --no-deps -W clippy::pedantic -A clippy::module_name_repetitions -A clippy::missing_errors_doc -A clippy::missing_panics_doc -A clippy::default_trait_access
```

PR #2007, opened 2024-04-11
Fix non-deterministic failing test

THenry14:
## Introduced changes

- modifies non-deterministic test, to avoid possible race condition problems

## Checklist

- [X] Linked relevant issue
- [X] Added relevant tests


MaksymilianDemitraszek:
> Linked relevant issue
Nothing is linked here

THenry14:
> > * [x]  Linked relevant issue
> 
> Nothing is linked here, or just an issue does not exist?

Yup, issue does not exist. 

> Also would be helpful to have one or two sentences what was the issue so we don't have to guess it from the code. I assume two script tests were operating on the same state?

I did not have any luck with reproducing this locally, so I had to make an educated guess. I don't think the thing you've mentioned was the case, as we operate on tempdirs that do not include state files. My guess was invalid nonce or generally an account issue (due to the fact that the same account was used for some other chain modifying tests) OR near-identical transactions that could be performed due to the fact that devnet executes them almost instantly. 

The first possible issue I try to resolve by using a new account, the second by creating a new script just for state-checking test case. I think these should prevent CI from failing non-deterministically, but as I couldn't reproduce it I will have to keep an eye on CI runs to make sure that is the case.

crates/sncast/tests/data/scripts/state_script/scripts/src/state_script.cairo
```diff
@@ -0,0 +1,40 @@
+use sncast_std::{
+    declare, deploy, invoke, call, DeclareResult, DeployResult, InvokeResult, CallResult, get_nonce
+};
+
+fn main() {
+    let max_fee = 99999999999999999;
+    let salt = 0x5;
+
+    let declare_nonce = get_nonce('latest');
```
ddoktorski, line 9: What's the rationale for using `latest` instead of `pending` here? I would expect that it might be the root issue of the failing test.

THenry14: this may be valid for tests on real nodes. If I remember correctly atm devnet just returns the same nonce for pending and latest, so this does not really make much difference for devnet tests (as blocks are being accepted instantly).

Arcticae (approved):
Hope that fixes it

ddoktorski (approved):
In my opinion, the actual fix involves just changing the account to another that doesn't modify the chain state in other tests. However, given our uncertainty, it's fine to also  add a separate script.

PR #1983, opened 2024-04-05
Refactored development docs

MaksymilianDemitraszek:
Closes #1976 

## Introduced changes

- made prerequisites a clear section
- added contributions.md link

## Checklist

- [x] Linked relevant issue
- [x] Updated relevant documentation
- [x] Added relevant tests
- [x] Performed self-review of the code
- [x] Added changes to `CHANGELOG.md`

PR #1948, opened 2024-03-28
Stop caching contracts artifacts

THenry14:
Closes https://github.com/foundry-rs/starknet-foundry/issues/1800

## Introduced changes

- remove caching script

## Checklist

- [X] Linked relevant issue
- [X] Updated relevant documentation
- [X] Added relevant tests
- [X] Performed self-review of the code
- [X] Added changes to `CHANGELOG.md`

PR #2008, opened 2024-04-11
Use devnet fork in sncast tests

### LINKED ISSUE #1800: Remove pre-compilation of contracts/scripts for sncast tests
In order to cache scarb artifacts we once created a `build_contracts.sh` script, that compiles contracts if there are no artifacts, and skips the compilation if artifacts are already present. The script is no longer needed, and in fact generates some problems when we update scarb. We should just get rid of it.

### LINKED ISSUE #1985: Update devnet to support forking
Since devnet started supporting forking, we can leverage this feature in our tests. Currently we manually set up contracts used in tests by building, declaring and deploying them. WIth this change, all contracts can be directly retrieved from the Sepolia testnet and will act as if they were deployed to the local devnet. This could also be used in the future for testing other account types that are not predeployed on devnet.

ddoktorski:
Closes #1985 
Closes #1800

## Introduced changes

- Fork the Sepolia testnet network in the sncast devnet tests
- Use contracts from the Sepolia instead of declaring/deploying them locally on devnet
- Stop building contracts locally

## Checklist

- [X] Linked relevant issue
- [X] Added relevant tests


crates/sncast/tests/helpers/devnet.rs
```diff
@@ -59,28 +61,10 @@ fn start_devnet() {
...
-        "CAST_MAP",
-    ));
-    rt.block_on(declare_contract(
-        "user4",
-        "/constructor_with_params/target/dev/constructor_with_params_ConstructorWithParams",
-        "CAST_WITH_CONSTRUCTOR",
-    ));
 
     rt.block_on(deploy_keystore_account());
     rt.block_on(deploy_cairo_0_account());
```
THenry14, line 67: would it be possible to also get rid of those?

THenry14 (commented):
glanced through, seems to be looking ok. will make more thorough review when marked as ready

.github/workflows/ci.yml
```diff
@@ -96,7 +96,11 @@ jobs:
       - uses: software-mansion/setup-scarb@v1.3.2
       - uses: software-mansion/setup-universal-sierra-compiler@v1
       - name: Run tests
-        run: cargo test --release -p sncast
+        env:
+          NODE_URL: ${{ secrets.NODE_URL }}
+        run: |
+          export FORK_NETWORK_URL="${NODE_URL}:7070/rpc/v0_7"
```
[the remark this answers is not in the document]
ddoktorski: If we construct the exact URL in the code, it will enforce the use of our node URL, which could be problematic, especially for external contributors. We might consider appending `/rpc/v0_x` at the code level, but only if this path is common across all node providers.

crates/sncast/tests/helpers/devnet.rs
```diff
@@ -34,6 +30,10 @@ fn start_devnet() {
         }
     }
 
+    dotenv::dotenv().ok();
+    let fork_network_url =
+        from_env("FORK_NETWORK_URL").expect("Failed to get FORK_NETWORK_URL environment variable");
```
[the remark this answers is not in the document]
ddoktorski: Renamed to `SEPOLIA_RPC_URL`

crates/sncast/tests/helpers/devnet.rs
ddoktorski: As discussed offline, it's technically possible to eliminate this. However, we would then need to either keep the private keys of the accounts deployed to Sepolia in the repo or find a way to conceal them. Given these considerations, it seems better to leave it as it is.

MaksymilianDemitraszek:
@ddoktorski I think https://github.com/foundry-rs/starknet-foundry/pull/1948 is a PR, so it cannot be closed

ddoktorski:
Thanks! It was supposed to be the issue linked in that PR, so #1800 The PR itself probably can be closed, but I will confirm it with @THenry14

MaksymilianDemitraszek:
@ddoktorski @

docs/src/development/environment-setup.md
```diff
@@ -29,6 +29,11 @@ $ cargo check
 > and use the newest [universal-sierra-compiler](https://github.com/software-mansion/universal-sierra-compiler)
 > after setting up the development environment, otherwise the tests will fail.
 
+> 📝 **Note**
+>
+> To run `sncast` tests, it is required to set `SEPOLIA_RPC_URL` environment variable, which is
+> used to fork the Sepolia testnet network in the devnet tests. This variable can be set in the `.env` file.
+
```
MaksymilianDemitraszek, line 36: I think notes should be used for additional info. Thise seems like core prerequisite, can se just write it as a standard text?

crates/sncast/tests/helpers/constants.rs
```diff
@@ -18,3 +20,15 @@ pub const DEVNET_PREDEPLOYED_ACCOUNT_ADDRESS: &str =
 // https://github.com/0xSpaceShard/starknet-devnet-rs/blob/1a76e9d29541af2667ca815c47bb332cead27c55/crates/starknet/src/constants.rs#L17
 pub const DEVNET_OZ_CLASS_HASH_CAIRO_1: &str =
     "0x61dac032f228abef9c6626f995015233097ae253a7f72d68552db02f2971b8f";
+
+pub const MAP_CONTRACT_ADDRESS_SEPOLIA: &str =
+    "0xcd8f9ab31324bb93251837e4efb4223ee195454f6304fcfcb277e277653008";
```
MaksymilianDemitraszek, line 25: I wonder if it wouldn't be better to use forked class hashes but deploy contract instance per-test. Would prevent race conditions.

docs/src/development/environment-setup.md
ddoktorski: Changed to a standard text. I think it would be beneficial to modify the environment setup docs. For instance running `install_denvet.sh` seems to be a core prerequisite as well, but it's mentioned in the note above.

crates/sncast/tests/helpers/constants.rs
ddoktorski: Generally agree that ideally it should be done this way. However, given that it has been done in a similar manner so far, and that tests don't depend on each other (the order doesn't matter) - I am not sure that making this change would be beneficial.

docs/src/development/environment-setup.md
MaksymilianDemitraszek: Yup, there is even a open PR for that #1983

crates/sncast/tests/helpers/constants.rs
MaksymilianDemitraszek: Just a suggestion, ok with leaving it as it is.

THenry14 (approved):

PR #2018, opened 2024-04-12
Fold appendix in docs by default

Arcticae:
## Introduced changes
Resulting menu fold:
image

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


drknzz (approved):
It's not folding just the appendix, but it's still fine since we don't have nested sections anywhere else

I like 👍

PR #2028, opened 2024-04-16
Delete mention of enabling the casm explicitly

Arcticae:


piotmag769 (approved):

MaksymilianDemitraszek:
Have we remove this from the template?

MaksymilianDemitraszek:
image
We also could probably remove it from our codebase.

piotmag769:
@MaksymilianDemitraszek enabling casm in Scarb.toml is fine, it's just not necessary - if the casm is present we will use it, if not, we will use USC

MaksymilianDemitraszek:
Imo would be better to have one way to do this.

piotmag769:
>Don't see any reason why we would want to consume scarb produced casm.

It is embedded in our logic though (in scarb-api I believe), I remember that I supported this idea. If Scarb already produced casm why should we recompile the sierra?

crates/sncast/tests/utils/build_contracts.sh
```diff
@@ -15,10 +15,22 @@ if [ ! -d "${REPO_ROOT}/crates/sncast/tests/utils/devnet/bin" ]; then
   exit 1
```
THenry14, line 15: this file will be removed in https://github.com/foundry-rs/starknet-foundry/pull/2008 as it is no longer needed

THenry14 (changes requested):
Do you mind giving some more context? what is this change related to?

piotmag769:
>Do you mind giving some more context? what is this change related to?

@THenry14 You don't need the `casm = true` in Scarb since we can compile with USC

https://github.com/foundry-rs/starknet-foundry/blob/master/crates/scarb-api/src/lib.rs#L55

THenry14:
> > Do you mind giving some more context? what is this change related to?
> 
> @THenry14 You don't need the `casm = true` in Scarb since we compile contracts with USC when needed
> 
> https://github.com/foundry-rs/starknet-foundry/blob/master/crates/scarb-api/src/lib.rs#L55

ahh, thanks. this is the context I was missing 🙏

PR #2038, opened 2024-04-17
Add paths to sierra to profiler call trace

piotmag769:
## Introduced changes

- as in the title - needed for decoupling profiler from specific sierra artifacts location. Profiler will try to parse the content of the file found under the path as:
1. `ContractClass`
2. `VersionedProgram`
3. `VersionedProgram` from `CompiledTestCrateRaw` (it is a temporary coupling of logic that won't affect other tools, when snforge migrates to test artifacts from Scarb it will be replaced by the test artifacts format, which I assume will be public)

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/cheatnet/src/runtime_extensions/forge_runtime_extension/contracts_data.rs
```diff
@@ -81,6 +81,13 @@ impl ContractsData {
             .map(|contract| &contract.class_hash)
     }
 
+    #[must_use]
+    pub fn get_source_sierra_path_for_contract(&self, name: &str) -> Option<&Utf8PathBuf> {
```
MaksymilianDemitraszek, line 85:
```suggestion
    pub fn get_source_sierra_path(&self, contract_name: &str) -> Option<&Utf8PathBuf> {
```

MaksymilianDemitraszek (commented):
Should we add some test veryfing if this is added 🤔

piotmag769:
@MaksymilianDemitraszek the test is already there, we check if the field with CairoExecutionInfo is not None (previously we only checked the vm trace)

crates/forge/tests/e2e/build_trace_data.rs
```diff
@@ -110,7 +110,8 @@ fn trace_has_vm_trace() {
 
 fn assert_vm_trace_exists(trace: &ProfilerCallTrace) {
     assert!(
-        trace.vm_trace.is_some() || trace.entry_point.function_name == Some(String::from("fail"))
+        trace.cairo_execution_info.is_some()
+            || trace.entry_point.function_name == Some(String::from("fail"))
```
piotmag769, line 114: @MaksymilianDemitraszek it is tested here

crates/forge-runner/src/build_trace_data.rs
```diff
@@ -35,18 +36,39 @@ pub const TEST_CODE_FUNCTION_NAME: &str = "SNFORGE_TEST_CODE_FUNCTION";
...
     let vm_trace = value.vm_trace.as_ref().map(|trace_data| {
         trace_data
             .iter()
             .map(build_profiler_trace_entry)
             .collect_vec()
     });
 
+    let cairo_execution_info = if vm_trace.is_some() && entry_point.contract_name.is_some() {
+        let contract_name = entry_point.contract_name.as_ref().unwrap();
+        let source_sierra_path = if contract_name == TEST_CODE_CONTRACT_NAME {
```
Arcticae, line 53: Wouldn't the logic be simpler if we just inserted this as a contract data into a map?

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/contracts_data.rs
```diff
@@ -86,6 +86,13 @@ impl ContractsData {
             .map(|contract| &contract.class_hash)
     }
 
+    #[must_use]
+    pub fn get_source_sierra_path(&self, name: &str) -> Option<&Utf8PathBuf> {
+        self.contracts
+            .get(name)
+            .map(|contract| &contract.source_sierra_path)
```
Arcticae, line 93: the logic could be placed here perhaps

crates/forge-runner/src/build_trace_data.rs
piotmag769: But it is not a contract, so I didn't want to do that

Arcticae: not a deal breaker. you can resolve if you want

Arcticae: buut if you did that this way, you could avoid prop-drilling this one special path

Arcticae: let that sink in

Arcticae (approved):

crates/forge-runner/src/build_trace_data.rs
```diff
@@ -35,18 +36,39 @@ pub const TEST_CODE_FUNCTION_NAME: &str = "SNFORGE_TEST_CODE_FUNCTION";
...
+            test_artifacts_path
+        } else {
+            contracts_data
+                .get_source_sierra_path(contract_name)
+                .unwrap()
+        }
+        .clone();
```
MaksymilianDemitraszek, line 60: I have no idea what this large if statement does. Please provide some abstraction.

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/contracts_data.rs
```diff
@@ -86,6 +86,13 @@ impl ContractsData {
             .map(|contract| &contract.class_hash)
     }
 
+    #[must_use]
+    pub fn get_source_sierra_path(&self, name: &str) -> Option<&Utf8PathBuf> {
```
MaksymilianDemitraszek, line 90:
```suggestion
    pub fn get_source_sierra_path(&self, contract_name: &str) -> Option<&Utf8PathBuf> {
```
what is `name` in this context?

MaksymilianDemitraszek (changes requested):

crates/forge-runner/src/build_trace_data.rs
piotmag769: Done

crates/forge-runner/src/build_trace_data.rs
```diff
@@ -55,20 +60,57 @@ pub fn build_profiler_call_trace(
...
+
+fn build_cairo_execution_info(
+    entry_point: &ProfilerCallEntryPoint,
+    vm_trace: Option<Vec<ProfilerTraceEntry>>,
+    contracts_data: &ContractsData,
+    test_artifacts_path: &Utf8Path,
+) -> Option<CairoExecutionInfo> {
+    let cairo_execution_info_available = vm_trace.is_some() && entry_point.contract_name.is_some();
+
+    if cairo_execution_info_available {
```
MaksymilianDemitraszek, line 77: This should not be decision of forge, this is profiler abstraction layer. Forge should just provide all available information and its profiler role to determine if it can interpret it or not.

MaksymilianDemitraszek (changes requested):

Issue #2042, opened 2024-04-18
Use executable attribute


THenry14:
### Which components does the task require to be changed? (think hard pls)

sncast

### Description

After https://github.com/software-mansion/scarb/issues/1254 we should start using executable attribute to mark scripts' main functions. Having this will also allow us to get rid of double-compilation of scripts in sncast main.

PR #2043, opened 2024-04-18
Workaround missing function debug name info in release artifacts

THenry14:
## Introduced changes

- Works around an issue, where there are no function debug name info in release artifacts. The issue will be fully fixed by https://github.com/foundry-rs/starknet-foundry/issues/2042

## Checklist

- [X] Linked relevant issue
- [X] Added relevant tests

### DIFF
--- a/crates/sncast/src/starknet_commands/script/run.rs
+++ b/crates/sncast/src/starknet_commands/script/run.rs
@@ -454,6 +454,7 @@ fn inject_lib_artifact(
         .target_dir
         .clone()
         .unwrap_or_else(|| metadata.workspace.root.join("target"));
+    // TODO(#2042)
     let sierra_path = &target_dir.join("dev").join(sierra_filename);
 
     let lib_artifacts = ScriptStarknetContractArtifacts {
--- a/crates/sncast/src/starknet_commands/script/run.rs
+++ b/crates/sncast/src/starknet_commands/script/run.rs
@@ -326,7 +326,8 @@ pub fn run(
     .with_context(|| "Failed to set up runner")?;
 
     let name_suffix = module_name.to_string() + "::main";
-    let func = runner.find_function(name_suffix.as_str())?;
+    let func = runner.find_function(name_suffix.as_str())
+        .context("Failed to find main function in script - please make sure `sierra-replace-ids` is not set to `false` for `dev` profile in script's Scarb.toml")?;
 
     let (entry_code, builtins) = runner.create_entry_code(func, &Vec::new(), usize::MAX)?;
     let footer = SierraCasmRunner::create_code_footer();
--- a/crates/sncast/src/starknet_commands/script/run.rs
+++ b/crates/sncast/src/starknet_commands/script/run.rs
@@ -454,9 +454,7 @@ fn inject_lib_artifact(
         .target_dir
         .clone()
         .unwrap_or_else(|| metadata.workspace.root.join("target"));
-    let sierra_path = &target_dir
-        .join(&metadata.current_profile)
-        .join(sierra_filename);
+    let sierra_path = &target_dir.join("dev").join(sierra_filename);
 
     let lib_artifacts = ScriptStarknetContractArtifacts {
         sierra: fs::read_to_string(sierra_path)?,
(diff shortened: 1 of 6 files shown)


crates/sncast/tests/e2e/script/general.rs
```diff
@@ -508,3 +508,41 @@ async fn test_state_file_rerun_failed_tx() {
         .unwrap();
     assert_tx_entry_success(invoke_tx_entry, "invoke");
 }
+
+#[tokio::test]
+async fn test_using_release_profile() {
```
ddoktorski, line 513: It seems that this test also passes successfully without the changes from this PR (for example on the master branch). Could you check this?

crates/sncast/src/starknet_commands/script/run.rs
```diff
@@ -454,9 +454,7 @@ fn inject_lib_artifact(
         .target_dir
         .clone()
         .unwrap_or_else(|| metadata.workspace.root.join("target"));
-    let sierra_path = &target_dir
-        .join(&metadata.current_profile)
-        .join(sierra_filename);
+    let sierra_path = &target_dir.join("dev").join(sierra_filename);
```
ddoktorski, line 457: Let's also add `TODO #2042` here, so we won't forget to update it.

piotmag769 (commented):
Maybe a better solution would be to tell users to override the release profile config https://docs.swmansion.com/scarb/docs/reference/profiles.html#overriding-built-in-profile-properties? It will affect contract compilation though cc @maciektr

maciektr:
@piotmag769 I believe this is specifically done to allow a build with not replaced ids.

piotmag769:
>I believe this is specifically done to allow a build with not replaced ids.

Doesn't calling build twice replace these artifacts anyways?

piotmag769:
>I believe this is specifically done to allow a build with not replaced ids.

If that's the rationale then it's ok

crates/sncast/tests/e2e/script/general.rs
THenry14: it fails without the changes from this PR (clean master, only this test added)

```
---- e2e::script::general::test_using_release_profile stdout ----
thread 'e2e::script::general::test_using_release_profile' panicked at crates/sncast/tests/e2e/script/general.rs:540:32:

--- Expected
++++ actual:   stdout
   1      - ...
   2      - command: script run
   3      - status: success
        1 +    Compiling lib(map_script) map_script v0.1.0 (/private/var/folders/rd/q9ymr6vd6w150wm7g3f36hf00000gn/T/.tmp5uDNi6/Scarb.toml)
        2 +    Compiling starknet-contract(map_script) map_script v0.1.0 (/private/var/folders/rd/q9ymr6vd6w150wm7g3f36hf00000gn/T/.tmp5uDNi6/Scarb.toml)
        3 +     Finished release target(s) in 2 seconds

Exit status: 0
stderr:

command: script run
error: No such file or directory (os error 2)

```

crates/sncast/src/starknet_commands/script/run.rs
THenry14: done in newest commit

THenry14:
> Maybe a better solution would be to tell users to override the release profile config (`sierra-replace-ids` property) https://docs.swmansion.com/scarb/docs/reference/profiles.html#overriding-built-in-profile-properties? It will affect contract compilation though cc @maciektr

Apart from what Maciek said, this should not be considered a proper fix, rather a workaround. We wait for `executable attributes` in scarb to properly resolve this issue.

piotmag769 (commented):
Can we add a test for a case when dev profile is overriden like this in Scarb.toml:
```
[profile.dev.cairo]
sierra-replace-ids = false
```
and see what the error message is?

THenry14:
> Can we add a test for a case when dev profile is overriden like this in Scarb.toml:
> 
> ```
> [profile.dev.cairo]
> sierra-replace-ids = false
> ```
> 
> and see what the error message is?

when I do this for dev I get the error with missing function debug name (Function with suffix `script::main` to run not found)

Issue #1855, opened 2024-03-11
Support argent accounts


THenry14:
### Which component is your feature related to?

Cast

### Feature Request

We should have on par to OZ support for argent accounts in sncast. 
Let's start with argent account usage, it should be possible to:

- import already existing argent account to accounts file (account add)
- remove argent account from accounts file (account delete - this should already be possible, but let's double check)
- use an argent account to make transactions

It may be handy to annotate non-oz accounts with a type in the accounts file. While adding such accounts let's remember to include this info in it. 

The documentation should also be updated to include information on what is currently supported, what is the support's scope, and what is still planned for the future.

After this is done, we should think about creating and deploying such accounts using sncast. This should be rather easy, since it seems to be support for this in the sdk.

PR #2045, opened 2024-04-18
Add tests for argent account

ddoktorski:
Towards #1855 

## Introduced changes

- Add fixture to deploy an argent account
- Run tests using the deployed argent account

## Checklist

- [X] Linked relevant issue
- [X] Added relevant tests


crates/sncast/tests/data/accounts/accounts.json
```diff
@@ -47,19 +47,27 @@
...
         },
-        "cairo1": {
+        "oz_cairo_1": {
             "private_key": "0x88ecc06581d81c76cef06d6f4f0c1b28",
             "public_key": "0x8cdbe26bc82084b04eccb3c6f8a76f12ad6c4015b3dc8ab90dc840d42cac29",
             "address": "0x691a61b12a7105b1372cc377f135213c11e8400a546f6b0e7ea0296046690ce",
             "legacy": false
         },
-        "cairo0": {
+        "oz_cairo_0": {
```
THenry14, line 56: I'd rather add a new field to accounts in here, something like "type", where it can be either one of oz, argent or braavos. The name could stay as well, but I realy think it may be a good idea to also unclude a type in there.

crates/sncast/tests/helpers/constants.rs
```diff
@@ -32,3 +32,6 @@ pub const MAP_CONTRACT_DECLARE_TX_HASH_SEPOLIA: &str =
 
 pub const CONSTRUCTOR_WITH_PARAMS_CONTRACT_CLASS_HASH_SEPOLIA: &str =
     "0x59426c817fb8103edebdbf1712fa084c6744b2829db9c62d1ea4dce14ee6ded";
+
+pub const ARGENT_ACCOUNT_CLASS_HASH: &str =
+    "0x029927c8af6bccf3f6fda035981e765a7bdbf18a2dc0d630494f8758aa908e2b";
```
THenry14, line 37: where was it taken from?

THenry14 (changes requested):

crates/sncast/tests/data/accounts/accounts.json
ddoktorski: It will be necessary if we add support for creating and deploying, but I wouldn't add it at this point, because it's currently not needed for anything.

crates/sncast/tests/helpers/constants.rs
ddoktorski: I just checked the class hash of the account that I deployed through Argent X. I couldn't find a specific class hash either in the Argent documentation or in their public repository. I tried to get some information from their Discord server but received no response. Notably, `starkli` also [utilizes this class hash](https://github.com/xJonathanLEI/starkli/blob/master/src/account.rs#L60), which is labeled as the `ArgentX` type on [Voyager](https://sepolia.voyager.online/class/0x029927c8af6bccf3f6fda035981e765a7bdbf18a2dc0d630494f8758aa908e2b).

crates/sncast/tests/data/accounts/accounts.json
THenry14: We want to start communicating with users that using argent is now officially possible. It may happen that we will not be ready with add/create commands in time for the next release. The users then will be forced to add accounts manually to the files, if they want to use such accounts. This will in turn make it very difficult for us to recognize the account type (we would have to assert this each and every time agains the chain, which is far from perfect). By adding this field now, we ensure that whatever happens we provide the users with a simple way to mark non-oz account, and do not have to do additional calls to chain in the future. It would be like "hey, this is an experimental way to use argent accounts, just mark them in the file if you really have to use them now and can't wait for add command". Does that make sense?

crates/sncast/tests/helpers/constants.rs
THenry14: fine, thanks. I assume it is the same for all networks, right?

crates/sncast/tests/data/accounts/accounts.json
THenry14: that reminds me - we should probably also update changelog with this information, right?

ddoktorski: > We want to start communicating with users that using argent is now officially possible

I'm not convinced to be honest, I would rather prefer to say that it's possible to use accounts that perform validation using the standard Stark signer. We don't make any changes in the code logic here.

>  By adding this field now, we ensure that whatever happens we provide the users with a simple way to mark non-oz account, and do not have to do additional calls to chain in the future. It would be like "hey, this is an experimental way to use argent accounts, just mark them in the file if you really have to use them now and can't wait for add command"

It doesn't make much sense to me. Once we add logic for creating and deploying argent accounts in a basic form then we will inform about it. What's the rationale to do this earlier?

THenry14: > I'm not convinced to be honest, I would rather prefer to say that it's possible to use accounts that perform validation using the standard Stark signer. We don't make any changes in the code logic here.

if it's possibe, it means it's officially supported. 

> It doesn't make much sense to me. Once we add logic for creating and deploying argent accounts in a basic form then we will inform about it. What's the rationale to do this earlier?

People want to use them accounts now, and there is no reason to prevent them from it. We should inform that this is indeed possible, and they can import already existing accounts and just use them.

ddoktorski: As discussed offline, we will add the `type` field along with changes to the `account add` command in a separate PR. This will allow the addition of an argent account to the accounts file.

PR #2050, opened 2024-04-19
Add `type` field to accounts file

ddoktorski:
Towards #1855

Discussed at https://github.com/foundry-rs/starknet-foundry/pull/2045#discussion_r1570812864

## Introduced changes

- Add required flag `--type` to `account add` command
- Save account type information in the accounts file

## Checklist

- [X] Linked relevant issue
- [X] Added relevant tests


crates/sncast/src/starknet_commands/account/add.rs
```diff
@@ -23,6 +24,10 @@ pub struct Add {
     #[clap(short, long, requires = "private_key_input")]
     pub address: FieldElement,
 
+    /// Type of the account
+    #[clap(short = 't', long = "type")]
+    pub account_type: AccountType,
```
MaksymilianDemitraszek, line 29: I am not sure if `type` is the correct word. `provider` maybe would be better? Or implementation?

Also, if i use help does ValueEnum list the options? if not we should document it.

crates/sncast/src/starknet_commands/account/mod.rs
```diff
@@ -34,18 +34,35 @@ pub enum Commands {
     Delete(Delete),
 }
 
+#[derive(ValueEnum, Clone, Debug)]
+pub enum AccountType {
+    Oz,
```
MaksymilianDemitraszek, line 39: Imo there is no point in using acronym, whole name would be more obvious and imo it does not add that much writing. Imo we should at least use one version, either `open_zeppelin` or `oz`

crates/sncast/tests/e2e/account/add.rs
```diff
@@ -194,6 +209,40 @@ pub async fn test_nonexistent_account_address() {
...
+    let output = snapbox.assert().failure();
+
+    assert_stderr_contains(
+        output,
+        indoc! {r"
+        error: invalid value 'unknown_type' for '--type <ACCOUNT_TYPE>'
+          [possible values: oz, argent]
+    "},
+    );
+}
```
MaksymilianDemitraszek, line 244: This is logic of clap library, not sure if it is useful test in the context of cast. It is basically isomorphic to our listing in the command args.

docs/src/starknet/account.md
```diff
@@ -9,7 +9,7 @@ and the second one deploys it to the network. After deployment, account can be u
 To remove an account from the accounts file, you can use  `sncast account delete`. Please note this only removes the account information stored locally - this will not remove the account from Starknet.
 
 > 💡 **Info**
-> Currently, only OpenZeppelin account creation is supported.
+> Currently, only OpenZeppelin account creation and deployment is supported.
```
MaksymilianDemitraszek, line 12: Though we list both oz and argent in the cli, no?

MaksymilianDemitraszek (changes requested):

crates/sncast/src/starknet_commands/account/mod.rs
ddoktorski: I think there's a significant difference between typing `--type oz` and `--type open_zeppelin`. On the other hand nothing prevents us from having it more descriptive In the accounts file. If you strongly think we should stick with one format, I'd prefer the acronym.

docs/src/starknet/account.md
ddoktorski: But only for the `account add` command at the moment.

crates/sncast/tests/e2e/account/add.rs
ddoktorski: It serves as a kind of sanity check. Personally I think that it makes some sense for e2e test, but I'm also fine with removing it.

docs/src/starknet/index.md
```diff
@@ -6,7 +6,8 @@ Starknet Foundry `sncast` is a command line tool for performing Starknet RPC cal
 > At the moment, `sncast` only supports contracts written in [Cairo](https://github.com/starkware-libs/cairo) v1 and v2.
 
 > ⚠️ **Warning**
-> Currently, only OpenZeppelin accounts are supported.
+> Currently, support is only provided for accounts that use the default signature based on the [Stark curve](https://docs.starknet.io/documentation/architecture_and_concepts/Cryptography/stark-curve).
+> It is recommended to use OpenZeppelin accounts.
```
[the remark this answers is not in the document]
THenry14: I actually think we **should** include the info that oz is the one type that is "recommended/fully supported" to not give users the wrong idea about our support of argent/braavos, until this support is on par with oz. The thing is, seeing just this page, someone might think every account is usable/deployable to the same degree. 

The message should stand imo, but perhaps with different wording like it was in `account` section of the docs (I'm fine reusing the whole sentence).

crates/sncast/src/starknet_commands/account/add.rs
MaksymilianDemitraszek: Starkli refers to those as `variants` https://book.starkli.rs/accounts imo it is better than `type` but still a bit off. 

If `provider` is taken I would fall back to `implementation`. `Type` seems off to me as it suggests set of accounts characterised by some properties. In our case each of those is a singleton, there are no set of properites, whose implementations implement the same interface.

docs/src/starknet/account.md
MaksymilianDemitraszek: I would make it a table instead with collumns:
Full name, cli arg, supported commands.

crates/sncast/src/starknet_commands/account/mod.rs
MaksymilianDemitraszek: I don't see a problem with typing open_zeppelin. This is not a command you run very often, full name is more discoverable. Currently somebody has to know what `oz` means.

crates/sncast/src/starknet_commands/account/add.rs
ddoktorski: > Starkli refers to those as variants

Only within the documentation, in account files it is also a `type`.

> Type seems off to me as it suggests set of accounts characterised by some properties

Actually, I think it can be interpreted this way if we think about different class hashes as instances of the same account type. However, I would think about it in a different way. There is an account at the `sncast` level that is characterised by the type. The cast interface is the same for every type, but under the hood, different types have different implementations, especially for constructor and signature logic.

crates/sncast/src/starknet_commands/account/mod.rs
ddoktorski: What about leaving the acronym, but improving the description for `--help` to have something like this?
```
          Possible values:
          - oz:     OpenZeppelin account implementation
          - argent: Argent account implementation
```

docs/src/starknet/account.md
ddoktorski: It would only make things more confusing in my opinion. The current sentence seems straightforward, but maybe it could be rephrased a bit e.g. `Currently, creation and deployment are only supported for OpenZeppelin accounts.`

> This sentence suggests bravos is not supported, it may be misleading.

Creation and deployment for braavos accounts are not supported at the moment.

crates/sncast/src/starknet_commands/account/mod.rs
MaksymilianDemitraszek: Sounds good

crates/sncast/src/starknet_commands/account/add.rs
MaksymilianDemitraszek: Account abstraction is a specific interface and those are different implementations, it is much better word and I don't see any valid argument why we shouldn't use it. 
We even use the word `implementation` in the explanation, why just don't name the set od implementations `implementation` instead of naming a set of implementations `type`. 
```
 Possible values:
          - oz:     OpenZeppelin account implementation
```


> Actually, I think it can be interpreted this way if we think about different class hashes as instances of the same account type. However, I would think about it in a different way. There is an account at the sncast level that is characterised by the type. The cast interface is the same for every type, but under the hood, different types have different implementations, especially for constructor and signature logic.

I don't understand this explanation, sorry 😞

THenry14: > Account abstraction is a specific interface and those are different implementations

technically, and correct me if I'm wrong, an account may support a different signer (eg braavos), or have a unique features in comparison to other account types/implementations, so it then technically alters an interface, right? point is, while the said contract must fulfill specific requirements in order to be considered an account, it may also augument them. In that case I would argue if a word 'implementation' is better that a 'type' (although generally I think both of those words are imperfect in our case).

In my opinion a big argument to leave this as 'type' is that (as Darek said) starkli already uses this nomenglature in their codebase, and I don't think having a 'type' field in starkli-style account files and something different like 'implementation' in legacy-starknet-style account files we utilize by default is a wise thing to do - it introduces unnecessary complexity to our code (handling two names for the same thing - remember we cannot introduce this breaking change to starkli accounts) and is just plain pointless.

The question then arises - if we use 'type' in the files holding accounts (which I very strongly suggest), shouldn't we use the same word in docs in order to not introduce more confusion among users? I'd say we should, but I see your point that the word does not properly describe differences between oz/argent/braavos. Perhaps there is a better word to describe it altogether, and I'm keen to hear your ideas lads, I'm just not sure.

MaksymilianDemitraszek (approved):
Everything except the `type` name looks good

crates/sncast/src/starknet_commands/account/add.rs
MaksymilianDemitraszek: > In my opinion a big argument to leave this as 'type' is that (as Darek said) starkli already uses this nomenglature in their codebase

In my opinion it is not an argument pro, and even should be considered slightly against, to correct our initial bias. We should pick the name that describes the property the best, we should not follow every mistake that starkli does, foundry has more stars on github. We should put an effort to build the best possible standard.

MaksymilianDemitraszek: > an account may support a different signer (eg braavos), or have a unique features in comparison to other account types/implementation

Agreed there could be multiple implementations of an account by a single provider with different features. If other providers would follow the same feature sets distinction, then this could eventually yield an account types. Though still I think we should allow users to select an `implementation` here as user selects specific code.

MaksymilianDemitraszek: > if we use 'type' in the files holding accounts (which I very strongly suggest), shouldn't we use the same word in docs in order to not introduce more confusion among users? I'd say we should

I think we should use the word implementation everywhere, because that's what it is. We should put an effort to correct the mistakes if we observe them, not just follow "others do that" or "we did it already somewhere else"

THenry14: > not just follow "others do that" or "we did it already somewhere else"

this is not the case at all - this is a case of "We must support starkli accounts, and they are using 'type' fields, so in order to not break things we should use the same". I don't think there is sensible effort we could put in to "correct the mistake", because we are depedend on starkli with this one - if we deviate it is a breaking change that makes starlki accounts unusable, and as I said - there is no point in my opinion to have "type" in one file and "implementation" in other, this just make our code more complex and can lead to issues.

MaksymilianDemitraszek: > because we are depedend on starkli with this one

I am not sure why this is a breaking change, changing the flag does not have to modify the fields in the account file. We just offer a compatibility layer over starkli, we don't have to do the same.

MaksymilianDemitraszek: Discussed offline, let's leave the name `type`

PR #2060, opened 2024-04-23
Improve deserialising values from cairo

Draggu:
Related #2059

## Introduced changes

- Add `FromReader` trait
- Add derive macro for `FromReader` trait

## Checklist

- [x] Linked relevant issue
- [x] Updated relevant documentation
- [ ] Added relevant tests
- [x] Performed self-review of the code
- [x] Added changes to `CHANGELOG.md`

PR #2062, opened 2024-04-23
Use from reader derive

### LINKED ISSUE #2059: Improve deserialising values from cairo

snforge

### Description

Currently we deserialize them manually like this
```rs
let version = input_reader.read_option_felt()?;
let account_contract_address = input_reader.read_option_felt()?;
let max_fee = input_reader.read_option_felt()?;
let signature = input_reader.read_option_vec()?;
let transaction_hash = input_reader.read_option_felt()?;
let chain_id = input_reader.read_option_felt()?;
let nonce = input_reader.read_option_felt()?;
```
we can make this easier, more readable and less error-prone with simply derive macro

Draggu:
Closes #2059

## Introduced changes

- Use `FromReader` derive macro

Follow-up to #2060

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/cheatnet/src/runtime_extensions/forge_runtime_extension/cheatcodes/spoof.rs
```diff
@@ -1,8 +1,35 @@
...
+impl FromReader for ResourceBounds {
+    fn from_reader(reader: &mut BufferReader<'_>) -> BufferReadResult<Self> {
+        let count = reader.read_felt()?;
+
+        let count = count.to_usize().ok_or(BufferReadError::ParseFailed)?;
+        let result = reader.read_slice(count * 3)?; // ResourceBounds struct has 3 fields
+
+        Ok(ResourceBounds(result.to_owned()))
+    }
+}
```
Arcticae, line 30: Why isn't this derivable? 🤔

Draggu: Because it want to read 3 felts instead of 1
>         let result = reader.read_slice(count * 3)?; // ResourceBounds struct has 3 fields

crates/runtime/src/utils/buffer_reader.rs
```diff
@@ -0,0 +1,76 @@
...
+        self.read()
+    }
+
+    pub fn read_vec<T: FromConv<Felt252>>(&mut self) -> BufferReadResult<Vec<T>> {
+        self.read()
+    }
+
+    pub fn read_string(&mut self) -> BufferReadResult<String> {
+        self.read()
+    }
```
piotmag769, line 71: Do we need these? We can just do self.read() anyways, right?

piotmag769 (approved):
Not rebased

piotmag769 (commented):
Based

PR #2063, opened 2024-04-24
Revamp `storage` cheatcode appendix

Arcticae:
## Introduced changes

- Moves store & load guides to advanced testing
- merged them into one instead of 2
- minimized the appendix 

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


Draggu (approved):

docs/src/testing/storage-cheatcodes.md
```diff
@@ -0,0 +1,164 @@
...
+        1,
+    );
+    
+    assert(loaded.len() == 1, 'Expected 1 felt loaded');
+    assert(*loaded.at(0) == 321, 'Expected 321 value loaded');
+}
+```
+
+## Example: Complex structures in storage
+This example uses a complex key and value, with `serde` as default serialization method.
```
piotmag769, line 66:
```suggestion
This example uses a complex key and value, with `Serde` as default serialization method.
```

docs/src/testing/storage-cheatcodes.md
```diff
@@ -0,0 +1,164 @@
...
+
+#[starknet::contract]
+mod Contract {
+    #[derive(Serde)]
+    struct MapKey {
+        a: felt252,
+        b: felt252,
+    }
+
+    // Required for indexing of complex_mapping
```
piotmag769, line 79:
```suggestion
    // Required to use the type as key in `LegacyMap`
```

docs/src/testing/storage-cheatcodes.md
```diff
@@ -0,0 +1,164 @@
...
+    // Serialization of key with `Serde` (other methods are possible)
+    impl MapKeyIntoSpan of Into<MapKey, Span<felt252>> {
+        fn into(self: MapKey) -> Span<felt252> {
+            let mut serialized_struct: Array<felt252> = array![];
+            self.serialize(ref serialized_struct);
+            serialized_struct.span()
+        }
+    }
+    
+     // Serialization of value with `Serde` (other methods are possible)
```
piotmag769, line 102: The description may be misleading - it seems like this implementation somehow affects the `Store` trait while it is just to make using cheatcodes easier. We should state the purpose of the code

docs/src/testing/storage-cheatcodes.md
piotmag769: We should rewrite it so that the user knows that deriving `Store` makes the type be stored as if it would be serialised with `Serde`, this is quite confusing

piotmag769 (approved):

docs/src/testing/storage-cheatcodes.md
Arcticae: No way that is more comprehensible

piotmag769: What does "indexing" mean?

PR #2065, opened 2024-04-24
Save sierra test code separately for profiler

piotmag769:
## Introduced changes

- Test artifacts file usually had 2 structs in it: `CompiledTestCrateRaw` for `src/` and `CompiledTestCrateRaw` for `tests/`, each with a separate sierra program. This made profiler unable to determine which sierra it should use for function level profiling (only path to the artifact file was passed to profiler). Now we save the relevant sierra program to another file and pass a path to the trace. This way profiler knows what sierra it should use. It also makes profiler less coupled with snforge (it doesn't have to know forge test artifact format)

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/forge-runner/src/sierra_test_code_path.rs
```diff
@@ -0,0 +1,35 @@
...
+use std::fs::File;
+use std::io::BufWriter;
+
+pub struct SierraTestCodePath {
+    pub path: Utf8PathBuf,
+}
+
+impl SierraTestCodePath {
+    pub fn new(test_crate: &CompiledTestCrateRunnable, output_path: Utf8PathBuf) -> Result<Self> {
+        let versioned_program: VersionedProgram = test_crate.sierra_program.clone().into();
```
piotmag769, line 16: Expensive clone, but `VersionedProgram` is an enum so I cannot do much

MaksymilianDemitraszek, line 16: Imo `Path` constructor copying a file is wery weird and unexpected behaviour.

crates/forge-runner/src/sierra_test_code_path.rs
```diff
@@ -0,0 +1,35 @@
+use crate::compiled_runnable::CompiledTestCrateRunnable;
+use anyhow::Context;
+use anyhow::Result;
+use cairo_lang_sierra::program::VersionedProgram;
+use camino::Utf8PathBuf;
+use fs4::FileExt;
+use std::fs::File;
+use std::io::BufWriter;
+
+pub struct SierraTestCodePath {
```
MaksymilianDemitraszek, line 10: I think `Code` is word which we don't need, how compiler refers to this structure?

crates/forge-runner/src/sierra_test_code_path.rs
```diff
@@ -0,0 +1,35 @@
...
+    pub path: Utf8PathBuf,
```
MaksymilianDemitraszek, line 11: This shouldn't be `pub` bcs this way we cannot guarantee that this is a valid path. 
We should have a function which copies sierra and retruns the path (not path constructor imo, a path is a path), this way we would guarantee on type level that every `SierraCodePath` actually points to a sierra file.

crates/forge-runner/src/lib.rs
```diff
@@ -38,6 +40,7 @@ mod printing;
 mod running;
 
 pub const CACHE_DIR: &str = ".snfoundry_cache";
+pub const SIERRA_TEST_CODE_DIR: &str = ".snfoundry_test_code";
```
MaksymilianDemitraszek, line 43: This copied sierras should are part of the trace, this name should be in `build_trace_data` module.

MaksymilianDemitraszek (changes requested):

crates/forge-runner/src/forge_config.rs
```diff
@@ -90,3 +90,19 @@ impl<'a> RuntimeConfig<'a> {
         }
     }
 }
+
+#[derive(Debug, PartialEq)]
+pub struct SierraTestCodePathConfig {
+    pub package_name: String,
+    pub sierra_test_code_dir: Utf8PathBuf,
+}
```
MaksymilianDemitraszek, line 98: No clue what it represents

crates/forge-runner/src/forge_config.rs
```diff
@@ -90,3 +90,19 @@ impl<'a> RuntimeConfig<'a> {
...
+
+impl SierraTestCodePathConfig {
+    #[must_use]
+    pub fn path_for_sierra(&self, tests_location: &str) -> Utf8PathBuf {
```
MaksymilianDemitraszek, line 102: No clue why we need a path config and we build path from it.

MaksymilianDemitraszek (changes requested):

crates/forge-runner/src/forge_config.rs
piotmag769: PathConfig is relevant for the whole package while the path is for a specific crate (the one from `src` or `tests`)

crates/forge-runner/src/forge_config.rs
piotmag769: The configuration for sierra code path that is relevant for the whole package

crates/forge/src/lib.rs
```diff
@@ -127,19 +124,31 @@ pub async fn run(
...
 
-        let summary =
-            forge_runner::run_tests_from_crate(compiled_test_crate, forge_config, tests_filter)
-                .await?;
+        let sierra_test_code_path = Arc::new(SierraTestCodePath::new(
+            &compiled_test_crate,
+            forge_config
+                .sierra_test_code_path_config
+                .path_for_sierra(tests_location),
+        )?);
```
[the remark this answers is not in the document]
piotmag769: The file? No, it does not, profiler needs this file to read sierra from it. 

Unless you ask about why do we create the struct here, then - we have to, `TestsLocation` is a struct from `forge`, doesn't exist in `forge-runner`

crates/forge-runner/src/sierra_test_code_path.rs
piotmag769: Removed the expensive clone, now only deserialisation is expensive, but it is also conditional
@MaksymilianDemitraszek @Arcticae

piotmag769: Changed

crates/forge-runner/src/sierra_test_code_path.rs
piotmag769: Done

crates/forge-runner/src/lib.rs
piotmag769: Done

crates/forge-runner/src/sierra_test_code_path.rs
piotmag769: Changed to `Program` + changed the name a bit

crates/forge-runner/src/sierra_test_code_path.rs
```diff
@@ -0,0 +1,35 @@
...
+}
+
+impl SierraTestCodePath {
+    pub fn new(test_crate: &CompiledTestCrateRunnable, output_path: Utf8PathBuf) -> Result<Self> {
```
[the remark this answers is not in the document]
piotmag769: Changed

crates/forge/src/lib.rs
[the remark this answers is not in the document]
piotmag769: Now the file is saved only when needed

crates/forge-runner/src/forge_config.rs
piotmag769: Removed

crates/forge-runner/src/forge_config.rs
piotmag769: Removed

crates/forge/src/lib.rs
```diff
@@ -127,19 +127,29 @@ pub async fn run(
...
-            compiled_test_crate.tests_location,
-            compiled_test_crate.test_cases.len(),
+            compiled_test_crate_raw.tests_location,
+            compiled_test_crate_raw.test_cases.len(),
+        );
+
+        let maybe_test_program_path = Arc::new(
+            test_program_paths_map
+                .get(&compiled_test_crate_raw.tests_location)
+                .cloned(),
```
piotmag769, line 139: I prefer to pass `Option` here explicitly and then use `Default` at the end instead of hiding the fact that the path does not exist. I have to pass sth around since we always build `ProfilerCallTrace`, even when we don't need it (it was introduced long time ago by @MaksymilianDemitraszek I believe)

crates/forge-runner/src/build_trace_data.rs
```diff
@@ -95,10 +99,14 @@ fn build_cairo_execution_info(
...
     contracts_data: &ContractsData,
-    test_artifacts_path: &Utf8Path,
+    maybe_test_program_path: &Option<TestProgramPath>,
 ) -> Option<Utf8PathBuf> {
     if contract_name == TEST_CODE_CONTRACT_NAME {
-        Some(Utf8PathBuf::from(test_artifacts_path))
+        Some(
+            maybe_test_program_path
+                .clone()
+                .map_or_else(Utf8PathBuf::new, Into::into),
```
piotmag769, line 108: If the path is `None` then we won't be saving trace anyways, so it's meaningless

Arcticae: So why is this function called then? This should be refactored

crates/forge/src/lib.rs
Arcticae: This should be refactored then somehow

crates/forge-runner/src/build_trace_data.rs
Arcticae: If someone makes mistake in the code, you will potentially output an invalid trace (impossible to interpret)

Arcticae (approved):
LGTM,
please make an issue for refactoring the trace building tho

crates/forge/src/lib.rs
piotmag769: True, but not in this PR. It was prob done to avoid relying on `Option` - to make sure its typesafe we would have to rewrite the whole type system responsible for building test summaries, preferably using type families

crates/forge-runner/src/build_trace_data.rs
piotmag769: Yep, but this requires a lot more refactoring, creating an issue

crates/forge/src/lib.rs
piotmag769: https://github.com/foundry-rs/starknet-foundry/issues/2068

crates/forge-runner/src/build_trace_data/test_program_path.rs
```diff
@@ -0,0 +1,46 @@
...
+use std::fs;
+use std::fs::File;
+use std::io::BufWriter;
+
+pub const TESTS_PROGRAMS_DIR: &str = ".snfoundry_test_code";
+
+#[derive(Clone)]
+pub struct TestProgramPath(Utf8PathBuf);
+
+impl TestProgramPath {
```
MaksymilianDemitraszek, line 15: Aren't those contracts as well? Also would add `Sierra` prefix to make it clear.

Would be nice to provide a comment what it exactly is.

crates/forge-runner/src/build_trace_data/test_program_path.rs
```diff
@@ -0,0 +1,46 @@
...
+    pub fn save_test_program(
```
MaksymilianDemitraszek, line 16: This shouldn't be a function of `TestProgramPath` imo, should be a free function. It suggests that it saves program under `TestProgramPath`.

crates/forge-runner/src/build_trace_data/test_program_path.rs
```diff
@@ -0,0 +1,46 @@
...
+        versioned_program: &VersionedProgram,
+        test_program_path: Utf8PathBuf,
```
MaksymilianDemitraszek, line 18:
```suggestion
        test_program_path: Utf8PathBuf,
```
Please name it differently, `source_test_program_path` maybe?

crates/forge/src/main.rs
```diff
@@ -240,6 +241,34 @@ fn warn_if_snforge_std_not_compatible(scarb_metadata: &Metadata) -> Result<()> {
...
+            "{package_name}_{:?}.sierra.json",
+            test_crate.tests_location
+        ));
+
+        if save_test_sierra_program {
+            let test_program_path =
+                TestProgramPath::save_test_program(&test_crate.sierra_program, test_program_path)?;
+            map.insert(test_crate.tests_location, test_program_path.clone());
+        }
+    }
```
MaksymilianDemitraszek, line 267: Please make it clearer what is the condition of saving the file. I would move `execution_data_to_save` as the first argument. Also I would yield `if save_test_sierra_program {` above the loop and create a separate, non-conditional saving function.

crates/forge/src/main.rs
```diff
@@ -240,6 +241,34 @@ fn warn_if_snforge_std_not_compatible(scarb_metadata: &Metadata) -> Result<()> {
     Ok(())
 }
 
+fn maybe_save_tests_sierra_programs(
```
MaksymilianDemitraszek, line 244: Why here, not in `build_trace_data`?

MaksymilianDemitraszek (changes requested):

crates/forge-runner/src/build_trace_data/test_program_path.rs
piotmag769: No, they are not contracts, I don't see how they would be

crates/forge-runner/src/build_trace_data/test_program_path.rs
piotmag769: I don't see how the word `source` is beneficial here, but okay

crates/forge/src/main.rs
piotmag769: It's too late to do it in `build_trace_data`, you don't have `CrateLocation` there + I would have to pass around this weird path to test artifacts

crates/forge-runner/src/build_trace_data/test_program_path.rs
piotmag769: Can change it, wanted it to be in one function to guarantee that the path is valid

crates/forge/src/main.rs
piotmag769: Moved the `if` above, imo creating a non conditional function is not that beneficial

crates/forge-runner/src/build_trace_data/test_program_path.rs
piotmag769: Refactored it a bit so the constructor also generates the path, guaranteeing its uniqueness

crates/forge-runner/src/build_trace_data/test_program_path.rs
piotmag769: Changed the logic here

crates/forge-runner/src/build_trace_data/test_program_path.rs
piotmag769: Changed the logi chere

crates/forge-runner/src/build_trace_data/test_program_path.rs
MaksymilianDemitraszek: I think we generally distinct those in ohter parts of the codebase, would be nice to do it here as well.

crates/forge/src/main.rs
MaksymilianDemitraszek: Can't we just put those into config? I think it is much less confusing for the reader than some weird saving procedure without context.

crates/forge-runner/src/build_trace_data/test_sierra_program_path.rs
```diff
@@ -0,0 +1,55 @@
...
+
+/// A path to a file with deserialized [`VersionedProgram`] that comes
+/// from compiled test crate. Needed to provide path to source sierra in
+/// [`trace_data::CairoExecutionInfo`].
+#[derive(Clone)]
+pub struct TestSierraProgramPath(Utf8PathBuf);
+
+impl TestSierraProgramPath {
+    pub fn save_sierra_test_program_from_test_crate(
+        versioned_program_from_crate: &VersionedProgram,
```
MaksymilianDemitraszek, line 20:
```suggestion
        versioned_program: &VersionedProgram,
```
This is not the part of this function logic

crates/forge-runner/src/build_trace_data/test_program_path.rs
piotmag769: I don't understand what you mean here, it is distinct from contracts

crates/forge-runner/src/build_trace_data/test_sierra_program_path.rs
```diff
@@ -0,0 +1,55 @@
...
+        crate_location: &str,
```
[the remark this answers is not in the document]
MaksymilianDemitraszek: We should convert it to Enum

crates/forge/src/main.rs
```diff
@@ -240,6 +243,35 @@ fn warn_if_snforge_std_not_compatible(scarb_metadata: &Metadata) -> Result<()> {
...
+    execution_data_to_save: ExecutionDataToSave,
+    compiled_test_crates: &[CompiledTestCrateRaw],
+    tests_programs_dir: &Utf8Path,
+    package_name: &str,
+) -> Result<HashMap<CrateLocation, TestSierraProgramPath>> {
```
MaksymilianDemitraszek, line 251: It should be a result Option.

crates/forge/src/main.rs
```diff
@@ -240,6 +243,35 @@ fn warn_if_snforge_std_not_compatible(scarb_metadata: &Metadata) -> Result<()> {
...
+        ExecutionDataToSave::None => false,
+    };
+    let mut map = HashMap::new();
+
+    if save_test_sierra_program {
+        for test_crate in compiled_test_crates {
+            let test_sierra_program_path =
+                TestSierraProgramPath::save_sierra_test_program_from_test_crate(
+                    &test_crate.sierra_program,
+                    &format!("{:?}", test_crate.tests_location),
```
MaksymilianDemitraszek, line 263: We should add a field `tests_location` to `RunnableTestCrate`

crates/forge/src/main.rs
MaksymilianDemitraszek: We should run it from `maybe_save_execution_data` and add test_location to `TestCrateRunnable`

crates/forge-runner/src/build_trace_data/test_sierra_program_path.rs
```diff
@@ -0,0 +1,55 @@
...
+pub const TESTS_PROGRAMS_DIR: &str = ".snfoundry_test_code";
+
+/// A path to a file with deserialized [`VersionedProgram`] that comes
+/// from compiled test crate. Needed to provide path to source sierra in
+/// [`trace_data::CairoExecutionInfo`].
+#[derive(Clone)]
+pub struct TestSierraProgramPath(Utf8PathBuf);
+
+impl TestSierraProgramPath {
+    pub fn save_sierra_test_program_from_test_crate(
```
MaksymilianDemitraszek, line 19:
```suggestion
    pub fn save_versioned_program(
```

crates/forge-runner/src/lib.rs
```diff
@@ -75,20 +77,27 @@ pub async fn run_tests_from_crate(
     tests: CompiledTestCrateRunnable,
     forge_config: Arc<ForgeConfig>,
     tests_filter: &impl TestCaseFilter,
+    package_name: &str,
```
piotmag769, line 80: Idk if it should be put in `ForgeConfig` maybe?

MaksymilianDemitraszek (approved):

Issue #2040, opened 2024-04-18
Implement `cheatcode<'cheat_execution_info'>`


MaksymilianDemitraszek:
Should be implemented as described in the relevant design doc

PR #2067, opened 2024-04-24
Add logic of `cheat_execution_info`

Draggu:
Related #2040

## Introduced changes

- Add logic of `cheat_execution_info`

Next PR (#2080) make current cheatcodes use it internally

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/cheatnet/src/runtime_extensions/deprecated_cheatable_starknet_extension/mod.rs
Arcticae, line 1: Are those changes relevant?

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/cheatcodes/cheat_execution_info.rs
```diff
@@ -0,0 +1,189 @@
...
+pub struct ExecutionInfoMock {
+    pub block_info: BlockInfoMock,
+    pub tx_info: TxInfoMock,
+    pub caller_address: CheatStatus<ContractAddress>,
+    pub contract_address: CheatStatus<ContractAddress>,
+    pub entry_point_selector: CheatStatus<EntryPointSelector>,
+}
+
+#[derive(Clone, Default, Debug)]
+pub struct TxInfoDto {
```
Arcticae, line 73: DTO doesn't say really what it's for

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/cheatcodes/cheat_execution_info.rs
```diff
@@ -0,0 +1,189 @@
...
+#[derive(Clone, Default, Debug)]
+pub struct ExecutionInfoDto {
+    pub block_info: BlockInfoDto,
+    pub tx_info: TxInfoDto,
+    pub caller_address: Operation<ContractAddress>,
+    pub contract_address: Operation<ContractAddress>,
+    pub entry_point_selector: Operation<EntryPointSelector>,
+}
+
+macro_rules! for_all_fields {
```
Arcticae, line 105: Wouldn't keeping those things in a map, and having simple functions (ops) be simpler in terms of code readability?

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/cheatcodes/cheat_execution_info.rs
Draggu: I don't understand. It will only be used only in communication `cairo -> forge` so DTO says exactly what it's for

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/cheatcodes/cheat_execution_info.rs
Draggu: They have different types (not only different felt types) so there would be need for enum of all them and then some kind of unwraping. Sadly it would not be more readable

crates/cheatnet/src/runtime_extensions/deprecated_cheatable_starknet_extension/mod.rs
Draggu: Related only. This was done very poorly in terms of type system. If you prefer can be done in separate PR

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/cheatcodes/cheat_execution_info.rs
Arcticae: It's not clear for the reader that it's the purpose of this object. I would name it differently, something like `TxInfoMockOperations` or something like that. New terminology here is more confusing than reuse of existing terms.

crates/cheatnet/src/runtime_extensions/deprecated_cheatable_starknet_extension/mod.rs
Arcticae: For the future - please do, you can resolve this one

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/cheatcodes/cheat_execution_info.rs
Arcticae: Meh i kind of dislike this macrology for this kind of stuff since adding or modifying of the code is harder
I can live with that though, but would wait for another opinion on that

crates/cheatnet/src/state.rs
```diff
@@ -267,6 +276,9 @@ pub struct TraceData {
 }
 
 pub struct CheatnetState {
+    pub cheated_execution_info_contracts: HashMap<ContractAddress, ExecutionInfoMock>,
```
drknzz, line 279: So CheatStatus is not needed here?

crates/cheatnet/src/state.rs
```diff
@@ -148,6 +150,13 @@ impl<T> CheatStatus<T> {
             }
         }
     }
+
+    pub fn as_option(self) -> Option<(T, CheatSpan)> {
+        match self {
+            Self::Cheated(value, span) => Some((value, span)),
+            Self::Uncheated => None,
+        }
+    }
```
drknzz, line 159: Is this used anywhere?

crates/cheatnet/src/runtime_extensions/deprecated_cheatable_starknet_extension/mod.rs
```diff
@@ -73,12 +73,17 @@ impl<'a> DeprecatedExtensionLogic for DeprecatedCheatableStarknetRuntimeExtensio
...
+                if let Some(caller_address) = self
+                    .cheatnet_state
+                    .get_cheated_caller_address(&contract_address)
+                {
                     // Increment, since the selector was peeked into before
                     syscall_handler.syscall_ptr += 1;
                     increment_syscall_count(syscall_handler, selector);
 
-                    let response = get_caller_address(self, contract_address);
+                    let response = GetContractAddressResponse {
```
drknzz, line 84: Is there no GetCallerAddressResponse or sth like that?

crates/cheatnet/src/runtime_extensions/deprecated_cheatable_starknet_extension/mod.rs
```diff
@@ -113,12 +128,18 @@ impl<'a> DeprecatedExtensionLogic for DeprecatedCheatableStarknetRuntimeExtensio
...
                     syscall_handler.syscall_ptr += 1;
                     increment_syscall_count(syscall_handler, selector);
 
-                    let response =
-                        get_sequencer_address(self, syscall_handler, contract_address).unwrap();
+                    syscall_handler.verify_not_in_validate_mode("get_sequencer_address")?;
+
+                    let response = GetContractAddressResponse {
+                        address: sequencer_address,
+                    };
```
drknzz, line 142: Same here

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/cheatcodes/cheat_execution_info.rs
```diff
@@ -0,0 +1,190 @@
...
+
+                        cheated_info.$($path).+ = CheatStatus::Uncheated;
+                    }
+                    Operation::StartGlobal(value) => {
+                        self.global_cheated_execution_info.$($path).+ =
+                            CheatStatus::Cheated(value.clone(), CheatSpan::Indefinite);
+
+                        for val in self.cheated_execution_info_contracts.values_mut() {
+                            val.$($path).+ = CheatStatus::Cheated(value.clone(), CheatSpan::Indefinite);
+                        }
```
drknzz, line 167: That's kinda expensive, but I guess the logic is simpler

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/cheatcodes/cheat_execution_info.rs
drknzz: I'd say it's readable enough right now

crates/cheatnet/src/state.rs
Draggu: `CheatStatus` is used in `ExecutionInfoMock`

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/cheatcodes/cheat_execution_info.rs
Draggu: Actually, both `.clone()` and `HashMap` iteration are not expensive here. The cloned value is the value of one of the fields listed in the `for_all_fields!` macro (`Vec` in the worst case, `Felt252` commonly). `self.cheated_execution_info_contracts` will contain only contracts that have previously been cheated. There should not be a lot of them, as this is per test.

Arcticae (approved):

drknzz (approved):

PR #2075, opened 2024-04-25
Rework `ResourceBounds` struct

Draggu:
## Introduced changes

- Rework `ResourceBounds` to make it more readable

### DIFF
--- a/crates/cheatnet/src/runtime_extensions/forge_runtime_extension/cheatcodes/spoof.rs
+++ b/crates/cheatnet/src/runtime_extensions/forge_runtime_extension/cheatcodes/spoof.rs
@@ -1,32 +1,13 @@
 use crate::state::{start_cheat, stop_cheat, CheatSpan, CheatTarget};
 use crate::CheatnetState;
 use cairo_felt::Felt252;
-use num_traits::ToPrimitive;
-use runtime::utils::buffer_reader::{BufferReadError, BufferReadResult, BufferReader};
-use runtime::utils::from_reader::FromReader;
 use runtime::FromReader;
-use std::ops::Deref;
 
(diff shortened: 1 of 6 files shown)


crates/cheatnet/src/runtime_extensions/forge_runtime_extension/cheatcodes/spoof.rs
```diff
@@ -38,7 +19,7 @@ pub struct TxInfoMock {
     pub transaction_hash: Option<Felt252>,
     pub chain_id: Option<Felt252>,
     pub nonce: Option<Felt252>,
-    pub resource_bounds: Option<ResourceBounds>,
+    pub resource_bounds: Option<Vec<ResourceBounds>>,
```
Arcticae, line 22: how come this is now vec and we don't need to change anything in cairo? 🤔

Draggu: Previously it was `ResourceBounds(pub Vec<Felt252>)` now it's `<Vec<ResourceBounds>>` where 
```rs
struct ResourceBounds {
    pub resource: Felt252,
    pub max_amount: u64,
    pub max_price_per_unit: u128,
}
```

Previously was this way to keep deserialising logic. Current is semantically correct

piotmag769: Nothing changed in cairo, we just changed the interface in Rust (btw we aligned it with the cairo one). It still serializes to felts the same way

piotmag769 (approved):
Rly gj with the changes

PR #2092, opened 2024-05-06
chore: fix typo

xiaoxianBoy:
fix a typo `verifiation` -> `verification`


Arcticae:
Those files are for internal usage, so it's not needed to update those retrospectively after they've been merged

PR #2100, opened 2024-05-09
Enable creating argent accounts

ddoktorski:
Towards #1855 

## Introduced changes

- Enable creating an argent account to the accounts file and starkli format

## Checklist

- [X] Linked relevant issue
- [X] Added relevant tests


crates/sncast/src/starknet_commands/account/create.rs
```diff
@@ -127,15 +142,30 @@ async fn generate_account(
...
+            let factory = ArgentAccountFactory::new(
+                class_hash,
+                chain_id,
+                FieldElement::ZERO,
+                signer,
+                provider,
+            )
+            .await?;
+            get_address_and_deployment_fee(factory, salt).await?
+        }
```
THenry14, line 167: what happens if someone tries to create braavos atm? 
I suppose we should add a new case in here that will gather all the other possible AccountTypes and return some error

THenry14: probably applicable to other, similar match statements

docs/src/appendix/sncast/account/create.md
```diff
@@ -13,6 +13,11 @@ Required.
 
 Account name under which account information is going to be saved.
 
+## `--type, -t <ACCOUNT_TYPE>`
+Optional.
```
THenry14, line 17: this might have to be changed depending on where we land with type flag for custom class hashes

crates/sncast/src/starknet_commands/account/create.rs
```diff
@@ -7,22 +7,29 @@ use camino::Utf8PathBuf;
...
-use starknet::core::utils::get_contract_address;
 use starknet::providers::jsonrpc::HttpTransport;
 use starknet::providers::JsonRpcClient;
 use starknet::signers::{LocalWallet, SigningKey};
 
 #[derive(Args, Debug)]
 #[command(about = "Create an account with all important secrets")]
 pub struct Create {
+    /// Type of the account
+    #[clap(value_enum, short = 't', long = "type", default_value_t = AccountType::Oz)]
```
THenry14, line 30: I think this should be required for custom class hashes, otherwise we have no way to know the account type.

crates/sncast/src/starknet_commands/account/create.rs
ddoktorski: > what happens if someone tries to create braavos atm?

It is not possible to pass `--type braavos` because it's an enum, and the only permitted values are `oz` and `argent` atm. If someone attempts to pass a braavos class hash with one of these allowable types, it would result in a failure during deployment.

THenry14: ok, so the clap handles it then, that's great :+1:

docs/src/appendix/sncast/account/create.md
THenry14: yes, let's do that :+1:

Draggu (approved):

PR #2113, opened 2024-05-14
Update forge template

Draggu:
Related https://github.com/software-mansion/scarb/pull/1308

## Introduced changes

- Add test script to template
- Run `scarb new` only if not run by scarb (see https://github.com/software-mansion/scarb/pull/1308)

### DIFF
--- a/crates/forge/src/init.rs
+++ b/crates/forge/src/init.rs
@@ -107,6 +107,8 @@ pub fn run(project_name: &str) -> Result<()> {
 
     let scarb = std::env::var("SCARB").ok();
 
+    // this check if command is run by scarb itself by checking if SCARB env exists
+    // this way we will not run `scarb new` again when `scarb new/init --snforge` is used
     if scarb.is_none() {
         ScarbCommand::new_with_stdio()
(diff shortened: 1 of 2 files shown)


Arcticae (approved):

crates/forge/src/init.rs
```diff
@@ -81,19 +89,39 @@ fn extend_gitignore(path: &Path) -> Result<()> {
...
+    // this check if command is run by scarb itself by checking if SCARB env exists
+    // this way we will not run `scarb new` again when `scarb new/init --snforge` is used
+    if scarb.is_none() {
+        ScarbCommand::new_with_stdio()
+            .current_dir(std::env::current_dir().context("Failed to get current directory")?)
+            .arg("new")
+            .arg(&project_path)
+            .run()
+            .context("Failed to initialize a new project")?;
+    }
```
maciektr, line 119: I am not really sure if we can rely on this? 🤔 
What if someone just sets the env var manually? 
The purpose of this env var is rather general - choose Scarb binary to use with Forge (whatever the usage may be). 
CC @Draggu @Arcticae

Draggu: #2121

PR #2151, opened 2024-05-21
Replace `map_string_errors` with a utility function

### LINKED ISSUE #2081: Rethink the `map_error_to_string` interface

snforge

### Description

It seems to be not as intuitive and easy to use as we first thought, we should think about introducing a helper function instead, which parses the felt array to bytearray and returns a result (which is a failure if it's unparseable).

Arcticae:
Closes #2081 
## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


snforge_std/src/byte_array.cairo
```diff
@@ -1,7 +1,21 @@
...
+
 fn byte_array_as_felt_array(self: @ByteArray) -> Array<felt252> {
     let mut serialized = array![];
 
     self.serialize(ref serialized);
 
     serialized
 }
+
+fn try_deserialize_bytearray_error(x: Array<felt252>) -> Result<ByteArray, ()> {
```
Draggu, line 11:
```suggestion
fn try_deserialize_bytearray(x: Span<felt252>) -> Result<ByteArray, ()> {
```

`_error` suffix does not tell us enything.

This function does not use `Array`, but `Span`

snforge_std/src/byte_array.cairo
```diff
@@ -1,7 +1,21 @@
...
+    if x.len() > 0 && *x.at(0) == BYTE_ARRAY_MAGIC {
+        let mut x_span = x.span().slice(1, x.len() - 1);
+        let deserialized = match Serde::<ByteArray>::deserialize(ref x_span) {
+            Option::Some(str) => str,
+            Option::None => panic!("panic string not deserializable")
```
Draggu, line 16: Panic in function returning `Result` is unexpected

Arcticae: This should never happen, because it means that something is seriously wrong with the serialization/deserialization itself
Do you think it should still be handleable? I mean it can happen if a user calls it on a random array, but seems pretty good idea to crash then since it's usage error

snforge_std/src/byte_array.cairo
Arcticae: It does not do `Serde::deserialize<ByteArray>()` - it also verifies if the *error* (emitted with `panic!` or `assert!`) string is also prepended with the magic byte which distinguishes it from other types of errors. That's why i added the `error` suffix, since it's used purely for that - the `ByteArray` deserialization itself does not include the magic byte verification.

Arcticae: With the input type i completely agree, missed that

snforge_std/src/byte_array.cairo
Draggu: This function is literraly called `try_deserialize_...`. Why function trying to deserialize something and returning `Result` should panic when deserialization fails?

snforge_std/src/byte_array.cairo
Draggu: > It does not do `Serde::deserialize<ByteArray>()` - it also verifies if the _error_ (emitted with `panic!` or `assert!`) string is also prepended with the magic byte which distinguishes it from other types of errors. That's why i added the `error` suffix, since it's used purely for that - the `ByteArray` deserialization itself does not include the magic byte verification.

Ok, I missed this. Agree here

snforge_std/src/byte_array.cairo
```diff
@@ -1,7 +1,27 @@
...
+/// Returns the parsed `ByteArray`, or an `Err` if the parsing failed.
+fn try_deserialize_bytearray_error(x: Span<felt252>) -> Result<ByteArray, ByteArray> {
+    if x.len() > 0 && *x.at(0) == BYTE_ARRAY_MAGIC {
+        let mut x_span = x.slice(1, x.len() - 1);
+        let deserialization = Serde::<ByteArray>::deserialize(ref x_span);
+            if deserialization.is_none() {
+                return Result::Err("Malformed input provided");
+            }
+            
+            return Result::Ok(deserialization.unwrap());
```
Draggu: `match` on `deserialization` not `.unwrap()`

snforge_std/src/byte_array.cairo
```diff
@@ -1,7 +1,27 @@
...
+        let mut x_span = x.slice(1, x.len() - 1);
+        let deserialization = Serde::<ByteArray>::deserialize(ref x_span);
+            if deserialization.is_none() {
+                return Result::Err("Malformed input provided");
+            }
+            
+            return Result::Ok(deserialization.unwrap());    
+        }
+    
+    Result::Err("Input is not a ByteArray-formatted error")
```
Draggu, line 26: Custom `enum` for error type would be better i think. It is more discoverable and easier to document and `match` on. it would also be consistent with other functions

Arcticae: Overkill imho, it's not something that you want to catch really. We just want to provide some context here.

snforge_std/src/byte_array.cairo
Arcticae: Done

Draggu (approved):

PR #2164, opened 2024-05-24
Use `from_serializable` and `build_readable_text `

Draggu:
## Introduced changes

- Use `from_serializable` instead of `Handled`
- Use `build_readable_text` from `shared` instead of copy-pasted logic


Arcticae:
What is this PR needed for?

Draggu:
> What is this PR needed for?

Just some minor improvments. Nothing important

PR #2167, opened 2024-05-28
Update docs since specifying account file path is not an option anymore

FredCoen:
The docs indicate that one can specify a path to the accounts file via the flag --accounts-file.

But that does not seem to work. Should i also make updates to the code anywhere? I am happy to dive into the code base and make the changes if you guide me a little. Recently my company moved to the starknet ecosystem so i am happy to contribute here!

I do think however on this issue its just an error message and the docs that seem to be needed correction.

Thanks

### DIFF
--- a/crates/sncast/src/lib.rs
+++ b/crates/sncast/src/lib.rs
@@ -582,8 +582,7 @@ pub fn raise_if_empty(value: &str, value_name: &str) -> Result<()> {
 
 pub fn check_account_file_exists(accounts_file_path: &Utf8PathBuf) -> Result<()> {
     if !accounts_file_path.exists() {
-        bail! {"Accounts file = {} does not exist! If you do not have an account create one with `account create` command \
-        or if you're using a custom accounts file, make sure to supply correct path to it with `--accounts-file` argument.", accounts_file_path}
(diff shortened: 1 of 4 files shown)


Draggu:
Hi @FredCoen 

Thanks for your contribution! But `--account-file` flag seems to work as expected with substantial tests covering this area.

Could you provide steps to reproduce the issue you've found? It'd help us understand any potential problems better.

FredCoen:
sure, i am using sncast  version 0.24.0

when i run the command 

`sncast --url $(SEPOLIA_RPC) account create --name $(name) --account-file account-file`

i get 
`error: unexpected argument '--account-file' found`

Also, when i run 
`sncast account create --help`

i dont get the option of "account-file"

Draggu:
> sure, i am using sncast version 0.24.0
> 
> when i run the command
> 
> `sncast --url $(SEPOLIA_RPC) account create --name $(name) --account-file account-file`
> 
> i get `error: unexpected argument '--account-file' found`
> 
> Also, when i run `sncast account create --help`
> 
> i dont get the option of "account-file"

`--account-file` is a common flag https://foundry-rs.github.io/starknet-foundry/appendix/sncast/common.html so it should be used before subcommand (`account create` in this case)

FredCoen:
ah ok sry thanks @Draggu

PR #2178, opened 2024-06-06
Revert env changes

### LINKED ISSUE #2172: Revert CI changes related to `.env` and secrets

snforge

### Description

It's impossible right now to run the actions from the contributor side, which is a burden to explain. Properly solving it would require too much effort for the payoff.

Arcticae:
Closes #2172

## Introduced changes

- Reverts the changes that were needed to extract testing RPC URL from envs

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests

### DIFF
--- a/crates/shared/src/test_utils/node_url.rs
+++ b/crates/shared/src/test_utils/node_url.rs
@@ -1,5 +1,3 @@
-use anyhow::{Context, Result};
-use std::env;
 use url::Url;
 
 /// Loads node RPC URL from `SEPOLIA_RPC_URL` environmental variable set in dotenv.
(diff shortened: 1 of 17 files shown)


Draggu:
Im not sure if we want to revert all changes made there. Keeping url in one place is good idea. Previously we had it copy pasted all over making troubles when bumping rpc version. We can keep this changes while reverting env related ones only

Arcticae:
@Draggu yeah probably that's a good point, was a bit lazy here :P

crates/shared/src/test_utils/node_url.rs
```diff
@@ -8,14 +8,8 @@ use url::Url;
...
-    dotenv::dotenv().ok();
-    let node_rpc_url = env::var("SEPOLIA_RPC_URL")
-        .context("The required environmental variable `SEPOLIA_RPC_URL` is not set. Please set it manually or in .env file"
-    )?;
-
-    Url::parse(&node_rpc_url).with_context(|| {
-        format!("Failed to parse the URL from the `SEPOLIA_RPC_URL` environmental variable: {node_rpc_url}")
-    })
+    Url::parse("http://188.34.188.184:7070/rpc/v0_7")
+        .with_context(|| "Failed to parse the sepolia RPC URL")
```
Draggu, line 12: `.unwrap()` here would be better, it should not fail

Arcticae: Hmm i think it's true but i prefer to have that in case somebody changes the URL - it's easier to debug then if the URL is incorrect

Arcticae: I can unwrap it probably here though after the context

PR #2193, opened 2024-06-12
Refactor `snforge_std` events part

tomek0123456789:
commit-id:d9896c8f

---

**Stack**:
- #2224
- #2195
- #2194
- #2193 ⬅

### DIFF
--- a/snforge_std/src/cheatcodes/events.cairo
+++ b/snforge_std/src/cheatcodes/events.cairo
@@ -4,22 +4,12 @@ use starknet::testing::cheatcode;
 use starknet::ContractAddress;
 use super::super::_cheatcode::handle_cheatcode;
 
-/// Allows specifying which contracts you want to capture events from.
-#[derive(Drop, Serde)]
-enum SpyOn {
-    All: (),
-    One: ContractAddress,
-    Multiple: Array<ContractAddress>
-}
+/// Creates `EventSpy` instance that spies on all events emitted after its creation.
+fn spy_events() -> EventSpy {
+    let mut event_offset = handle_cheatcode(cheatcode::<'spy_events'>(array![].span()));
+    let parsed_event_offset: usize = Serde::<usize>::deserialize(ref event_offset).unwrap();
 
-/// Creates `EventSpy` instance which spies on events emitted by contracts defined under the
-/// `spy_on` argument.
-fn spy_events(spy_on: SpyOn) -> EventSpy {
-    let mut inputs = array![];
-    spy_on.serialize(ref inputs);
-    let output = handle_cheatcode(cheatcode::<'spy_events'>(inputs.span()));
-
-    EventSpy { _id: *output[0], events: array![] }
+    EventSpy { _event_offset: parsed_event_offset }
 }
 
 /// Raw event format (as seen via the RPC-API), can be used for asserting the emitted events.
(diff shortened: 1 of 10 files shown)


snforge_std/src/cheatcodes/events.cairo
```diff
@@ -4,22 +4,11 @@ use starknet::testing::cheatcode;
...
-/// Allows specifying which contracts you want to capture events from.
-#[derive(Drop, Serde)]
-enum SpyOn {
-    All: (),
-    One: ContractAddress,
-    Multiple: Array<ContractAddress>
-}
+/// Creates `EventSpy` instance which spies on all events that will be emitted after its creation.
+fn spy_events() -> EventSpy {
+    let event_offset: usize = (*handle_cheatcode(cheatcode::<'spy_events'>(array![].span())).at(0)).try_into().unwrap();
```
Draggu: Please use `Serde` here. This is more convenient way

snforge_std/src/cheatcodes/events.cairo
```diff
@@ -29,50 +18,71 @@ struct Event {
...
 struct EventSpy {
-    _id: felt252,
-    events: Array<(ContractAddress, Event)>,
+    _event_offset: usize
+}
+
+/// A wrapper structure on an array of events to handle filtering smoothly.
+#[derive(Drop, Serde)]
+struct Events {
+    events: Array<(ContractAddress, Event)>
```
tomek0123456789, line 31: Might possibly change that into `Array<Event>` and change `Event` into:
```
#[derive(Drop, Clone, Serde)]
struct Event {
    emitted_by: ContractAddress,
    keys: Array<felt252>,
    data: Array<felt252>
}
```
but then we'd have to change `assert_...` methods so that the interface would be the same (they would have to accept `events: @Array<Event>` instead of `events: @Array<(ContractAddress, T)>`). 

Then we'd have the problem with the fact that `assert_...` methods accept `T, impl TEvent: starknet::Event<T>, impl TDrop: Drop<T>` - it doesn't ensure that the event we want to assert against would have the `emitted_by` field inside - we'd need another method on our `Event` struct like:
```
fn from<T, impl TEvent: starknet::Event<T>, impl TDrop: Drop<T>>(starknet_event: T, contract_address: ContractAddress) -> Event {
    // conversion from starknet::Event to our Event
}
```

snforge_std/src/cheatcodes/events.cairo
```diff
@@ -29,50 +18,71 @@ struct Event {
     data: Array<felt252>
 }
 
-
-/// An event spy structure, along with the events collected so far in the test.
-/// `events` are mutable and can be updated with `fetch_events`.
+/// An event spy structure containing an offset marking from which event given Spy
+/// should fetch events.
```
Draggu, line 24:
```suggestion
/// An event spy structure allowing to only get events emitted after it was created.
```

Offset is in fact implementation detail for the point in time from which we want to spy

snforge_std/src/cheatcodes/events.cairo
```diff
@@ -29,50 +20,73 @@ struct Event {
...
+#[derive(Drop, Serde)]
+struct Events {
+    events: Array<(ContractAddress, Event)>
 }
 
-trait EventFetcher {
-    /// Allows to update the structs' events field, from the spied contracts
-    fn fetch_events(ref self: EventSpy);
+trait EventSpyTrait {
+    /// Fetches all events accounting for the offset given [`EventSpy`] has.
```
Draggu, line 37: Method change from `fetch` to `get` while comment to `fetch`. Keep it consistent

snforge_std/src/cheatcodes/events.cairo
```diff
@@ -4,22 +4,13 @@ use starknet::testing::cheatcode;
...
-/// Allows specifying which contracts you want to capture events from.
-#[derive(Drop, Serde)]
-enum SpyOn {
-    All: (),
-    One: ContractAddress,
-    Multiple: Array<ContractAddress>
-}
+/// Creates `EventSpy` instance which spies on all events that will be emitted after its creation.
+fn spy_events() -> EventSpy {
+    let event_offset: usize = (*handle_cheatcode(cheatcode::<'spy_events'>(array![].span())).at(0))
```
Draggu, line 9: Please use `Serde`. It is more convenient way of doing this

snforge_std/src/cheatcodes/events.cairo
Arcticae: Do you think there is an advantage to be gained?

Arcticae: I kind of struggle to find one, it's strictly preferential, is it?

snforge_std/src/cheatcodes/events.cairo
Arcticae: The [``] reference doesn't do anything as of now

Arcticae: I also think the explanation is convoluted and too technical

snforge_std/src/cheatcodes/events.cairo
```diff
@@ -29,50 +20,73 @@ struct Event {
...
-        }
+        Events { events }
     }
 }
 
-/// Allows to assert the expected events emission (or lack thereof), in the scope of the spy
-trait EventAssertions<T, impl TEvent: starknet::Event<T>, impl TDrop: Drop<T>> {
-    fn assert_emitted(ref self: EventSpy, events: @Array<(ContractAddress, T)>);
-    fn assert_not_emitted(ref self: EventSpy, events: @Array<(ContractAddress, T)>);
+/// A trait on [`Events`] struct, which handles filtering.
```
Arcticae, line 52:
```suggestion
```
it doesn't really need it

snforge_std/src/cheatcodes/events.cairo
```diff
@@ -29,50 +20,73 @@ struct Event {
...
+    fn get_events(ref self: EventSpy) -> Events;
 }
 
-impl EventFetcherImpl of EventFetcher {
-    fn fetch_events(ref self: EventSpy) {
-        let mut output = handle_cheatcode(cheatcode::<'fetch_events'>(array![self._id].span()));
+impl EventSpyTraitImpl of EventSpyTrait {
+    fn get_events(ref self: EventSpy) -> Events {
+        let mut output = handle_cheatcode(
+            cheatcode::<'get_events'>(array![self._event_offset.into()].span())
```
Arcticae, line 44: Feels like a waste to always fetch those and not cache them somehow 🤔

Arcticae: It's an idea for the future i think though

snforge_std/src/cheatcodes/events.cairo
tomek0123456789: I'm having a dilemma right now, whether we should leave `fetch_events` as is, or go with `get_events`. There's also a `get_events` function in RPC - don't know if us naming it the same is good or bad. Waiting for your opinions.

Draggu: I would go with `fetch` as we are gaining events from backend

snforge_std/src/cheatcodes/events.cairo
Arcticae: There is also an option to not flatten the original event here, but rather nest it:

```
#[derive(Drop, Clone, Serde)]
struct EmittedEvent<T: starknet::Event<T>> {
    emitted_by: ContractAddress,
    event_data: T
}
```

tomek0123456789: It is preferential, yes, though I find having one structure with all the data nicer than passing the tuple (`contract_address`, event) to `assert_...` functions

snforge_std/src/cheatcodes/events.cairo
tomek0123456789: > The [``] reference doesn't do anything as of now
I know, but I'm trying to be forward-compatible with cairo-doc

snforge_std/src/cheatcodes/events.cairo
Arcticae: I meant something that would not be exposed to the user, for performance

snforge_std/src/cheatcodes/events.cairo
Arcticae: I'm not sure it will be the format of tagging. You can probably leave it as-is tho

snforge_std/src/cheatcodes/events.cairo
tomek0123456789: Added an issue #2217

Draggu (approved):

Arcticae (approved):

PR #2233, opened 2024-06-20
Bump starknet-rs version

ksew1:
**Stack**:
- #2261
- #2260
- #2236
- #2235
- #2234
- #2233 ⬅


Cargo.toml
```diff
@@ -62,8 +62,8 @@ rayon = "1.10"
 regex = "1.10.5"
 serde = { version = "1.0.203", features = ["derive"] }
 serde_json = "1.0.117"
-starknet = { git = "https://github.com/xJonathanLEI/starknet-rs", rev = "d980869" }
-starknet-crypto = { git = "https://github.com/xJonathanLEI/starknet-rs", rev = "d980869" }
+starknet = { git = "https://github.com/xJonathanLEI/starknet-rs", rev = "56177d2" }
+starknet-crypto = { git = "https://github.com/xJonathanLEI/starknet-rs", rev = "56177d2" }
```
Arcticae, line 66: I'd rather tag it - there seems to be one recent in the repo

crates/sncast/src/helpers/braavos.rs
```diff
@@ -77,12 +80,58 @@ where
...
-        deployment: &RawAccountDeployment,
+        deployment: &RawAccountDeploymentV1,
     ) -> Result<Vec<FieldElement>, Self::SignError> {
         let tx_hash =
-            PreparedAccountDeployment::from_raw(deployment.clone(), self).transaction_hash();
+            PreparedAccountDeploymentV1::from_raw(deployment.clone(), self).transaction_hash();
+
+        let signature = self.signer.sign_hash(&tx_hash).await?;
+
+        let mut aux_data = vec![
```
Arcticae, line 92: Can you please leave a comment here (or a link to the doc + comment) why it's done like this? I know, you know but probably somebody later can be confused by this

crates/sncast/src/helpers/braavos.rs
```diff
@@ -77,12 +80,58 @@ where
         self.block_id
     }
 
-    async fn sign_deployment(
+    async fn sign_deployment_v1(
```
Arcticae, line 83: This method is exactly the same as the one below, with the deployment type only being different. Can't we merge those two?

crates/sncast/src/lib.rs
```diff
@@ -554,7 +554,8 @@ async fn get_revert_reason(
         .await
         .map_err(SNCastProviderError::from)?;
 
-    if let starknet::core::types::ExecutionResult::Reverted { reason } = receipt.execution_result()
+    if let starknet::core::types::ExecutionResult::Reverted { reason } =
+        receipt.receipt.execution_result()
```
Arcticae, line 558: isn't this wrong name now, if we have to destructure this twice?

Arcticae (commented):
I don't really understand the multitude of different versions that are used here, why are some v1, some v2 and some v3s? If it's supposed to be a bump, i think all of those should be v0 for this PR?

ksew1:
> I don't really understand the multitude of different versions that are used here, why are some v1, some v2 and some v3s? If it's supposed to be a bump, i think all of those should be v0 for this PR?

Before, they weren't v0. Look into [this](https://github.com/xJonathanLEI/starknet-rs/commit/de633f4076f1bd350e0677979c274d9f8428fbae#diff-c83605e471fe03e3c6fcc8bc8fec58c6b118092b91ad89401e071ebff40ed447R28) commit in `starknet-rs` that added v3 support. Also, [here](https://docs.starknet.io/architecture-and-concepts/network-architecture/transactions/), you can see that for `Declare` it doesn't even support `v0`, and `Invoke` never had `v2`. Unfortunately, the naming is confusing.

Cargo.toml
ksew1: The recent tag introduced breaking changes:

> The `FieldElement` type has been replaced by `Felt`.

 and the tag before doesn't have support for `v3` transactions.

I would pretty much have to change every file, and I would rather do that migration in a separate PR, not connected to this issue.

crates/sncast/src/helpers/braavos.rs
ksew1: I could not find any documentation on that, except for the comment in the code of `braavos-account-contract`. You can find the link [here](https://github.com/myBraavos/braavos-account-cairo/blob/6efdfd597bb051e99c79a512fccd14ee2523c898/src/presets/braavos_account.cairo#L104). However, if we look into the deserialize [code](https://github.com/myBraavos/braavos-account-cairo/blob/6efdfd597bb051e99c79a512fccd14ee2523c898/src/presets/braavos_account.cairo#L141), it explicitly explains why the code data needs to be structured that way.

Arcticae (approved):

crates/sncast/src/helpers/braavos.rs
```diff
@@ -45,6 +48,53 @@ where
...
+            // multisig_threshold
+            FieldElement::ZERO,
+            // withdrawal_limit_low
+            FieldElement::ZERO,
+            // fee_rate
+            FieldElement::ZERO,
+            // stark_fee_rate
+            FieldElement::ZERO,
+            // chain_id
+            self.chain_id,
```
tomek0123456789, line 84: Where does it come from? It's not listed as a parameter in the function you linked above

ksew1: You can see why it is needed [here](https://github.com/myBraavos/braavos-account-cairo/blob/6efdfd597bb051e99c79a512fccd14ee2523c898/src/presets/braavos_base_account.cairo#L79). I added this as comment. As said before there are no docs 😭

tomek0123456789: You could add a link next to that comment explaining why it is needed with that link

tomek0123456789 (approved):

ksew1:
✓ Commit merged in pull request [#2261](https://github.com/foundry-rs/starknet-foundry/pull/2261)

PR #2191, opened 2024-06-12
Forge scarb plugin backend

Draggu:
Closes #1905

## Introduced changes

- changes on forge side to execute config runs

Follow-up to #2044

## Checklist

- [x] Linked relevant issue
- [ ] Updated relevant documentation
- [x] Added relevant tests
- [x] Performed self-review of the code
- [ ] Added changes to `CHANGELOG.md`

PR #2237, opened 2024-06-20
Forge scarb plugin - forge tests fixes

Draggu:
## Introduced changes

- Fix forge tests after scarb, forge, forge macros log change

Follow-up to #2191

Requires new scarb to pass


crates/forge/tests/e2e/running.rs
```diff
@@ -854,14 +828,15 @@ fn printing_in_contracts() {
...
-        [PASS] tests::test_contract::test_cannot_increase_balance_with_zero_value [..]
+        [PASS] contract_printing_integrationtest::test_contract::test_increase_balance [..]
+        [PASS] contract_printing_integrationtest::test_contract::test_cannot_increase_balance_with_zero_value [..]
         Tests: 2 passed, 0 failed, 0 skipped, 0 ignored, 0 filtered out
         "#},
     );
 }
 
 #[test]
+#[ignore] //TODO unignore when there exists prvious version that supports new attributes
```
Arcticae, line 839: don't understand this one

crates/forge/tests/e2e/diagnostics_and_plugins.rs
Arcticae, line 1: these are getting restored sometime soon from what i understand?

crates/forge/tests/e2e/running.rs
Draggu: This requires there to be previous version, but because we introduce now incompatibility with previous one, there must exists at least 2 versions with this new system to make this test work again

crates/forge/tests/e2e/diagnostics_and_plugins.rs
Draggu: We have to release this changes and 1 more version to restore this

crates/forge/test_utils/src/runner.rs
```diff
@@ -73,7 +73,7 @@ impl Contract {
                 sierra = true
 
                 [dependencies]
-                starknet = "2.4.0"
+                starknet = "2.6.4"
```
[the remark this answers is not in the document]
Draggu: Not sure if this is REQUIRED now, but it is used to generate package used in integration tests so better keep it up to date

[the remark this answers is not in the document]
Arcticae: I mean if this wasn't bumped before, why bump it now? That's the thing i am concerned about

crates/forge/tests/e2e/running.rs
Arcticae: Ah okay. Please leave a task though for this and mark it here by the TODO

crates/forge/test_utils/src/runner.rs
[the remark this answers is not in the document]
Draggu: Its not required but it should be bumped. In fact it should be bumped earlier but it happen to work so no one noticed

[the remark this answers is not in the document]
Arcticae: In this case it's required (scarb ships with a specific version of starknet plugin)
But for future versions i think we want to avoid doing this FYI

Arcticae (approved):

PR #2286, opened 2024-07-09
Add Verifying Contracts guide

romanbarbo:
## Introduced changes

This Pull Request adds the Verifying Contracts guide for the `verify` command.

## Checklist

### DIFF
--- a/docs/src/starknet/verify.md
+++ b/docs/src/starknet/verify.md
@@ -34,10 +34,8 @@ $ sncast --url http://127.0.0.1:5050/rpc \
 You are about to submit the entire workspace's code to the third-party chosen verifier at walnut, and the code will be publicly available through walnut's APIs. Are you sure? (Y/n) Y
 
 command: verify
-message: Contract successfully verified
+message: Contract has been successfully verified. You can check the verification status at the following link: https://api.walnut.dev/v1/sn_main/classes/0x03498e7edbc5f953315db118401fe7ea1eef637f63c56b45bd54e35150929ca3
 ```
-
-After verification with Walnut verifier, the contract can be debugged using Walnut's transaction step-by-step debugger.
 
 > 📝 **Note**
 > Contract name is a part after the `mod` keyword in your contract file. It may differ from package name defined in `Scarb.toml` file.


docs/src/starknet/verify.md
```diff
@@ -0,0 +1,39 @@
...
+    --contract-name SimpleBalance \
+    --verifier walnut \
+    --network mainnet
+
+You are about to submit the entire workspace's code to the third-party chosen verifier at walnut, and the code will be publicly available through walnut's APIs. Are you sure? (Y/n) Y
+
+command: verify
+message: Contract successfully verified
+```
+
```
Arcticae, line 37: I think we should add an explanation what happens next in the flow (where to look for the status)

docs/src/starknet/verify.md
```diff
@@ -0,0 +1,39 @@
+# Verifying Contracts
+
+## Overview
+
+Starknet Foundry `sncast` supports verifying Cairo contract classes with the `sncast verify` command by submitting the source code to a selected verification provider.
```
Arcticae, line 5: This doesn't explain why you should verify and what it really means

docs/src/starknet/verify.md
```diff
@@ -0,0 +1,39 @@
...
+
+For detailed CLI description, see [verify command reference](../appendix/sncast/verify.md).
+
+> ⚠️ **Warning**
+> Please be aware that submitting the source code means it will be publicly exposed through the provider's APIs.
```
Arcticae, line 10: I think it's also worth explaining here what happens to the code on the backend (how it can be used)

Arcticae (changes requested):

barabanovro:
Thank you for the review. I have addressed the comments

docs/src/starknet/verify.md
```diff
@@ -0,0 +1,43 @@
...
+    --verifier walnut \
+    --network mainnet
+
+You are about to submit the entire workspace's code to the third-party chosen verifier at walnut, and the code will be publicly available through walnut's APIs. Are you sure? (Y/n) Y
+
+command: verify
+message: Contract successfully verified
+```
+
+After verification with Walnut verifier, the contract can be debugged using Walnut's transaction step-by-step debugger.
```
Arcticae, line 40: If this is a general section (see the title marking that you did at `Example`), this does not belong here 

Also i don't think this is relevant at all

Arcticae: It still lacks the explanation where should i look for the confirmation, the "green check mark" that i'm after when doing the verification etc.

Is there a URL that i can use?

Arcticae (changes requested):

Arcticae (approved):

Issue #2222, opened 2024-06-19
Add `account list` command to cast


Arcticae:
### Which components does the task require to be changed? (think hard pls)

sncast

### Description

In order to know what account i can use, i should be able to list the available accounts, like  i.e. `docker image ls`

PR #2296, opened 2024-07-11
Added `sncast account list` subcommand

integraledelebesgue:
Closes [#2222](https://github.com/orgs/foundry-rs/projects/3/views/2?filterQuery=assignee%3Aintegraledelebesgue&pane=issue&itemId=67984329)

## Introduced changes
- added new `list` subcommand to `sncast account`
- it is used to list all accounts available to deploy contracts from
- accounts are retrieved from `accounts-file` parameter received from CLI or read from `.toml` config 

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests

### DIFF
--- a/crates/sncast/src/lib.rs
+++ b/crates/sncast/src/lib.rs
@@ -104,21 +104,12 @@ pub struct AccountData {
 
 impl Display for AccountData {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        macro_rules! repr_or_unspecified {
-            ($item:expr) => {
-                ($item).map_or("unspecified".to_string(), |it| it.to_string())
-            };
-        }
-
-        macro_rules! hex {
-            ($item:expr) => {
-                format!("{:#x}", $item)
-            };
-        }
-
-        macro_rules! hex_or_unspecified {
-            ($item:expr) => {
-                ($item).map_or("unspecified".to_string(), |it| hex!(it))
+        macro_rules! write_some {
+            ($dest:expr, $format_spec:expr, $item:expr) => {
+                match $item {
+                    Some(it) => write!($dest, $format_spec, it),
+                    None => Ok(()),
+                }
             };
         }
 
@@ -126,24 +117,21 @@ impl Display for AccountData {
             f,
             "
             Account data:
-              private key: {}
-              public key: {}
-              address: {}
-              salt: {}
-              class hash: {}
-              deployed: {}
-              legacy: {}
-              type: {}
+              private key: {:#x}
+              public key: {:#x}
             ",
-            hex!(self.private_key),
-            hex!(self.public_key),
-            hex_or_unspecified!(self.address),
-            hex_or_unspecified!(self.salt),
-            hex_or_unspecified!(self.class_hash),
-            repr_or_unspecified!(self.deployed),
-            repr_or_unspecified!(self.legacy),
-            repr_or_unspecified!(self.account_type)
-        )
+            self.private_key,
+            self.public_key,
+        )?;
+
+        write_some!(f, "  address: {:#x}\n", self.address)?;
+        write_some!(f, "  salt: {:#x}\n", self.salt)?;
+        write_some!(f, "  class hash: {:#x}\n", self.class_hash)?;
+        write_some!(f, "  deployed: {}\n", self.deployed)?;
+        write_some!(f, "  legacy: {}\n", self.legacy)?;
+        write_some!(f, "  type: {}\n", self.account_type)?;
+
+        Ok(())
     }
 }
 
--- a/crates/sncast/src/lib.rs
+++ b/crates/sncast/src/lib.rs
@@ -2,7 +2,6 @@ use anyhow::{anyhow, bail, Context, Error, Result};
 use camino::Utf8PathBuf;
 use clap::ValueEnum;
 use helpers::constants::{KEYSTORE_PASSWORD_ENV_VAR, UDC_ADDRESS};
-use indoc::indoc;
 use rand::rngs::OsRng;
 use rand::RngCore;
 use serde::{Deserialize, Serialize};
@@ -105,15 +104,15 @@ impl Display for AccountData {
         }
 
         let repr = format!(
-            indoc! {"
-                private key: {}
-                public key: {}
-                address: {}
-                salt: {}
-                class hash: {}
-                deployed: {}
-                legacy: {}
-            "},
+            concat!(
+                "  private key: {}\n",
+                "  public key: {}\n",
+                "  address: {}\n",
+                "  salt: {}\n",
+                "  class hash: {}\n",
+                "  deployed: {}\n",
+                "  legacy: {}\n"
+            ),
             hex!(self.private_key),
             hex!(self.public_key),
             hex_or_unspecified!(self.address),
--- a/crates/sncast/src/lib.rs
+++ b/crates/sncast/src/lib.rs
@@ -2,6 +2,7 @@ use anyhow::{anyhow, bail, Context, Error, Result};
 use camino::Utf8PathBuf;
 use clap::ValueEnum;
 use helpers::constants::{KEYSTORE_PASSWORD_ENV_VAR, UDC_ADDRESS};
+use itertools::Itertools;
 use rand::rngs::OsRng;
 use rand::RngCore;
 use serde::{Deserialize, Serialize};
@@ -27,7 +28,7 @@ use starknet::{
 use crate::helpers::constants::{DEFAULT_STATE_FILE_SUFFIX, WAIT_RETRY_INTERVAL, WAIT_TIMEOUT};
 use crate::response::errors::SNCastProviderError;
 use conversions::serde::serialize::CairoSerialize;
-use indoc::writedoc;
+use indoc::formatdoc;
 use serde::de::DeserializeOwned;
 use shared::rpc::create_rpc_client;
 use starknet::accounts::{AccountFactory, AccountFactoryError};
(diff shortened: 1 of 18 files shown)


crates/sncast/src/starknet_commands/account/list.rs
Arcticae, line 1: Does the command work with a default OZ-style account file?

crates/sncast/tests/e2e/account/list.rs
```diff
@@ -0,0 +1,116 @@
...
+
+    let expected = indoc! {"
+        Available accounts:
+        Network \"alpha-sepolia\":
+        - user0:
+        Account data:
+          private key: 0x1e9038bdc68ce1d27d54205256988e85
+          public key: 0x2f91ed13f8f0f7d39b942c80bfcd3d0967809d99e0cc083606cbe59033d2b39
+          address: 0x4f5f24ceaae64434fa2bc2befd08976b51cf8f6a5d8257f7ec3616f61de263a
+          salt: unspecified
```
Arcticae, line 37: If something is unspecified i think it should be omitted

Arcticae:
We should also add some info to `--help` and to docs that this will display your private keys, this might be somewhat dangerous

crates/sncast/src/lib.rs
```diff
@@ -63,6 +64,24 @@ impl FromStr for AccountType {
...
 
+impl AccountType {
+    fn to_string_pretty(self) -> String {
+        let repr = match self {
+            AccountType::Oz => "OpenZeppelin",
+            AccountType::Argent => "Argent",
+            AccountType::Braavos => "Braavos",
+        };
+
+        repr.to_string()
```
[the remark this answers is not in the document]
integraledelebesgue: Function `to_string_pretty` operates on the `AccountType` enum itself. In fact, it exists only because of the `AccountType::Oz` variant which  isn't descriptive enough to print. In my opinion, it should be renamed to `AccountType::OpenZeppelin` for consistence with other variants.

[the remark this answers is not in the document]
Arcticae: I am strictly talking about the fallback `.to_string()` here
Though if you want to rename the enum variants, go ahead

crates/sncast/src/starknet_commands/account/list.rs
```diff
@@ -0,0 +1,61 @@
...
+    about = "List available accounts",
+    before_help = "Warning! This command exposes cryptographic information, e.g. accounts' private keys"
+)]
+pub struct List {}
+
+pub fn print_account_list(accounts_file: &Utf8PathBuf) -> anyhow::Result<()> {
+    check_account_file_exists(accounts_file)?;
+
+    let accounts_file_path = accounts_file.canonicalize()?;
+    let accounts_file_path = accounts_file_path.to_str().ok_or(anyhow!(
```
ksew1, line 22: You can use context method instead

crates/sncast/src/starknet_commands/account/list.rs
```diff
@@ -0,0 +1,61 @@
...
+use clap::Args;
+use indoc::printdoc;
+use itertools::Itertools;
+use sncast::{check_account_file_exists, read_and_parse_json_file, AccountData};
+
+#[derive(Args, Debug)]
+#[command(
+    name = "list",
+    about = "List available accounts",
+    before_help = "Warning! This command exposes cryptographic information, e.g. accounts' private keys"
```
ksew1, line 14: How about adding a flag that will determine whether to display the private key or not?

crates/sncast/src/starknet_commands/account/list.rs
```diff
@@ -0,0 +1,61 @@
...
+    let accounts_file_path = accounts_file.canonicalize()?;
+    let accounts_file_path = accounts_file_path.to_str().ok_or(anyhow!(
+        "Failed to resolve an absolute path to the accounts file"
+    ))?;
+
+    let networks: HashMap<String, HashMap<String, AccountData>> =
+        read_and_parse_json_file(accounts_file)?;
+
+    let no_networks = networks.is_empty();
+    let no_accounts = networks.values().all(|net| net.values().len() == 0);
```
ksew1, line 30: You don't need `no_networks` variable, `all` returns true for empty iterator

crates/sncast/src/starknet_commands/account/list.rs
integraledelebesgue: My code just reads a hashmap. As long as the file is decoded properly and hashmap contains all data, everything should work fine.

crates/sncast/tests/e2e/account/helpers.rs
```diff
@@ -0,0 +1,67 @@
+use indoc::indoc;
+use tempfile::{tempdir, TempDir};
+use tokio::fs::File;
+use tokio::io::AsyncWriteExt;
+
+#[must_use]
+pub async fn create_tempdir_with_empty_json(file_name: &str) -> TempDir {
```
ksew1, line 7: What about making this function sync?

crates/sncast/tests/e2e/account/helpers.rs
```diff
@@ -0,0 +1,67 @@
...
+
+    let mut file = File::create(tempdir.path().join(file_name))
+        .await
+        .expect("Could not create temporary accounts file!");
+
+    file.write_all(data.as_bytes())
+        .await
+        .expect("Could not write temporary testing accounts");
+
+    let _ = file.flush().await;
```
ksew1, line 20: You do not check whether it is an `Err` or `Ok`.

crates/sncast/tests/e2e/account/helpers.rs
```diff
@@ -0,0 +1,67 @@
...
+
+    let mut file = File::create(tempdir.path().join(file_name))
+        .await
+        .expect("Could not create temporary accounts file!");
+
+    file.write_all(json_data.as_bytes())
+        .await
+        .expect("Could not write temporary testing accounts");
+
+    let _ = file.flush().await;
```
ksew1, line 64: Same here

crates/sncast/tests/e2e/account/helpers.rs
ksew1, line 1: How about making this one function? The only difference is the data that is written.

crates/sncast/src/lib.rs
```diff
@@ -83,6 +102,39 @@ pub struct AccountData {
     pub account_type: Option<AccountType>,
 }
 
+impl Display for AccountData {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        macro_rules! write_some {
+            ($dest:expr, $format_spec:expr, $item:expr) => {
+                match $item {
+                    Some(it) => write!($dest, $format_spec, it),
```
ksew1, line 110: use writeln! instead

ksew1 (changes requested):
Rest later

crates/sncast/tests/e2e/account/helpers.rs
integraledelebesgue: It used to be async before I moved it into the common module, however I don't insist on it

crates/sncast/tests/e2e/account/helpers.rs
integraledelebesgue: I prefer to add a short message just in case the call doesn't succeed somehow, despite the rarity of such scenario

ksew1: That's what I mean, but currently, the call can fail, and the function will not panic. So add `unwrap` or `expect`

crates/sncast/tests/e2e/account/helpers.rs
ksew1: Please make it sync then

integraledelebesgue: I mean - it's not my function. I just moved an already existing one into a common module to use it widely

docs/src/appendix/sncast/account/list.md
```diff
@@ -0,0 +1,20 @@
+# `list`
+List all available accounts.
+
+Account information will be retrieved from the file specified by `--accounts-file` argument,
+which is `~/.starknet_accounts/starknet_open_zeppelin_accounts.json` by default.
+Hides user's private keys by default.
+
+> ⚠️ **Warning**
```
Arcticae, line 8:
```suggestion
> ⚠️ **Warning**
>
```

docs/src/appendix/sncast/account/list.md
```diff
@@ -0,0 +1,20 @@
+# `list`
+List all available accounts.
+
+Account information will be retrieved from the file specified by `--accounts-file` argument,
+which is `~/.starknet_accounts/starknet_open_zeppelin_accounts.json` by default.
```
Arcticae, line 5: The command should read this file if `accounts-file` is not specified

crates/sncast/src/main.rs
```diff
@@ -412,13 +421,20 @@ async fn run_async_command(
                 )?;
                 Ok(())
             }
+
+            account::Commands::List(options) => {
+                print_account_list(&config.accounts_file, options.display_private_keys)?;
```
ksew1, line 426: You do not respect format flags `--int-format`, `--hex-format` and `--json`

docs/src/appendix/sncast/account/list.md
```diff
@@ -0,0 +1,20 @@
+# `list`
+List all available accounts.
+
+Account information will be retrieved from the file specified in user's environment.
+Hides user's private keys by default.
+
+> ⚠️ **Warning**
+> This command outputs cryptographic information about accounts, e.g. user's private key.
```
ksew1, line 8:
```suggestion
> This command may output cryptographic information about accounts, e.g. user's private key.
```

docs/src/starknet/account.md
```diff
@@ -101,6 +101,28 @@ result: Account successfully removed
...
+```shell
+$ sncast --accounts-file my-account-file.json account list
+
+Available accounts (at <current-directory>/my-account-file.json):
+
+Network "alpha-sepolia":
+- user0
+public key: 0x2f91ed13f8f0f7d39b942c80bfcd3d0967809d99e0cc083606cbe59033d2b39
+address: 0x4f5f24ceaae64434fa2bc2befd08976b51cf8f6a5d8257f7ec3616f61de263a
+salt: unspecified
```
ksew1, line 117:
```suggestion
```

docs/src/starknet/account.md
```diff
@@ -101,6 +101,28 @@ result: Account successfully removed
...
+class hash: unspecified
```
ksew1, line 118:
```suggestion
```

docs/src/starknet/account.md
ksew1, line 1: Can you add example with private keys?

ksew1 (changes requested):

Arcticae (approved):

integraledelebesgue:
### About a `--json` flag
* I have introduced support for this setting. However I assumed it is going to be used in specific cases (e.g. in user scripts)
* In JSON mode, the command outputs no information beyond the JSON itself
* In particular, it does not inform about the accounts file location and additional flags

integraledelebesgue:
## JSON mode - design question
There are two possibilites to output information in `--json` mode:
* Print a 1:1 structure read from `accounts-file` (accounts nested in networks)
* Transform it in a manner similar to the normal mode, i.e. add the `network` to the account document
Which one do we prefer?

Arcticae:
> ## JSON mode - design question
> There are two possibilites to output information in `--json` mode:
> 
> * Print a 1:1 structure read from `accounts-file` (accounts nested in networks)
> * Transform it in a manner similar to the normal mode, i.e. add the `network` to the account document
> 
> ### Which one do we prefer?

the latter

crates/sncast/src/starknet_commands/account/list.rs
```diff
@@ -0,0 +1,185 @@
...
+            data_repr.set_network(network);
+            result.insert(name.to_owned(), data_repr);
+        }
+    }
+
+    Ok(result)
+}
+
+fn print_as_json(networks: &HashMap<String, AccountDataRepresentation>) -> anyhow::Result<()> {
+    if networks.is_empty() {
```
ksew1, line 112: You can remove this if

ksew1 (approved):

PR #2297, opened 2024-07-11
Change trace directory name

piotmag769:
Better discoverability and more user friendly


integraledelebesgue (approved):
Very nice! :sparkles:

Issue #2302, opened 2024-07-12
Make RPCs URL be required only in commands it's relevant to


Arcticae:
### Which components does the task require to be changed? (think hard pls)

sncast

### Description

At the moment it's usage problematic because it needs to be included even if the command itself does not use it

PR #2305, opened 2024-07-16
Changed the `--url` argument semantics

integraledelebesgue:
Closes [#2302](https://github.com/orgs/foundry-rs/projects/3/views/2?filterQuery=assignee%3Aintegraledelebesgue&pane=issue&itemId=70733212)

## Introduced changes
- `--url` is no longer a common argument passed to `sncast`. Instead, each command that actually uses RPC handles it by itself
- `JsonRpcClient` is no longer created in `sncast`'s `main` function
- Instead, commands that require RPC instantiate their `provider`

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/sncast/src/main.rs
```diff
@@ -175,19 +172,19 @@ fn main() -> Result<()> {
...
-    verify_and_warn_if_incompatible_rpc_version(&provider, &config.url).await?;
-
     let wait_config = WaitForTx {
         wait: cli.wait,
         wait_params: config.wait_params,
     };
 
     match cli.command {
         Commands::Declare(declare) => {
+            let provider = get_provider(&config.url)?;
```
ksew1, line 185:
```rust
let provider = get_provider(&config.url)?;
verify_and_warn_if_incompatible_rpc_version(&provider, &config.url).await?;
```
You always repeat this sequence; please make this a function. Also, it could be declared in the top scope as lazy value, but it is up to you

ksew1 (changes requested):
1. Add this to the change log: which commands don't need a provider.
2. I don't like that, after these changes, this remains a common flag. The issue with this is that it is not respected by all commands, and you can't tell by running help if the command needs a URL or not. I would prefer to make this a local flag, but as this is a breaking change, I would also like to know the opinions of other team members.

Arcticae:
Agree with @ksew1 here - the purpose is to make using the RPC urls non-compulsory for the users
I realize that it's a breaking change, but it's necessary for the command interface to make sense

integraledelebesgue:
## `url` argument in `snfoundry.toml`
Do we still want to read the `url` argument from the config file since it becomes command-specific?

crates/sncast/src/main.rs
ksew1, line 1: You didn't delete rpc from `struct Cli`

crates/sncast/src/starknet_commands/account/add.rs
```diff
@@ -52,6 +52,10 @@ pub struct Add {
     #[allow(clippy::struct_field_names)]
     #[clap(long)]
     pub add_profile: Option<String>,
+
+    /// RPC provider url address; overrides url from snfoundry.toml
+    #[clap(short = 'u', long = "url")]
```
ksew1, line 57: You often repeat this. Instead create struct:
```rust
struct RpcArgs {
    /// RPC provider url address; overrides url from snfoundry.toml
    #[clap(short = 'u', long = "url")]
    pub rpc_url: Option<String>,
}
``` 
and use like this 

```rust
    #[clap(flatten)]
    pub rpc_args: RpcArgs,
```

crates/sncast/src/main.rs
```diff
@@ -569,4 +585,31 @@ fn update_cast_config(config: &mut CastConfig, cli: &Cli) {
         ),
         clone_or_else!(cli.wait_timeout, config.wait_params.get_timeout()),
     );
+
+    config.url = clone_or_else!(cli.rpc_url, config.url);
```
ksew1, line 589: I think this is a sign that we should go the other way around. Instead of updating `CastConfig` and then creating a provider from it, we should create the provider in the command. If RPC is not present, try to get it from `snfoundry.toml`.

crates/sncast/src/main.rs
```diff
@@ -569,4 +585,31 @@ fn update_cast_config(config: &mut CastConfig, cli: &Cli) {
...
+            command: multicall::Commands::Run(multicall::run::Run { ref rpc_url, .. }),
+        })
+        | Commands::ShowConfig(show_config::ShowConfig { ref rpc_url, .. })
+        | Commands::Script(script::Script {
+            command: script::Commands::Run(script::run::Run { ref rpc_url, .. }),
+        })
+        | Commands::TxStatus(tx_status::TxStatus { ref rpc_url, .. }) => rpc_url,
+        Commands::Account(Account { ref command }) => match command {
+            account::Commands::Add(account::add::Add { ref rpc_url, .. })
+            | account::Commands::Create(account::create::Create { ref rpc_url, .. })
```
ksew1, line 606: I think this is a sign that we should go the other way around. Instead of updating `CastConfig` and then creating a provider from it, we should create the provider in the command. If RPC is not present, try to get it from `snfoundry.toml`.

crates/sncast/src/lib.rs
```diff
@@ -177,6 +177,14 @@ pub fn get_provider(url: &str) -> Result<JsonRpcClient<HttpTransport>> {
     create_rpc_client(url)
 }
 
+pub async fn get_provider_and_verify_rpc_version(
```
ksew1, line 180: We can delete it as it is used only in one place

crates/sncast/src/starknet_commands/account/add.rs
```diff
@@ -52,6 +53,18 @@ pub struct Add {
     #[allow(clippy::struct_field_names)]
     #[clap(long)]
     pub add_profile: Option<String>,
+
+    #[clap(flatten)]
+    pub rpc_args: RpcArgs,
+}
+
+impl Provider for Add {
```
ksew1, line 61: I don't think it is worth to add additional layer here. I would just do `add.rpc_args.get_provider(&config).await?`

crates/sncast/src/starknet_commands/rpc.rs
ksew1, line 1: I doesn't belong here as it is not command

crates/sncast/src/starknet_commands/verify.rs
```diff
@@ -139,6 +139,10 @@ pub struct Verify {
     /// Specifies scarb package to be used
     #[clap(long)]
     pub package: Option<String>,
+
+    /// RPC provider url address; overrides url from snfoundry.toml
```
ksew1, line 143: I think you forgot to change this

integraledelebesgue: In fact it wasn't meant to be there :)

crates/sncast/src/starknet_commands/rpc.rs
integraledelebesgue: @ksew1 However it is closely related to commands as a helper struct. Do you have any suggestions where should I place this module?

crates/sncast/src/starknet_commands/deploy.rs
```diff
@@ -31,7 +32,7 @@ pub struct Deploy {
     pub salt: Option<FieldElement>,
 
     /// If true, salt will be modified with an account address
-    #[clap(short, long)]
+    #[clap(long)]
     pub unique: bool,
```
cptartur, line 36: Why was this changed?

crates/sncast/src/starknet_commands/rpc.rs
```diff
@@ -0,0 +1,25 @@
...
+    pub url: Option<String>,
+}
+
+impl RpcArgs {
+    pub async fn get_provider(
+        &self,
+        config: &CastConfig,
+    ) -> anyhow::Result<JsonRpcClient<HttpTransport>> {
+        let url = self.url.as_ref().unwrap_or(&config.url).to_owned();
+        let provider = get_provider(&url)?;
```
cptartur, line 19: There's no need to clone it fir we are using a reference in the next line anyway.

```suggestion
        let url = self.url.as_ref().unwrap_or(&config.url);
        let provider = get_provider(url)?;
```

crates/sncast/src/starknet_commands/deploy.rs
integraledelebesgue: `-u` from `unique` cannot coexist with `-u` from `url`. As `unique` is a very specialised flag, used only in one command, I've chosen to delete its short form to resolve the conflict.

crates/sncast/src/starknet_commands/rpc.rs
integraledelebesgue: That's a good point :)

cptartur (approved):

crates/sncast/src/starknet_commands/rpc.rs
ksew1: helper module

ksew1 (approved):

PR #2308, opened 2024-07-16
Enable declare v3

ksew1:
**Stack**:
- #2310
- #2309
- #2308 ⬅
- #2307

### DIFF
--- a/crates/sncast/src/helpers/fee.rs
+++ b/crates/sncast/src/helpers/fee.rs
@@ -4,7 +4,7 @@ use conversions::serde::deserialize::CairoDeserialize;
 use starknet::core::types::{BlockId, FieldElement};
 use starknet::providers::Provider;
 
-#[derive(Args, Debug)]
+#[derive(Args, Debug, Clone)]
 pub struct FeeArgs {
     /// Token that transaction fee will be paid in
     #[clap(long)]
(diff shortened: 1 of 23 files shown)


crates/sncast/src/helpers/fee.rs
```diff
@@ -151,3 +151,38 @@ pub enum FeeSettings {
         max_gas_unit_price: Option<u128>,
     },
 }
+
+#[macro_export]
+macro_rules! impl_from_version_for_fee_token {
```
Arcticae, line 156: I really like the solution, although i wonder: 
Couldn't this be solved with a trait? 
Something like `PayableTransaction`, which would expose the mapping of version -> token 
And also would define a method for formatting the error

WDYT?

Arcticae (approved):
lgtm overall, one nitpick/q

crates/sncast/src/helpers/fee.rs
ksew1: I like this idea but I want to be sure we're on the same page:

I am thinking to have a trait like this:

```rust
trait PayableTransaction {
    fn error_message(&self, token: &str, version: &str) -> String;
    fn validate(&self) -> Result<()>;
    fn version_from_token(&self) -> Option<FeeToken>;
}
```

and implement it via a macro:

```rust
impl_payable_transaction!(Invoke, token_not_supported_for_invoke,
    InvokeVersion::V1 => FeeToken::Eth,
    InvokeVersion::V3 => FeeToken::Strk
);
```

Arcticae: Yep i am not against implementing it via a macro
Having the clear interface without `cargo expand` is what i'm after

ksew1:
✓ Commit merged in pull request [#2310](https://github.com/foundry-rs/starknet-foundry/pull/2310)

Issue #2220, opened 2024-06-18
Add linking to block explorer on transaction outputs


Arcticae:
### Which components does the task require to be changed? (think hard pls)

sncast

### Description

It's easier to just click on the link and have the browser open for you in the exact right place
We should make this configurable just in case (have the user choose one of `starkscan`/`voyager` upon installation, or have a reasonable default). 
The default probably should be starkscan since it has better results on pagespeed, see reports:
[starkscan tx screen analysis](https://pagespeed.web.dev/analysis/https-starkscan-co-tx-0x06f250026f15fe6189d0e5e374daa63b1283e9619d4279ae12c4c433b694be19/1ns3ajpj3f?form_factor=desktop&category=performance&category=accessibility&category=best-practices&category=seo&hl=pl&utm_source=lh-chrome-ext)
[voyager tx screen analysis](https://pagespeed.web.dev/analysis/https-voyager-online-tx-0x6f250026f15fe6189d0e5e374daa63b1283e9619d4279ae12c4c433b694be19/1ljc1clbos?form_factor=desktop&category=performance&category=accessibility&category=best-practices&category=seo&hl=pl&utm_source=lh-chrome-ext)
[starkscan frontpage](https://pagespeed.web.dev/analysis/https-starkscan-co/aqry8b59xe?form_factor=desktop&category=performance&category=accessibility&category=best-practices&category=seo&hl=pl&utm_source=lh-chrome-ext)
[voyager frontpage](https://pagespeed.web.dev/analysis/https-voyager-online/sy5d5tha2w?form_factor=desktop&category=performance&category=accessibility&category=best-practices&category=seo&hl=pl&utm_source=lh-chrome-ext)

PR #2334, opened 2024-07-26
Added block explorer links on transaction output

integraledelebesgue:
Closes [#2220](https://github.com/orgs/foundry-rs/projects/3/views/2?pane=issue&itemId=67849333)

## Introduced changes
- All commands which commit a transaction, when invoked in human-readable mode, display a link
- Link leads to a search page in the block explorer, containing a detailed description of the transaction
- Used explorer is configurable via new configuration key - `block-explorer`
- We currently support all services listed [here](https://docs.starknet.io/documentation/tools/ref_block_explorers/)

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


integraledelebesgue:
## Explorer service
I have hardcoded two links (StarkScan and Voyager), but actual implementation uses StarkScan only. How should I introduce some customisation to it?
I think of three possibilities:
1. Hardcode one link and delete the other
2. Introduce new option to CastConfig and `snfoundry.toml` specifing the link
3. Use a custom `cfg` to choose the link at compile time

crates/sncast/src/response/print.rs
```diff
@@ -62,123 +68,150 @@ impl Display for OutputValue {
...
-    let formatted_output = output
-        .into_iter()
-        .map(|(k, v)| (k, apply_numbers_formatting(v, numbers_format)))
-        .collect();
-
-    for val in pretty_output(formatted_output, output_format)? {
-        match result {
-            Ok(_) => println!("{val}"),
-            Err(_) => eprintln!("{val}"),
+impl From<Value> for OutputValue {
```
ksew1, line 71: `NIT`: It's nice that you care about quality of code 😄, but I would do the refactor in a separate PR (Don't change this now, but it would be nice in the future). 

From the review perspective, it's hard to understand why some changes were made. Since it's not changing any behavior here, and it's not directly related to the PR topic, a separate refactor PR would be more appropriate.

crates/sncast/src/response/explorer_link.rs
```diff
@@ -0,0 +1,16 @@
+use super::{print::OutputFormat, structs::OutputLink};
+
+#[allow(dead_code)]
```
ksew1, line 3:
```suggestion
```

ksew1:
> ## Explorer service
> 
> I have hardcoded two links (StarkScan and Voyager), but actual implementation uses StarkScan only. How should I introduce some customisation to it? I think of three possibilities:
> 
>     1. Hardcode one link and delete the other
> 
>     2. Introduce new option to CastConfig and `snfoundry.toml` specifing the link
> 
>     3. Use a custom `cfg` to choose the link at compile time

Definitely not third option. I would go for second

crates/sncast/src/response/structs.rs
```diff
@@ -151,3 +151,35 @@ pub struct VerifyResponse {
 }
 
 impl CommandResponse for VerifyResponse {}
+
+pub trait OutputLink {
```
Arcticae, line 155: This looks like a better use case for a plain function

Arcticae (changes requested):

CHANGELOG.md
```diff
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Cast
 
+#### Added
+- Commands that commit transactions now display links to block explorers. When in human-readable mode, `invoke`, `declare`, `deploy`, `account create` and `account deploy` will display additional information with an url.
```
cptartur, line 13: Super nit, but probably doesn't need to mention that it's in human readable mode.

crates/sncast/src/response/explorer_link.rs
```diff
@@ -0,0 +1,16 @@
...
+const VOYAGER_SEARCH: &str = "https://voyager.online/tx";
+
+pub fn print_block_explorer_link_if_allowed<T: OutputLink>(
+    result: &anyhow::Result<T>,
+    output_format: OutputFormat,
+) {
+    if let (Ok(response), OutputFormat::Human) = (result, output_format) {
+        let url = response.format_url(STARKSCAN_SEARCH);
+        println!("\nVisit {url}\nto see transaction details");
```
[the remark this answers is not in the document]
cptartur: Going by that logic, declare could probably link to the deployed class_hash as well.

crates/sncast/src/response/structs.rs
integraledelebesgue: I have simplified this trait and moved its definition to the module

CHANGELOG.md
integraledelebesgue: I think it's worth emphasising. I can write that this behaviour is disabled in JSON mode instead

crates/sncast/tests/e2e/deploy.rs
```diff
@@ -84,8 +84,9 @@ async fn test_happy_case_human_output() {
             contract_address: [..]
             transaction_hash: [..]
 
-            Visit [..]
-            to see transaction details
+            Details:
```
cptartur, line 87: What about something like

```
See transaction (or some other word here) details:
  Contract: ...
  Transaction: ...
```

Just `Details:` is a bit "raw" compared to previous version with `Visit ... to see transaction details`.

crates/sncast/src/response/structs.rs
```diff
@@ -151,3 +154,49 @@ pub struct VerifyResponse {
...
+}
+
+impl OutputLink for DeclareResponse {
+    const TITLE: &'static str = "declaration";
+
+    fn format_links(&self, service: &str) -> String {
+        formatdoc!(
+            "
+            class: {service}/{:x}
+            transaction: {service}/{:x}
```
tomek0123456789, line 188: Not sure if it's going to be that simple. In case of Starkscan we have an endpoint `/search`, that simply redirects to the class/contract/etc. we're looking for. In case of e.g. Voyager, there is no such endpoint (or at least I haven't found it). This implementation implies that every explorer would have such an endpoint, which is not that obvious.

tomek0123456789 (changes requested):

crates/sncast/src/response/structs.rs
integraledelebesgue: Indeed - Voyager has separate endpoints for block, transactions etc.
What we can do is to narrow available block explorer services to have a control over URL formatting.
I propose introducing an enum with supported services.

CHANGELOG.md
```diff
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Cast
 
+#### Added
+- Commands that commit transactions now display links to block explorers. When in human-readable mode, `invoke`, `declare`, `deploy`, `multicall run`, `account create` and `account deploy` will display additional information with an url. A new key in Cast congirutation - `block-explorer-search-url` determines which block explorer service the displayed link leads to
```
tomek0123456789, line 13:
```suggestion
- Commands that commit transactions now display links to block explorers. When in human-readable mode, `invoke`, `declare`, `deploy`, `multicall run`, `account create` and `account deploy` will display additional information with an url. A new key in Cast congirutation - `block-explorer` determines which block explorer service the displayed link leads to
```

tomek0123456789 (approved):
LGTM

crates/sncast/src/response/explorer_link.rs
```diff
@@ -0,0 +1,21 @@
+use super::print::OutputFormat;
+use crate::helpers::block_explorer::Service;
+
+pub trait OutputLink {
+    const TITLE: &'static str;
+
+    fn format_links(&self, service: Service) -> String;
```
ksew1, line 7:
```suggestion
    fn format_links(&self, provider: Box<dyn LinkProvider>) -> String;
```

The you only use `.as_provider once`

crates/sncast/src/helpers/block_explorer.rs
```diff
@@ -0,0 +1,138 @@
...
+            Service::OkLink => Box::new(OkLink),
+            Service::NftScan => Box::new(NftScan),
+        }
+    }
+}
+
+pub trait LinkProvider {
+    fn transaction(&self, hash: FieldElement) -> String;
+    fn class(&self, hash: FieldElement) -> String;
+    fn account(&self, address: FieldElement) -> String;
```
ksew1, line 36: Isn't account treated as contract so they will always have the same link?

ksew1 (approved):

Arcticae (approved):

PR #2336, opened 2024-07-30
propose verification spec changes

rjnrohit:
## Introduced changes

This PR introduces changes to verification proposal document mainly as follows:
1. Actual verification to be performed on class level
2. Keep `contract_address` (user should be able to provide contract_address, which we resolve to `class_hash`)
Add `class_hash`
** Either contract_address or class_hash should be provided by user
3. Change `contract_name` to `class_name` OR keep `contract_name`

## Checklist

- [ ] Linked relevant issue
- [ ] Added relevant tests


Arcticae:
@rjnrohit I am lacking some rationale for those changes here. They make sense on a technical level but i don't understand why we need them, could you provide some details?

rjnrohit:
> @rjnrohit I am lacking some rationale for those changes here. They make sense on a technical level but i don't understand why we need them, could you provide some details?

Sure!
Since, Verification status labelled on class hash not on the contract addresses. And also ,If contract has been declared, it can be verified using class hash. 
Renaming of `contract-name` to `class-name` has been done because technically it is more aligned term, as requested by the voyager team

Arcticae:
@rjnrohit i asked for folks from Walnut to review this to make sure we're on the same page with this change, since it's your common interface and i would really like it to be aligned.

barabanovro:
We agree and all points seem reasonable. Please ping us when you have the PR with the changes so we can also implement changes on `WalnutVerificationInterface` and our backend. Thanks!

Arcticae (approved):

PR #2341, opened 2024-07-31
Change declare return

ksew1:
**Stack**:
- #2342
- #2341 ⬅

### DIFF
--- a/crates/cheatnet/tests/common/assertions.rs
+++ b/crates/cheatnet/tests/common/assertions.rs
@@ -2,6 +2,8 @@ use cairo_vm::Felt252;
 use cheatnet::runtime_extensions::call_to_blockifier_runtime_extension::rpc::{
     CallFailure, CallResult,
 };
+use cheatnet::runtime_extensions::forge_runtime_extension::cheatcodes::declare::DeclareResult;
+use starknet_api::core::ClassHash;
 
 #[inline]
 pub fn assert_success(call_contract_output: CallResult, expected_data: &[Felt252]) {
@@ -32,4 +34,19 @@ pub fn assert_error(call_contract_output: CallResult, expected_data: impl Into<S
         )
         if msg == expected_data.into(),
     ));
+}
+
+pub trait ClassHashAssert {
+    fn unwrap_success(self) -> ClassHash;
+}
+
+impl ClassHashAssert for DeclareResult {
+    fn unwrap_success(self) -> ClassHash {
+        match self {
+            DeclareResult::Success(class_hash) => class_hash,
+            DeclareResult::AlreadyDeclared(class_hash) => {
+                panic!("Class hash {class_hash} is already declared")
+            }
+        }
+    }
 }
--- a/crates/forge/tests/integration/cheat_block_number.rs
+++ b/crates/forge/tests/integration/cheat_block_number.rs
@@ -15,7 +15,10 @@ fn cheat_block_number_basic() {
             use traits::TryInto;
             use starknet::ContractAddress;
             use starknet::Felt252TryIntoContractAddress;
-            use snforge_std::{ declare, ContractClassTrait, start_cheat_block_number, stop_cheat_block_number, stop_cheat_block_number_global, start_cheat_block_number_global };
+            use snforge_std::{
+                declare, ContractClassTrait, DeclareResultTrait, start_cheat_block_number,
+                stop_cheat_block_number, stop_cheat_block_number_global, start_cheat_block_number_global
+            };
 
             #[starknet::interface]
             trait ICheatBlockNumberChecker<TContractState> {
@@ -24,7 +27,7 @@ fn cheat_block_number_basic() {
 
             #[test]
             fn test_stop_cheat_block_number() {
-                let contract = declare("CheatBlockNumberChecker").unwrap();
+                let contract = declare("CheatBlockNumberChecker").unwrap().contract_class();
                 let (contract_address, _) = contract.deploy(@ArrayTrait::new()).unwrap();
                 let dispatcher = ICheatBlockNumberCheckerDispatcher { contract_address };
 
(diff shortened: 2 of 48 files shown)


crates/forge/tests/integration/cheat_block_number.rs
ksew1, line 1: After fixing these tests manually, I think the current design for integration tests is pretty bad. Writing tests as a string is hard to correct; you have no LSP support, and these tests aren't formatted correctly. I think all of them should be read from a file.

crates/cheatnet/tests/common/assertions.rs
```diff
@@ -33,3 +35,18 @@ pub fn assert_error(call_contract_output: CallResult, expected_data: impl Into<S
...
+
+impl ClassHashAssert for DeclareResult {
+    fn success_class_hash(self) -> ClassHash {
+        match self {
+            DeclareResult::Success(class_hash) => class_hash,
+            DeclareResult::AlreadyDeclared(class_hash) => {
+                panic!("Class hash {class_hash} is already declared")
+            }
+        }
+    }
```
cptartur, line 51: This is confusing to me. At least, if you want to code it this way, name it in a way that suggests that this function is panicking/asserting something. E.g. `unwrap_success_class_hash`. 

At a caller site, it's not obvious what is it doing at all.

cptartur: Or we could change the `DeclareResult` a bit, so it returns a struct and then implement something along the lines of `unwrap_success. 

Then we could write `declare(state, contract_name, &contracts_data).unwrap_success().class_hash`.

crates/forge/tests/integration/cheat_block_number.rs
cptartur: Agreed no lsp is annoying, but having all tests in a separate file can become confusing with large enough amount of tests. Where you have to look for a relevant test code declaration if anything needs to be changed.

snforge_std/src/cheatcodes/contract_class.cairo
```diff
@@ -87,11 +94,33 @@ impl ContractClassImpl of ContractClassTrait {
...
+            DeclareResult::Success(contract_class) => contract_class,
+            DeclareResult::AlreadyDeclared(contract_class) => contract_class
+        }
+    }
+
+    fn success_contract_class(self: @DeclareResult) -> @ContractClass {
+        match self {
+            DeclareResult::Success(contract_class) => contract_class,
+            DeclareResult::AlreadyDeclared(_) => panic!("Class hash is already declared")
+        }
```
cptartur, line 115: Don't like this as a public interface, at least not named like this. It is not obvious it can panic. 

Either add `unwrap`/`assert` to the name or not include these at all. I'm not sure if there is relevant usecase for this method for end users.

cptartur (changes requested):

crates/cheatnet/tests/common/assertions.rs
ksew1: I like the name `unwrap_success`, but I don't understand why it should return a struct. We should just return the hash, not a struct containing it. The `DeclareResult` is a container. The same behavior is seen in Rust's `Result` enum. When you call `unwrap_err`, you get the inner value, not a struct containing it. So, I think it should just be `declare(state, contract_name, &contracts_data).unwrap_success()`.

snforge_std/src/cheatcodes/contract_class.cairo
ksew1: It's just an analogy to Rust's `Result` enum, so there definitely should be a method like this. I know the naming is unfortunate, but what about `unwrap_success`? The second method name could remain the same.

crates/forge/tests/integration/cheat_block_number.rs
ksew1: Well, one test per file would solve this issue. I don't see any benefit from hardcoding the test. If you want to change anything in them, you're pretty much going to copy it to an editor to get LSP support in most cases. Additionally, reading tests without syntax coloring is also not helpful.

crates/cheatnet/tests/common/assertions.rs
cptartur: `unwrap_success` is okay to me, the idea with returning a struct was just an different option, but agreed that it might not make that much sense.

snforge_std/src/cheatcodes/contract_class.cairo
cptartur: I think `unwrap_sucess` is okay.

cptartur: Not sure I get the usecase for asserting/unwrapping if the contract got successfully declared. In both cases of DeclareResult, you are getting an information that the contract is declared and can be used, not sure in which cases would the user want to verify if it was not previously declared.

ksew1: After discussing this offline with other team member, we decided to remove this

crates/forge/tests/integration/declare.rs
```diff
@@ -106,21 +109,25 @@ fn redeclare() {
...
-                Result::Ok(contract) => contract,
+                Result::Ok(result) => result.contract_class().clone(),
                 Result::Err(_) => panic!("Failed to declare contract"),
             };
             let class_hash = contract.class_hash.into();
             assert(class_hash != 0, 'proper class hash');
-            assert!(declare("HelloStarknet").is_err(), "Contract redeclared");
+            match declare("HelloStarknet").unwrap() {
+                DeclareResult::AlreadyDeclared(_) => {},
+                _ => panic!("Contract redeclared")
```
cptartur, line 129: I think it should be that? 

```suggestion
                _ => panic!("Contract not redeclared")
```

cptartur (approved):

crates/forge/tests/integration/declare.rs
ksew1: I think it is okay as it is. Re-declaration is a mistake, so it should panic with `Contract redeclared`.

ksew1:
✓ Commit merged in pull request [#2342](https://github.com/foundry-rs/starknet-foundry/pull/2342)

PR #2404, opened 2024-08-27
Fix `snforge init` logic, exit smoke tests on any error

cptartur:
Closes #

## Introduced changes

- `snforge init` will first create a project using scarb's `cairo-test` template and then override dependencies accordingly. This is so scarb doesn't call snforge init again internally
- Added `set -e` to `smoke_tests.sh` so it fails on any error

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests

### DIFF
--- a/crates/forge/src/init.rs
+++ b/crates/forge/src/init.rs
@@ -98,7 +98,7 @@ pub fn run(project_name: &str) -> Result<()> {
             .current_dir(current_dir)
             .arg("new")
             .arg(&project_path)
-            .env("SCARB_INIT_TEST_RUNNER", "starknet-foundry")
+            .env("SCARB_INIT_TEST_RUNNER", "cairo-test")
             .run()
             .context("Failed to initialize a new project")?;
     }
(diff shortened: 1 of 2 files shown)


integraledelebesgue (approved):

crates/forge/src/init.rs
```diff
@@ -98,13 +98,23 @@ pub fn run(project_name: &str) -> Result<()> {
             .current_dir(current_dir)
             .arg("new")
             .arg(&project_path)
-            .env("SCARB_INIT_TEST_RUNNER", "starknet-foundry")
+            .env("SCARB_INIT_TEST_RUNNER", "cairo-test")
```
Arcticae, line 101: Why

Arcticae (changes requested):

crates/forge/src/init.rs
cptartur: So scarb doesn't call `snforge init` again under the hood https://github.com/software-mansion/scarb/blob/09590f5fc07365c3e4188b0e8fa2e774a8269d68/scarb/src/ops/new.rs#L220. We are replacing the template generated anyway.

Arcticae (approved):

Issue #2344, opened 2024-08-01
Automatic uploads to registry


THenry14:
### Which components does the task require to be changed? (think hard pls)

snforge, sncast

### Description

Since the registry is up (scarbs.xyz) we should think about automatically uploading new versions to it on our CI.
For now this can be achieved by curl. After scarb 2.8.0 there should be support for uploads using it.

THenry14:
Child issues:

https://github.com/foundry-rs/starknet-foundry/issues/2462 (Remove scarb version constraint for foundry_std registry uploads)
https://github.com/foundry-rs/starknet-foundry/issues/2463 (Add publish workflow for snforge_scarb_plugin)

PR #2461, opened 2024-09-11
Add upload to registry workflow for foundry_std

### LINKED ISSUE #2463: Add publish workflow for snforge_scarb_plugin

other (describe below)

### Description

Related to https://github.com/foundry-rs/starknet-foundry/issues/2344

Under said issue automatic uploads for snforge_std and sncast_std was added. At the time of creating relevant PRs there was still no support for scarb package for procmacros. The aim of this issue is to update the publish workflow, to also automatically upload new snforge_scarb_plugin whenever its version changes. This could probably be a separate publish flow for each push to master, where `crates/snforge_scarb_plugin/Scarb.toml` is changed and its version is not already deployed to the registry.

Blocked by https://github.com/software-mansion/scarb/issues/1527

Up until now, whenever a new version of snforge_scarb_plugin is needed please construct the package manually. If needed, seek help from scarb team.

THenry14:
Closes https://github.com/foundry-rs/starknet-foundry/issues/2344
Closes #2463 

This PR is a part of the stack:
-- Add missing info recommended for package uploads (https://github.com/foundry-rs/starknet-foundry/pull/2460)
--> Add github action workflow for automatic registry uploads during release (This PR)

## Introduced changes

- Updates github workflow to also upload std libraries to scarbs registry during release
- Sets snforge_scarb_plugin to specific version downloadable from the registry, as required by scarb

## Checklist

- [X] Linked relevant issue
- [X] Added relevant tests


franciszekjob:
Blocked by https://github.com/software-mansion/scarb/pull/1605

kkawula:
It needs clarification, but plugins weren't correctly handled by `scarb` until https://github.com/software-mansion/scarb/pull/1712 was merged, I suppose that it will be needed to update `scarb` version in the entire project. 
In the earlier versions `scarb` had problem with access to the cache in database which caused dead lock. 
CI to lookup: https://github.com/foundry-rs/starknet-foundry/actions/runs/11577774705

maciektr:
> It needs clarification, but plugins weren't correctly handled by scarb until https://github.com/software-mansion/scarb/pull/1712 was merged, I suppose that it will be needed to update scarb version in the entire project.

It shouldn't have anything to do with uploading packages though

.github/workflows/publish_packages.yml
```diff
@@ -0,0 +1,56 @@
...
+      - name: Check if package version exist in the registry
+        run: |
+          set -eo pipefail
+          snforge_scarb_plugin=$(curl -s https://scarbs.xyz/api/v1/index/sn/fo/snforge_scarb_plugin.json | jq --arg version "${{ steps.scarb-init.outputs.CAIRO_VERSION }}" '[.[] | select(.v == $version)] | length > 0')
+          snforge_std=$(curl -s https://scarbs.xyz/api/v1/index/sn/fo/snforge_std.json | jq --arg version "${{ steps.scarb-init.outputs.CAIRO_VERSION }}" '[.[] | select(.v == $version)] | length > 0')
+          sncast_std=$(curl -s https://scarbs.xyz/api/v1/index/sn/ca/sncast_std.json | jq --arg version "${{ steps.scarb-init.outputs.CAIRO_VERSION }}" '[.[] | select(.v == $version)] | length > 0')
+
+          echo "snforge_scarb_plugin=$snforge_scarb_plugin"
+          echo "snforge_std=$snforge_std"
+          echo "sncast_std=$sncast_std"
```
kkawula, line 35: It gives cool info on what was published, in the sense that we do not want update the plugin every release.
image

.github/workflows/publish_packages.yml
```diff
@@ -0,0 +1,56 @@
...
+        if: steps.check-plugins.outputs.snforge_scarb_plugin != 'true'
+        working-directory: crates/snforge-scarb-plugin
+        run: scarb publish
+
+      - name: Package and upload snforge_std
+        if: steps.check-plugins.outputs.snforge_std != 'true'
+        working-directory: snforge_std
+        run: |
+          ../scripts/set_plugin_version.sh
+          scarb publish --allow-dirty
```
kkawula, line 51: `--allow-dirty` because we modify `snforge_std` by executing the script without committing changes

.github/workflows/publish_packages.yml
```diff
@@ -0,0 +1,56 @@
...
+      - uses: actions/checkout@v4
+
+      - uses: dtolnay/rust-toolchain@7b1c307e0dcbda6122208f10795a713336a9b35a
+        with:
+          toolchain: stable
+
+      - uses: software-mansion/setup-scarb@v1
+        with:
+          # todo: Use scarb stable version that support publishing plugins (after 2.8.4)
+          scarb-version: "nightly-2024-11-09"
```
kkawula, line 24: created issue https://github.com/foundry-rs/starknet-foundry/issues/2669

kkawula:
it works, tested on dev registry.
- [workflow](https://github.com/kkawula/starknet-foundry/actions/runs/11815728427/job/32917498713)
- [dev_registry](https://scarbs.dev/packages/snforge_std/0.33.0)

.github/workflows/publish_packages.yml
```diff
@@ -0,0 +1,56 @@
...
+
+      - name: Check if package version exist in the registry
```
DelevoXDG, line 26: Is this step necessary?

We implemented this for publishing Scarb plugins because their version matches the Cairo version. Since multiple Scarb versions can correspond to a single Cairo version, running the workflow with every Scarb release could cause errors by publishing plugins with the same version. 

Afaik, this doesn’t apply to snfoundry plugins because their version isn’t tied to any Cairo version. Normally, the workflow should only be triggered once per tag.

.github/workflows/publish_packages.yml
```diff
@@ -0,0 +1,56 @@
+name: publish_packages.yml
+on:
+  release:
+    types:
+      - published
+  workflow_dispatch:
```
DelevoXDG, line 6: I think we only want to publish on tag release

DelevoXDG (commented):
Partial review, please re-request when ready 👍🏻

.github/workflows/publish_packages.yml
```diff
@@ -0,0 +1,56 @@
+name: publish_packages.yml
+on:
+  release:
+    types:
+      - published
```
cptartur, line 5: I think `published` will also trigger on creating a draft release which our workflow does. I think it should be `released` so it only uploads to registry if we publish a new release.

cptartur: Also a concern, in case we ever create a pre-release, do we want to upload to registry as well? Does it support pre-release versions?

cc @maciektr

.github/workflows/publish_packages.yml
cptartur: I think there might be cases where we want to trigger upload manually. I think we can keep it

.github/workflows/publish_packages.yml
cptartur: We might decouple foundry plugin version from foundry itself so it might be the case that for consecutive foundry releases, plugin version will stay the same.

So we have to make sure we don't reupload the plugin to the registry in these cases.

kkawula: Sometimes we do patch plugin, without release then we would like to publish it and sometimes we will not do release plugin together with snforge_std and sncast_std, we want to set it on 1.0.0 and keep it until there are changes, and then I would fail the workflow and CI will be red at the release

.github/workflows/publish_packages.yml
```diff
@@ -0,0 +1,61 @@
...
+          echo "snforge_scarb_plugin=$snforge_scarb_plugin"
+          echo "snforge_std=$snforge_std"
+          echo "sncast_std=$sncast_std"
+
+          echo "snforge_scarb_plugin=$snforge_scarb_plugin" >> $GITHUB_OUTPUT
+          echo "snforge_std=$snforge_std" >> $GITHUB_OUTPUT
+          echo "sncast_std=$sncast_std" >> $GITHUB_OUTPUT
+
+      - name: Package and upload snforge_scarb_plugin
+        if: steps.check-plugins.outputs.snforge_scarb_plugin != 'true'
```
cptartur, line 47: Nit, but maybe this is more readable?
```suggestion
        if: steps.check-plugins.outputs.snforge_scarb_plugin == 'false'
```

scripts/set_plugin_version.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+PLUGIN_FILE_PATH="../crates/snforge-scarb-plugin/Scarb.toml"
+
+VERSION=$(grep '^version' "$PLUGIN_FILE_PATH" | sed -E 's/version = "([^"]*)"/\1/')
```
cptartur, line 5: Assuming this is meant to extract the version, why not use this code again?

```shell
snforge_std_version=$(cat snforge_std/Scarb.toml | grep version | cut -d '"' -f 2
```

scripts/set_plugin_version.sh
```diff
@@ -0,0 +1,11 @@
...
+
+STD_FILE_PATH="../snforge_std/Scarb.toml"
+
+sed -i.bak "/snforge_scarb_plugin/ s/\(snforge_scarb_plugin = \).*/\1\"^${VERSION}\"/" $STD_FILE_PATH
+
+rm ${STD_FILE_PATH}.bak 2> /dev/null
```
cptartur, line 11: If we are always removing the `.bak` file anyway, why not just omit creating it?

cptartur (changes requested):

.github/workflows/publish_packages.yml
kkawula: You are right, it should be changed to `released`
https://docs.github.com/en/webhooks/webhook-events-and-payloads?actionType=released#release

.github/workflows/publish_packages.yml
DelevoXDG: As I pointed out offline, if you're going in the direction of decoupling plugin from foundry itself, it seems to me it a separate workflow should be considered

.github/workflows/publish_packages.yml
kkawula: done

.github/workflows/publish_packages.yml
DelevoXDG: Why not require from tag that triggers the workflow to already include changes introduced by `set_plugin_version.sh`? 🤔

kkawula: For development reason, we always would like to have a relative path specified in the dependency

.github/workflows/publish_packages.yml
kkawula: It is one of the opportunities, but isn't it good enough? @cptartur

cptartur: As on how the decoupling will go exactly is yet to be determined. I think the workflow in the current form is okay, and we can introduce changes in the future.

.github/workflows/publish_packages.yml
cptartur: Okay, if we want to upload pre-releases as well, let's keep it at published.

.github/workflows/publish_packages.yml
```diff
@@ -0,0 +1,62 @@
+name: publish_packages.yml
+on:
+  release:
+    types:
+      - released
```
cptartur, line 5: Following the discussion from different comment 

```suggestion
      - published
```

.github/workflows/publish_packages.yml
```diff
@@ -0,0 +1,62 @@
...
+          echo "snforge_scarb_plugin=$snforge_scarb_plugin"
+          echo "snforge_std=$snforge_std"
+          echo "sncast_std=$sncast_std"
+
+          echo "snforge_scarb_plugin=$snforge_scarb_plugin" >> $GITHUB_OUTPUT
+          echo "snforge_std=$snforge_std" >> $GITHUB_OUTPUT
+          echo "sncast_std=$sncast_std" >> $GITHUB_OUTPUT
+
+      - name: Package and upload snforge_scarb_plugin
+        if: steps.check-versions.outputs.snforge_scarb_plugin == 'false'
```
cptartur, line 48: Actually, shouldn't it be without quotes in case of booleans? So just `false`? 

I've checked in the GH docs and it's not 100% clear to me.

cptartur: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions

https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions

cptartur (commented):
@THenry14 Can you please review as well? I know it's technically your PR but it's changed a bit 😅

.github/workflows/publish_packages.yml
THenry14: Similarly to @DelevoXDG , I would probably be in favour of splitting release flows for macros and standard libraries as well. 

My understanding is that the plugin won't necessarily follow the *_std version, so it may make sense to release it independently. For this manual workflow dispatch only is good enough. Most of logic from this file is fine, though you wouldn't need anything from this step (I mean the check-versions, which I find very hacky) so I'd actually argue it could become far simpler.

Having this, we could probably incorporate *_std upload to the `release.yml` workflow, as it seems to me this is more natural place for this. We already verify the version there, so chances are we do not need any further modifications - we could probably just add a new job (standard-libraries-release, or scarbs-release or sth) that will set the plugin version (this is unnecessary if you decide to set the version in Scarb.toml in repo) and use scarb to publish and call it a day. No need to check versions against the registry, verify what should/should not be published etc. 

Thoughts?

scripts/set_plugin_version.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+PLUGIN_FILE_PATH="../crates/snforge-scarb-plugin/Scarb.toml"
+
+VERSION=$(cat "$PLUGIN_FILE_PATH" | grep version | cut -d '"' -f 2)
```
THenry14, line 5: useless use of cat

https://www.shellcheck.net/wiki/SC2002

.github/workflows/publish_packages.yml
THenry14: Is it a hard requirement to set the version in snforge only for uploads? What are the downsides of having this set in repo?

scripts/set_plugin_version.sh
THenry14: This probably explains it better
https://en.wikipedia.org/wiki/Cat_(Unix)#Useless_use_of_cat

.github/workflows/publish_packages.yml
kkawula: I generally agree with you, but I can still see a few problems we could encounter. In scenario where this workflow is splited and moved to `release` we can get error while releasing new versions of _std and plugins simultaneously. Then plugin should be then uploaded before _std, to make `snforge_std` dependency working. As other job in can be made sequential, but if we like to provide workflow_dispatcher then its generates a lot of unnecessary boilerplate code (`ifs`) in entire workflow, because we can’t actually specify exact job to run on dispatch. In scenario of two different workflows it actually can’t be made sequential and then we can only hope that plugin will be uploaded first.

.github/workflows/publish_packages.yml
kkawula: Yes, we would like always to test the latest plugin version while developing, so relative path has to be provided. e.g. e2e test
There is always a case where it should be changed this way or another, definitely here will be the easiest.

scripts/set_plugin_version.sh
kkawula: Cool, removed

.github/workflows/publish_packages.yml
THenry14: > In scenario where this workflow is splited and moved to release we can get error while releasing new versions of _std and plugins simultaneously

Well I do not think we necessarily need to release macro in the same time we release package, and because of that workflow_dispatch would be ideal. I can easily imagine macro not being updated for months, or be updated multiple times during one week when we do not want to release new snfoundry version.

If you wanted to the release workflow could call the plugin publish workflow, to have the functionality we're aiming for and be sure that the release operations are indeed sequential.

DelevoXDG: > If you wanted to the release workflow could call the plugin publish workflow, to have the functionality we're aiming for and be sure that the release operations are indeed sequential.

That's how I see it as well, we should have:
1. Plugin publish workflow with `workflow_dispatch` and `workflow_call`
2. Foundry publish workflow on `release` and `workflow_dispatch`. It should first should check if plugin is up to date in first job and re-use the first workflow if necessary to upload the plugin, and release the foundry libs afterwards

> As other job in can be made sequential, but if we like to provide workflow_dispatcher then its generates a lot of unnecessary boilerplate code (`ifs`) in entire workflow, because we can’t actually specify exact job to run on dispatch.

Do you have any specific examples of why that might be the case? 🤔 I don't think this should really cause any boilerplate or significantly more ifs than we have now

kkawula: Seems convincing. Updated.

.github/workflows/release.yml
```diff
@@ -249,3 +249,53 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Check version
+        id: check-version
+        run: |
+          set -eo pipefail
+
+          snforge_scarb_plugin_version=$(grep version crates/snforge-scarb-plugin/Scarb.toml | cut -d '"' -f 2)
+          snforge_scarb_plugin=$(curl -s https://scarbs.xyz/api/v1/index/sn/fo/snforge_scarb_plugin.json | jq --arg version $snforge_scarb_plugin_version '[.[] | select(.v == $version)] | length > 0')
```
THenry14, line 265: Are you sure this curl works as intended? I'm getting:

```
$ curl -s https://scarbs.xyz/api/v1/index/sn/fo/snforge_scarb_plugin.json | jq --arg version 0.31.1 '[.[] | select(.v == $version)] | length > 0'
false

$ curl -s https://scarbs.xyz/api/v1/index/sn/fo/snforge_scarb_plugin.json | jq --arg version 31.31.1 '[.[] | select(.v == $version)] | length > 0'
false
```

So this returns false not matter if the version exists or not. Maybe I messed something though?

.github/workflows/release.yml
```diff
@@ -249,3 +249,53 @@ jobs:
...
+          echo "snforge_scarb_plugin=$snforge_scarb_plugin"
```
THenry14, line 266: nit: you can add -x to set flags for this script, and get rid of this echo (as you'll be able to see the checked version during the jq check)

.github/workflows/release.yml
```diff
@@ -249,3 +249,53 @@ jobs:
...
+      - uses: actions/checkout@v4
+
+      - uses: dtolnay/rust-toolchain@7b1c307e0dcbda6122208f10795a713336a9b35a
+        with:
+          toolchain: stable
+
+      - uses: software-mansion/setup-scarb@v1
+        with:
+          # todo: Use scarb stable version that support publishing plugins (after 2.8.4)
+          scarb-version: "nightly-2024-11-09"
```
THenry14, line 291: the same as in plugin workflow

scripts/set_plugin_version.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+PLUGIN_FILE_PATH="../crates/snforge-scarb-plugin/Scarb.toml"
+
+VERSION=$(grep version "$PLUGIN_FILE_PATH" | cut -d '"' -f 2)
+
+STD_FILE_PATH="../snforge_std/Scarb.toml"
```
THenry14, line 7:
```suggestion
PLUGIN_FILE_PATH="../crates/snforge-scarb-plugin/Scarb.toml"
STD_FILE_PATH="../snforge_std/Scarb.toml"

VERSION=$(grep version "$PLUGIN_FILE_PATH" | cut -d '"' -f 2)
```
nitpick

btw https://en.wikipedia.org/wiki/STD_(disambiguation) ( ͡° ͜ʖ ͡°)

Let's maybe use full `SNFORGE_STD_PATH` name

THenry14 (commented):
great job mate

.github/workflows/release.yml
```diff
@@ -249,3 +249,53 @@ jobs:
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           TAG: ${{ steps.create-release.outputs.computed-prefix }}${{ steps.create-release.outputs.version }}
+
+  check-plugin-version:
+    name: Check if plugin version exist in the registry and upload if not
```
cptartur, line 254: Should probably add `needs: [ create-release ]` so it only triggers on successful release.

.github/workflows/release.yml
```diff
@@ -249,3 +249,53 @@ jobs:
...
+
+      - name: Publish snforge_scarb_plugin
+        if: steps.check-version.outputs.snforge_scarb_plugin == 'false'
+        uses: ./.github/workflows/publish_plugin.yml
+        run: echo "Publishing snforge_scarb_plugin"
+
+  upload-to-registry:
+    name: Upload packages to the registry
+    runs-on: ubuntu-latest
+    needs: [ check-plugin-version, verify-version ]
```
cptartur, line 277:
```suggestion
    needs: [ create-release ]
```

How about this, I think then it will only run on successfully created release.

.github/workflows/release.yml
```diff
@@ -249,3 +249,53 @@ jobs:
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           TAG: ${{ steps.create-release.outputs.computed-prefix }}${{ steps.create-release.outputs.version }}
+
+  check-plugin-version:
```
DelevoXDG, line 253: Nit: 
```suggestion
  upload-plugin-version:
```
would be more accurate

.github/workflows/release.yml
DelevoXDG: It works, there's just no version `0.31.1` in the registry 😄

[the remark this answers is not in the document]
DelevoXDG: Nit: 
`snforge_scarb_plugin_exists`, `snforge_scarb_plugin_uploaded` or `snforge_scarb_plugin_published` would be most accurate since we're dealing with true or false statement

.github/workflows/publish_plugin.yml
```diff
@@ -0,0 +1,26 @@
+name: publish_plugin.yml
```
DelevoXDG, line 1:
```suggestion
name: Publish snforge_scarb_plugin
```

.github/workflows/release.yml
```diff
@@ -249,3 +249,53 @@ jobs:
...
+          snforge_scarb_plugin=$(curl -s https://scarbs.xyz/api/v1/index/sn/fo/snforge_scarb_plugin.json | jq --arg version $snforge_scarb_plugin_version '[.[] | select(.v == $version)] | length > 0')
+          echo "snforge_scarb_plugin=$snforge_scarb_plugin"
+          echo "snforge_scarb_plugin=$snforge_scarb_plugin" >> $GITHUB_OUTPUT
+
+      - name: Publish snforge_scarb_plugin
+        if: steps.check-version.outputs.snforge_scarb_plugin == 'false'
+        uses: ./.github/workflows/publish_plugin.yml
+        run: echo "Publishing snforge_scarb_plugin"
+
+  upload-to-registry:
```
DelevoXDG, line 274: Nit:
```suggestion
  publish-to-registry:
```
(for consistency)

.github/workflows/release.yml
DelevoXDG: > How about this, I think then it will only run on successfully created release.

Yes, but we still need to ensure that latest plugin version is released before publishing the stds.  

https://github.com/foundry-rs/starknet-foundry/pull/2461#discussion_r1847852740

.github/workflows/release.yml
THenry14: This is what I am

obraz

.github/workflows/publish_plugin.yml
```diff
@@ -0,0 +1,36 @@
...
+      - uses: dtolnay/rust-toolchain@7b1c307e0dcbda6122208f10795a713336a9b35a
+        with:
+          toolchain: stable
+
+      - uses: software-mansion/setup-scarb@v1
+        with:
+          scarb-version: "2.8.5"
+
+      - name: Publish snforge_scarb_plugin
+        if: steps.check-version.outputs.snforge_scarb_plugin_uploaded == 'false'
```
DelevoXDG, line 34: https://github.com/foundry-rs/starknet-foundry/pull/2461/files#r1853809134

This should also always run on `workflow_dispatch`

DelevoXDG (approved):
Please address the comments before merging

.github/workflows/release.yml
```diff
@@ -249,3 +249,35 @@ jobs:
...
+        with:
+          toolchain: stable
+
+      - uses: software-mansion/setup-scarb@v1
+        with:
+          scarb-version: "2.8.5"
+
+      - name: Publish sncast_std
+        working-directory: sncast_std
+        run: scarb publish --allow-dirty
```
cptartur, line 277: I don't think we need that here
```suggestion
        run: scarb publish
```

cptartur (approved):

.github/workflows/release.yml
kkawula: For unknown reasons, I got a warning about uncommitted changes if `workflow_call` was triggered. Leaving it as it is

Issue #1533, opened 2024-01-15
Implement docs tests in cairo from mdbook for forge


Arcticae:
### Which components does the task require to be changed?

snforge

### Description

Cairo book has this implemented and we can verify if their architecture is usable with our tests. If so, we can implement it, if not, let's design a custom solution (if this get's too complex, a design doc may be necessary).

piotmag769:
Maybe we can reuse this
https://github.com/cairo-book/cairo-book/tree/main?tab=readme-ov-file#work-locally-cairo-programs-verification
https://rust-lang.github.io/mdBook/format/mdbook.html#including-files

PR #2475, opened 2024-09-17
Add support overriding `block id/tag/hash` with a named fork binding

### LINKED ISSUE #2450: Support overriding block id/tag/hash with a named fork binding

snforge

### Description

We should be able to make this work when resolving the macro config: 

Image

kkawula:
Relates #2450 

## Introduced changes

- Added new fork config resolving option

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


Draggu (approved):

crates/snforge-scarb-plugin/src/utils.rs
```diff
@@ -1,30 +1,51 @@
...
+                }
+            }
+        )+
+
+        if let Some(result) = result {
+            Ok(result)
+        } else {
+            Err(Diagnostic {
+                message: $crate::utils::format_error_message(&messages),
+                severity: messages.into_iter().fold(Severity::Warning, |acc, x| $crate::utils::higher_severity(acc, x.severity))
```
cptartur, line 47: Nit but I think this could just be `reduce` since this branch will only execute if there is at least one message in `messages`.

cptartur (changes requested):
Good work, however this is missing documentation, an entry to CHANGELOG.md and most importantly some actual tests verifying this behaves as intended in `crates/forge/tests/integration/setup_fork.rs`.

crates/forge/src/run_tests/resolve_config.rs
```diff
@@ -121,6 +129,15 @@ fn replace_id_with_params(
...
         }
+        RawForkConfig::Overridden(overridden) => {
+            let fork_target_from_runner_config =
+                get_fork_target_from_runner_config(fork_targets, &overridden.name)?;
+
+            Ok(InlineForkConfig {
+                url: fork_target_from_runner_config.url.parse()?,
+                block: overridden.block,
+            })
+        }
```
integraledelebesgue, line 140: A tiny aesthetic `nit` since you can unpack the struct. Totally depending on your preferences:
```suggestion
        RawForkConfig::Overridden(OverridenForkConfig { name, block }) => {
            let fork_target_from_runner_config =
                get_fork_target_from_runner_config(fork_targets, &name)?;
                
            let url = fork_target_from_runner_config.url.parse()?;

            Ok(InlineForkConfig { url, block })
        }
```

integraledelebesgue (approved):

crates/forge/src/run_tests/resolve_config.rs
kkawula: It looks great

crates/snforge-scarb-plugin/src/utils.rs
kkawula: I suppose that `fold` will be more idiomatic in this case, we do not have to think about that there always will be at least one member in the list, we can avoid doing `uwrap` etc.

crates/snforge-scarb-plugin/src/utils.rs
```diff
@@ -1,30 +1,51 @@
...
-    })
+    formatdoc! {"
+        All options failed
+        {}
+        Resolve at least one of them
+    ", formatted_variants.join("\n")}
+}
+
+#[macro_export]
+macro_rules! branch {
```
ksew1, line 24: I don't know if this is a macro. It would be more readable as a function.

```rs
pub fn branch(results: Vec<Result<String, Diagnostic>>) -> Result<String, Diagnostic> {
    let mut messages: Vec<Diagnostic> = Vec::new();

    for result in results {
        match result {
            Ok(val) => return Ok(val),
            Err(err) => messages.push(err),
        }
    }

    Err(Diagnostic {
        message: format_error_message(&messages),
        severity: messages.iter().fold(Severity::Warning, |acc, diagnostic| {
            higher_severity(acc, diagnostic.severity)
        }),
    })
}
```

kkawula: Macro is lazy

ksew1: Still can be done with function though 

```rs
pub fn branch(results: Vec<Box<dyn Fn() -> Result<String, Diagnostic>>>) -> Result<String, Diagnostic> {
    let mut messages: Vec<Diagnostic> = Vec::new();

    for result in results {
        match result() {
            Ok(val) => return Ok(val),
            Err(err) => messages.push(err),
        }
    }

    Err(Diagnostic {
        message: format_error_message(&messages),
        severity: messages.iter().fold(Severity::Warning, |acc, diagnostic| {
            higher_severity(acc, diagnostic.severity)
        }),
    })
}

```

kkawula: I'll stay with the macro because it has better readability

crates/snforge-scarb-plugin/src/utils.rs
kkawula: Docs are blocked by #1533, page about forks is reconstructed there. I'll wait till they're done

crates/snforge-scarb-plugin/src/utils.rs
ksew1: No problem 😄. But I think you are mistaking verbosity for readability."

ksew1 (approved):

crates/snforge-scarb-plugin/src/utils.rs
ksew1: `Nit`: Please rename this x to more descriptive name

crates/forge/tests/integration/setup_fork.rs
```diff
@@ -178,6 +178,97 @@ fn fork_aliased_decorator() {
...
+                        )
+                        .unwrap()
+                        .join(VERSIONED_PROGRAMS_DIR),
+                    }),
+                }),
+                fork_targets: vec![ForkTarget::new(
+                    "FORK_NAME_FROM_SCARB_TOML".to_string(),
+                    node_rpc_url().to_string(),
+                    "tag".to_string(),
+                    "latest".to_string(),
```
cptartur, line 262: I'd change it to something stable. Let's use specific block number, just different that it is used to override in test.

crates/snforge-scarb-plugin/src/utils.rs
```diff
@@ -1,30 +1,55 @@
 use cairo_lang_macro::{Diagnostic, Severity};
 use indoc::formatdoc;
 
-fn higher_severity(a: Severity, b: Severity) -> Severity {
+pub fn higher_severity(a: Severity, b: Severity) -> Severity {
```
cptartur, line 4: Why was this done?

kkawula: This has to be changed because the [macro](https://github.com/foundry-rs/starknet-foundry/pull/2475/files#diff-13330e2784a18e63e1241cc441655b1332d63a2991f843ed73ae0242cfa2edf6R51) uses this function

cptartur (approved):

crates/snforge-scarb-plugin/src/utils.rs
kkawula: Docs added in #2500

PR #2486, opened 2024-09-19
Pad addresses to correct length

### LINKED ISSUE #2248: Pad the contract/account addresses and class hashes in the output

sncast

### Description

Wallets don't do this by their own and we should make sure we output the addresses not only correct numerically, but also correct in the terms of their length. They should be padded with zeros for display (especially for hex format).

### LINKED ISSUE #2424: sncast account commands print the address with incorrect length

sncast

### Foundry Version

0.28.0

MacOS

x86

### What happened

All `sncast account` commands I tried print the address with 63 characters after `0x`. There should be 64 characters(32 bytes). Accounts are missing a zero after `0x`. This is a simple error, but can make people that are unaware loose time since pasting this address on block explorers or faucets gives you an error about the address being invalid.

### Trace

Output of `sncast account create` on Sepolia
```sh
command: account create
address: 0x56dda3be03ae9013f6c6cafaede2834c524018164a6a42218c288aa544eeb8e
max_fee: 6048364968880
message: Account successfully created. Prefund generated address with at least <max_fee> STRK tokens or an equivalent amount of ETH tokens. It is good to send more in the case of higher demand.
```

cptartur:
Closes #2248 
Closes #2424 

## Introduced changes

- Addresses used in `sncast account create`, `sncast declare` and `sncast deploy` are now correctly padded to 66 (64 after `0x`) characters length.

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/sncast/src/response/structs.rs
```diff
@@ -24,6 +26,30 @@ where
     serializer.serialize_str(&format!("{value:#}"))
 }
 
+#[derive(Clone, Copy, Debug, PartialEq, Deserialize, CairoSerialize)]
+pub struct Address(pub Felt);
```
ddoktorski, line 30: I'm not a big fan of using the `Address` name since it is also used for `class_hash`. Generally without additional context, it could be confusing to understand why it was introduced. Apart from this, I think we should also print `transaction_hash` with padded `0` for consistency.

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 #### Fixed
 
 - `sncast declare` no longer fails for flat contracts (i.e. CASM artifacts with `bytecode_segment_lengths` being a number)
+- Addresses outputted when calling `sncast account create`, `sncast deploy` and `sncast declare` are now padded to the correct length and prefixed with `0x0`
```
ddoktorski, line 19: The term `correct length` implies that it has been incorrect so far, which I think is not true (but we would need to define what this means first). How about rephrasing it to note that it's now padded with 0 and always 64 characters long?

franciszekjob (commented):
Looks fine, let's just add tests for serializing padded felts.

crates/sncast/src/response/structs.rs
cptartur: Renamed to `PaddedFelt`

crates/conversions/src/lib.rs
```diff
@@ -7,10 +7,13 @@ pub mod entrypoint_selector;
 pub mod eth_address;
 pub mod felt252;
 pub mod nonce;
+pub mod padded_felt;
 pub mod primitive;
 pub mod serde;
 pub mod string;
 
+extern crate self as conversions;
```
cptartur, line 15: This was added to use `CairoSerialize` derives within the crate itself.

crates/sncast/src/response/print.rs
```diff
@@ -87,6 +87,9 @@ impl Format for OutputValue {
                 if let Ok(field) = Felt::from_str(&input) {
                     return match numbers {
                         NumbersFormat::Decimal => OutputValue::String(format!("{field:#}")),
+                        NumbersFormat::Hex if input.len() == 66 && input.starts_with("0x0") => {
+                            OutputValue::String(input)
+                        }
```
cptartur, line 92: I guess this might not be 100% foolproof but with current printing setup it's the best one I found working.

ddoktorski (approved):
LGTM

crates/sncast/src/state/state_file.rs
```diff
@@ -443,8 +445,8 @@ mod tests {
         let transaction1 = ScriptTransactionEntry {
             name: "declare".to_string(),
             output: ScriptTransactionOutput::DeclareResponse(DeclareResponse {
-                class_hash: Felt::try_from_hex_str("0x1").unwrap(),
-                transaction_hash: Felt::try_from_hex_str("0x2").unwrap(),
+                class_hash: PaddedFelt(Felt::try_from_hex_str("0x1").unwrap()),
+                transaction_hash: Felt::try_from_hex_str("0x2").unwrap().into_(),
```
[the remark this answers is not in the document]
cptartur: I'll change it. it's probably a leftover from some previous version

PR #2494, opened 2024-09-20
Validate Scarb.toml and Scarb version when running coverage

ksew1:
Closes #https://github.com/software-mansion/cairo-coverage/issues/48

## Introduced changes

- `snforge` now validates if your project will generate needed debug info for `cairo-coverage` when running  `--coverage` flag

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/forge-runner/src/coverage_api.rs
```diff
@@ -49,3 +54,50 @@ pub fn run_coverage(saved_trace_data_paths: &[PathBuf]) -> Result<()> {
...
+    ensure!(
+        has_need_entries,
+        formatdoc! {
+            "Scarb.toml must have the following Cairo compiler configuration to run coverage:
+
+            [profile.{profile}.cairo]
+            unstable-add-statements-functions-debug-info = true
+            unstable-add-statements-code-locations-debug-info = true
+            inlining-strategy = \"avoid\"
+            ... other entries ...
```
cptartur, line 85:
```suggestion
            "Scarb.toml must contain the following Cairo compiler configuration to run coverage:

            [profile.{profile}.cairo]
            unstable-add-statements-functions-debug-info = true
            unstable-add-statements-code-locations-debug-info = true
            inlining-strategy = \"avoid\"
```

crates/forge-runner/src/coverage_api.rs
```diff
@@ -49,3 +54,50 @@ pub fn run_coverage(saved_trace_data_paths: &[PathBuf]) -> Result<()> {
...
+            ",
+            profile = scarb_metadata.current_profile
+        },
+    );
+
+    ensure!(
+        scarb_metadata.app_version_info.version >= MINIMAL_SCARB_VERSION,
+        "Coverage generation requires scarb version >= {MINIMAL_SCARB_VERSION}",
+    );
```
cptartur, line 94: Nit, but this could be checked before we start validating the toml

crates/forge/tests/e2e/coverage.rs
```diff
@@ -13,3 +18,49 @@ fn test_coverage_project() {
...
+fn test_fail_on_scarb_version_lt_2_8_0() {
+    let temp = setup_package("coverage_project");
+
+    let output = test_runner(&temp).arg("--coverage").assert().failure();
+
+    assert_stdout_contains(
+        output,
+        "[ERROR] Coverage generation requires scarb version >= 2.8.0\n",
+    );
+}
```
cptartur, line 32: This test will have a different behavior depending on a local scarb setup

CHANGELOG.md
```diff
@@ -23,6 +23,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Project generated by `snforge` contains `assert_macros` dependency with version 0.1.0 for Scarb <= 2.8.0, otherwise equal to Scarb
 
+
+#### Changed
+
+- `snforge` now validates if your project will generate needed debug info for `cairo-coverage` when running  `--coverage` flag
```
cptartur, line 29: This could be a little bit more in depth about what it does I think

crates/forge-runner/src/coverage_api.rs
ksew1: I think I would leave that as it is, since people might think that configuration  should look exactly like this, which is not true.

crates/forge/tests/e2e/coverage.rs
ksew1: We have 2.7.0 in tool versions so this should always pass

PR #2497, opened 2024-09-23
Improve forge configuration acquired from tomls

### LINKED ISSUE #2433: Improve Forge configuration acquired from `.toml`s
### Which component is the issue related to?
Forge

### Issue
* Partially fixed by @piotmag769 in [PR](https://github.com/foundry-rs/starknet-foundry/pull/2432#issue-2497325264)
* Fork tests receive improper configuration (upper-case "Latest")
* Block id (e.g. "Latest") is hardcoded, we need to introduce constants
* Existing tests need to be improved
* We need to add e2e tests

kkawula:
Closes #2433

## Introduced changes

- Configuration parsing logic was moved to the stage of loading it from the file, instead of the previous version - when performing tests

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


kkawula:
In the linked issue there is listed `Block id (e.g. "Latest") is hardcoded, we need to introduce constants`, is this still required? 
Now we validate this before running tests and this value is used only in two places.

cptartur (changes requested):
Let's add a test to see if it's actually being correctly loaded from file. We can improve the overall flow of loading the configuration later.

kkawula:
Fortunately, these changes are still covered by older tests, e.g. here -> https://github.com/foundry-rs/starknet-foundry/blob/f30acafda387dfa85c72b2aeb9149a25f562b957/crates/forge/src/scarb.rs#L282-L463

kkawula:
Further changes are addressed in issue #2514

crates/forge/src/run_tests/resolve_config.rs
```diff
@@ -78,27 +77,6 @@ async fn resolve_fork_config(
...
 }
 
-fn parse_block_id(fork_target: &ForkTarget) -> Result<BlockId> {
-    let block_id = match fork_target.block_id_type.as_str() {
-        "number" => BlockId::BlockNumber(fork_target.block_id_value.parse()?),
-        "hash" => {
-            let block_hash = fork_target.block_id_value.parse::<BigInt>()?;
-
-            BlockId::BlockHash(block_hash.into())
-        }
```
cptartur, line 88: It seems like previously it was possible to pass felt as either decimal or hex. We should verify if that was true and if so if this is still possible

kkawula: It is possible, I've added tests

cptartur (approved):

crates/forge/src/scarb/config.rs
```diff
@@ -149,3 +158,106 @@ impl TryFrom<RawForgeConfig> for ForgeConfigFromScarb {
...
+        let name = "TestFork";
+        let url = "http://example.com";
+        let block_id_type = "tag";
+        let block_id_value = "latest";
+
+        let fork_target = ForkTarget::new(name, url, block_id_type, block_id_value).unwrap();
+
+        assert_eq!(fork_target.name, name);
+        assert_eq!(fork_target.url, Url::parse(url).unwrap());
+        if let BlockId::BlockTag = fork_target.block_id {
```
Draggu, line 215: @cptartur why approve XD

PR #2505, opened 2024-09-23
Remove todo for hash validation

### LINKED ISSUE #1179: Add validation of hash value in `BlockId::Hash` variant
Tagged with `TODO #1179`
We parsed the `BigInt` from db with `Felt252::from` which does sth like

```rs
value.try_into().unwrap_or_else(|_| &*CAIRO_PRIME_BIGUINT - (-value as u128)),
```

After moving test collection to Scarb, the appropriate range should be checked there, then we can pass the number as a String (see #1119)

franciszekjob:
Closes #1179 

## Introduced changes

Remove todo for hash validation. Block hash is calculated by [Poseidon hash](https://docs.starknet.io/architecture-and-concepts/network-architecture/block-structure/#block_hash) (for more please see [implementation](https://github.com/starkware-libs/sequencer/blob/bb361ec67396660d5468fd088171913e11482708/crates/starknet_api/src/block_hash/block_hash_calculator.rs#L68)). `BlockHash` holds value of type `StarkHash` which is simply `Felt` so ultimately there's no need to add extra range validation.

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests

### DIFF
--- a/crates/snforge-scarb-plugin/src/attributes/fork/block_id.rs
+++ b/crates/snforge-scarb-plugin/src/attributes/fork/block_id.rs
@@ -78,8 +78,6 @@ impl ParseFromExpr<(BlockIdVariants, &Expr)> for BlockId {
                     BlockIdVariants::Hash.as_ref(),
                 )?;
 
-                // TODO(#1179): Add hash range validation here
-
                 Ok(Self::Hash(hash))
             }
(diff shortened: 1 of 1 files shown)


Draggu (approved):
👍

cptartur (approved):

Issue #2408, opened 2024-08-28
Omit compilation of lib target when using `--no-optimization`


cptartur:
Right now we run Scarb compilation two times: First to build contracts used in tests and again to build test targets. We should implement some mechanism that allows the tests to be compiled alongside with tests artifacts.

1. This needs to be coordinated with and implemented in Scarb first
2. This feature should be configurable. Contracts behavior can change when building in test mode: We should still allow to do compilation in two separate steps to allow to test more closely to real network behavior.

First, we should create a design doc for this.

Arcticae:
@RealJohnnyTime we are working on a solution

ddoktorski:
The one thing left to do is to omit compiling `lib` target when `--no-optimization` flag is enabled. This should be done by leveraging `--target-kinds` flag in the `scarb build` command (added in Scarb 2.8.3).

PR #2507, opened 2024-09-24
Use contract artifacts produced by test target 

ddoktorski:
Towards #2408 

## Introduced changes

- Modified logic to compile contracts as part of the test target (for Scarb >= 2.8.3)
- Refactored code for getting contract artifacts
- Added `--no-optimization` flag
- Added separate set of basic tests for Scarb 2.8.3

## Checklist

- [X] Linked relevant issue
- [X] Added relevant tests


CHANGELOG.md
```diff
@@ -23,6 +23,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Project generated by `snforge` contains `assert_macros` dependency with version 0.1.0 for Scarb <= 2.8.0, otherwise equal to Scarb
 - Support for overriding fork configuration in test attribute with a different block ID, tag, or hash.
+- `--no-optimization` flag that can be used to build contracts using the [starknet contract target](https://docs.swmansion.com/scarb/docs/extensions/starknet/contract-target.html#starknet-contract-target. This is the default behavior when using Scarb < 2.8.3
+
+### Changed
+
+- For Scarb >= `2.8.3` contract artifacts are built as part of the test target now. This process speeds up the compilation time. It can be disabled using the [--no-optimization flag](https://foundry-rs.github.io/starknet-foundry/appendix/snforge/test.html#--no-optimization)
```
cptartur, line 30: Maybe let's mention that this may result in behavior not being 100% consistent with the real network

crates/forge/src/scarb.rs
```diff
@@ -8,12 +8,16 @@ use forge_runner::package_tests::TestTargetLocation;
...
 use scarb_metadata::{PackageMetadata, TargetMetadata};
 use scarb_ui::args::{FeaturesSpec, PackagesFilter};
+use semver::Version;
 use std::collections::HashMap;
 use std::fs::read_to_string;
 use std::io::ErrorKind;
 
 pub mod config;
 
+// TODO: Temporary set to 2.8.2, update to 2.8.3 after scarb release and before the PR merge
```
cptartur, line 18: TODO

crates/forge/test_utils/src/runner.rs
```diff
@@ -97,12 +97,16 @@ impl Contract {
...
-            get_contracts_artifacts_and_source_sierra_paths(&scarb_metadata, &package.id, None)
-                .unwrap()
-                .remove(&self.name)
-                .ok_or(anyhow!("there is no contract with name {}", self.name))?
-                .0;
+        let contract = get_contracts_artifacts_and_source_sierra_paths(
+            &scarb_metadata,
+            &package.id,
+            None,
+            false,
```
cptartur, line 104: Why is it hardcoded here?

crates/forge/tests/e2e/running.rs
```diff
@@ -150,7 +150,7 @@ fn with_failing_scarb_build() {
         output,
         indoc!(
             r"
-                [ERROR] Failed to build test artifacts with Scarb: `scarb` exited with error
+                [ERROR] Failed to build contracts with Scarb: `scarb` exited with error
```
cptartur, line 153: Why this error changed? Is it just because order of compilation changed?

crates/forge/tests/e2e/contract_artifacts.rs
```diff
@@ -0,0 +1,91 @@
...
+    [PASS] simple_package_integrationtest::test_simple::test_two [..]
+    [PASS] simple_package_integrationtest::test_simple::test_two_and_two [..]
+    [FAIL] simple_package_integrationtest::test_simple::test_failing
+
+    Failure data:
+        0x6661696c696e6720636865636b ('failing check')
+
+    [FAIL] simple_package_integrationtest::test_simple::test_another_failing
+
+    Failure data:
```
cptartur, line 52: Maybe it would make sense to assert that `.test.` json files are produced here?

crates/scarb-api/src/lib.rs
```diff
@@ -79,48 +79,54 @@ fn artifacts_for_package(path: &Utf8Path) -> Result<StarknetArtifacts> {
 ///
 /// Try getting the path to `starknet_artifacts.json` file that is generated by `scarb build` command.
 /// If the file is not present, for example when package doesn't define the starknet target, `None` is returned.
```
cptartur, line 81: This could be updated

crates/scarb-api/src/lib.rs
```diff
@@ -282,19 +288,88 @@ mod tests {
...
+            true,
+        )
+        .unwrap();
+
+        assert_eq!(
+            path,
+            temp.path()
+                .join("target/dev/basic_package_integrationtest.test.starknet_artifacts.json")
+        );
+    }
```
cptartur, line 371: Should we also assert that unittests are still there?

crates/forge/tests/e2e/running.rs
ddoktorski: Yup exactly, the order of compilation changed. Once we update scarb to 2.8.3 in the repository the error will be for "test artifacts" again

crates/scarb-api/src/lib.rs
ddoktorski: We can, but this test is specifically to check that when we have both contract artifacts for `unittest` and `integrationtest` then it should load `integrationtest`

crates/forge/test_utils/src/runner.rs
ddoktorski: This is used for tests. We build the contracts above using `scarb build`, currently at version `2.7.0`. Generally this interface, which uses boolean to decide whether artifacts should be taken from the test target is not ideal. An alternative could be to split `get_contracts_artifacts_and_source_sierra_paths` into multiple functions, but this would require more effort.

cptartur: Okay, I didn't notice this was in test utils

crates/scarb-api/src/lib.rs
cptartur: Okay, this is not 100% clear that this is the purpose of the test. Perhaps rename or restructure it in a way that it indicates that better.

ddoktorski: Renamed to `get_starknet_artifacts_path_for_test_build_when_integration_tests_exist`

PR #2516, opened 2024-09-25
Add `Scarb.toml` docs and template generation

### LINKED ISSUE #1017: Add `Scarb.toml` forge reference
Add Scarb.toml forge tool reference to documentation. It should show all possible configuration keys and values for forge with explanations.

Also generate the template with commented out parts, similar to `tsconfig.json`.

franciszekjob:
Closes #1017 

## Introduced changes

- Add `Scarb.toml` reference in the docs
- Overwrite `Scarb.toml` with possible configuration options when new project is generated by `snforge init`

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/forge/tests/e2e/running.rs
```diff
@@ -717,7 +730,7 @@ fn validate_init(temp: &TempDir) {
             sierra = true
 
             [scripts]
-            test = "snforge test"
+            test = "snforge test"{appended_content}
```
franciszekjob, line 733: note: When initializing project with `snforge init`, we need to append `Scarb.toml` template to the expected value, in case of `scarb new` we keep `Scarb.toml` as it is.

crates/forge/src/scarb/config.rs
```diff
@@ -6,6 +6,38 @@ use std::{
...
+# name = "SOME_SECOND_NAME"
+# url = "http://your.second.rpc.url"                         
+# block_id.number = "123"                                    # Block to fork from (block number)
+
+# [[tool.snforge.fork]]
+# name = "SOME_THIRD_NAME"
+# url = "http://your.third.rpc.url"
+# block_id.hash = "0x123"                                    # Block to fork from (block hash)
+
+# [profile.dev.cairo]                                        # Define Cairo compiler configuration section
```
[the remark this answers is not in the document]
franciszekjob: Sure, added.

docs/src/appendix/scarb-toml.md
```diff
@@ -0,0 +1,236 @@
+# The Manifest Format
+
+The `Scarb.toml` contains the package manifest that is needed in package compilation process. It can be used to provide configuration for Starknet Foundry Forge.
```
cptartur, line 3: Maybe let's also link to relevant scarb doc page here?

docs/src/appendix/scarb-toml.md
```diff
@@ -0,0 +1,236 @@
...
+The `fuzzer_runs` field specifies the number of runs of the random fuzzer.
+
+#### `fuzzer_seed`
+The `fuzzer_seed` field specifies the seed for the random fuzzer.
+
+#### Example of fuzzer configuration
+```toml
+[tool.snforge]
+fuzzer_runs = 1234
+fuzzer_seed = 1111
```
cptartur, line 31: Let's link to fuzzer section in docs

docs/src/appendix/scarb-toml.md
```diff
@@ -0,0 +1,236 @@
...
+block_id.hash = "0x123"
+```
+
+#### Example configuration with two forks
+```toml
+[[tool.snforge.fork]]
+name = "SOME_NAME"
+url = "http://your.rpc.url"
+block_id.tag = "latest"
+
```
cptartur, line 68: Let's link fork docs

docs/src/appendix/scarb-toml.md
```diff
@@ -0,0 +1,236 @@
...
+fuzzer_runs = 1234
+fuzzer_seed = 1111
+```
+
+### `[[tool.snforge.fork]]`
+```toml
+[[tool.snforge.fork]]
+# ...
+```
+Allows to configure forked tests. While in use, all below fields must be set at the same time.
```
cptartur, line 39:
```suggestion
Allows to configure forked tests. If defined, all fields outlined below must also be defined.
```

docs/src/appendix/scarb-toml.md
```diff
@@ -0,0 +1,236 @@
...
+block_id.tag = "latest"
+
+[[tool.snforge.fork]]
+name = "SOME_SECOND_NAME"
+url = "http://your.second.rpc.url"
+block_id.number = "123"
+```
+
+### `[profile.<dev|release>.cairo]`
+Adjust Cairo compiler configuration parameters when compiling this package. These options are not taken into consideration when this package is used as a dependency for another package. All fields are optional.
```
cptartur, line 76:
```suggestion
By default, these arguments do not need to be defined. Only set them to use profiler (link docs) or coverage (link docs).

Adjust Cairo compiler configuration parameters when compiling this package. These options are not taken into consideration when this package is used as a dependency for another package. All fields are optional.
```

docs/src/appendix/scarb-toml.md
```diff
@@ -0,0 +1,236 @@
...
+[profile.dev.cairo]
+unstable-add-statements-functions-debug-info = true
+```
+
+#### `inlining-strategy`
+This field is responsible for setting the inlining strategy to be used by compiler when building the package. The possible values are `default` or `avoid`. If avoid strategy is set, the compiler will only inline function annotated with `#[inline(always)]` attribute.
+```toml
+[profile.dev.cairo]
+inlining-strategy = "avoid"
+```
```
cptartur, line 103: Maybe let's just link to specific sections of scarb docs here instead of maintaining these ourselves?

docs/src/appendix/scarb-toml.md
```diff
@@ -0,0 +1,236 @@
...
+```toml
+# ...
+[dependencies]
+starknet = ">=2.8.2"
+openzeppelin = { git = "https://github.com/OpenZeppelin/cairo-contracts.git", branch = "cairo-2" }
+
+[[target.starknet-contract]]
+build-external-contracts = ["openzeppelin::account::account::Account"]
+# ...
+```
```
cptartur, line 182: While I get why this was included, this is kind of duplicating Scarb docs and adds an additional burden for us of maintaining it. 

Maybe let's just say that we handle `[features] and link to both a section in our docs that explain conditional compilation and to scarb docs? Similar on how scarb does it in their reference

image

Same goes for starknet contract target and maybe even for `unstable-...`.

crates/forge/tests/e2e/running.rs
```diff
@@ -727,9 +728,11 @@ fn validate_init(temp: &TempDir, validate_snforge_std: bool) {
 
             [scripts]
             test = "snforge test"
+            {SCARB_MANIFEST_TEMPLATE_CONTENT}
```
cptartur, line 731: Isn't this missing adding these contents?

crates/forge/src/scarb/config.rs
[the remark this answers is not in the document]
cptartur: Maybe let's just say that these should be enabled if you want to use profiling/coverage and link to documentation (or say to check it). These explanations are kind of long for the purpose of these comments

cptartur (changes requested):

crates/forge/tests/e2e/running.rs
franciszekjob: Resolved offline.

crates/forge/src/scarb/config.rs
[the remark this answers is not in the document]
franciszekjob: Reduced 👍 .

docs/src/appendix/scarb-toml.md
```diff
@@ -0,0 +1,245 @@
...
+}
+```
+
+Then update Scarb.toml so it includes the following lines:
+```toml
+[features]
+enable_for_tests = []
+```
+
+See [features](https://docs.swmansion.com/scarb/docs/reference/conditional-compilation.html#features) in Scarb docs.
```
cptartur, line 148: Let's also mention this at the top of the section

docs/src/appendix/scarb-toml.md
```diff
@@ -0,0 +1,245 @@
...
+```toml
+[[target.starknet-contract]]
+casm = true
+```
+
+See more about [targets](https://docs.swmansion.com/scarb/docs/reference/targets.html) in scarb docs.
+
+
+#### `build-external-contracts`
+The `build-external-contracts` accepts a list of strings, each of which is a reference to a contract defined in a dependency. The package that implements this contracts need to be declared as a dependency of the project in `[dependencies]`.
```
cptartur, line 175: This is a super formal description but doesn't really say what's the purpose of this in terms of forge.

Let's say that this is needed to use the contracts from dependencies in tests and also link to relevant scarb doc. You can keep the mention that these packages also need to be defined as dependencies, but this can be reworded.

crates/forge/src/scarb/config.rs
```diff
@@ -8,6 +8,38 @@ use std::{
...
+# unstable-add-statements-code-locations-debug-info = true   # Should be used if you want to use coverage
+# unstable-add-statements-functions-debug-info = true        # Should be used if you want to use coverage/profiler
+# inlining-strategy = "avoid"                                # Should be used if you want to use coverage
+
+# [features]                                                 # Define features section
```
cptartur, line 39: Let's say that this is used for conditional compilation. User should be able to look it up easily in docs.

cptartur (changes requested):

docs/src/appendix/scarb-toml.md
franciszekjob: Hope this is better.

> The `build-external-contracts` allows to use contracts from your dependencies inside your tests. It accepts a list of strings, each of which is a reference to a contract defined in a dependency. You need to add dependency which  implements this contracts  to your `Scarb.toml`.

I've added a sentence which explains what this option actually does in regards of forge, should be clearer now.

docs/src/appendix/scarb-toml.md
```diff
@@ -0,0 +1,248 @@
...
+
+Then update Scarb.toml so it includes the following lines:
+```toml
+[features]
+enable_for_tests = []
+```
+
+### `[[target.starknet-contract]]`
+Scarb supports registering targets that are handled by Scarb extensions. Such targets are called external. All fields are optional. See more about [targets](https://docs.swmansion.com/scarb/docs/reference/targets.html) in Scarb documentation.
+
```
cptartur, line 154: Let's just focus on `starknet-contract` target and say that this is required to build starknet contracts and link to this https://docs.swmansion.com/scarb/docs/extensions/starknet/contract-target.html#starknet-contract-target page. 

`sierra` and `casm` can just link to relevant subsections.

cptartur (changes requested):

docs/src/appendix/scarb-toml.md
franciszekjob: Ok, looks as below now:
image

cptartur (approved):

PR #2528, opened 2024-09-26
Implemented kosher data transformer

integraledelebesgue:
Things to do:
* more integration tests for transformation itself


integraledelebesgue:
It's done 🥹

PR #2529, opened 2024-09-26
Set `global` to `true` in `Cli` args

### LINKED ISSUE #2492: Use `global` for cast common arguments using clap

sncast

### Description

It should be possible to pass common arguments both before and after subcommand name, e.g. both `sncast --account default declare --contract_name name ...` and `sncast declare --account default --contract_name name ...` should be allowed.

This can be achieved using this method https://docs.rs/clap/latest/clap/struct.Arg.html#method.global

franciszekjob:
Closes #2492 

## Introduced changes

Allow common arguments to be used after subcommand.

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/sncast/tests/e2e/invoke.rs
```diff
@@ -52,6 +52,45 @@ async fn test_happy_case(account: &str) {
...
+        "put",
+        "--calldata",
+        "0x1 0x2",
+        "--max-fee",
+        "99999999999999999",
+        "--fee-token",
+        "eth",
+    ];
+
+    let snapbox = runner(&args);
```
ksew1, line 85:
```suggestion
    runner(&args).assert().success();
```

It is enough to check if it works. Still, we kind of check the clap functionality, so I don't think this test is necessary at all.

franciszekjob: Suggestion is reasonable, we can just check if it succeeds. 
This test will fail if we stop supporting the `global` (for some reason), so I think we should keep it.

crates/sncast/tests/e2e/invoke.rs
```diff
@@ -52,6 +52,39 @@ async fn test_happy_case(account: &str) {
     assert!(matches!(receipt, Invoke(_)));
 }
 
+#[test_case("oz_cairo_0"; "cairo_0_account")]
+#[test_case("oz_cairo_1"; "cairo_1_account")]
+#[test_case("oz"; "oz_account")]
+#[test_case("argent"; "argent_account")]
+#[test_case("braavos"; "braavos_account")]
+#[tokio::test]
+
```
cptartur, line 61:
```suggestion
```

cptartur (approved):

PR #2531, opened 2024-09-27
Explain `assert_macros` in the guide

### LINKED ISSUE #2481: Explain `assert_macros` in the guide

snforge

### Description

This needs to be described (and perhaps could also be added automatically to dev deps when generating the template?)

kkawula:
Closes #2481

## Introduced changes

- Added info about `assert_macros` in docs

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests

### DIFF
--- a/docs/src/getting-started/scarb.md
+++ b/docs/src/getting-started/scarb.md
@@ -26,9 +26,9 @@ Last but not least, remember that in order to use Starknet Foundry, you must hav
 ### `assert_macros`
 > ⚠️ **Recommended only for development** ️⚠️
 > 
->Assert macros package provides a set of macros that can be used to write assertions.
+>Assert macros package provides a set of macros that can be used to write assertions such as `assert_eq!`.
 In order to use it, your project must have the `assert_macros` dependency added to the `Scarb.toml` file.
-These macros are very expensive to run on Starknet, they generate a huge amount of steps and are not recommended for production use. 
+These macros are very expensive to run on Starknet, as they result a huge amount of steps and are not recommended for production use. 
 They are only meant to be used in tests.
 For snforge `v0.31.0` and later, this dependency is added by default. But for earlier versions, you need to add it manually.
 
--- a/docs/src/getting-started/scarb.md
+++ b/docs/src/getting-started/scarb.md
@@ -21,3 +21,27 @@ described [here](https://docs.swmansion.com/scarb/docs/extensions/testing.html#u
 
 Last but not least, remember that in order to use Starknet Foundry, you must have Scarb
 [installed](https://docs.swmansion.com/scarb/download.html) and added to the `PATH` environment variable.
+
+
+### `assert_macros`
+> ⚠️ **Recommended only for development** ️⚠️
+> 
+>Assert macros package provides a set of macros that can be used to write assertions.
+In order to use it, your project must have the `assert_macros` dependency added to the `Scarb.toml` file.
+These macros are very expensive to run on Starknet, they generate huge amount of steps and are not recommended for production use. 
+They are only meant to be used in tests.
+For snforge `v0.31.0` and later, this dependency is added by default. But for earlier versions, you need to add it manually.
+
+```toml
+[dev-dependencies]
+snforge_std = ...
+assert_macros = "<scarb-version>"
+```
+
+Available assert macros are 
+- `assert_eq!`
+- `assert_ne!`
+- `assert_lt!`
+- `assert_le!`
+- `assert_gt!`
+- `assert_ge!`
--- a/docs/src/getting-started/scarb.md
+++ b/docs/src/getting-started/scarb.md
@@ -26,11 +26,11 @@ Last but not least, remember that in order to use Starknet Foundry, you must hav
 ### `assert_macros`
 > ⚠️ **Recommended only for development** ️⚠️
 > 
->Assert macros package provides a set of macros that can be used to write assertions such as `assert_eq!`.
+>***Assert macros package provides a set of macros that can be used to write assertions such as `assert_eq!`.
 In order to use it, your project must have the `assert_macros` dependency added to the `Scarb.toml` file.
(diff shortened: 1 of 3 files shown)


docs/src/getting-started/scarb.md
```diff
@@ -21,3 +21,27 @@ described [here](https://docs.swmansion.com/scarb/docs/extensions/testing.html#u
 
 Last but not least, remember that in order to use Starknet Foundry, you must have Scarb
 [installed](https://docs.swmansion.com/scarb/download.html) and added to the `PATH` environment variable.
+
+
+### `assert_macros`
+> ⚠️ **Recommended only for development** ️⚠️
+> 
+>Assert macros package provides a set of macros that can be used to write assertions.
```
cptartur, line 29:
```suggestion
>Assert macros package provides a set of macros that can be used to write assertions such as `assert_eq!`.
```

docs/src/getting-started/scarb.md
```diff
@@ -21,3 +21,27 @@ described [here](https://docs.swmansion.com/scarb/docs/extensions/testing.html#u
...
+In order to use it, your project must have the `assert_macros` dependency added to the `Scarb.toml` file.
+These macros are very expensive to run on Starknet, they generate a huge amount of steps and are not recommended for production use.
```
cptartur, line 31:
```suggestion
These macros are very expensive to run on Starknet, as they result a huge amount of steps and are not recommended for production use. 
```

docs/src/getting-started/scarb.md
```diff
@@ -21,3 +21,27 @@ described [here](https://docs.swmansion.com/scarb/docs/extensions/testing.html#u
...
 [installed](https://docs.swmansion.com/scarb/download.html) and added to the `PATH` environment variable.
+
+
+### `assert_macros`
+> ⚠️ **Recommended only for development** ️⚠️
+> 
+>Assert macros package provides a set of macros that can be used to write assertions.
+In order to use it, your project must have the `assert_macros` dependency added to the `Scarb.toml` file.
+These macros are very expensive to run on Starknet, they generate a huge amount of steps and are not recommended for production use. 
+They are only meant to be used in tests.
```
cptartur, line 32: This could be bold

docs/src/getting-started/scarb.md
```diff
@@ -21,3 +21,27 @@ described [here](https://docs.swmansion.com/scarb/docs/extensions/testing.html#u
...
+assert_macros = "<scarb-version>"
+```
+
+Available assert macros are 
+- `assert_eq!`
+- `assert_ne!`
+- `assert_lt!`
+- `assert_le!`
+- `assert_gt!`
+- `assert_ge!`
```
cptartur, line 47: Do we want to list them all if we are not documenting them?

docs/src/getting-started/scarb.md
cptartur, line 1: I think this would make more sense either in "First Steps With Starknet Foundry` or in "Writing Tests".

Also, I think `assert_eq!` is used extensively in "Using Cheatcodes" section so maybe it would make sense to add a link to part you written in the place where it is first used.

cptartur (changes requested):

docs/src/getting-started/scarb.md
integraledelebesgue: There isn't any other documentation explaining them so I think we should list them and even provide a short description for each

kkawula: I think the semantics are pretty obvious, eg. lt, le. So this list is enough

docs/src/testing/testing.md
```diff
@@ -105,3 +105,26 @@ Tests: 0 passed, 0 failed, 0 skipped, 1 ignored, 0 filtered out
 
 To run only tests marked with the  `#[ignore]` attribute use `snforge test --ignored`.
 To run all tests regardless of the `#[ignore]` attribute use `snforge test --include-ignored`.
+
+## `assert_macros`
```
cptartur, line 109:
```suggestion
## Writing Assertions and `assert_macros` Package
```

docs/src/testing/testing.md
```diff
@@ -105,3 +105,26 @@ Tests: 0 passed, 0 failed, 0 skipped, 1 ignored, 0 filtered out
...
+> ⚠️ **Recommended only for development** ️⚠️
+> 
+>***Assert macros package provides a set of macros that can be used to write assertions such as `assert_eq!`.
+In order to use it, your project must have the `assert_macros` dependency added to the `Scarb.toml` file.
+These macros are very expensive to run on Starknet, as they result a huge amount of steps and are not recommended for production use. 
+They are only meant to be used in tests.
+For snforge `v0.31.0` and later, this dependency is added by default. But for earlier versions, you need to add it manually.***
```
cptartur, line 116:
```suggestion
For snforge `v0.31.0` and later, this dependency is added automatically when creating a project using `snforge init`. But for earlier versions, you need to add it manually.***
```

docs/src/getting-started/scarb.md
cptartur: I think we should ink it in the first place where it is used. I don't think this was resolved.

kkawula: I've added a warning in "Using Cheatcodes", other usages occur in each of the "snforge Advanced Features" sections eg. "Fuzz", and "Fork" testing. So I would skip them and assert that 'advanced' users know about 'assert_macros', to avoid text duplication. wdyt?

cptartur (approved):

docs/src/testing/testing.md
```diff
@@ -105,3 +105,26 @@ Tests: 0 passed, 0 failed, 0 skipped, 1 ignored, 0 filtered out
...
+assert_macros = "<scarb-version>"
+```
+
+Available assert macros are 
+- `assert_eq!`
+- `assert_ne!`
+- `assert_lt!`
+- `assert_le!`
+- `assert_gt!`
+- `assert_ge!`
```
piotmag769, line 130: We shouldn't document stuff that we don't control at all

Issue #2403, opened 2024-08-27
Print details about USC compilation process when running tests


cptartur:
### Which components does the task require to be changed? (think hard pls)

snforge

### Description

Right now it doesn't give any feedback. If USC compilation is long running it might see like a forge problem.

piotmag769:
I think it should be printed by USC binary btw

PR #2534, opened 2024-09-30
Keep sierra file path in `TestTargetRaw`

ddoktorski:
Towards #2403

## Introduced changes

- Keep sierra file path in `TestTargetRaw` so it can be printed later


crates/universal-sierra-compiler-api/src/lib.rs
```diff
@@ -37,49 +37,44 @@ pub struct AssembledCairoProgramWithSerde {
...
-pub fn compile_sierra_to_casm(sierra_program: &Program) -> Result<AssembledProgramWithDebugInfo> {
-    let assembled_with_info_raw = compile_sierra(
-        &serde_json::to_value(sierra_program)?,
-        None,
-        &SierraType::Raw,
-    )?;
+pub fn compile_sierra_to_casm(
+    sierra_program: &Program,
+    sierra_file_path: Option<&Utf8PathBuf>,
+) -> Result<AssembledProgramWithDebugInfo> {
```
franciszekjob, line 43: nit: Can we maintain same order of params as fields in `TestTargetRaw`? 
```suggestion
pub fn compile_sierra_to_casm(
    sierra_file_path: Option<&Utf8PathBuf>,
    sierra_program: &Program,
) -> Result<AssembledProgramWithDebugInfo> {
```

franciszekjob (approved):

cptartur (changes requested):

crates/universal-sierra-compiler-api/src/lib.rs
```diff
@@ -37,49 +37,44 @@ pub struct AssembledCairoProgramWithSerde {
...
-    )?;
+pub fn compile_sierra_to_casm(
+    sierra_file_path: Option<&Utf8PathBuf>,
+    sierra_program: &Program,
+) -> Result<AssembledProgramWithDebugInfo> {
+    let assembled_with_info_raw = if let Some(sierra_file_path) = sierra_file_path {
+        compile_sierra_at_path(sierra_file_path, &SierraType::Raw)?
+    } else {
+        compile_sierra(&serde_json::to_value(sierra_program)?, &SierraType::Raw)?
+    };
```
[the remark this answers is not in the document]
ddoktorski: Thanks for the comment. I decided to modify the interface in the following way:
- Removed completely `compile_sierra_to_casm` function
- Added a function `compile_sierra_program`

Basically, the point is to use `compile_sierra_at_path` where possible and for unsaved sierra use `compile_sierra_program` for raw sierra nad `compile_sierra` for either raw sierra or contracts. Other than that I also implemented `From<String>` for `AssembledProgramWithDebugInfo` to manage deserialization separately. Let me know what you think.

crates/universal-sierra-compiler-api/src/lib.rs
```diff
@@ -37,49 +36,47 @@ pub struct AssembledCairoProgramWithSerde {
...
 
-pub fn compile_sierra_to_casm(sierra_program: &Program) -> Result<AssembledProgramWithDebugInfo> {
-    let assembled_with_info_raw = compile_sierra(
-        &serde_json::to_value(sierra_program)?,
-        None,
-        &SierraType::Raw,
-    )?;
-    let assembled_with_info: AssembledProgramWithDebugInfo =
-        serde_json::from_str(&assembled_with_info_raw)?;
+pub trait UscOutputTransformer {
```
cptartur, line 39: Nit, maybe rename it to just `UscOutput`

Function could still be named `transform`, but probably something like `convert` could work as well.

crates/universal-sierra-compiler-api/src/lib.rs
```diff
@@ -37,49 +36,47 @@ pub struct AssembledCairoProgramWithSerde {
...
-) -> Result<String> {
+) -> Result<T> {
     let mut temp_sierra_file = Builder::new().tempfile()?;
-    let _ = temp_sierra_file.write(serde_json::to_vec(sierra_contract_class)?.as_slice())?;
+    let _ = temp_sierra_file.write(serde_json::to_vec(sierra_json)?.as_slice())?;
 
     compile_sierra_at_path(
-        temp_sierra_file.path().to_str().unwrap(),
-        current_dir,
+        Utf8Path::new(&temp_sierra_file.path().to_string_lossy().to_string()),
```
cptartur, line 63:
```suggestion
        Utf8Path::from_path(temp_sierra_file.path()).unwrap(),
```
Or instead of unwrap, you can convert it to anyhow error and bubble it. Not sure what makes more sense here, probably the latter since we are bubbling all other tmp file related error.

cptartur (changes requested):

crates/universal-sierra-compiler-api/src/lib.rs
ddoktorski: Renamed to `UniversalSierraCompilerOutput` and `convert`

cptartur (approved):

PR #2540, opened 2024-10-01
Fix duplicated short options

### LINKED ISSUE #2538: `sncast` CLI panics using debug profile

sncast

### Description

This was undetected using our CI because we are running with `--release` flag which doesn't do debug assertions. However, when running tests just with `sncast test`, cast fails with

```
Command add: Short option names must be unique for each argument, but '-a' is in use by both 'address' and 'account'
```

This is due to conflict introduced by global arguments.

- [ ] Resolve the conflict so it no longer fails
- [ ] Update tests in CI so we detect this kind of problems in the future. Probably this can be used to detect/test these https://docs.rs/clap/latest/clap/struct.Command.html#method.debug_assert

franciszekjob:
Closes #2538 

## Introduced changes

- Remove short option for `--accounts-file` flag.
- Change short option to `-d` for `--contract-address` flag.
- Add step to run one test in debug profile. This is because clap can suppress some errors in the release profile due to removed debug assertions.

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests

### DIFF
--- a/crates/sncast/src/main.rs
+++ b/crates/sncast/src/main.rs
@@ -76,7 +76,7 @@ struct Cli {
     account: Option<String>,
 
     /// Path to the file holding accounts info
-    #[clap(short = 'p', long = "accounts-file", global = true)]
+    #[clap(long = "accounts-file", global = true)]
     accounts_file_path: Option<Utf8PathBuf>,
 
     /// Path to keystore file; if specified, --account should be a path to starkli JSON account file
--- a/crates/sncast/src/main.rs
+++ b/crates/sncast/src/main.rs
@@ -72,7 +72,7 @@ struct Cli {
     /// Account to be used for contract declaration;
     /// When using keystore (`--keystore`), this should be a path to account file
     /// When using accounts file, this should be an account name
-    #[clap(short = 'a', long, global = true)]
+    #[clap(long, global = true)]
     account: Option<String>,
 
     /// Path to the file holding accounts info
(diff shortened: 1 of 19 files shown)


crates/sncast/src/starknet_commands/call.rs
```diff
@@ -11,7 +11,7 @@ use starknet::providers::{JsonRpcClient, Provider};
 #[command(about = "Call a contract instance on Starknet", long_about = None)]
 pub struct Call {
     /// Address of the called contract (hex)
-    #[clap(short = 'a', long)]
+    #[clap(short = 'd', long)]
```
cptartur, line 14: Maybe `-c` would make more sense? From `c`ontract

crates/sncast/src/main.rs
```diff
@@ -76,7 +76,7 @@ struct Cli {
     account: Option<String>,
 
     /// Path to the file holding accounts info
-    #[clap(short = 'f', long = "accounts-file", global = true)]
+    #[clap(long = "accounts-file", global = true)]
```
franciszekjob, line 79: note: imo it's better to get rid of short option if we can't use `f` or `a` (they are take by other flags) than to use something unrelated letter.

crates/sncast/src/starknet_commands/call.rs
franciszekjob: If you refer to `Call` only - it's possible, but I think it's better to maintain consistency and have short option 
 of `contract_address` the same for every command. Then it's not possible as  `c` is already used by `calldata` in `Invoke`.

cptartur (approved):

PR #2547, opened 2024-10-01
Implemented data transformation logic

integraledelebesgue:
**Stack**:
- #2555
- #2548
- #2547 ⬅
- #2546
- #2545

⚠️ *Part of a stack created by [spr](https://github.com/ejoffe/spr). Do not merge manually using the UI - doing so may have unexpected results.*

PR #2545, opened 2024-10-01
Implemented representations of Cairo `u256` and `u512`

integraledelebesgue:

PR #2548, opened 2024-10-01
Integrated data transformer with `sncast` commands

integraledelebesgue:
**Stack**:
- #2555
- #2548 ⬅
- #2547
- #2546
- #2545

### DIFF
--- a/crates/sncast/src/main.rs
+++ b/crates/sncast/src/main.rs
@@ -6,6 +6,7 @@ use crate::starknet_commands::{
 };
 use anyhow::{Context, Result};
 use configuration::load_global_config;
+use data_transformer::Calldata;
 use sncast::response::explorer_link::print_block_explorer_link_if_allowed;
 use sncast::response::print::{print_command_result, OutputFormat};
 
@@ -20,8 +21,8 @@ use sncast::helpers::scarb_utils::{
 };
 use sncast::response::errors::handle_starknet_command_error;
 use sncast::{
-    chain_id_to_network_name, get_account, get_block_id, get_chain_id, get_default_state_file_name,
-    NumbersFormat, ValidatedWaitParams, WaitForTx,
+    chain_id_to_network_name, get_account, get_block_id, get_chain_id, get_class_hash_by_address,
+    get_contract_class, get_default_state_file_name, NumbersFormat, ValidatedWaitParams, WaitForTx,
 };
 use starknet::accounts::ConnectedAccount;
 use starknet::core::utils::get_selector_from_name;
@@ -249,9 +250,19 @@ async fn run_async_command(
                 .try_into_fee_settings(&provider, account.block_id())
                 .await?;
 
+            // safe to unwrap because "constructor" is a standardized name
+            let selector = get_selector_from_name("constructor").unwrap();
+
+            let contract_class = get_contract_class(deploy.class_hash, &provider).await?;
+
+            let serialized_calldata = constructor_calldata
+                .map(|data| Calldata::from(data).serialized(contract_class, &selector))
+                .transpose()?
+                .unwrap_or_default();
+
             let result = starknet_commands::deploy::deploy(
                 deploy.class_hash,
-                &constructor_calldata,
+                &serialized_calldata,
                 deploy.salt,
                 deploy.unique,
                 fee_settings,
@@ -283,14 +294,21 @@ async fn run_async_command(
             let provider = rpc.get_provider(&config).await?;
 
             let block_id = get_block_id(&block_id)?;
+            let class_hash = get_class_hash_by_address(&provider, contract_address).await?;
+            let contract_class = get_contract_class(class_hash, &provider).await?;
 
-            let entry_point_selector = get_selector_from_name(&function)
+            let selector = get_selector_from_name(&function)
                 .context("Failed to convert entry point selector to FieldElement")?;
+
+            let serialized_calldata = calldata
+                .map(|data| Calldata::from(data).serialized(contract_class, &selector))
+                .transpose()?
+                .unwrap_or_default();
 
             let result = starknet_commands::call::call(
                 contract_address,
-                entry_point_selector,
-                calldata,
+                selector,
+                serialized_calldata,
                 &provider,
                 block_id.as_ref(),
             )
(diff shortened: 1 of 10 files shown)


crates/sncast/src/main.rs
```diff
@@ -249,9 +250,18 @@ async fn run_async_command(
                 .try_into_fee_settings(&provider, account.block_id())
                 .await?;
 
+            let selector = get_selector_from_name("constructor").unwrap();
+
+            let contract_class = get_contract_class(deploy.class_hash, &provider).await?;
```
cptartur, line 255: Below we are adding additional context to this, but here not, why is that?

crates/sncast/src/main.rs
```diff
@@ -284,13 +294,25 @@ async fn run_async_command(
...
 
+            let class_hash = get_class_hash_by_address(&provider, contract_address)
+                .await?
+                .with_context(|| {
+                    format!(
+                    	"Couldn't retrieve class hash of a contract with address {contract_address:#x}"
+                    )
+                })?;
+
+            let contract_class = get_contract_class(class_hash, &provider).await?;
```
cptartur, line 305: Is calldata not optional here? What in case of functions that have no arguments, do we event support these?

Can be a separate issue.

crates/sncast/src/main.rs
```diff
@@ -331,9 +353,18 @@ async fn run_async_command(
             let selector = get_selector_from_name(&function)
                 .context("Failed to convert entry point selector to FieldElement")?;
 
+            let class_hash = get_class_hash_by_address(&provider, contract_address)
+                .await
+                .with_context(|| format!("Couldn't retrieve class hash of a contract with address {contract_address:#x}"))?
+                .with_context(|| format!("Couldn't retrieve class hash of a contract with address {contract_address:#x}"))?;
```
cptartur, line 359: This is duplicated

crates/sncast/src/starknet_commands/call.rs
```diff
@@ -18,9 +18,9 @@ pub struct Call {
     #[clap(long)]
     pub function: String,
 
-    /// Arguments of the called function (list of hex)
+    /// Arguments of the called function (a list of Cairo-like expressions or values serialized as felts)
```
cptartur, line 21: I don't think `Cairo-like expression` is a term all users would understand. Perhaps this should show an example, be explained or just reworded?

crates/sncast/tests/e2e/deploy.rs
```diff
@@ -308,7 +308,7 @@ fn test_wrong_calldata() {
         output,
         indoc! {r"
         command: deploy
-        error: An error occurred in the called contract[..]Failed to deserialize param #2[..]
+        error: An error occurred in the called contract[..]('Input too long for arguments')[..]
```
cptartur, line 311: Why was this changed?

crates/sncast/tests/e2e/deploy.rs
```diff
@@ -330,14 +330,11 @@ async fn test_contract_not_declared() {
...
-    let output = snapbox.assert().success();
+    let output = snapbox.assert().failure();
 
     assert_stderr_contains(
         output,
-        indoc! {r"
-        command: deploy
-        error: An error occurred in the called contract[..]Class with hash[..]is not declared[..]
-        "},
+        "Error: Couldn't retrieve contract class with hash: 0x1",
```
cptartur, line 337: This is kind of "breaking" in terms of that we are loosing an explanation of why is this contract missing. Previously it said that it is not required. I'm not sure if that's good.

crates/sncast/tests/e2e/invoke.rs
```diff
@@ -310,14 +310,11 @@ async fn test_contract_does_not_exist() {
...
-    let output = snapbox.assert().success();
+    let output = snapbox.assert().failure();
 
     assert_stderr_contains(
         output,
-        indoc! {r"
-        command: invoke
-        error: An error occurred in the called contract[..]Requested contract address[..]is not deployed[..]
-        "},
+        "Error: Couldn't retrieve class hash of a contract with address 0x1",
```
cptartur, line 317: Same here

crates/sncast/tests/e2e/main_tests.rs
```diff
@@ -27,14 +27,11 @@ async fn test_happy_case_from_sncast_config() {
...
-    let output = snapbox.assert().success();
+    let output = snapbox.assert().failure();
 
     assert_stderr_contains(
         output,
-        indoc! {r"
-        command: call
-        error: There is no contract at the specified address
-        "},
+        "Error: Couldn't retrieve class hash of a contract with address 0x0",
```
cptartur, line 34: And here

crates/sncast/src/main.rs
integraledelebesgue: Indeed. `get_class_hash_by_address` returns `Result<Option<Felt>>`. We need to handle both `Err` and `None`

crates/sncast/tests/e2e/deploy.rs
integraledelebesgue: Made an issue [#2554](https://github.com/foundry-rs/starknet-foundry/issues/2554#issue-2564353877)

cptartur (commented):
Don't like we are changing the errors so much. However, I think behavior to previous can still be retained, as I've outlined in the relevant comment.

crates/sncast/src/starknet_commands/call.rs
integraledelebesgue: Sure, I will rewrite it in other words

crates/sncast/tests/helpers/constants.rs
```diff
@@ -10,7 +10,7 @@ pub const NETWORK: &str = "testnet";
 pub const SEED: u32 = 1_053_545_548;
 
 // Block number used by devnet to fork the Sepolia testnet network in the tests
-pub const FORK_BLOCK_NUMBER: u32 = 68_000;
+pub const FORK_BLOCK_NUMBER: u32 = 100_000;
```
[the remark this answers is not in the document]
integraledelebesgue: That's because of the data transformer contract - we need it in our fork of Sepolia. It's been deployed later in the chain so I bumped the block number.

integraledelebesgue:
✓ Commit merged in pull request [#2555](https://github.com/foundry-rs/starknet-foundry/pull/2555)

PR #2565, opened 2024-10-07
Revamp account import

### LINKED ISSUE #2218: Revamp `account add` procedure

sncast

### Description

1. We should probably name it `account import` since it imports an account
2. We should validate the provided account details
3. We should have a visual guide (with screenshots) on how to do it from (argent/braavos/oz)
4. We should have a separate section for this in the documentation (preferably "Importing accounts from wallets?")
5. We should probably also cover how to import an account to a wallet
6. We should remove RPC URL and class hash (they are probably not needed - this needs to be verified thoroughly)
7. We should support configuring global keystores and leaving them in `cast` GLOBAL configuration (same concept as #2212), to have them as first class citizen as well (this can be deprecated in the future as UAFF gains adoption, but for now keystore should have better UX)

franciszekjob:
Closes #2218

## Introduced changes

- Update `account add` to `account import`
- Remove `--public-key` flag
- Add address computation and validation based on provided salt, class hash, private key and account type
- Allow to pass private key in interactive mode when neither `--private-key` nor `--private-key-file` is passed
- Add "Importing Accounts" guide section

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/sncast/src/starknet_commands/account/import.rs
```diff
@@ -16,13 +17,13 @@ use starknet::signers::SigningKey;
...
-    /// Name of the account to be added
+pub struct Import {
+    /// Name of the account to be imported
     #[clap(short, long)]
     pub name: String,
 
     /// Address of the account
-    #[clap(short, long, requires = "private_key_input")]
+    #[clap(short, long)]
     pub address: Felt,
```
franciszekjob, line 27: note: `private_key_input` is not required for `address` now because user will be asked for private key  in interactive mode if it's not passed as arg.

crates/sncast/src/starknet_commands/account/import.rs
```diff
@@ -98,36 +98,62 @@ pub async fn add(
...
     };
 
-    let legacy = check_if_legacy_contract(class_hash, add.address, provider).await?;
+    let chain_id = get_chain_id(provider).await?;
+    if import.salt.is_some() && class_hash.is_some() {
+        let sncast_account_type = match import.account_type {
+            AccountType::Argent => SNCastAccountType::Argent,
+            AccountType::Braavos => SNCastAccountType::Braavos,
+            AccountType::Oz => SNCastAccountType::OpenZeppelin,
+        };
```
franciszekjob, line 107: note: As described in https://github.com/foundry-rs/starknet-foundry/issues/2571, there are two `AccountType` enums, unfortunately we need to map one to another here. This should be also refactored in mentioned issue.

crates/sncast/src/starknet_commands/account/import.rs
```diff
@@ -139,3 +158,9 @@ fn get_private_key_from_file(file_path: &Utf8PathBuf) -> Result<Felt> {
     let private_key_string = std::fs::read_to_string(file_path.clone())?;
     Ok(private_key_string.parse()?)
 }
+
+fn get_private_key_from_input() -> Felt {
+    let private_key =
+        rpassword::prompt_password("Enter private key: ").expect("Failed to read private key");
+    Felt::try_from_hex_str(&private_key).expect("Failed to parse private key into Felt")
+}
```
franciszekjob, line 166: note: I didn't add test for entering private key interactively on purpose, because  `prompt_password` uses TTY and adding input with `.stdin("<private_key>")` will not work. Maybe I should add a test which is skipped when running all tests?

crates/sncast/src/starknet_commands/account/import.rs
```diff
@@ -139,3 +154,9 @@ fn get_private_key_from_file(file_path: &Utf8PathBuf) -> Result<Felt> {
     let private_key_string = std::fs::read_to_string(file_path.clone())?;
     Ok(private_key_string.parse()?)
 }
+
+fn get_private_key_from_input() -> Felt {
+    let private_key =
+        rpassword::prompt_password("Enter private key: ").expect("Failed to read private key");
+    Felt::try_from_hex_str(&private_key).unwrap()
+}
```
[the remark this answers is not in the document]
cptartur: This should be definetely tested somehow. I've tried it locally and it fails on invalid felt input. 

1. We should test that 
2. We should handle felt errors instead of unwrapping.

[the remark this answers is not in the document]
cptartur: Also this should allow integer input as well, not only hex

crates/sncast/src/starknet_commands/account/import.rs
```diff
@@ -58,36 +57,33 @@ pub struct Add {
...
+    let private_key = match (&import.private_key, &import.private_key_file_path) {
+        (Some(passed_private_key), None) => passed_private_key,
+        (None, Some(passed_private_key_file_path)) => &{
+            get_private_key_from_file(passed_private_key_file_path).with_context(|| {
+                format!("Failed to obtain private key from the file {passed_private_key_file_path}")
+            })?
+        },
+        (None, None) => &get_private_key_from_input(),
+        _ => unreachable!("Checked on clap level"),
     };
```
cptartur, line 75: I'm not a fan of this construct forcing matches like this, why not write it as `if let(...) = {} else if let (...) = {} else {}`

crates/sncast/src/starknet_commands/account/import.rs
```diff
@@ -58,36 +57,33 @@ pub struct Add {
...
-            "The private key does not match the public key"
-        );
-    }
+    let private_key = &SigningKey::from_secret_scalar(*private_key);
 
-    let fetched_class_hash = get_class_hash_by_address(provider, add.address).await?;
+    let fetched_class_hash = get_class_hash_by_address(provider, import.address).await?;
     let deployed = fetched_class_hash.is_some();
-    let class_hash = match (fetched_class_hash, add.class_hash) {
+    let class_hash = match (fetched_class_hash, import.class_hash) {
```
cptartur, line 80: The whole match below is weird as well, why not write it as

```rust
let fetched_class_hash = get_class_hash_by_address(provider, import.address).await?;

if let Some(provided_class_hash) = import.class_hash {
    ensure!(
                fetched_class_hash == provided_class_hash,
                "Incorrect class hash {:#x} for account address {:#x}",
                from_user,
                import.address
            );
            fetched_class_hash
}
```

I don't understand in which case would we not be able to fetch the class_hash for given address and at the same time be able to correctly validate the one that the user provided. I think they must match, no?

cptartur: Also it should probably say that the incorrect class hash **was provided**.

crates/sncast/src/starknet_commands/account/import.rs
cptartur: Let's add relevant TODO to the code with issue number

crates/sncast/src/starknet_commands/account/import.rs
```diff
@@ -98,36 +94,55 @@ pub async fn add(
...
     };
 
-    let legacy = check_if_legacy_contract(class_hash, add.address, provider).await?;
+    let chain_id = get_chain_id(provider).await?;
+    if let (Some(salt), Some(class_hash)) = (import.salt, class_hash) {
+        let sncast_account_type = match import.account_type {
+            AccountType::Argent => SNCastAccountType::Argent,
+            AccountType::Braavos => SNCastAccountType::Braavos,
+            AccountType::Oz => SNCastAccountType::OpenZeppelin,
+        };
```
cptartur, line 103: In what case should it be possible for `class_hash` to be `None`? It should be just `Felt` and we shouldn't have `if let` it here.

crates/sncast/src/starknet_commands/account/import.rs
```diff
@@ -98,36 +94,55 @@ pub async fn add(
...
+        let computed_address =
+            compute_account_address(salt, private_key, class_hash, sncast_account_type, chain_id);
```
cptartur, line 105: If we successfully fetched the class_hash for the given address doesn't that mean it's correct already? And getting there without `class_hash` is weird. Maybe this verification should be done at an earlier stage?

docs/src/starknet/account-import-export.md
```diff
@@ -0,0 +1,146 @@
+# Importing And Exporting Accounts
```
cptartur, line 1: I think it should just be named `Importing Accounts`

docs/src/starknet/account-import-export.md
```diff
@@ -0,0 +1,146 @@
...
+
+You can export your account (more precisely its private key) from wallet (Argent, Braavos) and import it into the file holding the accounts info (`~/.starknet_accounts/starknet_open_zeppelin_accounts.json` by default).
+
+## Exporting an Account
```
cptartur, line 5: I'd name it something along the lines of `Getting Your Private Key` or similar.

docs/src/starknet/account-import-export.md
```diff
@@ -0,0 +1,146 @@
...
+$ sncast \
+    account import \
+	--url http://127.0.0.1:5050 \
+    --name account_123 \
+    --address 0x1 \
+    --private-key 0x2 \
+    --type oz
+```
+
+#### Passing Private Key in an Interactive, Obscured Mode
```
cptartur, line 84:
```suggestion
#### Passing Private Key in an Interactive Mode
```
Maybe just that is enough?

docs/src/starknet/account-import-export.md
```diff
@@ -0,0 +1,146 @@
...
+
+If you don't want to pass the private key in the command (because of visibility aspect), you can skip `--private-key` flag. You will be prompted to enter the private key in interactive mode.
```
cptartur, line 86:
```suggestion
If you don't want to pass the private key in the command (because of privacy aspect), you can skip `--private-key` flag. You will be prompted to enter the private key in interactive mode.
```
or maybe it should say safety or security. Though the interactive mode doesn't improve it that much.

docs/src/starknet/account-import-export.md
```diff
@@ -0,0 +1,146 @@
...
+	--url http://127.0.0.1:5050 \
+    --name account_123 \
+    --address 0x1 \
+    --private-key 0x2 \
+    --class-hash 0x3 \
+    --salt 0x4 \
+    --add-profile account_123
+    --accounts-file path/to/accounts.json
+    --type oz
+```
```
cptartur, line 113: Not sure if that's necessary here. There should already be section in appendix showing that. Here we should have relevant examples.

docs/src/starknet/account-import-export.md
```diff
@@ -0,0 +1,146 @@
...
+#### OpenZeppelin
+
+To import OpenZeppelin account, set the `--type` flag to `oz` or  `open_zeppelin`.
+
+```shell
+$ sncast \
+    account import \
+    ...
+    --type oz
+```
```
cptartur, line 146: This is super verbose, but at the same time doesn't show complete examples. I think if you want it to be helpful for some users it should show the whole commands so it can be copy-pasted.

docs/src/starknet/account.md
cptartur, line 1: This should not show any examples of importing accounts. Just creating. If you think that's necessary, we could have a relevant link somewhere on this page.

cptartur (changes requested):

crates/sncast/src/starknet_commands/account/import.rs
franciszekjob: Fine, refactored.

crates/sncast/src/starknet_commands/account/import.rs
franciszekjob: > I don't understand in which case would we not be able to fetch the class_hash for given address and at the same time be able to correctly validate the one that the user provided. I think they must match, no?

When an account contract is declared and not deployed, we will not be able to fetch its class hash so such situation can surely happen. That's why we later check whether it's declared.

```rust
 (None, Some(from_user)) => {
            check_class_hash_exists(provider, from_user).await?;
            Some(from_user)
        }
```

crates/sncast/src/starknet_commands/account/import.rs
franciszekjob: You're right, at this point in code we know that class hash is not `None`. I refactored class hash resolution logic, should be better now.

crates/sncast/src/starknet_commands/account/import.rs
franciszekjob: I don't really understand what's the issue here 😅 We compute the account address and check if it's the same (as the one from user) right after resolving class hash which is needed for computation. Could you please explain what you mean?

docs/src/starknet/account-import-export.md
franciszekjob: There's also section dedicated for exporting so I think the title is adequate 😄 .

docs/src/starknet/account-import-export.md
franciszekjob: `Exporting Your Private Key` should be fine 👍

docs/src/starknet/account-import-export.md
franciszekjob: `safety` should be adequate.

> Though the interactive mode doesn't improve it that much.

I know, but least it helps just a bit.

docs/src/starknet/account-import-export.md
franciszekjob: There isn't such section in appendix. Do you want me to add it there?

docs/src/starknet/account-import-export.md
franciszekjob: Actually, do we want to include separate examples which differ just in the `--type` flag? I think we can get rid of this.

docs/src/starknet/account.md
franciszekjob: Right, let's get rid of it here.

crates/sncast/src/starknet_commands/account/import.rs
franciszekjob: I refactored class hash resolution, it should be better now.

crates/sncast/src/starknet_commands/account/import.rs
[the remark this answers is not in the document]
franciszekjob: I've added function for parsing and relevant tests.

crates/sncast/src/starknet_commands/account/import.rs
```diff
@@ -0,0 +1,243 @@
...
+    })
+}
+
+fn get_private_key_from_file(file_path: &Utf8PathBuf) -> Result<Felt> {
+    let private_key_string = std::fs::read_to_string(file_path.clone())?;
+    Ok(private_key_string.parse()?)
+}
+
+fn parse_input_to_felt(input: &String) -> Result<Felt> {
+    let felt_re = Regex::new(r"^0x(0|[a-fA-F1-9]{1}[a-fA-F0-9]{0,62})$").unwrap();
```
cptartur, line 167: Let's link to spec with this

docs/src/starknet/account-import.md
```diff
@@ -0,0 +1,146 @@
...
+	--url http://127.0.0.1:5050 \
+    --name account_123 \
+    --address 0x1 \
+    --private-key 0x2 \
+    --class-hash 0x3 \
+    --salt 0x4 \
+    --add-profile account_123
+    --accounts-file path/to/accounts.json
+    --type oz
+```
```
cptartur, line 113: I'd remove this section or move it to appendix if it's necessary

docs/src/starknet/account-import.md
```diff
@@ -0,0 +1,146 @@
...
+#### OpenZeppelin
+
+To import OpenZeppelin account, set the `--type` flag to `oz` or  `open_zeppelin`.
+
+```shell
+$ sncast \
+    account import \
+    ...
+    --type oz
+```
```
cptartur, line 146: Let's either remove these or show complete examples that can be copy-pasted.

cptartur (changes requested):

docs/src/starknet/account-import.md
franciszekjob: Let's remove it, I don't consider it very handy tbh.

cptartur (approved):

PR #2569, opened 2024-10-07
Optimize tests compilation with `--exact` flag

### LINKED ISSUE #2530: Prototype filtering tests before compilation

snforge

### Description

Let's investigate if this can bring any major performance improvements and see what it would take to implement that properly.

cptartur:
Closes #2530 

## Introduced changes

- Optimize tests performance when using `--exact` flag
- Display `other filtered out` instead of number of filtered tests. This is should be brought back and addressed in https://github.com/foundry-rs/starknet-foundry/issues/2574

    Before
    ```rust
    Tests: 1 passed, 0 failed, 0 skipped, 0 ignored, 5 filtered out
    ```

    After
    ```rust
    Tests: 1 passed, 0 failed, 0 skipped, other ignored, other filtered out
    ```

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Short option for `--accounts-file` flag has been removed.
 - Short option for `--contract-address` is now `-d` instead of `-a`.
+- Test performance has been optimized. When using a filter, only the matching tests are compiled. The summary no longer displays the number of filtered-out tests when running all tests. When a filter is applied, it now shows `other filtered out` rather than the previous `<n> filtered out`.
```
cptartur, line 16: This should be in forge section, not cast.

Also I'd reword it so it says that this optimized compilation only happens when filtering by name. Not in case you pass a `path::to::test` or `path::to::module::`

crates/forge/src/run_tests/package.rs
```diff
@@ -131,9 +136,9 @@ pub async fn run_for_package(
     }: RunForPackageArgs,
     block_number_map: &mut BlockNumberMap,
 ) -> Result<Vec<TestTargetSummary>> {
+    clean_up_forge_test_filter();
```
cptartur, line 139: Why do we call this twice?

crates/forge/src/run_tests/package.rs
```diff
@@ -99,6 +100,10 @@ impl RunForPackageArgs {
     }
 }
 
+fn clean_up_forge_test_filter() {
```
cptartur, line 103: Maybe name it `unset_` or something similar?

crates/forge/src/run_tests/workspace.rs
```diff
@@ -43,6 +44,13 @@ pub async fn run_for_workspace(args: TestArgs) -> Result<ExitStatus> {
         .context("Failed to find any packages matching the specified filter")?;
 
     let filter = PackagesFilter::generate_for::<Metadata>(packages.iter());
+    let test_filter = args.test_filter.clone();
+
+    if let Some(last_filter) =
+        test_filter.and_then(|filter| filter.split("::").last().map(String::from))
+    {
+        setup_forge_test_filter(last_filter);
```
cptartur, line 52: Maybe just `set_`?

crates/forge/src/scarb.rs
```diff
@@ -66,11 +66,14 @@ fn build_contracts_with_scarb(filter: PackagesFilter, features: FeaturesSpec) ->
...
-    ScarbCommand::new_with_stdio()
+    let mut command = ScarbCommand::new_with_stdio();
+    command
         .arg("build")
         .arg("--test")
         .packages_filter(filter)
-        .features(features)
+        .features(features);
+
+    command
```
cptartur, line 76: This can be combined again

crates/forge/src/run_tests/package.rs
cptartur: Also, not sure if this should be called here. We execute this function multiple times, asynchronously. I think we should be unsetting the filter at the level we are setting it, before we return an `ExitStatus`.

docs/src/testing/running-tests.md
```diff
@@ -24,9 +26,11 @@ Collected 2 test(s) from package_name package
 Running 2 test(s) from src/
 [PASS] package_name::tests::calling
 [PASS] package_name::tests::calling_another
-Tests: 2 passed, 0 failed, 0 skipped, 0 ignored, 1 filtered out
+Tests: 2 passed, 0 failed, 0 skipped, 0 ignored, other filtered out
 ```
 
+When running with a filter, `other filtered out` is displayed in the summary.
```
cptartur, line 32: I think that instead of explaining cases in which it is displayed we should briefly explain optimized compilation logic here - maybe as a separate section. Just say something along the lines of "if you filter out tests using just a name, not paths (`::`), we will only compile these tests resulting in faster compilation`.

CHANGELOG.md
franciszekjob: I reworded it to 
```
Test performance has been optimized - using filter and `--exact` flag will result in compilation of only the selected test.```
because (as we discussed) it will be optimized only with `--exact` flag.

File: docs/src/testing/running-tests.md (hunk 6 above)
Comment (franciszekjob) [reply docs/src/testing/running-tests.md:32]: Actually I've reworded it and added a note, should draw user's attention more 👍 .

**Comment by franciszekjob** [comment]
discussed offline: We decided to optimize compilation only when `--exact` flag is passed.

File: CHANGELOG.md [hunk 7]
```diff
@@ -19,6 +19,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 #### Fixed
 - `account delete` command: It is no longer necessary to provide the `--url` argument each time. Either the `--url` or `--network` argument must be provided, but not both, as they are mutually exclusive.
 
+### Forge
+
+#### Changed
+
+- Test performance has been optimized - using filter and `--exact` flag will result in compilation of only the selected test.
```
cptartur, line 26:
```suggestion
- When using test name filter with `--exact` flag, forge will try to compile only the selected test.
```

It's not 100% true that it will only compile this tests. In case there are multiple matches by name, we'll compile all of them before filtering.

docs/src/testing/running-tests.md
```diff
@@ -32,12 +32,17 @@ Tests: 2 passed, 0 failed, 0 skipped, 0 ignored, 1 filtered out
 To run a specific test, you can pass a filter string along with an `--exact` flag.
 Note, you have to use a fully qualified test name, including a module name.
 
+> 📝 **Note**
+>
+> Running a specific test results in optimized compilation. Only the desired test will be compiled, unlike the case of running all tests where all of them are compiled.
```
cptartur, line 37:
```suggestion
> Running a specific test results in optimized compilation. `snforge` will try to compile only the desired test, unlike the case of running all tests where all of them are compiled.
```

cptartur (commented):
I think we should add a test case where there are two tests with exactly the same name, but at different paths and using `--exact` flag. To validate we only run one.

CHANGELOG.md
franciszekjob: Correct, it's because out false-positive approach, thanks for catching

cptartur (commented):
✅

PR #2570, opened 2024-10-07
Get features from all targets

### LINKED ISSUE #2567: Forge doesn't support collecting tests artifacts from some custom tests targets

snforge

### Description

See https://github.com/foundry-rs/starknet-foundry/blob/25044b09e491fdaf60b49bfdef11372ee353ac97/crates/scarb-api/src/lib.rs#L79

If someone defines a custom `[[test]]` target in scarb toml, that doesn't define `test-type` of `integration` or `unit` as we are expecting or doesn't have `name` in a format we are expecting, we will not collect any contracts to be used in tests.

### LINKED ISSUE #2568: Scarb features don't work with optimized compilation
When using scarb `--feature <some feature>`s with optimized (default for scarb versions that support it) compilation flow, any contract defined in `src/` under `#[cfg(feature: <some feature>)` attribute will not be available for declaring in tests.

cptartur:
Closes #2568 
Closes #2567

## Introduced changes

- Fixed features support for optimized compilation
- Changed the starknet artifacts loading logic so they work with custom test targets.
- Artifacts are now loaded from all test targets
- Artifacts are now compiled to casm concurrently
- Refactored the scarb-api crate

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/scarb-api/src/lib.rs
```diff
@@ -75,64 +78,121 @@ fn artifacts_for_package(path: &Utf8Path) -> Result<StarknetArtifacts> {
...
+#[derive(PartialEq, Debug)]
+struct ContractArtifactData {
+    path: Utf8PathBuf,
+    test_type: Option<String>,
+}
+
+fn build_target_path(target_dir: &Utf8Path, file_name: &str) -> Option<Utf8PathBuf> {
+    let path = target_dir.join(file_name);
+
+    if path.exists() {
```
piotmag769, line 90: I think the path should always exist in the places you use it, otherwise there is an error in our logic (we forget to call scarb/ignore its failure) or scarb doesn't work, in which case we should error anyways I think

crates/scarb-api/src/lib.rs
```diff
@@ -75,64 +78,121 @@ fn artifacts_for_package(path: &Utf8Path) -> Result<StarknetArtifacts> {
...
-    );
-
-    let map = match maybe_contracts_path {
-        Some(contracts_path) => load_contracts_artifacts_and_source_sierra_paths(&contracts_path)?,
-        None => HashMap::default(),
+    let contracts_paths = if use_test_target_contracts {
+        let test_targets = test_targets_by_name(package);
+        get_starknet_artifacts_paths_from_test_targets(target_dir, &test_targets)
+    } else {
+        let target_name = target_name_for_package(metadata, &package.id)?;
```
piotmag769, line 150: This function is not needed anymore together with `compilation_unit_for_package`. Here we should instead look for `starknet-contract` target name only and return `Option`. We can also drop passing whole metadata to this function, just do
```rs
let starknet_target_name = package.targets.iter().find(|target| target.kind == "starknet-contract).map(|target| target.name);
```

 If not found, we should return no artifacts and emit a warning that `starknet-contract` target was not defined but `--no-optimization` flag was used and no contracts will be available for declare

crates/scarb-api/src/lib.rs
```diff
@@ -75,64 +78,121 @@ fn artifacts_for_package(path: &Utf8Path) -> Result<StarknetArtifacts> {
     Ok(starknet_artifacts)
 }
 
+#[derive(PartialEq, Debug)]
+struct ContractArtifactData {
+    path: Utf8PathBuf,
+    test_type: Option<String>,
+}
```
piotmag769, line 85: This name is bad imo, there are already a lot of structs in this file that are named similarly, but have different semantics. Maybe there is another way to express in the typesystem that one of the paths should be treated as "default/base" (should be documented what it actually means here) one?

```rs
struct StarknetArtifactsFiles {
    default_artifacts_file: Utf8PathBuf,
    other_files: Vec<Utf8PathBuf>,
}
```

then it could have `load_contracts_artifacts` as a method. WDYT?

crates/scarb-api/src/lib.rs
```diff
@@ -75,64 +78,121 @@ fn artifacts_for_package(path: &Utf8Path) -> Result<StarknetArtifacts> {
...
+    let starknet_artifacts_file_name =
+        |name: &str, metadata: &TargetMetadata| -> Option<ContractArtifactData> {
+            let path = format!("{name}.test.starknet_artifacts.json");
+            let path = build_target_path(target_dir, &path);
+            let test_type = metadata
+                .params
+                // TODO get rid of hardcoding
+                .get("test-type")
+                .and_then(|value| value.as_str())
+                .unwrap();
```
piotmag769, line 110: I don't think this unwrap is safe, is it? Is it guaranteed that the `test-type` will be known for manually defined test targets?

piotmag769 (changes requested):
Just reviewed the logic

[nit] `name_for_package` is used only in a unittest, we can remove it
[nit] please move public stuff up in this file and private stuff below it, will be much easier to read

Ofc should be in a separate PR

crates/forge/tests/data/targets/custom_target_custom_names/metadata.json
maciektr, line 1: remove (I know you know, just to not miss it :) )

crates/forge/tests/data/targets/only_integration/Scarb.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "targets"
```
maciektr, line 2: Won't you get warnings from Scarb when using snforge_std as git dep if you use the same name for all this packages?

crates/scarb-api/src/lib.rs
```diff
@@ -75,64 +78,121 @@ fn artifacts_for_package(path: &Utf8Path) -> Result<StarknetArtifacts> {
...
+    contracts_paths: &[ContractArtifactData],
+) -> Result<HashMap<String, (StarknetContractArtifacts, Utf8PathBuf)>> {
+    if contracts_paths.is_empty() {
+        return Ok(HashMap::new());
+    }
+
+    // TODO use const
+    let base_artifacts = contracts_paths
+        .iter()
+        .find(|paths| paths.test_type == Some("integration".to_string()))
```
maciektr, line 173: Shouldn't this prioritize `None` i.e. contracts from `starknet-contract`? 🤔

crates/scarb-api/src/lib.rs
```diff
@@ -75,64 +78,121 @@ fn artifacts_for_package(path: &Utf8Path) -> Result<StarknetArtifacts> {
...
+
+    let mut base_artifacts =
+        load_contracts_artifacts_and_source_sierra_paths(&base_artifacts.path)?;
+
+    for artifact in other_artifacts {
+        let artifact = load_contracts_artifacts_and_source_sierra_paths(&artifact.path)?;
+        for (key, value) in artifact {
+            base_artifacts.entry(key).or_insert(value);
+        }
+    }
```
maciektr, line 193: Do I understand correctly, this is only meant to be a fallback? 
If so, isn't it extremely wasteful, that it compiles all contract files with USC before checking if you even need them?

maciektr: Btw. I still think it should be possible to compile all contract files in parallel with multiple usc processes.

maciektr (changes requested):

crates/scarb-api/src/lib.rs
cptartur: I think I'll just default to `None` since I'd have to hardcode a string here.

crates/scarb-api/src/lib.rs
```diff
@@ -75,64 +78,121 @@ fn artifacts_for_package(path: &Utf8Path) -> Result<StarknetArtifacts> {
     Ok(starknet_artifacts)
 }
 
+#[derive(PartialEq, Debug)]
+struct ContractArtifactData {
+    path: Utf8PathBuf,
+    test_type: Option<String>,
+}
+
+fn build_target_path(target_dir: &Utf8Path, file_name: &str) -> Option<Utf8PathBuf> {
```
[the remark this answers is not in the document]
cptartur: Removed

crates/scarb-api/src/lib.rs
```diff
@@ -75,64 +79,139 @@ fn artifacts_for_package(path: &Utf8Path) -> Result<StarknetArtifacts> {
...
+            })
+        };
+
+    test_targets
+        .iter()
+        .filter_map(|(target_name, metadata)| starknet_artifacts_file_name(target_name, metadata))
+        .collect()
+}
+
+/// Try getting the path to `starknet_artifacts.json` file that is generated by `scarb build` command
```
piotmag769, line 123:
```suggestion
/// Try getting the path to `starknet_artifacts.json` related to `starknet-contract` target. This file that is generated by `scarb build` command.
```

crates/scarb-api/src/lib.rs
```diff
@@ -75,64 +79,139 @@ fn artifacts_for_package(path: &Utf8Path) -> Result<StarknetArtifacts> {
...
+                    .collect::<Vec<_>>()
+            })
+            .unwrap_or_default()
     };
 
-    if use_test_target_contracts {
-        ["integration", "unit"]
-            .iter()
-            .find_map(|test_type| check_path(&starknet_artifacts_file_name(Some(test_type))))
+    if contracts_paths.is_empty() {
```
piotmag769, line 172: This check should be in a callee, rn a caller needs to remember to check this

crates/scarb-api/src/lib.rs
cptartur: Logic was changed. But there's no case where we have contracts both from "starknet-contract" and from "integration". It's either first if we build with `scarb build` or the second when using `scarb build --test`.

crates/scarb-api/src/lib.rs
cptartur: We can use `par_iter` to compile these in parallel, it should be safe.

crates/scarb-api/src/lib.rs
cptartur: Refactored

piotmag769 (approved):
Logic looks fine

crates/scarb-api/src/lib.rs
```diff
@@ -75,64 +80,150 @@ fn artifacts_for_package(path: &Utf8Path) -> Result<StarknetArtifacts> {
...
+        let compiled_artifacts = self
+            .other_files
+            .par_iter()
+            .map(load_contracts_artifacts_and_source_sierra_paths)
+            .collect::<Result<Vec<_>>>()?;
+
+        for artifact in compiled_artifacts {
+            for (key, value) in artifact {
+                base_artifacts.entry(key).or_insert(value);
+            }
```
maciektr, line 104: It still compiles those files, just to most probably discard them anyway? 👀 
Can't we check if we need specific contract artifact without compiling it first?

crates/scarb-api/src/lib.rs
```diff
@@ -75,64 +80,150 @@ fn artifacts_for_package(path: &Utf8Path) -> Result<StarknetArtifacts> {
...
+///
+/// If artifacts with `test_type` of `INTEGRATION_TEST_TYPE` are present, we use them base path
+/// and extend with paths to other artifacts.
+/// If `INTEGRATION_TEST_TYPE` is not present, we take first artifacts found.
+fn get_starknet_artifacts_paths_from_test_targets(
+    target_dir: &Utf8Path,
+    test_targets: &HashMap<String, &TargetMetadata>,
+) -> Option<StarknetArtifactsFiles> {
+    #[derive(PartialEq, Debug, Clone)]
+    struct ContractArtifactData {
```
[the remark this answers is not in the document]
maciektr: ^ What's wrong with naming variables? :P

crates/scarb-api/src/lib.rs
```diff
@@ -75,64 +80,143 @@ fn artifacts_for_package(path: &Utf8Path) -> Result<StarknetArtifacts> {
...
+    other_files: Vec<Utf8PathBuf>,
+}
+
+impl StarknetArtifactsFiles {
+    fn load_contracts_artifacts(
+        self,
+    ) -> Result<HashMap<String, (StarknetContractArtifacts, Utf8PathBuf)>> {
+        let mut base_artifacts = load_contracts_artifacts_and_source_sierra_paths(&self.base_file)?;
+
+        let compiled_artifacts = self
```
maciektr, line 95: Btw. is `scarb-api` the right place to hide USC calls? 🤔

crates/forge/tests/data/targets/only_lib_integration/tests/tests.cairo
```diff
@@ -0,0 +1,28 @@
...
+    let (contract_address, _) = contract.deploy(constructor_calldata).unwrap();
+    let dispatcher = IHelloStarknetDispatcher { contract_address };
+
+    let balance = dispatcher.get_balance();
+    assert(balance == 0, 'balance == 0');
+
+    dispatcher.increase_balance(100);
+
+    let balance = dispatcher.get_balance();
+    assert(balance == 100, 'balance == 100');
```
kkawula, line 27: [nit] shouldn't error message indicate what was wrong?
```
assert(balance == 100, 'balance != 100');
```

crates/scarb-api/src/artifacts/deserialized.rs
```diff
@@ -0,0 +1,39 @@
...
+pub(super) struct StarknetContractArtifactPaths {
+    pub sierra: Utf8PathBuf,
+}
+
+/// Get deserialized contents of `starknet_artifacts.json` file generated by Scarb
+///
+/// # Arguments
+///
+/// * `path` - A path to `starknet_artifacts.json` file.
+pub(super) fn artifacts_for_package(path: &Utf8Path) -> Result<StarknetArtifacts> {
```
piotmag769, line 32: It is only used in `represensation.rs`, no? Why not put it there?

crates/scarb-api/src/artifacts/deserialized.rs
```diff
@@ -0,0 +1,39 @@
...
+    let starknet_artifacts =
+        fs::read_to_string(path).with_context(|| format!("Failed to read {path:?} contents"))?;
+    let starknet_artifacts: StarknetArtifacts =
+        serde_json::from_str(starknet_artifacts.as_str())
+            .with_context(|| format!("Failed to parse {path:?} contents. Make sure you have enabled sierra code generation in Scarb.toml"))?;
```
piotmag769, line 37: Rn it has nothing to do with sierra code gen anymore I think?

crates/scarb-api/src/artifacts/artifacts.rs
piotmag769, line 1: This code should be in `crates/scarb-api/src/artifacts.rs` and rest of the modules should be private so only the `artifacts` module can use them and their content. Then you can remove `pub(super)` etc. from `deserialized` and `representation`

crates/scarb-api/src/artifacts/deserialized.rs
piotmag769, line 1: This looks like represenatation too, no? Those are kinda like internal structs for `StarknetArtifactsRepresentation`, so best to declare them where they are used

crates/scarb-api/src/artifacts/representation.rs
```diff
@@ -0,0 +1,78 @@
+use crate::artifacts::deserialized::{artifacts_for_package, StarknetArtifacts};
+use anyhow::anyhow;
+use camino::{Utf8Path, Utf8PathBuf};
+
+pub(super) struct StarknetArtifactsRepresentation {
+    path: Utf8PathBuf,
```
piotmag769, line 6:
```suggestion
    base_path: Utf8PathBuf,
```

crates/scarb-api/src/artifacts/representation.rs
```diff
@@ -0,0 +1,78 @@
...
+    artifacts: StarknetArtifacts,
+}
+
+impl StarknetArtifactsRepresentation {
+    pub(super) fn try_from_path(path: &Utf8Path) -> anyhow::Result<Self> {
```
piotmag769, line 11:
```suggestion
    pub(super) fn try_from_path(path_to_artifacts: &Utf8Path) -> anyhow::Result<Self> {
```

crates/scarb-api/src/artifacts/representation.rs
```diff
@@ -0,0 +1,78 @@
+use crate::artifacts::deserialized::{artifacts_for_package, StarknetArtifacts};
+use anyhow::anyhow;
+use camino::{Utf8Path, Utf8PathBuf};
+
+pub(super) struct StarknetArtifactsRepresentation {
```
piotmag769, line 5: This struct could be a function, you always use `try_new` and `artifacts` in a sequence anyways. This struct is dropped in the same function it is created in

crates/scarb-api/src/artifacts/artifacts.rs
```diff
@@ -0,0 +1,89 @@
...
+    /// Compiled sierra code
+    pub sierra: String,
+    /// Compiled casm code
+    pub casm: String,
+}
+
+#[derive(PartialEq, Debug)]
+pub(crate) struct StarknetArtifactsFiles {
+    base_file: Utf8PathBuf,
+    other_files: Vec<Utf8PathBuf>,
```
piotmag769, line 23:
```suggestion
    base: Utf8PathBuf,
    others: Vec<Utf8PathBuf>,
```

kkawula (commented):
Could you write a short summary in the description of PR

crates/scarb-api/src/lib.rs
cptartur: #2602

crates/scarb-api/src/lib.rs
```diff
@@ -213,18 +160,40 @@ pub fn package_matches_version_requirement(
...
     use assert_fs::fixture::{FileWriteStr, PathChild, PathCopy};
-    use assert_fs::prelude::FileTouch;
     use assert_fs::TempDir;
     use camino::Utf8PathBuf;
     use indoc::{formatdoc, indoc};
+    use std::fs;
     use std::str::FromStr;
 
-    fn setup_package(package_name: &str) -> TempDir {
+    pub(crate) fn setup_package(package_name: &str) -> TempDir {
```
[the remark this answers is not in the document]
cptartur: I don't think it hurts us to limit visibility of this

crates/scarb-api/src/artifacts/deserialized.rs
cptartur: This deserializes into structs defined above. I though it makes sense to have it along the structs it uses.

crates/scarb-api/src/artifacts/deserialized.rs
cptartur: It's tricky. Because this can be both sierra generation for older flow and just an error with test targets in new flow. I'll try to update this.

crates/scarb-api/src/artifacts/deserialized.rs
cptartur: I think having this split into meaningful files is more readable than one large one. Agreed that these are related however `deserialized.rs` is strictly code for deserialization using serde so I think it makes sense.

crates/scarb-api/src/artifacts/representation.rs
cptartur: Doesn't hurt us and I prefer to use structs to represent intermediate data than passing random tuples.

piotmag769 (approved):
Approve, I guess the last commits were just refactors

cptartur:
The docs on how this work needs to be updated in #2627

CHANGELOG.md
```diff
@@ -40,12 +48,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - When using test name filter with `--exact` flag, forge will try to compile only the selected test.
 
-### Forge
-
-#### Changed
-
-- Cairo Edition in `snforge init` template set to `2024_07`
-
```
piotmag769, line 48: ?

PR #2590, opened 2024-10-17
Add doc strings to snforge init template

maciektr:
I am open to any changes.

### DIFF
--- a/starknet_forge_template/src/lib.cairo
+++ b/starknet_forge_template/src/lib.cairo
@@ -1,14 +1,19 @@
+/// Interface representing `HelloContract`.
+/// This interface allows modification and retrieval of the contract balance.
 #[starknet::interface]
 pub trait IHelloStarknet<TContractState> {
+    /// Increase contract balance.
     fn increase_balance(ref self: TContractState, amount: felt252);
+    /// Retrieve contract balance.
     fn get_balance(self: @TContractState) -> felt252;
 }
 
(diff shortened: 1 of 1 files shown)


cptartur (commented):
I think maybe it'd make sense to add a comment to the template saying something like "use this command to generate documentation"

maciektr:
@cptartur In the code? Seems a bit excessive, we don't really have comments instructing to run `scarb build` or `snforge test` either.

cptartur:
> @cptartur In the code? Seems a bit excessive, we don't really have comments instructing to run `scarb build` or `snforge test` either.

Right, but maybe a comment like that would improve the discoverability of scarb doc? Alternatively this could be mentioned somewhere else, maybe we should generate a `README.md` file with some commands?

maciektr:
@cptartur 
> maybe a comment like that would improve the discoverability of scarb doc

I am a bit afraid this is a rabbit hole 😓 
We have a lot of cool futures, why not promote them all?

cptartur:
> @cptartur
> 
> > maybe a comment like that would improve the discoverability of scarb doc
> 
> I am a bit afraid this is a rabbit hole 😓 We have a lot of cool futures, why not promote them all?

I guess we can think about promoting them better if/when we implement forge cheatsheet. Good to merge for now.

cptartur (approved):

PR #2592, opened 2024-10-18
Pass args to coverage and profiler

ksew1:
Closes https://github.com/software-mansion/cairo-coverage/issues/80

## Introduced changes

We added:
```rust
  #[clap(last = true)]
    additional_args: Vec<OsString>,
```

So all arguments passed after `--`, will be collected to string and passed `cairo-coverage` and `cairo-profiler`
- You can't use `--coverage` and `--build-profile` together 

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Forge
+
+#### Changed
+
+- You can now pass arguments to `cairo-profiler` or `cairo-coverage`,
+  e.g. `snforge test --coverage --include tests-functions`. If flags name duplicates a flag in `snforge test`, you can
+  use `--` to separate them, e.g. `snforge test --coverage -- --help`
+- You can't use now `--coverage` and `--build-profile` flags at the same time. If you want to use both, you need to run
+  `snforge test` twice with different flags.
+
 ## [0.32.0] - 2024-10-16
 
 ### Cast
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,9 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 #### Changed
 
-- You can now pass arguments to `cairo-profiler` or `cairo-coverage`,
-  e.g. `snforge test --coverage --include tests-functions`. If flags name duplicates a flag in `snforge test`, you can
-  use `--` to separate them, e.g. `snforge test --coverage -- --help`
+- You can now pass arguments to `cairo-profiler` and `cairo-coverage`. Everything after `--` will be passed to it. E.g.
+  `snforge test --build-profile -- --show-inlined-functions`
 - You can't use now `--coverage` and `--build-profile` flags at the same time. If you want to use both, you need to run
   `snforge test` twice with different flags.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 #### Changed
 
-- You can now pass arguments to `cairo-profiler` and `cairo-coverage`. Everything after `--` will be passed to it. E.g.
+- You can now pass arguments to `cairo-profiler` and `cairo-coverage`. Everything after `--` will be passed to underlying binary. E.g.
   `snforge test --build-profile -- --show-inlined-functions`
 - You can't use now `--coverage` and `--build-profile` flags at the same time. If you want to use both, you need to run
   `snforge test` twice with different flags.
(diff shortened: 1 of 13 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 ## [Unreleased]
 
+### Forge
+
+#### Changed
+
+- You can now pass arguments to `cairo-profiler` or `cairo-coverage`,
+  e.g. `snforge test --coverage --include tests-functions`. If flags name duplicates a flag in `snforge test`, you can
+  use `--` to separate them, e.g. `snforge test --coverage -- --help`
+- You can't use now `--coverage` and `--build-profile` flags at the same time. If you want to use both, you need to run
```
ksew1, line 17: We can change this behavior to: 
> you can't use the `--coverage` and `--build-profile` flags if you are passing additional arguments

but it might  confuse users. won't be having breaking change though

crates/forge/src/lib.rs
```diff
@@ -173,7 +179,16 @@ pub fn main_execution() -> Result<ExitStatus> {
...
-        ForgeSubcommand::Test { args } => {
+        ForgeSubcommand::Test { mut args } => {
+            // clap can capture the first flag in test filter,
+            // let's move it to coverage_or_profiler_args
+            if let Some(test_filter) = &args.test_filter {
+                if test_filter.starts_with('-') {
+                    args.coverage_or_profiler_args.insert(0, test_filter.into());
+                    args.test_filter = None;
+                }
+            }
```
kkawula, line 190: I suppose that such heuristic can generate problems in the future, and introducing such condition disturbs the readability of the code. Maybe we should just add coverage subcommand to catch arguments in simpler and cleaner way? What do you think?

kkawula (changes requested):

crates/forge/src/lib.rs
kkawula: > I think we could at least check if --coverage or --build-profile were actually passed. Because I think this heuristic should be only enabled in this case, otherwise, we shouldn't be doing anything with the filter.

yes, it's the correct one, but it is going to be the same thing as if there was just a subcommand.

crates/forge/tests/e2e/coverage.rs
```diff
@@ -19,6 +19,21 @@ fn test_coverage_project() {
...
+
+    test_runner(&temp)
+        .arg("--coverage")
+        .arg("--output-path")
+        .arg("./my_file.lcov")
+        .assert()
+        .success();
+
+    assert!(temp.join("my_file.lcov").is_file());
+}
```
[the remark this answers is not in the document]
ksew1: Zrzut ekranu 2024-10-22 o 12 47 50

It is executed

crates/forge-runner/src/coverage_api.rs
```diff
@@ -47,10 +44,19 @@ pub fn run_coverage(saved_trace_data_paths: &[PathBuf]) -> Result<()> {
...
-        .arg(&path_to_save_coverage)
+    let mut command = Command::new(coverage);
+
+    if coverage_args.iter().all(|arg| arg != "--output-path") {
+        let dir_to_save_coverage = PathBuf::from(COVERAGE_DIR);
+        fs::create_dir_all(&dir_to_save_coverage).context("Failed to create a coverage dir")?;
+        let path_to_save_coverage = dir_to_save_coverage.join(OUTPUT_FILE_NAME);
+
+        command.arg("--output-path").arg(&path_to_save_coverage);
+    }
```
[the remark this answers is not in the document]
ksew1: It is responsibility of the user to create that directory, that is how it is handled in coverage.

crates/forge/src/lib.rs
ksew1: > I suppose that such heuristic can generate problems in the future, and introducing such condition disturbs the readability of the code. Maybe we should just add coverage subcommand to catch arguments in simpler and cleaner way? What do you think?

I guess we were discussing this earlier i don't know why we bring it here now and block it with the request changes 😅

CHANGELOG.md
```diff
@@ -7,6 +7,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Forge
+
+#### Changed
+
+- You can now pass arguments to `cairo-profiler` or `cairo-coverage`,
+  e.g. `snforge test --coverage --include tests-functions`. If flags name duplicates a flag in `snforge test`, you can
```
kkawula, line 15: This sentence doesn't really sound clear to me. Is `snforge test --coverage --include tests-functions` proper or not?

crates/forge/src/lib.rs
```diff
@@ -145,6 +147,10 @@ pub struct TestArgs {
     /// Build contracts separately in the scarb starknet contract target
     #[arg(long)]
     no_optimization: bool,
+
+    /// Additional arguments for cairo-coverage or cairo-profiler
+    #[clap(last = true)]
+    additional_args: Vec<OsString>,
```
kkawula, line 153: Could you please describe shortly in pr summary why `allow_hyphen_values` and `last` were introduced?

kkawula (approved):
Taylor LGTM

cptartur (approved):

PR #2606, opened 2024-10-24
Fix error while calling nonexistent entry point selector

### LINKED ISSUE #1812: Different errors when calling unknown entry point selector error from contract and from test

snforge

### Description

Calling inexistent entry point:
- from test - fixed in #1741 
- from another contract

When calling from test this is the entered arm:
https://github.com/foundry-rs/starknet-foundry/blob/b02aa0c7ef1aadc48a2d51467fdf55045387c695/crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/rpc.rs#L138

However when calling from another contract, this is the entered arm:
https://github.com/foundry-rs/starknet-foundry/blob/b02aa0c7ef1aadc48a2d51467fdf55045387c695/crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/rpc.rs#L131
It enters [`try_extract_panic_data`](https://github.com/foundry-rs/starknet-foundry/blob/e33d4db5822b805408ab9546051d84657d1407b2/crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/panic_data.rs#L6C8-L6C30)

One solution would be to match the entry point error message via a regex in there.

---
The use case for calling an inexistent entry point is testing entry points after calling `replace_class`.

Example error message:
```
Failure data:
    Got an exception while executing a hint: Hint Error: Error in the called contract (0x0480ddec869b0aa42b46971088ba8701df26fcabb2d3e9d324fd6fd3d013fef7):
Error at pc=0:927:
Got an exception while executing a hint.
Cairo traceback (most recent call last):
Unknown location (pc=0:358)
Unknown location (pc=0:739)

Error in the called contract (0x01f78e70f8b1e91ac2c327a35e451c1985ba46c4c8f4f19488e5a5e135e5be74):
Error at pc=0:1197:
Got an exception while executing a hint: Hint Error: Entry point EntryPointSelector(StarkFelt("0x0017c00f03de8b5bd58d2016b59d251c13056b989171c5852949903bc043bc27")) not found in contract.
Cairo traceback (most recent call last):
Unknown location (pc=0:475)
Unknown location (pc=0:926)
```

kkawula:
Closes #1812

The `Err` message chosen to be thrown by `safe-dispatcher` is intended to replicate `cairo-test` behavior. These felts are taken explicitly from [`CairoHintProcessor`](https://github.com/starkware-libs/cairo/blob/2ad7718591a8d2896fec2b435c509ee5a3da9fad/crates/cairo-lang-runner/src/casm_run/mod.rs#L1055-L1057)

## Introduced changes

- Add proper regex to catch falling error message

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


kkawula:
The error message was created specifically to reflect and be in consequence with the cairo-test. Now both tests give the same output: [tests](https://gist.github.com/andrew-fleming/ab52b92897fcd3c89afc689ca6eb9217)

CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Forge
+
+#### Fixed
+
+- When a contract call function via safe-dispatcher, calling non-existent entrypoint now returns `Err` instead of panicking.
```
kkawula, line 14: Not quite sure about this sentence.

franciszekjob, line 14:
```suggestion
- Calling contract functions via safe-dispatcher now returns an `Err` when attempting to invoke a non-existent entry point, instead of causing a panic.
```

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/panic_data.rs
```diff
@@ -10,6 +10,10 @@ pub fn try_extract_panic_data(err: &str) -> Option<Vec<Felt252>> {
     let re_string = Regex::new(r#"(?s)Execution failed\. Failure reason: "(.*?)"\."#)
         .expect("Could not create string panic_data matching regex");
 
+    let re_entry_point =
+        Regex::new(r"Entry point EntryPointSelector\((0x[0-9a-fA-F]+)\) not found in contract\.")
+            .unwrap();
+
```
franciszekjob, line 16: Could you please tell where this regex is taken from? From what I see in [spec](https://github.com/starkware-libs/starknet-specs/blob/6d88b7399f56260ece3821c71f9ce53ec55f830b/api/starknet_api_openrpc.json#L2455), entry point selector points to felt which has different regex 🤔 .

Also, if it's from spec - we should include a link in the comment as reference.

kkawula: Upadated

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/panic_data.rs
```diff
@@ -31,6 +36,14 @@ pub fn try_extract_panic_data(err: &str) -> Option<Vec<Felt252>> {
...
     }
 
+    if let Some(_captures) = re_entry_point.captures(err) {
+        let panic_data_felts = vec![
+            Felt252::from_bytes_be_slice("ENTRYPOINT_NOT_FOUND".as_bytes()),
+            Felt252::from_bytes_be_slice("ENTRYPOINT_FAILED".as_bytes()),
+        ];
+        return Some(panic_data_felts);
+    }
+
```
franciszekjob, line 46: Could you also explain why we return this specific panic data felts if entry point regex matches? I think we can add brief info about it in the PR description, it might be not clear at first glance.

cptartur: Agreed, maybe we can link some relevant lines from somewhere else?

cptartur (changes requested):
Can we add a test that verifies this behavior was fixed?

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/panic_data.rs
```diff
@@ -10,6 +10,11 @@ pub fn try_extract_panic_data(err: &str) -> Option<Vec<Felt252>> {
     let re_string = Regex::new(r#"(?s)Execution failed\. Failure reason: "(.*?)"\."#)
         .expect("Could not create string panic_data matching regex");
 
+    let re_entry_point = Regex::new(
+        r"Entry point EntryPointSelector\((0x[a-fA-F0-9]{0,64})\) not found in contract\.",
+    )
+    .expect("Could not create entry point panic_data matching regex");
+
```
kkawula, line 17: Unfortunately, felt returned from cairoVM doesn't fulfill felt regex definition from spec and as a result doesn't capture the selector returned in error.
I don't think it will be a problem to left it in `(0x[a-fA-F0-9]{0,64})` from.

franciszekjob: It looks like cairoVM returns felts in padded form (that's why it's 64 chars after 0x, instead of 63 which is in spec). It's fine, let's leave it as it is 👍  but I think we should add a brief comment explaining why this regex is different that the spec's one.

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/panic_data.rs
franciszekjob: https://github.com/starkware-libs/cairo/blob/2ad7718591a8d2896fec2b435c509ee5a3da9fad/crates/cairo-lang-runner/src/casm_run/mod.rs#L1055-L1057 can be linked in the comment.

franciszekjob (approved):
There are two places in which we should add comments with reference links, otherwise it's good to go 👌 .

cptartur (approved):

PR #2610, opened 2024-10-24
feat(sncast): generate default name when using account import

### LINKED ISSUE #2581: Generate default `name` when using `account import`

Cast

### Feature Request

When using `account import` we should be able to use it without passing `name` which is currently required. Cast should use default (available) or automatically generated name.

**Current usage**

```
sncast account import \
    --name abc111 \
    --address 0x1bbc22958cbc7a37bf8d787fc693ec296c0f46834fe6027054e3fb47d3edf75 \
    --private-key 0x1 \
    --type oz  \
    --url http://127.0.0.1:5050/rpc
```

**Usage after implementation**
```
sncast account import \
    --address 0x1bbc22958cbc7a37bf8d787fc693ec296c0f46834fe6027054e3fb47d3edf75 \
    --private-key 0x1 \
    --type oz  \
    --url http://127.0.0.1:5050/rpc
```

Both command such succeed with such output:
```
command: account import
```

khayss:
Closes #2581 

## Introduced changes

All changes made are focused on `sncast`.
In `crates/sncast/src/starknet_commands/account/import.rs`, I added a utility function to generate a random account name which is used to supply the account construct with a name where the user does not specify any to the  Also updated the import function in this file by changing the function parameter to accept an `Option<String>` type for `account` parameter. This way, users can import accounts without requiring `name` to be specified. Also, updated the `Import` struct `name` key to have a type of `Option<String>`.

In `crates/sncast/src/main.rs`, I updated the argument feed into the `import` function `account` parameter.

I updated the necessary documentation and logs.

-

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/sncast/src/starknet_commands/account/import.rs
```diff
@@ -196,6 +204,15 @@ fn get_private_key_from_input() -> Result<Felt> {
...
 
+fn generate_default_account_name() -> String {
+    let start = SystemTime::now();
+    let since_epoch = start
+        .duration_since(UNIX_EPOCH)
+        .expect("Time went backwards");
+    let timestamp = since_epoch.as_millis();
+    format!("account-{}", timestamp)
+}
+
```
franciszekjob, line 215: Let's refactor this function to create account with sequential names (`account-1`, `account-2`, ...). Name with timestamp (e.g. `account-1729865041172`) is too complicated.
Our logic should find next available numer (and handle gaps, e.g. having `account-1` and `account-3` in file, imported account should receive name `account-2`).

You can see how accounts are read from the file [here](https://github.com/foundry-rs/starknet-foundry/blob/e5918db054d7264fd08e79dee1c33d6391009eda/crates/sncast/src/starknet_commands/account/mod.rs#L102).

franciszekjob (changes requested):
1. Please fix linting.
2. Add test for importing account without `--name` flag.
3. Mention the change in `CHANGELOG.md`.

crates/sncast/src/starknet_commands/account/import.rs
khayss: Alright. This was my initial plan. I'll fall back to it.

khayss:
Hello @franciszekjob 
In this draft PR, I have implemented a way to generate accounts in a more sequential and friendly manner as requested.
However, I feel there are some challenges and possibly bugs.

I implemented the feature by tracking the number of default accounts and incrementing or decrementing as new accounts are added or removed. Also, I keep track of deleted ids in an array this way, we can reuse ids and cover gaps.

This implementation meant the structure of how the account is stored had to be restructured. This affect features asides from the `import` as they also read and write to the same account.

I still believe, the best approach, is defining a custom type that all features use, this way, if any change is made to one area, it'll cause an error that can be caught at compile time.

Also, due to the use of `serde_json::Value` type which is quite loose, we can run into some bugs and break other features by updating one.

For example, the following code can cause unexpected bugs.
https://github.com/foundry-rs/starknet-foundry/blob/f7cfe3e7b6792acb824ba344d4c63f9853a58fc6/crates/sncast/src/starknet_commands/account/deploy.rs#L340

If for instance there's a network with the name `deployment`, the line of code below will assign different data to the same key as above, which means one will overwrite the other.
https://github.com/foundry-rs/starknet-foundry/blob/f7cfe3e7b6792acb824ba344d4c63f9853a58fc6/crates/sncast/src/starknet_commands/account/mod.rs#L114C4-L114C50

This shows the need to use a common source of truth so we're certain, there's no conflict.

crates/sncast/src/starknet_commands/account/import.rs
```diff
@@ -80,6 +80,12 @@ pub async fn import(
     };
     let private_key = &SigningKey::from_secret_scalar(*private_key);
 
+    // let account = if let Some(name) = account {
+    //     name
+    // } else {
+    //     generate_default_account_name()
+    // };
+
```
franciszekjob, line 88: We shouldn't keep comments unless they are related with todo or there is a good reason to keep it in the codebase.

crates/sncast/src/starknet_commands/account/mod.rs
```diff
@@ -21,6 +21,21 @@ pub mod deploy;
...
+//              account details
+//              },
+//          }
+//      },
+//  settings: {
+//    default_account_count: u32,
+//    available_default_account_ids: Vec<u32>
+//      }
+//},
+
```
franciszekjob, line 38: Same as above.

franciszekjob (changes requested):
First of all, please address the changes I requested in previous review (formatting, tests etc.)

The only thing this feature should include, is to generate a default name if user doesn't pass one. This should be achieved by one function, there's no need to introduce new account structures etc.

Also, this feature shouldn't affect account deletion and any other functionalities.
Please keep it simple :).

crates/sncast/src/starknet_commands/account/import.rs
```diff
@@ -252,4 +305,18 @@ mod tests {
...
+        let (account, account_id) = super::get_import_name(None, &accounts_file);
+        let (account_2, account_id_2) = super::get_import_name(None, &accounts_file);
+
+        assert_eq!(account, "account-1");
+        assert_eq!(account_id, Some(1));
+
+        assert_eq!(account_2, "account-2");
+        assert_eq!(account_id_2, Some(2));
+    }
```
khayss, line 321: For some reason, this test and all other tests in the module are not captured.

khayss:
I have made the changes you asked for.

I believe the linting is correct.
Also updated the change-log

khayss:
Back to this @franciszekjob 
I see 2 checks are failing, one is a typo which I'll fix in a new commit very soon and the other is Cast test. This again goes to show how tricky it is to implement a default name that needs to be deterministic in a way with the current setup. The reason is, to achieve this, you need to somehow add a new structure (even the easiest one such as incremental ids without accounting for gaps) to the accounts file, and adding anything there disrupts the initial setup.

I'll be expecting your review to know next steps to take.
Thank you.

crates/sncast/src/starknet_commands/account/import.rs
```diff
@@ -196,9 +201,57 @@ fn get_private_key_from_input() -> Result<Felt> {
     parse_input_to_felt(&input)
 }
 
+pub fn get_import_name(
+    import_name: Option<String>,
+    accounts_file: &Utf8PathBuf,
+) -> (String, Option<u32>) {
```
franciszekjob, line 207: 1. I think `generate_account_name` could be better.
2. The function should return only one value which is the account name.

crates/sncast/src/starknet_commands/account/import.rs
```diff
@@ -196,9 +201,57 @@ fn get_private_key_from_input() -> Result<Felt> {
...
+        std::fs::write(
+            accounts_file.clone(),
+            serde_json::to_string_pretty(&items).unwrap(),
+        )
+        .expect("Failed to write to file");
+
+        (format!("account-{default_id}"), Some(default_id))
+    }
+}
+
```
franciszekjob, line 250: The idea of calculating newly added account index should be simplified, there's no need to store information about total ids etc. 

The idea could look as follows:
- read all accounts with name matching `account-{i}` pattern
- extract their numbers
- find the first free smallest number which will be then used for new account name

Is that idea clear to you?

crates/sncast/src/starknet_commands/account/import.rs
franciszekjob: Please base on https://github.com/foundry-rs/starknet-foundry/blob/master/crates/sncast/tests/e2e/account/import.rs while adding tests.

franciszekjob (changes requested):

franciszekjob:
Hi @khayss, please have a look and address the comments I've added. Thanks.

franciszekjob:
Hey @khayss , just wanted to check out how's the progress :) ?

khayss:
Hey @franciszekjob I'll make a PR on this soon. Got all the requests and suggestions.

khayss:
Hello, @franciszekjob Made the requested changes. I'll be awaiting your review :)

crates/sncast/src/starknet_commands/account/import.rs
```diff
@@ -1,3 +1,5 @@
+use std::collections::HashSet;
```
khayss, line 1: I am using HashSet because it's faster considering there's the possibility of gaps between names. e.g. `account-1`, `account-2`, `account-3`. This way, we don't need to first sort if we were to use an array (`Vec`)

crates/sncast/src/starknet_commands/account/mod.rs
```diff
@@ -21,6 +21,8 @@ pub mod deploy;
 pub mod import;
 pub mod list;
 
+type NestedMap<T> = HashMap<String, HashMap<String, T>>;
```
khayss, line 24: I moved this up here since it's also used in the `read_and_flatten` function `list.rs` and `generate_account_name` function in `import.rs`. Thus, preventing duplicity.

khayss:
I see some tests are failing. Even the previous ones. Something about not able to read `account.json` file. Any idea about this?

crates/sncast/src/starknet_commands/account/import.rs
```diff
@@ -196,6 +201,30 @@ fn get_private_key_from_input() -> Result<Felt> {
...
+
+    let mut id = 1;
+
+    while result.contains(&id) {
+        id += 1;
+    }
+
+    Ok(format!("account-{id}"))
+}
+
```
franciszekjob, line 227: The logic here is good, it just needs to handle the case when there is no json file with the accounts. In such case, `read_and_parse_json_file` will return `Err`. We should handle it and return `Ok(format!("account-{id}"))` (`id` should be equal to `1` at that time, because it will be first imported account, obvious).

It could look like this:
```rust
fn generate_account_name(accounts_file: &Utf8PathBuf) -> anyhow::Result<String> {
    let mut id: u32 = 1;
    let networks: NestedMap<AccountData> = match read_and_parse_json_file(accounts_file) {
        Ok(networks) => networks,
        Err(_) => return Ok(format!("account-{id}")),
    };
    
    let mut result = HashSet::new();

    for (_network, accounts) in networks {
        for (name, _data) in accounts {
            if let Some(id) = name
                .strip_prefix("account-")
                .and_then(|id| id.parse::<u32>().ok())
            {
                result.insert(id);
            };
        }
    }

    while result.contains(&id) {
        id += 1;
    }

    Ok(format!("account-{id}"))
}
```

crates/sncast/tests/e2e/account/import.rs
```diff
@@ -683,3 +683,58 @@ pub async fn test_invalid_address_computation() {
         error: Computed address {computed_address} does not match the provided address 0x123. Please ensure that the provided salt, class hash, and account type are correct.
     "});
 }
+
+#[test_case("oz", "open_zeppelin"; "oz_account_type")]
+#[test_case("argent", "argent"; "argent_account_type")]
+#[test_case("braavos", "braavos"; "braavos_account_type")]
```
franciszekjob, line 689: No need to put it here, we can hardcode type to `"oz"`.

crates/sncast/tests/e2e/account/import.rs
```diff
@@ -683,3 +683,58 @@ pub async fn test_invalid_address_computation() {
...
+        "--url",
+        URL,
+        "--address",
+        "0x123",
+        "--private-key",
+        "0x456",
+        "--class-hash",
+        DEVNET_OZ_CLASS_HASH_CAIRO_0,
+        "--type",
+        input_account_type,
```
franciszekjob, line 709: Please change this to `"oz"`.

crates/sncast/tests/e2e/account/import.rs
```diff
@@ -683,3 +683,58 @@ pub async fn test_invalid_address_computation() {
...
+            {
+                "alpha-sepolia": {
+                  "account-1": {
+                    "address": "0x123",
+                    "class_hash": DEVNET_OZ_CLASS_HASH_CAIRO_0,
+                    "deployed": false,
+                    "legacy": true,
+                    "private_key": "0x456",
+                    "public_key": "0x5f679dacd8278105bd3b84a15548fe84079068276b0e84d6cc093eb5430f063",
+                    "type": saved_type
```
franciszekjob, line 734: Should be changed to `"open_zeppelin`".

crates/sncast/tests/e2e/account/import.rs
```diff
@@ -683,3 +683,58 @@ pub async fn test_invalid_address_computation() {
         error: Computed address {computed_address} does not match the provided address 0x123. Please ensure that the provided salt, class hash, and account type are correct.
     "});
 }
+
+#[test_case("oz", "open_zeppelin"; "oz_account_type")]
+#[test_case("argent", "argent"; "argent_account_type")]
+#[test_case("braavos", "braavos"; "braavos_account_type")]
+#[tokio::test]
+pub async fn test_happy_case_unnamed_import(input_account_type: &str, saved_type: &str) {
```
franciszekjob, line 691: After above comment, these params should be removed as well.

franciszekjob (changes requested):
Looks good 👍 
Please address the changes I requested and also refactor test. 
The test should import e.g. 3 accounts, then remove `account-2` and then use import again. We should check if first three imported accounts have correct names and whether re-import will result in correct name (should be `account-2, to fill the gap).

khayss:
Made the requested changes and refactored the test.

crates/sncast/tests/e2e/account/import.rs
```diff
@@ -683,3 +683,97 @@ pub async fn test_invalid_address_computation() {
...
+        add_profile: --add-profile flag was not set. No profile added to snfoundry.toml
+    "});
+    }
+
+    let snapbox = runner(&delete_args).current_dir(tempdir.path());
+    snapbox.assert().success().stdout_matches(indoc! {r"
+        command: account delete
+        result: Account successfully removed
+    "});
+
```
franciszekjob, line 733: Let's check the accounts file content here. It should only have `account-1` and `account-3`.

crates/sncast/tests/e2e/account/import.rs
```diff
@@ -683,3 +683,97 @@ pub async fn test_invalid_address_computation() {
         error: Computed address {computed_address} does not match the provided address 0x123. Please ensure that the provided salt, class hash, and account type are correct.
     "});
 }
+
+#[tokio::test]
+pub async fn test_happy_case_unnamed_import() {
```
franciszekjob, line 688: nit:
```suggestion
pub async fn test_happy_case_default_name_generation() {
```

crates/sncast/tests/e2e/account/import.rs
```diff
@@ -683,3 +683,97 @@ pub async fn test_invalid_address_computation() {
...
+                  },
+                    "account-3": {
+                        "address": "0x123",
+                        "class_hash": DEVNET_OZ_CLASS_HASH_CAIRO_0,
+                        "deployed": false,
+                        "legacy": true,
+                        "private_key": "0x456",
+                        "public_key": "0x5f679dacd8278105bd3b84a15548fe84079068276b0e84d6cc093eb5430f063",
+                        "type": "open_zeppelin"
+                    }
```
franciszekjob, line 774: Please fix indentation here.

crates/sncast/tests/e2e/account/import.rs
```diff
@@ -683,3 +683,97 @@ pub async fn test_invalid_address_computation() {
...
+    let snapbox = runner(&import_args).current_dir(tempdir.path());
+    snapbox.assert().stdout_matches(indoc! {r"
+        command: account import
+        add_profile: --add-profile flag was not set. No profile added to snfoundry.toml
+    "});
+
+    let contents = fs::read_to_string(tempdir.path().join(accounts_file))
+        .expect("Unable to read created file");
+    let contents_json: serde_json::Value = serde_json::from_str(&contents).unwrap();
+    assert_eq!(
```
franciszekjob, line 743:
```suggestion
    assert_eq(
```

franciszekjob (changes requested):
It's almost good to go 😄 .

Just please address the changes and fix test.

The test should work this way:
- import 3 accounts
- check the file contents
- delete `account-2`
- check the file contents
- import account again
- check the file contents

crates/sncast/tests/e2e/account/import.rs
khayss: I tried removing the exclamation mark and it threw an error, so I kept it.

crates/sncast/tests/e2e/account/import.rs
khayss: fixed

crates/sncast/tests/e2e/account/import.rs
khayss: renamed. Thanks :)

khayss:
reworked the test

khayss:
updated the docs.

CHANGELOG.md
```diff
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Cast
+
+#### Added
+
+- Feature to import an account without requiring the `--name` flag. This will under the hood generate a default name for the account This allows for easier and more streamlined account imports.
```
franciszekjob, line 14: Actually I liked the previous description more. Could you please bring it back?

khayss:
I have made the updates @franciszekjob 
The tests and linting should pass now.
Also updated the change-log

crates/sncast/src/starknet_commands/account/import.rs
```diff
@@ -196,6 +201,33 @@ fn get_private_key_from_input() -> Result<Felt> {
...
+            };
+        }
+    }
+
+    while result.contains(&id) {
+        id += 1;
+    }
+
+    format!("account-{id}")
+}
```
franciszekjob, line 229:
```suggestion
fn generate_account_name(accounts_file: &Utf8PathBuf) -> Result<String> {
    let mut id = 1;

    if !accounts_file.exists() {
        return Ok(format!("account-{id}"));
    }

    let networks: NestedMap<AccountData> = read_and_parse_json_file(accounts_file)?;
    let mut result = HashSet::new();

    for (_network, accounts) in networks {
        for (name, _data) in accounts {
            if let Some(id) = name
                .strip_prefix("account-")
                .and_then(|id| id.parse::<u32>().ok())
            {
                result.insert(id);
            };
        }
    }

    while result.contains(&id) {
        id += 1;
    }

    Ok(format!("account-{id}"))
}
```

crates/sncast/src/starknet_commands/account/import.rs
```diff
@@ -80,6 +83,8 @@ pub async fn import(
     };
     let private_key = &SigningKey::from_secret_scalar(*private_key);
 
+    let account = account.unwrap_or(generate_account_name(accounts_file));
```
franciszekjob, line 86:
```suggestion
    let account = account.unwrap_or(generate_account_name(accounts_file)?);
```

franciszekjob (changes requested):
Looks good! Please address suggestions that I've left and we we'll be close to merge.  

@cptartur @kkawula @ddoktorski could someone of you have a look as well?

khayss:
Made the requested changes

CHANGELOG.md
```diff
@@ -11,6 +11,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 #### Added
 
+- You can skip `--name` flag when using `account import` - a default name will be generated.
+
+## [Unreleased]
+
+### Cast
+
+#### Added
```
cptartur, line 20: You've duplicated the unreleased section

crates/sncast/src/starknet_commands/account/import.rs
```diff
@@ -196,6 +201,34 @@ fn get_private_key_from_input() -> Result<Felt> {
...
+
+    if !accounts_file.exists() {
+        return Ok(format!("account-{id}"));
+    }
+
+    let networks: NestedMap<AccountData> = read_and_parse_json_file(accounts_file)?;
+    let mut result = HashSet::new();
+
+    for (_network, accounts) in networks {
+        for (name, _data) in accounts {
```
cptartur, line 215:
```suggestion
    for (_, accounts) in networks {
        for (name, _) in accounts {
```

docs/src/starknet/account-import.md
```diff
@@ -136,4 +136,17 @@ $ sncast \
...
+To import an account with an auto-generated name, skip the `--name` flag.
+
+```shell
+$ sncast \
+    account import \
+    --url http://127.0.0.1:5050 \
+    --address 0x1 \
+    --private-key 0x2 \
+    --type oz
 ```
```
cptartur, line 152: Instead of this, I'd, under the shell in `#### General Example`, add something like this "If the `--name` is not provided a default name will be generated instead.

cptartur (commented):
One problem is see with this is that how does the user know what name was generated for them? Does it show in the console?

crates/sncast/src/starknet_commands/account/import.rs
khayss: I'll effect this suggestion.

CHANGELOG.md
khayss: Oopsie. I will remove it.

khayss: There's another unreleased for a different feature added there. Seems to be different and not a duplicate

cptartur: There should be one `[Unreleased]` section. We change this into a relevant version when releasing.

khayss: Got it. Thanks

khayss:
Yeah, I'll make the updates ASAP

crates/sncast/src/starknet_commands/account/import.rs
```diff
@@ -142,30 +149,27 @@ pub async fn import(
...
-            "--add-profile flag was not set. No profile added to snfoundry.toml".to_string()
-        },
+        add_profile: import.add_profile.as_ref().map_or_else(
+            || "--add-profile flag was not set. No profile added to snfoundry.toml".to_string(),
+            |profile_name| format!("Profile {profile_name} successfully added to snfoundry.toml"),
+        ),
+        account_name: account.map_or_else(
+            || Some(format!("Account imported with name: {account_name}")),
+            |_| None,
+        ),
```
khayss, line 172: Due to the limitation on the acceptable lines of code for a function, I have to refactor this to save up some lines as `cargo clippy` will flag the `import` function as having too many lines (>100).

crates/sncast/src/response/structs.rs
```diff
@@ -66,6 +66,7 @@ impl CommandResponse for AccountCreateResponse {}
 #[derive(Serialize)]
 pub struct AccountImportResponse {
     pub add_profile: String,
+    pub account_name: Option<String>,
```
khayss, line 69: This is added so we can show the generated account name on the terminal when `--name` flag is skipped.

khayss:
I have made the requested changes. Kindly review.

CHANGELOG.md
```diff
@@ -13,6 +13,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 #### Added
 
+- You can skip `--name` flag when using `account import` - a default name will be generated.
+
+### Cast
+
+#### Added
+
```
franciszekjob, line 21:
```suggestion
- You can skip `--name` flag when using `account import` - a default name will be generated.
```

franciszekjob (changes requested):
Please apply changes.

crates/sncast/src/starknet_commands/account/import.rs
```diff
@@ -80,6 +83,10 @@ pub async fn import(
     };
     let private_key = &SigningKey::from_secret_scalar(*private_key);
 
+    let account_name = account
+        .clone()
+        .unwrap_or_else(|| generate_account_name(accounts_file).unwrap());
```
khayss, line 88: I'm using `unwrap_or_else` here because, `unwrap_or` is eagerly compiled and since we were using the return of a function call in the `unwrap_or`, it means, the function will be evaluated even when we don't need the value.

franciszekjob (approved):

franciszekjob:
Lets wait for @cptartur approve as well.

docs/src/starknet/account.md
```diff
@@ -88,6 +88,7 @@ $ sncast \
     --private-key 0x2 \
     --type oz
 ```
+The `--name` can be omitted as this is optional. A default name will be generated for the account.
```
cptartur, line 91: I'd not add that here, just in the other file you added.

docs/src/starknet/account-import.md
```diff
@@ -80,6 +80,7 @@ $ sncast \
     --private-key 0x2 \
     --type oz
 ```
+The `--name` can be omitted as this is optional. A default name will be generated for the account.
```
cptartur, line 83:
```suggestion
> 📝 **Note** 
> The `--name` can be omitted as this is optional. A default name will be generated for the account.
```

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 #### Added
 
+- You can skip `--name` flag when using `account import` - a default name will be generated.
```
cptartur, line 16: Let's move this to `[Unreleased]` section

cptartur (approved):

cptartur:
@khayss Please just address the suggested changes and we are good to merge.

docs/src/starknet/account.md
khayss: removed

docs/src/starknet/account-import.md
khayss: Updated.

khayss:
@cptartur Made the requested changes.

cptartur (approved):

franciszekjob:
Thanks for contribution @khayss!

khayss:
My pleasure. Thanks for the great support throughout.

PR #2617, opened 2024-10-26
Add cast script to template when using snforge --init

CollinsC1O:
## Summary

This pull request adds a default deployment script template that is automatically included when 
initializing a new project with `snforge --init`. The script:

## Implementation Details:

- Adds deployment script in `scripts/deploy.cairo`
- Includes contract declaration and deployment logic
- Handles nonce and balance checks
- Uses project name substitution for contract name
- Provides helpful comments for users

Issue: #<2268>


Arcticae:
Hey @CollinsC1O this is not the solution we were hoping for, this is usage of a regular bash script for deployment, we have our own solution developed just for this - [cast script](https://foundry-rs.github.io/starknet-foundry/starknet/script.html).
Please rewrite the solution so that it uses it. 

Also, this doesn't add the script to the `init template`, please fix that

cptartur (changes requested):
As outlined above, please show usages of Cast script: https://foundry-rs.github.io/starknet-foundry/starknet/script.html

Arcticae:
@CollinsC1O bump

crates/forge/src/init.rs
```diff
@@ -75,6 +117,17 @@ fn overwrite_files_from_scarb_template(
     Ok(())
 }
 
+fn create_deployment_script(base_path: &Path, project_name: &str) -> Result<()> {
+    let scripts_dir = base_path.join("scripts");
+    fs::create_dir_all(&scripts_dir)?;
+    
+    let script_path = scripts_dir.join(DEPLOY_SCRIPT_NAME);
+    let script_contents = replace_project_name(DEPLOY_SCRIPT_TEMPLATE.as_bytes(), project_name)?;
```
Arcticae, line 125: I would advise using the file contents instead of embedding it in a string

crates/forge/src/init.rs
```diff
@@ -75,6 +117,17 @@ fn overwrite_files_from_scarb_template(
     Ok(())
 }
 
+fn create_deployment_script(base_path: &Path, project_name: &str) -> Result<()> {
+    let scripts_dir = base_path.join("scripts");
```
Arcticae, line 121: This is not enough, it needs its' own package 

Please read this part for how it should look:
https://foundry-rs.github.io/starknet-foundry/starknet/script.html#1-scripts-directory-with-all-the-scripts-in-the-same-workspace-with-cairo-contracts-default-for-sncast-script-init

Arcticae (changes requested):
Please add a test which runs the init and the script afterwards 🙏

crates/forge/src/init.rs
```diff
@@ -16,6 +16,48 @@ static TEMPLATE: Dir = include_dir!("starknet_forge_template");
...
+    ContractClassType,
+};
+
+// Default deployment script for {{ PROJECT_NAME }}
+fn main() {
+    // Getting the current nonce for deployment
+    let nonce = get_nonce().unwrap();
+    println!("Current nonce: {}", nonce);
+    
+    // Checkinng account balance before deployment
```
cptartur, line 35:
```suggestion
    // Checking account balance before deployment
```

crates/forge/src/init.rs
```diff
@@ -16,6 +16,48 @@ static TEMPLATE: Dir = include_dir!("starknet_forge_template");
...
+    let balance = get_balance().unwrap();
+    println!("Account balance: {}", balance);
+
+    // Declaring the contract using project name placeholder
+    // This will be replaced with the actual project name during initialization
```
cptartur, line 40: Don't think we should say anything about placeholder here, the comment will be included in the actual user generated template

crates/forge/src/init.rs
```diff
@@ -16,6 +16,48 @@ static TEMPLATE: Dir = include_dir!("starknet_forge_template");
...
+    let declare_result = declare("{{ PROJECT_NAME }}", ContractClassType::V2).expect('Declaration failed');
```
cptartur, line 41: Declare uses name of the contract not of the project, this can be either hardcoded to `HelloStarknet` or we can somehow pass it to this.

crates/forge/src/init.rs
```diff
@@ -16,6 +16,48 @@ static TEMPLATE: Dir = include_dir!("starknet_forge_template");
...
+    println!("Contract declared with class hash: {}", declare_result.class_hash);
+    
+    // Preparing to deploy your contract
+    let constructor_calldata = array![];    // empty array, incase of contracts without a constructor. You cad add constructor arguments if needed
```
cptartur, line 45:
```suggestion
    let constructor_calldata = array![];    // empty array, in case of contracts without a constructor. You can add constructor arguments if needed
```

crates/forge/src/init.rs
```diff
@@ -220,4 +275,4 @@ pub fn run(project_name: &str) -> Result<()> {
         .context("Failed to fetch created project")?;
 
     Ok(())
-}
+}
```
cptartur, line 278: Newline missing here

crates/forge/src/init.rs
cptartur: If we are defining the `deploy.cairo` file (or a package once we have the correct solution) let's just copy from there. This string is duplicating that file.

starknet_forge_template/scripts/deploy/Scarb.toml 
```diff
@@ -0,0 +1,7 @@
+[package]
+name = "deploy"
+version = "0.1.0"
+
+[dependencies]
+starknet = ">=2.8.0"
+sncast_std = { git = "https://github.com/foundry-rs/starknet-foundry.git", tag = "v0.32.0" }
```
Arcticae, line 7:
```suggestion
sncast_std = { git = "https://github.com/foundry-rs/starknet-foundry.git", tag = "v0.32.0" }

```

crates/forge/src/init.rs
```diff
@@ -75,6 +78,39 @@ fn overwrite_files_from_scarb_template(
     Ok(())
 }
 
+fn create_deployment_script(base_path: &Path, project_name: &str) -> Result<()> {
+    let scripts_dir = base_path.join("scripts");
+    let deploy_dir = scripts_dir.join(DEPLOY_PACKAGE_NAME);
```
Arcticae, line 83: why not move the dir altogether in one move?
`overwrite_files_from_scarb_template` is used for other directories, i think it fits the use case too

Arcticae (changes requested):
Please remember about implementing a test for this as well 🙏

crates/forge/src/init.rs
cptartur: Not sure if `overwrite_files_from_scarb_template` supports nested dirs, so it might require creating `scripts/deploy` and then `scripts/deploy/src` then calling `overwrite_files_from_scarb_template` for both of them correctly. 

But I think it's still better than duplicating the logic.

starknet_forge_template/scripts/deploy/src/lib.cairo
```diff
@@ -0,0 +1 @@
+pub mod deploy;
```
cptartur, line 1:
```suggestion
pub mod deploy;

```

Arcticae:
Closing due to inactivity

PR #2618, opened 2024-10-27
Byte array implementation

### LINKED ISSUE #2147: Remove `CairoSerialize` impl for String. Use `ByteArray` explicitly

snforge

### Description

title

Nemezjusz:
Closes #2147

## Introduced changes

- I removed the Implementation for for String and &str
- Runed Linting and Testing
- Update Structs/Enums

Let me know if anything needs to be changed or updated! :)

### DIFF
--- a/crates/cheatnet/tests/cheatcodes/deploy.rs
+++ b/crates/cheatnet/tests/cheatcodes/deploy.rs
@@ -155,7 +155,7 @@ fn deploy_contract_on_predefined_address_after_its_usage() {
         matches!(
             output,
             CallResult::Failure(CallFailure::Error { msg, .. })
-            if msg.contains("Requested contract address") && msg.contains("is not deployed")
+            if msg.to_string().as_str().contains("Requested contract address") && msg.to_string().as_str().contains("is not deployed")
         ),
         "Wrong error message"
     );
--- a/crates/cheatnet/tests/cheatcodes/deploy.rs
+++ b/crates/cheatnet/tests/cheatcodes/deploy.rs
@@ -155,7 +155,7 @@ fn deploy_contract_on_predefined_address_after_its_usage() {
         matches!(
             output,
             CallResult::Failure(CallFailure::Error { msg, .. })
-            if msg.to_string().as_str().contains("Requested contract address") && msg.to_string().as_str().contains("is not deployed")
+            if msg.to_string().contains("Requested contract address") && msg.to_string().contains("is not deployed")
         ),
         "Wrong error message"
     );
--- a/crates/conversions/src/byte_array.rs
+++ b/crates/conversions/src/byte_array.rs
@@ -5,6 +5,8 @@ use cairo_lang_runner::short_string::as_cairo_short_string_ex;
 use cairo_lang_utils::byte_array::{BYTES_IN_WORD, BYTE_ARRAY_MAGIC};
 use cairo_serde_macros::{CairoDeserialize, CairoSerialize};
 use starknet_types_core::felt::Felt as Felt252;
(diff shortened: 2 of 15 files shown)


crates/conversions/src/byte_array.rs
```diff
@@ -68,3 +70,32 @@ impl From<ByteArray> for String {
...
 }
+
+
+impl fmt::Display for ByteArray {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{self:?}") 
+    }
+}
+
+// we can implement that or dont use Serialize/Deserialize in CallResult/CallFailure in rpc.rs
```
ksew1, line 81: Let's delete the derivation of `Serialize`/`Deserialize` in `CallResult`/`CallFailure`, as it is unused anyway.

crates/sncast/src/lib.rs
```diff
@@ -531,20 +533,20 @@ pub fn get_block_id(value: &str) -> Result<BlockId> {
...
 #[derive(Debug, CairoSerialize)]
 pub struct ErrorData {
-    pub data: String,
+    pub data: ByteArray,
 }
 
 impl ErrorData {
     #[must_use]
-    pub fn new(data: String) -> Self {
+    pub fn new(data: ByteArray) -> Self {
```
ksew1, line 541: I would delete this method since there's no need for it. It is merely a proxy for the constructor.

crates/sncast/src/lib.rs
```diff
@@ -531,20 +533,20 @@ pub fn get_block_id(value: &str) -> Result<BlockId> {
...
         ErrorData { data }
     }
 }
 
 impl From<ContractErrorData> for ErrorData {
```
ksew1, line 546: Let's also delete this, as it is unused.

crates/conversions/src/byte_array.rs
```diff
@@ -68,3 +70,32 @@ impl From<ByteArray> for String {
         format!("{full_words_string}{pending_word_string}")
     }
 }
+
+
+impl fmt::Display for ByteArray {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{self:?}")
```
ksew1, line 77: In the `Display` implementation, we want to display the original string.

I would remove the `From<ByteArray> for String` conversion and integrate its code into the `Display` implementation.

ksew1 (commented):
A couple of comments overall everything looks good 😁. Don't forget to run `cargo fmt` before pushing so the CI jobs won't fail.

crates/cheatnet/tests/cheatcodes/deploy.rs
```diff
@@ -155,7 +155,7 @@ fn deploy_contract_on_predefined_address_after_its_usage() {
         matches!(
             output,
             CallResult::Failure(CallFailure::Error { msg, .. })
-            if msg.contains("Requested contract address") && msg.contains("is not deployed")
+            if msg.to_string().as_str().contains("Requested contract address") && msg.to_string().as_str().contains("is not deployed")
```
cptartur, line 158: Do we need both `to_string` and `as_str` here? I think contains should exist on String as well.

ksew1: We don't need `as_str` as compiler automatically casts this to `&str`. @Nemezjusz could you remove that?

ksew1 (approved):
LGTM, please address @cptartur comment

cptartur (approved):

PR #2624, opened 2024-10-30
Fix data transformer integration

### LINKED ISSUE #2588: Fix data transformer shell integration
The current format where data transformer uses `()` to indicate transformable input doesn't play nicely with bash/zsh.

Investigate possible interactions with different ways of passing calldata to shell (with focus on different brackets and parenthesis which are special characters in bash) and propose a new way of passing input in shell. Implement proposed method.

cptartur:
Closes #2588

## Introduced changes

Previous data transformer integration used `--calldata` argument (along with the old serialized input) and used `()` to indicate data used for transformation. `()` are special characters in bash so it did not work in most shells, requiring wrapping in `''` anyway. This PR changes this integration so data for transformation is passed as a single string to a new `--arguments` argument and doesn't require any special wrapping.

- Changed the syntax of data transformer so it works better with shells
- Introduced new `--arguments` argument to `sncast` to distinguish data transformation
- Fixed support for shortstrings in data transformer
- Added tests using `.sh` scripts for data transformation

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/data-transformer/src/transformer.rs
```diff
@@ -43,7 +43,8 @@ pub fn transform(
 }
 
 fn split_expressions(input: &str, db: &SimpleParserDatabase) -> Result<Vec<Expr>> {
-    let expr = parse_expression(input, db)?;
+    let input = format!("({input},)");
```
ddoktorski, line 46: Could you explain this?

crates/sncast/src/starknet_commands/deploy.rs
```diff
@@ -50,6 +49,18 @@ pub struct Deploy {
     pub rpc: RpcArgs,
 }
 
+#[derive(Debug, Clone, clap::Args)]
+#[group(multiple = false)]
+pub struct DeployArguments {
+    /// Arguments of the called function serialized as a series of felts
+    #[clap(short, long, value_delimiter = ' ', num_args = 1..)]
+    pub constructor_calldata: Option<Vec<String>>,
```
ddoktorski, line 57: Is this name the only reason that we need a separate struct instead of using `Arguments`?

franciszekjob (commented):
Could you also provide a more granular description on what has been changed and what was causing the problem?

crates/sncast/src/main.rs
```diff
@@ -146,6 +148,54 @@ enum Commands {
...
+
+impl From<DeployArguments> for Arguments {
+    fn from(value: DeployArguments) -> Self {
+        let DeployArguments {
+            constructor_calldata,
+            arguments,
+        } = value;
+        Self {
+            arguments,
+            calldata: constructor_calldata,
```
kkawula, line 194:
```suggestion
            calldata: constructor_calldata,
            arguments,
```

crates/data-transformer/src/calldata.rs
```diff
@@ -44,11 +23,6 @@ impl Calldata {
         class_definition: ContractClass,
         function_selector: &Felt,
     ) -> anyhow::Result<Vec<Felt>> {
-        match self {
-            Calldata::Serialized(serialized) => Ok(serialized),
-            Calldata::Expressions(ref expressions) => {
-                transform(expressions, class_definition, function_selector)
-            }
-        }
+        transform(&self.expressions, class_definition, function_selector)
```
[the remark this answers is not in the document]
cptartur: Removing entirely

docs/src/starknet/calldata-transformation.md
```diff
@@ -65,100 +66,122 @@ $ sncast --account myuser \
...
-> Calldata transformation is enabled only when following requirements are met:
->
-> * **Arguments form a tuple** (on a contrary to a standard serialized input which is whitespace-separated). Note the lack of commas in the example above - it signals Cast that the input is serialized
->
-> * If there is only one single argument, **it is also written as a tuple**\
-> (e.g. `--calldata (0x2137,)`)
->
-> * Expressions should match the **Cairo syntax** and use **allowed items** - see [Supported expressions](./calldata-transformation.md#supported-expressions)
+Instead of serializing calldata yourself, `sncast` allows passing it in far more handy, human-readable form - as a tuple
+of Cairo expressions with `--arguments` flag.
```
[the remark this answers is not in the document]
cptartur: Thanks for catching that

crates/data-transformer/src/transformer.rs
cptartur: We have to pass the arguments as a valid cairo expression to the parser. I though that it'd be easiest to just wrap everything in tuple since we already had code to handle that - since previous version of calldata passing used `--calldata (1_u8, 234)` syntax.

crates/sncast/src/starknet_commands/deploy.rs
cptartur: Yes, it's only the name. Don't want to break this, it was called `constructor_calldata` previously.

crates/sncast/tests/e2e/deploy.rs
```diff
@@ -286,25 +288,27 @@ async fn test_happy_case_with_constructor() {
 
 #[tokio::test]
 async fn test_happy_case_with_constructor_cairo_expression_calldata() {
+    let tempdir = create_and_deploy_oz_account().await;
+
     let args = vec![
         "--accounts-file",
-        ACCOUNT_FILE_PATH,
+        "accounts.json",
```
kkawula, line 295: Can you explain this update?

crates/data-transformer/src/transformer.rs
kkawula: Maybe there should be a short comment? It seems unclear to me too.

crates/sncast/tests/e2e/deploy.rs
cptartur: See above, the `let tempdir = create_and_deploy_oz_account().await;`. It deploys a new account and stores it files in the structure that is hardcoded within tempdir.

We change the current_dir of snapbox to tempdir below and there it finds this files.

It's hardcoded in other tests that use `create_and_deploy_oz_account`, don't think that's problematic.

crates/data-transformer/src/transformer.rs
cptartur: Added a comment

cptartur:
> Could you also provide a more granular description on what has been changed and what was causing the problem?

@franciszekjob Added to description.

crates/sncast/src/starknet_commands/deploy.rs
ddoktorski: It seems reasonable to me to make this breaking change and reuse the `Arguments` struct here. Maybe in a different PR and for a future release (not today's) but generally I think it should be done.

docs/src/starknet/calldata-transformation.md
```diff
@@ -65,100 +65,118 @@ $ sncast --account myuser \
...
-          0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff \
-      ) \
+      --arguments \
+'array![array![1, 2], array![3, 4, 5], array![6]],'\
+'12,'\
+'-128_i8,'\
+'"Some string (a ByteArray)",'\
+"('a shortstring', 32_u32),"\
+'true,'\
+'0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff' \
```
ddoktorski, line 138: nit:
Actually in this form (single quotes each line + bakslash) indentations and whitespaces seem to work correctly. If this is true, it would improve the readability of this example. Also the note below could be moved further down to highlight its relevance primarily for cases where a pair of single quotes is expanded on multiple lines.

ddoktorski (approved):

franciszekjob (approved):

docs/src/starknet/calldata-transformation.md
cptartur: I'll verify

cptartur: Seems to break with indents, I'll leave it as is.

PR #2631, opened 2024-10-31
Print USC compilation details

### LINKED ISSUE #2403: Print details about USC compilation process when running tests

ddoktorski:
Closes #2403 

## Introduced changes

- Print information about compiling Sierra as a spinner message
- Skip printing the path for unsaved files

## Checklist

- [X] Linked relevant issue
- [X] Added relevant tests


crates/universal-sierra-compiler-api/src/lib.rs
```diff
@@ -59,16 +59,26 @@ pub fn compile_sierra(sierra_contract_class: &Value, sierra_type: &SierraType) -
...
 ) -> Result<String> {
-    let mut usc_command = UniversalSierraCompilerCommand::new();
+    // Skip printing details when compiling unsaved sierra
+    // which occurs during test execution for some cheatcodes e.g. `replace_bytecode`
+    if is_saved_file {
+        println!(
+            "Compiling Sierra to Casm ({})",
+            sierra_file_path.canonicalize_utf8()?
+        );
+    }
```
franciszekjob, line 79: I think we can slightly improve styling displayed message 😄 .

Before:
image

After:
image

Note that I've also added `println!("\n")` in [this line](https://github.com/foundry-rs/starknet-foundry/blob/bdec9564cf5d0c05d83a66acb2945eb8d426cdb2/crates/forge/src/run_tests/workspace.rs#L70).

Here's an example code that achieves this
```rust
if is_saved_file {
        println!(
            "{} ({})",
            style("Compiling Sierra to Casm").bold(),
            sierra_file_path.canonicalize_utf8()?,
    );
}
```

franciszekjob: nit: Just for the sake of uniformity, can we use `CASM` instead of `Casm` 😅 ? In the vast majority of places in the docs we use screaming case version.

kkawula (commented):
Spinner looks cool, but when compilation takes milliseconds it appears and disappears as fast we can't read it. So maybe to inform user whats going on, we should add heading about compilation?
Sth like:
```
Compilation Sierra to Casm...
<spinner>
```

crates/universal-sierra-compiler-api/src/utils.rs
```diff
@@ -0,0 +1,30 @@
+use anyhow::Result;
+use camino::Utf8PathBuf;
+use indicatif::{ProgressBar, ProgressStyle};
+use std::env;
+use std::time::Duration;
+
+pub fn get_spinner_message(sierra_file_path: &Utf8PathBuf) -> Result<ProgressBar> {
```
cptartur, line 7: I'd rename this since it spawns a working spinner

crates/universal-sierra-compiler-api/src/utils.rs
```diff
@@ -0,0 +1,30 @@
...
+    };
+    spinner.set_message(message);
+
+    Ok(spinner)
+}
+
+fn is_temp_file(file_path: &Utf8PathBuf) -> Result<bool> {
+    let temp_dir = env::temp_dir().canonicalize()?;
+    Ok(file_path.canonicalize()?.starts_with(temp_dir))
+}
```
cptartur, line 30: I wonder if there are cases in which this will not correctly recognise the temp dir. 
Though since it's just for outputs I guess it's not a problem anyway.

crates/universal-sierra-compiler-api/src/lib.rs
ddoktorski: Renamed both to `SIERRA` and `CASM`

crates/universal-sierra-compiler-api/src/utils.rs
ddoktorski: It'd be problematic if we want to test it, but I haven't found any simple way to test it anyway. Other than that as you said it is just for output.

crates/universal-sierra-compiler-api/src/utils.rs
```diff
@@ -0,0 +1,30 @@
...
+    spinner.set_style(ProgressStyle::with_template("\n{spinner} {msg}\n")?);
+    spinner.enable_steady_tick(Duration::from_millis(100));
+
+    // Skip printing path when compiling unsaved sierra
+    // which occurs during test execution for some cheatcodes e.g. `replace_bytecode`
+    let message = if is_temp_file(sierra_file_path)? {
+        "Compiling SIERRA to CASM".to_string()
+    } else {
+        format!(
+            "Compiling SIERRA to CASM ({})",
```
franciszekjob, line 18: nit: I think `Sierra` should stay as it was.
```suggestion
        "Compiling Sierra to CASM".to_string()
    } else {
        format!(
            "Compiling Sierra to CASM ({})",
```

franciszekjob:
> Spinner looks cool, but when compilation takes milliseconds it appears and disappears as fast we can't read it. So maybe to inform user whats going on, we should add heading about compilation? Sth like:
> 
> ```
> Compiling Sierra to Casm...
> <spinner>
> ```

Do you mean by that `Compiling Sierra to Casm...` should stay but spinner with file path would disappear? If so, I don't think it's a good idea. The motivation behind this PR was to inform user that the compilation is actually happening (when it takes longer) and that there isn't any problem with forge. And the current solution provides exactly the functionality that corresponds to this.

franciszekjob (approved):
We can think about adding [a simple message](https://github.com/foundry-rs/starknet-foundry/pull/2631#pullrequestreview-2413056927) that would disappear with spinner as well.

Overall, I like this, it provides exactly the functionality that solves the issue.

cptartur:
> We can think about adding [a simple message](https://github.com/foundry-rs/starknet-foundry/pull/2631#pullrequestreview-2413056927) that would disappear with spinner as well but this is not necessary.
> 
> Overall, I like this 👍 .

@franciszekjob if it's going to disappear then I don't think it would change much in terms of readability

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 #### Added
 
 - `generate_random_felt()` for generating (pseudo) random felt value.
+- Printing information about compiling Sierra files by `universal-sierra-compiler` before tests execution
```
cptartur, line 15:
```suggestion
- Printing information about compiling Sierra files by `universal-sierra-compiler`
```

It can also happen during tests execution right?

crates/universal-sierra-compiler-api/src/lib.rs
```diff
@@ -67,8 +69,9 @@ pub fn compile_sierra_at_path(
     sierra_file_path: &Utf8PathBuf,
     sierra_type: &SierraType,
 ) -> Result<String> {
-    let mut usc_command = UniversalSierraCompilerCommand::new();
+    let printed_message = spawn_spinner_message(sierra_file_path)?;
```
cptartur, line 72: Nit, 
```suggestion
    let spinner = spawn_spinner_message(sierra_file_path)?;
```

crates/universal-sierra-compiler-api/src/utils.rs
cptartur, line 1: Maybe let's name it `spinner.rs` or similar or eventually it will become a file with random collection of functions.

franciszekjob:
> > We can think about adding [a simple message](https://github.com/foundry-rs/starknet-foundry/pull/2631#pullrequestreview-2413056927) that would disappear with spinner as well but this is not necessary.
> > Overall, I like this 👍 .
> 
> @franciszekjob if it's going to disappear then I don't think it would change much in terms of readability

@cptartur I don't refer to situation when it compilation is very quick. My thought is that it will be helpful when compilation takes loner time (in fact that's the clue of issue) so people will know that this spinner indicates Sierra to CASM compilation and not something else.

franciszekjob:
> We can think about adding https://github.com/foundry-rs/starknet-foundry/pull/2631#pullrequestreview-2413056927 that would disappear with spinner as well but this is not necessary.

Nevermind, I didn't notice that the message is already getting printed 😅 , all fine.

CHANGELOG.md
ddoktorski: Correct, forgot to update this

cptartur (approved):

crates/universal-sierra-compiler-api/src/utils.rs
ddoktorski: On second thoughts, these are actually no acronyms, hence we should stick to the forms `Sierra` and `Casm`. This is also how they are mostly referred to in the starknet documentation.

PR #2649, opened 2024-11-07
Add edition entry in each `Scarb.toml`

### LINKED ISSUE #2632: Add missing `edition` field to Scarb packages in the repository
In the upcoming Scarb 2.9.0, a warning will be thrown in case of package without defined `edition` field.

Set it so that using packages from our repo (like `snforge_std`) doesn't cause these warnings for the users.

kkawula:
Closes #2632

## Introduced changes

-

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests

### DIFF
--- a/crates/cheatnet/tests/contracts/Scarb.toml
+++ b/crates/cheatnet/tests/contracts/Scarb.toml
@@ -1,7 +1,7 @@
 [package]
 name = "cheatnet_testing_contracts"
 version = "0.1.0"
-edition = "2024_07"
+edition = "2023_01"
 
 # See more keys and their definitions at https://docs.swmansion.com/scarb/docs/reference/manifest
 [[target.starknet-contract]]
--- a/crates/cheatnet/tests/contracts/Scarb.toml
+++ b/crates/cheatnet/tests/contracts/Scarb.toml
@@ -1,6 +1,7 @@
 [package]
 name = "cheatnet_testing_contracts"
 version = "0.1.0"
+edition = "2024_07"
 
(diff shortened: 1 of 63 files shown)


kkawula:
Tried to add the latest version possible.

franciszekjob (commented):
What if for some reason someone adds a package which doesn't have `edition` entry in its `Scarb.toml`? I believe we will not notice it. I think we should add test to check if every `Scarb.toml` has `edition` entry.

See [this CI step](https://github.com/foundry-rs/starknet-foundry/blob/41dd99ef9dee9fa8767760d4b9bb3c87ed5c7c46/.github/workflows/ci.yml#L228) which searches for all packages and check their formatting. We could add a similar one (display content of file and capture edition) - its shouldn't be a big effort.

cptartur:
> What if for some reason someone adds a package which doesn't have `edition` entry in its `Scarb.toml`? I believe we will not notice it. I think we should add test which would check this.
> 
> See [this CI step](https://github.com/foundry-rs/starknet-foundry/blob/41dd99ef9dee9fa8767760d4b9bb3c87ed5c7c46/.github/workflows/ci.yml#L228) which searches for all packages and check their formatting. We could add a similar one (display content of `Scarb.toml` and capture `edition` entry) - its shouldn't be a big effort.

Scarb 2.9.0 will emit a warning. Once we have tests running on different versions of Scarb, this could be testable without writing a script.

I'd create an issue and postpone resolving that until that is resolved.

cptartur (approved):
Let's create a follow-up issue to migrate them all to 2024_07.

PR #2657, opened 2024-11-08
Ensure tests can be run with different Scarb versions

ddoktorski:
Towards #2578 

## Introduced changes

- Modified tests to ensure that they can be run successfully for different Scarb versions
- Added logic to set correct `assert_macros` version in tests
- Removed `printing_in_contracts` test since it doesn't test any foundry logic, but only Scarb output for a specific case
- Removed `test_fail_on_scarb_version_lt_2_8_0`

## Checklist

- [X] Linked relevant issue
- [X] Added relevant tests


crates/forge/tests/e2e/coverage.rs
```diff
@@ -36,7 +36,7 @@ fn test_coverage_project_and_pass_args() {
 }
 
 #[test]
-#[cfg_attr(feature = "scarb_2_8_3", ignore)]
+#[cfg_attr(not(feature = "scarb_2_7_1"), ignore)]
 fn test_fail_on_scarb_version_lt_2_8_0() {
```
cptartur, line 40: Let's maybe rename it then

crates/forge/tests/e2e/running.rs
```diff
@@ -958,37 +958,6 @@ fn should_panic() {
     );
 }
 
-#[test]
-fn printing_in_contracts() {
```
cptartur, line 962: Why was this removed?

ddoktorski: Copied from the PR description:
- Removed `printing_in_contracts` test since it doesn't test any foundry logic, but only Scarb output for a specific case

In more details, for some Scarb versions it returns a warning and for other it doesn't, which is a problem when we want to test different versions. Regardless, I think this test is simply not relevant to Foundry.

crates/forge/tests/e2e/coverage.rs
ddoktorski: I believe this name is still valid as we want to test that it fails for versions less than `2.8.0`. We're currently using Scarb `2.7.1` for this, but we could technically use any other version.

cptartur (approved):

Issue #2479, opened 2024-09-17
Automatic validation of shell commands in docs


cptartur:
### Which components does the task require to be changed? (think hard pls)

snforge, sncast

### Description

Right now, we do not have any mechanism for validating if shell commands (e.g. calls to `snforge test` etc) in docs are up to date and correct. We should ensure that any command that user can copy from the documentation works and behaves as described.

We should introduce a mechanism to validate if outputs included in the documentation match these produced by the real program.

PR #2658, opened 2024-11-08
Add expandable shell snippet outputs

franciszekjob:
Relates #2479 

## Introduced changes

Update command snippets in docs:
- Added JS script which removes `$` from copied command snippet
- Split the way that output is an expandable block, by default it's expanded

Before:
image

After:
image

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


docs/src/getting-started/first-steps.md
```diff
@@ -6,14 +6,17 @@ We demonstrate how to create a new project, compile, and test it.
...
 Let's check out the project structure
 
 ```shell
-$ cd project_name
-$ tree . -L 1
+cd project_name
+tree . -L 1
+```
+
+```shell
```
franciszekjob, line 19: note: I wonder if having result of `tree` command is better in following form 🤔 :
image

Imo it should be always visible as the structure of directories is a key point here.

docs/codeSnippets.js
```diff
@@ -0,0 +1,29 @@
...
+ */
+function removePrompt(text) {
+    return text.split('\n').map(line => {
+        if (line.startsWith("$ ")) {
+            return line.slice(2);
+        }
+        return line;
+    }).join('\n');
+}
+
```
cptartur, line 15: Shouldn't it just remove the leading `$`?

docs/src/starknet/fees-and-versions.md
```diff
@@ -57,6 +57,7 @@ $ sncast account deploy \
     --version v1 \
     --max-fee 9999999999999
 ```
+</details>
```
cptartur, line 60: This has no beggining

cptartur (changes requested):
Can we default to details being closed?

docs/codeSnippets.js
franciszekjob: No because we can have multiline snippets, e.g.
```shell
$ echo abc
$ echo def
```

docs/src/starknet/fees-and-versions.md
franciszekjob: Removed, was there my mistake 😅 .

franciszekjob:
> Can we default to details being closed?

We can, but I wonder if that's good for a user? I believe people can sometimes not notice that the output is expandable 🤔 
If the code is runnable and output will be shown once "run" arrow is click then it's a bit more natural I think. 

I don't have a strong opinion on this though.

docs/codeSnippets.js
cptartur: Okay, do we have any snippets like this at the moment?

cptartur:
> > Can we default to details being closed?
> 
> We can, but I wonder if that's good for a user? I believe people can sometimes not notice that the output is expandable 🤔 If the code was runnable and output would be shown once "run" button was clicked then it's a bit more natural.
> 
> I don't have a strong opinion on this though.

I think showing the outputs by default just introduce clutter to the page. They are here for those who need it, but I think most users don't really care about them.

cptartur (approved):
Looks good. See my comment above

docs/codeSnippets.js
franciszekjob: Yes, e.g. [here](https://github.com/foundry-rs/starknet-foundry/blob/franciszekjob/2479-expandable-shell-snippets/docs/src/getting-started/first-steps.md).

image

franciszekjob:
> > > Can we default to details being closed?
> > 
> > 
> > We can, but I wonder if that's good for a user? I believe people can sometimes not notice that the output is expandable 🤔 If the code was runnable and output would be shown once "run" button was clicked then it's a bit more natural.
> > I don't have a strong opinion on this though.
> 
> I think showing the outputs by default just introduce clutter to the page. They are here for those who need it, but I think most users don't really care about them.

Ok, let's make the details closed as default then 👍 .

PR #2660, opened 2024-11-12
Automatic docs shell snippets validation

### LINKED ISSUE #2479: Automatic validation of shell commands in docs

franciszekjob:
Closes #2479 

## Introduced changes

-

## Checklist

- [ ] Linked relevant issue
- [ ] Added relevant tests

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4772,11 +4772,18 @@ dependencies = [
  "console",
  "regex",
  "semver",
+ "shell-words",
  "snapbox",
  "starknet",
  "starknet-types-core",
  "url",
 ]
+
+[[package]]
+name = "shell-words"
+version = "1.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "24188a676b6ae68c3b2cb3a01be17fbf7240ce009799bb56d5b1409051e78fde"
 
 [[package]]
 name = "shellexpand"
--- a/crates/shared/Cargo.toml
+++ b/crates/shared/Cargo.toml
@@ -13,3 +13,4 @@ starknet.workspace = true
 url.workspace = true
 regex.workspace = true
 snapbox.workspace = true
+shell-words = "1.0.0"
--- a/crates/shared/src/test_utils/docs_validation.rs
+++ b/crates/shared/src/test_utils/docs_validation.rs
@@ -1,0 +1,76 @@
+use regex::Regex;
+use std::env;
+use std::fs;
+use std::io;
+use std::path::Path;
+use std::path::PathBuf;
+
+pub fn extract_content_from_file(
+    file_path: &Path,
+    re: &Regex,
+) -> Result<Vec<String>, std::io::Error> {
+    let content = fs::read_to_string(file_path)?;
+    let matches: Vec<String> = re
+        .captures_iter(&content)
+        .filter_map(|caps| caps.get(1).map(|m| m.as_str().to_string()))
+        .collect();
+
+    Ok(matches)
+}
+
+pub fn extract_from_directory(
+    dir_path: &Path,
+    re: &Regex,
+    extension: Option<&str>,
+) -> io::Result<Vec<String>> {
+    let mut all_snippets = Vec::new();
+
+    for entry in fs::read_dir(dir_path)? {
+        let entry = entry?;
+        let path = entry.path();
+
+        if path.is_dir() {
+            all_snippets.extend(extract_from_directory(&path, re, extension)?);
+        } else if let Some(ext) = extension {
+            if path.extension().and_then(|e| e.to_str()) == Some(ext) {
+                let snippets = extract_content_from_file(&path, re)?;
+                if !snippets.is_empty() {
+                    all_snippets.extend(snippets);
+                }
+            }
+        } else {
+            let snippets = extract_content_from_file(&path, re)?;
+            if !snippets.is_empty() {
+                all_snippets.extend(snippets);
+            }
+        }
+    }
+
+    Ok(all_snippets)
+}
+
+#[must_use]
+pub fn get_parent_dir(levels_up: usize) -> PathBuf {
+    let mut dir = env::current_dir().expect("Failed to get the current directory");
+
+    for _ in 0..levels_up {
+        dir = dir.parent().expect("Failed to navigate up").to_owned();
+    }
+
+    dir
+}
+
+pub fn parse_snippet_str_to_command_args(snippet: &str) -> Vec<String> {
+    let cleaned_snippet = snippet
+        .lines()
+        .map(str::trim_end)
+        .collect::<Vec<&str>>()
+        .join(" ")
+        .replace(" \\", "");
+
+    shell_words::split(&cleaned_snippet)
+        .expect("Failed to parse snippet string")
+        .into_iter()
+        .map(|arg| arg.trim().to_string())
+        .collect()
+}
--- a/crates/shared/src/test_utils/docs_validation.rs
+++ b/crates/shared/src/test_utils/docs_validation.rs
@@ -38,11 +38,6 @@ pub fn extract_from_directory(
                     all_snippets.extend(snippets);
                 }
             }
-        } else {
-            let snippets = extract_content_from_file(&path, re)?;
-            if !snippets.is_empty() {
-                all_snippets.extend(snippets);
-            }
         }
     }
 
--- a/crates/shared/src/test_utils/mod.rs
+++ b/crates/shared/src/test_utils/mod.rs
@@ -1,2 +1,3 @@
+pub mod docs_validation;
 pub mod node_url;
 pub mod output_assert;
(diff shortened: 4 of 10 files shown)


crates/sncast/tests/docs_snippets/validation.rs
franciszekjob, line 1: note: As mentioned in https://github.com/foundry-rs/starknet-foundry/issues/2678, ultimately we should use `Cli::try_parse_from` instead of `runner`.

crates/docs/src/validation.rs
```diff
@@ -0,0 +1,70 @@
+use regex::Regex;
+use std::{
+    env, fs, io,
+    path::{Path, PathBuf},
+};
+
+pub fn extract_matches_from_file(
+    file_path: &Path,
+    re: &Regex,
+) -> Result<Vec<String>, std::io::Error> {
```
cptartur: Can this use `io::Result` like the function below?

crates/docs/src/validation.rs
```diff
@@ -0,0 +1,70 @@
...
+    let content = fs::read_to_string(file_path)?;
+    let matches: Vec<String> = re
+        .captures_iter(&content)
+        .filter_map(|caps| caps.get(1).map(|m| m.as_str().to_string()))
+        .collect();
```
cptartur: Nit but maybe this can be flattened & renamed for readability

```suggestion
    let matches: Vec<String> = re
        .captures_iter(&content)
        .filter_map(|captures| captures.get(1))
        .map(|m| m.as_str().to_string())
        .collect();
```

crates/docs/src/validation.rs
```diff
@@ -0,0 +1,70 @@
...
+    re: &Regex,
+    extension: Option<&str>,
+) -> io::Result<Vec<String>> {
+    let mut all_snippets = Vec::new();
+
+    for entry in fs::read_dir(dir_path)? {
+        let path = entry?.path();
+
+        if path.is_dir() {
+            all_snippets.extend(extract_matches_from_directory(&path, re, extension)?);
```
cptartur: Maybe let's use this library instead? https://github.com/BurntSushi/walkdir

We already have it in workspace dependencies.

crates/docs/src/validation.rs
```diff
@@ -0,0 +1,70 @@
...
+            let snippets = extract_matches_from_file(&path, re)?;
+            all_snippets.extend(snippets);
+        }
+    }
+
+    Ok(all_snippets)
+}
+
+#[must_use]
+pub fn get_parent_dir(levels_up: usize) -> PathBuf {
```
cptartur, line 81: If this is only used for tests, can we make it available only to tests? Perhaps some combination of `#[cfg(test)]` could do that?

crates/docs/src/validation.rs
```diff
@@ -0,0 +1,70 @@
...
+        dir = dir
+            .parent()
+            .expect("Failed to navigate to parent directory")
+            .to_owned();
+    }
+
+    dir
+}
+
+pub fn snippet_to_command_args(snippet: &str) -> Vec<String> {
```
cptartur: Same goes for this

crates/forge/tests/e2e/docs_snippets_validation.rs
```diff
@@ -0,0 +1,43 @@
...
+        vec!["snforge", "--version"],
+    ];
+
+    for snippet in &snippets {
+        let args = snippet_to_command_args(snippet.as_str());
+        let args: Vec<&str> = args.iter().map(String::as_str).collect();
+
+        if skipped_args.contains(&args) {
+            continue;
+        }
```
cptartur, line 29: I think this should maybe print a warning or something when skipping. I see that eventually someone might extend `skipped_args` with something for whatever reason and we won't be aware of that.

crates/sncast/tests/docs_snippets/validation.rs
```diff
@@ -0,0 +1,75 @@
...
+        .expect("Failed to extract sncast command snippets");
+
+    let snippets = docs_snippets
+        .into_iter()
+        .chain(readme_snippets)
+        .collect::<Vec<String>>();
+
+    let skipped_args = [
+        // snippet "$ sncast <subcommand>"
+        vec!["<subcommand>"],
```
cptartur, line 33: Why do we need that?

crates/sncast/tests/docs_snippets/validation.rs
```diff
@@ -0,0 +1,75 @@
...
+    ];
+
+    for snippet in &snippets {
+        let args = snippet_to_command_args(snippet.as_str());
+        let mut args: Vec<&str> = args.iter().map(String::as_str).collect();
+        args.remove(0);
+
+        if skipped_args.contains(&args) {
+            continue;
+        }
```
cptartur, line 56: Same warning as in forge.

crates/sncast/tests/docs_snippets/validation.rs
```diff
@@ -0,0 +1,75 @@
...
+            "--url",
+            "http://127.0.0.1:5050",
+            "--name",
+            "account_123",
+            "--address",
+            "0x1",
+            "--type",
+            "oz",
+        ],
+    ];
```
cptartur, line 47: I wonder if we should be defining what to skip here. 
This is probably okay as an initial solution but I think ideally snippets to be skipped should probably somehow marked in markdown instead.

cptartur: You can create a follow-up task for that if this cannot be resolved easily.

crates/sncast/tests/docs_snippets/validation.rs
```diff
@@ -0,0 +1,75 @@
...
+    extract_matches_from_directory, extract_matches_from_file, get_parent_dir,
+    snippet_to_command_args,
+};
+use regex::Regex;
+use tempfile::tempdir;
+
+use crate::helpers::runner::runner;
+
+#[test]
+fn test_docs_snippets() {
```
cptartur, line 11: This is kind of duplicating the code we have in forge. Could this be made a reusable function or parts of it somehow?

cptartur (changes requested):

crates/sncast/tests/docs_snippets/validation.rs
franciszekjob: It's because in docs we have such example
https://foundry-rs.github.io/starknet-foundry/starknet/index.html?highlight=subcommand#how-to-use-sncast

crates/sncast/tests/docs_snippets/validation.rs
franciszekjob: > I wonder if we should be defining what to skip here. This is probably okay as an initial solution but I think ideally snippets to be skipped should probably somehow marked in markdown instead.

Yup, makes sense.

crates/docs/src/validation.rs
```diff
@@ -0,0 +1,115 @@
...
+
+    Ok(snippets)
+}
+
+pub fn extract_snippets_from_directory(dir_path: &Path, re: &Regex) -> io::Result<Vec<Snippet>> {
+    let mut all_snippets = Vec::new();
+
+    let entries = walkdir::WalkDir::new(dir_path)
+        .into_iter()
+        .filter_map(Result::ok);
```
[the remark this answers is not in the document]
franciszekjob: I think we can leave it as standard print, we skip command on purpose so warning may be too much?

crates/docs/src/validation.rs
```diff
@@ -0,0 +1,115 @@
...
+
+pub fn print_success_message(snippets_len: usize, tool_name: &str) {
+    println!("Successfully validated {snippets_len} {tool_name} docs snippets");
+}
+
+pub fn print_skipped_snippet_message(snippet: &Snippet, tool_name: &str) {
+    println!(
+        "Skipped validation of {} snippet in the docs in file: {} at line {}",
+        tool_name, snippet.file_path, snippet.line_start
+    );
```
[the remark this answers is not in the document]
franciszekjob: I think we can leave it as standard print, we skip command on purpose so warning may be too much?

crates/docs/src/validation.rs
```diff
@@ -0,0 +1,115 @@
...
+    }
+}
+
+pub fn extract_snippets_from_file(file_path: &Path, re: &Regex) -> io::Result<Vec<Snippet>> {
+    let content = fs::read_to_string(file_path)?;
+    let mut snippets = Vec::new();
+
+    for caps in re.captures_iter(&content) {
+        if let Some(command_match) = caps.get(1) {
+            let match_position = content[..caps.get(0).unwrap().start()].lines().count();
```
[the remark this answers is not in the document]
franciszekjob: It simply iterates over the regex matches and gathers information about the snippet command, starting line and file path. Refactored, should be a bit cleaner now.

cptartur (approved):

PR #2662, opened 2024-11-12
Add scheduled workflow to test against different Scarb versions

### LINKED ISSUE #2578: Automatic testing of foundry against different Scarb versions

ddoktorski:
Closes #2578

## Introduced changes

- Added a scheduled workflow to run all unit, integration and e2e tests for `sncast` and `snforge` for different Scarb versions, every Wednesday and Sunday at 0:00 UTC

## Checklist

- [X] Linked relevant issue
- [X] Added relevant tests

### DIFF
--- a/.github/workflows/scheduled.yml
+++ b/.github/workflows/scheduled.yml
@@ -1,0 +1,128 @@
+name: Scheduled
+
+on:
+  pull_request:
+    paths:
+      - scripts/get_scarb_versions.sh
+      - .github/workflows/scheduled.yml
+  schedule:
+    - cron: '0 0 * * 3,6'
+
+jobs:
+  get-scarb-versions:
+    name: Get Scarb versions
+    outputs:
+      versions: ${{ steps.get_versions.outputs.versions }}
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: asdf-vm/actions/install@05e0d2ed97b598bfce82fd30daf324ae0c4570e6
+        with:
+          tool_versions: |
+            scarb latest
+
+      - name: Get versions
+        id: get_versions
+        run: |
+          scarb_versions=$(./scripts/get_scarb_versions.sh)
+          echo ${scarb_versions[@]}
+          echo "versions=[${scarb_versions[@]}]" >> "$GITHUB_OUTPUT"
+
+  test-forge-unit-and-integration:
+    runs-on: ubuntu-latest
+    needs: get-scarb-versions
+    strategy:
+      fail-fast: false
+      matrix:
+        version: ${{ fromJSON(needs.get-scarb-versions.outputs.versions) }}
+
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@stable
+      - uses: Swatinem/rust-cache@82a92a6e8fbeee089604da2575dc567ae9ddeaab
+      - uses: software-mansion/setup-scarb@v1
+        with:
+          scarb-version: ${{ matrix.version }}
+      - uses: software-mansion/setup-universal-sierra-compiler@v1
+      - name: Install cairo-profiler
+        run: |
+          curl -L https://raw.githubusercontent.com/software-mansion/cairo-profiler/main/scripts/install.sh | sh
+
+      - run: cargo test --release --lib -p forge
+      - run: cargo test --release -p forge integration
+
+  test-forge-e2e:
+    runs-on: ubuntu-latest
+    needs: get-scarb-versions
+    strategy:
+      fail-fast: false
+      matrix:
+        version: ${{ fromJSON(needs.get-scarb-versions.outputs.versions) }}
+
+    steps:
+      - name: Extract branch name
+        if: github.event_name != 'pull_request'
+        run: echo "BRANCH_NAME=$(echo ${GITHUB_REF#refs/heads/})" >> $GITHUB_ENV
+
+      - name: Extract branch name on pull request
+        if: github.event_name == 'pull_request'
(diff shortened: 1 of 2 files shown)


.github/workflows/scheduled.yml
```diff
@@ -0,0 +1,128 @@
+name: Scheduled
+
+on:
+  pull_request:
+    paths:
+      - scripts/get_scarb_versions.sh
+      - .github/workflows/scheduled.yml
+  schedule:
+    - cron: '0 0 * * 3,6'
```
cptartur, line 9: Nit, this could maybe run on Wednesday and Sunday instead? If it fails on a weekend, it will have to wait until Monday to be resolved anyway.

.github/workflows/scheduled.yml
```diff
@@ -0,0 +1,128 @@
...
+    outputs:
+      versions: ${{ steps.get_versions.outputs.versions }}
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: asdf-vm/actions/install@05e0d2ed97b598bfce82fd30daf324ae0c4570e6
+        with:
+          tool_versions: |
+            scarb latest
+
```
cptartur, line 23: What's the purpose of this?

scripts/get_scarb_versions.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+set -e
+
+# This script is used to find the following Scarb versions:
+# The current major.minor version along with all its patch versions
+# The latest patch versions for the two versions preceding the current one
+
+function get_all_patch_versions() {
+  # Omit version 2.8.0 as it has a bug when using the `assert_macros` package
```
cptartur, line 9: Let's create an issue to remove this step once we stop supporting `2.8.0`.

scripts/get_scarb_versions.sh
```diff
@@ -0,0 +1,29 @@
...
+  asdf list all scarb $1 | grep -v "rc" | grep -v "2.8.0"
+}
+
+function get_latest_patch_version() {
+  get_all_patch_versions $1 | sort -u | tail -1
+}
+
+major_minor_versions=($(get_all_patch_versions | cut -d . -f 1,2 | sort -u | tail -3))
```
cptartur, line 17: `sort` has a dedicated option for sorting versions `-V`, maybe that should be used here?

cptartur (commented):
@THenry14 can you review please?

.github/workflows/scheduled.yml
ddoktorski: To execute commands like `asdf list all scarb` the Scarb plugin must first be installed. I expected that the asdf action would automatically install Scarb since it is listed in the `.tool_versions` file in the repo, but for some  reason this was not the case and it was necessary to explicitly provide this argument here. So the purpose is solely to install Scarb asdf plugin.

scripts/get_scarb_versions.sh
ddoktorski: https://github.com/foundry-rs/starknet-foundry/issues/2685

scripts/get_scarb_versions.sh
ddoktorski: Good catch, thanks

cptartur (approved):

scripts/get_scarb_versions.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+set -e
+
+# This script is used to find the following Scarb versions:
+# The current major.minor version along with all its patch versions
+# The latest patch versions for the two versions preceding the current one
+
+function get_all_patch_versions() {
+  # Omit version 2.8.0 as it has a bug when using the `assert_macros` package
+  asdf list all scarb $1 | grep -v "rc" | grep -v "2.8.0"
```
THenry14, line 10:
```suggestion
  asdf list all scarb "$1" | grep -v "rc" | grep -v "2.8.0"
```

Please always quote your variables to prevent globbing and/or splitting 🙏 (applicable to the rest fo the code as well)

scripts/get_scarb_versions.sh
```diff
@@ -0,0 +1,29 @@
...
+
+if [[ ${major_minor_versions[0]} != "2.6" ]]; then
+  scarb_versions+=($(get_latest_patch_version ${major_minor_versions[0]}))
+fi
+
+scarb_versions+=($(get_latest_patch_version ${major_minor_versions[1]}))
+
+scarb_versions+=($(get_all_patch_versions ${major_minor_versions[2]}))
+
+echo \"$(echo ${scarb_versions[@]} | sed 's/ /", "/g')\"
```
THenry14, line 29: This looks a bit wonky, maybe something like:

```
printf '"%s", ' "${scarb_versions[@]}" | sed 's/, $/\n/'
```

Could do?

.github/workflows/scheduled.yml
```diff
@@ -0,0 +1,128 @@
+name: Scheduled
```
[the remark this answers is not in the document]
ddoktorski: Potentially we are considering using `cargo-nextest` for most tests on the CI workflow to speed up the process (currently we use it only for the forge e2e tests). Then the two workflows will be different. We could use `cargo-nextest` here too, but considering speed is not a crucial factor for the scheduled workflow I think it is better to maintain its current state as it is more transparent and seems easier to maintain and extend if necessary.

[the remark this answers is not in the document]
THenry14: Yup, it's not crucial, but if having better speed goes hand in hand with better maintainability, I wouldn't sneeze at that :) Either way, I think this is something that could be looked into as there are no potential downsides I can think of right now apart from the work being put, but as it's a nitpick You can resolve this :+1:

THenry14 (approved):

PR #2663, opened 2024-11-12
Show using `snforge_std` and `sncast_std` from registry

### LINKED ISSUE #2599: Show using `snforge_std` and `sncast_std` from registry in docs
In various places in documentation, we show using `snforge_std` and `sncast_std` from `git`. We should instead show how to use them from package registry.

ddoktorski:
Closes #2599


cptartur (approved):
🙏

PR #2670, opened 2024-11-13
Update test collection docs

### LINKED ISSUE #2465: Update `Test collection` docs

snforge

### Description

https://foundry-rs.github.io/starknet-foundry/testing/test-collection.html

We should describe that tests are collected from Scarb's [test targets](https://docs.swmansion.com/scarb/docs/reference/targets.html#test-targets). We should keep info that is specific to snforge and its plugin like treating functions tagged with `#[test]` as tests, but remove info that is specific to Scarb test targets (e.g. virtual lib.cairo, explanation of autodetected test targets etc.).

We can have an example for a standard tests structure (tests in both `src/` under `#[cfg(test)]` and in `tests/`) as this is unlikely to change

cptartur:
Closes #2465 

## Introduced changes

- Added information that Scarb handles test collection and linked relevant sections in Scarb docs
- Reworked test collection docs

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests

### DIFF
--- a/docs/src/testing/test-collection.md
+++ b/docs/src/testing/test-collection.md
@@ -1,111 +1,109 @@
-## Test Collection
+# How Tests Are Collected
 
-`snforge` considers all functions in your project marked with `#[test]` attribute as tests.
-By default, test functions run without any arguments.
-However, adding any arguments to function signature will enable [fuzz testing](../snforge-advanced-features/fuzz-testing.md) for this
-test case.
+Snforge executes tests however it does not compile them.  
+Instead, to compile tests, snforge internally runs `scarb build --test` command.
 
-`snforge` will collect tests only from these places:
+The `snforge_scarb_plugin` dependency, which is included with `snforge_std` dependency makes all functions
+marked with `#[test]` executable and indicates to Scarb they should be compiled.
+Without the plugin, no snforge tests can be compiled, that's why `snforge_std` dependency is always required in all
+snforge projects.
 
-- any files reachable from the package root (declared as `mod` in `lib.cairo` or its children) - 
-these have to be in a module annotated with `#[cfg(test)]`
-- files inside the [`tests`](#the-tests-directory) directory
+Thanks to that, Scarb collects all functions marked with `#[test]`
+from [valid locations](https://docs.swmansion.com/scarb/docs/extensions/testing.html#tests-organization) and compiles
+them into tests that are executed by snforge.
 
-## The `tests` Directory
+## `[[test]]` Target
 
-`snforge` collects tests from `tests` directory.
-Depending on the presence of `tests/lib.cairo` file, the behavior of the test collector will be different.
+Under the hood, Scarb utilizes the `[[test]]` target mechanism to compile the tests. More information about the
+`[[test]]` target is available in
+the [Scarb documentation](https://docs.swmansion.com/scarb/docs/reference/targets.html#test-targets).
 
-### With `tests/lib.cairo`
+By default, `[[test]]]` target is implicitly configured and user does not have to define it.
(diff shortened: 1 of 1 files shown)


docs/src/testing/test-collection.md
```diff
@@ -1,95 +1,111 @@
...
-│   ├── test_contract.cairo
-│   └── not_included/
-│       └── ignored.cairo
-└── src/
-    └── lib.cairo
+$ tree
+tests/
+├── module1.cairo <-- is collected
+├── module2.cairo <-- is collected
+└── module3.cairo <-- is collected
```
franciszekjob, line 85: Please split this shell snippet, similarly as in [this PR](https://github.com/foundry-rs/starknet-foundry/pull/2658) (just to maintain consistency 😄 ).

Btw, I think here we should make the output expanded by default.

docs/src/testing/test-collection.md
```diff
@@ -1,95 +1,111 @@
...
 
-tests from `tests/test_contract.cairo`, `tests/common.cairo` and `tests/common/utils.cairo` will be collected.
+```shell
+$ tree
+tests/
+├── lib.cairo
+├── module1.cairo  <-- is collected
+├── module2.cairo  <-- is collected
+└── module3.cairo  <-- is not collected
+```
```
franciszekjob, line 104: Same as above.

docs/src/testing/test-collection.md
```diff
@@ -1,95 +1,111 @@
-## Test Collection
+# How Tests Are Collected
 
-`snforge` considers all functions in your project marked with `#[test]` attribute as tests.
-By default, test functions run without any arguments.
-However, adding any arguments to function signature will enable [fuzz testing](../snforge-advanced-features/fuzz-testing.md) for this
-test case.
```
kkawula, line 6: Do we want to get rid of this information, is it written somewhere else?

docs/src/testing/test-collection.md
```diff
@@ -1,95 +1,111 @@
...
-    └── lib.cairo
-```
+You can read more about tests organization
+in [Scarb documentation](https://docs.swmansion.com/scarb/docs/extensions/testing.html#tests-organization).
+
+### Unit Tests
 
-with `tests/lib.cairo` content:
+Test placed in `src` directory are often called unit tests.
+For these test to function in snforge, they must be wrapped in a module marked with `test` attribute.
```
ddoktorski, line 36:
```suggestion
For these test to function in snforge, they must be wrapped in a module marked with `#[cfg(test)]` attribute.
```

docs/src/testing/test-collection.md
```diff
@@ -1,95 +1,111 @@
...
+tests/
+├── module1.cairo <-- is collected
+├── module2.cairo <-- is collected
+└── module3.cairo <-- is collected
 ```
 
-and `tests/common.cairo` content:
+Scarb will collect each of the file and compile it as a
+separate [test target](https://docs.swmansion.com/scarb/docs/reference/targets.html#test-targets).
+Each of this target will be run separately by `snforge`.
```
ddoktorski, line 90:
```suggestion
Scarb will collect each file and compile it as a
separate [test target](https://docs.swmansion.com/scarb/docs/reference/targets.html#test-targets).
Each of these targets will be run separately by `snforge`.
```

docs/src/testing/test-collection.md
cptartur: We have a separate page on fuzz testing where we explain how to enable them. I don't think the page on test collection benefits from having this explained.

ddoktorski (approved):

franciszekjob (approved):

kkawula (approved):

Issue #2468, opened 2024-09-13
Implement printing out latest executed line on failures


Arcticae:
### Which components does the task require to be changed? (think hard pls)

snforge

### Description

This task depends on creation of the aforementioned library that exposes the function used to extract the mapping.

If it fails to parse the mappings we should fall-back to not printing the result

Currently blocked by https://github.com/software-mansion/cairo-coverage/issues/62

PR #2679, opened 2024-11-15
Add backtrace to forge

ksew1:
Closes https://github.com/foundry-rs/starknet-foundry/issues/2468

## Introduced changes

- Forge will display backtrace for contracts in format similar to rust

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


ksew1:
This was inspired by Rust backtrace, but there are a couple of things to note in terms of user experience.

When Forge encounters an error, it suggests enabling a backtrace for more context:

```plaintext
Failure data:
    (0x454E545259504F494E545F4E4F545F464F554E44 ('ENTRYPOINT_NOT_FOUND'), 0x454E545259504F494E545F4641494C4544 ('ENTRYPOINT_FAILED'))

note: run with `SNFORGE_BACKTRACE=1` environment variable to display a backtrace
```

Enabling backtrace provides a call stack:

```plaintext
Failure data:
    (0x454E545259504F494E545F4E4F545F464F554E44 ('ENTRYPOINT_NOT_FOUND'), 0x454E545259504F494E545F4641494C4544 ('ENTRYPOINT_FAILED'))

error occurred in contract 'InnerContract' at pc: '72'
possible stack backtrace:
   0: InnerContract::inner_call
       at src/lib.cairo:47:9
   1: InnerContract::inner
       at src/lib.cairo:38:13
   2: InnerContract::__wrapper__inner
       at src/lib.cairo:37:9

[...]

Tests: 0 passed, 1 failed, 0 skipped
```

### However ...

- If functions are not inlined, only the line causing the failure is shown, without the functions that call them:

  ```plaintext
  0: InnerContract::inner_call
       at src/lib.cairo:47:9
  ```
  
  The Cairo compiler inlines aggressively, so often you will get the whole call stack from the entry point, which is really helpful when debugging. But in the worst-case scenario, you only get the last line.  I was thinking because of that should we always display the last line, but I believe the whole call stack is really helpful.

- Currently, the VM does not halt immediately upon encountering a panic and doesn't bubble it up, so backtraces may display only the contract's entry point.

In terms of UX, if you prefer something different, we can easily change this as the backend will probably stay the same.

ksew1:
Also, I will add documentation in a separate PR to avoid making this one any bigger.

crates/forge-runner/src/backtrace.rs
```diff
@@ -0,0 +1,224 @@
...
+            contract_name: &self.contract_name,
+            origins: code_locations
+                .iter()
+                .zip(function_names)
+                .map(|(code_location, function_name)| Backtrace {
+                    code_location,
+                    function_name,
+                })
+                .collect(),
+        })
```
franciszekjob, line 151: nit:
```suggestion
        let origins = code_locations
            .iter()
            .zip(function_names)
            .map(|(code_location, function_name)| Backtrace {
                code_location,
                function_name,
            })
            .collect();

        Some(BacktraceStack {
            pc,
            contract_name: &self.contract_name,
            origins,
        })
```

crates/forge/tests/data/backtrace_panic/Scarb.toml
```diff
@@ -0,0 +1,22 @@
...
+
+[[target.starknet-contract]]
+sierra = true
+
+[scripts]
+test = "snforge test"
+
+[profile.dev.cairo]
+unstable-add-statements-functions-debug-info = true
+unstable-add-statements-code-locations-debug-info = true
```
franciszekjob, line 22: Newline 😅

crates/forge/tests/data/backtrace_vm_error/Scarb.toml
```diff
@@ -0,0 +1,22 @@
...
+
+[[target.starknet-contract]]
+sierra = true
+
+[scripts]
+test = "snforge test"
+
+[profile.dev.cairo]
+unstable-add-statements-functions-debug-info = true
+unstable-add-statements-code-locations-debug-info = true
```
franciszekjob, line 22: Same as above

crates/forge/tests/e2e/backtrace.rs
```diff
@@ -0,0 +1,129 @@
...
+                   at [..]/src/lib.cairo:17:13
+               2: backtrace_vm_error::OuterContract::__wrapper__OuterContract__outer
+                   at [..]/src/lib.cairo:15:9"
+        },
+    );
+}
+
+#[test]
+#[cfg_attr(not(feature = "scarb_2_8_3"), ignore)]
+fn test_wrong_configuration() {
```
franciszekjob, line 63:
```suggestion
fn test_wrong_scarb_toml_configuration() {
```

crates/forge-runner/src/backtrace.rs
```diff
@@ -0,0 +1,224 @@
...
+use itertools::Itertools;
+use rayon::prelude::*;
+use starknet_api::core::ClassHash;
+use std::collections::HashMap;
+use std::fmt::Display;
+use std::{env, fmt};
+
+const BACKTRACE_ENV: &str = "SNFORGE_BACKTRACE";
+
+pub fn add_back_trace_footer(
```
franciszekjob, line 24: nit: Should be applied to other usages as well
```suggestion
pub fn add_backtrace_footer(
```

franciszekjob (changes requested):

crates/forge-runner/src/backtrace.rs
```diff
@@ -0,0 +1,226 @@
...
+            })?;
+
+        let VersionedProfilerAnnotations::V1(profiler_annotations) =
+            VersionedProfilerAnnotations::try_from_debug_info(sierra_debug_info).context(indoc! {
+                "perhaps the contract was compiled without the following entry in Scarb.toml under [profile.dev.cairo]:
+                unstable-add-statements-functions-debug-info = true
+
+                or scarb version is less than 2.8.0
+                "
+            })?;
```
cptartur, line 100: Should we always try to add these annotations?

crates/forge-runner/src/backtrace.rs
```diff
@@ -0,0 +1,226 @@
...
+
+        // Not optimal, but USC doesn't produce debug info for the contract class
```
cptartur, line 102: Is this resolvable in the USC? Perhaps a follow-up task should be created

crates/forge-runner/src/backtrace.rs
```diff
@@ -0,0 +1,226 @@
...
+        let (_, debug_info) = CasmContractClass::from_contract_class_with_debug_info(
```
cptartur, line 103: This compiles to casm right? 
We should probably clearly state somewhere that running with backtrace is quite expensive.

crates/forge-runner/src/backtrace.rs
```diff
@@ -0,0 +1,226 @@
...
+            pc,
+            contract_name: &self.contract_name,
+            stack,
+        })
+    }
+}
+
+struct BacktraceContractFormatter(HashMap<ClassHash, ContractBacktraceData>);
+
+impl BacktraceContractFormatter {
```
cptartur, line 159: Not sure if Formatter is a correct name for this.

crates/forge-runner/src/backtrace.rs
```diff
@@ -0,0 +1,226 @@
...
+}
+
+impl Display for BacktraceStack<'_> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        writeln!(
+            f,
+            "error occurred in contract '{}' at pc: '{}'",
+            self.contract_name, self.pc
+        )?;
+        writeln!(f, "possible stack backtrace:")?;
```
cptartur, line 213: Why is it called "possible"?

crates/forge-runner/src/backtrace.rs
```diff
@@ -0,0 +1,226 @@
...
+        for (i, backtrace) in self.stack.iter().enumerate() {
+            writeln!(f, "   {i}: {backtrace}")?;
+        }
+        Ok(())
+    }
+}
+
+fn truncate_at_char(input: &str, delimiter: char) -> &str {
```
cptartur, line 221: It's only used in one place, maybe we can inline this?

cptartur: By the way, cannot one of the `trim_` functions of `str` be used instead?

crates/forge-runner/src/test_case_summary.rs
```diff
@@ -216,11 +217,13 @@ impl TestCaseSummary<Single> {
...
         used_resources: UsedResources,
         call_trace: &Rc<RefCell<InternalCallTrace>>,
+        encountered_errors: &[EncounteredError],
         contracts_data: &ContractsData,
         maybe_versioned_program_path: &Option<VersionedProgramPath>,
     ) -> Self {
         let name = test_case.name.clone();
-        let msg = extract_result_data(&run_result, &test_case.config.expected_result);
+        let msg = extract_result_data(&run_result, &test_case.config.expected_result)
+            .map(|msg| add_backtrace_footer(msg, contracts_data, encountered_errors));
```
cptartur, line 226: Why are we calling `add_backtrace_footer` both in forge and forge-runner?

cptartur: Okay, never mind, it's both in forge-runner and I didn't realize it was two different branches.

crates/forge-runner/src/backtrace.rs
cptartur: Worth mentioning in the docs that this is needed as well. And let's try communicate with Scarb team if this is somehow implicitly resolvable.

cptartur (approved):

franciszekjob (approved):
Good job 👌

crates/forge-runner/src/backtrace.rs
```diff
@@ -0,0 +1,226 @@
...
+            .sierra_program_debug_info
+            .as_ref()
+            .context("debug info not found")?;
+
+        let VersionedCoverageAnnotations::V1(coverage_annotations) =
+            VersionedCoverageAnnotations::try_from_debug_info(sierra_debug_info).context(indoc! {
+                "perhaps the contract was compiled without the following entry in Scarb.toml under [profile.dev.cairo]:
+                unstable-add-statements-code-locations-debug-info = true
+
+                or scarb version is less than 2.8.0
```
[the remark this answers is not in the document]
ksew1: A similar implementation has already been done for [coverage](https://github.com/foundry-rs/starknet-foundry/blob/5bbfa804a1d7aac97a9df9d35b1fde3f2c08cc6d/crates/forge-runner/src/coverage_api.rs#L70). 

However, to achieve this, we need metadata for the operation. I haven’t implemented this because it would either require:
- Propagating the metadata throughout the entire codebase up until this point, or
- Implementing it at the start of the `snforge` process, which is not optimal as you could run the backtrace feature without encountering any error. In such cases, the annotations wouldn’t even be needed, yet the process would still fail.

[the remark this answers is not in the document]
cptartur: Please create a task for this and we can resolve the details there.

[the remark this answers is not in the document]
ksew1: https://github.com/foundry-rs/starknet-foundry/issues/2710

Issue #2650, opened 2024-11-07
Show using `.serialize` in Cairo in documentation


cptartur:
### Which components does the task require to be changed? (think hard pls)

snforge

### Description

Right now, we do not show how to use `.serialize` in Cairo in our documentation, even though it is required to prepare calldata for `deploy`, unless one wants to manually serialize the calldata

It is currently only shown in unrelated page "Direct Storage Access" and not explained on any page.

Documentation of testing smart contracts should be updated to indicate that it is possible to use `.serialize` in Cairo. We can also consider linking a relevant chapter in Cairo Book.

Supa-mega:
May I take care of this?

cptartur:
Let's add a proper utility for this in the future

PR #2682, opened 2024-11-18
Add data transformer examples to docs

### LINKED ISSUE #2557: Update `sncast` docs to use data transformer instead of serialized calldata as a default exampl

sncast

### Description

Once data transformer has proven itself to be working correctly we could update the documentation so default way of passing arguments is shown as data transformation. 

Current way of passing already serialized calldata could be documented as an option.

kkawula:
Closes #2557

## Introduced changes

-

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


docs/src/starknet/index.md
```diff
@@ -51,14 +50,16 @@ response: [0x0]
 > In the above example we supply `sncast` with `--account` and `--url` flags. If `snfoundry.toml` is present, and have these properties set, values provided using these flags will override values from `snfoundry.toml`. Learn more about `snfoundry.toml` configuration [here](../projects/configuration.md#sncast).
 
 
-### Calldata
+### Arguments and Calldata
```
cptartur, line 53: How about we change this section to `### Arguments` and explain only how to use `--arguments` flag.

Below, we can add a section `### Using Serialized Calldata` and keep the docs for `--calldata`.

docs/src/starknet/index.md
```diff
@@ -51,14 +50,16 @@ response: [0x0]
...
 
 Some `sncast` commands (namely `call`, `deploy` and `invoke`) allow passing *calldata* - a series of arguments to perform an action with on blockchain.
 
-In the example above we called a function with an argument: `0x0`, passed using `--calldata` flag.
+In the example above we called a function with an argument: `'"Charizard"'`, passed using `--arguments` flag, but it can
+be also obtained by passing a serialized form, in this example equivalent to passing `--calldata` with the following value: `0x0 0x43686172697a617264 0x9`.
 
 Please note the notation of the argument. The default way of passing calldata is a list of hexadecimally encoded field elements - the *serialized* calldata.
 To obtain the serialized form of the wished data, one must write a Cairo program calling `Serde::serialize` on subsequent arguments and displaying the results.
```
cptartur, line 61: This is outdated, serialized is no longer default - there is no default in code at all.

cptartur (changes requested):

docs/src/starknet/index.md
```diff
@@ -51,21 +50,29 @@ response: [0x0]
...
+
+### Using Serialized Calldata
+
+Equivalent to the previous example, the same result can be obtained by passing serialized calldata - a list of hexadecimal encoded field elements.
+
+In this example equivalent to passing `--calldata` with the following value: `0x0 0x43686172697a617264 0x9`.
+
+To obtain the serialized form of the wished data, one must write a Cairo program calling `Serde::serialize` on subsequent arguments and displaying the results.
+Read more about it in the [Cairo documentation](https://book.cairo-lang.org/appendix-03-derivable-traits.html?highlight=seri#serializing-with-serde).
+
```
franciszekjob, line 75:
```suggestion

### Using Serialized Calldata

The same result can be achieved by passing serialized calldata, which is a list of hexadecimal-encoded field elements.

For example, this is equivalent to using the --calldata option with the following value: 0x0 0x43686172697a617264  0x9.

To obtain the serialized form of the wished data, you can write a Cairo program that calls `Serde::serialize` on subsequent arguments and displays the results.

Read more about it in the [Cairo documentation](https://book.cairo-lang.org/appendix-03-derivable-traits.html?highlight=seri#serializing-with-serde).

```

franciszekjob (commented):
These examples (snippets + outputs) rely on example contract(s) present in our code, right?

docs/src/starknet/index.md
```diff
@@ -51,21 +50,30 @@ response: [0x0]
...
 > ⚠️ **Warning**
 > Cast will not verify the serialized calldata. Any errors caused by passing improper calldata in a serialized form will originate from the network.
 > Basic static analysis is possible only when passing expressions - see [calldata transformation](./calldata-transformation.md).
 
+
+### Using Serialized Calldata
+
+The same result can be achieved by passing serialized calldata, which is a list of hexadecimal-encoded field elements.
+
+For example, this is equivalent to using the --calldata option with the following value: 0x0 0x43686172697a617264  0x9.
```
cptartur, line 71:
```suggestion
For example, this is equivalent to using the --calldata option with the following value: 0x0 0x43686172697a617264 0x9.
```

docs/src/starknet/index.md
```diff
@@ -51,21 +50,30 @@ response: [0x0]
...
+
+To obtain the serialized form of the wished data, you can write a Cairo program that calls `Serde::serialize` on subsequent arguments and displays the results.
```
cptartur, line 73: Kind of resolves https://github.com/foundry-rs/starknet-foundry/issues/2650 but I think we might want to expand this in a separate PR anyway.

cptartur (approved):

PR #2689, opened 2024-11-19
Neque porro quisquam est qui dolorem

kkawula:
Closes #

## Introduced changes

-

## Checklist

- [ ] Linked relevant issue
- [ ] Added relevant tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - You can skip `--name` flag when using `account import` - a default name will be generated.
 - Addresses outputted when calling `sncast account create`, `sncast deploy` and `sncast declare` are now padded to 64 characters length and prefixed with `0x0`
+- Default global cast configuration.
 
 #### Changed
 
--- a/crates/forge/src/init.rs
+++ b/crates/forge/src/init.rs
@@ -20,15 +20,17 @@ fn create_snfoundry_manifest(path: &PathBuf) -> Result<()> {
     fs::write(
         path,
         formatdoc! {r#"
-        # Visit https://foundry-rs.github.io/starknet-foundry/appendix/snfoundry-toml.html for more information
+        # Visit https://foundry-rs.github.io/starknet-foundry/appendix/snfoundry-toml.html
+        # and https://foundry-rs.github.io/starknet-foundry/projects/configuration.html for more information
 
-        # [sncast.myprofile1]                                    # Define a profile name
-        # url = "http://127.0.0.1:5050/"                         # Url of the RPC provider
-        # accounts_file = "../account-file"                      # Path to the file with the account data
+        # [sncast.default]                                       # Define a profile name
+        # url = "https://starknet-sepolia.public.blastapi.io"    # Url of the RPC provider
+        # accounts-file = "../account-file"                      # Path to the file with the account data
         # account = "mainuser"                                   # Account from `accounts_file` or default account file that will be used for the transactions
         # keystore = "~/keystore"                                # Path to the keystore file
-        # wait_params = {{ timeout = 500, retry_interval = 10 }}   # Wait for submitted transaction parameters
-        # block_explorer = "StarkScan"                           # Block explorer service used to display links to transaction details
+        # wait-params = {{ timeout = 500, retry-interval = 10 }}   # Wait for submitted transaction parameters
+        # block-explorer = "StarkScan"                           # Block explorer service used to display links to transaction details
+        # show-explorer-links = true                             # Print links pointing to pages with transaction details in the chosen block explorer
         "#
         },
     )?;
--- a/crates/sncast/src/helpers/global_config.rs
+++ b/crates/sncast/src/helpers/global_config.rs
@@ -1,0 +1,57 @@
+use anyhow::Result;
+use camino::Utf8PathBuf;
+use indoc::formatdoc;
+use std::fs;
+use std::fs::File;
+use std::io::Write;
+
+pub fn get_global_config_path() -> Result<Utf8PathBuf> {
+    let config_dir = if cfg!(target_os = "windows") {
+        dirs::config_dir()
+    } else {
+        dirs::home_dir()
+    }
+    .ok_or_else(|| anyhow::anyhow!("Could not determine config directory"))?;
+
+    let global_config_dir = config_dir.join(if cfg!(target_os = "windows") {
+        "starknet-foundry"
+    } else {
+        ".config/starknet-foundry"
+    });
+
+    if !global_config_dir.exists() {
+        fs::create_dir_all(&global_config_dir)?;
+    }
+
+    let global_config_path =
+        Utf8PathBuf::from_path_buf(global_config_dir.join("snfoundry.toml")).unwrap();
+
+    if global_config_path.exists() {
+        Ok(global_config_path)
+    } else {
+        create_global_config(global_config_path.clone())?;
+        Ok(global_config_path)
+    }
+}
+
+fn create_global_config(global_config_path: Utf8PathBuf) -> Result<()> {
+    let mut file = File::create(global_config_path)?;
+    file.write_all(
+        formatdoc! {r#"
+        # Visit https://foundry-rs.github.io/starknet-foundry/appendix/snfoundry-toml.html
+        # and https://foundry-rs.github.io/starknet-foundry/projects/configuration.html for more information
+
+        [sncast.default]
+        url = "https://starknet-sepolia.public.blastapi.io"
+        # accounts-file = "~/.starknet_accounts/starknet_open_zeppelin_accounts.json"
+        # wait-params = {{ timeout = 500, retry-interval = 10 }}
+        # block-explorer = "StarkScan"
+        # show-explorer-links = true
+        # account = "default"
+        # keystore = ""
+        "#
+        }.as_bytes(),
+    )?;
+
+    Ok(())
+}
--- a/crates/sncast/src/main.rs
+++ b/crates/sncast/src/main.rs
@@ -16,6 +16,7 @@ use clap::{Parser, Subcommand};
 use sncast::helpers::configuration::CastConfig;
 use sncast::helpers::constants::{DEFAULT_ACCOUNTS_FILE, DEFAULT_MULTICALL_CONTENTS};
 use sncast::helpers::fee::PayableTransaction;
+use sncast::helpers::global_config::get_global_config_path;
 use sncast::helpers::scarb_utils::{
     assert_manifest_path_exists, build, build_and_load_artifacts, get_package_metadata,
     get_scarb_metadata_with_deps, BuildConfig,
@@ -209,8 +210,7 @@ fn main() -> Result<()> {
     if let Commands::Script(script) = &cli.command {
         run_script_command(&cli, runtime, script, numbers_format, output_format)
     } else {
-        let mut config = load_global_config::<CastConfig>(&None, &cli.profile)?;
-        update_cast_config(&mut config, &cli);
+        let config = get_cast_config(&cli)?;
 
         runtime.block_on(run_async_command(
             cli,
@@ -657,11 +657,8 @@ fn run_script_command(
             let manifest_path = assert_manifest_path_exists()?;
             let package_metadata = get_package_metadata(&manifest_path, &run.package)?;
 
-            let mut config = load_global_config::<CastConfig>(
-                &Some(package_metadata.root.clone()),
-                &cli.profile,
-            )?;
-            update_cast_config(&mut config, cli);
+            let config = get_cast_config(cli)?;
+
             let provider = runtime.block_on(run.rpc.get_provider(&config))?;
 
             let mut artifacts = build_and_load_artifacts(
@@ -739,4 +736,44 @@ fn update_cast_config(config: &mut CastConfig, cli: &Cli) {
         ),
         clone_or_else!(cli.wait_timeout, config.wait_params.get_timeout()),
     );
+}
+
+fn get_cast_config(cli: &Cli) -> Result<CastConfig> {
+    let global_config_path = get_global_config_path().unwrap_or(Utf8PathBuf::new());
+    let mut global_config =
+        load_global_config::<CastConfig>(&Some(global_config_path.clone()), &cli.profile)
+            .unwrap_or_else(|_| {
+                load_global_config::<CastConfig>(&Some(global_config_path), &None).unwrap()
+            });
+
+    let config = load_global_config::<CastConfig>(&None, &cli.profile)?;
+
+    combine_cast_configs(&mut global_config, &config);
+
+    update_cast_config(&mut global_config, &cli);
+    Ok(global_config)
+}
+
+fn combine_cast_configs(global_config: &mut CastConfig, config: &CastConfig) {
+    if config.url != String::default() {
+        global_config.url = config.url.clone();
+    }
+    if config.account != String::default() {
+        global_config.account = config.account.clone();
+    }
+    if config.accounts_file != Utf8PathBuf::default() {
+        global_config.accounts_file = config.accounts_file.clone();
+    }
+    if config.keystore != Option::default() {
+        global_config.keystore = config.keystore.clone();
+    }
+    if config.wait_params != ValidatedWaitParams::default() {
+        global_config.wait_params = config.wait_params.clone();
+    }
+    if config.block_explorer != Option::default() {
+        global_config.block_explorer = config.block_explorer.clone();
+    }
+    if config.show_explorer_links != true {
+        global_config.show_explorer_links = config.show_explorer_links;
+    }
 }
--- a/docs/src/projects/configuration.md
+++ b/docs/src/projects/configuration.md
@@ -105,6 +105,29 @@ response: [0x1, 0x23, 0x4]
 </details>
 <br>
 
+### Global configuration
+
+Since version 0.34.0 `sncast` will use the global configuration to combine configuration from local configs.
+Global configuration file will be automatically generated if it does not already exist, and will contain predefined `default` profile.
+
+#### Global Configuration File Location
+The global configuration is stored in a specific location depending on the operating system:
+
+- macOS/Linux : The global configuration file is located at `$HOME/.config/starknet-foundry/snfoundry.toml`
+- Windows : The file can be found at `C:\Users\<user>\AppData\Roaming\starknet-foundry\snfoundry.toml`
+
+#### Interaction between Local and Global profiles
+The interaction between local and global configurations follow to a clear set of overriding rules:
+
+By specifying `--profile` argument, you choose profile to set 
+
+- Local Configuration Priority: If there is a profile with the same name in both the local and global configurations, the settings from the local configuration will override those in the global configuration.
+
+- Global Default: In the absence of a local profile, the configuration will fall back to and override the default settings in the global profile, maintaining consistency across projects where explicit local configurations are not specified.
+Contents of the Global Configuration
+
+The global configuration file typically contains default profiles and shared settings that apply across all projects.
+
 ## Environmental variables
 
 Programmers can use environmental variables in both `Scarb.toml::tool::snforge` and in `snfoundry.toml`. To use an environmental variable as a value, use its name prefixed with `$`. 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4897,6 +4897,7 @@ dependencies = [
  "conversions",
  "ctor",
  "data-transformer",
+ "dirs",
  "fs_extra",
  "indoc",
  "itertools 0.12.1",
--- a/crates/sncast/Cargo.toml
+++ b/crates/sncast/Cargo.toml
@@ -56,6 +56,7 @@ serde_path_to_error.workspace = true
 walkdir.workspace = true
 const-hex.workspace = true
 regex.workspace = true
+dirs = "5.0.1"
 
 [dev-dependencies]
 ctor.workspace = true
--- a/crates/sncast/src/helpers/mod.rs
+++ b/crates/sncast/src/helpers/mod.rs
@@ -4,5 +4,6 @@ pub mod configuration;
 pub mod constants;
 pub mod error;
 pub mod fee;
+pub mod global_config;
 pub mod rpc;
 pub mod scarb_utils;
--- a/docs/src/appendix/snfoundry-toml.md
+++ b/docs/src/appendix/snfoundry-toml.md
@@ -25,19 +25,19 @@ The `url` field specifies the address of RPC provider.
 url = "http://example.com"
 ```
 
-#### `accounts_file`
+#### `accounts-file`
 
-The `accounts_file` field specifies the path to a file containing account information. 
+The `accounts-file` field specifies the path to a file containing account information. 
 If not provided, the default path is `~/.starknet_accounts/starknet_open_zeppelin_accounts.json`.
 
 ```toml
 [sncast.myprofile]
-accounts_file = "path/to/accounts.json"
+accounts-file = "path/to/accounts.json"
 ```
 
 #### `account`
 
-The `account` field specifies which account from the `accounts_file` to use for transactions.
+The `account` field specifies which account from the `accounts-file` to use for transactions.
 
 ```toml
 [sncast.myprofile]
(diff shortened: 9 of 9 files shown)


crates/sncast/src/main.rs
```diff
@@ -740,3 +737,49 @@ fn update_cast_config(config: &mut CastConfig, cli: &Cli) {
...
+
+    let config = load_global_config::<CastConfig>(&None, &cli.profile)?;
+
+    combine_cast_configs(&mut global_config, &config);
+
+    update_cast_config(&mut global_config, cli);
+    Ok(global_config)
+}
+
+fn combine_cast_configs(global_config: &mut CastConfig, config: &CastConfig) {
```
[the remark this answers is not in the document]
cptartur: With the current setup, if we added a new key to `CastConfig` we'd have to remember update this function. I'd rather have the compiler check this for us.

docs/src/projects/configuration.md
```diff
@@ -105,6 +105,28 @@ response: [0x1, 0x23, 0x4]
...
+> 📝 **Note**
+> Global configuration file will be created automatically if it does not exist and contain predefined `default` profile.
+
+
+#### Interaction between Local and Global profiles
+
+Interaction between global and local config is based on the overriding mechanism, the local config will be used to override 
+the global profile of the same name or in case it is not defined to override the default global profile.
+
+If `--profile` is not provided, the default profile from the local (if present) and global configuration will be used.
```
cptartur, line 128: This needs to be written in simpler terms, so it's obvious what overrides what and in which cases.

CHANGELOG.md
```diff
@@ -24,6 +24,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - You can skip `--name` flag when using `account import` - a default name will be generated.
 - Addresses outputted when calling `sncast account create`, `sncast deploy` and `sncast declare` are now padded to 64 characters length and prefixed with `0x0`
+- Globally available configuration to store profiles to share between projects.
+- Missing fields in `show-config` command.
```
ddoktorski, line 28: nit: Could you list the added fields here?

crates/shared/src/consts.rs
```diff
@@ -1,3 +1,4 @@
 pub const EXPECTED_RPC_VERSION: &str = "0.7.0";
 pub const RPC_URL_VERSION: &str = "v0_7";
 pub const SNFORGE_TEST_FILTER: &str = "SNFORGE_TEST_FILTER";
+pub const FREE_RPC_PROVIDER_URL: &str = "https://free-rpc.nethermind.io/sepolia-juno/v0_7";
```
ddoktorski, line 4: nit: Do we really want to hardcode a link to an external node in the template? This might require updates in the future. I suggest sticking with the previous devnet address `http://127.0.0.1:5050`.

crates/sncast/src/helpers/global_config.rs
```diff
@@ -0,0 +1,72 @@
...
+    } else {
+        ".config/starknet-foundry"
+    });
+
+    if !global_config_dir.exists() {
+        fs::create_dir_all(&global_config_dir)?;
+    }
+
+    let global_config_path =
+        Utf8PathBuf::from_path_buf(global_config_dir.join("snfoundry.toml")).unwrap();
```
ddoktorski, line 30: nit: Better to use `expect` in such cases

crates/sncast/src/helpers/global_config.rs
```diff
@@ -0,0 +1,72 @@
...
+
+    formatdoc! {r#"
+        # Visit https://foundry-rs.github.io/starknet-foundry/appendix/snfoundry-toml.html
+        # and https://foundry-rs.github.io/starknet-foundry/projects/configuration.html for more information
+
+        [sncast.default]
+        url = "{default_url}"
+        block-explorer = "{default_block_explorer}"
+        wait-params = {{ timeout = {default_wait_timeout}, retry-interval = {default_wait_retry_interval} }}
+        show-explorer-links = {default_show_explorer_links}
```
ddoktorski, line 51: Why this is not commented out?

crates/sncast/src/main.rs
```diff
@@ -740,3 +737,49 @@ fn update_cast_config(config: &mut CastConfig, cli: &Cli) {
...
+    let global_config_path = get_global_config_path().unwrap_or_else(|err| {
+        eprintln!("Error getting global config path: {err}");
+        Utf8PathBuf::new()
+    });
+
+    let mut global_config =
+        load_global_config::<CastConfig>(&Some(global_config_path.clone()), &cli.profile)
+            .unwrap_or_else(|_| {
+                load_global_config::<CastConfig>(&Some(global_config_path), &None).unwrap()
+            });
```
ddoktorski, line 751: Do we need this part? If `profile` is not provided it should be `None`.

crates/sncast/src/main.rs
```diff
@@ -740,3 +737,49 @@ fn update_cast_config(config: &mut CastConfig, cli: &Cli) {
         clone_or_else!(cli.wait_timeout, config.wait_params.get_timeout()),
     );
 }
+
+fn get_cast_config(cli: &Cli) -> Result<CastConfig> {
```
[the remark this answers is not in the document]
kkawula: The structure is called `CastConfig` so I suppose that it will be more accurate as it is now.

crates/sncast/src/helpers/global_config.rs
```diff
@@ -0,0 +1,72 @@
...
+
+    let global_config_dir = config_dir.join(if cfg!(target_os = "windows") {
+        "starknet-foundry"
+    } else {
+        ".config/starknet-foundry"
+    });
+
+    if !global_config_dir.exists() {
+        fs::create_dir_all(&global_config_dir)?;
+    }
```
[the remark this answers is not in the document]
kkawula: Why not? I suppose that `.config` dir idea is intended to be used by tools, and we should create it if is not present

crates/shared/src/consts.rs
kkawula: I suppose that it will be convenient for the user to have ready sepolia URL defined.

crates/sncast/src/helpers/configuration.rs
```diff
@@ -57,7 +58,7 @@ impl Default for CastConfig {
             keystore: None,
             wait_params: ValidatedWaitParams::default(),
             block_explorer: Some(block_explorer::Service::default()),
-            show_explorer_links: true,
+            show_explorer_links: show_explorer_links_default(),
```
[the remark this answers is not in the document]
kkawula: Hmmmmmm, we are using it in a few places so maybe we should use const as it is now

crates/sncast/tests/e2e/main_tests.rs
```diff
@@ -169,29 +169,6 @@ async fn test_missing_account_flag() {
...
-
-    let snapbox = runner(&args);
-    let output = snapbox.assert().failure();
-
-    assert_stderr_contains(
-        output,
-        "Error: RPC url not passed nor found in snfoundry.toml",
-    );
-}
-
```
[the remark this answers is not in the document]
kkawula: We provide it by default from now, so it isn't possible

crates/sncast/src/main.rs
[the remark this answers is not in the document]
kkawula: It isn't possible at all, because we can export `Cli` to different module/file, refactor should be addressed in other issue

crates/sncast/src/main.rs
```diff
@@ -740,3 +737,49 @@ fn update_cast_config(config: &mut CastConfig, cli: &Cli) {
...
+    let global_config_path = get_global_config_path().unwrap_or_else(|err| {
+        eprintln!("Error getting global config path: {err}");
+        Utf8PathBuf::new()
+    });
```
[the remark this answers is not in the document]
kkawula: discussed offline

docs/src/projects/configuration.md
```diff
@@ -105,6 +105,28 @@ response: [0x1, 0x23, 0x4]
 </details>
 <br>
 
+### Global configuration
```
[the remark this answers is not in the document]
kkawula: It seems to me that the composition as it is now is quite natural. Check the changes. I hope it will fit to you

crates/sncast/src/helpers/global_config.rs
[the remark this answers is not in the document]
cptartur: Seems bit weird to me to create those. 

But cc @THenry14 to verify

crates/sncast/src/main.rs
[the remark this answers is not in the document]
cptartur: Let's create an issue to refactor that then. `main.rs` is to big at this stage.

crates/sncast/src/helpers/config.rs
```diff
@@ -0,0 +1,114 @@
...
+}
+macro_rules! clone_field {
+    ($global_config:expr, $local_config:expr, $default_config:expr, $field:ident) => {
+        if $local_config.$field != $default_config.$field {
+            $local_config.$field.clone()
+        } else {
+            $global_config.$field.clone()
+        }
+    };
+}
```
cptartur, line 79: I like having new macros, however in this case I think it just could be a function :ccc

crates/shared/src/consts.rs
ddoktorski: Although it might be convenient for some users I still don't think we should include it in the template. If you add it here you are kind of responsible for it to work, which means extra effort to maintain it and potentially dealing with related support questions.

crates/sncast/src/main.rs
```diff
@@ -740,3 +735,23 @@ fn update_cast_config(config: &mut CastConfig, cli: &Cli) {
...
+        load_global_config::<CastConfig>(&Some(global_config_path.clone()), &cli.profile)
+            .unwrap_or_else(|_| {
+                load_global_config::<CastConfig>(&Some(global_config_path), &None).unwrap()
+            });
+
+    let local_config = load_global_config::<CastConfig>(&None, &cli.profile)?;
+
+    let mut combined_config = combine_cast_configs(&global_config, &local_config);
+
+    update_cast_config(&mut combined_config, cli);
```
cptartur, line 755: Can we rename this function? I know we had it before, but it's not a good name in my opinion for what it does.

docs/src/projects/configuration.md
```diff
@@ -105,6 +105,56 @@ response: [0x1, 0x23, 0x4]
...
+- **A:** Global configuration file containing the profiles `default` and `testnet`.
+- **B:** Local configuration file containing the profiles `default` and `mainnet`.
+
+In any directory in the file system, a user can run the `sncast` command using the `default` and `testnet` profiles, 
+because they are located in global config. 
+If no profiles are explicitly specified, the `default` profile from the global configuration file will be used.
+
+When running `sncast` from the `opus magnum` directory, there is a configuration file in the parent directory (B). 
+This setup allows for the use of the following profiles: `default`, `testnet`, and `mainnet`. If the `mainnet` profile is specified, 
+the configuration from the local file will be used to override the global `default` profile, as the `mainnet` profile does not exist in the global configuration.
```
cptartur, line 156: It's a nice example but maybe show a shortened contents of A and B instead?

docs/src/projects/configuration.md
```diff
@@ -105,6 +105,56 @@ response: [0x1, 0x23, 0x4]
 </details>
 <br>
 
+### Global configuration
+
+Global configuration file is usual [`snfoundry.toml`](https://foundry-rs.github.io/starknet-foundry/appendix/snfoundry-toml.html), 
+which is a common storage for configurations to apply to multiple projects across various directories.
+This file is stored in a predefined location and is managed differently compared to the local configuration file, which can be found in any parent directory.
+Profiles from global configurations are used to combine configuration with local profiles.
```
cptartur, line 113:
```suggestion
Global configuration file is usual [`snfoundry.toml`](https://foundry-rs.github.io/starknet-foundry/appendix/snfoundry-toml.html), 
which is a common storage for configurations to apply to multiple projects across various directories.
This file is stored in a predefined location and is used to store profiles that can be used from any location on your computer.
```

docs/src/projects/configuration.md
```diff
@@ -105,6 +105,56 @@ response: [0x1, 0x23, 0x4]
...
+
+#### Interaction between Local and Global profiles
+
+Interaction between global and local config is based on the overriding mechanism
```
cptartur, line 117:
```suggestion
Global config can be overridden by a local config.
```

docs/src/projects/configuration.md
```diff
@@ -105,6 +105,56 @@ response: [0x1, 0x23, 0x4]
...
+
+The local configuration will be used to override a global profile of the same name, or if no matching profile exists, it will override the global default profile. 
+If `--profile` is not explicitly specified, the default profiles from both the local and global configurations (if any is present) will be used.
```
cptartur, line 120:
```suggestion
If both local and global profiles with the same name are present, local profile will be combined with global profile. That is for any setting defined both in global and local profiles, local setting will be used. For settings not defined in local profile, value from the global profile will be used instead.

This same behavior applies for default profiles as well. A local default profile will override a global default profile.
```

cptartur: And link to a section above when talking about "default profiles".

cptartur (changes requested):

CHANGELOG.md
```diff
@@ -24,6 +24,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - You can skip `--name` flag when using `account import` - a default name will be generated.
 - Addresses outputted when calling `sncast account create`, `sncast deploy` and `sncast declare` are now padded to 64 characters length and prefixed with `0x0`
+- Globally available configuration to store profiles to share between projects.
+- Missing fields in `show-config` command, `block_explorer` and `show_explorer_links`.
+- Default free sepolia testnet RPC URL for any `sncast` command.
```
ddoktorski, line 29: Even if we decide to add it, which I am strongly against, what is the point of mentioning this in the changelog?

ddoktorski (commented):
In the commands `account add` and `account import` there is `--add-profile` flag. After introducing the global config it would be nice to explicitly mention that the created profile will be added to a local config if exists, or to change this logic. This can be done in this PR or a separate one, but it should be completed before the next release.

CHANGELOG.md
kkawula: I suppose that it is a significant change. It changes the expected behavior from the older version. Isn't it proper to mention here?

crates/sncast/src/main.rs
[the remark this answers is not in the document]
kkawula: #2700

CHANGELOG.md
ddoktorski: I see now that this also applies to the default `CastConfig`. I thought it was only for the template. That makes it even worse, but in such a case it should be added to the changelog.

docs/src/projects/configuration.md
kkawula: I've added links to proper sections, profile examples are defined above this section. I hope it will help avoid being redundant and also allow users to grab context. wdyt?

crates/sncast/src/helpers/config.rs
```diff
@@ -0,0 +1,114 @@
...
+        # url = "{default_url}"
+        # block-explorer = "{default_block_explorer}"
+        # wait-params = {{ timeout = {default_wait_timeout}, retry-interval = {default_wait_retry_interval} }}
+        # show-explorer-links = {default_show_explorer_links}
+        # accounts-file = "{default_accounts_file}"
+        # account = "{default_account}"
+        # keystore = "{default_keystore}"
+        "#,
+        default_url = FREE_RPC_PROVIDER_URL,
+        default_accounts_file = "~/.starknet_accounts/starknet_open_zeppelin_accounts.json",
```
ddoktorski, line 55: nit: There is a const `DEFAULT_ACCOUNTS_FILE`

crates/shared/src/consts.rs
ddoktorski, line 4: nit: Should be removed completely for now

crates/forge/src/init.rs
```diff
@@ -20,16 +21,19 @@ fn create_snfoundry_manifest(path: &PathBuf) -> Result<()> {
...
+
+        # [sncast.default]                                         # Define a profile name
+        # url = "{default_rpc_url}" # Url of the RPC provider
+        # accounts-file = "../account-file"                        # Path to the file with the account data
+        # account = "mainuser"                                     # Account from `accounts_file` or default account file that will be used for the transactions
+        # keystore = "~/keystore"                                  # Path to the keystore file
+        # wait-params = {{ timeout = 300, retry-interval = 10 }}     # Wait for submitted transaction parameters
+        # block-explorer = "StarkScan"                             # Block explorer service used to display links to transaction details
+        # show-explorer-links = true                               # Print links pointing to pages with transaction details in the chosen block explorer
+        "#,
```
ddoktorski, line 35: nit: Why not to share the default template between init and global config?

crates/sncast/src/main.rs
```diff
@@ -740,3 +735,23 @@ fn update_cast_config(config: &mut CastConfig, cli: &Cli) {
...
+        Utf8PathBuf::new()
+    });
+
+    let global_config =
+        load_global_config::<CastConfig>(&Some(global_config_path.clone()), &cli.profile)
+            .unwrap_or_else(|_| {
+                load_global_config::<CastConfig>(&Some(global_config_path), &None).unwrap()
+            });
+
+    let local_config = load_global_config::<CastConfig>(&None, &cli.profile)?;
```
cptartur, line 751: It's loading local config. Please rename this function.

crates/sncast/src/main.rs
```diff
@@ -740,3 +735,23 @@ fn update_cast_config(config: &mut CastConfig, cli: &Cli) {
...
+
+    let mut combined_config = combine_cast_configs(&global_config, &local_config);
+
+    update_config_with_cli(&mut combined_config, cli);
```
cptartur, line 755: Can we change the function `update_config_with_cli` to something like `config_with_cli` or similar and make it return a new config instead of taking a mutable reference?

docs/src/projects/configuration.md
```diff
@@ -105,6 +105,56 @@ response: [0x1, 0x23, 0x4]
 </details>
 <br>
 
+### Global configuration
+
+Global configuration file is usual [`snfoundry.toml`](https://foundry-rs.github.io/starknet-foundry/appendix/snfoundry-toml.html),
```
cptartur, line 110:
```suggestion
Global configuration file is a [`snfoundry.toml`](https://foundry-rs.github.io/starknet-foundry/appendix/snfoundry-toml.html), 
```

docs/src/projects/configuration.md
```diff
@@ -105,6 +105,56 @@ response: [0x1, 0x23, 0x4]
...
+
+Global configuration file is usual [`snfoundry.toml`](https://foundry-rs.github.io/starknet-foundry/appendix/snfoundry-toml.html), 
+which is a common storage for configurations to apply to multiple projects across various directories.
+This file is stored in a predefined location and is used to store profiles that can be used from any location on your computer.
+
+#### Interaction between Local and Global profiles
+
+Global config can be overridden by a local config.
+
+If both local and global profiles with the same name are present, local profile will be combined with global profile. That is for any setting defined both in global and local profiles, local setting will be used. For settings not defined in local profile, value from the global profile will be used instead.
```
cptartur, line 118:
```suggestion
If both local and global profiles with the same name are present, local profile will be combined with global profile. That is for any setting defined both in global and local profiles, local setting will be used. For settings not defined in local profile, values from the global profile will be used instead.
```

docs/src/projects/configuration.md
```diff
@@ -105,6 +105,56 @@ response: [0x1, 0x23, 0x4]
...
+#### Global Configuration File Location
+The global configuration is stored in a specific location depending on the operating system:
+
+- macOS/Linux : The global configuration file is located at `$HOME/.config/starknet-foundry/snfoundry.toml`
+- Windows : The file can be found at `C:\Users\<user>\AppData\Roaming\starknet-foundry\snfoundry.toml`
+
+> 📝 **Note**
+> If missing, global configuration file will be created automatically on running any `sncast` command for the first time.
+
+### Config interaction example
```
cptartur, line 131:
```suggestion
### Config Interaction Example
```

docs/src/projects/configuration.md
```diff
@@ -105,6 +105,56 @@ response: [0x1, 0x23, 0x4]
 </details>
 <br>
 
+### Global configuration
+
+Global configuration file is usual [`snfoundry.toml`](https://foundry-rs.github.io/starknet-foundry/appendix/snfoundry-toml.html), 
+which is a common storage for configurations to apply to multiple projects across various directories.
+This file is stored in a predefined location and is used to store profiles that can be used from any location on your computer.
+
+#### Interaction between Local and Global profiles
```
cptartur, line 114:
```suggestion
#### Interaction Between Local and Global Profiles
```

docs/src/projects/configuration.md
```diff
@@ -105,6 +105,56 @@ response: [0x1, 0x23, 0x4]
 </details>
 <br>
 
+### Global configuration
```
cptartur, line 108:
```suggestion
### Global Configuration
```

docs/src/projects/configuration.md
```diff
@@ -105,6 +105,56 @@ response: [0x1, 0x23, 0x4]
...
+
+In any directory in the file system, a user can run the `sncast` command using the `default` and `testnet` profiles, 
+because they are located in global config. 
+If no profiles are explicitly specified, the `default` profile from the global configuration file will be used.
+
+When running `sncast` from the `opus-magnum` directory, there is a configuration file in the parent directory (B). 
+This setup allows for the use of the following profiles: `default`, `testnet`, and `mainnet`. If the `mainnet` profile is specified, 
+the configuration from the local file will be used to override the global `default` profile, as the `mainnet` profile does not exist in the global configuration.
+
 ## Environmental variables
```
cptartur, line 158:
```suggestion
## Environmental Variables
```

docs/src/projects/configuration.md
```diff
@@ -105,6 +105,56 @@ response: [0x1, 0x23, 0x4]
...
+                └── opus-magnum/
+```
+
+**Glossary:**
+
+- **A:** Global configuration file containing the profiles [`default`](#default-profile) and [`testnet`](#defining-profiles-in-snfoundrytoml).
+- **B:** Local configuration file containing the profiles `default` and `mainnet`.
+
+In any directory in the file system, a user can run the `sncast` command using the `default` and `testnet` profiles, 
+because they are located in global config.
```
cptartur, line 151:
```suggestion
In any directory in the file system, a user can run the `sncast` command using the `default` and `testnet` profiles, 
because they are defined in global config (file A). 

```

docs/src/projects/configuration.md
```diff
@@ -105,6 +105,56 @@ response: [0x1, 0x23, 0x4]
...
+**Glossary:**
+
+- **A:** Global configuration file containing the profiles [`default`](#default-profile) and [`testnet`](#defining-profiles-in-snfoundrytoml).
+- **B:** Local configuration file containing the profiles `default` and `mainnet`.
+
+In any directory in the file system, a user can run the `sncast` command using the `default` and `testnet` profiles, 
+because they are located in global config. 
+If no profiles are explicitly specified, the `default` profile from the global configuration file will be used.
+
+When running `sncast` from the `opus-magnum` directory, there is a configuration file in the parent directory (B).
```
cptartur, line 154:
```suggestion
When running `sncast` from the `opus-magnum` directory, there is a configuration file in the parent directory (file B). 
```

docs/src/projects/configuration.md
```diff
@@ -105,6 +105,56 @@ response: [0x1, 0x23, 0x4]
...
+This file is stored in a predefined location and is used to store profiles that can be used from any location on your computer.
+
+#### Interaction between Local and Global profiles
+
+Global config can be overridden by a local config.
+
+If both local and global profiles with the same name are present, local profile will be combined with global profile. That is for any setting defined both in global and local profiles, local setting will be used. For settings not defined in local profile, value from the global profile will be used instead.
+
+This same behavior applies for [default profiles](#default-profile) as well. A local default profile will override a global default profile.
+
```
ddoktorski, line 121: I think the information that CLI values will always override values from both configs should be added somewhere

crates/sncast/src/helpers/configuration.rs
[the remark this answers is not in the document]
ddoktorski: What was decided here? Can we create a const in the `constants.rs` file instead?

crates/sncast/src/main.rs
kkawula: Renamed. I've also renamed `GlobalConfig` trait. This is no longer the correct name

crates/shared/src/consts.rs
kkawula: Yes, it was removed as a default from the CastConfig::Default, but I would like to leave it in the commented template.

cptartur (approved):

docs/src/projects/configuration.md
```diff
@@ -105,7 +105,62 @@ response: [0x1, 0x23, 0x4]
...
+Global configuration file is a [`snfoundry.toml`](https://foundry-rs.github.io/starknet-foundry/appendix/snfoundry-toml.html), 
+which is a common storage for configurations to apply to multiple projects across various directories.
+This file is stored in a predefined location and is used to store profiles that can be used from any location on your computer.
+
+#### Interaction Between Local and Global Profiles
+
+Global config can be overridden by a local config.
+
+If both local and global profiles with the same name are present, local profile will be combined with global profile. That is for any setting defined both in global and local profiles, local setting will be used. For settings not defined in local profile, values from the global profile will be used instead.
```
ddoktorski, line 118: The logic of overriding global default profil should be explicitly mention here, not only in the example below
```suggestion
If both local and global profiles with the same name are present, local profile will be combined with global profile. For any setting defined in both profiles, the local setting will take precedence. For settings not defined in the local profile, values from the corresponding global profile will be used, or if not defined, values from the global default profile will be used instead.
```

ddoktorski (approved):

PR #2730, opened 2024-11-29
Run some checks on windows

### LINKED ISSUE #2716: Run scheduled tests on windows machine
Currently, we do not run any checks on windows machines.

Update the scheduled job introduced in #2578, so tests are executed on windows as well.

cptartur:
Closes #2716 

## Introduced changes

-

## Checklist

- [ ] Linked relevant issue
- [ ] Added relevant tests


.github/workflows/ci.yml
```diff
@@ -23,56 +29,69 @@ jobs:
       - run: cargo test --release integration -p forge
 
   build-test-forge-e2e-nextest-archive:
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
```
ddoktorski, line 34: What is the reason to have `fail-fast` set to `false`?

.github/workflows/ci.yml
```diff
@@ -104,6 +123,10 @@ jobs:
         run: |
           curl -L https://raw.githubusercontent.com/software-mansion/cairo-coverage/main/scripts/install.sh | sh
 
+      - name: Install cairo-profiler
+        run: |
+          curl -L https://raw.githubusercontent.com/software-mansion/cairo-profiler/main/scripts/install.sh | sh
+
```
ddoktorski, line 129: Is it necessary here?

crates/forge-runner/src/lib.rs
```diff
@@ -68,9 +68,10 @@ pub fn maybe_save_trace_and_profile(
     }) = result
     {
         if execution_data_to_save.is_vm_trace_needed() {
-            let trace_path = save_trace_data(name, trace_data)?;
+            let name = sanitize_filename::sanitize(name.replace("::", "_"));
```
ddoktorski, line 71: This is a breaking change, please mention it in the changelog

crates/sncast/src/state/state_file.rs
```diff
@@ -396,7 +396,8 @@ mod tests {
     }
 
     #[test]
-    #[should_panic(expected = "No such file or directory (os error 2)")]
+    #[should_panic(expected = "Failed to load state file")]
+    // TODO handle system specific panics
```
ddoktorski, line 400: Is it going to be handled in this PR?

scripts/install_devnet.ps1
ddoktorski, line 1: Where are we using this script?

ddoktorski (commented):
Should we also update the scheduled workflow? Perhaps it's worth considering running tests on Windows only there instead of on every CI.

crates/forge/tests/e2e/io_operations.rs
```diff
@@ -31,7 +31,7 @@ fn file_reading() {
         [FAIL] file_reading_integrationtest::test::non_existent
         
         Failure data:
-            "No such file or directory [..]"
+            "[..] file [..]"
```
franciszekjob, line 34: Could we make this failure message dependent on platform (applies to other places)?

crates/forge/tests/e2e/running.rs
```diff
@@ -1045,7 +1045,7 @@ fn catch_runtime_errors() {
                 use snforge_std::fs::{FileTrait, read_txt};
 
                 #[test]
-                #[should_panic(expected: "No such file or directory (os error 2)")]
+                #[should_panic(expected: "file")]
```
franciszekjob, line 1048: Same as above

.github/workflows/ci.yml
```diff
@@ -188,23 +211,38 @@ jobs:
...
       - uses: dtolnay/rust-toolchain@7b1c307e0dcbda6122208f10795a713336a9b35a
         with:
           toolchain: stable
       - uses: Swatinem/rust-cache@82a92a6e8fbeee089604da2575dc567ae9ddeaab
-      - name: Install starknet-devnet-rs
-        run: ./scripts/install_devnet.sh
+      - name: Install starknet-devnet-rs on Linux/Macos
+        if: runner.os != 'Windows'
+        run: |
+          ./scripts/install_devnet.sh
```
franciszekjob, line 227: nit:
```suggestion
        run: ./scripts/install_devnet.sh
```

.github/workflows/ci.yml
```diff
@@ -1,5 +1,8 @@
 name: CI
 
+env:
+  DEVNET_TAG: "v0.2.2"
+
```
franciszekjob, line 5: We should use commit hash because it is immutable, unlike a tag which is mutable therefore could theoretically pose some risk.

crates/sncast/src/state/state_file.rs
franciszekjob: Please link issue number if it won't be implemented in this PR 🙏

franciszekjob (changes requested):

.github/workflows/ci.yml
cptartur: It's not, removing

crates/forge-runner/src/lib.rs
cptartur: It's only changing the names of the files, the directory remains the same. I can mention it, but I don't think we ever documented the format of the names anywhere.

crates/sncast/src/state/state_file.rs
cptartur: No longer needed, we are catching our error now. Removing todo.

scripts/install_devnet.ps1
cptartur: In sncast tests ctor

crates/forge/tests/e2e/io_operations.rs
cptartur: Actually, I'd do the opposite. We should wrap these errors so we don't output OS errors directly to the user. And specifically, we shouldn't assert specific os errors in tests.

cptartur: But until we do so, maybe I can assert the os specific error here

scripts/install_devnet.ps1
ddoktorski: Do you mean for local development on Windows? In the CI workflow it is currently not used to install devnet

crates/forge-runner/src/lib.rs
ddoktorski: Yes, but potentially it can be used as an input to profiler or coverage in some workflows

scripts/install_devnet.ps1
cptartur: Sorry, yeah. It's only for local development now. It could be used in CI, but since we do caching there, the script would have to be changed so it supports passing the installation path.

crates/forge/tests/e2e/io_operations.rs
cptartur: Changed all of them to be OS dependent except for one assert that asserted the path `scripts/`.

.github/workflows/ci.yml
cptartur: Updated here as well

scripts/install_devnet.ps1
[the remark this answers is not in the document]
cptartur: Removed

.github/workflows/ci.yml
```diff
@@ -23,56 +29,69 @@ jobs:
       - run: cargo test --release integration -p forge
```
ddoktorski, line 30: nit: `cargo test --release --bin snforge` above should be removed as it doesn't test anything

ddoktorski (approved):
I still think it would be better to test Windows only in the scheduled workflow. Testing it with every single change isn’t crucial and overall it will slow down our CI and make it harder to maintain.

franciszekjob:
@ddoktorski are you reffering to caching with setup-scarb? If so, I created https://github.com/foundry-rs/starknet-foundry/issues/2766.

cptartur:
> Can we create an issue to investigate whether the cache action functions correctly, considering that the first CI run on a branch takes significantly longer than subsequent runs, and on Windows it is much longer than on Linux?
> 
> Example: https://github.com/foundry-rs/starknet-foundry/actions/runs/12178894598/job/33969862907 <- first run does not find any cache https://github.com/foundry-rs/starknet-foundry/actions/runs/12179972960/job/33973292419

@ddoktorski I'd just merge this and we'll see if te build times are problematic. And if so, we can disable these checks at any time.

crates/forge/tests/e2e/mod.rs
```diff
@@ -11,7 +11,9 @@ mod coverage;
 mod docs_snippets_validation;
 mod env;
 mod features;
+#[cfg(not(target_os = "windows"))]
 mod fork_warning;
+#[cfg(not(target_os = "windows"))]
```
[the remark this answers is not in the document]
cptartur: https://github.com/foundry-rs/starknet-foundry/issues/2835

PR #2770, opened 2024-12-11
Add `snforge new` command

### LINKED ISSUE #2645: Rename `snforge init` to `snforge new`
`snforge init` behaves like `snforge new` - it creates a new directory instead of initializing an already existing one.

It should be renamed to `snforge new`.

This should be done in a non-breaking manner: Deprecated `snforge init` while adding `new` and only remove `init` after some time has passed since deprecation.

In the scope of this issue ensure that `scarb init` and `scarb new` works correctly when choosing the snforge test runner. Currently there is a problem with creating a new project when current working directory name is different than the provided name with the `--name` flag. Please see https://github.com/software-mansion/scarb/issues/1759 for reference.

ddoktorski:
Closes #2645 

## Introduced changes

- `snforge new` command, which allows specifying the name of the package and the path where it should be created
- Deprecation warning to the `snforge init` command

Docs and changelog will be updated in a separate PR.

## Checklist

- [X] Linked relevant issue
- [X] Added relevant tests


crates/forge/src/init.rs
```diff
@@ -1,223 +1,24 @@
...
-
-    add_template_to_scarb_manifest(&scarb_manifest_path)?;
-
-    if !snfoundry_manifest_path.is_file() {
-        create_snfoundry_manifest(&snfoundry_manifest_path)?;
+    let project_path = Utf8PathBuf::from_path_buf(current_dir)
+        .expect("Failed to create Utf8PathBuf for the current directory")
+        .join(project_name);
+
+    if !project_path.join("Scarb.toml").exists() {
```
ddoktorski, line 12: Note: This is a hack to prevent printing this warning when running `scarb init/new` with an older version of Scarb.

crates/forge/src/lib.rs
```diff
@@ -76,6 +78,11 @@ enum ForgeSubcommand {
         /// Name of a new project
         name: String,
     },
+    /// Create a new Forge project at <PATH>
```
franciszekjob, line 81: `PATH` here is a bit reminiscent of `$PATH` environment variable.
```suggestion
    /// Create a new Forge project at <path>
```

crates/forge/src/lib.rs
```diff
@@ -160,6 +167,20 @@ pub struct TestArgs {
...
+#[derive(Parser, Debug)]
+pub struct NewArgs {
+    path: Utf8PathBuf,
+    /// Name of a new project, defaults to the directory name
+    #[arg(short, long)]
+    name: Option<String>,
+    /// Do not initialize a new Git repository
+    #[arg(long)]
+    no_vcs: bool,
+    /// Try to create the project even if the specified directory at <PATH> is not empty, which can result in overwriting existing files
```
franciszekjob, line 179: Same as `PATH` above

crates/forge/src/lib.rs
```diff
@@ -160,6 +167,20 @@ pub struct TestArgs {
     additional_args: Vec<OsString>,
 }
 
+#[derive(Parser, Debug)]
+pub struct NewArgs {
+    path: Utf8PathBuf,
```
franciszekjob, line 172: Doc comment is missing

crates/forge/src/lib.rs
ddoktorski: I see, but this is how it will be shown after `snforge new --help`:
```
Usage: scarb new [OPTIONS] <PATH>
```
Scarb has a similar description for the `scarb new` command:
```
Create a new Scarb package at <PATH>
```

franciszekjob: Ok I see, let's leave it then, it's not a big thing.

franciszekjob: Ok I see, let's leave it then, it's not a big thing.

crates/forge/src/init.rs
```diff
@@ -1,223 +1,24 @@
...
+        print_as_warning(&anyhow!(
+            "Command `snforge init` has been deprecated and will be removed in an upcoming version. Please use `snforge new` instead."
```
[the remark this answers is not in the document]
cptartur: Not sure if we'll remove it right away

crates/forge/src/lib.rs
cptartur: 👆

crates/forge/src/new.rs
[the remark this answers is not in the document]
cptartur: I'd prefer we keep the init/new naming the same across scarb and forge.

crates/forge/tests/e2e/running.rs
```diff
@@ -696,7 +697,61 @@ fn init_new_project() {
         .assert()
         .success();
 
-    validate_init(&temp, false);
+    validate_init(&temp.join("test_name"), false);
```
cptartur, line 700: Can we also assert that the deprecation warning is emitted

crates/forge/tests/e2e/running.rs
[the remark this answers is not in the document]
ddoktorski: It is tested by `init_new_project_from_scarb`

crates/forge/src/new.rs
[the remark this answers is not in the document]
ddoktorski: The only difference now is that `snforge new` allows to create a package in an existing location, `scarb new` does not allow this.

franciszekjob (approved):
Let's just change the PR description (it says `snforge add`) and we should be good to go 😄

crates/forge/src/init.rs
cptartur: Maybe let's document this in code then

cptartur (approved):

PR #2773, opened 2024-12-13
Default account mechanism for sncast

### LINKED ISSUE #2695: Default account mechanism for sncast
Currently, `--account` must must be always provided in sncast. It can be done through CLI or `snfoundry.toml` file. 

#2221 will allow defining a default account for all sncast projects. This is not very well integrated with our account commands though.

Add an interactive prompt to `account import` and `account create` that will prompt the user to make the account they imported default. It should be something like this

```shell
> Do you want to set the imported (created) account as default?
Yes
No
Don't ask again
```

On selecting `Yes`, user should be informed that the default was stored in global config and how to find/change it.

Additionally, `account list` should indicate which account is set as a default.

If using `sncast` command which requires `--account`, in case the flag is missing, the output should be updated to provide an information about setting the default account.

### LINKED ISSUE #2696: Use `strk` as a default `--fee-token`
Currently, user is required to provide `--fee-token` or `--version` everytime they make a transaction.
This is impractical in terms of UX.
Additionally, txs supporting eth fee token are most likely to be removed in the upcoming future https://community.starknet.io/t/snip-16-deprecation-of-transaction-versions-0-1-2/114443

Use `strk` as a default `--fee-token` in sncast operations that require providing it.

Additionally, documentation should be updated to stop showing the `--fee-token` or `--version` argument except for the "Fees and Versions" page.

kkawula:
Closes #2695

## Introduced changes

- Added interactive prompt when creating or importing new account
- Added `silent` flag, to disable interactive selection dialog

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


cptartur (commented):
Think the behavior looks good. Not sure if always asking to create a local config won't cause in users creating multiple local configs at different directories. Maybe it only should ask about that if cwd is a starknet foundry project. 
@franciszekjob @ddoktorski what do you think?

franciszekjob:
> Think the behavior looks good. Not sure if always asking to create a local config won't cause in users creating multiple local configs at different directories. Maybe it only should ask about that if cwd is a starknet foundry project. @franciszekjob @ddoktorski what do you think?

I think we should prompt to add default in local config only when snfoundry.toml is present. For example: someone is using sncast for one of the first times, doesn't have snfoundry.toml, and is creating/importing an account. For such a person, it would probably be better if only the option to add to global pops up, because if they see two options (global and local), they might need to think it over for a while. A beginner user is likely not familiar with configs yet, so it’s better to limit the "level of difficulty".

crates/sncast/src/helpers/interactive.rs
```diff
@@ -0,0 +1,114 @@
...
+        let option = format!(
+            "Yes, global default account ({}).",
+            to_tilde_path(global_path.as_str())
+        );
+        options.push(option);
+    }
+
+    if let Ok(current_path) = current_dir() {
+        let current_path_utf8 = Utf8PathBuf::from_path_buf(current_path.clone())
+            .expect("Failed to convert current directory to Utf8PathBuf");
```
cptartur, line 26: I don't think `expect` is a good approach here. I'd either bubble the error or handle it by not displaying the local option.

crates/sncast/src/helpers/interactive.rs
```diff
@@ -0,0 +1,114 @@
...
+    options.push("No".to_string());
+
+    let selection = Select::with_theme(&ColorfulTheme::default())
+        .with_prompt("Do you want to make this account default?")
+        .items(&options)
+        .default(0)
+        .interact()
+        .context("Failed to display selection dialog")?;
+
+    match options[selection].as_str() {
```
cptartur, line 46: Could we use an enum here somehow? So our logic doesn't depend on matching strings.

crates/sncast/src/helpers/interactive.rs
```diff
@@ -0,0 +1,114 @@
...
+        selected if selected.starts_with("Yes, global default") => {
+            if let Ok(global_path) = get_global_config_path() {
+                edit_config(&global_path, "default", "account", account)?;
+            }
+        }
+        selected if selected.starts_with("Yes, local default") => {
+            if let Ok(current_path) = current_dir() {
+                let current_path_utf8 = Utf8PathBuf::from_path_buf(current_path)
+                    .expect("Failed to convert current directory to Utf8PathBuf");
```
cptartur, line 55: This should bubble not panic

crates/sncast/src/helpers/interactive.rs
```diff
@@ -0,0 +1,114 @@
...
+        .parse::<DocumentMut>()
+        .expect("Failed to parse TOML");
+    update_config(&mut toml_doc, profile, key, value);
+
+    File::create(config_path)?.write_all(toml_doc.to_string().as_bytes())?;
+
+    Ok(())
+}
+
+pub fn update_config(toml_doc: &mut DocumentMut, profile: &str, key: &str, value: &str) {
```
cptartur, line 82: Does this need to be public?

crates/sncast/tests/integration/config_edit.rs
cptartur, line 1: I think this could just be a unit test in the module defining this function. 

Also, if we want to have an utility like this, let's at least add a test case where the key is missing and needs to be created.

crates/sncast/src/helpers/interactive.rs
```diff
@@ -0,0 +1,114 @@
...
+
+    let profile_table = sncast_table
+        .get_mut(profile)
+        .and_then(|item| item.as_table_mut())
+        .expect("Failed to create or access profile table");
+
+    profile_table[key] = Value::from(value).into();
+}
+
+fn to_tilde_path(path: &str) -> String {
```
cptartur, line 104: I think maybe on windows we should just print the whole path? 
It can be handled with `cfg!(target_os = "windows")`

crates/sncast/src/helpers/interactive.rs
kkawula: I don't think so. I can see that it is common way to achieve this: https://github.com/software-mansion/scarb/blob/main/scarb/src/bin/scarb/interactive.rs#L40

franciszekjob: I see that in Scarb it's implemented this way but I wouldn't consider this a criterion. Imo num fits here well - selection message is defined only once and it looks neater.

kkawula: https://github.com/foundry-rs/starknet-foundry/pull/2792

kkawula: That's how it would have to look like

franciszekjob (approved):
LGTM

crates/sncast/src/helpers/interactive.rs
```diff
@@ -0,0 +1,199 @@
...
+    if let Some(option) = get_global_config_path().ok().map(|global_path| {
+        format!(
+            "Yes, global default account ({}).",
+            to_tilde_path(global_path.as_str())
+        )
+    }) {
+        options.push(option);
+    }
+
+    if let Some(option) = current_dir()
```
ddoktorski, line 25: nit: Also in other places
```suggestion
    if let Some(option) = env::current_dir()
```

crates/sncast/src/helpers/interactive.rs
```diff
@@ -0,0 +1,199 @@
...
+        })
+    {
+        options.push(option);
+    }
+
+    options.push("No".to_string());
+
+    let selection = Select::with_theme(&ColorfulTheme::default())
+        .with_prompt("Do you want to make this account default?")
+        .items(&options)
```
ddoktorski, line 43: Wasn't it supposed to allow only one option either local or global, not both?

crates/sncast/src/helpers/interactive.rs
```diff
@@ -0,0 +1,199 @@
...
+        }
+        _ => {}
+    }
+
+    Ok(())
+}
+
+fn edit_config(config_path: &Utf8PathBuf, profile: &str, key: &str, value: &str) -> Result<()> {
+    let mut file_content = String::new();
+    File::open(config_path)?.read_to_string(&mut file_content)?;
```
ddoktorski, line 73:
```suggestion
    let file_content = fs::read_to_string(config_path)?;
```

crates/sncast/src/helpers/interactive.rs
```diff
@@ -0,0 +1,199 @@
...
+
+    let mut toml_doc = file_content
+        .parse::<DocumentMut>()
+        .context("Failed to parse TOML")?;
+    update_config(&mut toml_doc, profile, key, value);
+
+    File::create(config_path)?.write_all(toml_doc.to_string().as_bytes())?;
```
ddoktorski, line 80:
```suggestion
    fs::write(config_path, toml_doc.to_string())?;
```

crates/sncast/src/helpers/interactive.rs
```diff
@@ -0,0 +1,199 @@
...
+
+    let profile_table = sncast_table
+        .get_mut(profile)
+        .and_then(|item| item.as_table_mut())
+        .expect("Failed to create or access profile table");
+
+    profile_table[key] = Value::from(value).into();
+}
+
+fn to_tilde_path(path: &str) -> String {
```
ddoktorski, line 107: Why does it accept `&str` and not path related type?

crates/sncast/src/helpers/interactive.rs
```diff
@@ -0,0 +1,199 @@
...
+    if cfg!(not(target_os = "windows")) {
+        if let Some(home_dir) = dirs::home_dir() {
+            let input_path = Path::new(path);
+
+            if let Ok(stripped_path) = input_path.strip_prefix(&home_dir) {
```
ddoktorski, line 112: Is `input_path` always in canonicalized form?

crates/sncast/src/helpers/interactive.rs
kkawula: So in the end it was decided to show user his local config if there is one

cptartur: I think both make more sense

crates/sncast/src/helpers/interactive.rs
cptartur: Okay, I'll take a look at this later.

crates/sncast/src/helpers/interactive.rs
```diff
@@ -0,0 +1,193 @@
...
+                edit_config(&global_path, "default", "account", account)?;
+            }
+        }
+        selected if selected.starts_with("Yes, local default") => {
+            if let Ok(current_path) = env::current_dir() {
+                let current_path_utf8 = Utf8PathBuf::from_path_buf(current_path).map_err(|_| {
+                    anyhow!("Failed to convert current directory path to Utf8PathBuf")
+                })?;
+
+                if let Ok(local_path) = search_config_upwards_relative_to(&current_path_utf8) {
```
cptartur, line 55: We're duplicating the search logic, can we store the local path if it exists and use it here and in the line 22?

crates/sncast/src/helpers/interactive.rs
```diff
@@ -0,0 +1,193 @@
...
+                }
+            }
+        }
+    }
+
+    path.to_string()
+}
+
+#[cfg(test)]
+mod tests {
```
cptartur, line 116: Let's also add a test case where there is no `[sncast]` section at all so it has to be created for a given profile.

And another case where there's `[sncast.default]` but no `[sncast.testnet]`, which has to be created.

crates/sncast/src/helpers/interactive.rs
```diff
@@ -0,0 +1,196 @@
...
+        .context("Failed to display selection dialog")?;
+
+    match &options[selection] {
+        PromptSelection::GlobalDefault(path) => {
+            edit_config(path, "default", "account", account)?;
+        }
+        PromptSelection::LocalDefault(path) => {
+            edit_config(path, "default", "account", account)?;
+        }
+        PromptSelection::No => {}
```
Draggu, line 63:
```suggestion
        PromptSelection::GlobalDefault(path) | PromptSelection::LocalDefault(path) => {
            edit_config(path, "default", "account", account)?;
        }
        PromptSelection::No => {}
```

crates/sncast/src/helpers/interactive.rs
```diff
@@ -0,0 +1,270 @@
...
+        assert_eq!(toml_doc.to_string(), expected);
+    }
+
+    #[test]
+    fn test_create_table_empty_file() {
+        let original = "";
+
+        let expected = formatdoc! {
+            r#"
+            [sncast]
```
kkawula, line 258: I am unsure if it is possible to avoid the creation of this artifact.  It's probably due to how `edit_toml` work

kkawula: And it doesn't change anything

CHANGELOG.md
```diff
@@ -16,6 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Minimal Rust version in requirements check is the same as in docs (`1.80.1`)
 - `snforge` produces trace for contracts even if they fail or panic (assuming test passed)
 
+
```
kkawula, line 26:
```suggestion
```

crates/sncast/src/helpers/interactive.rs
```diff
@@ -0,0 +1,267 @@
...
+    GlobalDefault(Utf8PathBuf),
+    LocalDefault(Utf8PathBuf),
+    No,
+}
+
+impl Display for PromptSelection {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let str = match self {
+            PromptSelection::LocalDefault(path) => {
+                format!("Yes, local default ({})", to_tilde_path(path))
```
Draggu, line 21:
```suggestion
                write!("Yes, local default ({})", to_tilde_path(path))
```

crates/sncast/src/helpers/interactive.rs
```diff
@@ -0,0 +1,267 @@
...
+            edit_config(path, "default", "account", account)?;
+        }
+        PromptSelection::No => {}
+    }
+
+    Ok(())
+}
+
+fn edit_config(config_path: &Utf8PathBuf, profile: &str, key: &str, value: &str) -> Result<()> {
+    let file_content = fs::read_to_string(config_path)?;
```
Draggu, line 67: `.context()`

crates/sncast/src/helpers/interactive.rs
```diff
@@ -0,0 +1,267 @@
...
+
+    let mut toml_doc = file_content
+        .parse::<DocumentMut>()
+        .context("Failed to parse TOML")?;
+    update_config(&mut toml_doc, profile, key, value);
+
+    fs::write(config_path, toml_doc.to_string())?;
```
Draggu, line 74: `.context()`

crates/sncast/src/main.rs
```diff
@@ -489,6 +492,18 @@ async fn run_async_command(
                 )
                 .await;
 
+                if !import.silent && result.is_ok() && io::stdout().is_terminal() {
+                    if let Some(account_name) =
+                        result.as_ref().ok().and_then(|r| r.account_name.clone())
+                    {
+                        prompt_to_add_account_as_default(account_name.as_str()).unwrap_or_else(
```
Draggu, line 499:
```suggestion
                        if let Err(err) = prompt_to_add_account_as_default(account_name.as_str()) {
```
you don't use value here so no need to `unwrap`

crates/sncast/src/main.rs
```diff
@@ -516,6 +531,12 @@ async fn run_async_command(
                 )
                 .await;
 
+                if !create.silent && result.is_ok() && io::stdout().is_terminal() {
+                    prompt_to_add_account_as_default(&account).unwrap_or_else(|err| {
```
Draggu, line 535: vide supra

Draggu (approved):

ddoktorski (approved):

crates/sncast/src/helpers/interactive.rs
```diff
@@ -0,0 +1,266 @@
...
+    }
+
+    let sncast_table = toml_doc
+        .get_mut("sncast")
+        .and_then(|item| item.as_table_mut())
+        .expect("Failed to create or access 'sncast' table");
+
+    if !sncast_table.contains_key(profile) {
+        sncast_table[profile] = Item::Table(Table::new());
+    }
```
cptartur, line 90: Wouldn't that solve the problem with empty `[sncast]` table being created?

```suggestion
    let sncast_table = toml_doc
        .get_mut(format!("sncast.{profile}"))
        .and_then(|item| item.as_table_mut())
        .expect("Failed to create or access 'sncast' table");
```

kkawula: Nope

Issue #2697, opened 2024-11-21
Invoking txs fail if fee-token strk is selected and --max-fee is used


amanusk:
### Which component is your bug related to?

sncast

### Foundry Version

0.33.0

### What operating system are you using?

Linux

### What system architecture are you using?

x86

### What happened

Running a command like:

```
sncast -a sep_acc_1 invoke  -d 0x04718f5a0fc34cc1af16a1cdee98ffb20c31f5cd61d6ab07201858f4287c938d -f transfer -c 0x7f3ebe4b3851f8a266dd4fa3dd77bad41aa5c0e18dab00efc505b5ceb96a286 --url https://starknet-sepolia.public.blastapi.io/rpc/v0_7 --fee-token strk  --max-fee 9999999999999 
```

Yields an error
`error: Unexpected RPC error occurred: Transaction must commit to pay a positive amount on fee.`

max-fee is a param that should only work with V1 tx (paying ETH), while V3 txs require specifying gas per unit.

In this case it seems the max-fee overrides the selection of strk token, and expects ETH payment. The error should be more informative, or to automatically derive the needed params for the specified max-fee.

This is also the way it is presetned in the docs:
https://foundry-rs.github.io/starknet-foundry/starknet/account.html#deploy-account-with-the-sncast-account-deploy-command

The command yields a similar error

### Is there an existing issue for this?

- [ ] I have searched the existing issues and verified no issue exits for this problem.

ddoktorski:
Hi @amanusk

In the current implementation `--max-fee` is treated the same for both tokens: it represents the total maximum gas that can be paid for the transaction. In your example, what really happens under the hood is that `sncast` first makes an RPC call to get the current gas price in FRI. Then it divides the provided `max_fee` by this value to calculate the `max_gas` (referred to as `max_amount` in the RPC). [Here](https://github.com/foundry-rs/starknet-foundry/blob/master/crates/sncast/src/helpers/fee.rs#L115-L132) is the link to the implementation for your reference. Following this, it attempts to execute the request.

With the value `--max-fee 9999999999999`, the current gas price in FRI is higher, and the resulting division is 0, which ultimately causes the above error. The same error can be replicated if you use the following flags: `--fee-token strk --max-gas 0 --max-gas-unit-price X`.

cptartur:
Created #2707 and #2706 from this

PR #2796, opened 2024-12-18
Validate max gas is greater than zero for strk fee settings

### LINKED ISSUE #2706: Validate fee args is greater than 0

franciszekjob:
Closes #2706

## Introduced changes

Validate if fee args are greater than 0
- Change type of `max_fee`, `max_gas` and `max_gas_unit_price` from `Option<Felt>` to `Option<NonZeroFelt>`

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/sncast/src/helpers/fee.rs
```diff
@@ -106,29 +113,36 @@ impl FeeArgs {
...
-                        max_gas_unit_price: Some(
-                            max_fee
-                                .floor_div(&NonZeroFelt::from_felt_unchecked(max_gas))
-                                .try_into_()?,
-                        ),
-                    },
+                    (Some(max_fee), Some(max_gas), None) => {
+                        let max_gas_unit_price =
+                            max_fee.floor_div(&NonZeroFelt::from_felt_unchecked(max_gas));
+                        if max_gas_unit_price == Felt::ZERO {
```
ksew1, line 119: The unchecked word should signal that it is not something we should use, as handling it incorrectly can lead to undefined behavior. In Rust, unchecked constructs like this are typically marked as `unsafe`. For example, see [`std::num::NonZero`](https://github.com/rust-lang/rust/blob/3378a5e084fa51c2e7f5c3f53b87e8ad89f1c517/library/core/src/num/nonzero.rs#L397). I’m not sure why Jonathan didn’t make it `unsafe`.

Previously, this functionality was clearly incorrect because we didn’t check whether it was zero. Now, we do check it, but the check happens *after* the operation. 

The sane approach would look something like this:

```rust
&NonZeroFelt::try_from(max_gas_unit_price)
    .context("--max-gas calculated from --max-fee should be greater than 0. Please increase --max-fee")?
```

Additionally, please remove the other places where `from_felt_unchecked` is used.

ksew1: Also (probably) the best way to handle this situation is to create our own wrapper type with the `FromStr` trait, which disallows the number `0`. This way, `clap`, while parsing the command, would immediately return an error:

```rs
struct NonZeroFelt(Felt);

impl FromStr for NonZeroFelt {
    type Err = String;

        match s.parse::<Felt>() {
            Ok(felt) => {
                if felt == Felt::ZERO {
                    Err(String::from("value should be greater than 0"))
                } else {
                    Ok(NonZeroFelt(felt))
                }
            }
            Err(_) => Err(String::from("value should be a number")),
        }
    }
}
```

ksew1: I’m sorry if I seem mean—I don’t intend to be. But together, we can forge a better place. 
a1d2b4446d5bbf9659a5279470acae9c

Draggu: > I’m not sure why Jonathan didn’t make it unsafe.

Because it is not how `unsafe` is designed to be used. It is not for invariant correctness but for memory correctness.
Rust's `NonZero` types have the `unchecked` method marked as `unsafe` because it can violate memory safety. These types exist to tell the compiler: "Hey, I will NOT use a 0 value, you can do whatever you want with it." As such, initializing it with 0 not only breaks the invariant but also the memory layout. For invariant correctness you should use `Result` and `.unwrap()`.

ksew1: Your understanding of how `unsafe` is designed in Rust is incomplete. `unsafe` is not restricted solely to ensuring memory safety; it also encompasses avoiding undefined behavior. Rust programs, whether they use `unsafe` code or not, **must never cause undefined behavior.** The difference is that, with `unsafe` code, the programmer assumes the responsibility of upholding certain invariants to ensure this requirement is met. 

As explicitly noted in the [Rust reference](https://doc.rust-lang.org/reference/behavior-considered-undefined.html): 

> *"Unsafe only means that avoiding undefined behavior is on the programmer; it does not change anything about the fact that Rust programs must never cause undefined behavior."*

Moreover, the reference also clarifies that **there is no formal model yet for all possible undefined behavior in unsafe Rust.** This means Rust leaves it to the programmer to carefully adhere to invariants that are necessary to prevent UB. The list of behaviors currently considered undefined is subject to change, as stated in the documentation:  

> *"This list is not exhaustive; it may grow or shrink. There is no formal model of Rust’s semantics for what is and is not allowed in unsafe code..."*

So, invariants in unsafe code are not limited to memory correctness but are crucial for avoiding any form of undefined behavior. For example, the documentation for methods like `std::slice::get_unchecked` explicitly states:

```
/// Calling this method with an out-of-bounds index is _[undefined behavior]_
/// even if the resulting reference is not used.
```

This demonstrates that the scope of invariants in `unsafe` code extends beyond just memory layout and memory access patterns—it includes anything that could lead to undefined behavior.

To apply this to the specific context of your example, the method `from_felt_unchecked` explicitly states in its documentation:

> *"This can lead to undefined behavior and big security issues."*

Thus, it is reasonable to treat this code as unsafe because violating its invariants can lead directly to undefined behavior, which is what the `unsafe` keyword is designed to guard against. Violating the invariant is not just a correctness issue—it is a potential source of UB, which makes `unsafe` absolutely appropriate here.

ksew1: In the context of `NonZero`, memory layout optimization guarantees that `Option<NonZero<T>>` has the same layout and size as `T`, with `0` representing `None`. Really cool stuff!

kkawula: boom

kkawula: boom

Draggu: While your explanation of the broader scope of `unsafe` in rust is accurate, for now, every instance of undefined behavior (UB) explicitly mentioned in the official rust documentation relates to memory safety. Rust's current model of UB, though incomplete, consistently centers around violations that affect memory layout, access, or usage in unsafe code.

The elements in the Rust reference's list suggest that the current model of `unsafe` in rust is de facto primarily focused on preserving memory safety. Even cases like using `std::slice::get_unchecked` with out-of-bounds indices fall under this umbrella, as they involve accessing memory in an unchecked manner, potentially leading to memory corruption.

The argument about invariants is valid in the theoretical scope of `unsafe`, but in practice, Rust’s usage of `unsafe` is grounded in addressing memory-related concerns.

In the case of `from_felt_unchecked`, its invariants can lead to undefined behavior due to violations impacting memory correctness, so marking it as `unsafe` is justified.

A better API in this case would be something like `from_felt(v).expect("v must not be 0")`.

ksew1: Just because most instances of undefined behavior (UB) currently documented in Rust relate to memory safety doesn't mean that the scope of `unsafe` is limited only to memory-related issues. The Rust Reference itself does not make such a statement or impose such a restriction. 

The `unsafe` keyword, as described in the Rust Reference, serves a broader purpose: it marks code that either **defines** or **satisfies** additional safety requirements. Specifically, it plays several roles depending on where it is used:

- It is used to **define** extra safety conditions in contexts such as `unsafe fn` and `unsafe trait`.
- It is used to **satisfy** extra safety conditions in contexts such as `unsafe {}`, `unsafe impl`, `unsafe extern`, or even `unsafe fn` under certain circumstances.

This perspective underscores that `unsafe` is not exclusively tied to ensuring memory safety but is instead about acknowledging and upholding additional safety invariants—whatever they may be. While memory-related issues dominate the documented cases of UB, this doesn't limit the theoretical or semantic scope of `unsafe`.

And I completely agree that a better API would involve using `Result`. In fact, such an API already exists in the form of [`TryFrom`](https://github.com/starknet-io/types-rs/blob/7ef30234da7957f7585ea832d57d13083c76995c/crates/starknet-types-core/src/felt/mod.rs#L514). So, @franciszekjob, let’s just use that, and I think we should be done here.

Draggu: Agree that it is non-exhaustive list, but as previously stated **in practice** it is memmory related. It was simplification so we don't have to make this discussion :sweat_smile:. Please stick with `TryFrom`.

crates/sncast/tests/e2e/invoke.rs
kkawula, line 1: Do these tests refer to the reported issue?
https://github.com/foundry-rs/starknet-foundry/issues/2697
I suppose that even if max_fee is greater than 0 an error may occur

franciszekjob: These ones refer to situation when user explicitly passes 0 to fee args. 
The case when **max gas or max gas unit price** is calculated based on other values, is handled e.g. [here](https://github.com/foundry-rs/starknet-foundry/blob/969e37b0551bc6ee68880d17216c5b0c03711af5/crates/sncast/src/helpers/fee.rs#L111).

crates/sncast/src/helpers/fee.rs
```diff
@@ -15,24 +15,24 @@ pub struct FeeArgs {
...
+    pub max_gas: Option<NonZeroFelt>,
 
     /// Max gas price in Fri. If not provided, will be automatically estimated. (Only for STRK fee payment)
-    #[clap(long)]
-    pub max_gas_unit_price: Option<Felt>,
+    #[clap(value_parser = parse_non_zero_felt, long)]
+    pub max_gas_unit_price: Option<NonZeroFelt>,
 }
 
 impl From<ScriptFeeSettings> for FeeArgs {
```
ksew1, line 30: I would say that we should go step further and do:

```rs
#[derive(Debug, PartialEq)]
pub enum FeeSettings {
    Eth {
        max_fee: Option<NonZeroFelt>,
    },
    Strk {
        max_gas: Option<NonZeroU64>,
        max_gas_unit_price: Option<NonZeroU128>,
    },
}
```

and 

```rs
#[derive(Debug, PartialEq, CairoDeserialize)]
pub enum ScriptFeeSettings {
    Eth {
        max_fee: Option<NonZeroFelt>,
    },
    Strk {
        max_fee: Option<NonZeroFelt>,
        max_gas: Option<NonZeroU64>,
        max_gas_unit_price: Option<NonZeroU128>,
    },
}
```

crates/sncast/src/helpers/fee.rs
```diff
@@ -92,43 +96,41 @@ impl FeeArgs {
...
-                    (Some(max_fee), Some(max_gas), None) => FeeSettings::Strk {
-                        max_gas: Some(max_gas.try_into_()?),
-                        max_gas_unit_price: Some(
-                            max_fee
-                                .floor_div(&NonZeroFelt::from_felt_unchecked(max_gas))
-                                .try_into_()?,
-                        ),
-                    },
+                    (Some(max_fee), None, Some(max_gas_unit_price)) => {
+                        let max_gas = NonZeroFelt::try_from(Felt::from(max_fee).floor_div(&max_gas_unit_price)).expect("Calculated max gas from provided --max-fee and --max-gas-unit-price is zero. Please increase --max-fee to obtain a positive gas amount");
```
ksew1, line 111:
```suggestion
                        let max_gas = NonZeroFelt::try_from(Felt::from(max_fee).floor_div(&max_gas_unit_price)).context("Calculated max gas from provided --max-fee and --max-gas-unit-price is zero. Please increase --max-fee to obtain a positive gas amount")?;
```

crates/sncast/src/helpers/fee.rs
```diff
@@ -92,43 +96,41 @@ impl FeeArgs {
...
+                        FeeSettings::Strk {
+                            max_gas: Some(Felt::from(max_gas).try_into_()?),
+                            max_gas_unit_price: Some(Felt::from(max_gas_unit_price).try_into()?),
+                        }
+                    }
+                    (Some(max_fee), Some(max_gas), None) => {
+                        let max_gas_unit_price = NonZeroFelt::try_from(Felt::from(max_fee).floor_div(&max_gas)).expect("Calculated max gas unit price from provided --max-fee and --max-gas is zero. Please increase --max-fee or decrease --max-gas to ensure a positive gas unit price");
```
ksew1, line 118:
```suggestion
                        let max_gas_unit_price = NonZeroFelt::try_from(Felt::from(max_fee).floor_div(&max_gas)).context("Calculated max gas unit price from provided --max-fee and --max-gas is zero. Please increase --max-fee or decrease --max-gas to ensure a positive gas unit price")?;
```

crates/sncast/src/helpers/fee.rs
```diff
@@ -92,43 +96,41 @@ impl FeeArgs {
...
+                    }
                     (Some(max_fee), None, None) => {
                         let max_gas_unit_price = provider
                             .get_block_with_tx_hashes(block_id)
                             .await?
                             .l1_gas_price()
                             .price_in_fri;
-
+                        let max_gas = NonZeroFelt::try_from(Felt::from(max_fee)
+                            .floor_div(&NonZeroFelt::try_from(max_gas_unit_price)?)).expect("Calculated max-gas from provided --max-fee and the current network gas price is zero. Please increase --max-fee to obtain a positive gas amount");
```
ksew1, line 131:
```suggestion
                            .floor_div(&NonZeroFelt::try_from(max_gas_unit_price)?)).context("Calculated max-gas from provided --max-fee and the current network gas price is zero. Please increase --max-fee to obtain a positive gas amount")?;
```

crates/conversions/src/serde/deserialize/deserialize_impl.rs
```diff
@@ -71,6 +71,27 @@ impl CairoDeserialize for bool {
...
 }
 
+impl CairoDeserialize for NonZeroFelt {
+    fn deserialize(reader: &mut BufferReader<'_>) -> BufferReadResult<Self> {
+        let felt = reader.read::<Felt>()?;
+        NonZeroFelt::try_from(felt).map_err(|_| BufferReadError::ParseFailed)
+    }
+}
+
+impl CairoDeserialize for NonZeroU64 {
```
ksew1, line 81: Maybe let's do this with macro for convenience:

```rs
macro_rules! impl_deserialize_for_nonzero_num_type {
    ($type:ty) => {
        impl CairoDeserialize for NonZero<$type> {
            fn deserialize(reader: &mut BufferReader<'_>) -> BufferReadResult<Self> {
                let val = <$type>::deserialize(reader)?;
                NonZero::new(val).ok_or(BufferReadError::ParseFailed)
            }
        }
    };
}
```

crates/sncast/src/helpers/fee.rs
```diff
@@ -92,44 +95,44 @@ impl FeeArgs {
...
-                            max_fee
-                                .floor_div(&NonZeroFelt::from_felt_unchecked(max_gas))
-                                .try_into_()?,
-                        ),
+                            .and_then(|val| NonZeroU128::try_from_(Felt::from(val)).ok()),
                     },
+                    (Some(max_fee), None, Some(max_gas_unit_price)) => {
+                        let max_gas = NonZeroFelt::try_from(Felt::from(max_fee).floor_div(&max_gas_unit_price)).context("Calculated max gas from provided --max-fee and --max-gas-unit-price is zero. Please increase --max-fee to obtain a positive gas amount")?;
+                        FeeSettings::Strk {
+                            max_gas: NonZeroU64::try_from_(Felt::from(max_gas)).ok(),
```
ksew1, line 108: This `.ok()` that you are using is something we really don’t want to do.  
If the conversion fails, we should fail the command and not proceed with the `None` variant.

Please change all occurrences

crates/conversions/src/felt.rs
```diff
@@ -39,6 +42,42 @@ impl FromConv<PaddedFelt> for Felt {
     }
 }
 
+impl TryFromConv<Felt> for NonZeroU64 {
```
ksew1, line 45:
```rs
impl TryFromConv<NonZeroFelt> for NonZeroU64 {
    type Error = String;
    fn try_from_(value: NonZeroFelt) -> Result<Self, Self::Error> {
        let value: u64 = Felt::from(value)
            .try_into()
            .map_err(|_| "felt was too large to fit in u64")?;
        Ok(NonZero::new(value)
            .unwrap_or_else(|| unreachable!("non zero felt is always greater than 0")))
    }
}

impl TryFromConv<NonZeroFelt> for NonZeroU128 {
    type Error = String;
    fn try_from_(value: NonZeroFelt) -> Result<Self, Self::Error> {
        let value: u128 = Felt::from(value)
            .try_into()
            .map_err(|_| "felt was too large to fit in u128")?;
        Ok(NonZero::new(value)
            .unwrap_or_else(|| unreachable!("non zero felt is always greater than 0")))
    }
}
```

crates/sncast/src/helpers/fee.rs
cptartur, line 1: If you implemented `TryFromConv` for types used here, why not use `try_into_()`?

crates/sncast/src/helpers/fee.rs
```diff
@@ -92,44 +95,44 @@ impl FeeArgs {
...
+                            max_gas_unit_price: NonZeroU128::try_from_(Felt::from(
+                                max_gas_unit_price,
+                            ))
+                            .ok(),
+                        }
+                    }
+                    (Some(max_fee), Some(max_gas), None) => {
+                        let max_gas_unit_price = NonZeroFelt::try_from(Felt::from(max_fee).floor_div(&max_gas)).context("Calculated max gas unit price from provided --max-fee and --max-gas is zero. Please increase --max-fee or decrease --max-gas to ensure a positive gas unit price")?;
```
cptartur, line 116:
```suggestion
                        let max_gas_unit_price = NonZeroFelt::try_from(Felt::from(max_fee).floor_div(&max_gas)).context("Calculated max gas unit price from provided --max-fee and --max-gas is 0. Please increase --max-fee or decrease --max-gas to ensure a positive gas unit price")?;
```

I think it makes more sense to use decimal value here and in other places.

crates/sncast/tests/e2e/invoke.rs
cptartur, line 1: The most important cases to test here are the ones where values calculated from `--max-fee` are 0. We should add these.

crates/sncast/tests/integration/fee.rs
```diff
@@ -184,8 +186,8 @@ async fn test_strk_fee_get_max_fee_with_max_gas() {
 
     let args = FeeArgs {
         fee_token: Some(FeeToken::Strk),
-        max_fee: Some(MAX_FEE.into()),
-        max_gas: Some(1_000_000_u32.into()),
+        max_fee: Some(Felt::from(MAX_FEE).try_into().unwrap()),
+        max_gas: Some(Felt::from(1_000_000_u32).try_into().unwrap()),
```
cptartur, line 190: Why are you using `NonZeroU64::new` in some test cases and `Felt::from...` in other cases for the same field? Let's pick a consistent behavior if it's possible.

crates/sncast/src/helpers/fee.rs
cptartur: Or even `= 0` but not sure about that

crates/conversions/src/serde/deserialize/deserialize_impl.rs
cptartur, line 1: cc @Draggu please take a look at these

crates/conversions/src/felt.rs
cptartur, line 1: Any conversions here should be tested, take a look at integration tests for this crate.

CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Cast
+
+### Changed
+
+- Values passed to the `--max-fee`, `--max-gas`, and `--max-gas-unit-price` flags must be greater than 0
```
cptartur, line 14: That's one thing, but it also verifies in case of automatic (implicit) conversion. Please add that, it's more important change to most users.

cptartur (changes requested):

crates/conversions/src/felt.rs
```diff
@@ -39,6 +42,42 @@ impl FromConv<PaddedFelt> for Felt {
...
+    type Error = String;
+    fn try_from_(value: Felt) -> Result<Self, Self::Error> {
+        if value == Felt::ZERO {
+            Err("value should be greater than 0".to_string())
+        } else {
+            let value: u64 = value.try_into().expect("failed to convert Felt to u64");
```
Draggu, line 51: As @ksew1 said ^ don't panic in fallible conversion. It is highly unexpected.

crates/conversions/src/serde/deserialize/deserialize_impl.rs
Draggu: No need as there are only 2 impls but please do it for consistency, there are already macro calls for regular num types.

crates/sncast/src/helpers/fee.rs
Draggu: Good call, use our conversion in all places where possible.

crates/sncast/tests/integration/fee.rs
franciszekjob: It's because fee-related fields in `FeeArgs` are use `NonZeroFelt` while in `FeeSettings::Strk/Eth` they use `NonZeroFelt`, `NonZeroU64` or `NonZeroU128`.

crates/conversions/src/felt.rs
franciszekjob: Added

CHANGELOG.md
ddoktorski: This is not a change, these values must be greater than 0 also before. It is still possible to pass 0, but it will now return a different error than before.

crates/sncast/tests/e2e/invoke.rs
franciszekjob: Fine, so let's break it down:

- [case when max fee and max gas unit price are passed](https://github.com/foundry-rs/starknet-foundry/blob/franciszekjob/2706-validate-fee-args-greater-than-zero/crates/sncast/src/helpers/fee.rs#L100-L118) - calculated `max_gas` will always be >= 1 because of previous check for `max_fee < max_gas_unit_price`, so it's unreachable for it to be equal to 0
- [case when max fee and max gas are passed](https://github.com/foundry-rs/starknet-foundry/blob/franciszekjob/2706-validate-fee-args-greater-than-zero/crates/sncast/src/helpers/fee.rs#L119-L135) - same as above
- [case when only max fee is passed](https://github.com/foundry-rs/starknet-foundry/blob/franciszekjob/2706-validate-fee-args-greater-than-zero/crates/sncast/src/helpers/fee.rs#L136-L155) - it's tested by [test_calculated_max_gas_equal_to_zero_when_max_fee_passed](https://github.com/foundry-rs/starknet-foundry/blob/a784e6987ff3449cc6bff8a10c1dacdf398a20ad/crates/sncast/tests/e2e/invoke.rs#L427).

CHANGELOG.md
franciszekjob: > This is not a change, these values must be greater than 0 also before. It is still possible to pass 0, but it will now return a different error than before.

The difference is that now they are enforced to be > 0 while parsing. I know what you mean that practically just the error message is be different, but I would still leave it for the sake of correctness. Wdyt?

ddoktorski: Ok, let's leave it then

crates/sncast/src/helpers/fee.rs
```diff
@@ -83,53 +85,72 @@ impl FeeArgs {
...
-                            .price_in_fri;
+                    (Some(max_fee), None, Some(max_gas_unit_price)) => {
+                        if max_fee < max_gas_unit_price {
+                            bail!(
+                                "--max-fee should be greater than or equal to --max-gas-unit-price"
+                            );
+                        }
+
+                        let max_gas = NonZeroFelt::try_from(Felt::from(max_fee).floor_div(&max_gas_unit_price))
+                        .unwrap_or_else(|_| unreachable!("Calculated max gas must be >= 1 because max_fee >= max_gas_unit_price ensures a positive result"));
```
cptartur, line 108: Couldn't some rounding errors still make this <1 in special cases?

franciszekjob: No, rounding errors can't make this < 1, because we explicitly ensure that `max_fee >= max_gas_unit_price` before this calculation. Since `floor_div` performs exact integer division (rounding down), and both `max_fee` and `max_gas_unit_price` are non-zero, the result of `floor_div` will always be `>=1`.

crates/sncast/tests/e2e/invoke.rs
```diff
@@ -388,6 +388,77 @@ fn test_too_low_max_fee() {
...
+    ];
+
+    let snapbox = runner(&args);
+    let output = snapbox.assert().success();
+
+    assert_stderr_contains(
+        output,
+        indoc! {r"
+        command: invoke
+        error: Calculated max-gas from provided --max-fee and the current network gas price is 0. Please increase --max-fee to obtain a positive gas amount: Tried to create NonZeroFelt from 0
```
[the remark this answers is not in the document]
franciszekjob: It's because of `context` [here](https://github.com/foundry-rs/starknet-foundry/blob/091e4847834ed93b1239ce604cf8739fddee00d9/crates/sncast/src/helpers/fee.rs#L145).

https://github.com/foundry-rs/starknet-foundry/pull/2796#discussion_r1893824846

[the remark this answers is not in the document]
cptartur: If this can be easily changed let's do so, but if not, please create a follow-up issue.

cptartur (approved):
See my comment, but otherwise looks good

crates/sncast/tests/e2e/invoke.rs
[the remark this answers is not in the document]
franciszekjob: Added #2852

ddoktorski (approved):

PR #2797, opened 2024-12-18
Print max fee conversion info when using v3 txs

franciszekjob:
Closes #

## Introduced changes

-

## Checklist

- [ ] Linked relevant issue
- [ ] Added relevant tests

### DIFF
--- a/crates/sncast/src/helpers/fee.rs
+++ b/crates/sncast/src/helpers/fee.rs
@@ -79,6 +79,9 @@ impl FeeArgs {
                 })
             }
             FeeToken::Strk => {
+                if self.max_fee.is_some() {
+                    print_max_fee_conversion_info();
+                }
                 let settings = match (self.max_fee, self.max_gas, self.max_gas_unit_price) {
                     (Some(_), Some(_), Some(_)) => {
                         bail!("Passing all --max-fee, --max-gas and --max-gas-unit-price is conflicting. Please pass only two of them or less")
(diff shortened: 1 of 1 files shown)


crates/sncast/src/helpers/fee.rs
```diff
@@ -113,14 +116,23 @@ impl FeeArgs {
...
-                        max_gas_unit_price: Some(
-                            max_fee
-                                .floor_div(&NonZeroFelt::from_felt_unchecked(max_gas))
-                                .try_into_()?,
-                        ),
-                    },
+
+                    (Some(max_fee), Some(max_gas), None) => {
+                        let max_gas_unit_price =
+                            max_fee.floor_div(&NonZeroFelt::from_felt_unchecked(max_gas));
```
ksew1, line 122: The unchecked word should signal that it is not something we should use, as handling it incorrectly can lead to undefined behavior. In Rust, unchecked constructs like this are typically marked as `unsafe`. For example, see [`std::num::NonZero`](https://github.com/rust-lang/rust/blob/3378a5e084fa51c2e7f5c3f53b87e8ad89f1c517/library/core/src/num/nonzero.rs#L397). I’m not sure why Jonathan didn’t make it `unsafe`.

Previously, this functionality was clearly incorrect because we didn’t check whether it was zero. Now, we do check it, but the check happens *after* the operation. 

The sane approach would look something like this:

```rust
&NonZeroFelt::try_from(max_gas_unit_price)
    .context("--max-gas calculated from --max-fee should be greater than 0. Please increase --max-fee")?
```

Additionally, please remove the other places where `from_felt_unchecked` is used.

ksew1: Also (probably) the best way to handle this situation is to create our own wrapper type with the `FromStr` trait, which disallows the number `0`. This way, `clap`, while parsing the command, would immediately return an error:

```rs
struct NonZeroFelt(Felt);

impl FromStr for NonZeroFelt {
    type Err = String;

        match s.parse::<Felt>() {
            Ok(felt) => {
                if felt == Felt::ZERO {
                    Err(String::from("value should be greater than 0"))
                } else {
                    Ok(NonZeroFelt(felt))
                }
            }
            Err(_) => Err(String::from("value should be a number")),
        }
    }
}
```

ksew1: Moved here https://github.com/foundry-rs/starknet-foundry/pull/2796 as the code was also moved here

ksew1 (approved):
200w

crates/sncast/src/helpers/fee.rs
```diff
@@ -279,6 +295,12 @@ fn parse_fee_token(s: &str) -> Result<FeeToken, String> {
     Ok(parsed_token)
 }
 
+fn print_max_fee_conversion_info(max_fee: Felt, max_gas: Felt, max_gas_unit_price: Felt) {
```
Draggu, line 298:
```suggestion
fn print_max_fee_conversion_info(max_fee: impl Into<Felt>, max_gas: impl Into<Felt>, max_gas_unit_price: impl Into<Felt>) {
```

PR #2817, opened 2024-12-20
Update test being run in `Run test in debug profile`

franciszekjob:
Closes #

## Introduced changes

`test_happy_case_common_arguments_after_subcommand` which was previously run in `Run test in debug profile` is non-existent, now `test_happy_case_human_readable` is run

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


.github/workflows/ci.yml
```diff
@@ -203,8 +203,6 @@ jobs:
       - uses: software-mansion/setup-universal-sierra-compiler@v1
       - name: Run tests
         run: cargo test --release -p sncast
-      - name: Run test in debug profile
-        run: cargo test -p sncast test_happy_case_common_arguments_after_subcommand
```
cptartur, line 207: Maybe it' make more sense to update it so it runs a test that exists. I think it was added because clap checks (e.g. checking if no arguments conflict with each other) only run in debug profile.

piotmag769: https://docs.rs/clap/latest/clap/?search=debug pretty sure that's not true

ddoktorski (approved):

.github/workflows/ci.yml
cptartur: > https://docs.rs/clap/latest/clap/?search=debug pretty sure that's not true

You can call the debug assert anywhere, but if you don't I believe it only runs in debug profile by default.

piotmag769: Oh you meant debug_assert, okay

kkawula (approved):

PR #2826, opened 2025-01-07
Use `parse_virtual` instead of creating virtual file

### LINKED ISSUE #2357: Use `db.parse_virtual` instead of creating virtual file manually

snforge

### Description

In plugin code, see:

https://github.com/foundry-rs/starknet-foundry/pull/2044/files#r1606384872

ddoktorski:
Closes #2357


maciektr (approved):

franciszekjob (approved):

Draggu (approved):
Please note this won't work with `inline_macro!` if we want them in the future as these require `kind: FileKind::Expr` to parse correctly.

PR #2833, opened 2025-01-09
Refactor `snforge_std` - `typed_checked_cheatcode`

kkawula:
I am happy to announce the first pr from our mailing list! 🔥🔥
Thanks Peter for your contribution

## Introduced changes

-

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


snforge_std/src/_cheatcode.cairo
```diff
@@ -1,29 +1,32 @@
...
+        arr.append_span(output);
 
         panic(arr)
     } else {
-        input
+        output
+    }
+}
+
+pub(crate) fn typed_checked_cheatcode<const selector: felt252, T, +Serde<T>>(
```
cptartur, line 17:
```suggestion
pub(crate) fn typed_cheatcode<const selector: felt252, T, +Serde<T>>(
```

snforge_std/src/_cheatcode.cairo
```diff
@@ -1,29 +1,32 @@
-use core::option::OptionTrait;
-use core::array::ArrayTrait;
-use core::traits::Into;
-use core::array::SpanTrait;
-
-pub(crate) fn handle_cheatcode(input: Span<felt252>) -> Span<felt252> {
-    let first = *input.at(0);
-    let input = input.slice(1, input.len() - 1);
+pub(crate) fn checked_cheatcode<const selector: felt252>(input: Span<felt252>) -> Span<felt252> {
```
cptartur, line 1: Does this need to be pub at all?

snforge_std/src/_cheatcode.cairo
Draggu: Imo `checked_cheatcode`, it handles part of forge panicking logic, not only types.

snforge_std/src/_cheatcode.cairo
```diff
@@ -1,29 +1,32 @@
...
+    let output = starknet::testing::cheatcode::<selector>(input);
+    let first = *output.at(0);
```
ddoktorski, line 3: Is it actually the exit code in case of error or something else? Can we rename it to something more descriptive?

snforge_std/src/_cheatcode.cairo
ddoktorski: Both names seem a bit off to me since they don’t imply any action. Why not use `execute_cheatcode` and `execute_cheatcode_and_deserialize` or something similar?

snforge_std/src/_cheatcode.cairo
kkawula: yes, it is used by multiple modules

ddoktorski (approved):

PR #2834, opened 2025-01-09
Add deprecation info of `version` flag

### LINKED ISSUE #2828: Deprecate versions other than `v3` in sncast commands that accept `--version`
Show deprecation warnings in case someone uses other version of transaction than `v3`. For all kinds of transactions.

In case someone uses `--version v3` show deprecation warning for the whole `--version` flag.

kkawula:
Closes #2828

## Introduced changes

-

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/sncast/tests/e2e/account/deploy.rs
cptartur, line 1: Can we add similar tests for other commands?

crates/sncast/src/helpers/version.rs
```diff
@@ -0,0 +1,34 @@
...
+                    _ => Err(format!(
+                        "Invalid value '{}'. Possible values: {}, {}",
+                        s, var1, var2
+                    )),
+                }
+            }
+        }
+
+        pub fn parse_version(s: &str) -> Result<$enum_name, String> {
+            let deprecation_message = "The '--version' flag is deprecated and will be removed in the future. Version 3 transactions will become the default and only available version.";
```
cptartur, line 26:
```suggestion
            let deprecation_message = "The '--version' flag is deprecated and will be removed in the future. Version 3 will become the only type of transaction available.";
```

How about that?

crates/sncast/src/helpers/version.rs
ksew1, line 1: It think this would be more readable instead of current macro

```rs
use anyhow::Error;
use clap::ValueEnum;
use shared::print::print_as_warning;

const DEPRECATION_MESSAGE: &str = "The '--version' flag is deprecated and will be removed in the future. Version 3 will become the only type of transaction available.";

pub fn parse_version<T: ValueEnum>(s: &str) -> Result<T, String> {
    print_as_warning(&Error::msg(DEPRECATION_MESSAGE));

    let ignore_case = false; // please check if it should be true of false

    T::from_str(s, ignore_case)
}
```

ksew1: Additionally, I think it would be better to change the signature of `print_as_warning` from:

```rust
pub fn print_as_warning(error: &Error)
```

to:

```rust
pub fn print_as_warning(error: impl Display)
```

This way, we wouldn't need to use the awkward workaround `&Error::msg(DEPRECATION_MESSAGE)` and could simply pass `DEPRECATION_MESSAGE` directly 😊

franciszekjob (commented):
1. Shouldn't we update [this message](https://github.com/foundry-rs/starknet-foundry/blob/bddfce703c74d392b4ca20f6cf82651f4699aef6/crates/sncast/src/helpers/fee.rs#L251) then? If someone uses `--token` we say `Use '--version' instead` while version will be deprecated too. I think we can just remove `Use '--version' instead` sentence.
2. [CI](https://github.com/foundry-rs/starknet-foundry/actions/runs/12698724525) is failing 👀 

Otherwise looks fine 👍

crates/sncast/src/helpers/version.rs
kkawula: Cool, it will simplify a lot, also some tests can be removed then

franciszekjob (approved):
Please resolve conflicts in changelog

crates/sncast/src/helpers/fee.rs
```diff
@@ -248,7 +248,8 @@ impl FromStr for FeeToken {
 }
 
 fn parse_fee_token(s: &str) -> Result<FeeToken, String> {
-    let deprecation_message = "Specifying '--fee-token' flag is deprecated and will be removed in the future. Use '--version' instead";
+    let deprecation_message =
+        "Specifying '--fee-token' flag is deprecated and will be removed in the future.";
     print_as_warning(&Error::msg(deprecation_message));
 
     let parsed_token: FeeToken = s.parse()?;
```
ksew1, line 255: Would here also be better to use `FeeToken::from_str(s, true)?;`

kkawula: Probably yes, but it is going to be removed in a future release so I wouldn't pay a lot of attention to it

cptartur (approved):

ksew1 (approved):

PR #2865, opened 2025-01-21
Remove `--version` and `--fee-token` flags

### LINKED ISSUE #2856: Remove deprecated txs versions and tokens

cptartur:
Closes #2856 

## Introduced changes

- Removed `--version` and `--fee-token` flags in sncast
- Removed ETH token support in cast scripts
- Updated and/or removed related documentation

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/sncast/src/starknet_commands/account/create.rs
```diff
@@ -85,7 +85,7 @@ pub async fn create(
         .context("Invalid address")?
         .parse()?;
 
-    let mut message = "Account successfully created. Prefund generated address with at least <max_fee> STRK tokens or an equivalent amount of ETH tokens. It is good to send more in the case of higher demand.".to_string();
+    let mut message = "Account successfully created. Prefund generated address with at least <max_fee> STRK tokens. It is good to send more in the case of higher demand.".to_string();
```
ddoktorski, line 88:
```suggestion
    let mut message = format!("Account successfully created. Prefund generated address with at least {max_fee} STRK tokens. It is good to send more in case of higher demand.");
```
And actually we should either replace `STRK` with `FRI` in the message or divide this value by `10^18`. It could also be renamed to `estimated_fee` or something like that.

crates/sncast/tests/e2e/script/general.rs
```diff
@@ -390,6 +390,7 @@ async fn test_run_script_twice_with_state_file_enabled() {
 }
 
 #[tokio::test]
+// #[ignore]
```
ddoktorski, line 393:
```suggestion
```

crates/sncast/tests/helpers/fixtures.rs
```diff
@@ -537,6 +537,7 @@ pub fn assert_tx_entry_failed(
     status: ScriptTransactionStatus,
     msg_contains: Vec<&str>,
 ) {
+    dbg!(&tx_entry);
```
ddoktorski, line 540:
```suggestion
```

franciszekjob (commented):
[This line](https://github.com/foundry-rs/starknet-foundry/blob/e78f981dbdc586d771bf3f29ce1ba2329e6167b2/docs/src/starknet/account.md?plain=1#L57) should be updated.

Looks like in [this file](https://github.com/foundry-rs/starknet-foundry/blob/2856-remove-deprecated-txs/design_documents/cairo_deployment_scripts.md?plain=1) we should also update mentioning eth to strk?

crates/sncast/src/starknet_commands/account/create.rs
cptartur: Agreed but let me do this as follow-up PR so we don't mix unrelated changes.

cptartur:
> [This line](https://github.com/foundry-rs/starknet-foundry/blob/e78f981dbdc586d771bf3f29ce1ba2329e6167b2/docs/src/starknet/account.md?plain=1#L57) should be updated.
> 
> Looks like in [this file](https://github.com/foundry-rs/starknet-foundry/blob/2856-remove-deprecated-txs/design_documents/cairo_deployment_scripts.md?plain=1) we should also update mentioning eth to strk?

Updated the docs. The second one is a design document, don't think we should be retroactively changing these.

franciszekjob:
> The second one is a design document, don't think we should be retroactively changing these.

Right, didn't notice it was a design doc 👍

ddoktorski (approved):

franciszekjob (approved):

kkawula (approved):

Issue #2051, opened 2024-04-22
New fuzzer architecture


Arcticae:
### Which components does the task require to be changed? (think hard pls)

snforge

### Description

Implement the new fuzzer architecture utilising plugins as outlined in "New fuzzing architecture" design doc.

cptartur:
From #2108 

Which components does the task require to be changed? (think hard pls)
snforge

Description
Based off the conception in the design doc for fuzzing,
we should define the interfaces first and then implement accordingly

This should be a trait, which allows the user to implement the trait on their structure to be fuzzed automatically
It should be similar in concept to the [rust's arbitrary](https://docs.rs/arbitrary/latest/arbitrary/)

PR #2892, opened 2025-01-29
Add `generate_arg` cheatcode

ddoktorski:
Towards #2051

## Introduced changes

- Added a cheatcode for generating random number that is used for creating data for fuzz tests

This PR is part of the stack:

➡️ Add `generate_arg` cheatcode (https://github.com/foundry-rs/starknet-foundry/pull/2892)
-- Add `Fuzzable` trait (https://github.com/foundry-rs/starknet-foundry/pull/2893)
-- Move statements logic to separate function (https://github.com/foundry-rs/starknet-foundry/pull/2894)
-- Add new fuzzer logic to plugin (https://github.com/foundry-rs/starknet-foundry/pull/2895)
-- Add cheatcode to save fuzzer input (https://github.com/foundry-rs/starknet-foundry/pull/2923)
-- Remove old fuzzer logic (https://github.com/foundry-rs/starknet-foundry/pull/2896)
-- Add fuzzer tests with multiple attributes (https://github.com/foundry-rs/starknet-foundry/pull/2898)
-- Update fuzzer documentation and changelog (https://github.com/foundry-rs/starknet-foundry/pull/2899)

### DIFF
--- a/crates/cheatnet/src/runtime_extensions/forge_runtime_extension/fuzzer.rs
+++ b/crates/cheatnet/src/runtime_extensions/forge_runtime_extension/fuzzer.rs
@@ -1,0 +1,38 @@
+use anyhow::ensure;
+use num_bigint::RandBigInt;
+use rand::prelude::StdRng;
+use starknet_types_core::felt::Felt;
+use std::sync::{Arc, Mutex};
+
+pub(crate) fn generate_arg(
+    fuzzer_rng: Option<Arc<Mutex<StdRng>>>,
+    min_value: Felt,
+    max_value: Felt,
+) -> anyhow::Result<Felt> {
+    let min_big_int = if min_value > (Felt::MAX + Felt::from(i128::MIN)) && min_value > max_value {
+        // Negative value x is serialized as P + x, where P is the STARK prime number
+        // hence to deserialize and get the actual x we need to subtract P (== Felt::MAX + 1)
+        min_value.to_bigint() - Felt::MAX.to_bigint() - 1
+    } else {
+        min_value.to_bigint()
+    };
+
+    let max_big_int = max_value.to_bigint();
+
+    ensure!(
+        min_big_int <= max_big_int,
+        format!("`generate_arg` cheatcode: `min_value` must be <= `max_value`, provided values after deserialization: {min_big_int} and {max_big_int}")
+    );
+
+    let value = if let Some(fuzzer_rng) = fuzzer_rng {
+        fuzzer_rng
+            .lock()
+            .unwrap()
+            .gen_bigint_range(&min_big_int, &(max_big_int + 1))
+    } else {
+        // `generate_arg` cheatcode can be also used outside the fuzzer context
+        rand::thread_rng().gen_bigint_range(&min_big_int, &(max_big_int + 1))
+    };
+
+    Ok(Felt::from(value))
+}
--- a/crates/cheatnet/src/runtime_extensions/forge_runtime_extension/fuzzer.rs
+++ b/crates/cheatnet/src/runtime_extensions/forge_runtime_extension/fuzzer.rs
@@ -27,7 +27,7 @@ pub(crate) fn generate_arg(
     let value = if let Some(fuzzer_rng) = fuzzer_rng {
         fuzzer_rng
             .lock()
-            .unwrap()
+            .expect("Failed to acquire lock on fuzzer_rng")
             .gen_bigint_range(&min_big_int, &(max_big_int + 1))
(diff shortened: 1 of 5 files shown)


crates/cheatnet/src/runtime_extensions/forge_runtime_extension/fuzzer.rs
```diff
@@ -0,0 +1,38 @@
...
+    );
+
+    let value = if let Some(fuzzer_rng) = fuzzer_rng {
+        fuzzer_rng
+            .lock()
+            .unwrap()
+            .gen_bigint_range(&min_big_int, &(max_big_int + 1))
+    } else {
+        // `generate_arg` cheatcode can be also used outside the fuzzer context
+        rand::thread_rng().gen_bigint_range(&min_big_int, &(max_big_int + 1))
```
cptartur, line 34: It can, but do we want to allow it? I assume this can limit our ability to change it if we don't treat it as more of an internal thing.

ddoktorski: We can't make it internal because user should be able to implement the `Fuzzable` trait with custom logic and `generate_arg` is needed for this. We can limit its documentation and require it to be imported using the path `snforge_std::fuzzable::generate_arg` to indicate that it is intended for a fuzzing context.

cptartur: Okay, right. I leave the decision up to you, I think only promoting in in the context of fuzzing might be less problematic, but maybe I'm overly concerned.

cptartur (approved):

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/fuzzer.rs
```diff
@@ -0,0 +1,38 @@
...
+
+    ensure!(
+        min_big_int <= max_big_int,
+        format!("`generate_arg` cheatcode: `min_value` must be <= `max_value`, provided values after deserialization: {min_big_int} and {max_big_int}")
+    );
+
+    let value = if let Some(fuzzer_rng) = fuzzer_rng {
+        fuzzer_rng
+            .lock()
+            .unwrap()
```
franciszekjob, line 30: nit:
```suggestion
            .expect("Failed to acquire lock on fuzzer_rng")
```

franciszekjob (approved):

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
```diff
@@ -476,6 +480,14 @@ impl<'a> ExtensionLogic for ForgeExtension<'a> {
             "generate_random_felt" => Ok(CheatcodeHandlingResult::from_serializable(
                 generate_random_felt(),
             )),
+            "generate_arg" => {
+                let min_value: Felt = input_reader.read()?;
+                let max_value: Felt = input_reader.read()?;
```
Draggu, line 485:
```suggestion
                let min_value = input_reader.read()?;
                let max_value = input_reader.read()?;
```

Should be inferable

snforge_std/src/cheatcodes/generate_arg.cairo
```diff
@@ -0,0 +1,8 @@
+use super::super::_cheatcode::execute_cheatcode_and_deserialize;
+
+// Generates a random number that is used for creating data for fuzz tests
+pub fn generate_arg<T, +Serde<T>, +Drop<T>, +Into<T, felt252>>(min_value: T, max_value: T) -> T {
```
Draggu, line 4:
```
pub(crate) fn generate_arg(
    fuzzer_rng: Option<Arc<Mutex<StdRng>>>,
    min_value: Felt,
    max_value: Felt,
) -> anyhow::Result<Felt> {
```
This is always single felt so why do we create an arbitrary type from it? it will fail with any type bigger than 1 felt. Maybe it should return something like `From<T, felt252>` (not sure if it is correct cairo) instead of  `Serde<T>`

ddoktorski: Mostly for convenience, to omit serialization to felt, but also to avoid edge cases such as `i128::MIN` as min and maximum felt value as max. A constraint on the same type prevents this, whereas using felts not. `generate_arg` should operate on a concrete numeric type and this implementation ensures that.

Draggu: Discussed offline

PR #2893, opened 2025-01-29
Add `Fuzzable` trait

ddoktorski:
Towards #2051

## Introduced changes

- Added `Fuzzable` trait with implementation for basic types

This PR is part of the stack:

-- Add `generate_arg` cheatcode (https://github.com/foundry-rs/starknet-foundry/pull/2892)
➡️ Add `Fuzzable` trait (https://github.com/foundry-rs/starknet-foundry/pull/2893)
-- Move statements logic to separate function (https://github.com/foundry-rs/starknet-foundry/pull/2894)
-- Add new fuzzer logic to plugin (https://github.com/foundry-rs/starknet-foundry/pull/2895)
-- Add cheatcode to save fuzzer input (https://github.com/foundry-rs/starknet-foundry/pull/2923)
-- Remove old fuzzer logic (https://github.com/foundry-rs/starknet-foundry/pull/2896)
-- Add fuzzer tests with multiple attributes (https://github.com/foundry-rs/starknet-foundry/pull/2898)
-- Update fuzzer documentation and changelog (https://github.com/foundry-rs/starknet-foundry/pull/2899)

PR #2895, opened 2025-01-29
Add new fuzzer logic

ddoktorski:
Towards #2051

## Introduced changes

- Added new plugin logic for fuzzer attribute
- Split `#[fuzzer]` attribute into `#[__fuzzer_config]` and `#[__fuzzer_wrapper]`
- Modified tests execution for fuzz tests

This PR is part of the stack:

-- Add `generate_arg` cheatcode (https://github.com/foundry-rs/starknet-foundry/pull/2892)
-- Add `Fuzzable` trait (https://github.com/foundry-rs/starknet-foundry/pull/2893)
-- Move statements logic to separate function (https://github.com/foundry-rs/starknet-foundry/pull/2894)
➡️ Add new fuzzer logic to plugin (https://github.com/foundry-rs/starknet-foundry/pull/2895)
-- Add cheatcode to save fuzzer input (https://github.com/foundry-rs/starknet-foundry/pull/2923)
-- Remove old fuzzer logic (https://github.com/foundry-rs/starknet-foundry/pull/2896)
-- Add fuzzer tests with multiple attributes (https://github.com/foundry-rs/starknet-foundry/pull/2898)
-- Update fuzzer documentation and changelog (https://github.com/foundry-rs/starknet-foundry/pull/2899)


crates/snforge-scarb-plugin/src/config_statement.rs
```diff
@@ -82,7 +92,113 @@ pub fn append_config_statements(
...
+            {actual_body_fn_attrs}
+            #[{internal_config_attr}]
+            {function_kw} {actual_body_fn_name}{signature} {{
+                {statements}
+            }}
+        "
+    )
+}
+
+fn extract_and_modify_params<F>(
```
Draggu, line 182: Don't like this one, does to many things, but up 2 you

crates/snforge-scarb-plugin/src/config_statement.rs
```diff
@@ -50,17 +56,21 @@ where
         "
     );
 
-    Ok(append_config_statements(db, func, &config_cheatcode))
+    if Collector::ATTR_NAME == "fuzzer" {
```
Draggu, line 59: It should not be here, to make it clear we can rewrite `fuzzer` into 2 attributes, first being clear config, second doing wrapper function, then we can keep this fn doing single thing. But if you think it is unnecessary overhead it is ok, just leave comment here.

crates/snforge-scarb-plugin/tests/integration/multiple_attributes.rs
```diff
@@ -85,3 +87,185 @@ fn works_with_few_attributes() {
...
+
+                    empty_fn_actual_body(snforge_std::fuzzable::Fuzzable::blank());
+
+                    return;
+                }
+                let f: felt252 = snforge_std::fuzzable::Fuzzable::generate();
+                empty_fn_actual_body(f);
+            }
+            #[__internal_config_statement]
+            fn empty_fn_actual_body(f: felt252) {
```
Draggu, line 267: Not sure if it was in doc, but I think that fn with different name than user-provided one should be wrapper. Let me explain: 
There can be another (not from snforge) macro on user test that uses this name to generate some stuff, so we should not change this name. If we rename wrapper function instead we will get wrong test name on forge side, but this can be easily bypassed with something like `"{actual_name}_SOME_CONST_SUFFIX_FOR_SNFORGE_THAT_IS_UNIQUE"` and then on forge binary side just handle this case and remove suffix

Draggu: Discussed offline

crates/snforge-scarb-plugin/src/config_statement.rs
```diff
@@ -82,7 +92,113 @@ pub fn append_config_statements(
...
+
+    let blank_values_for_config_run =
+        extract_and_modify_params(db, func, |_param| format!("{FUZZABLE_PATH}::blank()"), ", ");
+
+    let params_call = extract_and_modify_params(db, func, |param| param.name(db).as_text(db), ", ");
+
+    let internal_config_attr = InternalConfigStatementCollector::ATTR_NAME;
+    let actual_body_fn_name = format!("{name}_actual_body");
+
+    let (statements, if_content) = get_statements(db, func);
```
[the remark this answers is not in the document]
ddoktorski: As discussed offline, config stays in the original function (if exists). It is used to enable adding blank call after existing config statements.

cptartur (approved):

crates/snforge-scarb-plugin/src/attributes/fuzzer/wrapper.rs
```diff
@@ -0,0 +1,137 @@
...
+            {actual_body_fn_attrs}
+            #[{internal_config_attr}]
+            fn {actual_body_fn_name}{signature} {{
+                {statements}
+            }}
+        "
+    ))
+}
+
+fn extract_and_modify_params<F>(
```
franciszekjob, line 120: nit: Wdyt?
```suggestion
fn extract_and_transform_params<F>(
```

crates/snforge-scarb-plugin/src/utils.rs
```diff
@@ -53,3 +57,71 @@ macro_rules! branch {
...
+            || is_config_run.identifier(db) != "_is_config_run"
+        {
+            return None;
+        }
+
+        let statements = if_expr.if_block(db).statements(db).elements(db);
+
+        // omit last one (`return;`) as it have to be inserted after all new statements
+        Some(
+            statements[..statements.len() - 1]
```
franciszekjob, line 109: If for some reason `statements` will be empty, we will get panic. Maybe let's use [`saturating_sub(1)`](https://doc.rust-lang.org/std/primitive.u32.html#method.saturating_sub)? Or ensure they aren't empty?

crates/snforge-scarb-plugin/tests/integration/single_attributes/fuzzer.rs
```diff
@@ -142,6 +382,21 @@ fn fail_with_invalid_args() {
...
+fn fail_with_unnamed_arg() {
+    let item = TokenStream::new(EMPTY_FN.into());
+    let args = TokenStream::new("(123)".into());
+
+    let result = fuzzer_config(args, item);
+
+    assert_diagnostics(
+        &result,
+        &[Diagnostic::error(
+            "#[fuzzer] can be used with named attributes only",
```
franciszekjob, line 395: Shouldn't the error message be `"...with named args only"`?

crates/snforge-scarb-plugin/src/utils.rs
ddoktorski: It cannot be empty as there is always at least a `return` statement. Also this is not new logic, it was only moved to a different place.

crates/snforge-scarb-plugin/tests/integration/single_attributes/fuzzer.rs
ddoktorski: Good point, @Draggu was it intentional to use `attributes` instead of `arguments` in these errors?

Draggu: If it is like this in other places it is a typo

ddoktorski: @franciszekjob 
I will update these messages as part of rewriting the plugin to the new macro syntax https://github.com/foundry-rs/starknet-foundry/issues/2944

franciszekjob (approved):

PR #2897, opened 2025-01-29
Refactor `DictStateReader`

### LINKED ISSUE #2891: Refactor `DictStateReader`
### Current State

ATM, [`DictStateReader`](https://github.com/foundry-rs/starknet-foundry/blob/a47c6a4262813a3bb185c50f7fa9e8747be9c3e9/crates/runtime/src/starknet/state.rs#L13) has a couple fields which are never modified.

In scope of this task we should:
- remove `storage_view`, `address_to_nonce`, `class_hash_to_compiled_class_hash` fields
- adjust methods in `DictStateReader` which read these fields, so they always return error

### Objective

`get_storage_at`, `get_nonce_at`, `get_compiled_class_hash` in `DictStateReader` should always return error

### Additional Context

_No response_

franciszekjob:
Closes #2891 

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


ddoktorski (approved):

crates/runtime/src/starknet/state.rs
```diff
@@ -57,11 +47,6 @@ impl StateReader for DictStateReader {
     }
 
     fn get_compiled_class_hash(&self, class_hash: ClassHash) -> StateResult<CompiledClassHash> {
-        let compiled_class_hash = self
-            .class_hash_to_compiled_class_hash
-            .get(&class_hash)
-            .copied()
-            .unwrap_or_default();
-        Ok(compiled_class_hash)
+        Err(StateError::UndeclaredClassHash(class_hash))
```
cptartur, line 50: I'd prefer we throw an error like above here so it's immediately obvious where it's coming from. It could simplify some debugging in the future.

cptartur (approved):

PR #2905, opened 2025-01-31
sncast argument argument parsing logic

### LINKED ISSUE #2760: Always use data transformer unless `--calldata` is passed

emarc99:
- Closes #2760 

## Introduced changes

-

## Checklist

- [ ] Linked relevant issue
- [ ] Added relevant tests

### DIFF
--- a/crates/sncast/src/main.rs
+++ b/crates/sncast/src/main.rs
@@ -169,9 +169,7 @@ impl Arguments {
         contract_class: ContractClass,
         selector: &Felt,
     ) -> Result<Vec<Felt>> {
-        if let Some(arguments) = self.arguments {
-            Calldata::new(arguments).serialized(contract_class, selector)
-        } else if let Some(calldata) = self.calldata {
+        if let Some(calldata) = self.calldata {
             calldata
                 .iter()
                 .map(|data| {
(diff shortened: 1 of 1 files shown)


cptartur:
@emarc99 please add tests verifying the behavior introduced by your changes.

emarc99:
On it right now!

cptartur:
Closing due to inactivity

PR #2920, opened 2025-02-05
Update LICENSE

maximevtush:
Updated the copyright year to 2025 in the LICENSE file


ddoktorski:
@maximevtush
Thanks for the PR. However, the change does not seem necessary and would require updating this file in the future, so we decided to leave it as it is.

PR #2940, opened 2025-02-13
Panic data regexes tests

### LINKED ISSUE #2931: Panic data regexes
### Current State

`crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/panic_data.rs` file has regexes for extracting panic data.

### Objective

Verify and improve regexes in this file and add unit tests

### Additional Context

The existing unit tests only test strings `""` but not extracting from shortstrings `''`.

The regexes in current version are working but maybe can be written in a better manner.

kkawula:
Closes #2931

## Introduced changes

-

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


franciszekjob (approved):
image

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/panic_data.rs
```diff
@@ -66,29 +64,32 @@ mod test {
...
+                    "
+                ),
+                    Some(vec![Felt::from(344_693_033_291_u64), Felt::from(293_154_149_441_u64)]); "two felts")]
+    #[test_case(indoc!(r"
+                    Got an exception while executing a hint: Execution failed. Failure reason:
+                    Error in contract (contract address: 0x03cda836debfed3f83aa981d7a31733da3ae4f903dde9d833509d2f985d52241, class hash: 0x07ca8b953cb041ee517951d34880631e537682103870b9b018a7b493363b9b63, selector: 0x00a4695e9e8c278609a8e9362d5abe9852a904da970c7de84f0456c777d21137):
+                    0x434d3232 ('AYY, LMAO').
+                    "
+                ),
+                    Some(vec![Felt::from(4_282_713_u64), Felt::from(1_280_131_407_u64)]); "also two felts")]
```
ddoktorski, line 82: What is the reason for having two tests that essentially verify the same case?

ddoktorski (commented):
I don't have much context about this change. As I understand the current regexes work as expected and do not require any modifications. Is that correct? @kkawula

cptartur (changes requested):

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/panic_data.rs
kkawula: It can be removed

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/panic_data.rs
```diff
@@ -66,29 +61,35 @@ mod test {
...
+                    End of trace
+                    "
+                ),
+                    Some(vec![Felt::from(344_693_033_283_u64)]); "felt")]
+    #[test_case(indoc!(r"
+                    Beginning of trace
+                    Got an exception while executing a hint: Execution failed. Failure reason:
+                    Error in contract (contract address: 0x03cda836debfed3f83aa981d7a31733da3ae4f903dde9d833509d2f985d52241, class hash: 0x07ca8b953cb041ee517951d34880631e537682103870b9b018a7b493363b9b63, selector: 0x00a4695e9e8c278609a8e9362d5abe9852a904da970c7de84f0456c777d21137):
+                    (0x54687265652073616420746967657273206174652077686561742e2054776f ('Three sad tigers ate wheat. Two'), 0x2074696765727320776572652066756c6c2e20546865206f74686572207469 (' tigers were full. The other ti'),
+                    0x676572206e6f7420736f206d75636800000000000000000000000000000000 ('ger not so much')).
```
kkawula, line 78: based on https://github.com/starkware-libs/sequencer/blob/8c1b65fd037b83c4e5c65c95ae0e28b925b0eaa9/crates/blockifier/src/execution/native/utils_test.rs#L47-L53

kkawula:
image

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/panic_data.rs
```diff
@@ -66,29 +61,35 @@ mod test {
...
-    //         ),
-    //         ("Custom Hint Error: Invalid trace: \"PANIC, DATA\"", None)
-    //     ];
-    //
-    //     for (str, expected) in cases {
-    //         assert_eq!(try_extract_panic_data(str), expected);
-    //     }
-    // }
+    #[test_case(indoc!(r"
+                    Beginning of trace
```
cptartur, line 65: Are the errors still wrapped in `Beginning of trace`, `End of trace`? I don't think this is longer a case.

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/panic_data.rs
```diff
@@ -1,14 +1,13 @@
...
 #[must_use]
 pub fn try_extract_panic_data(err: &str) -> Option<Vec<Felt>> {
-    // FIXME add support for comma separated errors
-    let re_felt_array = Regex::new(
-        r"[\s\S]*Execution failed\. Failure reason:\nError in contract \(.+\):\n.*\('([\s\S]*)'\).",
-    )
-    .expect("Could not create felt panic_data matching regex");
+    // Matches panic data formatted with:
+    // https://github.com/starkware-libs/sequencer/blob/8211fbf1e2660884c4a9e67ddd93680495afde12/crates/starknet_api/src/execution_utils.rs
+    let re_felt_array = Regex::new(r"(0x[a-fA-F0-9]+)\s\('[^']*'\)")
```
cptartur, line 9: I like the simplification, however I think we should still be catching the `Got an exception while executing a hint...` prefix. Otherwise we are risking mapping any error containing parsable felts to our format, which I think could be problematic in some weird cases.

cptartur, line 9: This doesn't work in cases where the returned string cannot be parsed to shortstring. If I understand the linked function correctly, it can return something like `0x123 ('shortstring1'), 0x456, 0x789 ('shortstring2').

See this, it could partially resolve the problem https://regex101.com/r/MWpuSJ/1 (note part before regex is matched as well, but my comment above could resolve it).

kkawula: > it can return something like `0x123 ('shortstring1'), 0x456, 0x789 ('shortstring2').

It isn't true, function used to format the error `format_panic_data` uses fn `format_next_item` to parse felts that fits into shortstring otherwise its returns `None`, therefore our formatted panic data always contain felts that can be parsed into shortstring

`format_next_item` description:

> Formats a string or a short string / felt252. Returns the formatted string and a boolean indicating whether it's a string. If can't format the item, returns None.

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/panic_data.rs
```diff
@@ -1,46 +1,48 @@
-use conversions::{byte_array::ByteArray, felt::FromShortString};
+use conversions::byte_array::ByteArray;
 use regex::Regex;
 use starknet_types_core::felt::Felt;
 
 #[must_use]
 pub fn try_extract_panic_data(err: &str) -> Option<Vec<Felt>> {
-    // FIXME add support for comma separated errors
+    // Regex used to extract panic data from panicking proxy contract
     let re_felt_array = Regex::new(
```
cptartur, line 8: ? Or something similar?

```suggestion
    let re_error_prefix = Regex::new(
```

cptartur (approved):
✅ 
Let me merge it manually into bump branch.

Issue #2807, opened 2024-12-20
Handle missing requirements in snforge compatibility check


cptartur:
Currently, snforge compatibility check displays nice output only when incorrect version of a requirement is installed. In case it is missing entirely, `Failed to run version command for {name}` error is bubbled.

Update compatibility check so it displays missing requirements similarly as incompatible. 
E.g. `"❌ {} is not installed"` or similar.

PR #2945, opened 2025-02-14
Fix required tools validation

### LINKED ISSUE #2947: Do not always require Rust in compatibility check
### Current State

Currently, we always check if USC, Scarb and Rust are installed when running forge.
Rust is no longer needed if using scarb supporting prebuilt packages in registry.

### Objective

Revamp rust checking logic

### Additional Context

If user is on the Scarb version that supports prebuilt packages we shouldn't require Rust to be installed.

One concern is that if user doesn't opt-in into prebuilt packages, then Rust will be required even on Scarb versions that support them - this should be handled somehow.

ddoktorski:
## Introduced changes

- Run `check_requirements` only for `snforge test` command
- Require Rust installed only for Scarb `<2.10.0`

### DIFF
--- a/crates/forge/src/lib.rs
+++ b/crates/forge/src/lib.rs
@@ -294,7 +294,7 @@ fn check_requirements(output_on_success: bool) -> Result<()> {
     requirements_checker.check()?;
 
     let scarb_version = ScarbCommand::version().run()?.scarb;
-    if scarb_version <= MINIMAL_SCARB_VERSION_PREBUILT_PLUGIN {
+    if scarb_version < MINIMAL_SCARB_VERSION_PREBUILT_PLUGIN {
         let mut requirements_checker = RequirementsChecker::new(output_on_success);
         requirements_checker.add_requirement(Requirement {
             name: "Rust".to_string(),
--- a/crates/forge/src/lib.rs
+++ b/crates/forge/src/lib.rs
@@ -33,6 +33,7 @@ pub const CAIRO_EDITION: &str = "2024_07";
 
 const MINIMAL_RUST_VERSION: Version = Version::new(1, 80, 1);
 const MINIMAL_SCARB_VERSION: Version = Version::new(2, 7, 0);
+const MINIMAL_SCARB_VERSION_PREBUILT_PLUGIN: Version = Version::new(2, 10, 0);
 const MINIMAL_USC_VERSION: Version = Version::new(2, 0, 0);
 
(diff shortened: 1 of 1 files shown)


kkawula (approved):

crates/forge/src/lib.rs
```diff
@@ -302,5 +291,29 @@ fn check_requirements(output_on_success: bool) -> Result<()> {
             r"universal-sierra-compiler (?<version>[0-9]+.[0-9]+.[0-9]+)",
         ),
     });
-    requirements_checker.check()
+    requirements_checker.check()?;
+
+    let scarb_version = ScarbCommand::version().run()?.scarb;
+    if scarb_version <= MINIMAL_SCARB_VERSION_PREBUILT_PLUGIN {
```
franciszekjob, line 297: Shouldn't that be `<`?
```suggestion
    if scarb_version < MINIMAL_SCARB_VERSION_PREBUILT_PLUGIN {
```

ddoktorski: Yes absolutely

franciszekjob (approved):

crates/forge/src/lib.rs
```diff
@@ -302,5 +291,29 @@ fn check_requirements(output_on_success: bool) -> Result<()> {
...
+                r"rustc (?<version>[0-9]+.[0-9]+.[0-9]+)",
+            ),
+            helper_text: "Follow instructions from https://www.rust-lang.org/tools/install"
+                .to_string(),
+        });
+
+        requirements_checker.check()?;
+    }
+
+    Ok(())
```
cptartur, line 318: Can't we only conditionally add the rust requirement, but run `requirements_checker.check()?;` only once?

ddoktorski: It seems more accurate to first check if `Scarb` is installed and then check the version, otherwise the command will fail when executing `ScarbCommand::version().run()?`. Tbh I just realized that we don't print a formatted message when a given tool is not installed, instead we return an error. I think we should change this, but maybe not in this PR.

cptartur: Right, didn't catch that.

> Tbh I just realized that we don't print a formatted message when a given tool is not installed, instead we return an error. I think we should change this, but maybe not in this PR.

Correct, this needs to be addressed. I've created #2807 sometime ago.

cptartur (approved):

Issue #2850, opened 2025-01-14
Different templates when using `snforge new`


cptartur:
Add different templates than the one we currently use when running `snforge new`.
E.g. erc20 template, pure cairo template.

Starkware also suggested showing components, maybe workspaces.

PR #2952, opened 2025-02-17
Add Cairo program template

### LINKED ISSUE #2841: Add a starknet-foundry template that is not for Starknet Development

Other (Please describe below)

### Feature Request

Starknet Foundry will be the de-factor dev-framework even for pure cairo programs due to integrations with Cairo Profiler, Cairo Coverage, etc. However, users and newcomers, when developing in pure cairo, are a bit disoriented at the pre-generated Scarb.toml when starting a project with Starknet Foundry.

As such, I propose adding a new template for pure cairo programs, that would only contain the basic properties for pure Cairo.

ddoktorski:
Closes #2841
Towards #2850

## Introduced changes

- Added a template for simple Cairo program - the same that is created by `scarb new` for cairo test runner
- Renamed templates directory `starknet_forge_template` -> `snforge_templates`
- Fixed `assert_macros` version set in a template

## Checklist

- [X] Linked relevant issue
- [X] Added relevant tests


crates/forge/src/new.rs
```diff
@@ -86,16 +87,19 @@ fn replace_project_name(contents: &[u8], project_name: &str) -> Result<Vec<u8>>
     Ok(contents.into_bytes())
 }
 
-fn update_config(config_path: &Path, scarb: &Version) -> Result<()> {
+fn update_config(config_path: &Path, template: &Template) -> Result<()> {
```
kkawula, line 90: apart from these changes maybe ` fn update_scarb_manifest` would better fit here

crates/forge/src/new.rs
```diff
@@ -56,25 +56,26 @@ fn add_template_to_scarb_manifest(path: &PathBuf) -> Result<()> {
     Ok(())
 }
 
-fn overwrite_files_from_scarb_template(
-    dir_to_overwrite: &str,
-    base_path: &Path,
+fn overwrite_or_copy_files(
```
cptartur, line 59: Nit, but maybe we somehow include the word `template` in the function name

crates/forge/src/new.rs
```diff
@@ -232,41 +238,63 @@ pub fn new(
...
+        add_dependency(project_path, "snforge_std", snforge_version, true)?;
     }
 
-    ScarbCommand::new_with_stdio()
-        .current_dir(&project_path)
+    match template {
+        Template::BalanceContract => {
+            add_dependency(project_path, "starknet", &cairo_version.to_string(), false)?;
+        }
+        Template::CairoProgram => {}
```
cptartur, line 270: I think this might make it hard to maintain eventually if we want to have multiple templates. What if the deps etc. are defined along with the template enum or in some other generic manner. So we can have generic logic for adding those.

You are kind of there already.

cptartur: Depending on how much effort would be needed to develop this, it might be worth it.

crates/forge/src/new.rs
```diff
@@ -271,6 +294,44 @@ pub fn new(
     Ok(())
 }
 
+fn add_dependencies(
```
franciszekjob, line 297: Wouldn't it be better to put is as method of `TemplateManifestConfig`?

crates/forge/src/new.rs
```diff
@@ -271,6 +294,44 @@ pub fn new(
...
+    }
+
+    for dep in &template_config.dependencies {
+        add_dependency(scarb_manifest_path, dep)?;
+    }
+
+    Ok(())
+}
+
+fn add_dependency(scarb_manifest_path: &PathBuf, dependency: &Dependency) -> Result<()> {
```
franciszekjob, line 320: Same as above but for `Dependency`

franciszekjob (commented):
Let's also mention that somewhere in the forge guide.

kkawula (commented):
`snforge new` fails, is it expected?

```
Created package.
    Removing cairo_test from dev-dependencies
error: cannot get dependencies of `my_project@0.1.0`

Caused by:
    cannot find package `assert_macros ^2.11.3`
[ERROR] Failed to fetch created project: `scarb` exited with error
```

franciszekjob:
> `snforge new` fails, is it expected?
> 
> ```
> Created package.
>     Removing cairo_test from dev-dependencies
> error: cannot get dependencies of `my_project@0.1.0`
> 
> Caused by:
>     cannot find package `assert_macros ^2.11.3`
> [ERROR] Failed to fetch created project: `scarb` exited with error
> ```

For me it works, both on scarb 2.8.5 and 2.10.0. Which version are you using @kkawula ? Are you sure you've pulled last changes?

ddoktorski:
> For me it works, both on scarb 2.8.5 and 2.10.0. Which version are you using @kkawula ? Are you sure you've pulled last changes?

It is due to setting assert macros to Scarb version instead of Cairo version, the problem occurs only for Scarb `2.11.3`. Will be fixed soon.

crates/forge/src/new.rs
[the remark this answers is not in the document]
ddoktorski: Will try to do it in a follow-up PR

docs/src/getting-started/first-steps.md
```diff
@@ -9,6 +9,11 @@ To start a new project with Starknet Foundry, run `snforge new`
 $ snforge new hello_starknet
 ```
 
+> 📝 **Note**
+>
+> By default, `snforge new` creates a project with a simple `HelloStarknet` contract. You can create a different project using the `--template` flag. 
+> To see the list of available templates, refer to the [snforge new documentation](../appendix/snforge/new.md#-t---template)
```
franciszekjob, line 15: nit:
```suggestion
> To see the list of available templates, refer to the [`snforge new` documentation](../appendix/snforge/new.md#-t---template)
```

franciszekjob (approved):

crates/forge/src/new.rs
```diff
@@ -56,25 +89,26 @@ fn add_template_to_scarb_manifest(path: &PathBuf) -> Result<()> {
     Ok(())
 }
 
-fn overwrite_files_from_scarb_template(
-    dir_to_overwrite: &str,
-    base_path: &Path,
+fn overwrite_or_copy_template_files(
+    dir: &Dir,
+    template_path: &Path,
```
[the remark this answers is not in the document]
cptartur: I think we can keep it. It makes sense now.

cptartur (approved):

PR #3003, opened 2025-02-26
Fix forge scheduled tests

### LINKED ISSUE #2998: Fix scheduled tests for 2.10.0 and rest
### Current State

Some of our tests aren't working in 2.10.0 and need updating.

### Objective

Fix scheduled tests for 2.10.0 and rest

### Additional Context

Should be mostly changing some cairo code and making sure that dependencies like cairo coverage and profiler are up to date.

ddoktorski:
Closes #2998 

## Introduced changes

- Fix scheduled tests except cast tests, which are blocked by devnet
- Add feature to assert exact gas values in CI and to assert with a margin in scheduled tests
- Bump `cairo-annotations`


crates/forge/test_utils/src/runner.rs
```diff
@@ -288,7 +288,7 @@ pub fn assert_gas(result: &[TestTargetSummary], test_case_name: &str, asserted_g
             }
             AnyTestCaseSummary::Single(case) => match case {
                 TestCaseSummary::Passed { gas_info: gas, .. } => {
-                    *gas == asserted_gas
+                    asserted_gas.abs_diff(*gas) <= 1
```
cptartur, line 291: Can we add a different `assert_gas` method that uses diff and choose which one to use according to the feature? Then the scheduled tests on version other than scarb from `.tool-version` could use it.

E.g.

```rust
    if cfg!(feature = "non_exact_gas") {
        assert_gas_range(...) 
    else {
        assert_gas(...)
     }
```

I'd prefer if we still kept exact gas assertions when developing code. I know difference of `1` is not that big but we should assert exact values here.

crates/forge/tests/data/steps/src/lib.cairo
cptartur, line 1: I guess one thing we are losing with these changes is that we no longer assert if the max steps is sufficiently close to the default of 10m. Maybe with some logic changes we could still retain this check?

cptartur: Or we could hide this check behind a feature as well, what do you think

crates/forge/tests/integration/fuzzing.rs
```diff
@@ -76,7 +76,7 @@ fn fuzzed_while_loop() {
 
     // TODO (#2926)
     assert_eq!(gas_info.min, 2);
-    assert_eq!(gas_info.max, 23);
+    assert!(gas_info.max.abs_diff(23) <= 1);
```
cptartur, line 79: Same here

crates/forge/tests/data/steps/src/lib.cairo
ddoktorski: This has already been changed a few times and it is unmaintainable. Unfortunately, I don't have any idea for a test that will have 10m steps and for sure remain unchanged with future Cairo versions.

crates/forge/test_utils/src/runner.rs
ddoktorski: As discussed offline, for now leaving it as it is

crates/forge/tests/data/steps/src/lib.cairo
ddoktorski: I modified those tests to get the following steps values:
* Scarb `2.8.5`: `8250085` and `10200085`
* Scarb `2.10.1`: `9350088` and `11560088`

It seems to be close enough to test the desired behavior (failing for >10m steps - tested in `should_default_to_10m` test) and hopefully it won't need to be updated with the new Scarb version 🙏

cptartur: Okay, we can do that

crates/forge/test_utils/src/runner.rs
ddoktorski: I eventually added a logic similar to what you suggested:
* On CI we assert exact values
* In scheduled tests we allow a margin (for now arbitrary set to 10)

This applies only to tests from `integration::gas`.

ddoktorski:
Please note that Scarb version `2.11.2` does not exist yet, but for some reason `asdf` finds it when running `asdf list all scarb`.

Cargo.toml
```diff
@@ -45,7 +45,7 @@ cairo-lang-starknet-classes = "2.10.0"
 cairo-lang-parser = "2.10.0"
 cairo-lang-macro = "0.1.0"
 cairo-vm = "1.0.2"
-cairo-annotations = "0.2.2"
+cairo-annotations = "=0.3.0"
```
cptartur, line 48: Why was this bumped, doesn't blockifier still use `0.2.2`?

crates/forge/test_utils/src/runner.rs
```diff
@@ -300,6 +300,16 @@ pub fn assert_gas(result: &[TestTargetSummary], test_case_name: &str, asserted_g
     }));
 }
 
+// This logic is used to assert exact gas values in CI for the minimal supported Scarb version
+// and to assert gas values with a margin in scheduled tests, as values can vary for different Scarb versions
```
cptartur, line 304: Let's add something "scary" to it like `FOR LOCAL DEVELOPMENT ALWAYS USE EXACT CALCULATIONS`

crates/forge/tests/e2e/running.rs
```diff
@@ -782,6 +782,7 @@ fn incompatible_snforge_std_version_warning() {
         .parse::<DocumentMut>()
         .unwrap();
     scarb_toml["dev-dependencies"]["snforge_std"] = value("0.34.1");
+    scarb_toml["dev-dependencies"]["snforge_scarb_plugin"] = value("0.34.1");
```
cptartur, line 785: After we add proper tests for plugin reexport we might want to revisit this. And use version that supports reexports here.

Please link issue #3069 here

Cargo.toml
ddoktorski: 1) Blockifier doesn't have a dependency on `cairo-annotations`
2) There is a new field for code locations, which causes serialization to fail, effectively `SNFORGE_BACKTRACE` does not work with Scarb `>=2.11.0`

Anyway, this is also bumped in one of the PRs for the new gas.

cptartur: Please add an issue to change this to TODO

cptartur (approved):

crates/forge/test_utils/src/runner.rs
```diff
@@ -300,6 +300,17 @@ pub fn assert_gas(result: &[TestTargetSummary], test_case_name: &str, asserted_g
...
+// FOR LOCAL DEVELOPMENT ALWAYS USE EXACT CALCULATIONS
+fn assert_gas_with_margin(gas: u128, asserted_gas: u128) -> bool {
+    if cfg!(feature = "assert_non_exact_gas") {
+        asserted_gas.abs_diff(gas) <= 10
```
franciszekjob, line 308: Is the value of 10 supported by any specific calculations or is it just what we chose as a sensible value?

ddoktorski: Nope, this is chosen arbitrarily

franciszekjob (approved):

kkawula (approved):

Issue #3005, opened 2025-02-26
Pretty print current calltrace


ksew1:
### Which component is your feature related to?

Forge

### Feature Request

We want similar ui to `eth froundry`

PR #3032, opened 2025-03-03
Add debugging crate

ksew1:
Towards #3005

---

**Stack**:
- #3066
- #3065
- #3034
- #3033
- #3032 ⬅

### DIFF
--- a/crates/debugging/src/trace.rs
+++ b/crates/debugging/src/trace.rs
@@ -1,0 +1,68 @@
+use crate::tree::TreeSerialize;
+use blockifier::execution::entry_point::CallType;
+use cheatnet::runtime_extensions::call_to_blockifier_runtime_extension::rpc::CallResult;
+use cheatnet::state::{CallTrace, CallTraceNode};
+use starknet_api::contract_class::EntryPointType;
+use starknet_api::core::{ContractAddress, EntryPointSelector};
+use starknet_api::transaction::fields::Calldata;
+use std::cell::RefCell;
+use std::fmt;
+use std::fmt::Display;
+use std::rc::Rc;
+
+#[derive(Debug, Clone)]
+pub struct Trace {
+    pub selector: EntryPointSelector,
+    pub trace_info: TraceInfo,
+}
+
+#[derive(Debug, Clone)]
+pub struct TraceInfo {
+    pub entry_point_type: EntryPointType,
+    pub calldata: Calldata,
+    pub storage_address: StorageAddress,
+    pub caller_address: CallerAddress,
+    pub call_type: CallType,
+    pub nested_calls: Vec<Trace>,
+    pub call_result: CallResult,
+}
+
+#[derive(Debug, Clone)]
+pub struct StorageAddress(pub ContractAddress);
+
+#[derive(Debug, Clone)]
+pub struct CallerAddress(pub ContractAddress);
+
+impl Trace {
+    /// Creates a new [`Trace`] from a given `cheatnet` [`CallTrace`].
+    pub fn from_call_trace(call_trace: &Rc<RefCell<CallTrace>>) -> Self {
+        let call_trace = call_trace.borrow();
+        let nested_calls = call_trace
+            .nested_calls
+            .iter()
+            .filter_map(CallTraceNode::extract_entry_point_call)
+            .map(Self::from_call_trace)
+            .collect();
+
+        let trace_info = TraceInfo {
+            entry_point_type: call_trace.entry_point.entry_point_type,
+            calldata: call_trace.entry_point.calldata.clone(),
+            storage_address: StorageAddress(call_trace.entry_point.storage_address),
+            caller_address: CallerAddress(call_trace.entry_point.caller_address),
+            call_type: call_trace.entry_point.call_type,
+            nested_calls,
+            call_result: call_trace.result.clone(),
+        };
+
+        Self {
+            selector: call_trace.entry_point.entry_point_selector,
+            trace_info,
+        }
+    }
+}
+
+impl Display for Trace {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}", self.serialize())
+    }
+}
--- a/crates/debugging/src/tree/building/builder.rs
+++ b/crates/debugging/src/tree/building/builder.rs
@@ -1,0 +1,73 @@
+use ptree::TreeBuilder;
+use ptree::item::StringItem;
+
+/// A wrapper around [`TreeBuilder`] that adds two guard nodes to the tree upon creation and removes them upon building.
+pub struct TreeBuilderWithGuard {
+    builder: TreeBuilder,
+}
+
+impl TreeBuilderWithGuard {
+    /// Creates a new [`TreeBuilderWithGuard`].
+    pub fn new() -> Self {
+        let mut builder = TreeBuilder::new("guard".to_string());
+        builder.begin_child("guard".to_string());
+        Self { builder }
+    }
+
+    /// Add a child to the current item and make the new child current.
+    pub fn begin_child(&mut self, text: String) {
+        self.builder.begin_child(text);
+    }
+
+    /// Add an empty child (leaf item) to the current item.
+    pub fn add_empty_child(&mut self, text: String) {
+        self.builder.add_empty_child(text);
+    }
+
+    /// Finish adding children, and make the current item's parent current.
+    pub fn end_child(&mut self) {
+        self.builder.end_child();
+    }
+
+    /// Finish building the tree and return the top level item, not accounting for the guard nodes.
+    pub fn build(mut self) -> StringItem {
+        let string_item = self.builder.build();
+        extract_guard(extract_guard(string_item))
+    }
+}
+
+/// Extracts the guard node from a [`StringItem`].
+fn extract_guard(mut string_item: StringItem) -> StringItem {
+    assert_eq!(string_item.text, "guard");
+    assert_eq!(string_item.children.len(), 1);
+    string_item.children.pop().unwrap_or_else(|| {
+        unreachable!("guard is guaranteed to have one child by the assertion above")
+    })
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_happy_path() {
+        let mut builder = TreeBuilderWithGuard::new();
+        builder.add_empty_child("leaf".to_string());
+
+        let tree = builder.build();
+
+        assert_eq!(tree.text, "leaf");
+        assert!(tree.children.is_empty());
+    }
+
+    #[test]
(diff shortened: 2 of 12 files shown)


crates/debugging/src/trace.rs
```diff
@@ -0,0 +1,81 @@
...
+    pub call_type: CallType,
+    pub nested_calls: Vec<Trace>,
+    pub call_result: CallResult,
+}
+
+/// Wrapper for [`ContractAddress`] to distinguish storage addresses at the type level.
+#[derive(Debug, Clone)]
+pub struct StorageAddress {
+    pub address: ContractAddress,
+}
```
cptartur, line 34: Nit but maybe let's just use this if it's just a wrapper. Don't have a strong opinion on it though.

```suggestion
pub struct StorageAddress(ContractAddress)
```

crates/debugging/src/trace.rs
```diff
@@ -0,0 +1,81 @@
...
+
+impl Trace {
+    /// Creates a new [`Trace`] from a given `cheatnet` [`CallTrace`].
+    pub fn from_call_trace(call_trace: &Rc<RefCell<CallTrace>>) -> Self {
+        let call_trace = call_trace.borrow();
+        let nested_calls = call_trace
+            .nested_calls
+            .iter()
+            .filter_map(CallTraceNode::extract_entry_point_call)
```
cptartur, line 49: Why are we filtering deploys from this?

crates/debugging/src/tree/display.rs
```diff
@@ -0,0 +1,97 @@
...
+        match self {
+            CallResult::Success { ret_data } => {
+                format!("success: {ret_data:?}")
+            }
+            CallResult::Failure(call_failure) => match call_failure {
+                CallFailure::Panic { panic_data } => {
+                    format!("panic: {panic_data:?}")
+                }
+                CallFailure::Error { msg } => {
+                    format!("error: {msg}")
```
cptartur, line 80: Should we color these as well?

crates/debugging/src/tree/display.rs
```diff
@@ -0,0 +1,97 @@
...
+/// of a type that can be converted to a [`Felt`].
+fn string_hex(data: impl Into<Felt>) -> String {
+    data.into().to_hex_string()
+}
+
+/// Helper function to get debug representation of a type as a string.
+/// Mainly used for enums that hold no data or vectors of felts.
+fn string_debug(data: impl Debug) -> String {
+    format!("{data:?}")
+}
```
cptartur, line 97: Just let's make sure we don't export these to other modules

crates/debugging/src/tree/node.rs
```diff
@@ -0,0 +1,80 @@
...
+/// Abstraction over a [`TreeBuilder`] that automatically ends building a current node when dropped.
+pub struct Node<'a> {
+    pub builder: &'a mut TreeBuilder,
+}
+
+impl Drop for Node<'_> {
+    fn drop(&mut self) {
+        self.builder.end_child();
+    }
+}
```
cptartur, line 15: This is kind of magic, why not just call this explicitly?

crates/debugging/src/tree/node.rs
```diff
@@ -0,0 +1,80 @@
...
+
+impl<'a> Node<'a> {
+    /// Creates a [`TreeBuilder`] that should be used in [`Node::new`].
+    /// ## Note:
+    /// It is advised to use [`create_builder`] over creating a builder manually.
```
cptartur, line 20: Maybe let's make `builder` not pub so it is not allowed then?

crates/debugging/src/tree/node.rs
```diff
@@ -0,0 +1,80 @@
...
+
+    /// Creates a leaf node which parent is the current node.
+    pub fn leaf(&mut self, tree_item: &impl NodeDisplay) {
+        self.builder.add_empty_child(tree_item.display());
+    }
+
+    /// Consumes the [`Node`] and returns the serialized tree as a string.
+    /// ## Note:
+    /// This function assumes that the builder was created with [`Node::create_builder`].
+    /// As it extracts two guard nodes from the tree, it will panic if the assumption is not met.
```
cptartur, line 55: Again, let's not allow creating it otherwise then

crates/debugging/src/tree/serialize.rs
```diff
@@ -0,0 +1,29 @@
+use crate::trace::{Trace, TraceInfo};
+use crate::tree::node::Node;
+
+/// Trait for serializing a type into a tree.
+/// Implementations of this trait should only focus on placements of nodes in a tree not display aspects of them.
+/// Display should be handled by the [`NodeDisplay`](super::display::NodeDisplay) trait.
+pub trait TreeSerialize {
```
cptartur, line 7: If this is just for structuring, it should be named something else than serialize?

crates/debugging/src/trace.rs
ksew1: To be honest, I don't like tuple structs. Accessing fields via `.0` feels weird to me. I would prefer using this

crates/debugging/src/trace.rs
ksew1: I thought they wouldn't matter to the user, but we can bring them back.

crates/debugging/src/tree/display.rs
ksew1: I want to wait until we have all the data so we can look at the bigger picture and choose a nice ui, color palette, etc.

crates/debugging/src/tree/node.rs
ksew1: I think it is much easier to understand the way it is currently designed. You get a handle for the node, and if it goes out of scope, you can no longer add to it

crates/debugging/src/tree/node.rs
ksew1: I'm not sure how this would help. I need to get a reference to the current builder to construct a node, and I can't validate what's inside the builder until I build it. It's not like someone would make that mistake, as everything is asserted and tested. I just wanted to write good documentation for you so maintaining this would be easier for you, even without my help.

crates/debugging/src/tree/serialize.rs
ksew1: Well, from the user's perspective, you serialize data to the tree. It's just that, thanks to the separation of abstraction layers, the implementer doesn't need to handle the display aspect. Anyway if you don't like my explanation, I am open to new names

crates/debugging/src/trace.rs
cptartur: From what I see in `build_trace_data.rs` they should be included in the trace data generated by foundry for profiler. However, they are not included in execution resources. 

I'd verify if profiler includes those and if so, maybe we should as well?

cptartur: I'm not sure if including them will make the outputs easier to understand or more cluttered.

crates/debugging/src/tree/display.rs
cptartur: Okay, let's wait until then

crates/debugging/src/tree/node.rs
cptartur: You could wrap `TreeBuilder` in a (tuple) struct and make your methods accept it instead of a `TreeBuilder` directly. Then you can only allow creating it through your utility `create_builder`. 

It would prevent the developer from using it incorrectly and wouldn't complicate your code much because the inner `TreeBuilder` would still be available within the file even if it was private.

crates/debugging/src/tree/serialize.rs
cptartur: I like that this is abstracted I just wonder if this is the most obvious way we can do it. 

For me, the most obvious way would be if we had a trait just for transforming something into a tree structure, which would you then after have to explicitly attach as `.child_node` to your parent.

But if that's to much restructuring, how about naming it something more explicit, along the lines`AsTree` or `AsTreeNode`. Having something with a `serialize` that doesn't return any value and takes a mutable reference to a different struct is not obvious to me.

Then it could be:

```rust
pub trait AsTreeNode {
    fn as_tree_node(&self, parent: &mut Node);
}
```

crates/debugging/src/trace.rs
ksew1: Will investigate it here #3067

crates/debugging/src/tree/building/builder.rs
```diff
@@ -0,0 +1,47 @@
...
+    /// Finish adding children, and make the current item's parent current.
+    pub fn end_child(&mut self) {
+        self.builder.end_child();
+    }
+
+    /// Finish building the tree and return the top level item, not accounting for the guard nodes.
+    pub fn build(mut self) -> StringItem {
+        let string_item = self.builder.build();
+        let string_item = extract_guard(string_item);
+        extract_guard(string_item)
```
franciszekjob, line 36: nit: Can be simplified:
```suggestion
        extract_guard(extract_guard(string_item))
```

crates/debugging/src/tree/building/builder.rs
```diff
@@ -0,0 +1,47 @@
...
+    builder: TreeBuilder,
+}
+
+impl TreeBuilderWithGuard {
+    /// Creates a new [`TreeBuilderWithGuard`].
+    pub fn new() -> Self {
+        let mut builder = TreeBuilder::new("guard".to_string());
+        builder.begin_child("guard".to_string());
+        Self { builder }
+    }
```
franciszekjob, line 15: Could we add a short test checking that new tree has desired structure (with exactly two guards)?

franciszekjob (commented):
Looks good 👍 , left 2 minor comments

crates/debugging/src/tree/ui/display.rs
```diff
@@ -0,0 +1,97 @@
...
+impl NodeDisplay for CallResult {
+    const TAG: &'static str = "call result";
+    fn string_pretty(&self) -> String {
+        match self {
+            CallResult::Success { ret_data } => {
+                format!("success: {ret_data:?}")
+            }
+            CallResult::Failure(call_failure) => match call_failure {
+                CallFailure::Panic { panic_data } => {
+                    format!("panic: {panic_data:?}")
```
cptartur, line 77: Eventually we may want to have `ret_data` and `panic_data` stored in some wrapper struct so we can implement `Display` for them correctly. Using debug print is okay for now, but it doesn't guard your implementation from eventual type changes of these (that could break/change formatting).

cptartur (approved):

franciszekjob (approved):

crates/debugging/src/tree/ui/display.rs
ksew1: You mean a wrapper for `Vec<Felts>`? We can definitely think about it, but I want to have a bigger picture as we see what will be part of the UI and what will not.

PR #3039, opened 2025-03-03
Pin cache version to the forge version

### LINKED ISSUE #3021: Forge cache invalidation or versioning mechanism
### Current State

External structs are deserialized and used to create forge cache contents. If these structs change, e.g. due to dependencies bump, and user doesn't delete the caches manually it can lead to hard to debug errors after upgrading foundry.

### Objective

Add cache invalidation or versioning mechanism

### Additional Context

For example cache could also have foundry version serialized somehow or included in the name.

kkawula:
Closes #3021

## Introduced changes

-

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/forge/tests/data/forking/src/lib.cairo
```diff
@@ -1,6 +1,5 @@
 #[cfg(test)]
 mod tests {
-    use starknet::ContractAddress;
```
kkawula, line 3: Removes warning

crates/cheatnet/src/forking/cache.rs
```diff
@@ -13,13 +13,16 @@ use std::collections::HashMap;
 use std::fs;
 use std::fs::OpenOptions;
 use std::io::{Read, Write};
+use std::string::ToString;
 use url::Url;
 
-pub const CACHE_VERSION: usize = 4;
+pub fn cache_version() -> String {
+    env!("CARGO_PKG_VERSION").replace('.', "_")
```
cptartur, line 21: Can we use something like `forge_0_38_0` so basically add a prefix to the version. So if we ever want to version it "correctly" again, we can just get rid of the prefix.

kkawula: I don't know if it is necessary. I mean that if we decide to version it by the next integers as it was before, it won't conflict with the current versioning.

CHANGELOG.md
```diff
@@ -9,6 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Forge
 
+#### Changed
+
+- From now the cache version is pinned to the forge version.
```
cptartur, line 14:
```suggestion
- Fork cache version is pinned to the forge version.
```

scripts/release.sh
```diff
@@ -19,6 +19,20 @@ rm snforge_std/Scarb.toml.bak 2> /dev/null
...
+OLD_FILE=$(ls "$DIRECTORY" | grep -E '_v[0-9_]+\.json')
+OLD_FILE_PATH="$DIRECTORY/$OLD_FILE"
+NEW_FILE_PATH=$(echo "$OLD_FILE_PATH" | sed -E "s/_v[0-9_]+\.json$/_v${VERSION_UNDERSCORED}.json/")
+
+mv "$OLD_FILE_PATH" "$NEW_FILE_PATH"
+
+sed -i.bak -E "s/\"cache_version\":\"[0-9_]+\"/\"cache_version\":\"${VERSION_UNDERSCORED}\"/" "$NEW_FILE_PATH"
+rm "$NEW_FILE_PATH.bak" 2> /dev/null
+# end
+
```
cptartur, line 35: @THenry14 can you take a look at this?

scripts/release.sh
```diff
@@ -19,6 +19,20 @@ rm snforge_std/Scarb.toml.bak 2> /dev/null
 sed -i.bak "/\[package\]/,/version =/ s/version = \".*/version = \"${VERSION}\"/" crates/snforge-scarb-plugin/Scarb.toml
 rm crates/snforge-scarb-plugin/Scarb.toml.bak 2> /dev/null
 
+# start: Update cache test data
+VERSION_UNDERSCORED=$(echo "$VERSION" | tr '.' '_')
+
+DIRECTORY="crates/forge/tests/data/forking/.snfoundry_cache"
+OLD_FILE=$(ls "$DIRECTORY" | grep -E '_v[0-9_]+\.json')
```
THenry14, line 26:
```suggestion
OLD_FILE=$(find "$DIRECTORY" -type f -regex '.*_v[0-9][0-9_]*\.json')
```
or something along those lines.

Generally it is not advisable to use ls in such applications (or in scripts generally, eg problem: you can't distinguish files vs directories).

My suggestion should work for both linux and bsd (mac) find, but please test it beforehand.

Btw what are possible file names? maybe the regex could be improved somehow?

kkawula: Script updated, it works well

THenry14 (approved):

cptartur (approved):

PR #3046, opened 2025-03-05
Remove plugin from workspace and pin dependencies

cptartur:
Closes #

## Introduced changes

-

## Checklist

- [ ] Linked relevant issue
- [ ] Added relevant tests


cptartur:
Will need to follow up #3047

kkawula (approved):

crates/snforge-scarb-plugin/Cargo.toml
```diff
@@ -9,15 +9,15 @@ rust-version = "1.80.1"
...
-smol_str.workspace = true
-num-bigint.workspace = true
+cairo-lang-macro = "=0.1.0"
+cairo-lang-parser = "=2.10.0"
+cairo-lang-utils = "=2.10.0"
+cairo-lang-syntax = "=2.10.0"
+url = "=2.5.4"
+indoc = "=2.0.5"
+smol_str = "=0.2.2"
+num-bigint = "0.4.6"
```
ksew1, line 19: Seems like here should also be pin?

kkawula (changes requested):
Doesn't work

crates/snforge-scarb-plugin/Cargo.toml
cptartur: My bad, yes it should be pinned. WIll fix that though 0.38.1 on registry doesn't have it pinned.

kkawula (approved):

crates/snforge-scarb-plugin/Cargo.toml
```diff
@@ -9,15 +9,15 @@ rust-version = "1.80.1"
...
 [dependencies]
-cairo-lang-macro.workspace = true
-cairo-lang-parser.workspace = true
-cairo-lang-utils.workspace = true
-cairo-lang-syntax.workspace = true
-url.workspace = true
-indoc.workspace = true
-smol_str.workspace = true
-num-bigint.workspace = true
+cairo-lang-macro = "=0.1.0"
```
ddoktorski, line 12: nit: `cairo-lang-macro` can be removed from the workspace config

ddoktorski (approved):
Given that the plugin is no longer part of the workspace, what about moving it to the main directory instead of `/crates`?

PR #3051, opened 2025-03-06
Adjust --detailed-resources to sierra-gas tracking

THenry14:
Related to #2977

This PR stack aims to add support for new resource tracking (sierra-gas) and new gas representation (GasVector triplet instead of single l1 gas value).

-- (https://github.com/foundry-rs/starknet-foundry/pull/3049) Add sierra gas tracking
-- (https://github.com/foundry-rs/starknet-foundry/pull/3050) Run already existing tests with CairoSteps set explicitly
-- (This PR) Add support for sierra gas in `--detailed-resources`
-- (https://github.com/foundry-rs/starknet-foundry/pull/3052) Add support for sierra gas in `--save-trace-data`
-- (https://github.com/foundry-rs/starknet-foundry/pull/3053) Add sierra version assertion
-- (https://github.com/foundry-rs/starknet-foundry/pull/3054) Introduce new gas representation
-- (https://github.com/foundry-rs/starknet-foundry/pull/3055) Add gas tests for sierra-gas tracking
-- (https://github.com/foundry-rs/starknet-foundry/pull/3056) Update docs with sierra-gas related changes

PR #3050, opened 2025-03-06
Run forge tests with cairo steps explicitly set in tests

THenry14:
Related to #2977

This PR stack aims to add support for new resource tracking (sierra-gas) and new gas representation (GasVector triplet instead of single l1 gas value).

-- (https://github.com/foundry-rs/starknet-foundry/pull/3049) Add sierra gas tracking
--> (This PR) Run already existing tests with CairoSteps set explicitly
-- (https://github.com/foundry-rs/starknet-foundry/pull/3051) Add support for sierra gas in `--detailed-resources`
-- (https://github.com/foundry-rs/starknet-foundry/pull/3052) Add support for sierra gas in `--save-trace-data`
-- (https://github.com/foundry-rs/starknet-foundry/pull/3053) Add sierra version assertion
-- (https://github.com/foundry-rs/starknet-foundry/pull/3054) Introduce new gas representation
-- (https://github.com/foundry-rs/starknet-foundry/pull/3055) Add gas tests for sierra-gas tracking
-- (https://github.com/foundry-rs/starknet-foundry/pull/3056) Update docs with sierra-gas related changes

PR #3053, opened 2025-03-06
Add sierra assertion when tracking sierra gas

THenry14:
Related to #2977

--> (This PR) Add sierra version assertion
-- (https://github.com/foundry-rs/starknet-foundry/pull/3054) Introduce new gas representation
-- (https://github.com/foundry-rs/starknet-foundry/pull/3055) Add gas tests for sierra-gas tracking

### DIFF
--- a/crates/cheatnet/src/runtime_extensions/forge_runtime_extension/contracts_data.rs
+++ b/crates/cheatnet/src/runtime_extensions/forge_runtime_extension/contracts_data.rs
@@ -10,7 +10,7 @@ use starknet::core::utils::get_selector_from_name;
 use starknet_api::core::{ClassHash, EntryPointSelector};
 use std::collections::HashMap;
 
-type ContractName = String;
+pub type ContractName = String;
 type FunctionName = String;
 
 #[derive(Debug, Clone, PartialEq, Default)]
@@ -104,6 +104,11 @@ impl ContractsData {
         entry_point_selector: &EntryPointSelector,
     ) -> Option<&FunctionName> {
         self.selectors.get(entry_point_selector)
+    }
+
+    #[must_use]
+    pub fn get_contract_names(&self) -> Option<Vec<&ContractName>> {
+        (!self.contracts.is_empty()).then(|| self.contracts.keys().collect())
     }
 }
 
--- a/crates/cheatnet/src/runtime_extensions/forge_runtime_extension/contracts_data.rs
+++ b/crates/cheatnet/src/runtime_extensions/forge_runtime_extension/contracts_data.rs
@@ -10,7 +10,7 @@ use starknet::core::utils::get_selector_from_name;
 use starknet_api::core::{ClassHash, EntryPointSelector};
 use std::collections::HashMap;
 
-pub type ContractName = String;
+type ContractName = String;
 type FunctionName = String;
 
 #[derive(Debug, Clone, PartialEq, Default)]
@@ -104,11 +104,6 @@ impl ContractsData {
         entry_point_selector: &EntryPointSelector,
     ) -> Option<&FunctionName> {
         self.selectors.get(entry_point_selector)
-    }
-
-    #[must_use]
-    pub fn get_contract_names(&self) -> Option<Vec<&ContractName>> {
-        (!self.contracts.is_empty()).then(|| self.contracts.keys().collect())
     }
 }
 
--- a/crates/forge/src/compatibility_check.rs
+++ b/crates/forge/src/compatibility_check.rs
@@ -134,7 +134,7 @@ fn verify_contracts_sierra_versions(forge_config: &ForgeConfig) -> Result<()> {
             let contract_version = parse_sierra_version(sierra_str)?;
             if MINIMAL_SIERRA_VERSION_FOR_SIERRA_GAS > contract_version {
                 bail!(
-                    "Contract version {} is lower than required minimal sierra version",
+                    "Contract version {} is lower than required minimal sierra version {MINIMAL_SIERRA_VERSION_FOR_SIERRA_GAS}",
                     contract_version
                 );
             }
@@ -149,7 +149,7 @@ fn parse_sierra_version(sierra_str: &str) -> Result<Version> {
 
     let parsed_values: Vec<u8> = sierra_json["sierra_program"]
         .as_array()
-        .context("Unable to read `sierra_program`. Ensure it is an array of felts.")?
+        .context("Unable to read `sierra_program`.")?
         .iter()
         .take(3)
         .filter_map(|x| x.as_str())
@@ -431,6 +431,12 @@ mod tests {
                 execution_data_to_save: ExecutionDataToSave::default(),
             }),
         };
-        assert!(check_sierra_gas_version_requirement(&forge_config).is_err());
+        let result = check_sierra_gas_version_requirement(&forge_config).unwrap_err();
+        dbg!(&result.to_string());
+        assert!(
+            result
+                .to_string()
+                .contains("Tracking SierraGas is not supported for sierra <= 1.7.0")
+        );
     }
 }
--- a/crates/forge/src/compatibility_check.rs
+++ b/crates/forge/src/compatibility_check.rs
@@ -1,6 +1,4 @@
-use crate::MINIMAL_SIERRA_VERSION_FOR_SIERRA_GAS;
-use anyhow::{Context, Result, anyhow, bail};
-use forge_runner::forge_config::{ForgeConfig, ForgeTrackedResource};
+use anyhow::{Context, Result, anyhow};
 use regex::Regex;
 use semver::Version;
 use std::cell::RefCell;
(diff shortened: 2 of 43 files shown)


ddoktorski (commented):
Shouldn't it be sufficient to verify that contracts were compiled with `Scarb >=2.10.0` rather than checking Sierra version explicitly?

THenry14:
> Shouldn't it be sufficient to verify that contracts were compiled with `Scarb >=2.10.0` rather than checking Sierra version explicitly?

Yes, it could be done this way. 
I have opted, however, with checking contract class data directly because I wanted to be explicit - after all I don't care about scarb per se, but about sierra; I already have the contract class loaded, it's just a matter of parsing it for version, which is not that expensive.

If you have strong opinion on this let's discuss, personally I don't see many advantages from checking scarb version over this, but am ready to be schooled :)

crates/forge/src/compatibility_check.rs
```diff
@@ -115,10 +117,80 @@ fn get_raw_version(name: &str, command: &RefCell<Command>) -> Result<String> {
...
+                .test_runner_config
+                .contracts_data
+                .get_artifacts(contract)
+                .context(format!("Missing Sierra JSON for contract: {contract}"))?
+                .sierra;
+
+            let contract_version = parse_sierra_version(sierra_str)?;
+            if MINIMAL_SIERRA_VERSION_FOR_SIERRA_GAS > contract_version {
+                bail!(
+                    "Contract version {} is lower than required minimal sierra version",
```
cptartur, line 137: Can we do something along the lines of this?

```suggestion
                    "Contract version {} is lower than required minimal sierra version {MINIMAL_SIERRA_VERSION_FOR_SIERRA_GAS}",
```

crates/forge/src/compatibility_check.rs
```diff
@@ -115,10 +117,80 @@ fn get_raw_version(name: &str, command: &RefCell<Command>) -> Result<String> {
...
+    Ok(())
+}
+
+fn parse_sierra_version(sierra_str: &str) -> Result<Version> {
+    let sierra_json: serde_json::Value =
+        serde_json::from_str(sierra_str).context("Failed to parse Sierra JSON")?;
+
+    let parsed_values: Vec<u8> = sierra_json["sierra_program"]
+        .as_array()
+        .context("Unable to read `sierra_program`. Ensure it is an array of felts.")?
```
cptartur, line 152: This is not user generated so it cannot be fixed by a user anyway.

```suggestion
        .context("Unable to read `sierra_program`.")?
```

crates/forge/src/compatibility_check.rs
```diff
@@ -115,10 +117,80 @@ fn get_raw_version(name: &str, command: &RefCell<Command>) -> Result<String> {
...
+        .as_str(),
+    )
+    .map_err(std::convert::Into::into)
+}
+
+pub(crate) fn check_sierra_gas_version_requirement(forge_config: &ForgeConfig) -> Result<()> {
+    if forge_config.test_runner_config.tracked_resource == ForgeTrackedResource::SierraGas {
+        verify_contracts_sierra_versions(forge_config).with_context(||
+            format!(
+                "Tracking SierraGas is not supported for sierra <= {MINIMAL_SIERRA_VERSION_FOR_SIERRA_GAS}"
```
cptartur, line 176: This is kind of duplicating the error you get from the inside of the function.

crates/forge/src/compatibility_check.rs
```diff
@@ -249,4 +321,116 @@ mod tests {
...
+                cache_dir: Utf8PathBuf::default(),
+                contracts_data: ContractsData::try_from(contracts).unwrap(),
+                environment_variables: HashMap::default(),
+            }),
+            output_config: Arc::new(OutputConfig {
+                detailed_resources: false,
+                execution_data_to_save: ExecutionDataToSave::default(),
+            }),
+        };
+        assert!(check_sierra_gas_version_requirement(&forge_config).is_err());
```
cptartur, line 434: Can we do `unwrap_err` and assert the error contents?

ksew1:
> Yes, it could be done this way.
I have opted, however, with checking contract class data directly because I wanted to be explicit - after all I don't care about scarb per se, but about sierra; I already have the contract class loaded, it's just a matter of parsing it for version, which is not that expensive.

> If you have strong opinion on this let's discuss, personally I don't see many advantages from checking scarb version over this, but am ready to be schooled :)

To be honest, I think we should check for `scarb` versions. It is way easier for the user to understand that they have the wrong `scarb` version and should change it, rather than realizing they have the wrong `sierra` version.

crates/forge/src/compatibility_check.rs
THenry14: Not really, it adds context - what I wanted to achieve? - track sierra. what was the error? (inner function - wrong sierra version).

crates/forge/src/compatibility_check.rs
THenry14: Added, please note that there is also a e2e test that asserts this `sierra_gas_with_older_scarb`

THenry14:
> > Yes, it could be done this way.
> > I have opted, however, with checking contract class data directly because I wanted to be explicit - after all I don't care about scarb per se, but about sierra; I already have the contract class loaded, it's just a matter of parsing it for version, which is not that expensive.
> 
> > If you have strong opinion on this let's discuss, personally I don't see many advantages from checking scarb version over this, but am ready to be schooled :)
> 
> To be honest, I think we should check for `scarb` versions. It is way easier for the user to understand that they have the wrong `scarb` version and should change it, rather than realizing they have the wrong `sierra` version.

I see your point 🤔 But I also really think this should be done on package level, when we actually need the artifacts.
Do you think it would be better, if the error also stated minimal scarb version? (so not checking scarb per se, but tying it when interfacing with the user). That way user would be able to quickly tell which scarb version is needed, but we still would be able to check this per artifact. Maybe it's a stretch though. 🤔

cptartur:
> > > Yes, it could be done this way.
> > > I have opted, however, with checking contract class data directly because I wanted to be explicit - after all I don't care about scarb per se, but about sierra; I already have the contract class loaded, it's just a matter of parsing it for version, which is not that expensive.
> > 
> > 
> > > If you have strong opinion on this let's discuss, personally I don't see many advantages from checking scarb version over this, but am ready to be schooled :)
> > 
> > 
> > To be honest, I think we should check for `scarb` versions. It is way easier for the user to understand that they have the wrong `scarb` version and should change it, rather than realizing they have the wrong `sierra` version.
> 
> I see your point 🤔 But I also really think this should be done on package level, when we actually need the artifacts. Do you think it would be better, if the error also stated minimal scarb version? (so not checking scarb per se, but tying it when interfacing with the user). That way user would be able to quickly tell which scarb version is needed, but we still would be able to check this per artifact. Maybe it's a stretch though. 🤔

Is it possible for new Scarb versions to still generate older sierra somehow? If yes, we should check that. But otherwise it may be a bit overkill.

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/contracts_data.rs
```diff
@@ -105,6 +105,11 @@ impl ContractsData {
     ) -> Option<&FunctionName> {
         self.selectors.get(entry_point_selector)
     }
+
+    #[must_use]
+    pub fn get_contract_names(&self) -> Option<Vec<&ContractName>> {
+        (!self.contracts.is_empty()).then(|| self.contracts.keys().collect())
+    }
```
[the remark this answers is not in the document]
cptartur: I didn't understand what this code does correctly

crates/forge/src/compatibility_check.rs
cptartur: Hmm, okay, but for me it looks kind of verbose

THenry14:
> > Yes, it could be done this way.
> > I have opted, however, with checking contract class data directly because I wanted to be explicit - after all I don't care about scarb per se, but about sierra; I already have the contract class loaded, it's just a matter of parsing it for version, which is not that expensive.
> 
> > If you have strong opinion on this let's discuss, personally I don't see many advantages from checking scarb version over this, but am ready to be schooled :)
> 
> To be honest, I think we should check for `scarb` versions. It is way easier for the user to understand that they have the wrong `scarb` version and should change it, rather than realizing they have the wrong `sierra` version.

I see your point 🤔 But I also really think this should be done on package level, when we actually need the artifacts.
Do you think it would be better, if the error also stated minimal scarb version? (so not checking scarb per se, but tying it when interfacing with the user). That way user would be able to quickly tell which scarb version is needed, but we still would be able to check this per artifact. Maybe it's a stretch though. 🤔 

> > > > Yes, it could be done this way.
> > > > I have opted, however, with checking contract class data directly because I wanted to be explicit - after all I don't care about scarb per se, but about sierra; I already have the contract class loaded, it's just a matter of parsing it for version, which is not that expensive.
> > > 
> > > 
> > > > If you have strong opinion on this let's discuss, personally I don't see many advantages from checking scarb version over this, but am ready to be schooled :)
> > > 
> > > 
> > > To be honest, I think we should check for `scarb` versions. It is way easier for the user to understand that they have the wrong `scarb` version and should change it, rather than realizing they have the wrong `sierra` version.
> > 
> > 
> > I see your point 🤔 But I also really think this should be done on package level, when we actually need the artifacts. Do you think it would be better, if the error also stated minimal scarb version? (so not checking scarb per se, but tying it when interfacing with the user). That way user would be able to quickly tell which scarb version is needed, but we still would be able to check this per artifact. Maybe it's a stretch though. 🤔
> 
> Is it possible for new Scarb versions to still generate older sierra somehow? If yes, we should check that. But otherwise it may be a bit overkill.

Scarb is always tied up to a specific cairo + sierra versions, so there is no way an older scarb suddenly starts supporting newer sierra and the other way around.

ksew1:
> > > > > Yes, it could be done this way.
> > > > > I have opted, however, with checking contract class data directly because I wanted to be explicit - after all I don't care about scarb per se, but about sierra; I already have the contract class loaded, it's just a matter of parsing it for version, which is not that expensive.
> > > > 
> > > > 
> > > > > If you have strong opinion on this let's discuss, personally I don't see many advantages from checking scarb version over this, but am ready to be schooled :)
> > > > 
> > > > 
> > > > To be honest, I think we should check for `scarb` versions. It is way easier for the user to understand that they have the wrong `scarb` version and should change it, rather than realizing they have the wrong `sierra` version.
> > > 
> > > 
> > > I see your point 🤔 But I also really think this should be done on package level, when we actually need the artifacts. Do you think it would be better, if the error also stated minimal scarb version? (so not checking scarb per se, but tying it when interfacing with the user). That way user would be able to quickly tell which scarb version is needed, but we still would be able to check this per artifact. Maybe it's a stretch though. 🤔
> > 
> > 
> > Is it possible for new Scarb versions to still generate older sierra somehow? If yes, we should check that. But otherwise it may be a bit overkill.
> 
> Scarb is always tied up to a specific cairo + sierra versions, so there is no way an older scarb suddenly starts supporting newer sierra and the other way around.

If I had to leak my personal opinion, I would likely prefer the "fail-fast" model.  

By making assertions at the start of the program, we can detect issues early rather than encountering errors halfway through execution. This will improve maintainability since all checks are clearly defined in one place, making it easier to understand how they relate to each other.  

Also, I would base the check purely on Scarb version. Most users are probably not familiar with Sierra or how it relates to Scarb, so keeping it Scarb-centric would make it more accessible.

cptartur:
> > > > > > Yes, it could be done this way.
> > > > > > I have opted, however, with checking contract class data directly because I wanted to be explicit - after all I don't care about scarb per se, but about sierra; I already have the contract class loaded, it's just a matter of parsing it for version, which is not that expensive.
> > > > > 
> > > > > 
> > > > > > If you have strong opinion on this let's discuss, personally I don't see many advantages from checking scarb version over this, but am ready to be schooled :)
> > > > > 
> > > > > 
> > > > > To be honest, I think we should check for `scarb` versions. It is way easier for the user to understand that they have the wrong `scarb` version and should change it, rather than realizing they have the wrong `sierra` version.
> > > > 
> > > > 
> > > > I see your point 🤔 But I also really think this should be done on package level, when we actually need the artifacts. Do you think it would be better, if the error also stated minimal scarb version? (so not checking scarb per se, but tying it when interfacing with the user). That way user would be able to quickly tell which scarb version is needed, but we still would be able to check this per artifact. Maybe it's a stretch though. 🤔
> > > 
> > > 
> > > Is it possible for new Scarb versions to still generate older sierra somehow? If yes, we should check that. But otherwise it may be a bit overkill.
> > 
> > 
> > Scarb is always tied up to a specific cairo + sierra versions, so there is no way an older scarb suddenly starts supporting newer sierra and the other way around.
> 
> If I had to leak my personal opinion, I would likely prefer the "fail-fast" model.
> 
> By making assertions at the start of the program, we can detect issues early rather than encountering errors halfway through execution. This will improve maintainability since all checks are clearly defined in one place, making it easier to understand how they relate to each other.
> 
> Also, I would base the check purely on Scarb version. Most users are probably not familiar with Sierra or how it relates to Scarb, so keeping it Scarb-centric would make it more accessible.

Sierra information could be relevant to the users that are aware of the USC existing, so maybe it can still be printed. But definitely an information about Scarb version is required.

THenry14:
Yeah, You're right guys, I'm convinced :+1:
Made the necessary changes:
- we now check scarb version, and not sierra version
- the check is moved to a requirements checker (as it makes more sense now to verify pre-compilation)
- Info about sierra is still retained in error messsage, as scarb is tied to specific sierra version

ddoktorski (approved):

cptartur (approved):

ksew1 (approved):

PR #3054, opened 2025-03-06
New gas representation

THenry14:
Related to #2977

-- (https://github.com/foundry-rs/starknet-foundry/pull/3053) Add sierra version assertion
--> (This PR) Introduce new gas representation
-- (https://github.com/foundry-rs/starknet-foundry/pull/3055) Add gas tests for sierra-gas tracking

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 #### Changed
 - gas is now reported using resource bounds triplet (l1_gas, l1_data_gas and l2_gas)
-- `available_gas` now accepts named arguments denoting resource bounds
+- `available_gas` now accepts named arguments denoting resource bounds (eg #[available_gas(l1_gas: 1, l1_data_gas: 2, l2_gas: 3)])
 
 ### Cast
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - new `--tracked-resource` flag, that will change currently tracked resource
   (`cairo-steps` for vm resources - default; `sierra-gas` for sierra gas consumed resources in cairo native)
 
+#### Changed
+- gas is now reported using resource bounds triplet (l1_gas, l1_data_gas and l2_gas)
+- `available_gas` now accepts named arguments denoting resource bounds
+
 ### Cast
 
 #### Added
--- a/crates/forge-runner/src/gas.rs
+++ b/crates/forge-runner/src/gas.rs
@@ -1,4 +1,5 @@
 use crate::test_case_summary::{Single, TestCaseSummary};
+use anyhow::anyhow;
 use blockifier::abi::constants;
 use blockifier::context::TransactionContext;
 use blockifier::execution::call_info::{ChargedResources, EventSummary, ExecutionSummary};
@@ -13,6 +14,7 @@ use blockifier::utils::u64_from_usize;
 use cheatnet::runtime_extensions::call_to_blockifier_runtime_extension::rpc::UsedResources;
 use cheatnet::runtime_extensions::forge_config_extension::config::RawAvailableGasConfig;
 use cheatnet::state::ExtendedStateReader;
+use shared::print::print_as_warning;
 use starknet_api::execution_resources::{GasAmount, GasVector};
 use starknet_api::transaction::EventContent;
 use starknet_api::transaction::fields::GasVectorComputationMode;
@@ -43,7 +45,7 @@ pub fn calculate_used_gas(
         vm_resources: resources.execution_resources.clone(),
         n_reverted_steps: 0,
         sierra_gas: resources.gas_consumed,
-        reverted_sierra_gas: GasAmount(0),
+        reverted_sierra_gas: GasAmount::ZERO,
     };
 
     let transaction_resources = TransactionResources {
@@ -164,7 +166,15 @@ pub fn check_available_gas(
             ..
         } if available_gas.is_some_and(|available_gas| match available_gas {
             RawAvailableGasConfig::MaxGas(gas) => {
-                gas_info.l1_gas + gas_info.l1_data_gas + gas_info.l2_gas > GasAmount(gas as u64)
+                print_as_warning(&anyhow!(
+                    "Setting available_gas with unnamed argument is deprecated. \
+                Consider setting resource bounds (l1_gas, l1_data_gas and l2_gas) explicitly."
+                ));
+                // convert resource bounds to classic l1_gas using formula
+                // l1_gas + l1_data_gas + (l2_gas / 40000)
+                // because 100 l2_gas = 0.0025 l1_gas
+                gas_info.l1_gas + gas_info.l1_data_gas + (gas_info.l2_gas / 40000)
+                    > GasAmount(gas as u64)
             }
             RawAvailableGasConfig::MaxResourceBounds(bounds) => {
                 let av_gas = bounds.to_gas_vector();
--- a/crates/forge-runner/src/gas.rs
+++ b/crates/forge-runner/src/gas.rs
@@ -166,6 +166,7 @@ pub fn check_available_gas(
             ..
         } if available_gas.is_some_and(|available_gas| match available_gas {
             RawAvailableGasConfig::MaxGas(gas) => {
+                // todo(3109): remove uunnamed argument in available_gas
                 print_as_warning(&anyhow!(
                     "Setting available_gas with unnamed argument is deprecated. \
                 Consider setting resource bounds (l1_gas, l1_data_gas and l2_gas) explicitly."
--- a/crates/forge-runner/src/gas.rs
+++ b/crates/forge-runner/src/gas.rs
@@ -41,9 +41,7 @@ pub fn calculate_used_gas(
     let computation_resources = ComputationResources {
         vm_resources: resources.execution_resources.clone(),
         n_reverted_steps: 0,
-        // TODO(#2977)
-        sierra_gas: GasAmount(0),
-        // TODO(#2977)
+        sierra_gas: resources.gas_consumed,
         reverted_sierra_gas: GasAmount(0),
     };
 
@@ -56,8 +54,7 @@ pub fn calculate_used_gas(
     Ok(transaction_resources.to_gas_vector(
         versioned_constants,
         use_kzg_da,
-        // TODO(#2977)
-        &GasVectorComputationMode::NoL2Gas,
+        &GasVectorComputationMode::All,
     ))
 }
 
@@ -154,7 +151,7 @@ fn get_state_resources(
 }
 
 pub fn check_available_gas(
-    available_gas: Option<usize>,
+    available_gas: Option<GasVector>,
     summary: TestCaseSummary<Single>,
 ) -> TestCaseSummary<Single> {
     match summary {
@@ -164,11 +161,17 @@ pub fn check_available_gas(
             gas_info,
             debugging_trace,
             ..
-        } if available_gas.is_some_and(|available_gas| gas_info > available_gas as u128) => {
+        } if available_gas.is_some_and(|available_gas| {
+            gas_info.l1_gas > available_gas.l1_gas
+                || gas_info.l1_data_gas > available_gas.l1_data_gas
+                || gas_info.l2_gas > available_gas.l2_gas
+        }) =>
+        {
             TestCaseSummary::Failed {
                 name,
                 msg: Some(format!(
-                    "\n\tTest cost exceeded the available gas. Consumed gas: ~{gas_info}"
+                    "\n\tTest cost exceeded the available gas. Consumed l1_gas: ~{}, l1_data_gas: ~{}, l2_gas: ~{}",
+                    gas_info.l1_gas, gas_info.l1_data_gas, gas_info.l2_gas
                 )),
                 arguments,
                 fuzzer_args: Vec::default(),
--- a/crates/forge-runner/src/gas.rs
+++ b/crates/forge-runner/src/gas.rs
@@ -11,6 +11,7 @@ use blockifier::state::errors::StateError;
 use blockifier::transaction::objects::HasRelatedFeeType;
 use blockifier::utils::u64_from_usize;
 use cheatnet::runtime_extensions::call_to_blockifier_runtime_extension::rpc::UsedResources;
+use cheatnet::runtime_extensions::forge_config_extension::config::RawAvailableGasConfig;
 use cheatnet::state::ExtendedStateReader;
 use starknet_api::execution_resources::{GasAmount, GasVector};
 use starknet_api::transaction::EventContent;
@@ -151,7 +152,7 @@ fn get_state_resources(
 }
 
 pub fn check_available_gas(
-    available_gas: Option<GasVector>,
+    available_gas: Option<RawAvailableGasConfig>,
     summary: TestCaseSummary<Single>,
 ) -> TestCaseSummary<Single> {
     match summary {
@@ -161,10 +162,16 @@ pub fn check_available_gas(
             gas_info,
             debugging_trace,
             ..
-        } if available_gas.is_some_and(|available_gas| {
-            gas_info.l1_gas > available_gas.l1_gas
-                || gas_info.l1_data_gas > available_gas.l1_data_gas
-                || gas_info.l2_gas > available_gas.l2_gas
+        } if available_gas.is_some_and(|available_gas| match available_gas {
+            RawAvailableGasConfig::MaxGas(gas) => {
+                gas_info.l1_gas + gas_info.l1_data_gas + gas_info.l2_gas > GasAmount(gas as u64)
+            }
+            RawAvailableGasConfig::MaxResourceBounds(bounds) => {
+                let av_gas = bounds.to_gas_vector();
+                gas_info.l1_gas > av_gas.l1_gas
+                    || gas_info.l1_data_gas > av_gas.l1_data_gas
+                    || gas_info.l2_gas > av_gas.l2_gas
+            }
         }) =>
         {
             TestCaseSummary::Failed {
--- a/crates/forge-runner/src/printing.rs
+++ b/crates/forge-runner/src/printing.rs
@@ -25,8 +25,21 @@ pub fn print_test_result(
                 gas_info,
                 ..
             } => Some(format!(
-                " (runs: {runs}, gas: {{max: ~{}, min: ~{}, mean: ~{:.2}, std deviation: ~{:.2}}})",
-                gas_info.max, gas_info.min, gas_info.mean, gas_info.std_deviation
+                " (runs: {runs}, l1_gas: {{max: ~{}, min: ~{}, mean: ~{:.2}, std deviation: ~{:.2}}},\
+                l1_data_gas: {{max: ~{}, min: ~{}, mean: ~{:.2}, std deviation: ~{:.2}}},\
+                l2_gas: {{max: ~{}, min: ~{}, mean: ~{:.2}, std deviation: ~{:.2}}})",
+                gas_info.l1_gas.max,
+                gas_info.l1_gas.min,
+                gas_info.l1_gas.mean,
+                gas_info.l1_gas.std_deviation,
+                gas_info.l1_data_gas.max,
+                gas_info.l1_data_gas.min,
+                gas_info.l1_data_gas.mean,
+                gas_info.l1_data_gas.std_deviation,
+                gas_info.l2_gas.max,
+                gas_info.l2_gas.min,
+                gas_info.l2_gas.mean,
+                gas_info.l2_gas.std_deviation
             )),
             TestCaseSummary::Failed {
                 fuzzer_args,
@@ -40,7 +53,10 @@ pub fn print_test_result(
 
     let gas_usage = match any_test_result {
         AnyTestCaseSummary::Single(TestCaseSummary::Passed { gas_info, .. }) => {
-            format!(" (gas: ~{gas_info})")
+            format!(
+                " (l1_gas: ~{}, l1_data_gas: ~{}, l2_gas: ~{})",
+                gas_info.l1_gas, gas_info.l1_data_gas, gas_info.l2_gas
+            )
         }
         _ => String::new(),
     };
--- a/crates/forge-runner/src/printing.rs
+++ b/crates/forge-runner/src/printing.rs
@@ -24,23 +24,7 @@ pub fn print_test_result(
                 test_statistics: FuzzingStatistics { runs },
                 gas_info,
                 ..
-            } => Some(format!(
-                " (runs: {runs}, l1_gas: {{max: ~{}, min: ~{}, mean: ~{:.2}, std deviation: ~{:.2}}},\
-                l1_data_gas: {{max: ~{}, min: ~{}, mean: ~{:.2}, std deviation: ~{:.2}}},\
-                l2_gas: {{max: ~{}, min: ~{}, mean: ~{:.2}, std deviation: ~{:.2}}})",
-                gas_info.l1_gas.max,
-                gas_info.l1_gas.min,
-                gas_info.l1_gas.mean,
-                gas_info.l1_gas.std_deviation,
-                gas_info.l1_data_gas.max,
-                gas_info.l1_data_gas.min,
-                gas_info.l1_data_gas.mean,
-                gas_info.l1_data_gas.std_deviation,
-                gas_info.l2_gas.max,
-                gas_info.l2_gas.min,
-                gas_info.l2_gas.mean,
-                gas_info.l2_gas.std_deviation
-            )),
+            } => Some(format!(" (runs: {runs}, {gas_info})",)),
             TestCaseSummary::Failed {
                 fuzzer_args,
                 test_statistics: FuzzingStatistics { runs },
--- a/crates/forge-runner/src/test_case_summary.rs
+++ b/crates/forge-runner/src/test_case_summary.rs
@@ -13,6 +13,7 @@ use cheatnet::state::{CallTrace as InternalCallTrace, EncounteredError};
 use conversions::byte_array::ByteArray;
 use num_traits::Pow;
 use shared::utils::build_readable_text;
+use starknet_api::execution_resources::GasVector;
 use starknet_types_core::felt::Felt;
 use std::cell::RefCell;
 use std::option::Option;
@@ -20,6 +21,13 @@ use std::rc::Rc;
 
 #[derive(Debug, PartialEq, Clone, Default)]
 pub struct GasStatistics {
+    pub l1_gas: GasStatisticsComponent,
+    pub l1_data_gas: GasStatisticsComponent,
+    pub l2_gas: GasStatisticsComponent,
+}
+
+#[derive(Debug, PartialEq, Clone, Default)]
+pub struct GasStatisticsComponent {
     pub min: u128,
     pub max: u128,
     pub mean: f64,
@@ -28,14 +36,49 @@ pub struct GasStatistics {
 
 impl GasStatistics {
     #[must_use]
-    pub fn new(gas_usages: &[u128]) -> Self {
-        let mean = Self::mean(gas_usages);
+    pub fn new(gas_usages: &[GasVector]) -> Self {
+        let l1_gas_values: Vec<u128> = gas_usages
+            .iter()
+            .map(|gv| u128::from(gv.l1_gas.0))
+            .collect();
+        let l1_data_gas_values: Vec<u128> = gas_usages
+            .iter()
+            .map(|gv| u128::from(gv.l1_data_gas.0))
+            .collect();
+        let l2_gas_values: Vec<u128> = gas_usages
+            .iter()
+            .map(|gv| u128::from(gv.l2_gas.0))
+            .collect();
+
+        let l1_gas_mean = Self::mean(l1_gas_values.as_ref());
+        let l1_data_gas_mean = Self::mean(l1_data_gas_values.as_ref());
+        let l2_gas_mean = Self::mean(l2_gas_values.as_ref());
 
         GasStatistics {
-            min: gas_usages.iter().min().copied().unwrap(),
-            max: gas_usages.iter().max().copied().unwrap(),
-            mean,
-            std_deviation: Self::std_deviation(mean, gas_usages),
+            l1_gas: {
+                GasStatisticsComponent {
+                    min: l1_gas_values.iter().min().copied().unwrap(),
+                    max: l1_gas_values.iter().max().copied().unwrap(),
+                    mean: l1_gas_mean,
+                    std_deviation: Self::std_deviation(l1_gas_mean, l1_gas_values.as_ref()),
+                }
+            },
+            l1_data_gas: {
+                GasStatisticsComponent {
+                    min: l1_data_gas_values.iter().min().copied().unwrap(),
+                    max: l1_data_gas_values.iter().max().copied().unwrap(),
+                    mean: l1_data_gas_mean,
+                    std_deviation: Self::std_deviation(l1_gas_mean, l1_data_gas_values.as_ref()),
+                }
+            },
+            l2_gas: {
+                GasStatisticsComponent {
+                    min: l2_gas_values.iter().min().copied().unwrap(),
+                    max: l2_gas_values.iter().max().copied().unwrap(),
+                    mean: l2_gas_mean,
+                    std_deviation: Self::std_deviation(l1_gas_mean, l2_gas_values.as_ref()),
+                }
+            },
         }
     }
 
@@ -77,7 +120,7 @@ impl TestType for Fuzzing {
 #[derive(Debug, PartialEq, Clone)]
 pub struct Single;
 impl TestType for Single {
-    type GasInfo = u128;
+    type GasInfo = GasVector;
     type TestStatistics = ();
     type TraceData = VersionedProfilerCallTrace;
 }
@@ -128,7 +171,6 @@ pub enum TestCaseSummary<T: TestType> {
     Skipped {},
 }
 
-#[expect(clippy::large_enum_variant)]
 #[derive(Debug)]
 pub enum AnyTestCaseSummary {
     Fuzzing(TestCaseSummary<Fuzzing>),
@@ -190,7 +232,7 @@ impl TestCaseSummary<Fuzzing> {
                 debugging_trace,
             } => {
                 let runs = results.len();
-                let gas_usages: Vec<u128> = results
+                let gas_usages: Vec<GasVector> = results
                     .into_iter()
                     .map(|a| match a {
                         TestCaseSummary::Passed { gas_info, .. } => gas_info,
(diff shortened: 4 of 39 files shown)


THenry14:
🔴 
Please take a look specifically at `cheat_max_fee` cheatcode while reviewing this PR - I have not touched this cheatcode at all, and am not sure if it should *just work* or perhaps I have missed something crucial here that is not causing tests to explode.

crates/forge-runner/src/test_case_summary.rs
```diff
@@ -28,14 +36,49 @@ pub struct GasStatistics {
...
+            .map(|gv| u128::from(gv.l1_data_gas.0))
+            .collect();
+        let l2_gas_values: Vec<u128> = gas_usages
+            .iter()
+            .map(|gv| u128::from(gv.l2_gas.0))
+            .collect();
+
+        let l1_gas_mean = Self::mean(l1_gas_values.as_ref());
+        let l1_data_gas_mean = Self::mean(l1_data_gas_values.as_ref());
+        let l2_gas_mean = Self::mean(l2_gas_values.as_ref());
```
ksew1, line 55: Also this function doesn't work. At least not how it is intended to be. The division `sum / len` is done using integer division because both values are `u128`, which means it truncates any decimals before converting to `f64`.

I would personally use `statrs`or any other statistical library. It would be less simpler and less error prone.

```rs
use statrs::statistics::Statistics;

let data = Data::new(vec![1.0, 2.0, 3.0, 4.0]);
println!("{}", data.mean().unwrap());
let data = vec![1.0, 2.0, 3.0, 4.0];
println!("{}", data.std_dev());
```

crates/forge/test_utils/src/runner.rs
```diff
@@ -288,7 +289,9 @@ pub fn assert_gas(result: &[TestTargetSummary], test_case_name: &str, asserted_g
             }
             AnyTestCaseSummary::Single(case) => match case {
                 TestCaseSummary::Passed { gas_info: gas, .. } => {
-                    *gas == asserted_gas
+                    gas.l1_gas.0 == asserted_gas.l1_gas.0
```
ksew1, line 292: We can just use `Eq` as `GasVector` implements it

crates/forge/tests/integration/fuzzing.rs
```diff
@@ -5,6 +5,8 @@ use test_utils::runner::{TestCase, assert_passed};
 use test_utils::running_tests::run_test_case;
 use test_utils::test_case;
 
+const ALLOWED_ERROR: f64 = 0.01;
```
ksew1, line 8: Why we don't use epsilon?

crates/snforge-scarb-plugin/src/attributes/available_gas.rs
```diff
@@ -24,14 +25,32 @@ impl AttributeCollector for AvailableGasCollector {
         args: Arguments,
         _warns: &mut Vec<Diagnostic>,
     ) -> Result<String, Diagnostics> {
-        let &[arg] = args.unnamed_only::<Self>()?.of_length::<1, Self>()?;
-
-        let gas = Number::parse_from_expr::<Self>(db, arg.1, arg.0.to_string().as_str())?;
-
-        let gas = gas.as_cairo_expression();
+        let named_args = args.named_only::<Self>()?;
```
cptartur, line 28: Maybe, to facilitate easier migration, we should still allow having a single unnamed attribute? 

The logic would be `either allow single unnamed attribute or only named attributes`. This way it would be backward compatible.

crates/forge/tests/integration/gas.rs
```diff
@@ -32,11 +36,20 @@ fn declare_cost_is_omitted() {
 
     assert_passed(&result);
     // 1 = cost of 230 steps (because int(0.0025 * 230) = 1)
-    assert_gas(&result, "declare_cost_is_omitted", 1);
+    // 0 l1_gas + 0 l1_data_gas + 1 * (100 / 0.0025) l2 gas
```
cptartur, line 39: So the calculation here is now that first it computes the amount of L1 gas (1 because of 230 steps) and then it converts to L2?

crates/forge/tests/integration/gas.rs
```diff
@@ -747,11 +945,20 @@ fn events_cost() {
     // 156 range_check_builtin ~= 7
     // 6 gas for 50 event values
     // ~13 gas for 50 event keys
-    assert_gas(&result, "events_cost", 7 + 6 + 13);
+    // 0 l1_gas + 0 l1_data_gas + (7 + 6 + ~13) * (100 / 0.0025) l2 gas
```
cptartur, line 948: Maybe if we are asserting exactly 13, let's remove `~` from both lines?

crates/forge/tests/integration/gas.rs
```diff
@@ -784,5 +991,14 @@ fn events_contract_cost() {
...
     // ~13 gas for 50 event keys
-    assert_gas(&result, "event_emission_cost", 11 + 96 + 6 + 13);
+    // 0 l1_gas + 96 l1_data_gas + (11 + 6 + ~13) * (100 / 0.0025) l2 gas
+    assert_gas(
+        &result,
+        "event_emission_cost",
+        GasVector {
+            l1_gas: GasAmount(0),
+            l1_data_gas: GasAmount(96),
+            l2_gas: GasAmount(1_208_000),
```
cptartur, line 1001: How does this work exactly? It's because it's `~30 * 40,000` not `30 * 40,000`?

crates/snforge-scarb-plugin/src/attributes/available_gas.rs
```diff
@@ -24,14 +25,32 @@ impl AttributeCollector for AvailableGasCollector {
...
+
+        let l1_gas = named_args
+            .as_once_optional("l1_gas")?
+            .map(|arg| Number::parse_from_expr::<Self>(db, arg, "l1_gas"))
+            .transpose()?
+            .unwrap_or(Number(BigInt::ZERO));
```
ddoktorski, line 34: Shouldn't the logic rather be: if not provided treat it as unlimited? I expect this to be a common use case to limit only a specific gas type. With the current approach it also enforces setting other types if they are not 0.

crates/forge-runner/src/gas.rs
```diff
@@ -163,11 +160,17 @@ pub fn check_available_gas(
             arguments,
             gas_info,
             ..
-        } if available_gas.is_some_and(|available_gas| gas_info > available_gas as u128) => {
+        } if available_gas.is_some_and(|available_gas| {
+            gas_info.l1_gas > available_gas.l1_gas
```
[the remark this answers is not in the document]
THenry14: GasVector is defined in `starknet_api` crate, so I would have to make a wrapper types for it. I didn't think it's worth it tbh, but maybe I'm wrong? 🤔 Thoughts/arguments?

crates/forge-runner/src/test_case_summary.rs
THenry14: Correct me if I'm wrong, but the module requires u128 -> f64 conversion, so won't it mean that we also loose precision and range? How is it better than operating on u128 and only as a last step converting to f64? (maybe I have a brainfart or something) 🤔

crates/forge/test_utils/src/runner.rs
THenry14: huh, my IDE marked it as an error for some reason but you're right, weird :+1:

crates/forge/tests/integration/fuzzing.rs
THenry14: clippy complained https://rust-lang.github.io/rust-clippy/master/index.html#float_cmp

CHANGELOG.md
```diff
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - new `--tracked-resource` flag, that will change currently tracked resource
   (`cairo-steps` for vm resources - default; `sierra-gas` for sierra gas consumed resources in cairo native)
 
+#### Changed
+- gas is now reported using resource bounds triplet (l1_gas, l1_data_gas and l2_gas)
+- `available_gas` now accepts named arguments denoting resource bounds
```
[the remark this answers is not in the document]
THenry14: Added an example attribute (since I think this is what you had in mind). Please verify that it 🙏

crates/forge-runner/src/gas.rs
```diff
@@ -56,8 +54,7 @@ pub fn calculate_used_gas(
     Ok(transaction_resources.to_gas_vector(
         versioned_constants,
         use_kzg_da,
-        // TODO(#2977)
-        &GasVectorComputationMode::NoL2Gas,
+        &GasVectorComputationMode::All,
```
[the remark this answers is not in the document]
THenry14: Tracked resource does not matter here, this decribes resource bounds.

crates/forge/tests/data/fuzzing/tests/multiple_attributes.cairo
```diff
@@ -1,7 +1,7 @@
 use fuzzing::adder;
 use fuzzing::fib;
 
-#[available_gas(100)]
+#[available_gas(l2_gas: 40000000)]
```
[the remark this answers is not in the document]
THenry14: The error is:

```
error: Plugin diagnostic: #[available_gas] can be used with named arguments only
 --> trace/tests/test_trace.cairo[proc_test][proc___internal_config_statement]:3:1
#[available_gas(123)]
^^^^^^^^^^^^^^^^^^^^^
note: this error originates in the attribute macro: `__internal_config_statement`
```

and it originates from Arguments structs (is the same to all attributes that use it's interface). 
Added test for this in newest patchset.

crates/snforge-scarb-plugin/src/attributes/available_gas.rs
THenry14: Hmm, this is to be consider for sure. 
My rationale for introducing breaking change is that this way _should be encouraged_, and setting just single max fee argument will be removed in next starknet version. Pretty similar situation to returning triplet in gas estimations.

for sure we can discuss :+1:

crates/forge/tests/integration/gas.rs
THenry14: no, l1_gas is totally different now, and it does not cover execution costs (afaik it just handles messaging).
the calculation here is: there was 1 execution cost from steps, and we jsut convert it to L2 gas (because l2 gas is now a measure of execution cost)

crates/forge/tests/integration/gas.rs
THenry14: here's the thing - afaik it was not exactly 13 :D that's why I have left the ~, as far as I remember it was actually 13.2 (but I didn't dig into why it was that, I assumed You have made a good job calculating this and I only verified new calculations, no conversions).

crates/forge/tests/integration/gas.rs
THenry14: The same as https://github.com/foundry-rs/starknet-foundry/pull/3054#discussion_r1987609375

the event does not cost exactly 13, but 13.2 (so your comments here were an approximation, I beliebe that is why someone said ~13 and not 13), hence discrepancy. Again: I did not verify those number, just assumed You did a good job doing it (I just did a conversion, which does not introduce any measuring errors). I only verified sierra gas calculations.

crates/snforge-scarb-plugin/src/attributes/available_gas.rs
THenry14: Ahh yes, you're right that is just plain better UX. 
the only caveat is, right now technically you can use l1 gas to pay for execution resources (normally paid for by l2), so you will need to be very careful with setting those two (eg if you were to set l2 to something that is not sufficient, l1 can be taken and chipped in to cover the fee). I believe this behaviour will not be permitted from next starknet version.

This can be pretty confusing, I think. Thoughts?

cptartur: Maybe if we could allow explicitly setting these to unlimited somehow (maybe through some kind of enum, but not sure if this is possible in plugin), then it would be the best. 
But that might be to complicated to implement.

crates/forge/tests/integration/gas.rs
cptartur: It was like this for a while now. Maybe you can create an issue to investigate this aferwards?

cptartur: And link it in code here

crates/forge/tests/integration/gas.rs
cptartur: https://github.com/foundry-rs/starknet-foundry/pull/3054#discussion_r1989865883

crates/forge/tests/integration/gas.rs
cptartur: Right, but this cost of `1` is measured from 230 steps which are currently priced in l1 gas? 
image

crates/snforge-scarb-plugin/src/attributes/available_gas.rs
ddoktorski: > I believe this behaviour will not be permitted from next starknet version

Do you mean from `0.14.0`? Can we quickly verify if setting `l2_gas` to a very low value with logic for unlimited `l1_gas` will result in the missing gas being taken from l1?

> Maybe if we could allow explicitly setting these to unlimited somehow 

The default behavior is unlimited, so for consistency not setting a given type should keep it as the default, so unlimited.

crates/forge-runner/src/gas.rs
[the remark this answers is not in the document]
ksew1: I don't think is worth upstream i would leave this as it is. It is honestly no biggie

crates/forge-runner/src/test_case_summary.rs
THenry14: Hmm, yeah that is probably a good call. Changed :+1:

crates/forge-runner/src/gas.rs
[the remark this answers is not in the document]
THenry14: This is because starknet 13.4 still supports just setting l1_gas (this support is scheduled to be removed in starknet 0.14). So If you were to pick NoL2Gas, the `to_gas_vector` just converts everything to l1_gas and do not return resource bounds. This is described [here](https://community.starknet.io/t/starknet-v0-13-4-pre-release-notes/115257#p-2358763-backward-compatibility-6)

crates/forge-runner/src/printing.rs
```diff
@@ -24,8 +24,21 @@ pub fn print_test_result(
...
+                gas_info.l1_gas.mean,
+                gas_info.l1_gas.std_deviation,
+                gas_info.l1_data_gas.max,
+                gas_info.l1_data_gas.min,
+                gas_info.l1_data_gas.mean,
+                gas_info.l1_data_gas.std_deviation,
+                gas_info.l2_gas.max,
+                gas_info.l2_gas.min,
+                gas_info.l2_gas.mean,
+                gas_info.l2_gas.std_deviation
```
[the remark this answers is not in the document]
THenry14: Hmm, let me create an issue for that, since this might need some investigation and proper planning before coming to a conclusion https://github.com/foundry-rs/starknet-foundry/issues/3077

crates/snforge-scarb-plugin/src/attributes/available_gas.rs
THenry14: I'm not sure what is the flow here tbh, but I don't think there are attributes that accept multiple types of arguments (named AND unnamed), and I don't that is a good idea to start doing this, so that means we would have to make a new attribute probably. Which might be ok actually. It's just a question if we want this, and I leave the decision up to you :+1:

crates/forge/tests/integration/gas.rs
THenry14: It's:

1 cairo step = 0.0025 L1 gas = 100 L2 gas

230 steps * 0.0025 l1 = 0.575, so 1 (because this is the smallest permissible value)
230 steps * 100 l2 = 23000 (but the same logic as above applies, so minimal value becomes 40000, since 1 l1 gas = 40000 l2 gas)

crates/forge/tests/integration/gas.rs
THenry14: done https://github.com/foundry-rs/starknet-foundry/issues/3078

crates/snforge-scarb-plugin/src/attributes/available_gas.rs
THenry14: Yes, I meant 0.14. To verify this, I have modified `integration::available_gas::available_gas_exceeded` test from `forge` crate; currently it has only l2_gas set to 5, the rest is set to 0 (default), and the test checks if a panic is being thrown because available gas was exceeded (it does). Now I change `available_gas` like so:

FROM:
```
            #[test]
            #[available_gas(l2_gas: 5)]
            fn keccak_cost() {
                keccak::keccak_u256s_le_inputs(array![1].span());
            }
```

TO:
```
            #[test]
            #[available_gas(l1_gas: 999999999999999999999999999999999, l2_gas: 5)]
            fn keccak_cost() {
                keccak::keccak_u256s_le_inputs(array![1].span());
            }
```

and the tests fails with:

```
thread 'integration::available_gas::available_gas_exceeded' panicked at crates/forge/test_utils/src/runner.rs:253:5:
Some tests didn't fail
```

upon inspection, to make sure it fails because test succeeds and not something totally unrelated, I see:

```
    TestTargetSummary {
        test_case_summaries: [
            Single(
                Passed {
                    name: "test_package_integrationtest::test_case::keccak_cost",
```

> The default behavior is unlimited, so for consistency not setting a given type should keep it as the default, so unlimited.

This was far easier, when only max_fee could be specified 😞 
I mean, I do not see a clear solution for this to be honest (apart from setting this to 0 as it is right now or just straight up requiring users to set all bounds)

crates/forge-runner/src/test_case_summary.rs
ksew1: It doesn't matter whether the sum is in u128 or f64 because for division, you need to cast the sum to f64 anyway.
This means that if you have a vector of u128, summing the elements first and then casting the final sum to f64 will yield  pretty much the same result as if you first cast each element of the vector to f64 and then summed them.

ksew1:
```rs
fn main() {
    let nums: Vec<u128> = vec![1 << 120, 1, 1, 1]; // Large numbers

    // Sum as u128, then convert
    let sum_u128: u128 = nums.iter().sum();
    let result1: f64 = sum_u128 as f64;

    // Convert each to f64, then sum
    let result2: f64 = nums.iter().map(|&x| x as f64).sum();

    println!("(Sum as u128, then to f64)   :   {}", result1);
    println!("(Convert each to f64, then sum): {}", result2);
}
```

The output is the same
```bash
(Sum as u128, then to f64)   :   1329227995784916000000000000000000000
(Convert each to f64, then sum): 1329227995784916000000000000000000000

```

THenry14: that's why I asked: https://github.com/foundry-rs/starknet-foundry/pull/3054#discussion_r1989239397
cuz I'm not sure using the crate you suggested would have helped, right?

ksew1: Now everything is correct. The previous code was wrong because it performed integer division instead of floating-point division.  

The crate won't change anything now, but I was just thinking that if we had used it from the beginning, we wouldn't have needed to make this change.

crates/forge/tests/integration/gas.rs
cptartur: Can we extend the comment so it explains what you just wrote? It wasn't trivial for me.

crates/snforge-scarb-plugin/src/attributes/available_gas.rs
cptartur: I think at least `fork` supports multiple.

crates/snforge-scarb-plugin/src/attributes/available_gas.rs
cptartur: The way we handle max fee is to verify after the test execution is completed if the value didn't exceed max fee. Couldn't we type these bounds as `Option` inside forge and treat them as unlimited when checking gas if they are `None`?

cptartur (commented):
Responded to previous comments

crates/snforge-scarb-plugin/src/attributes/available_gas.rs
ddoktorski: > upon inspection, to make sure it fails because test succeeds and not something totally unrelated

Tbh it doesn't make much sense to me looking at the current logic [here](https://github.com/foundry-rs/starknet-foundry/pull/3054/files#diff-5f4c6e6bd283c1d085b4ec4d8dc055fcdb9f53820fa682274675aef306f3be7dR163) and actually the example above doesn't really check if the available gas was exceeded, it only checks whether this test fails or not. My guess it that the number you provided above for `l1_gas` exceeds `u64` and therefore it silently fails conversion to `GasVector` without asserting available gas at the end, but I am not sure. Could you check it again? 🙏

THenry14: lol, you might be onto something with the overflow, great catch 🤔 that means there is a bug somewhere in the code (my guess is attributes?) as **every test just passes if you provide a number big enough**, for any field - that is wild (I just confirmed it with latest released snforge as well, but since it used max_fee it probably was not that big of a problem up until now). Will try to find it, and keep you updated.

The question with paying with l1 for l2 still remains (and will result in the behaviour I described) - I think it's time for a decision - do we want it or not? connects to this thread https://github.com/foundry-rs/starknet-foundry/pull/3054#discussion_r1987593228

cptartur (commented):
Responded to comments

crates/snforge-scarb-plugin/src/attributes/available_gas.rs
THenry14: Oh shit, you're right. Sorry for my ignorance :( 
Modified the attribute as requested, and added a few tests in newest commit(s)

crates/forge/tests/integration/gas.rs
THenry14: I mean, all this is already in the top comment (with linked docs on calculations), I can add this here for one test as well, but I don't think adding it everywhere will bring any value tbh. 🤔 

Done in newest.

crates/snforge-scarb-plugin/src/attributes/available_gas.rs
THenry14: Ok, so I found the issue - Once again thanks Darek for diggint into this, I wouldn't have found it otherwise 🙏 

The plugin took a number from the attribute as felt, and then it was casted to BigInt. Then, forge took it and passed it to the vm, which failed silently (tldr it couldn't fit bigger values into GasAmount, so it defaulted to None and returned it to the forge). Forge then operated on a config, where available gas was not set (or rather - been set to None, instead of the value from the attribute), that's why the tests always passed.

This bug was there from the beggining, I guess no-one ever spotted this because up until now, if someone set max_fee larger that u64::MAX, it might as well be infinite. The problem manifested itself when using bounds, as setting one of the bounds as larger than u64::MAX nullified all other bounds, making the gas infinite for all of them.

This is now fixed on the plugin side, values are verified for both max_fee and bounds cases (although you can probably make a case to not doing it for max_fee, as the issue is not that severe there).

Regarding the second part of it, I have made sure for now that no conversion of l1 <-> l2 for payment will be made implicitly, so we should be good for time being and decide what to do next when starknet 0.14 is up.

docs/src/testing/test-attributes.md
```diff
@@ -72,7 +72,19 @@ If the test exceeds the limit, it fails with an appropriate error.
 
 #### Usage
 
-Asserts that the test does not use more than 5 units of gas.
+Asserts that the test does not use more than 5 units of l2 gas.
```
ddoktorski, line 75: nit:
```suggestion
Asserts that the test does not use more than 5 units of l2 gas:
```
Also in the sentences below

crates/forge-runner/src/gas.rs
```diff
@@ -163,11 +161,23 @@ pub fn check_available_gas(
             arguments,
             gas_info,
             ..
-        } if available_gas.is_some_and(|available_gas| gas_info > available_gas as u128) => {
+        } if available_gas.is_some_and(|available_gas| match available_gas {
+            RawAvailableGasConfig::MaxGas(gas) => {
+                gas_info.l1_gas + gas_info.l1_data_gas + gas_info.l2_gas > GasAmount(gas as u64)
```
ddoktorski, line 166: I am not sure if this logic makes sense to me. If we really want to continue allowing the unnamed argument to avoid a breaking change, shouldn't we treat `gas` as L1 gas, so first converting `gas_info` to L1 gas and then comparing? Overall I think it might be confusing either way and I would personally prefer to introduce this breaking change now (after all it has to happen now or later).

THenry14: This was actually discussed offline with @cptartur , and my understanding was that we want to avoid breaking change on the interface (attribute) level 🤔 

Just to make sure I understand you clear here - you do not mean to have support for l1 gas per se, just to convert gas_info to classic l1 gas just here for this check? if so, I can see a logic in this. One downside I see would be that as a user you get triplet representation everywhere, so it would not be obvious for you what to put in unnamed argument straight away if it is being represented as classic l1.

I have no opinion on this, whatever you guys decide is fine with me (as long as it's not full-on support for classic l1 gas, that idea is bad)

cptartur: I wanted to retain the old logic where user just passed a value to `#[available_gas]`. Previously it was also a sum, just that of l1 and l1 data gas (implicitly, we never showed them separately). This is to avoid breaking changes. 

We can even consider deprecating this at some point, though I still think it has use cases for users who just want to ensure that their logic doesn't exceed some max gas (fee) value.

crates/forge-runner/src/gas.rs
```diff
@@ -163,11 +161,23 @@ pub fn check_available_gas(
...
+                    || gas_info.l2_gas > av_gas.l2_gas
+            }
+        }) =>
+        {
             TestCaseSummary::Failed {
                 name,
                 msg: Some(format!(
-                    "\n\tTest cost exceeded the available gas. Consumed gas: ~{gas_info}"
+                    "\n\tTest cost exceeded the available gas. Consumed l1_gas: ~{}, l1_data_gas: ~{}, l2_gas: ~{}",
+                    gas_info.l1_gas, gas_info.l1_data_gas, gas_info.l2_gas
```
cptartur, line 180: Maybe if user used `RawAvailableGasConfig::MaxGas` then we could also show a total here?

crates/forge/tests/integration/available_gas.rs
```diff
@@ -8,7 +8,24 @@ fn correct_available_gas() {
...
+
+    assert_passed(&result);
+}
+
+#[test]
+fn correct_available_gas_with_unnamed_argument() {
+    let test = test_utils::test_case!(indoc!(
+        r"
+            #[test]
+            #[available_gas(440000)]
```
cptartur, line 28: How did this test work previously with just `11` available gas?

crates/forge/tests/integration/gas.rs
cptartur: Can we add a todo in the code?

crates/forge-runner/src/gas.rs
ddoktorski: > Just to make sure I understand you clear here - you do not mean to have support for l1 gas per se, just to convert gas_info to classic l1 gas just here for this check?

Yup, that's correct.

> One downside I see would be that as a user you get triplet representation everywhere, so it would not be obvious for you what to put in unnamed argument straight away if it is being represented as classic l1.

Absolutely agree, but I think the current logic is not obvious as well. I suggest this approach because, as a matter of fact `#available_gas` currently accepts L1 gas (even though it is not called like this), so it would be consistent this way.

crates/forge-runner/src/test_case_summary.rs
```diff
@@ -53,6 +72,16 @@ impl GasStatistics {
 
         (sum_squared_diff / gas_usages.len() as f64).sqrt()
     }
+
+    fn calculate_component(gas_usages: &[u128]) -> GasStatisticsComponent {
```
ksew1, line 76: [nit] I know this was my suggestion but i am thinking that it would be better if this was constructor method  on `GasStatisticsComponent`

crates/snforge-scarb-plugin/src/types.rs
```diff
@@ -51,9 +51,24 @@ impl ParseFromExpr<Expr> for Felt {
...
+    pub fn validate_in_gas_range<T: AttributeInfo>(
+        &self,
+        arg_name: &str,
+    ) -> Result<(), Diagnostic> {
+        let max = u64::MAX;
+        if *self > Number(max.into()) {
+            Err(T::error(format!(
+                "{arg_name} it too large (max permissible value is {max})"
+            )))?;
+        }
```
cptartur, line 67: Shouldn't we also verify that the user doesn't provide a negative value? Or is that impossible?

crates/snforge-scarb-plugin/src/types.rs
```diff
@@ -51,9 +51,24 @@ impl ParseFromExpr<Expr> for Felt {
...
 pub struct Number(pub(crate) BigInt);
 
+impl Number {
+    pub fn validate_in_gas_range<T: AttributeInfo>(
+        &self,
+        arg_name: &str,
+    ) -> Result<(), Diagnostic> {
+        let max = u64::MAX;
+        if *self > Number(max.into()) {
+            Err(T::error(format!(
```
ksew1, line 64: [nit]: using `?` adds unnecessary call to `From` and in my opinions is less readable. I would just write `return`

crates/snforge-scarb-plugin/tests/integration/single_attributes/available_gas.rs
```diff
@@ -4,56 +4,99 @@ use indoc::formatdoc;
 use snforge_scarb_plugin::attributes::available_gas::available_gas;
 
 #[test]
-fn fails_with_empty() {
+fn works_with_empty() {
```
cptartur, line 7: Do we want to allow this? I'm not sure what was the use case of empty available gas previously.

cptartur: Maybe we could at least deprecate that?

crates/forge-runner/src/gas.rs
THenry14: Done

crates/forge-runner/src/test_case_summary.rs
```diff
@@ -28,20 +36,31 @@ pub struct GasStatistics {
...
-            mean,
-            std_deviation: Self::std_deviation(mean, gas_usages),
+            l1_gas: { Self::calculate_component(l1_gas_values.as_ref()) },
+            l1_data_gas: { Self::calculate_component(l1_data_gas_values.as_ref()) },
+            l2_gas: { Self::calculate_component(l2_gas_values.as_ref()) },
         }
     }
 
     #[expect(clippy::cast_precision_loss)]
     fn mean(gas_usages: &[u128]) -> f64 {
```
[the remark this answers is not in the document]
THenry14: I couldn't be arsed to be honest, so I asked AI to do this 🙈 thanks for the suggestion.

crates/forge/tests/integration/available_gas.rs
THenry14: it was in some of the threads, it was previously l1, this is (was) l2. As we discussed, I have changed the unnamed to be classic l1, so this test is now back to 11.

crates/snforge-scarb-plugin/src/types.rs
THenry14: it's impossible - it's on scarb plugin level afaik (if you try that, some error diagnostic will be raised as far as I remember)

crates/snforge-scarb-plugin/tests/integration/single_attributes/available_gas.rs
THenry14: It works as if the attribute was not set basically, as then available_gas is set to max

ksew1 (approved):

crates/forge/tests/integration/fuzzing.rs
```diff
@@ -76,8 +78,16 @@ fn fuzzed_while_loop() {
...
+    assert!(gas_info.l1_gas.mean < ALLOWED_ERROR);
+    assert!(gas_info.l1_gas.std_deviation < ALLOWED_ERROR);
+    assert_eq!(gas_info.l1_data_gas.min, 0);
+    assert_eq!(gas_info.l1_data_gas.max, 0);
+    assert!(gas_info.l1_data_gas.mean < ALLOWED_ERROR);
+    assert!(gas_info.l1_data_gas.std_deviation < ALLOWED_ERROR);
+    assert_eq!(gas_info.l2_gas.min, 80000);
+    assert_eq!(gas_info.l2_gas.max, 920_000);
+    assert!((gas_info.l2_gas.mean - 504_218.75).abs() < f64::EPSILON);
+    assert!((gas_info.l2_gas.std_deviation - 248_434.50).abs() < ALLOWED_ERROR);
```
ddoktorski, line 92: Please either assert all l2 gas values with a higher margin or run this test for the specific Scarb version using a feature 🙏 In its current form, this test will fail our scheduled tests.

crates/snforge-scarb-plugin/tests/integration/single_attributes/available_gas.rs
```diff
@@ -82,20 +129,159 @@ fn work_with_number() {
...
+
+    let result = available_gas(args, item);
+
+    assert_diagnostics(
+        &result,
+        &[Diagnostic::error(formatdoc!(
+            "
+                All options failed
+                - variant: <l2_gas> argument was specified 2 times, expected to be used only once
+                - variant: #[available_gas] can be used with unnamed arguments only
```
ddoktorski, line 143: This error message is a bit vague. Can we explicitly mention that it can be used with a single unnamed argument meaning L1 gas or something like that?

The same applies to the error `#[available_gas] can be used with named arguments only`.

crates/forge-runner/src/gas.rs
```diff
@@ -163,11 +163,31 @@ pub fn check_available_gas(
             arguments,
             gas_info,
             ..
-        } if available_gas.is_some_and(|available_gas| gas_info > available_gas as u128) => {
+        } if available_gas.is_some_and(|available_gas| match available_gas {
+            RawAvailableGasConfig::MaxGas(gas) => {
+                print_as_warning(&anyhow!(
+                    "Setting available_gas with unnamed argument is deprecated. \
+                Consider setting resource bounds (l1_gas, l1_data_gas and l2_gas) explicitly."
```
cptartur, line 170: Please create an issue to remove this following deprecation (can be tentatively scheduled for iteration 9) and link it here.

crates/snforge-scarb-plugin/tests/integration/single_attributes/available_gas.rs
cptartur: For me it's weird to support this, but if we also did support it previously then probably doesn't matter.

cptartur (approved):

crates/snforge-scarb-plugin/tests/integration/single_attributes/available_gas.rs
THenry14: https://github.com/foundry-rs/starknet-foundry/issues/3108

crates/forge-runner/src/gas.rs
THenry14: https://github.com/foundry-rs/starknet-foundry/issues/3109

Issue #2960, opened 2025-02-20
Verify correctness of asserted gas value in `l1_handler_cost` test


franciszekjob:
### Current State

While bumping blockifier to 0.14.0-rc.1, asserted gas value in `l1_handler_cost` has changed. The calculations are a bit different to what is currently shown in the [message cost docs](https://docs.starknet.io/architecture-and-concepts/network-architecture/fee-mechanism/#l_2-l_1_messages). We should verify if our calculations are correct (docs may be outdated).

### Objective

Verify correctness of asserted gas value in `l1_handler_cost` test

cptartur:
@ddoktorski relevant to gas tests

PR #3055, opened 2025-03-06
Add gas tests for sierra-gas tracked resource

THenry14:
Related to #2977

-- (https://github.com/foundry-rs/starknet-foundry/pull/3053) Add sierra version assertion
-- (https://github.com/foundry-rs/starknet-foundry/pull/3054) Introduce new gas representation
--> (This PR) Add gas tests for sierra-gas tracking

### DIFF
--- a/crates/forge/tests/integration/gas.rs
+++ b/crates/forge/tests/integration/gas.rs
@@ -1006,3 +1006,574 @@ fn events_contract_cost_cairo_steps() {
         },
     );
 }
+
+#[test]
+#[cfg_attr(not(feature = "scarb_since_2_10"), ignore)]
+fn declare_cost_is_omitted_sierra_gas() {
+    let test = test_case!(
+        indoc!(
+            r#"
+            use snforge_std::declare;
+            #[test]
+            fn declare_cost_is_omitted() {
+                declare("GasChecker").unwrap();
+            }
+        "#
+        ),
+        Contract::from_code_path(
+            "GasChecker".to_string(),
+            Path::new("tests/data/contracts/gas_checker.cairo"),
+        )
+        .unwrap()
+    );
+
+    let result = run_test_case(&test, ForgeTrackedResource::SierraGas);
+
+    assert_passed(&result);
+    // 23860 = reported consumed sierra gas
+    // 0 l1_gas + 0 l1_data_gas + 23860 l2 gas
+    assert_gas(
+        &result,
+        "declare_cost_is_omitted",
+        GasVector {
+            l1_gas: GasAmount(0),
+            l1_data_gas: GasAmount(0),
+            l2_gas: GasAmount(23860),
+        },
+    );
+}
+
+#[test]
+#[cfg_attr(not(feature = "scarb_since_2_10"), ignore)]
+fn deploy_syscall_cost_sierra_gas() {
+    let test = test_case!(
+        indoc!(
+            r#"
+            use core::clone::Clone;
+            use snforge_std::{declare, DeclareResultTrait};
+            use starknet::{SyscallResult, deploy_syscall};
+            #[test]
+            fn deploy_syscall_cost() {
+                let contract = declare("GasConstructorChecker").unwrap().contract_class().clone();
+                let (address, _) = deploy_syscall(contract.class_hash, 0, array![].span(), false).unwrap();
+                assert(address != 0.try_into().unwrap(), 'wrong deployed addr');
+            }
+        "#
+        ),
+        Contract::from_code_path(
+            "GasConstructorChecker".to_string(),
+            Path::new("tests/data/contracts/gas_constructor_checker.cairo"),
+        )
+        .unwrap()
+    );
+
+    let result = run_test_case(&test, ForgeTrackedResource::SierraGas);
+
+    assert_passed(&result);
+    // l = 1 (updated contract class)
+    // n = 1 (unique contracts updated - in this case it's the new contract address)
+    // ( l + n * 2 ) * felt_size_in_bytes(32) = 96 (total l1 data cost)
+    //
+    // 20000 = cost of 2 keccak syscall (because 2 * 100 * 100) (from constructor)
+    //      -> 1 keccak syscall costs 100 cairo steps
+    // 142810 = cost of 1 deploy syscall (because 1 * 1132 * 100 + 7 * 4050 + 18 * 70)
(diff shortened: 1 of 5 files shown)


ksew1 (approved):

ddoktorski (changes requested):
Even with Scarb `2.11.1` some of these tests already fail. Based on our experience so far, such tests seem unmaintainable in the long run. I don't have a specific idea on how to test gas related features, but could we at least assert all values with some acceptable margin? 🙏

cc @cptartur

crates/forge/tests/integration/gas.rs
```diff
@@ -1002,3 +1002,574 @@ fn events_contract_cost_cairo_steps() {
...
+    );
+
+    let result = run_test_case(&test, ForgeTrackedResource::SierraGas);
+
+    assert_passed(&result);
+    // l = 1 (updated contract class)
+    // n = 1 (unique contracts updated - in this case it's the new contract address)
+    // ( l + n * 2 ) * felt_size_in_bytes(32) = 96 (total l1 data cost)
+    //
+    // 20000 = cost of 2 keccak syscall (because 2 * 100 * 100) (from constructor)
```
cptartur, line 1073: Maybe we add some units to it, so it's more obvious: keccak is 100 steps, and each step is 100 sierra gas. 

And that the price of keccak is from versioned constants.

It will be much easier to maintain this in the future

crates/forge/tests/integration/gas.rs
```diff
@@ -1002,3 +1002,574 @@ fn events_contract_cost_cairo_steps() {
...
+        .unwrap()
+    );
+
+    let result = run_test_case(&test, ForgeTrackedResource::SierraGas);
+
+    assert_passed(&result);
+    // 96 = gas cost of onchain data (see `deploy_syscall_cost_sierra_gas` test)
+    // 20000 = cost of 2 keccak syscall (see `deploy_syscall_cost_sierra_gas` test)
+    // 142810 = cost of 1 deploy syscall (see `deploy_syscall_cost_sierra_gas` test)
+    // 487334 = reported consumed sierra gas
```
cptartur, line 1119: So this is just from code execution that doesn't directly translate to anything specific (like syscalls etc.)?

crates/forge/tests/integration/gas.rs
```diff
@@ -1002,3 +1002,574 @@ fn events_contract_cost_cairo_steps() {
...
+            "GasChecker".to_string(),
+            Path::new("tests/data/contracts/gas_checker.cairo"),
+        )
+        .unwrap()
+    );
+
+    let result = run_test_case(&test, ForgeTrackedResource::SierraGas);
+
+    assert_passed(&result);
+    // 29524 = gas cost of l2 -> l1 message
```
cptartur, line 1399: I assume this was just taken from what was reported back from the gas calculation code?

crates/forge/tests/integration/gas.rs
```diff
@@ -1002,3 +1002,574 @@ fn events_contract_cost_cairo_steps() {
...
+    //      -> we emit 50 keys, each taking up 1 felt of space
+    //      -> L2 gas cost for event key is 10240 gas/felt
+    // 256000 = 50 * 5120
+    //      -> we emit 50 keys, each having 1 felt of data
+    //      -> L2 gas cost for event data is 5120 gas/felt
+    // 10000 = cost of 1 emit event syscall (because 1 * 61 * 100 + 1 * 70 = 6170)
+    //      -> 1 emit event syscall costs 61 cairo steps and 1 range check builtin
+    //      -> 1 range check costs 70
+    //      -> the minimum total cost is `syscall_base_gas_cost`, which is pre-charged by the compiler (atm it is 100 * 100)
+    // 186880 = reported consumed sierra gas
```
cptartur, line 1517: Isn't this quite substantial compared to other costs (that were calculated)? I wonder if this isn't caused by not including something in the calculations.

crates/forge/tests/integration/gas.rs
THenry14: All this information is already provided 🤔 I mean, I see your point, but there are things in the top comment for a reason - to not repeat all the same stuff over and over. I have ammended the top comment with information that:
```
Costs of syscalls (if provided) are taken from versioned_constants (blockifier)
```

to be super-duper clear, but really I find this info redundant at this point tbh. However, you're the boss here, and if you insist I can add it ofc 🙏

crates/forge/tests/integration/gas.rs
THenry14: Yup

crates/forge/tests/integration/gas.rs
THenry14: Not necessarily - you must remember this is `l2_gas` for sierra steps. If you were to convert this into "classic" gas values, you'd see that its cost is equivalent to ~31 gas (and the cost before any of my changes is 30, slight increase but this is step heavy so this is somewhat expected).

To be super clear where did I get the numbers from:

l2 gas reported here (so the measure of execution cost) is 1239010 for sierra steps and 1208000 for cairo steps (see assert_gas for `events_contract_cost_sierra_gas` and `events_contract_cost_cairo_steps` respectively).

Converting this to classical L1 gas can be done from simple proportion:

100 L2 gas = 0.0025 L1 Gas
1239010 L2 gas = X 
(so X = 30.98 L1 gas)

1208000 L2 gas = X
(so X = 30.2 L1 gas)

As we can see from the same test on master, the numbers checks out - https://github.com/foundry-rs/starknet-foundry/blob/master/crates/forge/tests/integration/gas.rs#L786 :
`11 + 6 + 13` = 30 (we discard 96, as this is not related to execution - this is actually l1_data_gas, that is now taken by use in a separate resource bound; we can add it here, but we would also have to add it to calculations above; and if we did the numbers would still have to match, as we add the same number in both places).


Hope that clarifies things :+1:

crates/forge/tests/integration/gas.rs
cptartur: Maybe it makes sense to link this https://github.com/foundry-rs/starknet-foundry/issues/2960 issue here and in related?

crates/forge/tests/integration/gas.rs
cptartur: The cost `11+6+13=30` is for the test `events_contract_cost` while here we are in the equivalent of `events_cost`.

But here the converted value would be `~24` from the 26 on master which is also in line.

Seems fine for me.

cptartur (approved):

Issue #2858, opened 2025-01-17
Rewrite snforge plugin to new macro syntax


cptartur:

cptartur:
Not yet released

PR #3059, opened 2025-03-06
Assert plugin outputs with Cairo formatter

ddoktorski:
Towards #2858 

## Introduced changes

- With new macro syntax, the returned code is formatted differently than before, and our current logic does not normalize it correctly for asserts. Instead of replacing whitespaces with a single space, apply the Cairo formatter and remove newlines.


franciszekjob (approved):
👍

ksew1 (approved):

cptartur (approved):

Issue #2490, opened 2024-09-19
Support RPC version 0.8.0


cptartur:
### Which components does the task require to be changed? (think hard pls)

sncast

### Description

Release candidate was already published https://github.com/starkware-libs/starknet-specs/releases/tag/v0.8.0-rc0
We need to wait at least until starknet-rs is bumped and the spec is implemented in integration/devnet.

PR #3064, opened 2025-03-08
Support RPC 0.8.0

franciszekjob:
Towards #2490 

## Introduced changes

Bump RPC to 0.8.0:
- Introduce `--l1-gas`, `--l1-gas-price`, `--l2-gas`, `--l2-gas-price`, `--l1-data`, `--l1-data-gas` flags
- Remove `--max-gas`, `--max-gas-unit-price` flags
- Update RPC errors

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/sncast/tests/e2e/declare.rs
```diff
@@ -82,8 +85,18 @@ async fn test_happy_case(class_hash: Felt, account_type: AccountType) {
...
+        "--l1-gas-price",
+        "10000000000000",
+        "--l2-gas",
+        "1000000000",
+        "--l2-gas-price",
+        "100000000000000000000",
+        "--l1-data-gas",
+        "100000",
+        "--l1-data-gas-price",
+        "10000000000000",
```
franciszekjob, line 99: note: Using auto estimation fails because fee estimate returns insufficient resources (it's on devnet side). Due to it, we use fixed resource bounds in our tests (applies to all other places).

franciszekjob:
note: Docs and scripts tests are failing, they will be fixed in subsequent PRs

scripts/install_devnet.sh
```diff
@@ -2,8 +2,9 @@
 set -e
 
 DEVNET_INSTALL_DIR="$(git rev-parse --show-toplevel)/crates/sncast/tests/utils/devnet"
-DEVNET_REPO="https://github.com/0xSpaceShard/starknet-devnet-rs"
-DEVNET_VER="v0.2.2"
+# TODO(#3087): Use official devnet release
```
ddoktorski, line 5: Why not [v0.3.0-rc.0 release](https://github.com/0xSpaceShard/starknet-devnet/releases/tag/v0.3.0-rc.0)?

crates/sncast/src/helpers/constants.rs
```diff
@@ -20,7 +20,8 @@ pub const UDC_ADDRESS: Felt =
 pub const OZ_CLASS_HASH: Felt =
     felt!("0x00e2eb8f5672af4e6a4e8a8f1b44989685e668489b0a25437733756c5a34a1d6");
 pub const ARGENT_CLASS_HASH: Felt =
-    felt!("0x029927c8af6bccf3f6fda035981e765a7bdbf18a2dc0d630494f8758aa908e2b");
+    // felt!("0x029927c8af6bccf3f6fda035981e765a7bdbf18a2dc0d630494f8758aa908e2b");
+    felt!("0x036078334509b514626504edc9fb252328d1a240e4e948bef8d0c08dff45927f");
```
ddoktorski, line 24: Why this class hash?

crates/sncast/src/helpers/fee.rs
```diff
@@ -1,117 +1,174 @@
...
-                    max_gas_unit_price: Some(
-                        NonZeroU128::try_from_(max_gas_unit_price).map_err(anyhow::Error::msg)?,
-                    ),
-                }
+    pub fn try_into_fee_settings(&self, fee_estimate: &Option<FeeEstimate>) -> Result<FeeSettings> {
+        // If some resource bounds values are lacking, starknet-rs will estimate them automatically
+        // but in case someone passes --max-fee flag, we need to make estimation on our own
+        // to check if the fee estimate isn't higher than provided max fee
+        if let Some(max_fee) = self.max_fee {
+            self.check_conflicting_flags()?;
```
ddoktorski, line 67: It should be validated by clap

crates/sncast/src/helpers/fee.rs
```diff
@@ -1,117 +1,174 @@
...
+            self.validate_max_fee_meets_resource_bounds()?;
+
+            ensure!(
+                fee_estimate.is_some(),
+                "Fee estimate is required when passing max fee is provided"
+            );
```
ddoktorski, line 73:
```suggestion
            let fee_estimate = fee_estimate.clone().ok_or_else(|| {
                anyhow!("Fee estimate must be calculated when max_fee is provided")
            })?;
```

scripts/install_devnet.sh
franciszekjob: It says:
> This is a pre-release serving JSON-RPC API v0.7.1 and internally executing transactions using blockifier v0.14.0-rc.0. This means it's mostly adapted to Starknet 0.13.4, with the exception of L1-data + L2 gas semantics, which are tied to JSON-RPC v0.8.0.

crates/sncast/src/response/errors.rs
```diff
@@ -73,14 +75,14 @@ pub enum SNCastStarknetError {
     ClassAlreadyDeclared,
     #[error("Invalid transaction nonce")]
     InvalidTransactionNonce,
-    #[error("Max fee is smaller than the minimal transaction cost")]
-    InsufficientMaxFee,
+    #[error("The transaction's resources don't cover validation or the minimal transaction fee")]
```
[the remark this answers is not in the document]
franciszekjob: This is taken from [spec](https://github.com/starkware-libs/starknet-specs/blob/4b351773f92feb7a2b1b9a754ff525ee0ba194bb/api/starknet_write_api.json#L239) 😅 .

crates/sncast/src/starknet_commands/multicall/mod.rs
```diff
@@ -15,6 +15,6 @@ pub struct Multicall {
 
 #[derive(Debug, Subcommand)]
 pub enum Commands {
-    Run(Run),
+    Run(Box<Run>),
```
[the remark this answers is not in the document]
franciszekjob: It was suggested by clippy, but it's not needed anymore. Removed.

crates/sncast/src/helpers/fee.rs
```diff
@@ -1,117 +1,174 @@
...
-                let max_gas = NonZeroFelt::try_from(Felt::from(max_fee)
-                            .floor_div(&max_gas_unit_price)).context("Calculated max-gas from provided --max-fee and the current network gas price is 0. Please increase --max-fee to obtain a positive gas amount")?;
-                print_max_fee_conversion_info(max_fee, max_gas, max_gas_unit_price);
-                FeeSettings {
-                    max_gas: Some(NonZeroU64::try_from_(max_gas).map_err(anyhow::Error::msg)?),
-                    max_gas_unit_price: Some(
-                        NonZeroU128::try_from_(max_gas_unit_price).map_err(anyhow::Error::msg)?,
-                    ),
-                }
+    pub fn try_into_fee_settings(&self, fee_estimate: &Option<FeeEstimate>) -> Result<FeeSettings> {
```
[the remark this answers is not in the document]
franciszekjob: In fact, clippy enforces to use reference.

crates/sncast/src/helpers/fee.rs
franciszekjob: Not sure if that's possible  on clap level- we want to disallow passing all together, but still allow to pass any other subset. `conflicts_with_all` seemd helpful, but after all it doesn't allow for that.

crates/sncast/src/helpers/constants.rs
franciszekjob: I've restored it to previous version 0(.3.1). However some tests fail with it (there's no such problem with 0.4.0 which was present the change above). Will be addressed in #3089

crates/conversions/src/serde/serialize/serialize_impl.rs
```diff
@@ -27,14 +29,28 @@ impl CairoSerialize for CallEntryPoint {
...
+    fn serialize(&self, output: &mut BufferWriter) {
+        match &self {
+            ContractExecutionError::Nested(inner) => {
+                inner.class_hash.serialize(output);
+                inner.contract_address.serialize(output);
+                inner.selector.serialize(output);
+                inner.error.serialize(output);
+            }
+            ContractExecutionError::Message(msg) => ByteArray::from(msg.as_str()).serialize(output),
+        }
```
cptartur, line 46: Do we actually test these anywhere? I think it would be worth doing so.

crates/forge/tests/e2e/fork_warning.rs
```diff
@@ -39,6 +39,7 @@ fn setup_redirect_server() {
 }
 
 #[test]
+#[ignore = "TODO(#3088)"]
```
cptartur, line 42: The issue for this is closed, shouldn't these be restored?

crates/sncast/src/helpers/fee.rs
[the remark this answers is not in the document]
cptartur: Let's use `Option<&FeeEstimate>` it's more idiomatic.

crates/sncast/src/helpers/fee.rs
cptartur: I don't think we need this kind of level of granularity of settings fees: For me user should be either allowed to pass `--max-fee` or specific args for gas. I don't see use cases for setting max fee along specific gas values in a command line too.

This would make it far less complex.

cptartur: And we could validate everything at clap level.

crates/sncast/src/helpers/fee.rs
```diff
@@ -120,27 +188,22 @@ impl FeeArgs {
 #[derive(Debug, PartialEq, CairoDeserialize)]
 pub struct ScriptFeeSettings {
```
cptartur, line 189: At cast script level this should take an enum
```
pub enum ScriptFeeSettings {
  MaxFee(value),
  GasTriplet(...)
}
```

That's a feature for future versions but please create an issue for that.

crates/sncast/src/lib.rs
```diff
@@ -591,11 +596,11 @@ pub async fn wait_for_tx(
...
             ) => match execution_status {
-                starknet::core::types::TransactionExecutionStatus::Succeeded => {
-                    return Ok("Transaction accepted");
+                starknet::core::types::ExecutionResult::Succeeded => {
+                    return Ok("Transaction accepted".to_string());
                 }
-                starknet::core::types::TransactionExecutionStatus::Reverted => {
-                    return get_revert_reason(provider, tx_hash).await;
+                starknet::core::types::ExecutionResult::Reverted { reason } => {
+                    return Ok(reason);
```
cptartur, line 603: This isn't correct, but is being resolved by cast scripts PR.

crates/sncast/src/helpers/fee.rs
cptartur: Something along the lines of this

```
// Arg group "resources"
    pub l1_gas: Felt,
    pub l1_gas_price: Felt,
    pub l2_gas: Felt,
    pub l2_gas_price: Felt,
    pub l1_data_gas: Felt
    pub l1_data_gas_price: Felt,


// FeeArgs
max_fee: Felt // conflicts with arg group resources
resources: Resources // conflicts with max fee
```

crates/sncast/src/starknet_commands/declare.rs
```diff
@@ -78,24 +73,39 @@ pub async fn declare(
...
+        l2_gas_price,
+        l1_data_gas,
+        l1_data_gas_price,
+    } = fee_settings.expect("Failed to convert to fee settings");
+
+    let declaration = apply_optional(declaration, l1_gas, DeclarationV3::l1_gas);
+    let declaration = apply_optional(declaration, l1_gas_price, DeclarationV3::l1_gas_price);
+    let declaration = apply_optional(declaration, l2_gas, DeclarationV3::l2_gas);
+    let declaration = apply_optional(declaration, l2_gas_price, DeclarationV3::l2_gas_price);
+    let declaration = apply_optional(declaration, l1_data_gas, DeclarationV3::l1_data_gas);
```
cptartur, line 104: This logic is repeated in multiple places, can't this be extracted to some generic helper function?

cptartur: Or a macro if that's needed

crates/sncast/src/starknet_commands/multicall/mod.rs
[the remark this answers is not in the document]
cptartur: What's the lint name? Seems unusual.

crates/sncast/src/starknet_commands/multicall/run.rs
```diff
@@ -56,7 +56,7 @@ struct InvokeCall {
 }
 
 pub async fn run(
-    run: Run,
+    run: Box<Run>,
```
cptartur, line 59: Same here

crates/sncast/tests/e2e/account/deploy.rs
```diff
@@ -38,11 +40,10 @@ pub async fn test_happy_case(class_hash: &str, account_type: &str) {
...
         "my_account",
-        "--max-gas",
-        "1000",
-        "--max-gas-unit-price",
-        "100000000000",
-    ];
+    ]
+    .into_iter()
+    .chain(TEST_RESOURCE_BOUNDS_FLAGS.into_iter())
+    .collect::<Vec<&str>>();
```
cptartur, line 46: Nit, but maybe this could be hidden behind some helper function?

crates/sncast/tests/e2e/account/deploy.rs
```diff
@@ -217,34 +190,34 @@ pub async fn test_valid_class_hash() {
...
+//     let snapbox = runner(&args).current_dir(tempdir.path());
+
+//     snapbox.assert().success().stdout_matches(indoc! {r"
+//         command: account deploy
+//         transaction_hash: [..]
+
+//         To see invocation details, visit:
+//         transaction: [..]
+//     "});
+// }
```
cptartur, line 220: Please restore it or ignore it with a linked issue

crates/sncast/tests/e2e/call.rs
```diff
@@ -160,11 +169,12 @@ fn test_wrong_calldata() {
...
     let output = snapbox.assert().success();
 
-    assert_stderr_contains(
+    // 0x496e70757420746f6f206c6f6e6720666f7220617267756d656e7473 is "Input too long for arguments"
+    assert_stdout_contains(
         output,
         indoc! {r"
         command: call
-        error: An error occurred [..]Execution failed[..]Input too long for arguments[..]
+        response: [0x496e70757420746f6f206c6f6e6720666f7220617267756d656e7473]
```
cptartur, line 177: Why isn't it decoded now?

crates/sncast/tests/e2e/declare.rs
```diff
@@ -95,68 +100,69 @@ async fn test_happy_case(class_hash: Felt, account_type: AccountType) {
...
-
-    let output = output.get_output().stdout.clone();
-
-    let hash = get_transaction_hash(&output);
-    let receipt = get_transaction_receipt(hash).await;
-
-    assert!(matches!(receipt, Declare(_)));
-}
+// TODO(#3090)
+// #[test_case(Some("100000000000000000"), None, None; "max_fee")]
```
cptartur, line 104: I think that's a big one to actually ignore, I'd prefer we restore it in some form that before we release.

cptartur: Also let's uncomment this and ignore instead of making it commented out.

crates/sncast/tests/e2e/deploy.rs
```diff
@@ -91,56 +94,57 @@ async fn test_happy_case(class_hash: Felt, account_type: AccountType) {
...
-    }
-
-    let snapbox = runner(&args).current_dir(tempdir.path());
-    let output = snapbox.assert().success().get_output().stdout.clone();
-    let hash = get_transaction_hash(&output);
-    let receipt = get_transaction_receipt(hash).await;
-
-    assert!(matches!(receipt, Deploy(_)));
-}
+// TODO(#3090)
```
cptartur, line 97: Uncomment and ignore

crates/sncast/tests/e2e/invoke.rs
```diff
@@ -91,67 +97,61 @@ async fn test_happy_case(class_hash: Felt, account_type: AccountType) {
...
-
-    let snapbox = runner(&args).current_dir(tempdir.path());
-    let output = snapbox.assert().success().get_output().stdout.clone();
-
-    let hash = get_transaction_hash(&output);
-    let receipt = get_transaction_receipt(hash).await;
-
-    assert!(matches!(receipt, Invoke(_)));
-}
+// TODO(#3090)
```
cptartur, line 103: Uncomment and ignore

crates/sncast/tests/e2e/invoke.rs
```diff
@@ -267,78 +275,45 @@ fn test_too_low_gas() {
...
+
+//     // TODO(#2852)
+//     assert_stderr_contains(
+//         output,
+//         indoc! {r"
+//         command: invoke
+//         error: Calculated max-gas from provided --max-fee and the current network gas price is 0. Please increase --max-fee to obtain a positive gas amount: Tried to create NonZeroFelt from 0
+//         "},
+//     );
+// }
```
cptartur, line 316: Uncomment (and ignore with a linked issue if needed). Or remove if it is no longer relevant (kind of seems like it).

crates/sncast/tests/helpers/constants.rs
```diff
@@ -23,7 +24,7 @@ pub const DEVNET_PREDEPLOYED_ACCOUNT_ADDRESS: &str =
 
 // OpenZeppelin account contract v0.8.1
 pub const DEVNET_OZ_CLASS_HASH_CAIRO_1: Felt =
-    felt!("0x061dac032f228abef9c6626f995015233097ae253a7f72d68552db02f2971b8f");
+    felt!("0x044cab2e6a3a7bc516425d06d76c6ffd56ae308864dbc66f8e75028e3784aa29");
```
cptartur, line 27: Class hash changed but the comment above says the version is the same? Why is that?

crates/sncast/tests/helpers/fixtures.rs
```diff
@@ -215,15 +224,37 @@ pub async fn invoke_contract(
...
+    let execution = apply_optional(
+        execution,
+        fee_settings.l1_data_gas,
+        ExecutionV3::l1_data_gas,
+    );
+    let execution = apply_optional(
+        execution,
+        fee_settings.l1_data_gas_price,
+        ExecutionV3::l1_data_gas_price,
+    );
```
cptartur, line 249: If we create a helper for these it can be used here as well

cptartur (changes requested):

crates/sncast/src/helpers/fee.rs
franciszekjob: > I don't think we need this kind of level of granularity of settings fees: For me user should be either allowed to pass > --max-fee or specific args for gas. I don't see use cases for setting max fee along specific gas values in a > command line too.

> This would make it far less complex.

My initial idea was to allow using max fee with other args, but if we want to simplify it then that's fine.

crates/sncast/src/starknet_commands/multicall/mod.rs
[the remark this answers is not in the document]
franciszekjob: https://rust-lang.github.io/rust-clippy/master/index.html#/large_enum_variant

Seems it's not needed now, as the size is ok.

crates/sncast/tests/e2e/account/deploy.rs
franciszekjob: Sure, this should be better:

```rust
let args = vec![
        "--accounts-file",
        "accounts.json",
        "--account",
        "my_account",
        "deploy",
        "--url",
        URL,
        "--class-hash",
        MAP_CONTRACT_CLASS_HASH_SEPOLIA,
        "--salt",
        "0x2",
        "--unique",
    ];
    let args = apply_test_resource_bounds_flags(args);
```

crates/sncast/tests/e2e/invoke.rs
franciszekjob: Refactored it and linked #3100 . The test case with max fee has been commented out as it's not possible to ignore a specific one.

crates/sncast/tests/e2e/deploy.rs
franciszekjob: Refactored it and linked #3100 . The test case with max fee has been commented out as it's not possible to ignore a specific one.

crates/sncast/tests/e2e/declare.rs
franciszekjob: I restored it with resource bounds case, the one with max fee is ignored (linked to #3100).

crates/sncast/src/starknet_commands/declare.rs
franciszekjob: It can, sure - added a macro for that.

crates/sncast/tests/helpers/fixtures.rs
franciszekjob: Same as above, macro `apply_optional_fields` handles that now.

crates/conversions/src/serde/serialize/serialize_impl.rs
franciszekjob: I made output assertion in `test_wrong_function_name` and `test_wrong_calldata` more detailed, so now it checks the serialization, take a look.

crates/sncast/tests/e2e/call.rs
franciszekjob: It's coming from devnet (which gets that from blockifier). Previously it was returned as error, now it's a response (not sure why, it's not because of our codebase). Create #3107 for that.

crates/sncast/tests/helpers/constants.rs
franciszekjob: The comment with version should be changed, but tbh I don't think it's worth keeping at all, we don't do that for Cairo 0, nor for argent or braavos. Wdyt?

crates/sncast/src/helpers/fee.rs
```diff
@@ -1,146 +1,179 @@
...
+
+            ensure!(
+                Felt::from(max_fee) >= fee_estimate.overall_fee,
+                "Estimated fee ({}) is higher than provided max fee ({})",
+                fee_estimate.overall_fee,
+                Felt::from(max_fee)
+            );
 
-        Ok(settings)
+            let fee_settings = FeeSettings {
```
ddoktorski, line 80: Why not to implement `TryFrom<FeeEstimate>` and `TryFrom<FeeArgs>` for `FeeSettings`?

crates/sncast/src/response/errors.rs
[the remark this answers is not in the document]
ddoktorski: Up to you

crates/sncast/src/starknet_commands/account/create.rs
```diff
@@ -181,8 +181,9 @@ async fn generate_account(
         }
         AccountType::Argent => {
             let factory =
-                ArgentAccountFactory::new(class_hash, chain_id, Felt::ZERO, signer, provider)
+                ArgentAccountFactory::new(class_hash, chain_id, Some(Felt::ZERO), signer, provider)
```
ddoktorski, line 184: This is for a guardian public key. I guess they changed the interface and now it should be `None`, but please check this 🙏 
```suggestion
                ArgentAccountFactory::new(class_hash, chain_id, None, signer, provider)
```

ddoktorski: Perhaps this is related to issues with argent accounts that you mentioned

crates/sncast/src/starknet_commands/account/deploy.rs
```diff
@@ -194,7 +194,7 @@ async fn get_deployment_result(
             let factory = ArgentAccountFactory::new(
                 class_hash,
                 chain_id,
-                Felt::ZERO,
+                Some(Felt::ZERO),
```
ddoktorski, line 197: Same as above

crates/sncast/src/lib.rs
```diff
@@ -731,6 +718,19 @@ pub fn apply_optional<T, R, F: FnOnce(T, R) -> T>(initial: T, option: Option<R>,
     }
 }
 
+#[macro_export]
+macro_rules! apply_optional_fields {
```
cptartur, line 722: No way apply_optional got a sequel

crates/sncast/tests/helpers/constants.rs
cptartur: Okay, we can remove it, better than provide incorrect informations.

crates/sncast/src/starknet_commands/deploy.rs
```diff
@@ -92,35 +92,45 @@ pub async fn deploy(
...
+    //     None => execution,
+    //     Some(l1_gas_price) => execution.l1_gas_price(l1_gas_price),
+    // };
+    // let execution = match l2_gas {
+    //     None => execution,
+    //     Some(l2_gas) => execution.l2_gas(l2_gas),
+    // };
+    // let execution = match l2_gas_price {
+    //     None => execution,
+    //     Some(l2_gas_price) => execution.l2_gas_price(l2_gas_price),
```
cptartur, line 119: Please remove the commented out code

crates/sncast/src/starknet_commands/deploy.rs
```diff
@@ -92,35 +92,45 @@ pub async fn deploy(
...
-    let result = execution.send().await;
+    let deployment = apply_optional_fields!(
+        deployment,
+        l1_gas => DeploymentV3::l1_gas,
+        l1_gas_price => DeploymentV3::l1_gas_price,
+        l2_gas => DeploymentV3::l2_gas,
+        l2_gas_price => DeploymentV3::l2_gas_price,
+        l1_data_gas => DeploymentV3::l1_data_gas,
+        l1_data_gas_price => DeploymentV3::l1_data_gas_price,
+        nonce => DeploymentV3::nonce
```
cptartur, line 103: Assuming that the commented out code was just condensed to that?

crates/sncast/tests/docs_snippets/validation.rs
```diff
@@ -65,14 +64,6 @@ fn test_docs_snippets() {
...
         }
 
-        // check if one of "declare", "deploy" or "invoke" is included in args
-        if args
-            .iter()
-            .any(|arg| *arg == "declare" || *arg == "deploy" || *arg == "invoke" || *arg == "run")
-        {
-            args.extend(TEST_RESOURCE_BOUNDS_FLAGS);
-        }
-
```
cptartur: Why isn't it required anymore?

crates/sncast/src/starknet_commands/multicall/mod.rs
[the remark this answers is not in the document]
cptartur: TIL, okay, let's keep that

crates/sncast/src/lib.rs
franciszekjob: 😆

crates/sncast/tests/docs_snippets/validation.rs
franciszekjob: This change should was accidentally added to this PR, it should stay on https://github.com/foundry-rs/starknet-foundry/pull/3101

crates/sncast/src/starknet_commands/deploy.rs
franciszekjob: Yup, left the commented out one by accident.

crates/sncast/tests/helpers/constants.rs
ddoktorski: * Are you sure this is the correct class hash? In the devnet repo it is named [sierra hash](https://github.com/0xSpaceShard/starknet-devnet/blob/cb779e696c07c89198a64c4d9c3ffde46bc7a66d/tests/integration/common/constants.rs#L26) not class hash
* Why is it `felt` instead of `&str`?
* You can find official calculated OZ class hashes in https://docs.openzeppelin.com/contracts-cairo/1.0.0/presets

franciszekjob: > Are you sure this is the correct class hash?

Yes, it's the one used for oz accounts predeployement:
image

> Are you sure this is the correct class hash?
It's just the name on their side.

> Why is it felt instead of &str?
It's not a change I included in my PR.

> You can find official calculated OZ class hashes in https://docs.openzeppelin.com/contracts-cairo/1.0.0/presets

You have a good eye 👍

crates/sncast/src/starknet_commands/account/create.rs
franciszekjob: > This is for a guardian public key. I guess they changed the interface and now it should be None, but please check this 🙏

Yes, in `v0.4.0` guardian key is optional but none that I've used `Some(Felt::ZERO)` because we're still using class hash for v0.3.1. However, I think we're good to update to v0.4.1 (which is used while creating account) so the guardian key will be `None` then. Updated.

crates/sncast/Cargo.toml
```diff
@@ -30,7 +30,7 @@ runtime = { path = "../runtime" }
 conversions = { path = "../conversions" }
 data-transformer = { path = "../data-transformer" }
 configuration = { path = "../configuration" }
-shared = { path = "../shared" }
+shared = { path = "../shared", features = ["testing"] }
```
ddoktorski, line 33: I think this way `testing` feature is always enabled. Is it the intended behavior?

crates/sncast/src/helpers/fee.rs
```diff
@@ -1,146 +1,193 @@
...
+                fee_estimate.overall_fee,
+                Felt::from(max_fee)
+            );
+
+            let fee_settings = FeeSettings::try_from_(fee_estimate.clone())
+                .expect("Failed to convert FeeEstimate to FeeSettings");
+            Ok(fee_settings)
+        } else {
+            let fee_settings = FeeSettings::try_from_(self.clone())
+                .expect("Failed to convert FeeArgs to FeeSettings");
```
ddoktorski, line 85: We should not allow a panic scenario depending on user input. Please improve the error handling for this.

crates/sncast/src/helpers/fee.rs
```diff
@@ -1,146 +1,193 @@
...
-    max_gas: impl Into<Felt>,
-    max_gas_unit_price: impl Into<Felt>,
-) {
-    let max_fee: Felt = max_fee.into();
-    let max_gas: Felt = max_gas.into();
-    let max_gas_unit_price: Felt = max_gas_unit_price.into();
-    println!(
-        "Specifying '--max-fee' flag results in conversion to '--max-gas' and '--max-gas-unit-price' flags\nConverted {max_fee} max fee to {max_gas} max gas and {max_gas_unit_price} max gas unit price\n",
-    );
+impl TryFromConv<FeeEstimate> for FeeSettings {
```
ddoktorski, line 139: Why `TryFromConv` instead of Rust `TryFrom`?

crates/sncast/src/helpers/fee.rs
```diff
@@ -1,146 +1,193 @@
...
+    type Error = String;
```
ddoktorski, line 140: Why string?

crates/sncast/Cargo.toml
franciszekjob: Absolutely, changed.

crates/sncast/src/helpers/fee.rs
franciszekjob: Change to anyhow error.

crates/sncast/src/helpers/fee.rs
franciszekjob: > In what cases can it even fail? Ideally it'd be verified on clap level or somewhere else and you'd just use ::from_ here.

Change to `from`.

> Wouldn't it make sense to use `u128` and `u64` accordingly in `FeeArgs` instead of typing these as felts? If we are converting them anyway.

Yes, I assume that before we had these values as felt because we wanted to accept hex strings. However, in spec they are defined as u64/u128, so I think we can safely change them.

crates/sncast/src/helpers/constants.rs
franciszekjob: Yup.

crates/sncast/tests/e2e/deploy.rs
```diff
@@ -302,7 +306,7 @@ fn test_contract_already_deployed() {
         output,
         indoc! {r"
         command: deploy
-        error: [..]Deployment failed: contract already deployed at address [..]
+        error: Transaction execution error [..]
```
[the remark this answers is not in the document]
franciszekjob: I'm not sure what's the reason for that, have a look - below I've printed stderr for `test_contract_already_deployed` and `test_same_salt_and_class_hash_deployed_twice` (from scripts). We can clearly see that in script test there's additional message test.

image

image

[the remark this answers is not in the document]
cptartur: In scripts we also do additional print of the result from within the script code `println!("{:?}", deploy_result);`. 

Maybe there's something with out error handling/serialization logic that this get's omitted when running from the CLI?

crates/conversions/src/serde/serialize/serialize_impl.rs
cptartur: Okay, ideally we'd have some tests here as well but okay.

crates/sncast/src/response/errors.rs
```diff
@@ -73,14 +75,14 @@ pub enum SNCastStarknetError {
...
+    InsufficientResourcesForValidate,
```
cptartur, line 79: So there is no error for insufficient max fee now? Is it some kind of resources bounds error now?

crates/sncast/tests/integration/fee.rs
cptartur, line 1: We should have a test where estimate fee isn't none as well.

1. for testing if it get's correctly used
2. for testing if `max_fee` not being none and `estimate` being provided
3. for testing if `max_fee` being provided and `estimate` being none

Maybe some other?

crates/sncast/src/response/errors.rs
franciszekjob: Yes, when resource bounds are too low to validate tx, see [spec](https://github.com/starkware-libs/starknet-specs/blob/d70a2e57c9a66db1bbf86c388f38d6295a6a2a75/api/starknet_write_api.json#L42).

crates/sncast/tests/integration/fee.rs
franciszekjob: 1. `test_from_fee_estimate`
2. `test_max_fee_set`
3. `test_max_fee_set_and_fee_estimate_none`

> Maybe some other?
`test_max_fee_set_and_fee_estimate_higher`

crates/sncast/tests/e2e/deploy.rs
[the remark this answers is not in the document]
franciszekjob: Discussed offline: will be addresses in fixes before stable.

crates/sncast/tests/integration/fee.rs
```diff
@@ -1,247 +1,264 @@
...
-            max_gas: None,
-            max_gas_unit_price: None,
+            l1_gas: Some(1),
+            l1_gas_price: Some(2),
+            l2_gas: Some(3),
+            l2_gas_price: Some(4),
+            l1_data_gas: Some(5),
+            l1_data_gas_price: Some(6),
         }
     );
```
cptartur, line 263: This is basically a unit test, please place it directly in fee.rs file.

Maybe some other from here can be moved as well?

crates/sncast/src/helpers/fee.rs
```diff
@@ -1,146 +1,183 @@
...
-        };
+    pub fn try_into_fee_settings(&self, fee_estimate: Option<&FeeEstimate>) -> Result<FeeSettings> {
+        // If some resource bounds values are lacking, starknet-rs will estimate them automatically
+        // but in case someone passes --max-fee flag, we need to make estimation on our own
+        // to check if the fee estimate isn't higher than provided max fee
+        if let Some(max_fee) = self.max_fee {
+            self.validate_max_fee_meets_resource_bounds()?;
+
+            let fee_estimate = fee_estimate
+                .ok_or_else(|| anyhow!("Fee estimate must be passed when max_fee is provided"))?;
```
cptartur, line 70: Actually, now that I look at it, it doesn't make sense to bubble it as a handleable error. If this error handling code is reached it means we have a bug in our logic somewhere else and this should never be allowed. Let's replace it with `expect()`.

crates/sncast/src/helpers/fee.rs
```diff
@@ -1,146 +1,183 @@
...
-                    max_gas_unit_price: Some(
-                        NonZeroU128::try_from_(max_gas_unit_price).map_err(anyhow::Error::msg)?,
-                    ),
-                }
-            }
-        };
+    pub fn try_into_fee_settings(&self, fee_estimate: Option<&FeeEstimate>) -> Result<FeeSettings> {
+        // If some resource bounds values are lacking, starknet-rs will estimate them automatically
+        // but in case someone passes --max-fee flag, we need to make estimation on our own
+        // to check if the fee estimate isn't higher than provided max fee
```
cptartur, line 65: Let's add an assert for case when `max_fee` is none and `fee estimate` is some. This means a bug in our logic and we should panic.

franciszekjob: [This line](https://github.com/foundry-rs/starknet-foundry/blob/franciszekjob/2490-rpc-0.8.0/crates/sncast/src/helpers/fee.rs#L69-L70) checks it.

franciszekjob: I don't think such assert is needed, as we already check if fee estimate is not none, inside first if's block.

crates/sncast/tests/integration/fee.rs
franciszekjob: Good catch, moved.

> Maybe some other from here can be moved as well?

No, nothing apart of that.

crates/sncast/tests/integration/fee.rs
```diff
@@ -1,247 +1,238 @@
...
+    let err = args.try_into_fee_settings(None).unwrap_err();
 
     assert_eq!(
-        settings,
-        FeeSettings {
-            max_gas: Some(NonZeroU64::new(MAX_FEE / 1_000).unwrap()),
-            max_gas_unit_price: Some(NonZeroU128::new(1_000).unwrap()),
-        }
+        err.to_string(),
+        "Fee estimate must be passed when max_fee is provided"
```
cptartur, line 209: I think it should just be 

```rust
args.try_into_fee_settings(None).unwrap();
```

Without any other lines? If the tests is `should_panic` anyway.

cptartur (approved):

docs/src/appendix/sncast/account/deploy.md
```diff
@@ -23,14 +23,34 @@ Possible values: `mainnet`, `sepolia`.
 ## `--max-fee, -m <MAX_FEE>`
 Optional.
 
-Maximum fee for the `deploy_account` transaction in Fri or Wei depending on fee token or transaction version. When not used, defaults to auto-estimation. Must be greater than zero.
+Maximum fee for the `deploy_account` transaction depending on fee token or transaction version. If used, estimated overall fee can't be higher than this value. Must be greater than zero.
```
ddoktorski, line 26:
```suggestion
Maximum fee for the `deploy_account` denoted in FRI. Must be greater than zero. If provided, it is not possible to use any of the following fee related flags: `--l1-gas`, `--l1-data-price`, `--l2-gas`, `--l2-gas-price`, `--l1-data-gas`, `--l1-data-gas-price`.
```
Also for other commands

crates/sncast/src/helpers/fee.rs
```diff
@@ -1,150 +1,222 @@
...
+            let fee_settings = FeeSettings::try_from(fee_estimate.clone())
+                .expect("Failed to convert FeeEstimate to FeeSettings");
+            Ok(fee_settings)
+        } else {
+            let fee_settings = FeeSettings::from(self.clone());
+            Ok(fee_settings)
+        }
+    }
+
+    fn validate_max_fee_meets_resource_bounds(&self) -> Result<()> {
```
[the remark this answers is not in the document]
franciszekjob: No, let's remove it.

ddoktorski (approved):

PR #3033, opened 2025-03-03
Integrate debugging trace into forge

ksew1:
Towards #3005

---

**Stack**:
- #3066
- #3065
- #3034
- #3033 ⬅
- #3032

⚠️ *Part of a stack created by [spr](https://github.com/ejoffe/spr). Do not merge manually using the UI - doing so may have unexpected results.*

PR #3065, opened 2025-03-09
Print contract by name

### LINKED ISSUE #3008: Print contract by name

Forge

### Feature Request

As in title

ksew1:
Closes #3008

---

**Stack**:
- #3066
- #3065 ⬅
- #3034
- #3033
- #3032

### DIFF
--- a/crates/debugging/src/trace.rs
+++ b/crates/debugging/src/trace.rs
@@ -1,6 +1,7 @@
 use crate::tree::TreeSerialize;
 use blockifier::execution::entry_point::CallType;
 use cheatnet::runtime_extensions::call_to_blockifier_runtime_extension::rpc::CallResult;
+use cheatnet::runtime_extensions::forge_runtime_extension::contracts_data::ContractsData;
 use cheatnet::state::{CallTrace, CallTraceNode};
 use starknet_api::contract_class::EntryPointType;
 use starknet_api::core::{ContractAddress, EntryPointSelector};
@@ -18,6 +19,7 @@ pub struct Trace {
 
 #[derive(Debug, Clone)]
 pub struct TraceInfo {
+    pub contract_name: Option<ContractName>,
     pub entry_point_type: EntryPointType,
     pub calldata: Calldata,
     pub storage_address: StorageAddress,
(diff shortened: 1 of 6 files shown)


crates/debugging/src/trace.rs
```diff
@@ -27,24 +29,39 @@ pub struct TraceInfo {
...
         let trace_info = TraceInfo {
+            contract_name: contracts_data
+                .get_contract_name(
+                    &call_trace
+                        .entry_point
+                        .class_hash
+                        .expect("this should be set in `fn execute_call_entry_point` in cheatnet"),
+                )
+                .cloned()
+                .map(ContractName),
```
franciszekjob, line 64: nit: Let's move `contract_name` declaration before `trace_info`, I'm not a fan of these deep nestings 😅

franciszekjob (approved):

crates/debugging/src/trace.rs
cptartur: Maybe extract this to a helper function? This clutters this code too much.

crates/debugging/src/tree/ui/as_tree_node.rs
```diff
@@ -29,3 +31,11 @@ impl AsTreeNode for TraceInfo {
...
     }
 }
+
+impl<T: NodeDisplay> AsTreeNode for Option<&T> {
+    fn as_tree_node(&self, parent: &mut Node) {
+        if let Some(value) = *self {
+            parent.leaf(value);
+        }
+    }
+}
```
cptartur, line 41: Do we need anywhere else than in line 22 you added? Adding this is risking weird behaviors in case more `Option`s are used for trace.

crates/debugging/src/trace.rs
franciszekjob: Can be too

crates/debugging/src/tree/ui/as_tree_node.rs
ksew1: This is exactly the behavior that I want for the option. I want to have some type of marker to indicate that if there is no value, it shouldn't be added to the tree.

cptartur (approved):

PR #3093, opened 2025-03-13
Voyager verification support

### LINKED ISSUE #2287: Add voyager verification interface to the `verify` command

Cast

### Feature Request

Currently, the `cast`'s `verify` command only support `walnut` verifier for contract verification. Adding more options to the user will add more flexibility for the contract verification

jkopanski:
Closes #2287

This will submit the contract for verification successfully, but the verification itself will fail as we need to release backend changes for supporting this. Those are in review right now.

## Introduced changes

- adds another `verifier` argument parameter: `voyager` to verify using [Voyager block explorer](https://voyager.online/)

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/sncast/src/starknet_commands/verify/voyager.rs
```diff
@@ -0,0 +1,355 @@
...
+    let ancestors = Utf8Path::ancestors(first_guess.as_ref());
+    let mut biggest_prefix: &Utf8Path = first_guess.as_ref();
+    for prefix in ancestors {
+        if paths.iter().all(|src| src.starts_with(prefix)) {
+            biggest_prefix = prefix;
+            break;
+        }
+    }
+    biggest_prefix.to_path_buf()
+}
```
franciszekjob, line 176: `first_guess` seems unnecessary — we could compute the common prefix directly from `paths` without needing an external starting point 😅 .

crates/sncast/Cargo.toml
```diff
@@ -7,6 +7,12 @@ edition.workspace = true
 [dependencies]
 anyhow.workspace = true
 camino.workspace = true
+cairo-lang-casm.workspace = true
+cairo-lang-runner.workspace = true
+cairo-lang-runnable-utils.workspace = true
+cairo-lang-sierra.workspace = true
+cairo-lang-sierra-to-casm.workspace = true
+cairo-lang-utils.workspace = true
```
kkawula, line 15: What's the reason for this change?

crates/sncast/src/starknet_commands/verify/mod.rs
```diff
@@ -25,7 +27,7 @@ pub struct Verify {
     pub contract_name: String,
 
     /// Block explorer to use for the verification
-    #[clap(short, long, value_enum, default_value_t = Verifier::Walnut)]
+    #[clap(short, long, value_enum)]
```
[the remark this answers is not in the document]
kkawula: Imo a changelog entry should be added and the default verifier should be removed. We shouldn't promote any of these.

kkawula (commented):
> This will submit the contract for verification successfully, but the verification itself will fail as we need to release backend changes for supporting this. Those are in review right now.

- Can you link this issue/pr?

- Please add entry there - docs/src/appendix/sncast/verify.md

crates/sncast/src/starknet_commands/verify/voyager.rs
```diff
@@ -0,0 +1,355 @@
...
+                }
+                None => Err(anyhow!(
+                    "More than one package found in scarb metadata - specify package using --package flag"
+                )),
+            }
+        } else if let Some(p) = workspace_packages.pop() {
+            Ok(p)
+        } else {
+            Err(anyhow!("No packages found in scarb metadata"))
+        })?;
```
kkawula, line 279:
```suggestion
let selected = if workspace_packages.len() > 1 {
            match package {
                Some(ref package_name) => workspace_packages
                    .into_iter()
                    .find(|p| p.name == *package_name)
                    .ok_or_else(|| {
                        anyhow!("Package '{}' not found in scarb metadata", package_name)
                    }),
                None => Err(anyhow!(
                    "Multiple packages found in scarb metadata. Specify a package using the --package flag."
                )),
            }
        } else {
            workspace_packages
                .pop()
                .ok_or_else(|| anyhow!("No packages found in scarb metadata"))
        }?;
```

crates/sncast/src/starknet_commands/verify/voyager.rs
```diff
@@ -0,0 +1,355 @@
...
+
+        let (prefix, files) = self.gather_files()?;
+        let project_dir_path = self
+            .metadata
+            .workspace
+            .root
+            .strip_prefix(prefix)
+            // backend expects this: "." for cwd
+            .map(|p| {
+                if p == Utf8Path::new("") {
```
kkawula, line 289:
```suggestion
                if p.as_str().is_empty() {
```

crates/sncast/src/starknet_commands/verify/voyager.rs
```diff
@@ -0,0 +1,355 @@
...
+                    Utf8Path::new(".")
+                } else {
+                    p
+                }
+            })?;
+
+        println!("These are the files that I'm about to transfer:");
```
kkawula, line 296:
```suggestion
        println!("The following files will be transferred:");
```

docs/src/starknet/verify.md
```diff
@@ -17,6 +17,10 @@ For detailed CLI description, see [verify command reference](../appendix/sncast/
 
 Walnut is a tool for step-by-step debugging of Starknet transactions. You can learn more about Walnut here [walnut.dev](https://walnut.dev). Note that Walnut requires you to specify the Starknet version in your `Scarb.toml` config file.
 
+### Voyager
+
+[Voyager](https://voyager.online/) is the Starknet block explorer.
+
```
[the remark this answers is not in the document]
jkopanski: I was mimicking walnut here, but since both aren't mentioned anywhere right now, perhaps we should name it something like `VERIFIER_API_URL` for both?

crates/sncast/Cargo.toml
jkopanski: That was leftover from previous attempt

docs/src/starknet/verify.md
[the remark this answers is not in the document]
franciszekjob: Correct me if I'm wrong, but this will enforce to update the env every time you want to use different verifier? I know it may be a rare case, but theoretically that's how it would look.

crates/sncast/src/starknet_commands/verify/voyager.rs
jkopanski: yeah, it was there because `first_guess` will be right almost every time, but in the end it doesn't save anything.

crates/sncast/src/starknet_commands/verify/mod.rs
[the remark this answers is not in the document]
jkopanski: I've added a changelog entry

crates/sncast/src/starknet_commands/verify/voyager.rs
jkopanski: Thanks!

crates/sncast/src/starknet_commands/verify/voyager.rs
jkopanski: Actually failing test made me think about it again. While we could calculate it from `paths` that would require a _ton_ of iterations. Basically building the prefix from the smallest one and checking every path if it's still applicable.

By having good starting point (which we have) we can check if that works for us and be done if it is. Only if that fails, we look for the other options.

docs/src/starknet/verify.md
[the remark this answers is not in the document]
franciszekjob: >I'm not sure what you mean here.

I mean that if there will be `VERIFIER_API_URL` (shared between both verifiers), you'll have to change it every time you change verifier.

crates/sncast/src/starknet_commands/verify/voyager.rs
franciszekjob: Ok, let's leave it as it is.

docs/src/starknet/verify.md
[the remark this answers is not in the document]
jkopanski: but change verifier in what sense? Users change verifier through command line argument, the env variable is left _unset_ almost all the time.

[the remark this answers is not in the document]
franciszekjob: >but change verifier in what sense? 

Let's say you use walnut and you've set `VERIFIER_API_URL` to walnut one. Then you switch to voyager, hence you need to change the env again. I know it may be a rare case. Wdyt?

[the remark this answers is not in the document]
jkopanski: >  I know it may be a rare case. Wdyt?

Yes it's rare, because (as a `sncast` user) in order to use walnut *you don't set* any env variable. You pass `--verifier walnut`.

The env variable acts as an override so a developer (either walnut, voyager or foundry dev) could target non production instance. IMHO then having single env var makes more sense, but it's not a hill I'm willing to die on ;)

franciszekjob:
@jkopanski Is it ready to re-review?

crates/sncast/src/starknet_commands/verify/voyager.rs
```diff
@@ -0,0 +1,355 @@
...
+    fn new(network: Network, workspace_dir: Utf8PathBuf) -> Result<Self> {
+        let manifest_path = scarb_utils::get_scarb_manifest_for(workspace_dir.as_ref())?;
+        let metadata = scarb_utils::get_scarb_metadata_with_deps(&manifest_path)?;
+        let url_str = match env::var("STARKNET_RPC_URL") {
+            Ok(addr) => addr,
+            Err(_) => match network {
+                Network::Mainnet => "https://free-rpc.nethermind.io/mainnet-juno/v0_7".to_string(),
+                Network::Sepolia => "https://free-rpc.nethermind.io/sepolia-juno/v0_7".to_string(),
+            },
+        };
```
franciszekjob, line 224: 1. We don't have any mechanism which allows to set rpc url through env.
2. Have a look at `Network` impl, it provides methods which return provider url 😄 .
3. This won't apply if you follow step 2., but we should use rpc 0.8

CHANGELOG.md
franciszekjob, line 1: Shouldn't we mention that it's possible to use `VERIFIER_API_URL`?

crates/sncast/src/starknet_commands/verify/voyager.rs
jkopanski: > Have a look at Network impl, it provides methods which return provider url 😄 .

Thanks, that seemed like something I was missing. You're referring to [this one](https://github.com/foundry-rs/starknet-foundry/blob/034c433c1ee2d5e6897179b0cb3378a32e0b3f9f/crates/sncast/src/helpers/rpc.rs#L84)? All those functions are private.

Looking at that module perhaps I should add `RpcArgs` to verify?

CHANGELOG.md
jkopanski: Personally I don't think so. It wasn't mentioned before and I think it's not a sure facing feature. I suspect (you can confirm or deny) but it was added only for setting that in the tests.

crates/sncast/src/starknet_commands/verify/voyager.rs
franciszekjob: > Looking at that module perhaps I should add RpcArgs to verify?

Yes, that should solve the issue.

jkopanski: Unfortunately I couldn't reuse much of existing machinery, due to how `--network` interacts with `CastConfig`. Please see how it looks now. I think the test failure is unrelated and needs a restart.

franciszekjob:
Hi @jkopanski , is this ready to re-review?

jkopanski:
@franciszekjob yeah it is, given my last commentary on the rpc args solution

franciszekjob (approved):
Tested it, works good! I just have a small remark - maybe we could sort the files for slightly better UX?
image

jkopanski:
@franciszekjob I'm getting some unrelated failures.

franciszekjob:
> @franciszekjob I'm getting some unrelated failures.

It's a windows test, I've rerun it and it should pass.

franciszekjob (changes requested):
1. I've noticed that `--url` can't be used with `verify command`, please fix this 🙏 .

image

2. When do you expect the backend to be ready?

jkopanski:
> I've noticed that --url can't be used with verify command, please fix this 🙏 .

Yes, it's not and was not defined for verification. What do you mean by fix it? I think it's bit ambiguous in this context. Does it mean RPC or verification service url? Furthermore look at my comments above regarding re using existing `RpcArgs` if that is what you had in mind by this comment.

franciszekjob:
> > I've noticed that --url can't be used with verify command, please fix this 🙏 .
> 
> Yes, it's not and was not defined for verification. What do you mean by fix it? I think it's bit ambiguous in this context. Does it mean RPC or verification service url? Furthermore look at my comments above regarding re using existing `RpcArgs` if that is what you had in mind by this comment.

Nevermind - I see it's consistent with what we have, ignore it.

franciszekjob:
> When do you expect the backend to be ready so we can merge the PR?

How about this @jkopanski ?

jkopanski:
> > When do you expect the backend to be ready?
> 
> How about this @jkopanski ?

it's "done", needs to be released. Let me get back a bit later with more detailed info

jkopanski:
@franciszekjob we're planning to have a release on Thursday

jkopanski:
@franciszekjob it was released yesterday

crates/sncast/src/starknet_commands/verify/walnut.rs
```diff
@@ -86,13 +92,13 @@ impl VerificationInterface for WalnutVerificationInterface {
         }
     }
 
-    fn gen_explorer_url(&self) -> Result<String> {
+    fn gen_explorer_url(&self) -> String {
         let api_base_url =
-            env::var("WALNUT_API_URL").unwrap_or_else(|_| "https://api.walnut.dev".to_string());
+            env::var("VERIFIER_API_URL").unwrap_or_else(|_| "https://api.walnut.dev".to_string());
```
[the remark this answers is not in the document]
jkopanski: We've discussed this above.

This variable was *not* mentioned in the docs originally, and I think for good reason. It's not something that we want users to set. User should use cli parameters to set the target. This sort of _development mode_ and to run tests.

Are you sure, you want to advertise this in docs/changelog?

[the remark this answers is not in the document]
kkawula: Okey, in the sense of that it is good as it is

crates/sncast/src/starknet_commands/verify/mod.rs
```diff
@@ -49,30 +57,61 @@ pub struct Verify {
     /// Specifies scarb package to be used
     #[arg(long)]
     pub package: Option<String>,
+
+    /// RPC provider url address; overrides url from snfoundry.toml. Will use public provider if not set.
```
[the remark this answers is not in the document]
jkopanski: I've added the docs but it seems the ci is stuck?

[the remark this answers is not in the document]
franciszekjob: CI didn't start because you have some unresolved conflicts with master branch.

kkawula:
@jkopanski Could you please resolve conflicts?

PR #3094, opened 2025-03-13
Replace equilibrium rpc url with our internal node

### LINKED ISSUE #3086: Use internal node
### Current State

Once our internal node is updated, we should use it instead of public one.

### Objective

We should use our internal node

### Additional Context

_No response_

ddoktorski:
Closes #3086


franciszekjob (approved):

franciszekjob:
`e2e::forking::with_cache` has failed - it's because there's no cache file associated with the node url. We need to change version in `http___188_34_188_184_7070_rpc_v0_7_54060_v0_38_3.json` file name and also remove `https___rpc_pathfinder_equilibrium_co_testnet_sepolia_rpc_v0_8_54060_v0_38_3.json`.

PR #3110, opened 2025-03-17
Add `erc20-contract` template

### LINKED ISSUE #2850: Different templates when using `snforge new`

ddoktorski:
Closes #2850

## Introduced changes

- Based on https://github.com/amanusk/starknet-foundry-template

## Checklist

- [X] Linked relevant issue
- [X] Added relevant tests


crates/forge/src/new.rs
```diff
@@ -349,10 +400,24 @@ fn get_template_dir(template: &Template) -> Result<Dir> {
...
+
+    let oz_version = match scarb_version {
+        ver if ver >= Version::new(2, 9, 4) => Version::new(1, 0, 0),
+        ver if ver >= Version::new(2, 9, 1) => Version::new(0, 20, 0),
+        ver if ver >= Version::new(2, 8, 4) => Version::new(0, 19, 0),
+        _ => bail!("Minimal Scarb version to create a new project with ERC-20 template is 2.8.4"),
+    };
+
+    Ok(oz_version)
+}
```
cptartur, line 423: How will we ensure this stays up to date?

crates/forge/tests/e2e/new.rs
```diff
@@ -172,15 +178,31 @@ fn get_expected_manifest_content(template: &Template, validate_snforge_std: bool
...
+            name = "SEPOLIA_LATEST"
+            url = "{FREE_RPC_PROVIDER_URL}"
+            block_id = {{ tag = "latest" }}
+        "#
+        )
+    } else {
+        ""
+    };
+
     let (dependencies, target_contract_entry) = match template {
```
cptartur, line 196: Doesn't this shadow the `target_contract_entry` defined above? Or is this intended?

cptartur (commented):
Didn't review the contract code. I'm assuming it was taken from somewhere that makes sense.

crates/forge/tests/e2e/new.rs
ddoktorski: Yes intended, it is not present in the `cairo-program` template

crates/forge/src/new.rs
ddoktorski: It's not a problem to use newer Scarb with older OZ dependency, so it would be good to update it from time to time, but it is not strictly necessary for it to work.

crates/forge/tests/e2e/new.rs
```diff
@@ -191,18 +213,30 @@ fn get_expected_manifest_content(template: &Template, validate_snforge_std: bool
...
-        snforge_std_assert
-    ).trim_end().to_string() + "\n"
+        snforge_std_assert,
+        SCARB_MANIFEST_TEMPLATE_CONTENT.trim_end()
+    );
+
+    // Replace 3 or more consecutive newlines with exactly 2 newlines
+    RE_NEWLINES
+        .replace_all(&expected_manifest, "\n\n")
+        .to_string()
```
franciszekjob, line 239: Could you please explain that?

snforge_templates/erc20_contract/tests/test_erc20.cairo
```diff
@@ -0,0 +1,129 @@
+use openzeppelin_token::erc20::{ERC20ABIDispatcher, ERC20ABIDispatcherTrait, ERC20Component};
+use snforge_std::{
+    CheatSpan, ContractClassTrait, DeclareResultTrait, EventSpyAssertionsTrait,
+    cheat_caller_address, declare, spy_events,
+};
+use starknet::{ContractAddress, contract_address_const};
+
+const ETH_TOKEN_ADDRESS: felt252 =
+    0x049d36570d4e46f48e99674bd3fcc84644ddd6b96f7c741b1562b82f9e004dc7;
```
franciszekjob, line 9: Wouldn't it be better to use STRK?

franciszekjob (commented):
We should also mention somewhere in the forge guide that it's possible to use this template.

snforge_templates/erc20_contract/tests/test_erc20.cairo
```diff
@@ -0,0 +1,129 @@
...
+
+const INITIAL_SUPPLY: u256 = 10_000_000_000;
+
+fn setup() -> ContractAddress {
+    let erc20_class_hash = declare("MockERC20").unwrap().contract_class();
+
+    let mut calldata = ArrayTrait::new();
+    INITIAL_SUPPLY.serialize(ref calldata);
+
+    let sender_account = contract_address_const::<1>();
```
kkawula, line 19: `contract_address_const` is deprecated :C
> note: "Use `TryInto::try_into` in const context instead."
https://github.com/starkware-libs/cairo/blob/main/corelib/src/starknet.cairo#L51-L55

ddoktorski: I mentioned this during the call. It is possible to use `try_into` in this case only from Cairo `2.11`. To support earlier versions we need to use `contract_address_const` (or possibly something else that I am not aware of).

crates/forge/tests/e2e/new.rs
ddoktorski: We use `assert_matches` so any discrepancy such as additional or missing newline will result in a failed assertion. This is a simple way to normalize the expected string.

snforge_templates/erc20_contract/tests/test_erc20.cairo
ddoktorski: It doesn't make a difference to me. I can change it to STRK if you prefer.

crates/forge/src/new.rs
cptartur: Please create an issue to keep this up to date (some automatic flow maybe) and put it in the backlog. So we at least eventually revisit this.

cptartur (approved):

crates/forge/src/new.rs
ddoktorski: There is already a stable OZ version and as I said nothing will break here if it is not updated immediately after a new version release. I wouldn't add any issues that are not actionable and any automatic flow for this seems to be overkill.

franciszekjob (approved):

crates/forge/src/new.rs
cptartur: I guess the test for the template should catch any regressions in this.

cptartur (approved):

PR #3113, opened 2025-03-18
Fix fork caches on windows

### LINKED ISSUE #2303: Fix double locking file caches on windows

snforge

### Description

There seems to be a problem with file locking, double lock does not wait for acquiring but fails on the write itself:

Image

Related code:

https://github.com/foundry-rs/starknet-foundry/blob/8fa75bbdd1b0b0b80c6fc3e7fc9fa399a11a925a/crates/cheatnet/src/forking/cache.rs#L138C1-L141C1

The problem seems to be the way windows locks work in `fs2`
Related info about cross-platform locking in `fs2`: https://tikv.github.io/doc/fs2/trait.FileExt.html#notes-on-file-locks

cptartur:
Closes #2303

## Introduced changes

- Fixed fork caches on windows.
- Restored forking related tests on windows

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


kkawula (approved):

crates/cheatnet/src/forking/cache.rs
```diff
@@ -134,15 +134,17 @@ impl ForkCache {
     fn save(&self) {
         let mut file = OpenOptions::new()
```
Draggu, line 135: Probably should be stored in `ForkCache` instead of `Utf8PathBuf` to prevent using wrong handle in future

crates/cheatnet/src/forking/cache.rs
```diff
@@ -152,6 +154,8 @@ impl ForkCache {
             fs_fork_cache_content.to_string()
         };
 
+        file.set_len(0).expect("Failed to truncate file");
```
Draggu, line 157: You can make separate method `.clear()` for this

crates/cheatnet/src/forking/cache.rs
cptartur: This way we'd be keeping the file open for the duration of the program. I'd avoid changing this logic too much in this PR.

ddoktorski (approved):

crates/cheatnet/src/forking/cache.rs
```diff
@@ -99,6 +99,18 @@ impl Drop for ForkCache {
     }
 }
 
+trait FileTruncateExtension {
```
Draggu, line 102: Use `Ext`, `Extension` has meaning in cheatnet

PR #3134, opened 2025-03-24
Handle specially top level call in trace

### LINKED ISSUE #3079: Handle specially top level call in trace

Forge

### Feature Request

The top level call is not contract but test itself, because of that it needs special handling. Maybe custom display format?

ksew1:
Closes #3079

## Introduced changes

- display top level call as test name

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/debugging/src/trace/types.rs
```diff
@@ -0,0 +1,69 @@
...
+pub struct ContractName(pub String);
+
+#[derive(Debug, Clone)]
+pub struct StorageAddress(pub ContractAddress);
+
+#[derive(Debug, Clone)]
+pub struct CallerAddress(pub ContractAddress);
+
+impl Trace {
+    /// Creates a new [`Trace`] from a given `cheatnet` [`CallTrace`] and [`ContractsData`].and a test name.
```
kkawula, line 55:
```suggestion
    /// Creates a new [`Trace`] from a given `cheatnet` [`CallTrace`] and [`ContractsData`], and a test name.
```

crates/debugging/src/trace/collect.rs
```diff
@@ -0,0 +1,75 @@
...
+    contracts_data: &ContractsData,
+    test_name: String,
+) -> Trace {
+    let call_trace = call_trace.borrow();
+
+    Trace {
+        test_name: TestName(test_name),
+        nested_calls: nested_calls(&call_trace, contracts_data),
+    }
+}
```
kkawula, line 21: Is this abstraction needed here? Apart from that, maybe it will be better to provide just `&CallTrace` because it always calls `.borrow()` at the beginning

crates/debugging/src/trace/collect.rs
```diff
@@ -0,0 +1,75 @@
...
+    contracts_data
+        .get_contract_name(
+            &call_trace
+                .entry_point
+                .class_hash
+                .expect("this should be set in `fn execute_call_entry_point` in cheatnet"),
+        )
+        .cloned()
+        .map(ContractName)
+        .expect("this should be present in `ContractsData`")
```
kkawula, line 66: In the case of panic, pannic message wouldn't provide info on what actually is missing. It also applies applies to other `.expect()` occurrences 
```suggestion
        .expect("contract name should be present in `ContractsData`")
```

crates/debugging/src/trace/types.rs
```diff
@@ -0,0 +1,69 @@
...
+pub struct ContractName(pub String);
+
+#[derive(Debug, Clone)]
+pub struct StorageAddress(pub ContractAddress);
+
+#[derive(Debug, Clone)]
+pub struct CallerAddress(pub ContractAddress);
+
+impl Trace {
+    /// Creates a new [`Trace`] from a given `cheatnet` [`CallTrace`] and [`ContractsData`] and a test name.
```
kkawula, line 55:
```suggestion
    /// Creates a new [`Trace`] from a given `cheatnet` [`CallTrace`], [`ContractsData`] and a test name.
```

crates/debugging/src/trace/collect.rs
ksew1: > Is this abstraction needed here?  

I don't think it's abstraction; it's just a function. 😅 I also think it should be there, as all functions that collect should be in the collection module.  

> Maybe it would be better to provide just `&CallTrace`.  

Yep, I agree, as it creates unnecessary indirection. Changed. 🫡

kkawula (approved):

ddoktorski (approved):

PR #3140, opened 2025-03-25
Validate forge packages

### LINKED ISSUE #2975: Run `scarb lint` on forge packages
### Current State

Currently, we do not lint our packages at all

### Objective

Run `scarb lint` on forge packages

### Additional Context

Setup a CI workflow that will lint packages, including `forge new` templates. Additionally, fix the already broken lints.

### LINKED ISSUE #2976: Check template for build warnings
### Current State

Currently, I don't think we check for build warnings.
If we do, please close the issue.

### Objective

Validate if snforge template does not produce build warnings

### Additional Context

Probably can be resolved by running `scarb check` on it and validating the output doesn't contain errors.

It seems it can also be resolved by adding:
```
[cairo]
allow-warnings = false
```
to `Scarb.toml`.

franciszekjob:
Closes #2975
Closes #2976

## Introduced changes

- Validate `snforge_std` and `sncast_std` templates
- Validate (`scarb check` and `scarb lint`) on forge templates

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/forge/tests/integration/gas.rs
franciszekjob, line 1: note: Gas has changed because we're using `!=` instead of `<` comparison.

.github/workflows/ci.yml
```diff
@@ -267,22 +267,29 @@ jobs:
...
-          exit 0
+          chmod +x ./scripts/setup_snforge_templates.sh
+          ./scripts/setup_snforge_templates.sh
+      - name: Check packages formatting
+        run: ./scripts/scarb_fmt.sh
+      - uses: software-mansion/setup-scarb@v1
+        with:
+          scarb-version: "2.10.0"
+      - name: Check packages linting
+        run: ./scripts/scarb_lint.sh
```
franciszekjob, line 292: note: `scarb lint` is available from 2.10.0, hence need to use different scarb version

crates/forge/tests/integration/gas.rs
```diff
@@ -1201,15 +1201,15 @@ fn contract_keccak_cost_sierra_gas() {
...
+    // 1157519 = reported consumed sierra gas
+    // 0 l1_gas + 96 l1_data_gas + (142810 + 50000 + 87650 + 1157519) l2 gas
     assert_gas(
         &result,
         "contract_keccak_cost",
         GasVector {
             l1_gas: GasAmount(0),
             l1_data_gas: GasAmount(96),
-            l2_gas: GasAmount(1_439_395),
+            l2_gas: GasAmount(1_438_075),
```
franciszekjob, line 1212: note: Gas changes are because of modifications in gas checker contract - now we're using `!=` instead of `<` in loops.

scripts/scarb_lint.sh
ddoktorski, line 1: Tbh do we need checking lint in test packages at all? What are the benefits of this?

franciszekjob: It's a tiny improvement of devex. It's not a big effort so imo we should leave it.

ddoktorski: I must disagree. It seems like quite a lot of effort to maintain this compared to the value it brings, and it also makes our already fat CI even bigger.

.github/workflows/ci.yml
```diff
@@ -267,22 +267,29 @@ jobs:
...
-          output=$(find . -type f -name "Scarb.toml" -execdir sh -c '
-              echo "Running \"scarb fmt\" in directory: $PWD"
-              scarb fmt --check
-          ' \;)
-          echo "$output"
-          if grep -iq "Diff" <<< "$output"; then
-              exit 1
-          fi
-          exit 0
+          chmod +x ./scripts/setup_snforge_templates.sh
```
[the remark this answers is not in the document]
franciszekjob: good catch 👍

scripts/setup_snforge_templates.sh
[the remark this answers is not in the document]
franciszekjob: No, assuming that packages are correctly lined and can be compiled.

scripts/scarb_lint.sh
```diff
@@ -0,0 +1,43 @@
...
+' \;)
+
+echo "$output"
+
+# Check for lint warnings, we need to do it this way because `scarb lint` returns 0 even if there are warnings
+if grep -iq "warning: Plugin diagnostic:" <<< "$output"; then
+    exit 1
+# Check for missing `edition`` field
+elif grep -iq "warn: `edition` field not set in `[package]` section for package" <<< "$output"; then
+    exit 1
```
kkawula, line 40: Maybe it is worth creating an issue for that. If in the future, `scarb lint` can change warning into an error

kkawula (changes requested):

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 #### Added
 
 - `--template` flag to `snforge new` command that allows selecting a template for the new project. Possible values are `balance-contract` (default), `cairo-program` and `erc20-contract`
+- `Scarb.toml` created by the `snforge new` command now includes the entry `allow-warnings = false` for Cairo checks
```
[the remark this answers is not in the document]
franciszekjob: Fine, it may be too impactful. removing.

scripts/scarb_lint.sh
franciszekjob: #3148

franciszekjob:
@ddoktorski @kkawula
I've removed lint-related changes with internal packages and moved the tests from CI to e2e (as discussed with @ddoktorski ). Take a look

.github/workflows/ci.yml
```diff
@@ -267,23 +270,6 @@ jobs:
       - uses: software-mansion/setup-universal-sierra-compiler@v1
       - run: cargo test --release -p scarb-api
 
-  scarbfmt:
```
cptartur, line 270: Any reason we no longer do this?

.github/workflows/ci.yml
```diff
@@ -320,6 +306,8 @@ jobs:
       - uses: Swatinem/rust-cache@82a92a6e8fbeee089604da2575dc567ae9ddeaab
       - uses: actions/checkout@v4
       - uses: software-mansion/setup-scarb@v1
+        with:
+          scarb-version: "2.10.0"
```
cptartur, line 310: Why are we using a fixed scarb version here? 
Also I guess we should eventually remove this, an issue for this would be good.

crates/forge/tests/e2e/packages_validation.rs
```diff
@@ -0,0 +1,149 @@
...
+        let path = path.unwrap().path();
+        if path.is_dir() {
+            matches.append(&mut find_files_recursively(
+                Utf8PathBuf::from(path.to_string_lossy().to_string()),
+                file_name,
+            ));
+        } else if path.to_string_lossy().contains(file_name) {
+            matches.push(Utf8PathBuf::from(path.to_string_lossy().to_string()));
+        }
+    }
```
cptartur, line 26: I think we could use walkdir for this, no? https://docs.rs/walkdir/latest/walkdir/

crates/forge/tests/e2e/packages_validation.rs
```diff
@@ -0,0 +1,149 @@
+use std::{fs, process::Stdio};
+
+use super::common::runner::runner;
+use assert_fs::prelude::PathChild;
+use camino::Utf8PathBuf;
+use clap::ValueEnum;
+use forge::Template;
+use project_root::get_project_root;
```
cptartur, line 8: This package is just a single function with a couple of lines, maybe we could just implement it ourselves?

crates/forge/tests/e2e/packages_validation.rs
```diff
@@ -0,0 +1,149 @@
...
+            .success();
+
+        let package_path = temp_dir.child(package_name);
+        let package_path = Utf8PathBuf::from_path_buf(package_path.to_path_buf())
+            .expect("Failed to convert to Utf8PathBuf");
+
+        // Overwrite Scarb.toml with `allow-warnings = true`
+        let scarb_toml_path = package_path.join("Scarb.toml");
+        let mut scarb_toml = fs::read_to_string(&scarb_toml_path).unwrap();
+        scarb_toml.push_str("\n[cairo]\nallow-warnings = true\n");
```
cptartur, line 129: Why do we allow them?

crates/forge/tests/e2e/packages_validation.rs
```diff
@@ -0,0 +1,149 @@
...
+        fs::write(&scarb_toml_path, &scarb_toml).expect("Failed to write to Scarb.toml");
+
+        check_and_lint(package_path);
+    }
+}
+
+#[test_case("snforge_std")]
+#[test_case("sncast_std")]
```
cptartur, line 137: This validates sncast_std even though it is in `forge` crate. I think it'd make more sense in a separate crate that checks both of these crates (and we could expand it to other crates easily in the future if needed).

cptartur (changes requested):

.github/workflows/ci.yml
franciszekjob: We do it in e2e test

.github/workflows/ci.yml
franciszekjob: Not needed, it was needed because of some linting changes (which aren't present anymore). Removed.

crates/forge/tests/e2e/packages_validation.rs
franciszekjob: We can but don't wan to mix changes in this PR, can be addressed in separate one.

crates/forge/tests/e2e/packages_validation.rs
franciszekjob: We do this only to ensure that scarb check will fail if there are any warnings (that's even suggested in issue description + we've talked about it offline yesterday 😅 ) . Note that we do it only for test packages, we don't include it in `Scarb.toml` template.

crates/forge/tests/e2e/packages_validation.rs
franciszekjob: Good catch, will move it.

crates/forge/tests/e2e/packages_validation.rs
cptartur: But shouldn't it be `allow_warnings=false` then?

franciszekjob: Oops, should be, it was a typo.

.github/workflows/ci.yml
```diff
@@ -140,6 +140,8 @@ jobs:
       - uses: software-mansion/setup-universal-sierra-compiler@v1
 
       - run: cargo test --package forge --features scarb_since_2_10 sierra_gas
+      - run: cargo test --package forge --features scarb_since_2_10 validate_templates
+      - run: cargo test --package packages_validation --features scarb_since_2_10 validate_cairo_libs
```
ddoktorski, line 144: It would be better to extend `scarbfmt` job and add these steps there. It could be renamed to something like `scarb fmt and lint`

crates/forge/tests/e2e/templates_validation.rs
```diff
@@ -0,0 +1,41 @@
...
+use forge::Template;
+use packages_validation::check_and_lint;
+
+#[test]
+fn validate_templates() {
+    let temp_dir = assert_fs::TempDir::new().expect("Unable to create a temporary directory");
+
+    let templates = Template::value_variants();
+
+    for template in templates {
```
ddoktorski, line 16: nit: Why not to use `#[test_case]` instead for each template?

franciszekjob: To prevent from forgetting about adding test case when new template is added.

.github/workflows/ci.yml
franciszekjob: What's the rationale? Packages validation tests are run twice - normally in e2e suite and here with scarb 2.10.
I don't think we should spread tests run with `scarb_since_2_10` feature flag across many jobs.
Once our minimal supported scarb version is 2.10, then we will just remove this job and there won't be need to modify others.

franciszekjob: > That's true in the case of validate_templates, but what about the packages_validation tests?

Right, it should be added to run without feature flag, now I see your point.

scripts/scarbfmt.sh
```diff
@@ -1,12 +1,10 @@
...
-
-popd || exit
+output=$(find . -type f -name "Scarb.toml" -execdir sh -c '
+    echo "Running \"scarb fmt --check\" in directory: $PWD"
+    scarb fmt --check
+' \;)
+echo "$output"
+if grep -iq "Diff" <<< "$output"; then
+    exit 1
+fi
```
kkawula, line 10: nit. bash generally uses 2 spaces for indentation
you can easily format code with `shfmt -i 2 <file>`
```suggestion
output=$(find . -type f -name "Scarb.toml" -execdir sh -c '
    echo "Running \"scarb fmt --check\" in directory: $PWD"
    scarb fmt --check
' \;)
echo "$output"
if grep -iq "Diff" <<<"$output"; then
  exit 1
fi
```

crates/packages_validation/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "packages_validation"
+version = "1.0.0"
+edition.workspace = true
```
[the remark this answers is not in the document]
ddoktorski: Sounds good to me

cptartur:
@franciszekjob please re-request when it's ready for another review (xtask is implemented?)

crates/packages_validation/Cargo.toml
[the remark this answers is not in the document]
franciszekjob: as dicsussed offline: xtask seems good, but in order to validate templates we're using enum from forge, also the `check_and_lint` function is used both in libs and templates validation. Let's leave it in crate.

crates/packages_validation/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "packages_validation"
```
kkawula, line 2: Maybe it should be renamed? 
sth like `testing_packages_validatoin`, in order to distinguish this package from others

crates/packages_validation/src/lib.rs
```diff
@@ -0,0 +1,37 @@
+use camino::Utf8PathBuf;
+use scarb_api::ScarbCommand;
+use std::process::Stdio;
+
+pub fn check_and_lint(package_path: Utf8PathBuf) {
+    println!("Running `scarb check` in directory {package_path}");
```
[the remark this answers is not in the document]
franciszekjob: Without this we won't know for which package `scarb check` has failed - it's not cluttering.
We do the same for docs validation btw.

scripts/scarbfmt.sh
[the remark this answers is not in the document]
franciszekjob: In fact, it's in separate script 😅 . Am I missing something?

crates/packages_validation/Cargo.toml
franciszekjob: I think current name already distinguishes it from others, but I'm open for change. Wdyt @ddoktorski @cptartur ?

crates/packages_validation/src/lib.rs
```diff
@@ -0,0 +1,37 @@
...
+        .arg("check")
+        .command()
+        .stdout(Stdio::inherit())
+        .stderr(Stdio::inherit())
+        .output()
+        .expect("Failed to run `scarb check`");
+    assert!(check_output.status.success(), "`scarb check` failed");
+
+    // TODO(#3149)
+    if cfg!(feature = "scarb_since_2_10") {
```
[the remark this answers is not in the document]
franciszekjob: Could you clarify a bit? We run `scarb lint` only for scarb 2.10.0 (because it provides lint).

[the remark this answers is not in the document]
franciszekjob: discussed offline, it's fine as it is now

crates/packages_validation/Cargo.toml
cptartur: Does rust workspace structure allow for nesting? If so, maybe it'd make sense to have a `crates/testing` directory? Then we wouldn't need to include `testing_` in the crate name.

We could possible move code from shared::test_utils there as well.

cptartur:
@franciszekjob please re-request review after outstanding comments have been answered (or addressed) 🙏

scripts/scarbfmt.sh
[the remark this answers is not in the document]
franciszekjob: Restored.

crates/packages_validation/Cargo.toml
franciszekjob: I've added new crates `testing` and `packages_validation` inside. 

> We could possible move code from shared::test_utils there as well.

Will see it that makes sense and potentially move it in separate PR 👍

crates/packages_validation/tests/cairo_libs.rs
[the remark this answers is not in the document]
franciszekjob: Moved

.github/workflows/ci.yml
```diff
@@ -276,6 +284,12 @@ jobs:
       - uses: software-mansion/setup-universal-sierra-compiler@v1
       - name: Run tests
         run: cargo test --release -p sncast
+      # TODO(#3096): Remove running tests with `scarb_since_2_10`
+      - uses: software-mansion/setup-scarb@v1
+        with:
+          scarb-version: "2.10.1"
```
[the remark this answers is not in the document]
franciszekjob: But we need to set specific version of scarb here, in this case 2.10.1, that's why we use it again.

.github/workflows/ci.yml
```diff
@@ -44,6 +44,14 @@ jobs:
       - uses: software-mansion/setup-scarb@v1
       - uses: software-mansion/setup-universal-sierra-compiler@v1
       - run: cargo test --release --lib -p forge
+      # TODO(#3096): Remove running tests with `scarb_since_2_10`
+      - uses: software-mansion/setup-scarb@v1
+        with:
+          scarb-version: "2.10.1"
```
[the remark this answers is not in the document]
franciszekjob: Reply same as above.

.github/workflows/ci.yml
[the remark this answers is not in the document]
kkawula: We probably can removed one of these

crates/forge/tests/e2e/templates.rs
```diff
@@ -0,0 +1,38 @@
...
+
+    let package_path = temp_dir.child(package_name);
+    let package_path = Utf8PathBuf::from_path_buf(package_path.to_path_buf())
+        .expect("Failed to convert to Utf8PathBuf");
+
+    // Overwrite Scarb.toml with `allow-warnings = false`
+    let scarb_toml_path = package_path.join("Scarb.toml");
+    let mut scarb_toml = fs::read_to_string(&scarb_toml_path).unwrap();
+    scarb_toml.push_str("\n[cairo]\nallow-warnings = false\n");
+    fs::write(&scarb_toml_path, &scarb_toml).expect("Failed to write to Scarb.toml");
```
kkawula, line 35: Maybe to avoid bugs in the future, we should use `toml_edit` crate to add this entry without creating conflicts?

Cargo.toml
```diff
@@ -14,7 +14,9 @@ members = [
     "crates/configuration",
     "crates/universal-sierra-compiler-api",
     "crates/docs",
-    "crates/debugging"
+    "crates/debugging",
+    "crates/testing",
+    "crates/testing/packages_validation",
```
cptartur, line 19: Can't we do something along the lines of this?

```suggestion
    "crates/testing/*"
```

crates/testing/Cargo.toml
cptartur, line 1: I don't think we need an empty `testing` package (unless there is now way around it). It should just be a plain directory containing testing related crates.

crates/testing/packages_validation/src/lib.rs
```diff
@@ -0,0 +1,54 @@
...
+        .expect("Failed to get project root")
+        .join(package_name);
+    let package_path = package_path
+        .canonicalize()
+        .expect("Failed to canonicalize path");
+    let package_path =
+        Utf8PathBuf::from_path_buf(package_path).expect("Failed to convert to Utf8PathBuf");
+
+    check_and_lint(&package_path);
+}
```
cptartur, line 54: Not sure if resolving paths to packages should be responsibility of this package - it makes it less flexible. Why not provide paths at the caller site? 

The paths when running `cargo test` have a stable working directory, so we can just hardcode them at caller site https://doc.rust-lang.org/cargo/commands/cargo-test.html#working-directory-of-tests

crates/forge/tests/e2e/templates.rs
```diff
@@ -0,0 +1,38 @@
...
+            "--template",
+            template.to_string().as_str(),
+            &package_name,
+        ])
+        .assert()
+        .success();
+
+    let package_path = temp_dir.child(package_name);
+    let package_path = Utf8PathBuf::from_path_buf(package_path.to_path_buf())
+        .expect("Failed to convert to Utf8PathBuf");
```
cptartur, line 29: We can probably hardcode these, see my comment about paths below

crates/forge/tests/e2e/templates.rs
cptartur: +1 for `toml_edit`, we already use it in some places.

cptartur (changes requested):

.github/workflows/ci.yml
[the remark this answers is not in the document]
franciszekjob: No, because then we will use version from `.tool-versions` file, which ATM is 2.9.4.
Once minimal supported is 2.10, then we can remove this step (as mentioned in the comment).

crates/forge/tests/e2e/templates.rs
franciszekjob: Applied

Cargo.toml
franciszekjob: I don't think so. Note that above we do the same:

https://github.com/foundry-rs/starknet-foundry/blob/master/Cargo.toml#L9-L10

crates/testing/packages_validation/src/lib.rs
franciszekjob: > Not sure if resolving paths to packages should be responsibility of this package - it makes it less flexible. Why not provide paths at the caller site?

Valid point, applied

crates/forge/tests/e2e/templates.rs
franciszekjob: In this case w we shouldn't hardcode them - note that these the package name is dynamic, depending on specific test case.

crates/forge/tests/code_quality.rs
```diff
@@ -0,0 +1,18 @@
+use camino::Utf8PathBuf;
+use packages_validation::check_and_lint; // Ensure you have this at the top
+
```
kkawula, line 3: What is the idea behind this?

kkawula (approved):

crates/forge/tests/code_quality.rs
franciszekjob: Leftover, removed 😅

crates/forge/tests/e2e/templates.rs
cptartur: Test case can have a reference to the template and a path to it. But maybe that's more complicated than the current solution

crates/forge/tests/code_quality.rs
```diff
@@ -0,0 +1,18 @@
...
+    let package_path: Utf8PathBuf = std::env::current_dir()
+        .expect("Failed to get current directory")
+        .ancestors()
+        .nth(2)
+        .expect("Failed to get grandparent directory")
+        .join("snforge_std")
+        .canonicalize()
+        .expect("Failed to canonicalize path")
+        .try_into()
+        .expect("Failed to convert to Utf8PathBuf");
```
cptartur: Why not just provide this path directly if the working directory of tests is stable?

```suggestion
    let package_path = Utf8PathBuf::from("../../snforge_std")
        .canonicalize()
        .unwrap()
        .try_into()
        .unwrap();
```

crates/sncast/tests/code_quality.rs
```diff
@@ -0,0 +1,18 @@
+use camino::Utf8PathBuf;
+use packages_validation::check_and_lint;
+
+#[test]
+fn validate_sncast_std() {
+    let package_path: Utf8PathBuf = std::env::current_dir()
```
cptartur: Same here

crates/forge/tests/e2e/templates.rs
franciszekjob: I would stay with current one tbh.

cptartur (approved):

PR #3143, opened 2025-03-26
Add tests using `snforge_std` from registry

### LINKED ISSUE #3060: Add tests using `snforge_std` from registry
### Current State

Currently, our testing suite mostly uses `snforge_std` and plugin from local paths. This is not good for detecting errors with registry and building proces.

### Objective

Add tests using `snforge_std` from registry

### Additional Context

There should be tests for both using prebuilts and building locally.

We could utilize a dev environment on registry or some locally hosted registry for this.

The exact scope of these tests should be decided but they should at least cover the cases related to packages from registry building incident.

### LINKED ISSUE #3069: Add tests to ensure that `reexport-plugin` works correctly
### Current State

Not tested

### Objective

Add tests to ensure that `reexport-plugin` works correctly

### Additional Context

_No response_

### LINKED ISSUE #3124: Publish lockfiles with snforge plugin
### Current State

.

### Objective

Publish lockfiles with snforge plugin

### Additional Context

This requires bumping the publish workflow to use the latest scarb version (after they have published it).

kkawula:
Closes #3060 
Closes #3069 
Closes #3124

## Introduced changes

Current State:
- `smoke_test.sh` - run binary with
    - `snforge_std` dependency from github link and the appropriate rev
- `release workflow` - build foundry binaries, run smoke tests, and apart from it publish plugin and stds to the official registry, it doesn't matter if tests pass or not
- `scheduled tests` - run cargo test

New State:
- `smoke_test.sh` - run binary with
     - `snforge_std` dependency from github link and the appropriate rev,
     - dev registry dependency, also check if prebuild works
     - dev registry dependency, without prebuild
- `release workflow`- build foundry binaries, build plugin binaries, publish plugin and stds to dev registry, run smoke tests, and only if passed, publish it to the official registry
- `scheduled tests` - run cargo test, build foundry binaries, build plugin binaries, publish plugin and stds to dev registry, run smoke tests,

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


.github/workflows/_build-plugin-binaries.yml
kkawula, line 1: Added inputs and related IFs, the rest is moved from the original workflow.

.github/workflows/_build-binaries.yml
kkawula, line 1: Everything moved from original workflow

.github/workflows/_test-binaries.yml
kkawula, line 1: Added `std_version` input and consequently the VERSION variable

.github/workflows/publish_std.yml
kkawula, line 1: New one

.github/workflows/scheduled.yml
```diff
@@ -4,14 +4,14 @@ on:
...
-      - .github/workflows/scheduled.yml
+      # - .github/workflows/scheduled.yml
   schedule:
     - cron: '0 0 * * 3,0'
   workflow_dispatch:
 
 jobs:
   get-scarb-versions:
-    if: "! github.event.repository.fork"
+    if: github.event.repository.fork == false
```
kkawula, line 14: Changed after `actions` vscode extension suggention

.github/workflows/_test-binaries.yml
```diff
@@ -0,0 +1,98 @@
...
+      - uses: dtolnay/rust-toolchain@7b1c307e0dcbda6122208f10795a713336a9b35a
+        with:
+          toolchain: stable
+
+      - name: Download artifacts
+        uses: actions/download-artifact@v4
+        with:
+          path: artifacts-dl
+
+      - name: Move artifacts to staging director
```
cptartur, line 47:
```suggestion
      - name: Move artifacts to staging directory
```

.github/workflows/publish_plugin.yml
```diff
@@ -2,7 +2,27 @@ name: Publish snforge_scarb_plugin
...
+        required: false
+        type: string
+
   workflow_dispatch:
+    inputs:
+      dev_registry:
+        required: false
+        type: boolean
+        default: false
+      override_plugin_version:
```
cptartur, line 20: Same here with the rename

.github/workflows/publish_std.yml
```diff
@@ -0,0 +1,72 @@
...
+  workflow_call:
+    inputs:
+      dev_registry:
+        required: false
+        type: boolean
+        default: false
+      override_std_version:
+        required: false
+        type: string
+      # snforge_std in the repository has a plugin dependency specified as a relative path, which must be overridden each time before publishing.
```
cptartur, line 13: Is this comment relevant here?

.github/workflows/scheduled.yml
```diff
@@ -4,14 +4,14 @@ on:
   pull_request:
     paths:
       - scripts/get_scarb_versions.sh
-      - .github/workflows/scheduled.yml
+      # - .github/workflows/scheduled.yml
```
cptartur, line 7: Why not remove this entirely?

cptartur (commented):
@THenry14 can you take a look at this?

.github/workflows/_build-plugin-binaries.yml
```diff
@@ -0,0 +1,142 @@
...
+
+on:
+  workflow_call:
+    inputs:
+      override_plugin_version:
+        required: false
+        type: string
+  workflow_dispatch:
+    inputs:
+      override_plugin_version:
```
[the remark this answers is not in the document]
kkawula: Changed

.github/workflows/_build-plugin-binaries.yml
```diff
@@ -0,0 +1,142 @@
...
+      plugin_uploaded: ${{ steps.check-version.outputs.plugin_uploaded }}
+    steps:
+      - uses: actions/checkout@v4
+      - name: Check version
+        id: check-version
+        run: |
+          set -exo pipefail
+
+          if [ -z "${{ inputs.overridden_plugin_version }}" ]; then
+            snforge_scarb_plugin_version=$(grep version crates/snforge-scarb-plugin/Scarb.toml | cut -d '"' -f 2)
```
THenry14, line 29: nit: at the moment there is only one ocurence of "version" so it's fine, but maybe we could improve this in case anything more with `.*version*` appears in the toml? I think grep-ing with regex like `^version =` could be more than enough, but am not sure. You could use jq as well.

THenry14: applicable to other places like this as well

.github/workflows/publish_plugin.yml
```diff
@@ -2,7 +2,27 @@ name: Publish snforge_scarb_plugin
...
   workflow_dispatch:
+    inputs:
+      dev_registry:
+        required: false
+        type: boolean
+        default: false
+      overridden_plugin_version:
+        required: false
+        type: string
+      build_plugin_binaries:
```
THenry14, line 23: what is the use case for not building here? 🤔

.github/workflows/publish_plugin.yml
```diff
@@ -17,124 +37,39 @@ jobs:
...
-        with:
-          name: build-plugin-${{ matrix.target }}
-          path: ${{ env.OUTPUT_BINARY_PATH }}
-          compression-level: 0
 
   upload-to-registry:
     name: Upload snforge_scarb_plugin to the registry
     runs-on: ubuntu-latest
+    if: |
+      always() &&
```
THenry14, line 61: I don't quite understand this, can you please explain why this is needed?

.github/workflows/publish_plugin.yml
```diff
@@ -153,4 +88,15 @@ jobs:
...
+          
+          if [ -n "${{ inputs.overridden_plugin_version }}" ]; then
+            sed -i.bak "/\[package\]/,/version =/ s/version = \".*/version = \"${{ inputs.overridden_plugin_version }}\"/" Scarb.toml
+            rm Scarb.toml.bak 2> /dev/null
+
+            sed -i.bak "/\[package\]/,/version =/ s/version = \".*/version = \"${{ inputs.overridden_plugin_version }}\"/" Cargo.toml
+            rm Cargo.toml.bak 2> /dev/null
+          fi
+
+          scarb publish --allow-dirty ${{ inputs.dev_registry == true && '--index https://scarbs.dev/' || '' }}
```
THenry14, line 102: I'm a bit afraid of this, mainly if it won't be too easy to upload a new/test version to prod by mistake. Maybe we should set dev registry to be a default instead?

.github/workflows/scheduled.yml
```diff
@@ -117,10 +113,73 @@ jobs:
...
+
+  build-plugin-binaries:
+    name: Build plugin binaries
+    if: github.event_name == 'workflow_dispatch' || github.event_name == 'schedule'
+    needs: verify-version
+    uses: ./.github/workflows/_build-plugin-binaries.yml
+    with:
+      overridden_plugin_version: ${{ needs.verify-version.outputs.version }}-test.${{ github.sha }}${{ github.run_id }}
+          
+  publish-plugin:
```
THenry14, line 145: We might want to connect to scarb team and discuss how to tackle dev registry cleanup, may be periodically. With this, we'll very soon bloat the storage, which can be costly.

scripts/smoke_test.sh
```diff
@@ -8,20 +8,53 @@ SNFORGE_PATH="$2"
...
+scarb cache clean
 
 # Check cast
 
 if ! $SNCAST_PATH call \
     --url "$RPC_URL" \
     --contract-address 0x06b248bde9ce00d69099304a527640bc9515a08f0b49e5168e2096656f207e1d \
     --function "get" --calldata 0x1 | grep -q $'command: call\nresponse: [0x0]'; then
-  exit 1
+    exit 1
```
THenry14, line 59:
```suggestion
  exit 1
```

nit. bash generally uses 2 spaces for iindent

.github/workflows/publish_plugin.yml
THenry14: But we need built binaries in either scenario right? can't we make it that no matter what called this workflow, the build workflow is always called as well (because publish depends on it)? 

> plugin artifacts are built out of this workflow. So we only build plugin binaries in this workflow when we want to publish the plugin manually

shouldn't the flow be exactly the same at all times though? maybe I'm still misunderstanding something here

kkawula: > But we need built binaries in either scenario right? can't we make it that no matter what called this workflow, the build workflow is always called as well (because publish depends on it)?

Unfortunately, we can't because in `release` workflow we use this workflow `publish_plugin` twice, and because of that we can build binaries here. It is because building binaries also requires uploading artifacts with `uplaod artifact` step. In the scenario of uploading the same artifact twice, it generates an error.

.github/workflows/_build-binaries.yml
```diff
@@ -0,0 +1,87 @@
...
+    inputs:
+      version:
+          required: true
+          type: string
+
+jobs:
+  build-binaries:
+    name: Build ${{ matrix.target }}
+    runs-on: ${{ matrix.os }}
+    continue-on-error: true
```
THenry14, line 14: I know this one is moved, but do you know what was the rationale to have that?
This might just waste resources, but perhaps there is a good reason for it?

.github/workflows/_build-plugin-binaries.yml
```diff
@@ -0,0 +1,142 @@
...
+            snforge_scarb_plugin_version="${{ inputs.overridden_plugin_version }}"
+          fi
+          
+          plugin_uploaded=$(curl -s https://scarbs.xyz/api/v1/index/sn/fo/snforge_scarb_plugin.json | jq --arg version "$snforge_scarb_plugin_version" '[.[] | select(.v == $version)] | length > 0')
+          echo "plugin_uploaded=$plugin_uploaded" >> $GITHUB_OUTPUT
+
+  build-binaries:
+    name: Build ${{ matrix.target }}
+    needs: check-version
+    if: needs.check-version.outputs.plugin_uploaded == 'false' || github.event_name == 'workflow_dispatch'
```
THenry14, line 40: what happens if there is a 'workflow_dispatch' event but the plugin was already uploaded? shouldn't this be skipped as well?

.github/workflows/_publish_plugin.yml
```diff
@@ -0,0 +1,77 @@
...
+        run: |
+          set -exo pipefail
+
+          if [ -z "${{ inputs.overridden_plugin_version }}" ]; then
+            snforge_scarb_plugin_version=$(grep "^version =" crates/snforge-scarb-plugin/Scarb.toml | cut -d '"' -f 2)
+          else
+            snforge_scarb_plugin_version="${{ inputs.overridden_plugin_version }}"
+          fi
+          plugin_uploaded=$(curl -s https://scarbs.xyz/api/v1/index/sn/fo/snforge_scarb_plugin.json | jq --arg version "$snforge_scarb_plugin_version" '[.[] | select(.v == $version)] | length > 0')
+          echo "plugin_uploaded=$plugin_uploaded" >> $GITHUB_OUTPUT
```
THenry14, line 32: [nitpick]: it seems this could be externalized. What I mean is that the logic is not really straightforward, and exactly the same thing is being used n several places; it would be far simpler to keep it one place and only update it there is there is a need (it would also be less error-prone). It could be separate workflow or just a bash script somewhere in repo.

.github/workflows/_test-binaries.yml
```diff
@@ -0,0 +1,98 @@
...
+            tar xzvf ${{ env.BINARY_PATH }}
+          fi
+
+      - name: Install universal-sierra-compiler
+        uses: software-mansion/setup-universal-sierra-compiler@v1
+
+      - name: Smoke test
+        shell: bash
+        env:
+          RPC_URL: "http://188.34.188.184:7070/rpc/v0_8"
```
THenry14, line 78: *sight*, I guess this was here before? :( It was probably leaked number of times before so my devops heart won't be mended, but please let's rather use `secrets.NODE_URL` or `secrets.CHEATNET_RPC_URL` (not sure what are the values, because they cannot be seen directly - we might want to double check them) and not hardcode in yamls 🙏

.github/workflows/release.yml
```diff
@@ -256,32 +131,17 @@ jobs:
 
   publish-snforge-scarb-plugin:
     name: Publish snforge_scarb_plugin
-    uses: ./.github/workflows/publish_plugin.yml
+    needs: [test-binary]
```
THenry14, line 134: I love it ❤️

scripts/smoke_test.sh
```diff
@@ -8,14 +8,47 @@ SNFORGE_PATH="$2"
...
-$SNFORGE_PATH init my_project
-pushd my_project || exit
+$SNFORGE_PATH new my_project_0
+pushd my_project_0 || exit
 scarb add --dev snforge_std --git "$REPO_URL" --rev "$REVISION"
 $SNFORGE_PATH test || exit
 popd || exit
+scarb cache clean
+
+# Check forge std from registry prebuild
```
THenry14, line 22: nit: let's improve on the comments, `Check forge std from registry prebuild` vs `Check forge std from registry build` is not obvious at all

.github/workflows/_build-binaries.yml
cptartur: It seems to be there since the original workflow was created, probably copied over from Scarb. And they don't seem to no longer have it defined (@kkawula please verify that).

.github/workflows/_build-plugin-binaries.yml
cptartur: Probably should still be skipped, upload will fail anyway if they are already present on the registry.

.github/workflows/_test-binaries.yml
cptartur: It's already hardcoded in cast codebase itself 😅 

Agree we can move it to secrets for "correctness" sake but ideally it'd have to be moved in bunch of other places as well.

cptartur: @kkawula you can create a follow-up issue, this might actually be worth tackling.

.github/workflows/publish_std.yml
```diff
@@ -0,0 +1,70 @@
...
+      - name: Publish snforge_std
+        working-directory: snforge_std
+        run: |
+          if [ -n "${{ inputs.override_std_version }}" ]; then
+            sed -i.bak "/\[package\]/,/version =/ s/version = \".*/version = \"${{ inputs.override_std_version }}\"/" Scarb.toml
+            rm Scarb.toml.bak 2> /dev/null
+          fi
+          if ${{ inputs.prod_registry == true }}; then
+            scarb add snforge_scarb_plugin@${{ inputs.plugin_dep_version }}
+          else
```
cptartur, line 65: Will this override the version that is set as path in `snforge_std`?

.github/workflows/scheduled.yml
```diff
@@ -117,10 +113,71 @@ jobs:
 
       - run: cargo test --release -p sncast
 
+  verify-version:
```
cptartur, line 116: Maybe it should be named something like `get_version`? From what I understand it just get's the version from `Cargo.toml`.

scripts/smoke_test.sh
```diff
@@ -8,14 +8,47 @@ SNFORGE_PATH="$2"
...
+test_output=$($SNFORGE_PATH test)
+test_exit=$?
+
+if [[ $test_exit -ne 0 ]] || echo "$test_output" | grep -q 'Compiling snforge_scarb_plugin'; then
+    exit 1
+fi
+popd || exit
+scarb cache clean
+
+# Check forge std from registry build
```
cptartur, line 40: Aren't these opposite? Here you allow prebuild plugin, in the code above not.

.github/workflows/_build-binaries.yml
kkawula: Scarb team removed this line, but imo we can leave it as it is. It can be helpful to detect which targets are affected, as it was last time when only linux related targets had a building issue

.github/workflows/_test-binaries.yml
kkawula: I've updated this workflow to use secret. This url only appears once in workflows.

scripts/smoke_test.sh
kkawula: Not exactly, there is removal of `allow-prebuilt-plugins`, the last part of the sed command consists of `d` which means delete.

.github/workflows/publish_std.yml
kkawula: yes, this command overrides entire dependency

.github/workflows/_build-plugin-binaries.yml
kkawula: It can be removed, as I can see, but I will test it on the fork to be sure.

.github/workflows/_publish_plugin.yml
kkawula: Done

.github/workflows/_build-binaries.yml
```diff
@@ -0,0 +1,86 @@
...
+        uses: taiki-e/install-action@cross
+
+      - name: Enable cross-compilation
+        if: matrix.cross
+        shell: bash
+        run: |
+          echo "CARGO=cross" >> $GITHUB_ENV
+
+      - name: Build
+        run: ${{ env.CARGO }} build --release --locked --target ${{ matrix.target }}
```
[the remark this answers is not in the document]
franciszekjob: nit: Also, I think we can set default shell under `defaults`, see [docs](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/setting-a-default-shell-and-working-directory#setting-default-shell-and-working-directory)

.github/workflows/_publish_plugin.yml
```diff
@@ -0,0 +1,72 @@
...
+          
+          plugin_uploaded=$(curl -s https://scarbs.xyz/api/v1/index/sn/fo/snforge_scarb_plugin.json | jq --arg version "$snforge_scarb_plugin_version" '[.[] | select(.v == $version)] | length > 0')
+          echo "plugin_uploaded=$plugin_uploaded" >> $GITHUB_OUTPUT
+
+
+  upload-to-registry:
+    name: Upload snforge_scarb_plugin to the registry
+    runs-on: ubuntu-latest
+    if: needs.check-version.result == 'success'
+    needs: [check-version]
```
franciszekjob, line 38: Isn't this if redundant?

.github/workflows/scheduled.yml
```diff
@@ -117,10 +113,72 @@ jobs:
...
+        uses: mikefarah/yq@bbdd97482f2d439126582a59689eb1c855944955
+        with:
+          cmd: yq -oy '.workspace.package.version' 'Cargo.toml'
+          
+      - name: Compare versions
+        id: validVersion
+        run: |
+          COMMIT_VERSION=${{ steps.lookupVersion.outputs.result }}
+          echo "Project version from this commit = $COMMIT_VERSION"
+          echo "version=$COMMIT_VERSION" >> "$GITHUB_OUTPUT"
```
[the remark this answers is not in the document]
kkawula: Renamed

.github/workflows/_build-binaries.yml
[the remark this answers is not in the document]
kkawula: This part was moved from the original workflows, so I would prefer to leave it as it is

.github/workflows/publish_plugin.yml
```diff
@@ -1,156 +1,27 @@
 name: Publish snforge_scarb_plugin
 
 on:
-  workflow_call:
   workflow_dispatch:
+    inputs:
+      prod_registry:
+        required: false
+        type: boolean
```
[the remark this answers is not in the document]
kkawula: We are telling to use the prod registry, we do not check anything, then variant without `is_` seems more natural for me

.github/workflows/_build-binaries.yml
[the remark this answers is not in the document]
kkawula: There is ambiguity, I will fix that, but first underscore is to suggest that this workflow is in kind of utility, but other pauses should be dashes. I am not sure is there any naming convention, but eachof  the most popular workflows usesa  single dash as a delimiter

.github/workflows/_publish_plugin.yml
kkawula: it is, thanks

.github/workflows/publish_plugin.yml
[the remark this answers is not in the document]
franciszekjob: Right, then maybe `use_prod_registry`?

franciszekjob (approved):

.github/workflows/_publish-plugin.yml
```diff
@@ -0,0 +1,70 @@
...
+    inputs:
+      prod_registry:
+        required: false
+        type: boolean
+      overridden_plugin_version:
+        required: false
+        type: string
+
+jobs:
+  check-version:
```
cptartur, line 14: Nit, maybe rename to `check-uploaded`?

.github/workflows/_test-binaries.yml
```diff
@@ -0,0 +1,98 @@
...
+
+      - name: Get binary path
+        shell: bash
+        run: |
+          if [[ ${{ matrix.target }} == *windows* ]]; then
+            BINARY_PATH="artifacts/starknet-foundry-${{ inputs.bin_version }}-${{ matrix.target }}.zip"
+          else
+            BINARY_PATH="artifacts/starknet-foundry-${{ inputs.bin_version }}-${{ matrix.target }}.tar.gz"
+          fi
+          echo "BINARY_PATH=$BINARY_PATH" >> $GITHUB_ENV
```
cptartur, line 61: Nit, maybe this could be renamed `Get artifact path`?

cptartur: And variables renamed accordingly

cptartur (approved):
Looks good, left some nits about variable/steps names

.github/workflows/_test-binaries.yml
kkawula: Renamed

.github/workflows/_publish-plugin.yml
kkawula: Renamed

PR #3179, opened 2025-04-01
Fix test result printing

kkawula:
Closes #

## Introduced changes

-

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/forge-runner/src/printing.rs
```diff
@@ -147,8 +147,13 @@ fn result_header(any_test_result: &AnyTestCaseSummary) -> String {
 }
 
 fn result_debug_trace(any_test_result: &AnyTestCaseSummary) -> String {
```
ksew1, line 149: [nit] 

```rs
fn result_debug_trace(any_test_result: &AnyTestCaseSummary) -> String {
    any_test_result
        .debugging_trace()
        .map(|trace| format!("\n{trace}"))
        .unwrap_or_default()
}
```

crates/forge-runner/src/printing.rs
```diff
@@ -147,8 +147,13 @@ fn result_header(any_test_result: &AnyTestCaseSummary) -> String {
...
-    any_test_result
+    let trace = any_test_result
         .debugging_trace()
         .map(ToString::to_string)
-        .unwrap_or_default()
+        .unwrap_or_default();
+    if trace.is_empty() {
+        trace
+    } else {
+        format!("\n{trace}")
```
ksew1, line 157: Any reason why we add now one newline instead of two? 😅

franciszekjob (commented):
We should to have a test which checks exact output. Could you add it?

franciszekjob (approved):

crates/shared/src/test_utils/output_assert.rs
```diff
@@ -67,6 +67,14 @@ fn assert_output_contains(output: &str, lines: &str) {
 
     assert!(contains, "Output does not match:\n\n{out}");
 }
+fn assert_output_contains_exact(output: &str, lines: &str) {
```
cptartur, line 70: It should be just named `assert_output`. So it suggests that it basically compares the expected with the output.

crates/shared/src/test_utils/output_assert.rs
```diff
@@ -67,6 +67,14 @@ fn assert_output_contains(output: &str, lines: &str) {
...
+    let converted_pattern = regex::escape(lines).replace(r"\[\.\.\]", ".*");
```
cptartur, line 71: And we probably shouldn't allow patters in it at all.

kkawula: It will be quite annoying to update this test when resources change in the future. We never assert them.

crates/shared/src/test_utils/output_assert.rs
kkawula: renamed

cptartur (approved):

Issue #2806, opened 2024-12-20
Investigate tests that assert builtins


ddoktorski:
### Which components does the task require to be changed? (think hard pls)

snforge

### Description

Review the tests that check builtins in execution resources (mostly those using the `assert_builtin` function) to ensure they make sense and investigate whether they can be improved.

ddoktorski:
In the scope of this task consider adding more tests for calldata factor, especially related to gas when tracking Cairo steps.

PR #3238, opened 2025-04-18
Run tests through blockifier logic

cptartur:
**Stack**:
- #3333
- #3328
- #3327
- #3247
- #3246
- #3245
- #3244
- #3243
- #3242
- #3241
- #3240
- #3238 ⬅


crates/cheatnet/src/state.rs
```diff
@@ -343,7 +343,8 @@ impl Default for CheatnetState {
             ClassHash(TryFromHexStr::try_from_hex_str(TEST_CONTRACT_CLASS_HASH).unwrap());
         let test_call = Rc::new(RefCell::new(CallTrace {
             entry_point: test_code_entry_point.into(),
-            run_with_call_header: true,
+            // TODO this field could be removed entirely
+            run_with_call_header: false,
```
franciszekjob, line 347: Will it be removed then?

crates/forge-runner/src/running.rs
```diff
@@ -166,6 +160,55 @@ pub struct RunCompleted {
...
 
+pub struct VmExecutionContext<'a> {
+    pub runner: CairoRunner,
+    pub syscall_handler: SyscallHintProcessor<'a>,
+    pub initial_syscall_ptr: Relocatable,
+    // Additional data required for execution is appended after the program bytecode.
+    pub program_extra_data_length: usize,
+}
+
+fn initialize_execution_context<'a>(
```
franciszekjob, line 171: It could be a method of `VmExecutionContext`

crates/forge-runner/src/running.rs
```diff
@@ -264,44 +318,87 @@ pub fn run_test_case(
...
+                    .extended_runtime
+                    .extended_runtime
+                    .extended_runtime
+                    .hint_handler,
+                n_total_args,
+                program_extra_data_length,
+                tracked_resource,
+            )?;
+
+            // TODO this can be done better, we can take gas directly from call info
```
franciszekjob, line 378: Will it be addressed in this PR (or issue will be added)?

crates/forge-runner/src/running.rs
```diff
@@ -327,79 +424,34 @@ pub fn run_test_case(
     let transaction_context = get_context(&forge_runtime).tx_context.clone();
     let used_resources =
         get_all_used_resources(forge_runtime, &transaction_context, tracked_resource);
-    let gas_used = calculate_used_gas(
+    let gas = calculate_used_gas(
```
franciszekjob, line 427: Any specific reason for this change?

crates/forge-runner/src/running/copied_code.rs
```diff
@@ -0,0 +1,384 @@
...
+        }
+    }
+
+    Ok(())
+}
+
+// Reason copied: Private function
+pub(crate) fn prepare_program_extra_data(
+    runner: &mut CairoRunner,
+    // contract_class: &CompiledClassV1,
```
franciszekjob, line 137: Looks like it can be removed (?)

crates/forge-runner/src/running/copied_code.rs
```diff
@@ -0,0 +1,384 @@
...
+        resources: vm_resources,
+        storage_read_values: syscall_handler_base.read_values.clone(),
+        accessed_storage_keys: syscall_handler_base.accessed_keys.clone(),
+        read_class_hash_values: syscall_handler_base.read_class_hash_values.clone(),
+        accessed_contract_addresses: syscall_handler_base.accessed_contract_addresses.clone(),
+    })
+}
+
+#[allow(clippy::trivially_copy_pass_by_ref)]
+/// Reason copied: Required by `finalize_execution`
```
franciszekjob, line 258:
```suggestion
// Reason copied: Required by `finalize_execution`
```

crates/forge-runner/src/running/copied_code.rs
```diff
@@ -0,0 +1,384 @@
...
+    };
+    Ok(CallResult {
+        failed,
+        retdata: read_execution_retdata(runner, retdata_size, retdata_start)?,
+        gas_consumed,
+    })
+}
+
+#[allow(unused_variables)]
+/// Reason copied: Removed code for test function arguments
```
franciszekjob, line 315:
```suggestion
// Reason copied: Removed code for test function arguments
```

crates/forge-runner/src/running/copied_code.rs
```diff
@@ -0,0 +1,384 @@
...
+    }
+    // Pre-charge entry point's initial budget to ensure sufficient gas for executing a minimal
+    // entry point code. When redepositing is used, the entry point is aware of this pre-charge
+    // and adjusts the gas counter accordingly if a smaller amount of gas is required.
+    let call_initial_gas = call
+        .initial_gas
+        .checked_sub(entry_point_initial_budget)
+        .ok_or(PreExecutionError::InsufficientEntryPointGas)?;
+    // Push gas counter.
+    // dbg!(&call_initial_gas);
```
franciszekjob, line 362:
```suggestion
```

crates/forge-runner/src/running/copied_code.rs
```diff
@@ -0,0 +1,384 @@
...
+        .map_err(CairoRunError::VirtualMachine)?;
+
+    Ok(())
+}
+
+#[allow(clippy::items_after_statements, clippy::needless_pass_by_value)]
+/// Why copied: Required by `run_entry_point`
+///
+/// Fills the holes after running the entry point.
+/// Currently only fills the holes in the rc96 segment.
```
franciszekjob, line 63:
```suggestion
// Why copied: Required by `run_entry_point`
//
// Fills the holes after running the entry point.
// Currently only fills the holes in the rc96 segment.
```

crates/forge-runner/src/package_tests.rs
```diff
@@ -8,6 +17,20 @@ pub mod raw;
...
+    (BuiltinName::add_mod, AddModType::ID),
+    (BuiltinName::range_check96, RangeCheck96Type::ID),
+    (BuiltinName::segment_arena, SegmentArenaType::ID),
+    (BuiltinName::poseidon, PoseidonType::ID),
+    (BuiltinName::ec_op, EcOpType::ID),
+    (BuiltinName::bitwise, BitwiseType::ID),
+    (BuiltinName::range_check, RangeCheckType::ID),
+    (BuiltinName::pedersen, PedersenType::ID),
+];
+
```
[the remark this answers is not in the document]
kkawula: Maybe there is some kind of const defined in blockierfier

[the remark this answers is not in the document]
cptartur: > Can we somehow ensure that the order won't be broken (for instance by adding a test)?

This doesn't prevent anyone from changing the test along with the const. If the comment is not enough to stop them, then I'm not sure if tests changes anything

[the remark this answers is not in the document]
cptartur: > Maybe there is some kind of const defined in the blockifier

Couldn't find any const for that, the order here corresponds to `#[implicit_precedence(...)]` attribute that is generated by the cairo compiler for contracts

crates/cheatnet/src/state.rs
cptartur: In the last PR of the stack it is removed

crates/forge-runner/src/running.rs
cptartur: It's (very loosely at this point) based of blockifier code, I will add a relevant comment here.

crates/forge-runner/src/running.rs
cptartur: I want to do it in the separate PR. Linked https://github.com/foundry-rs/starknet-foundry/issues/3292

crates/forge-runner/src/running.rs
cptartur: Changed back to `gas_used`

crates/forge-runner/src/running/copied_code.rs
cptartur: The idea is to upstream as much of the code from this file as possible so don't want to modify it yet. Linked issue https://github.com/foundry-rs/starknet-foundry/issues/3293

crates/forge-runner/src/package_tests.rs
```diff
@@ -23,6 +46,20 @@ pub struct TestDetails {
...
+impl TestDetails {
+    #[must_use]
+    pub fn builtins(&self) -> Vec<BuiltinName> {
+        let mut builtins = vec![];
+        for (builtin_name, builtin_ty) in BUILTIN_ORDER {
+            if self.parameter_types.iter().any(|(ty, _)| ty == &builtin_ty) {
+                builtins.push(builtin_name);
+            }
+        }
+        builtins.reverse();
```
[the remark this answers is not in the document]
cptartur: Leftover logic, reversed the const instead

crates/forge-runner/src/running.rs
```diff
@@ -264,44 +318,87 @@ pub fn run_test_case(
...
+        initial_syscall_ptr,
+        &mut forge_runtime
+            .extended_runtime
+            .extended_runtime
+            .extended_runtime
+            .hint_handler
+            .read_only_segments,
+        &entry_point,
+        entry_point_initial_budget,
+    )?;
```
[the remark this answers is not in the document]
cptartur: Extracted `entry_point_initial_budget` but for `prepare_call_arguments` I think it may be better to accept it, since it's using a function from (copied from) blockifier and I don't want to wrap it to hide this.

crates/snforge-scarb-plugin/tests/integration/multiple_attributes.rs
```diff
@@ -1,10 +1,19 @@
...
 use snforge_scarb_plugin::attributes::{available_gas::available_gas, fork::fork, test::test};
 
+fn last_n_lines(result: &ProcMacroResult, n_lines: usize) -> TokenStream {
+    let lines = result.token_stream.to_string();
+    let lines = lines
+        .lines()
+        .skip(lines.lines().count() - n_lines)
+        .collect::<String>();
+    TokenStream::new(lines)
+}
```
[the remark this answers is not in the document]
cptartur: Maybe we can refactor the tests to greater extend once we merge them? The problem now is that the `#[test]` attribute now generates two functions and logic in these tests doesn't really place nicely with that.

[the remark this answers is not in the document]
franciszekjob: > Maybe we can refactor the tests to greater extend once we merge them?

👍

franciszekjob:
@cptartur please re-request review once tests are fixed 🙏

crates/forge/tests/integration/resources.rs
```diff
@@ -51,7 +54,8 @@ fn builtins_count() {
     assert_passed(&result);
 
     // No ECDSA and Keccak builtins
-    assert_builtin(&result, "range_check", BuiltinName::range_check, 4);
+    assert_builtin(&result, "empty", BuiltinName::range_check, 5);
```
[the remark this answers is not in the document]
cptartur: I've investigated and these are my conclusions. 

1st, if you check `assert_builtin` it has this logic
```rust
// TODO(#2806)
    let expected_count = if builtin == BuiltinName::range_check {
        expected_count - 1
    } else {
        expected_count
    };
```
so the actual asserted value is 4.

2nd in the empty test the 4 usages of builtin come respectively from these lines

Usage 1 https://github.com/foundry-rs/starknet-foundry/blob/737445f8d52480d712aa5d858d2fe564d2084fd8/crates/snforge-scarb-plugin/src/attributes/test.rs#L65 because https://github.com/starkware-libs/cairo/blob/3bcb85488dfd3a8c39a9208a4f350609b024ec39/corelib/src/gas.cairo#L28

Usage 2 https://github.com/foundry-rs/starknet-foundry/blob/737445f8d52480d712aa5d858d2fe564d2084fd8/crates/snforge-scarb-plugin/src/attributes/test.rs#L68 because https://github.com/starkware-libs/cairo/blob/3bcb85488dfd3a8c39a9208a4f350609b024ec39/corelib/src/gas.cairo#L34

Usage 3 and 4 come from this https://github.com/foundry-rs/starknet-foundry/blob/737445f8d52480d712aa5d858d2fe564d2084fd8/snforge_std/src/cheatcode.cairo#L30
specifically because array method .at and .slice have implicit GasBuiltin as well https://github.com/starkware-libs/cairo/blob/3bcb85488dfd3a8c39a9208a4f350609b024ec39/corelib/src/array.cairo#L93

So asserted value of 5 (actually 4) makes sense to me. I've compared and on master, before my changes, empty test has 2 actual RangeCheck usages (from .at and .slice) and test with explicit range check has actual 3 (4 asserted). 

One problem with this logic is that in my changed code, both empty and range_check cases have asserted value of 5, when it should be 5 and 6. 

I assume this is due to some optimization that the compiler is doing, since the actual test code is now wrapped https://github.com/foundry-rs/starknet-foundry/blob/737445f8d52480d712aa5d858d2fe564d2084fd8/crates/snforge-scarb-plugin/src/attributes/test.rs#L78

I've tried adding `#[inline(never)]` to `fn {name}_return_wrapper{signature} {{` and the value for range_check test case increased to 6, while for empty it remained 5.

crates/forge-runner/src/package_tests.rs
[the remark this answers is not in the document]
piotmag769: If you break it you will notice it, Starknet won't ever change it

franciszekjob (approved):

crates/forge-runner/src/running.rs
```diff
@@ -138,43 +140,133 @@ pub(crate) fn run_fuzz_test(
...
+pub struct RunCompleted {
+    pub(crate) status: RunStatus,
     pub(crate) call_trace: Rc<RefCell<CallTrace>>,
     pub(crate) gas_used: GasVector,
     pub(crate) used_resources: UsedResources,
     pub(crate) encountered_errors: EncounteredErrors,
     pub(crate) fuzzer_args: Vec<String>,
 }
 
+pub struct VmExecutionContext<'a> {
```
ddoktorski, line 163: Maybe add a comment why the struct with the same name from the blockifier is not used here?

crates/forge-runner/src/running.rs
```diff
@@ -138,43 +140,133 @@ pub(crate) fn run_fuzz_test(
...
+    pub runner: CairoRunner,
+    pub syscall_handler: SyscallHintProcessor<'a>,
+    pub initial_syscall_ptr: Relocatable,
+    // Additional data required for execution is appended after the program bytecode.
+    pub program_extra_data_length: usize,
+}
+
+// Based on code from https://github.com/starkware-libs/sequencer/blob/e417a9e7d50cbd78065d357763df2fbc2ad41f7c/crates/blockifier/src/execution/entry_point_execution.rs#L122
+// Enough of the logic of this had to be changed that probably it won't be possible to upstream it
+fn initialize_execution_context<'a>(
```
ddoktorski, line 173: What do you think about following convention from the cheatnet crate to include in the modifed code comments like `// region: Modified blockifier code`?

crates/forge-runner/src/running.rs
```diff
@@ -138,43 +140,133 @@ pub(crate) fn run_fuzz_test(
...
+    Completed(Box<RunCompleted>),
+    Error(RunError),
+}
+
+// Builtins field is private in program, so we need this workaround
+fn builtins_from_program(program: &Program) -> Vec<BuiltinName> {
+    program.iter_builtins().copied().collect::<Vec<_>>()
+}
+
+fn entry_point_initial_budget(forge_runtime: &ForgeRuntime) -> u64 {
```
ddoktorski, line 231: nit: It could take `SyscallHintProcessor` as a parameter

crates/forge-runner/src/running.rs
```diff
@@ -138,43 +140,133 @@ pub(crate) fn run_fuzz_test(
...
+        .extended_runtime
+        .hint_handler
+        .base
+        .context
+        .gas_costs()
+        .base
+        .entry_point_initial_budget
+}
+
+fn build_test_call_and_entry_point(
```
ddoktorski, line 244: nit: Some of these helper functions could be moved to a separate file

crates/forge/tests/e2e/running.rs
```diff
@@ -797,17 +797,21 @@ fn incompatible_snforge_std_version_warning() {
...
         [FAIL] steps::tests::steps_more_than_10000000
-
+        
         Failure data:
-            Could not reach the end of the program. RunResources has no remaining steps.
-            Suggestion: Consider using the flag `--max-n-steps` to increase allowed limit of steps
-
-        Tests: 1 passed, 1 failed, 0 skipped, 0 ignored, 0 filtered out
-
+            Memory addresses must be relocatable
```
ddoktorski, line 803: Will this change break using the new forge with an old plugin?

crates/snforge-scarb-plugin/src/attributes/test.rs
```diff
@@ -48,12 +49,35 @@ fn test_internal(
...
+    let signature = func.declaration(db).signature(db).as_text(db);
+    let statements = func.body(db).statements(db).as_text(db);
+    let attributes = func.attributes(db).as_text(db);
+
     if should_run_test {
         Ok(formatdoc!(
             "
             #[snforge_internal_test_executable]
+            #[implicit_precedence(core::pedersen::Pedersen, core::RangeCheck, core::integer::Bitwise, core::ec::EcOp, core::poseidon::Poseidon, core::SegmentArena, core::circuit::RangeCheck96, core::circuit::AddMod, core::circuit::MulMod, core::gas::GasBuiltin, System)]
+            fn {name}(mut _data: Span<felt252>) -> Span::<felt252> {{
```
ddoktorski, line 62: What is exactly the `_data` parameter?

crates/snforge-scarb-plugin/src/attributes/fuzzer/wrapper.rs
```diff
@@ -39,11 +39,7 @@ fn fuzzer_wrapper_internal(
         if let Some(test_attr) = attr_list.find_attr(db, TestCollector::ATTR_NAME) {
             vec![test_attr]
         } else {
-            [
-                attr_list.query_attr(db, "snforge_internal_test_executable"),
-                attr_list.query_attr(db, InternalConfigStatementCollector::ATTR_NAME),
-            ]
-            .concat()
+            [attr_list.query_attr(db, InternalConfigStatementCollector::ATTR_NAME)].concat()
```
ddoktorski, line 42:
```suggestion
            attr_list.query_attr(db, InternalConfigStatementCollector::ATTR_NAME)
```

crates/snforge-scarb-plugin/tests/integration/multiple_attributes.rs
```diff
@@ -172,7 +212,6 @@ fn works_with_fuzzer_config_wrapper() {
     // due to the function having a parameter
     let item = TokenStream::new(formatdoc!(
         r"
-        #[snforge_internal_test_executable]
```
ddoktorski, line 175: It was supposed to simulate `#test` attribute, please update it accordingly

crates/snforge-scarb-plugin/tests/integration/multiple_attributes.rs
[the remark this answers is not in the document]
ddoktorski: If you will do this I am ok with it, but I'd prefer to avoid making this kind of changes on my side in the scope of the migrating to the new macro syntax 🙏

crates/forge/tests/integration/resources.rs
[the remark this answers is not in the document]
ddoktorski: > I guess that partially clears issue #2806 because we now know (assuming what I wrote is correct) where are the range_check usages coming from.

Yup, agree

> Additionally, maybe we should add #[inline(never)] to the test.rs code so that compiler doesn't optimize away user code from tests resulting in different builtin usage etc. Though I'm not sure about this.

I’d prefer not to do this, we should mostly test the default behavior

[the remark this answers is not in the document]
cptartur: The default behavior in this case might actually be the `#[inline(never)]`. Previously user wrote their tests and we run that, now we wrap them in another function so it might change the way compiler optimizes things. But I don't know the details of that.

crates/forge-runner/src/running.rs
cptartur: I don't think it makes sense for this case. If you take a look at the linked function you'll see that even the arguments don't match.

crates/forge-runner/src/running.rs
cptartur: Done

crates/forge-runner/src/running.rs
cptartur: On the caller site we have to then call it as

```rust
let entry_point_initial_budget = entry_point_initial_budget(
        &forge_runtime
            .extended_runtime
            .extended_runtime
            .extended_runtime
            .hint_handler,
    );
```

I think in this case it is clearer to hide that logic in the function itself.

crates/forge-runner/src/running.rs
cptartur: Moved to `setup.rs`.

crates/snforge-scarb-plugin/src/attributes/test.rs
cptartur: Contract entry points take their arguments as Span of felts. Modified the signature of tests so it includes this as well so it matches. We aren't passing any calldata to the tests though, so it is unused.

crates/snforge-scarb-plugin/tests/integration/multiple_attributes.rs
cptartur: With my changes, the `#[snforge_internal_test_executable]` will be on the outermost function (one emulating contract entrypoint). So the fuzzer wrapper will not have it. I could emulate it properly with adding the outer wrapper here but we would have to then skip it anyway for the next step of this test.

crates/forge/tests/e2e/running.rs
ddoktorski: It might be worth mentioning this in the changelog or somewhere else

crates/forge/tests/integration/resources.rs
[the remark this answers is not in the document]
ddoktorski: We can consider this, but it seems fine to leave it as it is for now.

crates/snforge-scarb-plugin/tests/integration/multiple_attributes.rs
ddoktorski: Actually wouldn't it be better to use the `last_n_lines` function here as well? Then we can work only on the return wrapper item

crates/forge-runner/src/running.rs
ddoktorski: I meant to call this function before creating runtimes for example after `initialize_execution_context` function, anyway this is a nitpick

crates/snforge-scarb-plugin/tests/integration/multiple_attributes.rs
cptartur: Reworked the test as suggested

crates/forge/tests/e2e/running.rs
cptartur: Will add a PR about it to the stack

ddoktorski (approved):

cptartur:

PR #3246, opened 2025-04-18
Update `sncast` crate so it depends on `cairo-lang-runner` and `cairo-lang-runnable-utils` directly

cptartur:
**Stack**:
- #3333
- #3328
- #3327
- #3247
- #3246 ⬅
- #3245
- #3244
- #3243
- #3242
- #3241
- #3240
- #3238

⚠️ *Part of a stack created by [spr](https://github.com/ejoffe/spr). Do not merge manually using the UI - doing so may have unexpected results.*

PR #3245, opened 2025-04-18
Remove `cairo-lang-runner` form `shared` crate

cptartur:
**Stack**:
- #3333
- #3328
- #3327
- #3247
- #3246
- #3245 ⬅
- #3244
- #3243
- #3242
- #3241
- #3240
- #3238

⚠️ *Part of a stack created by [spr](https://github.com/ejoffe/spr). Do not merge manually using the UI - doing so may have unexpected results.*

PR #3243, opened 2025-04-18
Remove `cairo-lang-runner` from `runtime` crate

cptartur:
**Stack**:
- #3333
- #3328
- #3327
- #3247
- #3246
- #3245
- #3244
- #3243 ⬅
- #3242
- #3241
- #3240
- #3238

⚠️ *Part of a stack created by [spr](https://github.com/ejoffe/spr). Do not merge manually using the UI - doing so may have unexpected results.*

PR #3242, opened 2025-04-18
Remove `cairo-lang-runner` from `conversions` crate

cptartur:
**Stack**:
- #3333
- #3328
- #3327
- #3247
- #3246
- #3245
- #3244
- #3243
- #3242 ⬅
- #3241
- #3240
- #3238

⚠️ *Part of a stack created by [spr](https://github.com/ejoffe/spr). Do not merge manually using the UI - doing so may have unexpected results.*

PR #3247, opened 2025-04-18
Remove always `false` `run_with_call_header`

### LINKED ISSUE #3072: Investigate invalid syscall selector error

snforge

### Foundry Version

0.38.3

Linux

x86

### What happened

(This issue was reported by one of out users on telegram, ref link: https://t.me/starknet_foundry_support/5652)

Image

the last uncommented line (let expected peak = ... ) is the one that makes the second test fail due to "Invalid syscall selector" error, but that same very similar line (almost the same) in the first test is working fine

test fails when poseidon_hash_bn254 function is called and the endpoint append_leaf_aocl of the contract in the same test.

Worth noting that append_leaf_aocl does uses the same poseidon_hash_bn254 internally (within the contract), but interacting with the contract and this endpoint alone causes no issues. 

Code is here https://github.com/keep-starknet-strange/garaga/tree/msets/src/contracts/mutator_set (all test passes in this code)

Function: https://github.com/keep-starknet-strange/garaga/blob/main/src/src/hashes/poseidon_bn254.cairo#L16

(cairo 2.9.2 was used here)

### Trace

_No response_

- [x] I have searched the existing issues and verified no issue exits for this problem.

cptartur:
**Stack**:
- #3333
- #3328
- #3327
- #3247 ⬅
- #3246
- #3245
- #3244
- #3243
- #3242
- #3241
- #3240
- #3238

### DIFF
--- a/crates/forge-runner/src/build_trace_data.rs
+++ b/crates/forge-runner/src/build_trace_data.rs
@@ -50,7 +50,6 @@ pub fn build_profiler_call_trace(
         vm_trace,
         contracts_data,
         versioned_program_path,
-        value.run_with_call_header,
     );
 
     ProfilerCallTrace {
@@ -75,7 +74,6 @@ fn build_cairo_execution_info(
     vm_trace: Option<Vec<ProfilerTraceEntry>>,
     contracts_data: &ContractsData,
     versioned_program_path: &Utf8Path,
-    run_with_call_header: bool,
 ) -> Option<CairoExecutionInfo> {
     let contract_name = get_contract_name(entry_point.class_hash, contracts_data);
     let source_sierra_path = contract_name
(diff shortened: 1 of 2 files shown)


franciszekjob (approved):

kkawula:
Closes #https://github.com/foundry-rs/starknet-foundry/issues/3072

crates/forge-runner/src/build_trace_data.rs
```diff
@@ -50,7 +50,7 @@ pub fn build_profiler_call_trace(
         vm_trace,
         contracts_data,
         versioned_program_path,
-        value.run_with_call_header,
+        false,
```
ddoktorski, line 53: Why not to remove the parameter `run_with_call_header` from the `build_cairo_execution_info` function and simply hardcode `false` in the return struct in this function?

cptartur: Good call, removed

ddoktorski (approved):

cptartur:

Issue #3009, opened 2025-02-26
Decode call data


ksew1:
### Which component is your feature related to?

Forge

### Feature Request

Needs reverse data transformer

PR #3260, opened 2025-04-22
Add reverse transformer

ksew1:
Related #3009

---

**Stack**:
- #3289
- #3288
- #3287
- #3286
- #3285
- #3264
- #3263
- #3262
- #3261
- #3260 ⬅

### DIFF
--- a/crates/data-transformer/src/reverse_transformer/mod.rs
+++ b/crates/data-transformer/src/reverse_transformer/mod.rs
@@ -1,0 +1,65 @@
+mod transform;
+mod types;
+
+use crate::reverse_transformer::transform::{ReverseTransformer, TransformationError};
+use crate::shared::extraction::extract_function_from_selector;
+use starknet::core::types::contract::AbiEntry;
+use starknet_types_core::felt::Felt;
+
+#[derive(Debug, thiserror::Error)]
+pub enum ReverseTransformError {
+    #[error(r#"Function with selector "{0}" not found in ABI of the contract"#)]
+    FunctionNotFound(Felt),
+    #[error(transparent)]
+    TransformationError(#[from] TransformationError),
+}
+
+/// Transforms a calldata into a Cairo-like string representation of the arguments
+pub fn reverse_transform_input(
+    input: &[Felt],
+    abi: &[AbiEntry],
+    function_selector: &Felt,
+) -> Result<String, ReverseTransformError> {
+    let input_types: Vec<_> = extract_function_from_selector(abi, *function_selector)
+        .ok_or(ReverseTransformError::FunctionNotFound(*function_selector))?
+        .inputs
+        .into_iter()
+        .map(|input| input.r#type)
+        .collect();
+
+    reverse_transform(input, abi, &input_types)
+}
+
+/// Transforms a call output into a Cairo-like string representation of the return values
+pub fn reverse_transform_output(
+    output: &[Felt],
+    abi: &[AbiEntry],
+    function_selector: &Felt,
+) -> Result<String, ReverseTransformError> {
+    let output_types: Vec<_> = extract_function_from_selector(abi, *function_selector)
+        .ok_or(ReverseTransformError::FunctionNotFound(*function_selector))?
+        .outputs
+        .into_iter()
+        .map(|input| input.r#type)
+        .collect();
+
+    reverse_transform(output, abi, &output_types)
+}
+
+fn reverse_transform(
+    felts: &[Felt],
+    abi: &[AbiEntry],
+    types: &[String],
+) -> Result<String, ReverseTransformError> {
+    let mut reverse_transformer = ReverseTransformer::new(felts, abi);
+
+    Ok(types
+        .iter()
+        .map(|parameter_type| {
+            Ok(reverse_transformer
+                .parse_and_transform(parameter_type)?
+                .to_string())
+        })
+        .collect::<Result<Vec<String>, TransformationError>>()?
+        .join(", "))
+}
--- a/crates/data-transformer/src/reverse_transformer/transform.rs
+++ b/crates/data-transformer/src/reverse_transformer/transform.rs
@@ -1,0 +1,258 @@
+use crate::reverse_transformer::types::{
+    Enum, Primitive, Sequence, SequenceType, Struct, StructField, Tuple, Type,
+};
+use crate::shared::parsing::{ParseError, parse_expression};
+use crate::shared::path::{GenericArgsExtractionError, SplitResult, split};
+use cairo_lang_parser::utils::SimpleParserDatabase;
+use cairo_lang_syntax::node::ast::{Expr, ExprListParenthesized, ExprPath};
+use conversions::serde::deserialize::{BufferReadError, BufferReader};
+use starknet::core::types::contract::{AbiEntry, AbiEnum, AbiStruct};
+use starknet_types_core::felt::Felt;
+
+/// An error that can occur during the transformation process.
+#[derive(Debug, thiserror::Error)]
+pub enum TransformationError {
+    #[error("type unsupported by reverse transformer")]
+    UnsupportedType,
+    #[error("reading from buffer failed with error: {0}")]
+    BufferReaderError(#[from] BufferReadError),
+    #[error("{enum_name} enum does not have variant at position {variant_position}")]
+    NoSuchEnumVariant {
+        enum_name: String,
+        variant_position: usize,
+    },
+    #[error("abi is invalid")]
+    InvalidAbi,
+    #[error(transparent)]
+    ParseError(#[from] ParseError),
+    #[error(transparent)]
+    GenericArgsExtractionError(#[from] GenericArgsExtractionError),
+}
+
+/// An error that can occur when trying to transform a primitive type.
+#[derive(Debug, thiserror::Error)]
+enum PrimitiveError {
+    #[error("Type {0} is not primitive")]
+    NotFound(String),
+    #[error(transparent)]
+    BufferReaderError(#[from] BufferReadError),
+}
+
+pub struct ReverseTransformer<'a> {
+    abi: &'a [AbiEntry],
+    buffer_reader: BufferReader<'a>,
+    db: SimpleParserDatabase,
+}
+
+impl<'a> ReverseTransformer<'a> {
+    /// Creates a new instance of [`ReverseTransformer`].
+    pub fn new(inputs: &'a [Felt], abi: &'a [AbiEntry]) -> Self {
+        Self {
+            abi,
+            buffer_reader: BufferReader::new(inputs),
+            db: SimpleParserDatabase::default(),
+        }
+    }
+
+    /// Parses the given `&str` to an [`Expr`] and then transforms it to a [`Type`].
+    pub fn parse_and_transform(&mut self, expr: &str) -> Result<Type, TransformationError> {
+        self.transform_expr(&parse_expression(expr, &self.db)?)
+    }
+
+    /// Transforms the given [`Expr`] to a [`Type`].
+    fn transform_expr(&mut self, expr: &Expr) -> Result<Type, TransformationError> {
+        match expr {
+            Expr::Tuple(expr) => self.transform_tuple(expr),
+            Expr::Path(expr) => self.transform_path(expr),
+            _ => Err(TransformationError::UnsupportedType),
+        }
+    }
+
+    /// Transforms a tuple expression to a [`Type::Tuple`].
+    fn transform_tuple(
+        &mut self,
+        expr: &ExprListParenthesized,
+    ) -> Result<Type, TransformationError> {
+        let parsed_exprs = expr
+            .expressions(&self.db)
+            .elements(&self.db)
+            .into_iter()
+            .map(|expr| self.transform_expr(&expr))
+            .collect::<Result<Vec<_>, TransformationError>>()?;
+
+        Ok(Type::Tuple(Tuple(parsed_exprs)))
+    }
+
+    /// Transforms a [`ExprPath`] to a [`Type`].
+    fn transform_path(&mut self, expr: &ExprPath) -> Result<Type, TransformationError> {
+        match split(expr, &self.db)? {
+            SplitResult::Simple { splits } => self.transform_simple_path(&splits),
+            SplitResult::WithGenericArgs {
+                splits,
+                generic_args,
+            } => self.transform_generic_path(&splits, &generic_args),
+        }
+    }
+
+    /// Transforms a generic path to a [`Type::Sequence`].
+    ///
+    /// It first checks if the path is a known sequence type (Array or Span).
+    /// If it is, it reads the length of the sequence from the buffer.
+    /// Then it recursively transforms the elements of the sequence.
+    fn transform_generic_path(
+        &mut self,
+        splits: &[String],
+        generic_args: &str,
+    ) -> Result<Type, TransformationError> {
+        let sequence_type = match splits.join("::").as_str() {
+            "core::array::Array" => SequenceType::Array,
+            "core::array::Span" => SequenceType::Span,
+            _ => return Err(TransformationError::UnsupportedType),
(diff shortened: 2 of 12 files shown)


crates/data-transformer/src/shared/extraction.rs
```diff
@@ -0,0 +1,47 @@
...
+                // because it's searched for using Felt entrypoint selector, identically as functions.
+                // Also, we don't need any constructor-specific properties, just argument types.
+                map.insert(
+                    get_selector_from_name(constructor.name.as_str()).unwrap(),
+                    AbiFunction {
+                        name: constructor.name.clone(),
+                        inputs: constructor.inputs.clone(),
+                        outputs: vec![],
+                        state_mutability: StateMutability::View,
+                    },
```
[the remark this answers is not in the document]
ksew1: It doesn't have output field: https://github.com/xJonathanLEI/starknet-rs/blob/c1e636bfab2d46a2ce484c8313547950a27058e6/starknet-core/src/types/contract/mod.rs#L264

crates/data-transformer/src/reverse_transformer/mod.rs
```diff
@@ -0,0 +1,69 @@
...
+    abi: &[AbiEntry],
+    function_selector: &Felt,
+) -> Result<String, ReverseTransformError> {
+    let type_extractor = |function: AbiFunction| {
+        function
+            .inputs
+            .into_iter()
+            .map(|input| input.r#type)
+            .collect::<Vec<_>>()
+    };
```
cptartur, line 29: Why not make it a normal function instead?

crates/data-transformer/src/reverse_transformer/mod.rs
```diff
@@ -0,0 +1,69 @@
...
+#[derive(Debug, thiserror::Error)]
+pub enum ReverseTransformError {
+    #[error(r#"Function with selector "{0}" not found in ABI of the contract"#)]
+    FunctionNotFound(Felt),
+    #[error(transparent)]
+    TransformationError(#[from] TransformationError),
+}
+
+/// Transforms a calldata into a Cairo-like string representation of the arguments
+pub fn reverse_transform_input(
```
cptartur, line 18: I assume this is just needed for backtrace? Since sncast will only need to transform function outputs.

crates/data-transformer/src/reverse_transformer/mod.rs
```diff
@@ -0,0 +1,69 @@
...
+            .collect::<Vec<_>>()
+    };
+    reverse_transform(output, abi, function_selector, type_extractor)
+}
+
+fn reverse_transform(
+    felts: &[Felt],
+    abi: &[AbiEntry],
+    function_selector: &Felt,
+    type_extractor: fn(AbiFunction) -> Vec<String>,
```
cptartur, line 53: I get the generalization of this with type_extractor, but maybe a less cleaver approach would be clearer to follow: Have this function accept argument types to convert directly and handle the argument extraction on the higher level?

crates/data-transformer/src/reverse_transformer/transform.rs
```diff
@@ -0,0 +1,245 @@
...
+    pub fn parse_and_transform(&mut self, expr: &str) -> Result<Type, TransformationError> {
+        self.transform_expr(&parse_expression(expr, &self.db)?)
+    }
+
+    /// Transforms the given [`Expr`] to a [`Type`].
+    fn transform_expr(&mut self, expr: &Expr) -> Result<Type, TransformationError> {
+        match expr {
+            Expr::Tuple(expr) => self.transform_tuple(expr),
+            Expr::Path(expr) => self.transform_path(expr),
+            _ => Err(TransformationError::UnsupportedType),
```
cptartur, line 62: Why do we know it will only be tuple or a `Path`? Does things like structs etc. fall under `Path`?

crates/data-transformer/src/reverse_transformer/transform.rs
```diff
@@ -0,0 +1,245 @@
...
+        splits: &[String],
+        generic_args: &str,
+    ) -> Result<Type, TransformationError> {
+        let sequence_type = match splits.join("::").as_str() {
+            "core::array::Array" => SequenceType::Array,
+            "core::array::Span" => SequenceType::Span,
+            _ => return Err(TransformationError::UnsupportedType),
+        };
+
+        let length = self.buffer_reader.read::<usize>()?;
```
cptartur, line 108: We bubble these errors, but I wonder if that makes sense. Failing to read length from these means malformed ABI. Shouldn't it panic or fail with this kind of error instead?

crates/data-transformer/src/reverse_transformer/transform.rs
```diff
@@ -0,0 +1,245 @@
...
+    /// Then it tries to find the type in the ABI and transform it to the matching representation.
+    fn transform_simple_path(&mut self, parts: &[String]) -> Result<Type, TransformationError> {
+        let name = parts.last().expect("path should not be empty").to_owned();
+
+        match self.transform_primitive_type(&name) {
+            Err(PrimitiveError::NotFound(_)) => (),
+            Ok(primitive) => return Ok(Type::Primitive(primitive)),
+            Err(PrimitiveError::BufferReaderError(error)) => {
+                return Err(TransformationError::BufferReaderError(error));
+            }
```
cptartur, line 133: Nit, but do we have to wrap these again? Can't we support conversions with `.into()` somehow?

crates/data-transformer/src/reverse_transformer/transform.rs
```diff
@@ -0,0 +1,245 @@
...
+            "i32" => Ok(Primitive::I32(self.buffer_reader.read()?)),
+            "i64" => Ok(Primitive::I64(self.buffer_reader.read()?)),
+            "i128" => Ok(Primitive::I128(self.buffer_reader.read()?)),
+            "ByteArray" => Ok(Primitive::ByteArray(self.buffer_reader.read()?)),
+            "bytes31" => Ok(Primitive::CairoBytes31(self.buffer_reader.read()?)),
+            "ContractAddress" => Ok(Primitive::ContractAddress(self.buffer_reader.read()?)),
+            "ClassHash" => Ok(Primitive::ClassHash(self.buffer_reader.read()?)),
+            "StorageAddress" => Ok(Primitive::StorageAddress(self.buffer_reader.read()?)),
+            "EthAddress" => Ok(Primitive::EthAddress(self.buffer_reader.read()?)),
+            "felt" | "felt252" => Ok(Primitive::Felt(self.buffer_reader.read()?)),
```
cptartur, line 225: I guess the only "problem" is that the reverse data transformer doesn't support shortstrings, which would be need to do in some cases.

But we would have to rely on heuristic that checks if the felt can be converted to a shortstring and displaying it along with the raw value.

What do you think about that?

crates/data-transformer/src/reverse_transformer/types.rs
```diff
@@ -0,0 +1,172 @@
...
+#[derive(Debug)]
+pub struct Tuple(pub Vec<Type>);
+
+impl Display for Tuple {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let mut dbg = f.debug_tuple("");
+        for item in &self.0 {
+            dbg.field(&format_args!("{item}"));
+        }
+        dbg.finish()
```
cptartur, line 97: Do we want to use debug formatting in display?

crates/data-transformer/src/shared/extraction.rs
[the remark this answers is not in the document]
cptartur: What's the point of even supporting it if it cannot return anything?

crates/data-transformer/src/shared/path.rs
```diff
@@ -0,0 +1,73 @@
...
+    },
+    WithGenericArgs {
+        splits: Vec<String>,
+        generic_args: String,
+    },
+}
+pub fn split(
+    path: &ExprPath,
+    db: &SimpleParserDatabase,
+) -> Result<SplitResult, GenericArgsExtractionError> {
```
cptartur, line 27: Maybe it would be better to define this as a trait and implement it for `&ExprPath`? This way you could use `.split(...)`.

crates/data-transformer/src/shared/path.rs
```diff
@@ -0,0 +1,73 @@
...
+                let generic_args = extract_generic_args(&segment, db)?;
+                return Ok(SplitResult::WithGenericArgs {
+                    splits,
+                    generic_args,
+                });
+            }
+        }
+    }
+
+    Ok(SplitResult::Simple { splits })
```
cptartur, line 45: Why not have this return inside the match as well?

crates/data-transformer/src/shared/extraction.rs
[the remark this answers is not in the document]
ksew1: because we are interested of it's inputs

crates/data-transformer/src/reverse_transformer/transform.rs
ksew1: > Does things like structs etc. fall under Path?
Yes.  You can look at this: https://github.com/foundry-rs/starknet-foundry/pull/3263/files

crates/data-transformer/src/reverse_transformer/transform.rs
ksew1: Well, if we return this as an `Err`, we give the consumer of this reverse transformer a chance to handle it, e.g., by falling back to raw felts. If we panic instead, the consumer doesn't have that choice, so I would prefer returning it as an error.

crates/data-transformer/src/reverse_transformer/types.rs
ksew1: Why wouldn't we?

crates/data-transformer/src/shared/path.rs
ksew1: It it is much more lines of code for such a subtle difference. I don't think it is worth it

crates/data-transformer/src/reverse_transformer/transform.rs
ksew1: there is already implemented `From` via `Error` trait. I just prefer the more verbose version

crates/data-transformer/src/shared/path.rs
ksew1: Because we want to have all the splits. In the case of arrays and spans, we know that the generic segment is at the end of the path, so by the time we reach the generic, we have collected all the splits

crates/data-transformer/src/reverse_transformer/transform.rs
cptartur: Okay, for me it's a bit too verbose but no strong opinion.

crates/data-transformer/src/reverse_transformer/transform.rs
cptartur: Okay, let's wait if user start requesting it then

crates/data-transformer/src/shared/path.rs
cptartur: Okay, if it requires more code than it's probably fine to keep

crates/data-transformer/src/shared/path.rs
cptartur: > In the case of arrays and spans, we know that the generic segment is at the end of the path, so by the time we reach the generic, we have collected all the splits

This might be worth having in the comment in code

crates/data-transformer/src/reverse_transformer/transform.rs
ksew1: That's what `cairo_lang_runner::short_string::as_cairo_short_string` does, but let's say you try `123` — this becomes `{` as a short string. Did you mean that? Probably not.
There are many such cases where it will produce garbage short strings and clutter the output.

crates/data-transformer/src/shared/path.rs
```diff
@@ -0,0 +1,85 @@
...
+            PathSegment::Simple(segment) => {
+                splits.push(segment.ident(db).token(db).text(db).to_string());
+            }
+            PathSegment::WithGenericArgs(segment) => {
+                splits.push(segment.ident(db).token(db).text(db).to_string());
+                let generic_args = extract_generic_args(&segment, db)?;
+                return Ok(SplitResult::WithGenericArgs {
+                    splits,
+                    generic_args,
+                });
```
cptartur, line 52: Maybe there should be a safeguard assert here that checks if the iterator has been exhausted before we return? In case of malformed abi.

Or is it verified in some other step?

crates/data-transformer/src/shared/extraction.rs
```diff
@@ -0,0 +1,47 @@
...
+use std::collections::HashMap;
+
+pub fn extract_function_from_selector(
+    abi: &[AbiEntry],
+    function_selector: &Felt,
+) -> Option<AbiFunction> {
+    map_selectors_to_functions(abi).remove(function_selector)
+}
+
+fn map_selectors_to_functions(abi: &[AbiEntry]) -> HashMap<Felt, AbiFunction> {
```
ddoktorski, line 13: nit: If this is used only by the function `extract_function_from_selector` wouldn't it be better to modify it so that it stops traversing the ABI once the needed function has been found?

crates/data-transformer/src/reverse_transformer/transform.rs
```diff
@@ -0,0 +1,245 @@
...
+            }
+            AbiStructOrEnum::Struct(struct_type_definition) => {
+                self.transform_struct(struct_type_definition)
+            }
+        }
+    }
+
+    /// Transforms to a [`Type::Struct`].
+    ///
+    /// It first it recursively transforms the fields of the struct and then creates a [`Type::Struct`] instance.
```
ddoktorski, line 148:
```suggestion
    /// Recursively transforms the fields of the struct and then creates a [`Type::Struct`] instance.
```

crates/data-transformer/src/reverse_transformer/types.rs
```diff
@@ -0,0 +1,172 @@
...
+            Primitive::U128(value) => write!(f, "{value}_u128"),
+            Primitive::U256(value) => write!(f, "{value}_u256"),
+            Primitive::U384(value) => write!(f, "{value}_u384"),
+            Primitive::U512(value) => write!(f, "{value}_u512"),
+            Primitive::I8(value) => write!(f, "{value}_i8"),
+            Primitive::I16(value) => write!(f, "{value}_i16"),
+            Primitive::I32(value) => write!(f, "{value}_i32"),
+            Primitive::I64(value) => write!(f, "{value}_i64"),
+            Primitive::I128(value) => write!(f, "{value}_i128"),
+            Primitive::Felt(value) => write!(f, "{value}_felt252"),
```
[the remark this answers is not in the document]
ksew1: https://github.com/foundry-rs/starknet-foundry/pull/3262#discussion_r2066757691

We will go with hex and this suffix will be removed

crates/data-transformer/src/reverse_transformer/types.rs
```diff
@@ -0,0 +1,172 @@
...
+        }
+    }
+}
+
+#[derive(Debug)]
+pub struct Tuple(pub Vec<Type>);
+
+impl Display for Tuple {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let mut dbg = f.debug_tuple("");
```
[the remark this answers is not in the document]
ksew1: I'm not sure if I understand the question correctly. But here, we use Rust debug formatting, and Rust is known for its great stability and for not introducing breaking changes except in new editions.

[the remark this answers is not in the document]
cptartur: For some reason I though `debug_tuple` is from cairo crates not from rust directly. Disregard.

cptartur (approved):

ddoktorski (approved):

PR #3262, opened 2025-04-22
Add reverse transformer to sncast call

ksew1:
Related #3009

commit-id:583ae7d0

---

**Stack**:
- #3289
- #3288
- #3287
- #3286
- #3285
- #3264
- #3263
- #3262 ⬅


crates/sncast/README.md
```diff
@@ -178,7 +178,8 @@ $ sncast \
 
 ```shell
 command: call
-response: [0x6]
+response: 6_felt252
```
cptartur, line 181: Are felt always represented as dec strings now? Personally I would be expecting them to be hexes, at least in case where they are long

ksew1: Yes, felts are always dec. ContractAddress, ClassHash, StorageAddress, EthAddress, and CairoBytes31 are in hex. We can go for hex for felts but it would also remove the `_felt252` as it seems weird with hex

ddoktorski: I would also prefer to print felts just as hexes

cptartur (approved):
I'd prefer different felt formatting as commented, but otherwise looks good

ddoktorski (approved):

kkawula (approved):

PR #3268, opened 2025-04-23
Adjust backtrace tests

ddoktorski:
commit-id:f0069e59

---

**Stack**:
- #3274
- #3273
- #3269
- #3268 ⬅

### DIFF
--- a/crates/forge/tests/e2e/build_trace_data.rs
+++ b/crates/forge/tests/e2e/build_trace_data.rs
@@ -178,12 +178,13 @@ fn trace_is_produced_even_if_contract_panics() {
     let temp = setup_package("backtrace_panic");
     test_runner(&temp)
         .arg("--save-trace-data")
+        .arg("--ignored")
         .assert()
         .success();
 
     let trace_data = fs::read_to_string(
         temp.join(TRACE_DIR)
-            .join("backtrace_panic_Test_test_contract_panics.json"),
+            .join("backtrace_panic_Test_test_contract_panics_with_should_panic.json"),
     )
     .unwrap();
 
--- a/docs/src/snforge-advanced-features/backtrace.md
+++ b/docs/src/snforge-advanced-features/backtrace.md
@@ -36,7 +36,7 @@ If your contract fails and a backtrace can be generated, `snforge` will prompt y
 this:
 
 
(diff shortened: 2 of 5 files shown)


docs/src/snforge-advanced-features/backtrace.md
```diff
@@ -36,7 +36,7 @@ If your contract fails and a backtrace can be generated, `snforge` will prompt y
 this:
 
 
-<!-- { "package_name": "backtrace_vm_error" } -->
+<!-- { "ignored": true, "package_name": "backtrace_vm_error" } -->
```
ksew1, line 39: Shouldn't we just change the output?

ddoktorski: As for an example in the docs this one seems a bit better than the actual VM error. From the reader's perspective this is only an output example without showing the real package used, so it does not matter much

ksew1 (approved):

docs/src/snforge-advanced-features/backtrace.md
cptartur: Maybe it would be better to show the real error anyway? Adding ignores increases the risk of these snippets becoming totally outdated

cptartur: Please address this comment, otherwise it's good

ddoktorski: Changed the example output to use `backtrace_panic` package instead of `backtrace_vm_error`. It required also some other adjustments

cptartur (approved):

cptartur:
@ksew1 please take a look again

crates/forge/tests/e2e/build_trace_data.rs
```diff
@@ -178,12 +178,13 @@ fn trace_is_produced_even_if_contract_panics() {
     let temp = setup_package("backtrace_panic");
     test_runner(&temp)
         .arg("--save-trace-data")
+        .arg("--ignored")
```
ksew1, line 181: Wouldn't it be easier to run contract by name instead of running ignored tests?

franciszekjob (approved):

ksew1 (approved):

PR #3275, opened 2025-04-24
Add docs for `set_balance`

franciszekjob:
Towards #1979 

**Stack**:
- #3316
- #3276
- #3275 ⬅
- #3271
- #3256

## Introduced changes

Add docs and example for `set_balance` and `Token`

## Checklist

- [x] Linked relevant issue
- [x] Updated relevant documentation
- [x] Added relevant tests
- [x] Performed self-review of the code
- [x] Added changes to `CHANGELOG.md`

PR #3271, opened 2025-04-23
Add Cairo implementation of `set_balance`

franciszekjob:
Towards #1979

**Stack**:
- #3316
- #3276
- #3275
- #3271 ⬅
- #3256

## Introduced changes

This PR adds `set_balance` cheatcode to `snforge_std`

## Checklist

- [x] Linked relevant issue
- [x] Updated relevant documentation
- [x] Added relevant tests
- [x] Performed self-review of the code
- [x] Added changes to `CHANGELOG.md`

PR #3276, opened 2025-04-24
Allow disabling STRK predeployment

### LINKED ISSUE #1979: Add a `deal` cheatcode for getting erc20 token

snforge

### Description

It can be based on foundry, but the interface or name may vary if we want to change it.

We should consider some flexibility when it comes to address of the popular tokens (different addresses for different ETH/STRK tokens on different forked networks - we can provide some constants, or helper functions).

It should be implemented via a wrapper over `store` + `load`, and be working with standard tokens like `STRK`/`ETH` by default (taking into account the implementations they use)

franciszekjob:
Closes #1979

**Stack**:
- #3316
- #3276 ⬅
- #3275
- #3271
- #3256

## Introduced changes

Introduce `#[disable_predeployed_contracts]` attribute

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


cptartur (approved):

crates/snforge-scarb-plugin/src/lib.rs
```diff
@@ -62,3 +63,8 @@ fn available_gas(args: TokenStream, item: TokenStream) -> ProcMacroResult {
 fn should_panic(args: TokenStream, item: TokenStream) -> ProcMacroResult {
     should_panic(args, item)
 }
+
+#[attribute_macro]
+fn disable_strk_predeployment(args: TokenStream, item: TokenStream) -> ProcMacroResult {
```
ddoktorski, line 68: Thinking long term we should make it more generic, something like `disable_predeployed_contracts`.

franciszekjob: Thinking even further (considering we would add more tokens) that attribute would take an argument and could be used multiple times, e.g.

`#disable_predeployed_contract("STRK")`

or 

`#disable_predeployed_token("STRK")`

wdyt @ddoktorski @cptartur @kkawula ?

crates/forge/tests/integration/set_balance.rs
```diff
@@ -245,3 +247,47 @@ fn test_check_gas_in_empty_test_case() {
...
+                )
+                    .unwrap_syscall();
+                balance
+            }
+
+            #[test]
+            #[disable_strk_predeployment]
+            fn test_set_balance_strk_with_disabled_predeployment() {
+                let contract_address: ContractAddress = 0x123.try_into().unwrap();
+                get_balance(contract_address, Token::STRK);
```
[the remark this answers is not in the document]
franciszekjob: > What are we checking here?

It's a case when someone wants to interact with STRK but the predeployment is disabled

> Maybe it is worth adding a symentric test without `#[disable_strk_predeployment]` and show the difference directly?

I don't understand? In this case we indeed use the attribute

crates/snforge-scarb-plugin/src/lib.rs
franciszekjob: as discussed offline, let's use `#[disable_predeployed_contracts]`

crates/cheatnet/src/runtime_extensions/forge_config_extension.rs
```diff
@@ -30,6 +30,9 @@ impl<'a> ExtensionLogic for ForgeConfigExtension<'a> {
             "set_config_fork" => config_cheatcode!(fork),
             "set_config_available_gas" => config_cheatcode!(available_gas),
             "set_config_ignore" => config_cheatcode!(ignore),
+            "set_config_contracts_predeploy" => {
+                config_cheatcode!(disable_predeployed_contracts)
+            }
```
ddoktorski, line 35:
```suggestion
            "set_config_disable_contracts" => config_cheatcode!(disable_predeployed_contracts),
```

crates/forge/tests/integration/set_balance.rs
```diff
@@ -206,3 +206,47 @@ fn test_set_balance_strk_with_fork() {
...
+
+#[test]
+fn test_set_balance_strk_with_disabled_predeployment() {
+    let test = test_case!(
+        indoc!(
+            r#"
+            use snforge_std::TokenTrait;
+            use starknet::{ContractAddress, syscalls, SyscallResultTrait};
+
+            use snforge_std::{Token};
```
ddoktorski, line 218: nit:
```suggestion
            use snforge_std::{Token, TokenTrait};
            use starknet::{ContractAddress, syscalls, SyscallResultTrait};
```

crates/cheatnet/src/runtime_extensions/forge_config_extension/config.rs
```diff
@@ -167,6 +167,13 @@ pub struct RawIgnoreConfig {
     pub is_ignored: bool,
 }
 
+// disable strk predeployment
+
+#[derive(Debug, Clone, CairoDeserialize)]
+pub struct RawContractsPredeploymentConfig {
```
ddoktorski, line 173:
```suggestion
pub struct RawPredeployedContractsConfig {
```

crates/snforge-scarb-plugin/src/attributes/disable_predeployed_contracts.rs
```diff
@@ -0,0 +1,38 @@
...
+
+pub struct ContractsPredeploymentCollector;
+
+impl AttributeInfo for ContractsPredeploymentCollector {
+    const ATTR_NAME: &'static str = "disable_predeployed_contracts";
+}
+
+impl AttributeTypeData for ContractsPredeploymentCollector {
+    // Suffix is `..._predeploy` instead of `..._predeployment` because it's too big for felt
+    const CHEATCODE_NAME: &'static str = "set_config_contracts_predeploy";
```
ddoktorski, line 17:
```suggestion
    const CHEATCODE_NAME: &'static str = "set_config_disable_contracts";
```

ddoktorski (approved):

PR #3277, opened 2025-04-24
Feature/2927 cheatspan nonzero enforcement

### LINKED ISSUE #2927: Add check preventing setting `CheatSpan` to `0`
### Current State

Currently, there is no logic preventing the user from setting the `CheatSpan` to `0` directly from Cairo code. This would cause an underflow in `CheatStatus::decrement_cheat_span` code which, instead of cancelling the cheat, would make the cheat effective basically indefinitely.

### Objective

Add a check that prevents `CheatSpan` from being set to `0`

### Additional Context

We could also just handle the case where `CheatSpan` is `0` but it doesn't make any sense to set it to this value.

Uh-X3L:
**Summary**
========
This PR introduces validation to prevent setting `CheatSpan::TargetCalls(0)`, addressing a potential underflow bug that could cause cheats to persist indefinitely. The update enforces correct usage of cheat spans across all cheatcode entry points.

Details  
=====
CheatSpan Smart Constructor:
- Added a smart constructor `CheatSpan::target_calls(n)` in state.rs that returns an error if `n == 0`, preventing invalid construction at the source.
Example:
```rust
// Before
let span = CheatSpan::TargetCalls(n);

// After
let span = CheatSpan::target_calls(n)?;
```

Runtime Validation in Cheatcode Entry Points:
- Added runtime checks in all cheatcode entry points (`cheat_block_number`, `cheat_block_timestamp`, `cheat_caller_address`, `cheat_sequencer_address`, `mock_call`) to panic if `CheatSpan::TargetCalls(0)` is passed.
Example:
```rust
if let CheatSpan::TargetCalls(n) = span {
    if n == 0 {
        panic!("CheatSpan::TargetCalls(0) is not allowed");
    }
}
```

**Test Coverage:**
- Searched the codebase to ensure all production usages of `CheatSpan::TargetCalls(n)` are protected.
- Left direct construction in test code for testing and demonstration purposes.

**Impact:**
- Prevents accidental creation of cheats that never expire due to underflow.
- Enforces correct cheat span semantics and improves test reliability.
- No changes required in decrement logic or existing test helpers.

### DIFF
--- a/crates/cheatnet/src/state.rs
+++ b/crates/cheatnet/src/state.rs
@@ -41,6 +41,17 @@ pub enum CheatSpan {
     TargetCalls(usize),
 }
 
+impl CheatSpan {
+    /// Smart constructor for TargetCalls variant. Prevents zero value.
+    pub fn target_calls(n: usize) -> Result<Self, String> {
+        if n == 0 {
+            Err("CheatSpan::TargetCalls(0) is not allowed".to_string())
+        } else {
+            Ok(CheatSpan::TargetCalls(n))
+        }
+    }
+}
+
 #[derive(Debug)]
 pub struct ExtendedStateReader {
     pub dict_state_reader: DictStateReader,
--- a/crates/cheatnet/src/state.rs
+++ b/crates/cheatnet/src/state.rs
@@ -13,15 +13,19 @@ use blockifier::execution::syscalls::hint_processor::SyscallUsageMap;
 use blockifier::state::errors::StateError::UndeclaredClassHash;
 use blockifier::state::state_api::{StateReader, StateResult};
 use cairo_annotations::trace_data::L1Resources;
+use cairo_serde_macros::CairoDeserialize;
 use cairo_vm::Felt252;
 use cairo_vm::vm::runners::cairo_runner::ExecutionResources;
 use cairo_vm::vm::trace::trace_entry::RelocatedTraceEntry;
-use conversions::serde::deserialize::CairoDeserialize;
+use conversions::serde::deserialize::{
+    BufferReadError, BufferReadResult, BufferReader, CairoDeserialize as CairoDeserializeTrait,
+};
 use conversions::serde::serialize::{BufferWriter, CairoSerialize};
 use conversions::string::TryFromHexStr;
 use runtime::starknet::constants::TEST_CONTRACT_CLASS_HASH;
 use runtime::starknet::context::SerializableBlockInfo;
 use runtime::starknet::state::DictStateReader;
+use serde::{Deserialize, Serialize};
 use starknet_api::block::BlockInfo;
 use starknet_api::core::{ChainId, EntryPointSelector};
 use starknet_api::transaction::fields::ContractAddressSalt;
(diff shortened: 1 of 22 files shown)


crates/cheatnet/src/state.rs
```diff
@@ -41,6 +41,17 @@ pub enum CheatSpan {
...
+    /// Smart constructor for TargetCalls variant. Prevents zero value.
+    pub fn target_calls(n: usize) -> Result<Self, String> {
+        if n == 0 {
+            Err("CheatSpan::TargetCalls(0) is not allowed".to_string())
+        } else {
+            Ok(CheatSpan::TargetCalls(n))
+        }
+    }
+}
+
```
cptartur, line 54: Instead of that, we should change the definition of `CheatSpan` so that `NonZeroUsize` is used instead of `usize`.

This way we represent it not being zero in the type system.

cptartur: Logic for decrementing should be adjusted accordingly

cptartur: With that, the enum definition in `snforge_std` should be adjusted accordingly as well.

Cairo doesn't have `NonZeroUsize` definition but we can define one ourselves. Something along the lines of this

```cairo
#[derive(Copy, Drop, PartialEq, Debug)]
pub struct NonZeroUsize {
    value: usize
}

impl NonZeroUsizeSerdeImpl of Serde<NonZeroUsize> {
    fn serialize(self: @NonZeroUsize, ref output: Array<felt252>) {
        self.value.serialize(ref output)
    }

    fn deserialize(ref serialized: Span<felt252>) -> Option<NonZeroUsize> {
        let value = Serde::<usize>::deserialize(ref serialized)?;
        Option::Some(NonZeroUsize { value })
    }
}

#[generate_trait]
pub impl NonZeroUsizeImpl of NonZeroUsizeTrait {
    fn new(value: usize) -> NonZeroUsize {
        assert(value > 0, 'NonZeroUsize cannot be 0');
        NonZeroUsize { value }
    }
}
```

cptartur (changes requested):
Please add relevant tests as well and update the docs and changelog

Uh-X3L:
@cptartur  Hey Arthur, that was a little bit more than I expected to chew on this issue, but here it is, let me know your thoughts, had to change many things, and hopefully I did this in the safest way I could find.

crates/forge/tests/integration/cheat_sequencer_address.rs
```diff
@@ -246,7 +246,7 @@ fn cheat_sequencer_address_with_span() {
 
                 let target_sequencer_address: ContractAddress = 123.try_into().unwrap();
 
-                cheat_sequencer_address(dispatcher.contract_address, target_sequencer_address, CheatSpan::TargetCalls(1));
+                cheat_sequencer_address(dispatcher.contract_address, target_sequencer_address, CheatSpan::target_calls(1));
```
cptartur, line 249: This will not compile/work. There is no method `CheatSpan::target_calls` in the `cheatcodes.cairo`.

cptartur (changes requested):

cptartur:
The code doesn't work and too me looks like it is likely AI generated.

PR #3281, opened 2025-04-25
Always display return data if `should_panic` test passes

ddoktorski:
**Stack**:
- #3282
- #3281 ⬅

### DIFF
--- a/crates/forge-runner/src/test_case_summary.rs
+++ b/crates/forge-runner/src/test_case_summary.rs
@@ -418,7 +418,9 @@ fn extract_result_data(run_result: &RunResult, expectation: &ExpectedTestResult)
             };
 
             match expected_data {
-                Some(expected) if is_matching(panic_data, expected) => None,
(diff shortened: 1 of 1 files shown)


ksew1 (approved):

cptartur (commented):
Can you explain this change? If the tests that is marked with `#[should_panic(expected: '...')` passes, do we show the panic backtrace to the user? 

If yes, I'm not sure if this is a good behavior

ddoktorski:
> Can you explain this change?

When using `should_panic` attribute, behavior varies based on the `expected` argument. Without it `Success data:` message is printed, otherwise it is not. As a result of this inconsistency passing `SNFORGE_BACKTRACE=1` won't print a backtrace in the latter case. I suggest allowing backtrace printing for any panic, regardless of test success or failure. If you find the `Success data` message redundant, we can potentially fix it in a different way.

ddoktorski:
Closing pull request: commit has gone away

PR #3282, opened 2025-04-25
Omit displaying backtrace for handled errors

ddoktorski:


crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/entry_point.rs
```diff
@@ -479,3 +487,54 @@ pub(crate) fn extract_trace_and_register_errors(
         trace: Some(trace),
     }
 }
+
+// This helper function is used for backtrace to avoid displaying errors that were already handled
+// It checks if nested calls failed with the same error as the root call
+// If a nested call does not fail with the same error, the function clears the error state for that call
+// Note: This may not be accurate if a panic was initially handled and then the function panicked
+// again with the identical panic data
+fn clear_handled_errors(
```
ksew1, line 496: Currently, from what I see, it is DFS implemented using a stack. What do you think about using recursion? I find it easier to follow. Also, two things:

- You terminate the search from a node if its return data matches the root's return data, but that doesn't mean the children of the current node have the same data, so we still need to traverse them.
- We could improve this heuristic, because if a node's return data doesn't match the root's return data, then all its children should be considered clear without further checks.

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/entry_point.rs
```diff
@@ -479,3 +487,54 @@ pub(crate) fn extract_trace_and_register_errors(
...
+    root_call: &CallInfo,
+    root_class_hash: &ClassHash,
+    cheatnet_state: &mut CheatnetState,
+) {
+    let root_retdata = root_call.execution.retdata.0.clone();
+
+    let mut queue: Vec<CallInfo> = vec![];
+    queue.append(
```
ksew1, line 504: Nit: why not extend?

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/entry_point.rs
```diff
@@ -479,3 +487,54 @@ pub(crate) fn extract_trace_and_register_errors(
...
+        .iter()
+        .filter(|call| {
+            call.execution.failed
+                && !root_retdata.starts_with(&call.execution.retdata.0)
+                && call
+                    .call
+                    .class_hash
+                    .is_some_and(|class_hash| class_hash != *root_class_hash)
+        })
+        .cloned()
```
ksew1, line 538: Nit: we don't need to take ownership
```suggestion
```

ddoktorski: Actually we need to take ownership to be able to call `collect`, but please let me know if you see a better approach to this

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/entry_point.rs
ddoktorski: I completely modified the implementation. Now we first search for contracts that have a matching error and then we clear `encountered_errors` for all contracts that are not in our collected list. I think this approach is easier to follow. It's still an iterative implementation and I believe it's quite readable, but let me know if you strongly prefer it to be rewritten using recursion.

crates/cheatnet/src/state.rs
```diff
@@ -473,6 +473,10 @@ impl CheatnetState {
     pub fn register_error(&mut self, class_hash: ClassHash, pcs: Vec<usize>) {
         self.encountered_errors.insert(class_hash, pcs);
     }
+
+    pub fn clear_error(&mut self, class_hash: ClassHash) {
```
ksew1, line 477:
```suggestion
    pub fn clear_error(&mut self, class_hash: &ClassHash) {
```

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/entry_point.rs
ddoktorski: In the above example the set `contracts_matching_root_error` will contain only one contract: A. Therefore, we will clear the entries for both contracts B and C.

ksew1: Yeah my mistake i didn't saw it now is the inverse of the condition

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/entry_point.rs
ksew1: Can't we just do:

```rs
fn get_inner_calls_with_matching_panic_data<'a>(
    call_info: &'a CallInfo,
    root_retdata: &[Felt],
) -> Vec<&'a CallInfo> {
    call_info
        .inner_calls
        .iter()
        .filter(|call| call.execution.failed && root_retdata.starts_with(&call.execution.retdata.0))
        .collect()
}
```

ksew1 (approved):

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/entry_point.rs
ddoktorski: We can't because `collect` requires owned values

ksew1: I it not a big deal but this compile on my computer just fine

ddoktorski: Thanks that works, updated

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/entry_point.rs
```diff
@@ -479,3 +487,60 @@ pub(crate) fn extract_trace_and_register_errors(
...
+    root_class_hash: &ClassHash,
+) -> HashSet<ClassHash> {
+    let mut contracts_matching_root_error = HashSet::new();
+    contracts_matching_root_error.insert(*root_class_hash);
+
+    let mut failed_matching_calls: Vec<&CallInfo> = vec![root_call];
+
+    while let Some(call_info) = failed_matching_calls.pop() {
+        if let Some(class_hash) = call_info.call.class_hash {
+            contracts_matching_root_error.insert(class_hash);
```
cptartur, line 526: Doesn't this casue `root_class_hash` to be inserted into `contracts_matching_root_error` twice? I know it's a `HashSet` so it's not really a problem, but do we need an initial insert to a `contracts_matching_root_error`? 

Can `root_call.call.class_hash` be different than `root_class_hash`?

crates/forge/tests/e2e/backtrace.rs
```diff
@@ -249,6 +249,35 @@ fn test_backtrace_panic_without_inlines() {
...
+fn test_handled_error_not_display() {
+    let temp = setup_package("dispatchers");
+
+    let output = test_runner(&temp)
+        .arg("test_handle_and_panic")
+        .env("SNFORGE_BACKTRACE", "1")
+        .assert()
+        .success();
+
+    assert_stdout_contains(
```
cptartur, line 264: Maybe a comment telling what error we don't expect to appear in the output would be handy?

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/entry_point.rs
ddoktorski: > Doesn't this casue root_class_hash to be inserted into contracts_matching_root_error twice?

Yes it does. Let's omit the initial insert, it was added purely to be safe and explicit.

> Can root_call.call.class_hash be different than root_class_hash?

It should be the same as this is the same class hash used to create `ExecutableCallEntryPoint` in the `execute_call_entry_point` function that is later used in the `initialize_execution_context_with_runner_mode` to create `SyscallHandlerBase`.

crates/forge/tests/e2e/backtrace.rs
ddoktorski: Added a comment and an actual assert to check that no specific error appears in the output. So far it only checked what the output contained, which isn't a thorough test for this change.

cptartur (approved):

Issue #3195, opened 2025-04-07
CI speed optimization


cptartur:
### Current State

Currently our CI is quite slow and runs multiple jobs.

### Objective

Optimize CI performance

### Additional Context

Specific improvements to be researched then implemented.

We should also have some kind of measured baseline for these tasks performance we can reference.

Issue #3188, opened 2025-04-04
Pin actions by hash, not by tab


ksew1:
### Which component is your feature related to?

Other (Please describe below)

### Feature Request

Originated from: https://github.com/foundry-rs/starknet-foundry/pull/3172#discussion_r2026374442

PR #3283, opened 2025-04-26
Run forge integration tests on CI with cargo nextest

ddoktorski:
Related #3195 

## Introduced changes

- The job `test-forge-unit-and-integration` is currently a bottleneck in the CI process, significantly delaying execution. This change partially fixes the problem by running integration tests with `cargo nextest`
- To avoid increasing the number of necessary workers, the number of partitions for E2E tests has been decreased from 8 to 3 as these jobs are still executed relatively quickly (around 10 minutes each)
- The entire workflow for both Windows and Ubuntu has improved from around [50 minutes](https://github.com/foundry-rs/starknet-foundry/actions/runs/14679406363) to [35 minutes](https://github.com/ddoktorski/starknet-foundry/actions/runs/14679518349)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,7 +41,7 @@ jobs:
       - uses: software-mansion/setup-universal-sierra-compiler@v1
       - run: cargo test --release --lib -p forge
 
-  build-test-forge-e2e-nextest-archive:
+  build-test-forge-nextest-archive:
     needs: setup
     runs-on: ${{ matrix.os }}
     strategy:
@@ -65,7 +65,7 @@ jobs:
   test-forge-integration:
     name: Test Forge / Integration Tests
     runs-on: ${{ matrix.os }}
-    needs: [ build-test-forge-e2e-nextest-archive, setup ]
+    needs: [ build-test-forge-nextest-archive, setup ]
     strategy:
       fail-fast: false
       matrix:
@@ -87,7 +87,7 @@ jobs:
   test-forge-e2e:
     name: Test Forge / E2E Tests
     runs-on: ${{ matrix.os }}
-    needs: [ build-test-forge-e2e-nextest-archive, setup ]
+    needs: [ build-test-forge-nextest-archive, setup ]
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -63,14 +62,36 @@ jobs:
           name: nextest-archive-${{ matrix.os }}
           path: nextest-archive-${{ matrix.os }}.tar.zst
 
+  test-forge-integration:
+    name: Test Forge / Integration Tests
+    runs-on: ${{ matrix.os }}
+    needs: [ build-test-forge-e2e-nextest-archive, setup ]
```
ksew1, line 68: Maybe we can rename this as this also builds integration tests

ksew1 (approved):
Thanks!

.github/workflows/ci.yml
```diff
@@ -63,14 +62,36 @@ jobs:
...
+    strategy:
+      fail-fast: false
+      matrix:
+        partition: [ 1, 2, 3 ]
+        os: ${{ fromJson(needs.setup.outputs.matrix_os) }}
+    steps:
```
ksew1, line 74: [nit] It is also a good practice to pin actions by commit hash rather than by tag, as commit hashes are immutable and reduce the risk of supply chain attacks. I did it in several action created by me: https://github.com/foundry-rs/starknet-foundry/blob/05c42a26ae9420f81f0febd52f16c2019644a7b6/.github/workflows/ci.yml#L190

ddoktorski: Generally agree, but I am not sure if that matters that much for the actions that we maintain. If you think it is worth doing let's just update it in all places in a separate PR.

ksew1: Could you just add issue, so we don't forget about this one?

ksew1 (approved):

.github/workflows/ci.yml
ddoktorski: Actually it already exists https://github.com/foundry-rs/starknet-foundry/issues/3188

kkawula (approved):
Could you please update the required checks in repo settings? After these changes, they are no longer correct.

cptartur (approved):

PR #3287, opened 2025-04-28
Use reverse transformer in debugging for call result

ksew1:
Related #3009

commit-id:c19312f9

---

**Stack**:
- #3319
- #3318
- #3289
- #3288
- #3287 ⬅

### DIFF
--- a/crates/debugging/Cargo.toml
+++ b/crates/debugging/Cargo.toml
@@ -12,4 +12,4 @@ starknet-types-core.workspace = true
 starknet.workspace = true
 cheatnet = { path = "../cheatnet" }
 data-transformer = { path = "../data-transformer" }
-serde_json .workspace = true
+serde_json.workspace = true
--- a/crates/debugging/src/trace/collect.rs
+++ b/crates/debugging/src/trace/collect.rs
@@ -1,13 +1,17 @@
 use crate::Trace;
 use crate::trace::types::{
     CallerAddress, ContractName, ContractTrace, Selector, StorageAddress, TestName, TraceInfo,
-    TransformedCalldata,
+    TransformedCallResult, TransformedCalldata,
+};
+use cheatnet::runtime_extensions::call_to_blockifier_runtime_extension::rpc::{
+    CallFailure, CallResult as CheatnetCallResult,
 };
 use cheatnet::runtime_extensions::forge_runtime_extension::contracts_data::ContractsData;
 use cheatnet::state::{CallTrace, CallTraceNode};
-use data_transformer::reverse_transform_input;
+use data_transformer::{reverse_transform_input, reverse_transform_output};
 use serde_json::Value;
 use starknet::core::types::contract::AbiEntry;
+use starknet_api::execution_utils::format_panic_data;
 use std::cell::RefCell;
 use std::rc::Rc;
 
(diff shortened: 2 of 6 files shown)


crates/debugging/Cargo.toml
```diff
@@ -12,4 +12,5 @@ starknet-types-core.workspace = true
 starknet.workspace = true
 cheatnet = { path = "../cheatnet" }
 data-transformer = { path = "../data-transformer" }
+cairo-lang-runner.workspace = true
```
kkawula, line 15: Aren't we currently removing this dependency? xD cc @cptartur

cptartur: We are. @ksew1 stack to remove cairo-lang-runner will have a replacement for this, can we have some kind of placeholder/todo! for now here or can we wait with merging this PR?

cptartur (changes requested):

crates/debugging/Cargo.toml
ksew1: Used `starknet_api::execution_utils::format_panic_data` instead

crates/debugging/src/trace/collect.rs
```diff
@@ -105,3 +109,30 @@ fn collect_transformed_calldata(call_trace: &CallTrace, abi: &[AbiEntry]) -> Tra
...
+            CallFailure::Error { msg } => format_result_message("error", &msg.to_string()),
+        },
+    })
+}
+
+fn format_result_message(tag: &str, message: &str) -> String {
+    message
+        .is_empty()
+        .then(|| tag.to_string())
+        .unwrap_or_else(|| format!("{tag}: {message}"))
```
cptartur, line 137: Wouldn't this be clearer?

```suggestion
    if message.is_empty() {
        tag.to_string()
    } else {
        format!("{tag}: {message}")
    }
```

crates/debugging/src/trace/types.rs
```diff
@@ -33,6 +32,9 @@ pub struct TraceInfo {
     pub call_result: CallResult,
 }
 
+#[derive(Debug, Clone)]
+pub struct CallResult(pub String);
```
cptartur, line 36: Why are we reimplementing this struct?

crates/debugging/src/trace/collect.rs
ksew1: I would say that "clean code" is more a matter of preference here. I really like the functional style, so I went with the current solution. If you prefer the imperative style, we can go with that instead — no biggie for me.

crates/debugging/src/trace/types.rs
ksew1: Because it is not the old `CallResult` but already transformed `CallResult`. I should probably named this `TransformedCallResult` to avoid confusion

crates/forge/tests/e2e/debugging.rs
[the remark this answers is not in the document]
ksew1: Added

crates/debugging/src/trace/collect.rs
cptartur: I like functional style, but your solution seems bit too smart for what it's doing.

cptartur (approved):

franciszekjob (approved):

Issue #3164, opened 2025-03-31
Add option to turn on debugging print in forge


ksew1:
### Which component is your feature related to?

Forge

### Feature Request

This should be done via `--verbosity` flag

PR #3288, opened 2025-04-28
Add verbosity to debugging

ksew1:
Related #3164

---

**Stack**:
- #3319
- #3318
- #3289
- #3288 ⬅

### DIFF
--- a/crates/debugging/src/trace/collect.rs
+++ b/crates/debugging/src/trace/collect.rs
@@ -1,8 +1,8 @@
-use crate::Trace;
 use crate::trace::types::{
     CallerAddress, ContractName, ContractTrace, Selector, StorageAddress, TestName, TraceInfo,
     TransformedCallResult, TransformedCalldata,
 };
+use crate::{Trace, Verbosity};
 use cheatnet::runtime_extensions::call_to_blockifier_runtime_extension::rpc::{
     CallFailure, CallResult as CheatnetCallResult,
 };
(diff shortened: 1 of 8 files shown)


crates/debugging/src/verbosity.rs
```diff
@@ -0,0 +1,57 @@
...
+    pub fn full<T>(self, supply_fn: impl FnOnce() -> T) -> Full<T> {
+        Full::new(self, supply_fn)
+    }
+}
+
+/// Macro to define a verbosity-gated container.
+///
+/// Assumes that the struct name provided (e.g., `Medium`) exactly matches
+/// a variant in the `Verbosity` enum (e.g., `Verbosity::Medium`).
+macro_rules! impl_verbosity_container {
```
ksew1, line 28: Dopaminka 🥰

crates/debugging/src/verbosity.rs
```diff
@@ -0,0 +1,57 @@
+/// Verbosity for displaying trace.
```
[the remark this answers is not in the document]
ksew1: I change this. What do you think now?

franciszekjob (approved):

crates/debugging/src/verbosity.rs
```diff
@@ -0,0 +1,57 @@
+/// Verbosity that can be used to create verbosity-gated containers.
+#[derive(Debug, Clone, Copy, Ord, PartialOrd, Eq, PartialEq)]
+pub enum Verbosity {
+    /// Lowest verbosity level. Can't create any containers as it should always be computed.
+    Minimal,
+    /// Medium verbosity level. Can create [`Medium`] containers.
+    Medium,
+    /// Full verbosity level. Can create [`Full`] containers.
+    Full,
+}
```
franciszekjob, line 10: I wonder if such names would't be more accurate: basic - detailed - full. Wdyt?

ksew1: Let's see what others will say

cptartur: Maybe minimal - normal/standard - detailed.

But tbh I'm fine with all options

crates/forge-runner/src/test_case_summary.rs
```diff
@@ -323,8 +323,14 @@ impl TestCaseSummary<Single> {
...
-        let debugging_trace = cfg!(feature = "debugging")
-            .then(|| debugging::Trace::new(&call_trace.borrow(), contracts_data, name.clone()));
+        let debugging_trace = cfg!(feature = "debugging").then(|| {
+            debugging::Trace::new(
+                &call_trace.borrow(),
+                contracts_data,
+                debugging::Verbosity::Full,
+                name.clone(),
+            )
+        });
```
cptartur, line 333: Do we want to still keep the debugging behind a feature flag if we are going to enable it by default?

crates/debugging/src/tree/building/node.rs
```diff
@@ -39,4 +39,11 @@ impl<'a> Node<'a> {
     pub fn leaf(&mut self, tree_item: &impl NodeDisplay) {
         self.builder.add_empty_child(tree_item.display());
     }
+
+    /// Creates a leaf node which parent is the current node if the item is not `None`.
+    pub fn leaf_optional(&mut self, tree_item: Option<&impl NodeDisplay>) {
```
cptartur, line 44: That's a bit weird control flow to me. Wouldn't it make more sense to still have `.leaf` method and conditionally create it at caller site?

crates/debugging/src/trace/collect.rs
```diff
@@ -12,122 +12,154 @@ use data_transformer::{reverse_transform_input, reverse_transform_output};
...
 
-fn collect_abi(contract_name: &ContractName, contracts_data: &ContractsData) -> Vec<AbiEntry> {
-    let artifacts = contracts_data
-        .get_artifacts(&contract_name.0)
-        .expect("artifact should be present in `ContractsData`");
+        let trace_info = TraceInfo {
+            contract_name,
+            entry_point_type: self
+                .verbosity
+                .full(|| self.call_trace.entry_point.entry_point_type),
```
cptartur, line 63: Would it be possible to create some kind of general `Into<Full>` and `Into<Medium>` so we could just call `.into()` here? And would that be too complex?

My only nitpick is that verbosity levels of these are defined in the struct itself, so having a simpler way of creating it would maybe be a bit clearer.

crates/forge-runner/src/test_case_summary.rs
ksew1: This is changed in next PR

crates/debugging/src/tree/building/node.rs
ksew1: This method is used six times, and as the codebase grows, it will likely be used even more. If we handled this at the caller site, a single line would automatically become three. I don't see a significant benefit to that. Also, since this isn't exposed outside of this crate, I there aren't any external callers.

crates/debugging/src/trace/collect.rs
ksew1: You need two things: verbosity and a function to compute this lazily. So it would need to be implemented via From on a tuple, as `From` and `Into` only take a single argument. I haven’t found any other clean way of doing this.

crates/debugging/src/verbosity.rs
ksew1: went with minimal - standard - detailed

cptartur (approved):

crates/debugging/src/trace/collect.rs
cptartur: Okay, if we can't have a clean `From` or `Into` for this, then disregard.

PR #3289, opened 2025-04-28
Use debugging in forge without a feature flag

ksew1:
Related #3164

commit-id:fb008c1d

---

**Stack**:
- #3319
- #3318
- #3289 ⬅

### DIFF
--- a/crates/forge-runner/src/debugging.rs
+++ b/crates/forge-runner/src/debugging.rs
@@ -1,0 +1,39 @@
+use cheatnet::runtime_extensions::forge_runtime_extension::contracts_data::ContractsData;
+use cheatnet::state::CallTrace;
+use clap::ValueEnum;
+
+/// Trace verbosity level.
+#[derive(ValueEnum, Debug, Clone, Copy)]
+pub enum TraceVerbosity {
+    /// Display test name, contract name and selector.
+    Minimal,
+    /// Display test name, contract name, selector, calldata and call result.
+    Standard,
+    /// Display everything.
+    Detailed,
+}
+
+impl From<TraceVerbosity> for debugging::Verbosity {
+    fn from(verbosity: TraceVerbosity) -> Self {
+        match verbosity {
+            TraceVerbosity::Minimal => debugging::Verbosity::Minimal,
+            TraceVerbosity::Standard => debugging::Verbosity::Standard,
+            TraceVerbosity::Detailed => debugging::Verbosity::Detailed,
+        }
+    }
+}
+
+#[must_use]
+pub fn build_debugging_trace(
+    call_trace: &CallTrace,
+    contracts_data: &ContractsData,
+    trace_verbosity: Option<TraceVerbosity>,
+    test_name: String,
+) -> Option<debugging::Trace> {
+    Some(debugging::Trace::new(
+        call_trace,
+        contracts_data,
+        trace_verbosity?.into(),
+        test_name,
+    ))
+}
--- a/crates/forge-runner/src/lib.rs
+++ b/crates/forge-runner/src/lib.rs
@@ -1,4 +1,5 @@
 use crate::coverage_api::run_coverage;
+use crate::debugging::TraceVerbosity;
 use crate::forge_config::{ExecutionDataToSave, ForgeConfig, TestRunnerConfig};
 use crate::running::{run_fuzz_test, run_test};
 use crate::test_case_summary::TestCaseSummary;
@@ -33,6 +34,7 @@ pub mod test_case_summary;
 pub mod test_target_summary;
 
 pub mod backtrace;
+pub mod debugging;
 mod gas;
 pub mod printing;
 pub mod running;
@@ -103,6 +105,7 @@ pub fn run_for_test_case(
     forge_config: Arc<ForgeConfig>,
     versioned_program_path: Arc<Utf8PathBuf>,
     send: Sender<()>,
+    trace_verbosity: Option<TraceVerbosity>,
 ) -> JoinHandle<Result<AnyTestCaseSummary>> {
     if case.config.fuzzer_config.is_none() {
         tokio::task::spawn(async move {
@@ -112,6 +115,7 @@ pub fn run_for_test_case(
                 forge_config.test_runner_config.clone(),
                 versioned_program_path,
                 send,
+                trace_verbosity,
             )
             .await?;
             Ok(AnyTestCaseSummary::Single(res))
@@ -124,6 +128,7 @@ pub fn run_for_test_case(
                 forge_config.test_runner_config.clone(),
                 versioned_program_path,
                 send,
+                trace_verbosity,
             )
             .await??;
             Ok(AnyTestCaseSummary::Fuzzing(res))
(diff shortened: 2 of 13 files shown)


crates/forge/src/lib.rs
```diff
@@ -140,6 +141,11 @@ enum ColorOption {
 pub struct TestArgs {
     /// Name used to filter tests
     test_filter: Option<String>,
+
+    /// Trace verbosity level
+    #[arg(short = 'v', long, default_value = "none")]
+    trace_verbosity: TraceVerbosity,
+
```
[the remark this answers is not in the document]
ksew1: Mainly because how help looks: 
 ```
  -v, --trace-verbosity <TRACE_VERBOSITY>
          Trace verbosity level
          
          [default: none]

          Possible values:
          - none:    Don't display trace
          - minimal: Display test name, contract name and selector
          - medium:  Display test name, contract name, selector, calldata and call result
          - full:    Display everything
          
 ```
 
 ```
   -v, --trace-verbosity <TRACE_VERBOSITY>
          Trace verbosity level

          Possible values:
          - minimal: Display test name, contract name and selector
          - medium:  Display test name, contract name, selector, calldata and call result
          - full:    Display everything
```

I prefer the fist one, but it is matter of taste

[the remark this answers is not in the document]
Comment (franciszekjob) [reply crates/forge/src/lib.rs:148]: What's the benefit of having none option? It seems it's just redundant.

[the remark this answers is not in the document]
Comment (ksew1) [reply crates/forge/src/lib.rs]: The longer i thought about it the more it clear became to me that it should be option. Changed

File: crates/forge-runner/src/debugging.rs [hunk 2]
```diff
@@ -0,0 +1,39 @@
...
+impl From<TraceVerbosity> for debugging::Verbosity {
+    fn from(verbosity: TraceVerbosity) -> Self {
+        match verbosity {
+            TraceVerbosity::Minimal => debugging::Verbosity::Minimal,
+            TraceVerbosity::Medium => debugging::Verbosity::Medium,
+            TraceVerbosity::Full => debugging::Verbosity::Full,
+        }
+    }
+}
+
```
cptartur, line 25: We do we need to define a separate enum? To have a stable interface?

crates/forge-runner/src/lib.rs
```diff
@@ -103,6 +105,7 @@ pub fn run_for_test_case(
     forge_config: Arc<ForgeConfig>,
     versioned_program_path: Arc<Utf8PathBuf>,
     send: Sender<()>,
+    trace_verbosity: Option<TraceVerbosity>,
```
cptartur, line 108: Another argument that we needs to flow down the stream of these functions 🥰

We need to build some abstraction for that soon

crates/forge/Cargo.toml
```diff
@@ -10,7 +10,7 @@ smoke = []
 scarb_2_9_1 = []
 scarb_since_2_10 = []
 no_scarb_installed = []
-debugging = ["forge_runner/debugging"]
+debugging = []
```
cptartur, line 13: Do we need still this feature at all?

crates/forge/src/lib.rs
```diff
@@ -140,6 +141,11 @@ enum ColorOption {
 pub struct TestArgs {
     /// Name used to filter tests
     test_filter: Option<String>,
+
+    /// Trace verbosity level
+    #[arg(long)]
+    trace_verbosity: Option<TraceVerbosity>,
```
cptartur, line 147: Docs will be another PR right?

crates/forge/tests/e2e/debugging.rs
```diff
@@ -3,14 +3,46 @@ use indoc::formatdoc;
...
+    let output = test_runner(&temp)
+        .arg("--trace-verbosity")
+        .arg("full")
+        .assert()
+        .code(1);
 
-    assert_stdout_contains(
-        output,
-        formatdoc! {r"
+    assert_stdout_contains(output, test_output(full_debugging_trace_message));
```
cptartur, line 15: Do we reuse these functions anywhere? 
If not, I think it would actually be clearer to just inline these

crates/forge-runner/src/debugging.rs
franciszekjob: I think @ksew1 simply wanted to add contextual clarity here. It may seem redundant at first glance, but imo 
overall it makes sense.

I just wonder if we really need it. Are we planning to introduce other verbosity type? Maybe it's worth keeping it simple?

crates/forge/src/lib.rs
franciszekjob: Yes, that's what @ksew1 said

crates/forge-runner/src/debugging.rs
ksew1: Both of you are right. It was:

- stable interface  
- contextual clarity  
- no dependency on clap  
- separation from forge, as there are plans to use this crate in the future in the profiler  

To be honest, I would prefer to keep two separate enums unless you strongly want to avoid duplication.

crates/forge-runner/src/lib.rs
ksew1: An easy refactor would be to create a `RunContext` struct that holds all these arguments which simply flow downstream and are not modified. Nevertheless, this is out of scope for this PR, and I'm not sure if refactors are legal right know as the situation is changing like in a kaleidoscope

crates/forge/Cargo.toml
ksew1: Only to have separate jobs in ci for this tests. It is more quality of life improvement rather than necessity.

crates/forge/tests/e2e/debugging.rs
ksew1: `test_output` is used in multiple places. I think at this level, "clean code" is more a matter of preference. Since it is my code, this is my preference. However, I don't have a strong opinion on this one.

crates/forge-runner/src/debugging.rs
franciszekjob: It's fine, there's an actual reason to have it separated 👍

crates/forge-runner/src/lib.rs
cptartur: Right, it's I wouldn't do it in this stack, but whoever touches these args next is doing a refactor 😅

crates/forge/tests/e2e/debugging.rs
cptartur: Okay, test output can stay since it's contents aren't relevant for these tests. But I'd inline the trace messages if they aren't reused in multiple tests. I think it's easier to follow tests these way

crates/forge-runner/src/debugging.rs
cptartur: Good for me as well

cptartur (approved):
Looks good, but please address my comment about inlines

crates/forge/tests/e2e/debugging.rs
ksew1: This is not easy either, as each of the trace messages is used twice—once for a test that passes and once for a test that fails. This is done to check that they don't differ. Because of that, they are higher-order functions.

franciszekjob (approved):

PR #3309, opened 2025-04-30
Remove success data message for passing tests

### LINKED ISSUE #3304: Change backtrace behavior on should_panic
* Allow to display backtrace when test panicked and the variable is set
* Do not print a note about using `SNFORGE_BACKTRACE=1` when test passes
* Do not print `Success data` when `#[should_panic]` is used and test passes

ddoktorski:
Closes #3304 

commit-id:e6286a8e

---

**Stack**:
- #3282
- #3309 ⬅

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Forge
+
+#### Changed
+
+- "Success data" message is no longer printed when a test using the `#[should_panic]` attribute passes
+
 ## [0.42.0] - 2025-04-28
 
 ### Forge
@@ -34,11 +40,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 #### Fixed
 
 - Bug that prevented from passing values to `--arguments` that started with a leading minus `-` sign.
-
-### Cast
-
-### Fixed
-
 - User is now prompted to save an imported or deployed account in `sncast` config even when using `--network` flag
 
 ## [0.41.0] - 2025-04-08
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -30,11 +36,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 #### Fixed
 
 - Bug that prevented from passing values to `--arguments` that started with a leading minus `-` sign.
-
-### Cast
-
-### Fixed
-
```
ddoktorski, line 37: Note: Not related to this PR, fix of duplicated entries

crates/forge-runner/src/test_case_summary.rs
```diff
@@ -367,9 +367,9 @@ impl TestCaseSummary<Single> {
                     if matching {
                         TestCaseSummary::Passed {
                             name,
-                            msg: msg.map(|msg| {
-                                add_backtrace_footer(msg, contracts_data, &encountered_errors)
-                            }),
+                            msg: is_backtrace_enabled()
+                                .then(|| get_backtrace(contracts_data, &encountered_errors).ok())
+                                .flatten(),
```
ksew1, line 372: Don't we want to display error if something went  wrong?

ddoktorski: In this case test passes with the expected panic. We can only display the error that is expected (or its superstring, because for bytearrays we only check if the error contains the expected value). What do you mean by something went wrong here?

ksew1: If backtrace fails to be created, you won't be knowing why. Because we call `get_backtrace(contracts_data, &encountered_errors).ok()`

ddoktorski: I see, it can be changed so `get_backtrace` simply returns `String` which is either backtrace or something like `Failed to create backtrace: {err}`

ksew1: I would go with that

cptartur (approved):

franciszekjob (approved):

ksew1 (approved):

PR #3313, opened 2025-05-05
Explicitly print full path of files edited by `sncast account`

### LINKED ISSUE #3227: Explicitly print full path of files edited by `sncast account import`

Cast

### Feature Request

Currently, the result of `sncast account import` results in the following:
```
command: account import
add_profile: Profile argent_2_sepolia successfully added to snfoundry.toml
```

It is not clear without reading the docs, which files have been edited, and where is the information added. 

Would expect something like:
```
Account information added to: /home/amanusk/.starknet_accounts/starknet_open_zeppelin_accounts.json

Profile information added to: /home/amanusk/Code/Cairo/starknet-foundry-template/snfoundry.toml
```

kkawula:
Closes #3227

## Introduced changes

-

## Checklist

- [x] Linked relevant issue
- [ ] Added relevant tests

### DIFF
--- a/crates/configuration/src/lib.rs
+++ b/crates/configuration/src/lib.rs
@@ -1,9 +1,9 @@
 use anyhow::{Context, Result, anyhow};
+use camino::Utf8PathBuf;
 use scarb_metadata::{Metadata, PackageId};
 use serde_json::{Map, Number};
+use std::fs::File;
 use std::{env, fs};
-
-use camino::Utf8PathBuf;
 use tempfile::{TempDir, tempdir};
 use toml::Value;
 pub const CONFIG_FILENAME: &str = "snfoundry.toml";
@@ -51,6 +51,17 @@ pub fn get_profile(
         None if profile_name == "default" => Ok(serde_json::Value::Object(Map::default())),
         None => Err(anyhow!("Profile [{}] not found in config", profile_name)),
     }
+}
+
+#[must_use]
+pub fn resolve_config_file() -> Utf8PathBuf {
+    find_config_file().unwrap_or_else(|_| {
+        let path = Utf8PathBuf::from(CONFIG_FILENAME);
+        File::create(&path).expect("creating file in current directory should be possible");
+
+        path.canonicalize_utf8()
+            .expect("path canonicalize in current directory should be possible")
+    })
 }
 
 pub fn load_config<T: Config + Default>(
--- a/crates/sncast/src/starknet_commands/account/create.rs
+++ b/crates/sncast/src/starknet_commands/account/create.rs
@@ -27,6 +27,7 @@ use starknet::providers::jsonrpc::HttpTransport;
 use starknet::signers::{LocalWallet, SigningKey};
 use starknet_types_core::felt::Felt;
 use std::str::FromStr;
+use configuration::resolve_config_file;
 
 #[derive(Args, Debug)]
 #[command(about = "Create an account with all important secrets")]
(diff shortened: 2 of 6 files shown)


kkawula:
image

crates/sncast/tests/e2e/account/create.rs
```diff
@@ -185,12 +185,11 @@ pub async fn test_happy_case_add_profile() {
...
     ];
 
-    let snapbox = runner(&args).current_dir(tempdir.path());
-    let output = snapbox.assert().success();
+    let output = runner(&args).current_dir(tempdir.path()).assert();
 
     assert_stdout_contains(
         output,
-        "add_profile: Profile my_account successfully added to snfoundry.toml",
+        "add_profile: Profile my_account successfully added to [..]",
```
cptartur, line 192: Can't we use the tempdir path joined with something to assert proper paths in these tests?

crates/sncast/src/starknet_commands/account/create.rs
```diff
@@ -127,33 +128,34 @@ pub async fn create(
...
-        }
+    let mut add_profile_message =
+        String::from("--add-profile flag was not set. No profile added to snfoundry.toml");
+
+    if create.add_profile.is_some() {
+        let url = create
+            .rpc
+            .url
+            .clone()
+            .expect("Conflicting arguments should be handled in clap");
```
cptartur, line 139: Nit, this should never even appear, but maybe we can use a more specific message?

crates/sncast/src/starknet_commands/account/create.rs
```diff
@@ -127,33 +128,34 @@ pub async fn create(
...
+        let config = CastConfig {
+            url,
+            account: account.into(),
+            accounts_file: accounts_file.into(),
+            keystore,
+            ..Default::default()
+        };
+        let config_path = resolve_config_file();
+        let profile_name = add_profile.expect("add_profile should be Some");
+        add_profile_message =
```
cptartur, line 150: Maybe it would be safer to structure the code like

```rust
let add_profile_message = if create.add_profile.is_some() {
    // ...
    "message"
} else {
    String::from("--add-profile flag was not set. No profile added to snfoundry.toml");
};

File: crates/configuration/src/lib.rs [hunk 4]
```diff
@@ -53,6 +53,17 @@ pub fn get_profile(
...
+pub fn resolve_config_file() -> Utf8PathBuf {
+    find_config_file().unwrap_or_else(|_| {
+        let path = Utf8PathBuf::from(CONFIG_FILENAME);
+        File::create(&path).expect("creating file in current directory should be possible");
+
+        path.canonicalize_utf8()
+            .expect("path canonicalize in current directory should be possible")
+    })
+}
+
```
cptartur, line 66: While we are at it, can we change this PR to a stack, that first does some more refactors to the config? 

Your logic is better already but what if instead of having `resolve_config_file` that returns a path, could have a structure representing the config location? Then it could have methods for creating and a finding a default path. 

If you think this cannot be reasonably refactored anymore without reworking half of cast let me know though.

crates/sncast/src/starknet_commands/account/create.rs
kkawula: Actually, it is: https://github.com/foundry-rs/starknet-foundry/pull/3313/files#diff-ffea8fc06fa950e50c6ed65400b8f57b8a4c7766d53f5fe42854c75d8caa97e5R134-R135

crates/configuration/src/lib.rs
kkawula: I have split it into a stack, but I don't find it better with a structure representing the config location, it should be part of the bigger refactor related to configs.

PR #3240, opened 2025-04-18
Remove `cairo-lang-runner` dependency from forge runner

cptartur:
**Stack**:
- #3333
- #3328
- #3327
- #3247
- #3246
- #3245
- #3244
- #3243
- #3242
- #3241
- #3240 ⬅
- #3238

⚠️ *Part of a stack created by [spr](https://github.com/ejoffe/spr). Do not merge manually using the UI - doing so may have unexpected results.*

PR #3327, opened 2025-05-12
Update CHANGELOG.md

cptartur:
**Stack**:
- #3333
- #3328
- #3327 ⬅
- #3247
- #3246
- #3245
- #3244
- #3243
- #3242
- #3241
- #3240
- #3238


CHANGELOG.md
```diff
@@ -7,6 +7,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Forge
+
+#### Changed
+
+- Forge is incompatible with non-latest versions of `snforge_std` dependency. Changed the code generated by `snforge_std`'s `#[test]` attribute
```
ddoktorski, line 14: I previously suggested adding a point about this to the changelog, but maybe instead we should include a note or warning in the release notes? I guess the important thing is that this version is not compatible with old plugins, the fact that the `#[test]` attribute generates different code internally is probably not relevant here?

cptartur: I think we should probably do both. Keep a changelog for completeness and have a warning as well.

ddoktorski: I'm not sure, at least to me the current phrasing seems more confusing than informative

cptartur: Simplified the message and we can add a separate warning once we release

ddoktorski (approved):

CHANGELOG.md
```diff
@@ -7,6 +7,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Forge
+
+#### Changed
+
+- Changed the code generated by `snforge_std`'s `#[test]` attribute
```
franciszekjob, line 14: Imo it doesn't tell much to a regular user, could we rephrase it?

CHANGELOG.md
```diff
@@ -7,6 +7,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+
+#### Fixed
+
+- "invalid syscall selector" error appearing when using arithmetic circuits
```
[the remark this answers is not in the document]
cptartur: I haven't investigated it in more depth, simply adding it because the issue was supposedly resolved. @kkawula do we have more details to provide here?

Sidenote, don't think more details are 100% needed here

CHANGELOG.md
cptartur: I think the only important part from the user standpoint is that it changed and that they need to update `snforge_std` (which we will have a separate warning for). Generated code itself is an implementation detail.

CHANGELOG.md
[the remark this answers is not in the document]
franciszekjob: Ok, let's leave it then.

franciszekjob (approved):

cptartur:

Issue #3114, opened 2025-03-18
Add tracked resources tracking to cheatnet


cptartur:
### Current State

Currently, we do not push to `tracked_resource_stack` except for the initial push. This results in inconsistent behavior from real network.

### Objective

Correctly push tracked resources to the `tracked_resource_stack`  

### Additional Context

Relevant logic from blockifier should be reused/copied accordingly. Transactions should correctly fallback to vm resources from sierra resources when required.

Additionally, tests for transaction where mixed resource types are used should be added.

PR #3334, opened 2025-05-12
Remove usage of `SierraVersion::LATEST` constant

ddoktorski:
Towards #3114

## Introduced changes

- Replace all occurrences of `SierraVersion::LATEST` with explicit Sierra version values

### DIFF
--- a/crates/cheatnet/src/constants.rs
+++ b/crates/cheatnet/src/constants.rs
@@ -4,6 +4,7 @@ use std::sync::Arc;
 
 use blockifier::execution::entry_point::{CallType, ExecutableCallEntryPoint};
 use cairo_lang_starknet_classes::casm_contract_class::CasmContractClass;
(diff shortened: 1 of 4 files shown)


franciszekjob (approved):

crates/cheatnet/src/constants.rs
```diff
@@ -46,15 +47,15 @@ fn contract_class_no_entrypoints() -> ContractClass {
     let casm_contract_class: CasmContractClass = serde_json::from_str(raw_contract_class)
         .expect("`raw_contract_class` should be valid casm contract class");
 
-    ContractClass::V1((casm_contract_class, SierraVersion::LATEST))
+    ContractClass::V1((casm_contract_class, SierraVersion::default()))
```
cptartur, line 50: What version is the default? Should we use that?

cptartur (approved):

PR #3336, opened 2025-05-13
Fallback to VM resources tracking for older Sierra versions

### LINKED ISSUE #3114: Add tracked resources tracking to cheatnet

ddoktorski:
Closes #3114

## Introduced changes

- Update tracked resources stack in the context execution based on the executed contract
- After executing test, update top call resources for both sierra gas and VM resources

Note: This change is relevant only for tests that use contracts from a fork network compiled with Sierra `<1.7.0`

## Checklist

- [X] Linked relevant issue
- [X] Added relevant tests


crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
```diff
@@ -613,20 +613,12 @@ pub fn update_top_call_resources(runtime: &mut ForgeRuntime, tracked_resource: &
...
-        }
-        TrackedResource::SierraGas => {
-            let all_sierra_gas_consumed = add_sierra_gas_resources(&top_call);
-            let mut top_call = top_call.borrow_mut();
-            top_call.gas_consumed = all_sierra_gas_consumed;
-            top_call
-        }
-    };
+    let all_execution_resources = add_execution_resources(top_call.clone());
+    let all_sierra_gas_consumed = add_sierra_gas_resources(&top_call);
```
THenry14, line 617: what happens if we track sierra, but then out of the sudden a wild cairo-step appears at the end of the stack? I don't see a logic for conversion so I assume both sierra gas and steps will be accounted for? not sure it's the best way, I think resources should be converted accordingly (or maybe I'm missing/misunderstanding somehting?)

THenry14 (commented):
generally, I love it. great job 💪

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
ddoktorski: This logic only accumulates sierra gas and VM resources. For example if we track sierra gas, but there are no calls that track cairo steps, then the resources for cairo steps will effectively sum to zero. As a result, they won't affect the final results.

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/entry_point.rs
```diff
@@ -178,12 +178,16 @@ pub fn execute_call_entry_point(
             context,
         ),
     };
+    context
+        .tracked_resource_stack
+        .pop()
+        .expect("Unexpected empty tracked resource.");
```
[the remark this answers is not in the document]
THenry14: yup, this is where I'm coming from - technically it shouldn't be required to do pop() in here, you could just use current_tracked_resource (given that stack is not modified in execute). Let's check this, and if it turns out stack can change, you can just resolve this :+1: otherwise I would just remove this unnecessary pop, but up to you really.

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
THenry14: can we somehow make sure that is the case? maybe we could prepare a test for this, or something? I am just worried we miss something (now or in the future), and we end up with mixed resources in the output. To be clear, I am mostly worried about mixed scenario where there are cairo steps and sierra gas accounted for at different stages of the stack

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/entry_point.rs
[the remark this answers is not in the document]
ddoktorski: This `pop()` cannot be removed as it’s important for keeping the last element in the stack up to date. It is used in the function `get_current_tracked_resource` and also in some other places. Imagine a scenario in which an inner call uses cairo steps. If we don’t pop this element from the stack before returning to the outer call, we may continue tracking cairo steps even when we should be tracking sierra gas instead.

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
ddoktorski: > can we somehow make sure that is the case?

Actually I think quite reliable confirmation is that there was no need to adjust any gas values in the existing tests

> I am mostly worried about mixed scenario where there are cairo steps and sierra gas accounted for at different stages of the stack

I think it would make sense to expand the output of the `--detailed-resources` flag in the case of mixed resources. I would add this in a separate PR along with some tests to verify that both sierra gas and VM resources are displayed correctly. Does that work for you or would you prefer to include some specific tests within the scope of this PR?

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/entry_point.rs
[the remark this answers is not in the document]
THenry14: but shouldn't we though? if this is the scenario, all the next calls **should** track steps (ie if steps appeared at any moment in the stack, we must track them from now on as far as I remember); and also at the end we must represent resources as vm resources - all of them, even the first calls that tracked sierra gas, right?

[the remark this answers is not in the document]
ddoktorski: > all the next calls should track steps 

This is true, but I am referring to a situation that an inner call can affect independent outer call, which should not happen.

> and also at the end we must represent resources as vm resources - all of them, even the first calls that tracked sierra gas, right?

Why would that be the case?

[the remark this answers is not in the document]
ddoktorski: Actually perfect example is the test `test_tracked_resources` added in this PR. If you remove the pop logic, you'll see that the second call (`call_internal` in the contract test) tracks Cairo steps instead of Sierra gas. However both calls - the external call to the map contract and the internal call are completely independent. This happens because without removing the top element from the stack before executing the internal call, the top element remains `CairoSteps`. As a result `get_current_tracked_resource` function returns `Cairo steps`, because the logic is: if the last tracked resource was Cairo steps, continue tracking it. But for `call_internal` only what the parent call tracked should be relevant, the map contract doesn't matter.

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
ddoktorski: > I am still not sure how --tracked-resource flag should work now if we allow mixed resources. It might be very confusing if as a user I wanted to track sierra gas, but get cairo steps in the trace file, detailed resources or any other output

I agree that it might be confusing, but converting to a different resource seems hacky to me. If the call info contains a given resource because it had to track it, then I would prefer to keep the actual information there. Does it make any difference for the profiler though?

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/entry_point.rs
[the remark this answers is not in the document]
THenry14: Ah ok so you mean an external call, like a call to some external contract, cool thanks that is the thing that was vague to me. Then, I think, we arrived to the same place as in the other thread, so perhaps we can merge the conversations together. tldr; I think we should either allow users to override using the tracked-resource flag or (imo better) just treat the stack as a source of truth and return whatever resource results from it. 

Please also make sure that an external call does not change tracked resource for all stack - honestly I am not sure about this, and don't think this was described explicitly anywhere 🙏

crates/cheatnet/tests/starknet/execution.rs
```diff
@@ -46,3 +49,48 @@ fn test_state_reverted() {
...
+        TrackedResource::CairoSteps
+    );
+    assert_eq!(first_inner_call.execution.gas_consumed, 0);
+    assert_ne!(first_inner_call.resources.n_steps, 0);
+
+    let second_inner_call = main_call_info.inner_calls.last().unwrap();
+    assert_eq!(
+        second_inner_call.tracked_resource,
+        TrackedResource::SierraGas
+    );
```
cptartur, line 93: Technically to validate this logic entirely, we should also have a next level of nesting where call that used `CairoSteps` calls a method that could use `SierraGas` but we check it also used `CairoSteps`. 

Or just a separate tests similar to this where the order is reverted and we check everything used `CairoSteps`

See this https://community.starknet.io/t/starknet-v0-13-4-pre-release-notes/115257#p-2358763-sierra-gas-vs-vm-resources-what-do-we-count-10

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
cptartur: We could just treat `--tracked-resource` flag as a "hint" to `snfoundry` i.e. it tries to use the resource provided, but if for some reason we had to fallback from sierra to vm resources, then we explicitly show the user that it happened.

I assume the other case where it changes from vm resources to sierra is not possible.

But maybe I'm misunderstanding the problem?

crates/cheatnet/tests/starknet/execution.rs
ddoktorski: > we should also have a next level of nesting where call that used CairoSteps calls a method that could use SierraGas but we check it also used CairoSteps.

Good point, I updated the test to have an additional nested call to assert that it uses Cairo steps as well

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
ddoktorski: > We could just treat --tracked-resource flag as a "hint" to snfoundry i.e. it tries to use the resource provided, but if for some reason we had to fallback from sierra to vm resources, then we explicitly show the user that it happened.

There is a clear and specific reason why execution must switch to VM resources and our behavior is consistent with the network. Tbh I don't see any reason to modify the flag as there is still a use case to track cairo steps for contracts with sierra version greater than `1.7.0`.

> I assume the other case where it changes from vm resources to sierra is not possible.

That's correct.

THenry14 (approved):

crates/cheatnet/tests/starknet/execution.rs
```diff
@@ -46,3 +49,57 @@ fn test_state_reverted() {
     assert!(inner_call.execution.events.is_empty());
     assert!(inner_call.execution.l2_to_l1_messages.is_empty());
 }
+
+#[test]
```
cptartur, line 53: The test could be a bit easier to follow if you added comment with names of the functions being called or some other context

cptartur (approved):

crates/cheatnet/tests/starknet/execution.rs
ddoktorski: Added

PR #3328, opened 2025-05-12
Add runtime check for `snforge_std` version

cptartur:
**Stack**:
- #3333
- #3328 ⬅
- #3327
- #3247
- #3246
- #3245
- #3244
- #3243
- #3242
- #3241
- #3240
- #3238

⚠️ *Part of a stack created by [spr](https://github.com/ejoffe/spr). Do not merge manually using the UI - doing so may have unexpected results.*

PR #3341, opened 2025-05-13
Use a stable toolchain for all jobs

cptartur:
**Stack**:
- #3341 ⬅
- #3333
- #3328
- #3327
- #3247
- #3246
- #3245
- #3244
- #3243
- #3242
- #3241
- #3240
- #3239
- #3238


franciszekjob (commented):
Le't use stable in other workflows too
image

kkawula (changes requested):
There is a little misunderstanding of how `dtolnay/rust-toolchain` works
```
Rustup toolchain specifier e.g. stable, nightly, 1.42.0, nightly-2022-01-01. Important: the default is to match the @rev as described above. When passing an explicit `toolchain` as an input instead of @rev, you'll want to use "dtolnay/rust-toolchain@master" as the revision of the action.
```

Also it is worth mentioning that these actions should stay pinned to commit rev
https://github.com/foundry-rs/starknet-foundry/pull/3283#discussion_r2063126376

Issue #3022, opened 2025-02-27
Foundry UI


cptartur:
### Current State

There is no single, consistent way to format outputs of Starknet Foundry and related tools. Notably, even snforge and sncast do not provide the same UI or dev utilites.

### Objective

Create Foundry UI library

### Additional Context

The implementation can probably be inspired by Scarb UI. 

The library should be easily reusable between different projects aiming to integrate with Foundry.

Details and scope should be decided through a design document.

fathima200:
I’d like to resolve this.

PR #3347, opened 2025-05-15
Improve output of `sncast account create`

### LINKED ISSUE #3254: Improve output of `sncast account declare`
### Current State

It has been reported by Starkware that the current output can be improved, following the output from the other tool.

### Objective

Improve output of `sncast account declare`

### Additional Context

Will update with details once we have them

kkawula:
Closes #3254

## Introduced changes

-

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


franciszekjob (changes requested):

crates/sncast/src/response/structs.rs
```diff
@@ -83,9 +76,7 @@ impl CommandResponse for DeclareResponse {}
 
 #[derive(Serialize)]
 pub struct AccountCreateResponse {
-    pub address: PaddedFelt,
-    #[serde(serialize_with = "crate::response::structs::serialize_as_decimal")]
-    pub max_fee: Felt,
+    pub account_address: PaddedFelt,
```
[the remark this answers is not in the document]
kkawula: So, I can add an entry to changelog, or resign from this change at all, but I like this change and it is more informative for me. We verbose tell that it is account address, not transaction or sth else.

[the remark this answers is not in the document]
franciszekjob: I would resign at all, mostly because:
1. It seems to me that it's pretty clear that it's the address of the account, don't think that `account_` is a game-changer.
2. Soon our outputs will be revamped, and then it won't matter it the field is named `address`/`account_address` and at that point it will be a not needed change.

Let's see what others think.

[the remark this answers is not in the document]
cptartur: With #3022 we'll have more granularity so we'll be able to rename this. For now, let's avoid breaking it.

crates/sncast/tests/e2e/account/create.rs
```diff
@@ -49,11 +49,10 @@ pub async fn test_happy_case(account_type: &str) {
         indoc! {r"
         command: account create
         add_profile: --add-profile flag was not set. No profile added to snfoundry.toml
-        address: 0x0[..]
-        max_fee: [..]
-        message: Account successfully created. Prefund generated address with at least <max_fee> STRK tokens. It is good to send more in the case of higher demand.
+        account_address: 0x0[..]
+        message: Account successfully created. The estimated deployment fee is [..]
```
cptartur, line 53: Why aren't we asserting the whole message. We should have at least one tests that does that.

crates/sncast/src/starknet_commands/account/create.rs
```diff
@@ -379,7 +383,18 @@ fn generate_deploy_command_with_keystore(
...
+fn felt_to_bigdecimal<F, D>(felt: F, decimals: D) -> BigDecimal
+where
+    F: AsRef<Felt>,
+    D: Into<i64>,
+{
+    BigDecimal::new(
+        BigInt::from_bytes_be(Sign::Plus, &felt.as_ref().to_bytes_be()),
+        decimals.into(),
+    )
+}
```
cptartur, line 400: We have no tests for this conversion at all, at least we should add unit tests.

cptartur: And some e2e ensuring it is printed with correct format

crates/sncast/src/starknet_commands/account/create.rs
```diff
@@ -75,15 +77,18 @@ pub async fn create(
...
         generate_account(provider, salt, class_hash, create.account_type).await?;
 
     let address: Felt = account_json["address"]
         .as_str()
         .context("Invalid address")?
         .parse()?;
 
-    let mut message = "Account successfully created. Prefund generated address with at least <max_fee> STRK tokens. It is good to send more in the case of higher demand.".to_string();
+    let mut message = format!(
+        "Account successfully created. The estimated deployment fee is {} STRK. Prefund the account to cover deployment transaction fee",
```
cptartur, line 89:
```suggestion
        "Account successfully created but it needs to be deployed. The estimated deployment fee is {} STRK. Prefund the account to cover deployment transaction fee",
```

cptartur: Also starkli tells that the max fee is `x` but it is good to fund with `x + y` (we should check what's the reasonable overhead for that, maybe some starknet docs have it) so it doesn't fail on price change. I think it could good to have that as well

crates/sncast/src/starknet_commands/account/create.rs
cptartur, line 1: As part of account creation we also print

```
To see account creation details, visit:
account: https://sepolia.starkscan.co/[..]
```

I don't think it makes much sense to do so, let's only show the link after successfully running `account deploy`.

crates/sncast/src/starknet_commands/account/create.rs
kkawula: I don't really like that, they are just [multiplying everything](https://github.com/xJonathanLEI/starkli/blob/master/src/subcommands/account/deploy.rs#L502) by 1.5. I couldn't find any reasonable doc that describe that. Leaving the estimated fee should be fine

crates/sncast/src/starknet_commands/account/create.rs
kkawula: I can't agree, it makes sense in my opinion, because when creating an account we print link to the account itself, but when deploying the printed link refers toa  transaction

cptartur: Right but this a link to an account that isn't really on the network. I know that it becomes "active" after it is funded but I think it might be communicating a wrong thing to the user.

`sncast account deploy` could show both link to the transaction and account, same as `sncast declare` links to a transaction and a contract.

crates/sncast/src/starknet_commands/account/create.rs
cptartur: Not sure if 1.5 is correct multiplier, but we should ensure that the value returned by estimate fee is valid in most cases. Because it might take a user some time to prefund the account and local variations in the fee shouldn't break that.

crates/sncast/src/starknet_commands/account/create.rs
kkawula: I am not a big fan of this change. First output of `sncast account deploy` will get a little messy, secondly, providing a link to "inactive" account is a good handle to pre-fund contract address, imo it is just more verbose than the account address provided in the command output.

cptartur: I don't have a strong opinion on that but I think it can also be a bit misleading to print link to the account if it's not deployed yet. But up to you.

crates/sncast/src/response/structs.rs
```diff
@@ -84,8 +77,6 @@ impl CommandResponse for DeclareResponse {}
 #[derive(Serialize)]
 pub struct AccountCreateResponse {
     pub address: PaddedFelt,
-    #[serde(serialize_with = "crate::response::structs::serialize_as_decimal")]
-    pub max_fee: Felt,
```
cptartur, line 88: This is a breaking change technically, for the json output. Maybe let's included that in the changelog.

cptartur: And have a changelog entry for your improvement as well.

crates/sncast/src/starknet_commands/account/create.rs
```diff
@@ -379,7 +383,71 @@ fn generate_deploy_command_with_keystore(
...
+        let result = felt_to_bigdecimal(felt, 18);
+        assert_eq!(result, BigDecimal::new(BigInt::from(1), 18));
+    }
+
+    #[test]
+    fn test_felt_to_bigdecimal_with_large_number() {
+        let felt = Felt::from_hex("0x1234567890abcdef").unwrap();
+        let result = felt_to_bigdecimal(felt, 18);
+        let expected = BigDecimal::new(BigInt::from_bytes_be(Sign::Plus, &felt.to_bytes_be()), 18);
+        assert_eq!(result, expected);
```
cptartur, line 425: This isn't testing match though if we are using both values from the felt. Can't we create big decimal from some constant value we precompute?

cptartur: i.e. hardcode something

crates/sncast/src/starknet_commands/account/create.rs
kkawula: https://github.com/foundry-rs/starknet-foundry/issues/3403

crates/sncast/src/response/structs.rs
franciszekjob: Was this addresses/resolved?

franciszekjob: Shouldn't it be `estimated_fee` then?

franciszekjob: Shouldn't it be `estimated_fee` then?

kkawula: Renamed

franciszekjob (approved):

cptartur (approved):

PR #3350, opened 2025-05-16
Disable voyager provider in `sncast`

cptartur:
Closes #

## Introduced changes

- Voyager provider seems to be no longer available. Disabled it so we always use Blast

## Checklist

- [ ] Linked relevant issue
- [ ] Added relevant tests

### DIFF
--- a/crates/sncast/src/helpers/rpc.rs
+++ b/crates/sncast/src/helpers/rpc.rs
@@ -5,8 +5,6 @@ use clap::Args;
 use shared::consts::RPC_URL_VERSION;
 use shared::verify_and_warn_if_incompatible_rpc_version;
 use starknet::providers::{JsonRpcClient, jsonrpc::HttpTransport};
-use std::env::current_exe;
-use std::time::UNIX_EPOCH;
 
 #[derive(Args, Clone, Debug, Default)]
 #[group(required = false, multiple = false)]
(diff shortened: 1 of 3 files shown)


ksew1 (approved):

crates/sncast/src/helpers/rpc.rs
```diff
@@ -57,27 +55,13 @@ impl RpcArgs {
...
 
 impl FreeProvider {
     fn semi_random() -> Self {
-        let seed = installation_constant_seed().unwrap_or(2);
-        if seed % 2 == 0 {
-            return Self::Blast;
-        }
-        Self::Voyager
+        FreeProvider::Blast
     }
```
franciszekjob, line 65: It's called `semi_random` and it always returns blast - I know that this PR is a hotfix but this should be changed at some point, either by adding other provider or refactoring the code so the method name is not confusing. Could you please link an issue here?

franciszekjob (approved):

crates/sncast/src/helpers/rpc.rs
cptartur: I'd want to introduce an alternative provider in the future, there exist some other. But don't have time to research them right now.

cptartur: And generally rework this logic, so provider can be overriden by the user. Will do that post hotfix

franciszekjob: Yes sure, let's just not forget about it 👍

PR #3356, opened 2025-05-19
Use rustup in build matrix workflows

cptartur:
Closes #

## Introduced changes

- Uses rustup directly in build workflows to override the settings from `rust-toolchain.toml`

## Checklist

- [ ] Linked relevant issue
- [ ] Added relevant tests

### DIFF
--- a/.github/workflows/_build-binaries.yml
+++ b/.github/workflows/_build-binaries.yml
@@ -48,10 +48,9 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - uses: dtolnay/rust-toolchain@b3b07ba8b418998c39fb20f53e8b695cdcc8de1b
-        with:
-          toolchain: stable
-          target: ${{ matrix.target }}
+      - name: Setup rust
+        run: |
+          rustup target add ${{ matrix.target }} --toolchain 1.86.0
 
       - uses: Swatinem/rust-cache@9d47c6ad4b02e050fd481d890b2ea34778fd09d6
         with:
--- a/.github/workflows/_build-plugin-binaries.yml
+++ b/.github/workflows/_build-plugin-binaries.yml
@@ -68,9 +68,9 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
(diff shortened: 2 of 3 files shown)


.github/workflows/_build-binaries.yml
```diff
@@ -48,10 +48,9 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - uses: dtolnay/rust-toolchain@b3b07ba8b418998c39fb20f53e8b695cdcc8de1b
-        with:
-          toolchain: stable
-          target: ${{ matrix.target }}
+      - name: Setup rust
+        run: |
+          rustup target add ${{ matrix.target }} --toolchain 1.86.0
```
ddoktorski, line 53: Is it necessary when compiling with `cross`?

ddoktorski (approved):

.github/workflows/_build-binaries.yml
cptartur: From this docs `cross` should technically pull the target itself https://github.com/cross-rs/cross/blob/main/docs/getting-started.md#cross-compiling-your-first-package but in case of problems with that they tell you to add the target yourself. So it probably doesn't hurt us to just keep it.

kkawula, line 53: Do we need to provide the toolchain since it's already in `rust-toolchain.toml`?

cptartur: Here it says that if no `--toolchain` is passed it will use the default toolchain https://rust-lang.github.io/rustup/cross-compilation.html

And rust toolchain file is an local override https://rust-lang.github.io/rustup/overrides.html#overrides so it may not pull the correct version.

I've tried locally in our repo by running `rustup target add aarch64-unknown-linux-gnu` and `rustup target add aarch64-unknown-linux-gnu --toolchain 1.86`. Both downloaded new toolchains so they must be different (subsequent reruns of each command told me that toolchains are already installed), using the default configured in rustup.

kkawula (approved):

ddoktorski (approved):

PR #3377, opened 2025-05-19
Enable displaying mixed resources when using `--detailed-resources` flag

ddoktorski:
Related #3114

## Introduced changes

- When tracking `sierra-gas`, display both Sierra gas and VM resources if VM resources are non zero and `--detailed-resources` flag is used
- Remove redundant parentheses around `sierra_gas_consumed` value


crates/forge-runner/src/printing.rs
```diff
@@ -69,32 +70,44 @@ fn format_detailed_resources(
...
-                used_resources.gas_consumed.0, syscalls,
-            )
+        sierra_gas_consumed: {}
+        syscalls: ({syscalls})",
+                used_resources.gas_consumed.0
+            );
+
+            if used_resources.execution_resources != ExecutionResources::default() {
+                output.push_str(&vm_resources_output);
+            }
```
THenry14, line 107: are we making sure somewhere that resources for the same syscall aren't displayed here multiple times? or maybe quite contrary? 

I'm just afraid such situation might be confusing for users (did I just pay twice for a syscall or what happened really?)

ddoktorski: > If some of these costs are included twice

No resources are calculated twice. The only issue is that some syscalls, which should be counted as VM resources, are instead included in the Sierra gas consumed. However, they are still only counted once.

> Maybe just an information that we had to fall back to calculating vm resources in tests and that the values may not be accurate.

We can add a warning or something like that, but hopefully it will be fixed in #3399

THenry14 (approved):
lgtm

crates/forge-runner/src/printing.rs
ddoktorski: Added a warning message in https://github.com/foundry-rs/starknet-foundry/pull/3377/commits/f0681c098438ee08e04d44bf82869e58896620f7

cptartur (approved):

cptartur (approved):

PR #3392, opened 2025-05-22
Foundry UI

franciszekjob:
Towards #3022

**Stack**:
- #3409
- #3405
- #3402
- #3396
- #3393
- #3392 ⬅

## Introduced changes

- Introduce `foundry-ui` crate
    - Add `UI`, `Message`

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/foundry-ui/src/lib.rs
```diff
@@ -0,0 +1,65 @@
...
+impl OutputFormat {
+    #[must_use]
+    pub fn from_flag(json: bool) -> Self {
+        if json {
+            OutputFormat::Json
+        } else {
+            OutputFormat::Human
+        }
+    }
+}
```
franciszekjob, line 21: note: I know that flags responsible for output format is currently only present in sncast, but thinking ahead it's reasonable to have it at generic ui level.

crates/sncast/src/response/structs.rs
franciszekjob, line 1: note: Structs have been placed in their respective files.

franciszekjob, line 1: note: command response structs have been placed in their respective files

cptartur (commented):
Looks good I think, left some comments about possible improvements

crates/foundry-ui/src/message.rs
```diff
@@ -0,0 +1,60 @@
...
+pub trait Message {
+    /// Return textual (human) representation of this message.
+    ///
+    /// Default implementation returns empty string, making [`Ui`] skip printing this message.
+    fn text(&self) -> String
+    where
+        Self: Sized + Serialize,
+    {
+        String::new()
+    }
```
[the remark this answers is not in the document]
franciszekjob: Right, maybe a json would be suitable then, wdyt? Or we could also leave it without default implementation.

cptartur (commented):
Responded to comments

crates/foundry-ui/src/lib.rs
```diff
@@ -0,0 +1,67 @@
...
+
+impl OutputFormat {
+    #[must_use]
+    pub fn from_flag(json: bool) -> Self {
+        if json {
+            OutputFormat::Json
+        } else {
+            OutputFormat::Human
+        }
+    }
```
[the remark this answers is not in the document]
franciszekjob: I don't really see a point to remove it - look at the code where it's used, seems natural and sensible:
https://github.com/foundry-rs/starknet-foundry/blob/32302f6a6ed9b7f94c970ff32e6b8b4cfd15f92b/crates/sncast/src/main.rs#L217

crates/sncast/src/response/explorer_link.rs
```diff
@@ -1,16 +1,16 @@
...
+use foundry_ui::OutputFormat;
 use starknet_types_core::felt::Felt;
 
 pub trait OutputLink {
     const TITLE: &'static str;
 
     fn format_links(&self, provider: Box<dyn LinkProvider>) -> String;
 
-    fn print_links(&self, provider: Box<dyn LinkProvider>) {
+    fn print_links(&self, provider: Box<dyn LinkProvider>) -> String {
```
kkawula, line 10: Rename?

crates/sncast/src/response/print.rs
```diff
@@ -154,6 +127,14 @@ impl From<Value> for OutputData {
     }
 }
 
+impl<T: CommandResponse + Serialize> From<&T> for OutputData {
```
kkawula, line 130: CommandResponse + Serialize = CommandResponse
Is it correct?

crates/sncast/src/starknet_commands/account/deploy.rs
```diff
@@ -63,6 +64,7 @@ pub async fn deploy(
             keystore_path_,
         )
         .await
+        .map(Into::into)
```
kkawula, line 67: can we get rid of this map somehow?

crates/sncast/src/starknet_commands/multicall/run.rs
```diff
@@ -132,6 +132,7 @@ pub async fn run(
 
     execute_calls(account, parsed_calls, fee_args, None, wait_config)
         .await
+        .map(std::convert::Into::into)
```
kkawula, line 135: This formatting differ from other maps

crates/sncast/src/starknet_commands/account/deploy.rs
franciszekjob: Note that now we return a separate type `AccountDeployResponse` which is constructed from `InvokeResponse`, hence the usage of into.

crates/sncast/src/starknet_commands/multicall/run.rs
franciszekjob: Changed.

crates/foundry-ui/src/message.rs
```diff
@@ -0,0 +1,31 @@
...
+    ///
+    /// Default implementation returns empty string, making [`Ui`] skip printing this message.
+    fn text(&self) -> String
+    where
+        Self: Sized + Serialize;
+
+    /// Return JSON representation of this message.
+    fn json(&self) -> String
+    where
+        Self: Sized + Serialize,
```
ksew1, line 15: Sized is not needed here also i would just make Serialize super trait as from what i see we always do `Message + serde::Serialize`

crates/foundry-ui/src/message.rs
```diff
@@ -0,0 +1,31 @@
+use serde::Serialize;
+
+/// A typed object that can be either printed as a human-readable message or serialized as JSON.
+pub trait Message {
+    /// Return textual (human) representation of this message.
+    ///
+    /// Default implementation returns empty string, making [`Ui`] skip printing this message.
+    fn text(&self) -> String
+    where
+        Self: Sized + Serialize;
```
ksew1, line 10: This two are not needed

crates/foundry-ui/src/lib.rs
```diff
@@ -0,0 +1,67 @@
...
+
+    /// Get the output format of this [`UI`] instance.
+    #[must_use]
+    pub fn output_format(&self) -> OutputFormat {
+        self.output_format
+    }
+
+    /// Print the given message to stdout using the configured output format.
+    pub fn print<T>(&self, message: &T)
```
ksew1, line 47: Could we do rust naming's? So `println` and `eprintln`? This way it is less confusing

crates/foundry-ui/src/lib.rs
```diff
@@ -0,0 +1,67 @@
...
+impl UI {
+    /// Create a new [`UI`] instance configured with the given output format.
+    #[must_use]
+    pub fn new(output_format: OutputFormat) -> Self {
+        Self { output_format }
+    }
+
+    /// Get the output format of this [`UI`] instance.
+    #[must_use]
+    pub fn output_format(&self) -> OutputFormat {
```
ksew1, line 42: I don't like how we are using this. We are doing ui.output_format() -> format message according to output_format -> print message. The flow should be i give message to ui and it knows how to format it.

crates/foundry-ui/src/lib.rs
franciszekjob: Sure

crates/foundry-ui/src/lib.rs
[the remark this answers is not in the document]
cptartur: I think it's unnecessary indirection for something this simple. The usage you posted is obvious but from other places, it's not obvious at all what `from_flag` method does since arguments name aren't easily visible to the caller.

Why not 
```rust
let output_format = if cli.json {
  OutputFormat::Json
} else {
  OutputFormat::Human
};
```
at the caller site? We aren't saving any code from doing this.

crates/foundry-ui/src/lib.rs
cptartur: I think the idea is you create a UI instance per output format (json or human readable) and you pass it through the application. To me it makes sense that the UI has a (immutable) state like this.

crates/foundry-ui/src/message.rs
```diff
@@ -0,0 +1,31 @@
...
+    fn text(&self) -> String {
+        (*self).to_string()
+    }
+}
+
+impl Message for String {
+    fn text(&self) -> String {
+        self.to_string()
+    }
+}
```
cptartur, line 31: How does that play with the default conversion to json? The user will just get the encoded string?

Not sure how we can enforce this on the type level (or if), but for json output maybe it makes sense to return objects always? 

What I mean is that maybe there is a reason for the output to always be of the format

```
SomeObject { 
  "key1": "value1"
}
```

and with these default impls the user will get just `"string text"`.

cc @kkawula @ksew1

crates/sncast/src/response/explorer_link.rs
```diff
@@ -22,27 +24,29 @@ pub enum ExplorerError {
...
+        return None;
     };
     let Ok(network) = chain_id.try_into() else {
-        return;
+        return None;
     };
 
     if let Ok(provider) = explorer.unwrap_or_default().as_provider(network) {
-        response.print_links(provider);
+        return Some(response.print_links(provider));
```
cptartur, line 48: Maybe the method should be renamed. It doesn't seem like it is printing anything anymore

crates/foundry-ui/src/lib.rs
```diff
@@ -0,0 +1,67 @@
...
+/// All messaging (basically all writes to `stdout`) must go through this object.
+#[derive(Debug)]
+pub struct UI {
+    output_format: OutputFormat,
+    // TODO(3395): Add state here, that can be used for spinner
+}
+
+impl UI {
+    /// Create a new [`UI`] instance configured with the given output format.
+    #[must_use]
```
cptartur, line 35: Now that I think of it, we should probably define a trait for UI and implement it here.

So you can type our function as accepting the UI trait, not the specific implementation of it.
The benefit will be you will be able to mock the UI in tests or provide an implementation that allows us to just capture what is printed so we can assert it (e.g. for tests that show warnings about `snforge_std`).

cc @kkawula @ksew1

crates/foundry-ui/src/lib.rs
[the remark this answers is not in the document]
franciszekjob: > Why not at the caller site?

Imo it's not a good practice to write ifs logic directly at the level of `main` function, even if they are simple.

Let's follow your reasoning and assume we introduce other simple flags in future. Then our `main` function would get cluttered over the time:


```rust
fn main() -> Result<()> {
    let cli = Cli::parse();

    let numbers_format = NumbersFormat::from_flags(cli.hex_format, cli.int_format);

    let output_format = if cli.json {
        OutputFormat::Json
    } else {
        OutputFormat::Human
    };

    let other_value = cli.other_field {
        ...
    } else {
        ...
    }
   
   ...
}
```

I'm open for discussion but the arguments you've provided don't really convince me 😅 

@ksew1 @kkawula wdyt about this?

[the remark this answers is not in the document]
kkawula: I prefer the solution that @franciszekjob proposed, it takes away logic from main function, let's stick to it

crates/sncast/src/response/print.rs
kkawula: imo we should use`T: CommandResponse` only

crates/foundry-ui/src/lib.rs
```diff
@@ -0,0 +1,68 @@
...
+    /// Get the output format of this [`UI`] instance.
+    #[must_use]
+    pub fn output_format(&self) -> OutputFormat {
+        self.output_format
+    }
+
+    /// Print the given message to stdout using the configured output format.
+    pub fn println<T>(&self, message: &T)
+    where
+        T: Message + serde::Serialize,
```
kkawula, line 50: Isn't it the same?
```suggestion
        T: Message
```

crates/foundry-ui/src/lib.rs
```diff
@@ -0,0 +1,68 @@
...
+        match self.output_format {
+            OutputFormat::Human => println!("{}", message.text()),
+            OutputFormat::Json => println!("{}", message.json()),
+        }
+    }
+
+    /// Print the given message to stderr using the configured output format.
+    pub fn eprintln<T>(&self, message: &T)
+    where
+        T: Message + serde::Serialize,
```
kkawula, line 61:
```suggestion
        T: Message
```

crates/sncast/src/response/print.rs
franciszekjob: Right, we can leave just `T: CommandResponse`

crates/foundry-ui/src/message.rs
```diff
@@ -0,0 +1,23 @@
...
+/// A typed object that can be either printed as a human-readable message or serialized as JSON.
+pub trait Message: Serialize {
+    /// Return textual (human) representation of this message.
+    ///
+    /// Default implementation returns empty string, making [`Ui`] skip printing this message.
+    fn text(&self) -> String;
+
+    /// Return JSON representation of this message.
+    fn json(&self) -> String {
+        serde_json::to_string(self).unwrap_or_else(|_| "Invalid JSON".to_string())
```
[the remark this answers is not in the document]
franciszekjob: You're right, I've missed it. Changed to `serde_json::to_string(self).expect("Failed to serialize message to JSON")`

crates/foundry-ui/src/lib.rs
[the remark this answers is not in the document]
franciszekjob: >  I would personally create a function output_format_from_flag(flag: bool) -> OutputFormat on the caller side.

Why separate function if it's related only with `OutputFormat` (and it's not generic)? I would keep it as a method.

cptartur (commented):
Added my review above but it is telling me it's still requested

crates/sncast/src/response/account/create.rs
```diff
@@ -0,0 +1,49 @@
...
+use serde::{Serialize, Serializer};
+use starknet_types_core::felt::Felt;
+
+use crate::{
+    helpers::block_explorer::LinkProvider,
+    response::{command::CommandResponse, explorer_link::OutputLink},
+};
+
+#[derive(Clone)]
+pub struct Decimal(pub u64);
```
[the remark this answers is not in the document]
ksew1: But why use it instead of u64?

crates/foundry-ui/src/message.rs
franciszekjob: Right, valid point - I've removed default impl of `json` on `Message. Now both `text` and `json` need to be implemented, hope it makes more sense.

crates/foundry-ui/src/message.rs
```diff
@@ -0,0 +1,23 @@
...
+    }
+}
+
+impl<T: ToString> Message for T
```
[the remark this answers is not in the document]
franciszekjob: Seems like solution from https://github.com/foundry-rs/starknet-foundry/pull/3392#discussion_r2109636629 thread, resolves this.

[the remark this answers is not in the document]
cptartur: This is better probably, though I still don't 100% see use cases for having a plain string as a command return message. 
What are the cases where we'd use this impl you have here instead of a specialized struct?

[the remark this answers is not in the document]
franciszekjob: For instance a simple print 😅 :
```rust
ui.println("Example text");
```

crates/foundry-ui/src/lib.rs
```diff
@@ -0,0 +1,65 @@
...
+pub mod message;
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
+pub enum OutputFormat {
+    #[default]
+    Human,
+    Json,
+}
+
+pub trait Ui {
```
ksew1: I this trait even used?

crates/foundry-ui/src/lib.rs
```diff
@@ -0,0 +1,68 @@
...
+
+    // TODO: Will be removed in 3022-3-use-foundry-ui-in-sncast
+    /// Get the output format of this [`UI`] instance.
+    #[must_use]
+    pub fn output_format(&self) -> OutputFormat {
+        self.output_format
+    }
+
+    /// Print the given message to stdout using the configured output format.
+    pub fn println<T>(&self, message: &T)
```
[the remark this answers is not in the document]
ksew1: yeah `format_message` is a good name

crates/foundry-ui/src/message.rs
```diff
@@ -0,0 +1,24 @@
+use serde::Serialize;
+
+/// A typed object that can be either printed as a human-readable message or serialized as JSON.
+pub trait Message: Serialize {
+    /// Return textual (human) representation of this message.
+    fn text(&self) -> String;
+
+    /// Return JSON representation of this message.
+    fn json(&self) -> String;
```
ksew1: After some thought i think it would be better to return `serde::Value`

crates/foundry-ui/src/lib.rs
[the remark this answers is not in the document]
franciszekjob: But it won't be used anywhere now, so there's no point adding it.

crates/foundry-ui/src/lib.rs
franciszekjob: Yeah, now it's used. Btw I've changed it's name to `Printer`.

crates/foundry-ui/src/message.rs
franciszekjob: Makes sense

crates/foundry-ui/src/lib.rs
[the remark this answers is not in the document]
ksew1: It would de duplicate code

crates/foundry-ui/src/lib.rs
ksew1: Where? I only see implementation not the usage

crates/foundry-ui/src/lib.rs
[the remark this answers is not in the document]
ksew1: Also it would be nicer to use `impl` syntax instead of generics

crates/sncast/src/response/account/create.rs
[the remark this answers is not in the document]
franciszekjob: Looking at the code, it was used to display estimated fee (in account creation response) as decimal number, rather than hex (it was felt). Some time ago type of estimated fee changed from felt to u128, so we can now safely remove this code.

crates/foundry-ui/src/lib.rs
franciszekjob: https://github.com/foundry-rs/starknet-foundry/pull/3392/files#diff-58fbed0f1346540ed57607881e130f1cf525dcfe591519e6d90ea89969b64d91R52

ksew1: Yes, but we don't use this trait anywhere. It is just implemented

crates/foundry-ui/src/message.rs
```diff
@@ -0,0 +1,24 @@
...
+    /// Return textual (human) representation of this message.
+    fn text(&self) -> String;
+
+    /// Return JSON representation of this message.
+    fn json(&self) -> Value;
+}
+
+impl<T: ToString> Message for T
+where
+    T: Serialize,
```
ksew1, line 15:
```suggestion
    T: Serialize + ToString,
```

And remove `
impl<T: ToString>`

ksew1 (approved):
My only question is about this trait

crates/foundry-ui/src/message.rs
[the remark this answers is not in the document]
cptartur: All right, let's keep this implementation then.

crates/foundry-ui/src/message.rs
```diff
@@ -0,0 +1,23 @@
+use serde::Serialize;
+
+/// A typed object that can be either printed as a human-readable message or serialized as JSON.
+pub trait Message: Serialize {
```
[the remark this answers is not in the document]
cptartur: It's an nice idea but for now let's not expand the scope of this PR (stack).

crates/sncast/src/response/print.rs
```diff
@@ -76,6 +63,7 @@ impl From<Value> for OutputValue {
             ),
             Value::String(s) => OutputValue::String(s.to_string()),
             Value::Bool(b) => OutputValue::String(b.to_string()),
+            Value::Number(n) => OutputValue::String(n.to_string()),
```
cptartur, line 66: Why do we need this?

crates/foundry-ui/src/message.rs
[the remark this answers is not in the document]
kkawula: > Seems that should be a new variant of OutputFormat

Sounds good

crates/sncast/src/main.rs
```diff
@@ -24,7 +31,7 @@ use sncast::helpers::scarb_utils::{
     get_package_metadata, get_scarb_metadata_with_deps,
 };
 use sncast::response::errors::handle_starknet_command_error;
-use sncast::response::structs::{CallResponse, DeclareResponse, TransformedCallResponse};
+
```
kkawula:
```suggestion
```

crates/sncast/src/response/cast_message.rs
```diff
@@ -0,0 +1,43 @@
...
+    command::CommandResponse,
+    print::{Format, OutputData},
+};
+
+#[derive(Serialize)]
+pub struct SncastMessage<T: Serialize> {
+    pub command: String,
+    pub command_response: T,
+    pub numbers_format: NumbersFormat,
+}
```
kkawula:
```suggestion
#[derive(Serialize)]
pub struct SncastMessage<T> {
    pub command: String,
    pub command_response: T,
    pub numbers_format: NumbersFormat,
}
```
That Trait bound is required by derive

crates/sncast/src/starknet_commands/account/create.rs
```diff
@@ -86,9 +85,9 @@ pub async fn create(
         .context("Invalid address")?
         .parse()?;
 
+    let estimated_fee_strk = BigDecimal::new(estimated_fee.into(), 18.into());
     let mut message = format!(
-        "Account successfully created but it needs to be deployed. The estimated deployment fee is {} STRK. Prefund the account to cover deployment transaction fee",
-        felt_to_bigdecimal(estimated_fee, 18)
+        "Account successfully created but it needs to be deployed. The estimated deployment fee is {estimated_fee_strk} STRK. Prefund the account to cover deployment transaction fee"
```
kkawula, line 90: Is there a reason why this function should be removed?

crates/sncast/src/starknet_commands/account/create.rs
```diff
@@ -375,67 +374,3 @@ fn generate_deploy_command_with_keystore(
...
-        BigInt::from_bytes_be(Sign::Plus, &felt.as_ref().to_bytes_be()),
-        decimals.into(),
-    )
-}
-
-#[cfg(test)]
-mod tests {
-    use super::*;
-
-    #[test]
```
kkawula, line 394: Is there a reason why this function should be removed?

kkawula (changes requested):

crates/foundry-ui/src/message.rs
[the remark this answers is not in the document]
franciszekjob: https://github.com/foundry-rs/starknet-foundry/issues/3435

crates/sncast/src/starknet_commands/account/create.rs
franciszekjob: Yes, if you read the previous code, `estimated_fee` was of `Felt` type, which is quite artifical, because the estimated fee returned from `starknet-rs` is `u128`. I've changed the type in `AccountCreateResponse` to mentioned `u128` ([here](https://github.com/foundry-rs/starknet-foundry/pull/3392/files#diff-70a96365d3d53e9082de907ba5b9661e4940d5b55d3c7fd1227021447d0a8556R20)), hence we don't need this function anymore.

crates/sncast/src/starknet_commands/account/create.rs
franciszekjob: https://github.com/foundry-rs/starknet-foundry/pull/3392#discussion_r2125921742

crates/sncast/src/response/cast_message.rs
franciszekjob: In fact, it should be `CommandResponse` instead of `Serialized`, changed.

crates/sncast/src/response/print.rs
franciszekjob: It's a result of deleting `Decimal`, here https://github.com/foundry-rs/starknet-foundry/pull/3392/files#diff-b07be4e12cdea6b79c88ef81620caabdbf5d4250c3f73c912fc0eda9467b3bbeL11

kkawula (approved):

PR #3393, opened 2025-05-22
Add message components

franciszekjob:
Towards #3022 

**Stack**:
- #3409
- #3405
- #3402
- #3396
- #3393 ⬅
- #3392

## Introduced changes

Add `TypedMessage` and `LabeledMessage` components to foundry-ui

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/foundry-ui/src/components/mod.rs
```diff
@@ -0,0 +1,8 @@
+//! This module provides various ready to use message types for use with
+//! a [`Ui`][crate::Ui].
+
+pub use tagged::*;
+pub use typed::*;
+
+mod tagged;
+mod typed;
```
cptartur, line 8: If we are reexporting whole modules why not just do this?
```suggestion
pub mod tagged;
pub mod typed;
```

crates/foundry-ui/src/components/tagged.rs
```diff
@@ -0,0 +1,40 @@
...
+impl<'a> TaggedMessage<'a> {
+    #[must_use]
+    pub fn styled(tag: &'a str, text: &'a str, type_style: &'a str) -> Self {
+        Self {
+            tag,
+            text,
+            type_style: Some(type_style),
+        }
+    }
+}
```
cptartur, line 27: In which cases `type_style` will be able to be none if we don't allow creating such instances?

crates/foundry-ui/src/components/tagged.rs
```diff
@@ -0,0 +1,40 @@
...
+pub struct TaggedMessage<'a> {
+    tag: &'a str,
+    text: &'a str,
+    #[serde(skip)]
+    type_style: Option<&'a str>,
+}
+
+impl<'a> TaggedMessage<'a> {
+    #[must_use]
+    pub fn styled(tag: &'a str, text: &'a str, type_style: &'a str) -> Self {
```
cptartur, line 20: This is cool but if we want to have consistent styling we would probably want to have some enum with predefined styles fir different things in the future.

crates/foundry-ui/src/components/typed.rs
```diff
@@ -0,0 +1,40 @@
...
+use serde::Serialize;
+
+use crate::Message;
+
+/// Generic textual message with `ty` prefix.
+///
+/// The type prefix can be stylized in text mode.
+/// e.g. "Tests: 1 passed, 1 failed"
+#[derive(Serialize)]
+pub struct TypedMessage<'a> {
```
cptartur, line 11: Typed is a bit misleading naming for me, it suggest it has something to do with the type definition itself or the type used.

crates/foundry-ui/src/components/typed.rs
```diff
@@ -0,0 +1,40 @@
...
+    type_style: Option<&'a str>,
+}
+
+impl<'a> TypedMessage<'a> {
+    #[must_use]
+    pub fn styled(ty: &'a str, text: &'a str, type_style: &'a str) -> Self {
+        Self {
+            ty,
+            text,
+            type_style: Some(type_style),
```
cptartur, line 24: Same here, will we ever allow creating it with it being `None`?

crates/foundry-ui/src/components/tagged.rs
franciszekjob: Added `raw()` method, which doesn't have `tyg_style` param, then it's `None`.

crates/foundry-ui/src/components/typed.rs
franciszekjob: Same as https://github.com/foundry-rs/starknet-foundry/pull/3393#discussion_r2106255953

crates/foundry-ui/src/components/tagged.rs
franciszekjob: To me it makes more sense when these styles are predefined inside specific print methods on `UI` (e.g. `print_warning`, `print_error`), wdyt?

crates/foundry-ui/src/components/typed.rs
franciszekjob: Hmm, agree, maybe "labeled" would be better, will rename it.

crates/foundry-ui/src/components/tagged.rs
cptartur: Yeah it probably doesn't make sense now. But if we eventually get a lot of styles then it will make sense to abstract that somehow probably.

cptartur (approved):

crates/foundry-ui/src/components/tagged.rs
franciszekjob: Created https://github.com/foundry-rs/starknet-foundry/issues/3408

ksew1 (commented):
#

crates/foundry-ui/src/components/tagged.rs
```diff
@@ -0,0 +1,51 @@
...
+use crate::Message;
+
+/// Generic textual message with `tag` prefix.
+///
+/// The tag prefix can be stylized in text mode.
+/// e.g. "[WARNING]: An example warning message"
+#[derive(Serialize)]
+pub struct TaggedMessage<'a> {
+    tag: &'a str,
+    text: &'a str,
```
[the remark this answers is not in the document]
franciszekjob: > Shouldn't it be `&T where T: Message`?

Hmm, how would you see that?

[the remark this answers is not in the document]
ksew1: Suppose i want to print typed message as error. This holds for all components with  `text: &'a str`.

crates/foundry-ui/src/components/warning.rs
```diff
@@ -0,0 +1,25 @@
+use serde::Serialize;
+
+use crate::Message;
+
+use super::tagged::TaggedMessage;
+
+/// Warning textual message.
+#[derive(Serialize)]
+pub struct WarningMessage<'a> {
+    text: &'a str,
```
ksew1, line 10: This as JSON will be `{ "test": "something" }`, which is the same as the `ErrorMessage` serialization.

If there are mistakes at this stage, I believe there should not be an implicit implementation of the `json` method.

crates/foundry-ui/src/components/tagged.rs
```diff
@@ -0,0 +1,51 @@
...
+
+    /// Field which dictates the style of the tag as a string that `console::Style` can interpret.
+    #[serde(skip)]
+    tag_style: Option<&'a str>,
```
ksew1, line 17: Do we need this field? We could just pass styled tag to the component. Which gives more flexibility

crates/foundry-ui/src/components/tagged.rs
```diff
@@ -0,0 +1,51 @@
...
+    pub fn raw(tag: &'a str, text: &'a str) -> Self {
+        Self {
+            tag,
+            text,
+            tag_style: None,
+        }
+    }
+}
+
+impl Message for TaggedMessage<'_> {
```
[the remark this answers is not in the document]
franciszekjob: > I know that for `json` we don't want to have any coloring but what about formatting? We are loosing any information about tag/label there.
> 
> Maybe it's not important to implement right away, but it could make sense to have some field in the json about the tag/label.
> 
> cc @kkawula @ksew1

Current implementation supports displaying messages in json format, e.g. here is serialized `WarningMessage`:
```json
{"tag":"WARNING","text":"This is a serious warning!"}
```

crates/foundry-ui/src/components/warning.rs
franciszekjob: Changed, now it's:

```json
{"message_type":"warning","message":"This is a serious warning!"}
```

Wdyt @ksew1 ?

crates/foundry-ui/src/components/tagged.rs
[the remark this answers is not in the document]
franciszekjob: > Suppose i want to print typed message as error. This holds for all components with `text: &'a str`.

Why? It's possible:

```rust
ui.eprintln(&LabeledMessage::raw("label", "Some text"));
```

crates/foundry-ui/src/components/tagged.rs
franciszekjob: Right, it will be simpler if we don't hold it here, removed.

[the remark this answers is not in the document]
franciszekjob: Resolved by https://github.com/foundry-rs/starknet-foundry/pull/3393#discussion_r2121254825

crates/foundry-ui/src/components/tagged.rs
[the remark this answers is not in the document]
franciszekjob: Updated

crates/foundry-ui/src/components/tagged.rs
```diff
@@ -0,0 +1,35 @@
+use serde::Serialize;
+
+use crate::Message;
+
+/// Generic message with `tag` prefix.
+///
+/// The tag prefix can be stylized in text mode.
+/// e.g. "[WARNING]: An example warning message"
+#[derive(Serialize)]
+pub struct TaggedMessage<'a, T: Message + Serialize> {
```
kkawula: It isn't consistent across all `Message` types

kkawula: A it probably should stay as Message standalone

crates/foundry-ui/src/components/error.rs
```diff
@@ -0,0 +1,35 @@
...
+use serde::Serialize;
+
+use crate::Message;
+
+use super::tagged::TaggedMessage;
+
+/// Error message.
+#[derive(Serialize)]
+pub struct ErrorMessage<'a, T: Message> {
+    message_type: &'a str,
```
ksew1: I see that we are using this field only for json serialization. I think just using `json!` macro would be more convenient and would be easier to read.

crates/foundry-ui/src/components/tagged.rs
```diff
@@ -0,0 +1,35 @@
...
+    }
+}
+
+impl<T: Message> Message for TaggedMessage<'_, T> {
+    fn text(&self) -> String {
+        format!("[{}] {}", self.tag, self.message.text())
+    }
+
+    fn json(&self) -> String {
+        serde_json::to_string(self).expect("Failed to serialize as JSON")
```
ksew1: This needs to call self.message.json()

crates/foundry-ui/src/components/error.rs
```diff
@@ -0,0 +1,35 @@
...
+        Self {
+            message_type: "error",
+            message,
+        }
+    }
+}
+
+impl<T: Message> Message for ErrorMessage<'_, T> {
+    fn text(&self) -> String {
+        let tag = Style::new().red().apply_to("ERROR").to_string();
```
ksew1:
```suggestion
        let tag = style("ERROR").red().to_string();
```

crates/foundry-ui/src/components/labeled.rs
```diff
@@ -0,0 +1,35 @@
+use serde::Serialize;
+
+use crate::Message;
+
+/// Generic message with `label` prefix.
+///
+/// The label prefix can be stylized in text mode.
+/// e.g. "Tests: 1 passed, 1 failed"
+#[derive(Serialize)]
+pub struct LabeledMessage<'a, T> {
```
ksew1: Is this dead code?

crates/foundry-ui/src/components/warning.rs
```diff
@@ -0,0 +1,35 @@
...
+        Self {
+            message_type: "warning",
+            message,
+        }
+    }
+}
+
+impl<T: Message> Message for WarningMessage<'_, T> {
+    fn text(&self) -> String {
+        let tag = Style::new().yellow().apply_to("WARNING").to_string();
```
ksew1: Same suggestion as for Error

ksew1 (approved):

cptartur (approved):

kkawula (approved):

PR #3396, opened 2025-05-22
Use foundry UI in sncast

franciszekjob:
Towards #3022

**Stack**:
- #3409
- #3405
- #3402
- #3396 ⬅
- #3393
- #3392

## Introduced changes

All printed output from `sncast` is now done through foundry-ui

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/forge/src/lib.rs
```diff
@@ -7,6 +7,7 @@ use derive_more::Display;
 use forge_runner::CACHE_DIR;
 use forge_runner::debugging::TraceVerbosity;
 use forge_runner::forge_config::ForgeTrackedResource;
+use foundry_ui::Ui;
```
cptartur, line 10: Super nit, name it `UI` against convention because that looks better to me 😅

crates/forge/test_utils/src/runner.rs
cptartur, line 1: General comment for all functions, it could make sense to refactor signatures of some of them so they don't take a million (plus a new one) arguments.

Maybe that should be part of the stack, but we can talk it through at the next meeting.

crates/forge/tests/integration/cheat_block_number.rs
```diff
@@ -131,7 +132,8 @@ fn cheat_block_number_basic() {
         .unwrap()
     );
 
-    let result = run_test_case(&test, ForgeTrackedResource::CairoSteps);
```
cptartur, line 134: There should be a wrapper for doing `run_test_case` with a default UI, so we don't have to add it in every test

cptartur (commented):
Integration looks simple enough, except for the number of arguments.

Sidenote, maybe the work for sncast and forge could be separate PRs. It's bit hard to follow in which parts where the most significant changes

franciszekjob:
> Sidenote, maybe the work for sncast and forge could be separate PRs. It's bit hard to follow in which parts where the most significant changes

Thath's what I'm doing but due to the fact that we're using shared functions across our crates, I had to apply some changes in other crates than sncast :/

cptartur (approved):
Probably is fine, didn't review in detail assuming it's all trivial stuff.

crates/sncast/src/main.rs
```diff
@@ -846,7 +858,7 @@ fn process_command_result<T>(
     result: Result<T>,
     numbers_format: NumbersFormat,
     ui: &UI,
-    block_explorer_link: Option<String>,
+    block_explorer_link: Option<OutputLinkMessage>,
 ) where
     T: serde::Serialize + Clone + CommandResponse,
     SncastMessage<T>: Message + Serialize,
```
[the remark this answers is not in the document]
cptartur: Do we need this whole constraint `SncastMessage<T>: Message + Serialize`?

cptartur (commented):
Didn't review whole, will review the rest later

crates/sncast/src/main.rs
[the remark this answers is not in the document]
franciszekjob: Actually we can remove it - done.

crates/foundry-ui/src/lib.rs
```diff
@@ -25,7 +25,7 @@ impl OutputFormat {
 /// colour, etc.
 ///
 /// All messaging (basically all writes to `stdout`) must go through this object.
-#[derive(Debug)]
+#[derive(Debug, Default, Clone)]
```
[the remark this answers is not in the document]
franciszekjob: Leftover, removed.

crates/sncast/src/response/explorer_link.rs
```diff
@@ -45,7 +66,7 @@ pub fn block_explorer_link_if_allowed<T: OutputLink>(
     };
 
     if let Ok(provider) = explorer.unwrap_or_default().as_provider(network) {
-        return Some(response.print_links(provider));
+        return Some(OutputLinkMessage::new(response, provider));
```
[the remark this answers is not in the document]
franciszekjob: `ExplorerLinksMessage` seems more meaningful, renamed.

crates/sncast/tests/integration/wait_for_tx.rs
```diff
@@ -90,20 +93,28 @@ async fn test_wait_for_reverted_transaction() {
...
-        .unwrap();
+    let ui = UI::default();
+    wait_for_tx(
+        &provider,
+        transaction_hash,
+        ValidatedWaitParams::new(1, 3),
+        &ui,
+    )
+    .await
+    .map_err(std::convert::Into::<anyhow::Error>::into)
```
ksew1:
```suggestion
    .map_err(anyhow::Error::from)
```

ksew1 (approved):

crates/sncast/src/starknet_commands/account/list.rs
```diff
@@ -105,73 +106,116 @@ fn read_and_flatten(
...
-    print_if_some("address", data.address.as_ref());
-    print_if_some("salt", data.salt.as_ref());
-    print_if_some("class hash", data.class_hash.as_ref());
-    print_if_some("deployed", data.deployed.as_ref());
-    print_if_some("legacy", data.legacy.as_ref());
-    print_if_some("type", data.account_type.as_ref());
-    println!();
-}
+        if let Some(ref network) = self.network {
+            let _ = writeln!(result, "  network: {network}");
```
cptartur, line 114: Wouldn't we be fine with `pust_str` instead?

crates/sncast/src/starknet_commands/account/list.rs
```diff
@@ -105,73 +106,116 @@ fn read_and_flatten(
...
+impl Message for AccountsListMessage {
+    fn text(&self) -> String {
+        let accounts_file_path = self
+            .accounts_file
+            .canonicalize()
+            .expect("Failed to resolve the accounts file path");
+
+        let accounts_file_path = accounts_file_path
+            .to_str()
+            .expect("Failed to resolve an absolute path to the accounts file");
```
cptartur: That's a change in behavior no? We will now panic in case where accounts file has invalid data and previously it bubbled an error.

crates/sncast/tests/helpers/devnet.rs
```diff
@@ -68,7 +70,9 @@ fn start_devnet() {
         if verify_devnet_availability(&format!("{host}:{port}")) {
             break;
         } else if now.elapsed() >= timeout {
-            eprintln!("Timed out while waiting for devnet!");
+            UI::default().println(&ErrorMessage::new(
+                &"Timed out while waiting for devnet to start!",
+            ));
```
cptartur: I mean, we can use UI here but for printing in tests it's still fine to use `println!` etc.

crates/sncast/src/main.rs
```diff
@@ -457,7 +458,7 @@ async fn run_async_command(
 
                         process_command_result("multicall new", result, numbers_format, ui, None);
                     } else {
-                        println!("{DEFAULT_MULTICALL_CONTENTS}");
+                        ui.println(&DEFAULT_MULTICALL_CONTENTS.to_string());
```
kkawula:
```suggestion
                        ui.println(&DEFAULT_MULTICALL_CONTENTS));
```

crates/sncast/src/main.rs
```diff
@@ -512,7 +516,9 @@ async fn run_async_command(
                         result.as_ref().ok().and_then(|r| r.account_name.clone())
                     {
                         if let Err(err) = prompt_to_add_account_as_default(account_name.as_str()) {
-                            eprintln!("Error: Failed to launch interactive prompt: {err}");
+                            ui.eprintln(&format!(
+                                "Error: Failed to launch interactive prompt: {err}"
+                            ));
```
kkawula, line 521: Shouldn't it be `ErrorMessage`?

crates/sncast/src/main.rs
```diff
@@ -590,13 +596,14 @@ async fn run_async_command(
                             .name
                             .expect("Must be provided if not using a keystore"),
                     ) {
-                        eprintln!("Error: Failed to launch interactive prompt: {err}");
+                        ui.eprintln(&format!(
+                            "Error: Failed to launch interactive prompt: {err}"
+                        ));
```
kkawula, line 602: Shouldn't it be `ErrorMessage`?

kkawula (changes requested):

crates/sncast/src/starknet_commands/account/list.rs
franciszekjob: Thanks to `write!`  we avoid extra allocation, that's even suggested by clippy, see [`format_push_string`](https://rust-lang.github.io/rust-clippy/master/index.html#format_push_string)

crates/sncast/src/starknet_commands/account/list.rs
franciszekjob: Refactored, now it's bubbled in `new` method

crates/sncast/src/main.rs
franciszekjob: No, `ErrorMessage` is whole `[ERROR] Example error...`, this is different

crates/sncast/src/main.rs
franciszekjob: Same as above

crates/sncast/src/main.rs
kkawula: Why, this is different? I suppose that it shouldn't be, one of the main reasons why we would like to introduce UI was to be consistent in each message, and this example breaks that rule, 
```suggestion
                            ui.eprintln(&ErrorMessage::new(
                                format!("Failed to launch interactive prompt: {err}").as_str(),
                            ));
```

franciszekjob: Note that I didn't change the actual output, I've only adjusted it to use the introduced UI. Your remark is generally right, but it's tied to the aspect of consistency for displaying errors in snforge and sncast. Created #3436 for that.

kkawula (approved):

PR #3400, opened 2025-05-23
Add workflow to build nightly releases

### LINKED ISSUE #3300: Nightly releases
### Current State

.

### Objective

Nightly releases

### Additional Context

Create a pipeline for publishing nightly releases

ddoktorski:
Closes #3300


ddoktorski:
Example run: https://github.com/ddoktorski/starknet-foundry/actions/runs/15210212174
Example release: https://github.com/ddoktorski/starknet-foundry-nightlies/releases/tag/nightly-2025-05-20-15210212174 (run id included in the name only for testing)

ddoktorski:
TODO:
* Set `SNFOUNDRY_NIGHTLIES_CONTENTS_WRITE` secret in the repository as a token with write permission to https://github.com/foundry-rs/starknet-foundry-nightlies

kkawula:
#3423

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,187 @@
+name: Nightly
+
+on:
+  workflow_dispatch:
+    inputs:
+      dry_run:
+        description: "Dry run"
+        type: boolean
+        default: true
```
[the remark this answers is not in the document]
kkawula: As it has default set to true, I don't think so

.github/workflows/_publish-plugin.yml
```diff
@@ -28,10 +28,10 @@ jobs:
           
           snforge_scarb_plugin_version=$(get_version "${{ inputs.overridden_plugin_version }}")
           
+          # TODO: It should check dev registry if prod_registry is false
```
cptartur: What about this todo?

.github/workflows/_build-binaries.yml
```diff
@@ -7,6 +7,9 @@ on:
       version:
         required: true
         type: string
+      ref:
```
cptartur, line 10: Don't we want to use some more descriptive name?

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,187 @@
...
+
+      - name: Configure Git for committing
+        run: |
+          git config user.name github-actions
+          git config user.email github-actions@github.com
+
+      - name: Determine nightly version
+        id: version
+        shell: bash
+
```
cptartur:
```suggestion
```

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,187 @@
...
+    # Do not run on dry_run
+    if: ${{ !(inputs.dry_run) }}
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          ref: ${{ needs.prepare.outputs.nightly_branch }}
+
+      - name: Create source code archives
+        run: |
+          git archive "--prefix=starknet-foundry-${{ needs.prepare.outputs.nightly_tag }}/" -o "starknet-foundry-${{ needs.prepare.outputs.nightly_tag }}.zip" HEAD
```
cptartur, line 126: I think we can ignore windows specific builds from this workflow, will be a bit simpler to maintain.

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,187 @@
...
+            # We remove the version tag from the filename so it can
+            # be easily accessed in asdf and installation scripts.
+            #
+            # For example:
+            # starknet-foundry-v0.44.0+nightly-2025-05-22-aarch64-apple-darwin.tar.gz
+            # becomes:
+            # starknet-foundry-nightly-2025-05-22-aarch64-apple-darwin.tar.gz
+
+            if ! [[ $(grep "#" <<< $file) ]]; then
+              label=$(echo "$file" | sed -E "s/v[^+]*\+//" | sed -E "s|.*/||")
```
cptartur: `sed -E "s|.*/||"` 

Is this valid syntax? With vertical lines?

.github/workflows/_publish-plugin.yml
kkawula: https://github.com/foundry-rs/starknet-foundry/pull/3423

.github/workflows/_build-binaries.yml
ddoktorski: It seems natural to name it `ref`, as the same name is used by [actions/checkout](https://github.com/actions/checkout?tab=readme-ov-file#checkout-a-different-branch)

.github/workflows/nightly.yml
ddoktorski: Actually, it seems like more effort to add specific logic to exclude Windows than to just keep it. I would opt to leave it for now and we can always adjust it later if there are any maintenance issues.

.github/workflows/nightly.yml
ddoktorski: Yes, they are allowed in extended regular expressions: https://www.gnu.org/software/sed/manual/sed.html#ERE-syntax

kkawula (approved):

RELEASING.md
```diff
@@ -13,3 +13,16 @@ with the name format `vMAJOR.MINOR.PATCH`.
 
 Note that in this case `CHANGELOG.md`, `Cargo.toml` and `Cargo.lock` files
 have to be updated accordingly.
+
+# Nightly releases
+
+Nightly release must be triggered manually using the Nightly GitHub action.
+This action builds binaries from specified ref and uploads them to the [starknet-foundry-nightlies](https://github.com/foundry-rs/starknet-foundry-nightlies) repository.
+Additionally, there are `stds` and `plugin` version uploaded to the [_dev_ registry](https://scarbs.dev/).
+After a successful release, [Ma'at](https://github.com/software-mansion/maat) is automatically trigger to run experiments in nightly workspace. Results can be found [here](https://docs.swmansion.com/maat/)
```
franciszekjob:
```suggestion
After a successful release, [Ma'at](https://github.com/software-mansion/maat) is automatically triggered to run experiments in nightly workspace. Results can be found [here](https://docs.swmansion.com/maat/)
```

RELEASING.md
```diff
@@ -13,3 +13,16 @@ with the name format `vMAJOR.MINOR.PATCH`.
 
 Note that in this case `CHANGELOG.md`, `Cargo.toml` and `Cargo.lock` files
 have to be updated accordingly.
+
+# Nightly releases
+
+Nightly release must be triggered manually using the Nightly GitHub action.
+This action builds binaries from specified ref and uploads them to the [starknet-foundry-nightlies](https://github.com/foundry-rs/starknet-foundry-nightlies) repository.
+Additionally, there are `stds` and `plugin` version uploaded to the [_dev_ registry](https://scarbs.dev/).
```
franciszekjob:
```suggestion
Additionally, there are `stds` and `plugin` versions uploaded to the [_dev_ registry](https://scarbs.dev/).
```

franciszekjob (approved):

.github/workflows/nightly.yml
cptartur: We can remove windows entirely from build workflows. @kkawula will do so as a prerequisite to #3440 

But to avoid conflicts, you can keep it for now.

.github/workflows/nightly.yml
cptartur: TIL that you can use any value as delimiter in sed https://backreference.org/2010/02/20/using-different-delimiters-in-sed/index.html
Doesn't seem to require `-E` mode, I think it can be removed in case of your regex.

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,186 @@
...
+          do
+            # We remove the version tag from the filename so it can
+            # be easily accessed in asdf and installation scripts.
+            #
+            # For example:
+            # starknet-foundry-v0.44.0+nightly-2025-05-22-aarch64-apple-darwin.tar.gz
+            # becomes:
+            # starknet-foundry-nightly-2025-05-22-aarch64-apple-darwin.tar.gz
+
+            if ! [[ $(grep "#" <<< $file) ]]; then
```
cptartur: Why do we only change the files that don't include `#`?

ddoktorski: We rename build artifacts to remove the version from their names. Files that include `#` are [source code files](https://github.com/foundry-rs/starknet-foundry/pull/3400/files/5df1d87dc762b0cf8d4df8bf047f66c2e8dc4352#diff-0d5658b415099a82c11c03a06ca4ec765b4003a1f4b2f3f1943980a882cf8aa6R153-R154) and do not need to be renamed.

.github/workflows/nightly.yml
ddoktorski: I didn't go into much detail but doing for example
```
echo "starknet-foundry-v0.44.0+nightly-2025-05-22-aarch64-apple-darwin.tar.gz" | sed "s/v[^+]*\+//"
```
without `-E` fails

.github/workflows/nightly.yml
cptartur: If there are only two source files that need to be handled separately, maybe it'd be more readable if we had two different loops rather than adding this if?

cptartur: It'd be a slight duplication of code but I think it would be clearer to understand

ddoktorski: Done

cptartur (approved):

.github/workflows/nightly.yml
cptartur: Okay, doesn't matter that much, we can keep it

Issue #3174, opened 2025-04-01
Bump blockifier to v0.14.0


ksew1:
### Which component is your feature related to?

Other (Please describe below)

### Current State

Tracking issue

### Objective

Bump blockifier

PR #3416, opened 2025-05-28
Bump blockifier to 0.14.0

franciszekjob:
Towards #3174

**Stack**:
- #3417
- #3416 ⬅

## Introduced changes

Bump blockifier to v0.14.0

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/cheatnet/src/runtime_extensions/common.rs
```diff
@@ -38,7 +40,11 @@ pub fn get_syscalls_gas_consumed(
             let base_cost = if let SyscallSelector::Deploy = selector {
                 syscall_gas_cost.get_syscall_cost(0)
             } else {
-                syscall_gas_cost.base_syscall_cost()
+                versioned_constants
+                    .os_constants
+                    .gas_costs
+                    .base
+                    .syscall_base_gas_cost
```
ddoktorski: Why was it changed?

franciszekjob: It's not needed, leftover from initial adjustments. Restored.

crates/forge-runner/src/build_trace_data.rs
```diff
@@ -123,7 +121,7 @@ pub fn build_profiler_execution_resources(
 ) -> ProfilerExecutionResources {
     let mut profiler_syscall_counter = HashMap::new();
     for (key, val) in syscall_usage {
-        profiler_syscall_counter.insert(build_profiler_deprecated_syscall_selector(key), val);
+        profiler_syscall_counter.insert(key, val);
```
ddoktorski: Actually do we need `profiler_syscall_counter` at all?

crates/data-transformer/src/shared/path.rs
```diff
@@ -10,6 +10,8 @@ pub enum GenericArgsExtractionError {
     InvalidGenericArgs,
     #[error("Expected exactly one generic argument")]
     MoreThanOneGenericArg,
+    #[error("Path segment missing")]
+    PathSegmentMissing,
```
ddoktorski, line 14: Could you give some context why this error needs to be added now?

crates/cheatnet/src/runtime_extensions/common.rs
```diff
@@ -29,7 +28,10 @@ pub fn get_syscalls_gas_consumed(
     syscall_usage
         .iter()
         .map(|(selector, usage)| {
-            let syscall_gas_cost = versioned_constants.get_syscall_gas_cost(selector);
+            let syscall_gas_costs = &versioned_constants.os_constants.gas_costs.syscalls;
+            let syscall_gas_cost = syscall_gas_costs
+                .get_syscall_gas_cost(selector)
+                .expect("Failed to get syscall gas cost");
```
kkawula: How do we prevent it from failing?

crates/forge/tests/integration/gas.rs
```diff
@@ -419,15 +419,15 @@ fn contract_pedersen_cost_cairo_steps() {
 
     assert_passed(&result);
     // 96 = cost of deploy (see snforge_std_deploy_cost test)
-    // 7 = cost of 86 pedersen builtins (because int(0.08 * 86) = 7)
-    // 0 l1_gas + 96 l1_data_gas + 7 * (100 / 0.0025) l2 gas
+    // 10 = cost of 86 pedersen and 62 range check builtins (because int(0.08 * 86 + 0.04 * 62) = 10)
+    // 0 l1_gas + 96 l1_data_gas + 10 * (100 / 0.0025) l2 gas
```
ddoktorski: This appears to be incorrect, the logic for gas calculation can be a bit unintuitive in this case. It doesn't sum the builtin usage, but instead it only considers the usage of the most expensive builtin or steps.

In this situation it seems that the number of steps increased and exceeded the cost of pedersen operations (steps: `ceil(2856 * 0.0025) = 8` vs pedersen: `ceil(86 * 0.08) = 7`). To make this test behave as it did before, you’ll need to increase the number of pedersen calls in the `GasChecker` contract.

For reference you can take a look at:
https://github.com/starkware-libs/sequencer/blob/main-v0.14.0/crates/blockifier/src/fee/fee_utils.rs#L63-L110
https://docs.starknet.io/architecture-and-concepts/network-architecture/fee-mechanism/#overall_fee (note the max in the formula)

crates/cheatnet/src/runtime_extensions/common.rs
franciszekjob: I don't understand your question - `get_syscall_gas_cost` returns `Result` so there's no other way than to use `expect` 😅

crates/forge-runner/src/build_trace_data.rs
franciszekjob: Right, we don't need it, removed.

crates/data-transformer/src/shared/path.rs
franciszekjob: [`split`](https://github.com/foundry-rs/starknet-foundry/blob/7d1c6a199b4f7e4c337165a9d130701be25815cb/crates/data-transformer/src/shared/path.rs#L36) function returns `Result<SplitResult, GenericArgsExtractionError>` and when matching `p` (which is of type `PathSegment`), we need to handle `Missing` variant. But looking at it now, it seems that `GenericArgsExtractionError` should be renamed (e.g. to `PathSplitError` ?) so it has a broader scope. Wdyt?

crates/cheatnet/src/forking/state.rs
```diff
@@ -68,7 +68,7 @@ fn other_provider_error<T>(boxed: impl ToString) -> Result<T, StateError> {
     let err_str = boxed.to_string();
 
     Err(StateReadError(
-        if err_str.contains("error sending request for url") {
+        if err_str.contains("error sending request") {
```
franciszekjob: note: Seems that error message was reduced, with the old `error sending request for url`, test `test_calling_nonexistent_url ` is failing. See https://github.com/foundry-rs/starknet-foundry/actions/runs/15439759021/job/43454570622?pr=3416

cptartur: That's weird, these errors originate in starknet rs I think or somewhere related and we didn't bump those.

crates/cheatnet/src/runtime_extensions/common.rs
cptartur: Seems that a bunch of syscalls are deprecated and will fail if trying to use them
https://github.com/starkware-libs/sequencer/blob/b4d231909527de0949f8422c1889cbdce3b0790b/crates/blockifier/src/blockifier_versioned_constants.rs#L879

It seems that methods from `corelib::starknet::` don't utilized them (I've only checked `get_caller_address` though) so it should be good but we should make sure no tests on our side fail due to syscall problems.

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/cheatcodes/cheat_block_hash.rs
```diff
@@ -81,7 +81,7 @@ impl CheatnetState {
         &mut self,
         contract_address: ContractAddress,
         block_number: u64,
```
cptartur, line 83: Why does it need to be mutable now?

crates/data-transformer/src/shared/path.rs
cptartur: Do we handle these errors in anyway or just bubble them to the user eventually?

Renaming seems fine, segment path being invalid seems non-recoverable from data transformer side so I'd just bubble these.

cc @ksew1 can you take a look?

crates/forge-runner/src/running/setup.rs
```diff
@@ -40,7 +40,19 @@ pub fn initialize_execution_context<'a>(
     // Instantiate Cairo runner.
     let proof_mode = false;
     let trace_enabled = true;
-    let mut runner = CairoRunner::new(program, LayoutName::all_cairo, proof_mode, trace_enabled)?;
+    let mut runner = CairoRunner::new(
+        program,
+        LayoutName::all_cairo,
+        // The `dynamic_layout_params` argument should only be used with dynamic layout.
+        // It is ignored otherwise.
+        // See: https://github.com/lambdaclass/cairo-vm/blob/4f02f5aab30f9062c2db1f5addbf28d34843168e/vm/src/vm/runners/cairo_runner.rs#L171
```
cptartur: Don't think we need to be copying these comments

crates/forge-runner/src/running/setup.rs
```diff
@@ -40,7 +40,19 @@ pub fn initialize_execution_context<'a>(
...
+        None,
+        proof_mode,
+        trace_enabled,
+        // By default, padding should be enabled
+        // See: https://github.com/lambdaclass/cairo-vm/blob/e339ab7cea2794428cff791a84c45841b84ca854/vm/src/cairo_run.rs#L37-L44
+        false,
```
cptartur: This should be "extracted" to a variable above, like `proof_mode` and `trace_enabled`.

cptartur: https://github.com/lambdaclass/cairo-vm/blob/4f02f5aab30f9062c2db1f5addbf28d34843168e/vm/src/vm/runners/cairo_runner.rs#L233

Seems that it can only be true in `proof_mode` which is false for us so it's good. 
But it could be worth verifying what these paddings are doing to the trace beyond the comment.

crates/forge/tests/data/contracts/gas_checker.cairo
```diff
@@ -262,6 +262,45 @@ mod GasChecker {
...
+            core::pedersen::pedersen(1, 2);
+            core::pedersen::pedersen(1, 2);
+            core::pedersen::pedersen(1, 2);
+            core::pedersen::pedersen(1, 2);
+            core::pedersen::pedersen(1, 2);
+            core::pedersen::pedersen(1, 2);
+            core::pedersen::pedersen(1, 2);
+            core::pedersen::pedersen(1, 2);
+            core::pedersen::pedersen(1, 2);
+            core::pedersen::pedersen(1, 2);
```
cptartur: Why these additions?

cptartur (commented):
Seems like mostly harmless bump this time

crates/cheatnet/src/runtime_extensions/common.rs
kkawula: That was what I meant, Thanks

crates/cheatnet/src/forking/state.rs
franciszekjob: > That's weird, these errors originate in starknet rs I think or somewhere related and we didn't bump those.

I don't think it's from starknet-rs - `StateReadError` comes from blockifier, which doesn't use starknet-rs. It's probably from one of blockifier's request/network-related dependencies and the error string got shortened (?). Not sure if it's worth deeper investigation.

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/cheatcodes/cheat_block_hash.rs
franciszekjob: Yes, below we call

```rust
syscall_handler.base.get_block_hash(block_number)?,
```

and [`get_block_hash`](https://github.com/starkware-libs/sequencer/blob/e4fc6e4013b10055c45827ea2d8c6ac055c2e511/crates/blockifier/src/execution/syscalls/syscall_base.rs#L120) needs mutable reference.

crates/forge/tests/data/contracts/gas_checker.cairo
franciszekjob: See https://github.com/foundry-rs/starknet-foundry/pull/3416#discussion_r2126387619

crates/cheatnet/src/runtime_extensions/common.rs
franciszekjob: Ok I see - I don't think we can do it. But for better experience while potentially facing this error, maybe `unwrap_or_else(|e| panic!("Failed to get syscall gas cost: {e}"))`, wdyt?

crates/data-transformer/src/shared/path.rs
ksew1: Yep, i would just bubble it and rename the enum

crates/cheatnet/src/runtime_extensions/common.rs
kkawula: Looks good, but I would be great to test it using Ma'at with nightly fonudry build

crates/forge-runner/src/build_trace_data.rs
```diff
@@ -118,13 +113,8 @@ fn build_profiler_call_trace_node(
 #[must_use]
 pub fn build_profiler_execution_resources(
     execution_resources: ExecutionResources,
-    syscall_usage: SyscallUsageMap,
     gas_consumed: u64,
 ) -> ProfilerExecutionResources {
-    let mut profiler_syscall_counter = HashMap::new();
-    for (key, val) in syscall_usage {
-        profiler_syscall_counter.insert(build_profiler_deprecated_syscall_selector(key), val);
-    }
```
kkawula: @piotmag769 wdyt?

crates/forge-runner/src/build_trace_data.rs
```diff
@@ -206,80 +196,6 @@ fn build_profiler_entry_point_type(value: EntryPointType) -> ProfilerEntryPointT
...
-        DeprecatedSyscallSelector::DelegateL1Handler => {
-            ProfilerDeprecatedSyscallSelector::DelegateL1Handler
-        }
-        DeprecatedSyscallSelector::Deploy => ProfilerDeprecatedSyscallSelector::Deploy,
-        DeprecatedSyscallSelector::EmitEvent => ProfilerDeprecatedSyscallSelector::EmitEvent,
-        DeprecatedSyscallSelector::GetBlockHash => ProfilerDeprecatedSyscallSelector::GetBlockHash,
-
-        DeprecatedSyscallSelector::GetBlockNumber => {
-            ProfilerDeprecatedSyscallSelector::GetBlockNumber
-        }
```
kkawula: Is it still required by profiler?  @piotmag769

crates/sncast/Cargo.toml
```diff
@@ -33,8 +33,8 @@ data-transformer = { path = "../data-transformer" }
 configuration = { path = "../configuration" }
 shared = { path = "../shared" }
 forge_runner = { path = "../forge-runner" }
-cairo-lang-runner = "2.11.2"
-cairo-lang-runnable-utils = "2.11.2"
+cairo-lang-runner = "2.12.0-rc.0"
+cairo-lang-runnable-utils = "2.12.0-rc.0"
```
kkawula, line 38:
```suggestion
cairo-lang-runner.workspace = true
cairo-lang-runnable-utils.workspace = true
```

crates/forge-runner/src/build_trace_data.rs
piotmag769: Nope, but will be needed in https://github.com/software-mansion/cairo-profiler/issues/175 so imo mark with `#[expect(dead_code)]`

crates/sncast/Cargo.toml
ddoktorski: Afaik we don't want to include dependencies on `cairo-lang-runner` in the workspace config. There has been an effort to avoid relying on it from multiple sources. Eventually it should also be removed from the `sncast` crate.

crates/cheatnet/src/forking/state.rs
cptartur: We mostly throw these errors from `ForkStateReader`, I think it's unlikely it's coming from blockifier itself, it's not making any requests to the network.
It might be from reqwest or some other network related package maybe?

Cargo.toml
```diff
@@ -29,22 +29,22 @@ license = "MIT"
 license-file = "LICENSE"
 
 [workspace.dependencies]
-blockifier = { git = "https://github.com/software-mansion-labs/sequencer.git", branch = "main-v0.13.5-modified-sierra-dep", default-features = false, features = ["testing", "tracing"] }
+blockifier = { git = "https://github.com/starkware-libs/sequencer.git", branch = "main-v0.14.0", default-features = false, features = ["testing", "tracing"] }
```
cptartur: Supposedly there has been an RC crates release for blockifier but it's quite old at this point. We should probably still track a branch and monitor releases.

crates/cheatnet/src/runtime_extensions/common.rs
cptartur: Let's both improve the error and test it with maat eventually.

crates/data-transformer/src/shared/parsing.rs
```diff
@@ -18,6 +18,7 @@ pub fn parse_expression(source: &str, db: &SimpleParserDatabase) -> Result<Expr,
         content: source.to_string().into(),
         code_mappings: [].into(),
         kind: FileKind::Expr,
+        original_item_removed: false,
```
cptartur, line 21: Do we know what it does? (not super critical)

crates/forge/tests/data/contracts/gas_checker.cairo
cptartur: Did you link the correct comment? How are deprecated syscalls related to this?

crates/sncast/Cargo.toml
cptartur: `cairo-lang-runner` must not escape the containment of sncast 😅 
We'll remove it entirely eventually but it's a bigger lift for now.

cptartur (commented):
Looks good

crates/forge-runner/src/build_trace_data.rs
```diff
@@ -206,19 +199,21 @@ fn build_profiler_entry_point_type(value: EntryPointType) -> ProfilerEntryPointT
...
+    value: SyscallSelector,
 ) -> ProfilerDeprecatedSyscallSelector {
     match value {
         DeprecatedSyscallSelector::CallContract => ProfilerDeprecatedSyscallSelector::CallContract,
         DeprecatedSyscallSelector::DelegateCall => ProfilerDeprecatedSyscallSelector::DelegateCall,
         DeprecatedSyscallSelector::DelegateL1Handler => {
             ProfilerDeprecatedSyscallSelector::DelegateL1Handler
         }
-        DeprecatedSyscallSelector::Deploy => ProfilerDeprecatedSyscallSelector::Deploy,
+        DeprecatedSyscallSelector::Deploy | DeprecatedSyscallSelector::MetaTxV0 => {
```
ddoktorski: Maybe let's add a separate arm for `MetaTxV0` with a TODO comment?

crates/forge/tests/integration/gas.rs
```diff
@@ -1087,7 +1087,7 @@ fn deploy_syscall_cost_sierra_gas() {
         GasVector {
             l1_gas: GasAmount(0),
             l1_data_gas: GasAmount(96),
-            l2_gas: GasAmount(639_664),
+            l2_gas: GasAmount(643_974),
```
ddoktorski: Comment with calculations above not updated

crates/forge/tests/integration/gas.rs
```diff
@@ -1120,15 +1120,15 @@ fn snforge_std_deploy_cost_sierra_gas() {
     // 96 = gas cost of onchain data (see `deploy_syscall_cost_sierra_gas` test)
     // 20000 = cost of 2 keccak syscall (see `deploy_syscall_cost_sierra_gas` test)
     // 147660 = cost of 1 deploy syscall (see `deploy_syscall_cost_sierra_gas` test)
-    // 491944 = reported consumed sierra gas
-    // 0 l1_gas + 96 l1_data_gas + (20000 + 147660 + 491944) l2 gas
+    // 496158 = reported consumed sierra gas
+    // 0 l1_gas + 96 l1_data_gas + (20000 + 147660 + 496158) l2 gas
```
ddoktorski: It doesn't sum up to `663_914`

crates/forge/tests/integration/gas.rs
```diff
@@ -1200,15 +1200,15 @@ fn contract_keccak_cost_sierra_gas() {
     // 87650 = cost of 1 call contract syscall (because 1 * 866 * 100 + 15 * 70)
     //      -> 1 call contract syscall costs 866 cairo steps and 15 range check builtins
     //      -> 1 range check costs 70
-    // 1160975 = reported consumed sierra gas
-    // 0 l1_gas + 96 l1_data_gas + (142810 + 50000 + 87650 + 1160975) l2 gas
+    // 1169099 = reported consumed sierra gas
+    // 0 l1_gas + 96 l1_data_gas + (142810 + 50000 + 87650 + 1169099) l2 gas
```
ddoktorski: It doesn't add up either, please go through all the modified tests to double check the correctness of the values 🙏

crates/cheatnet/src/forking/state.rs
franciszekjob: > It might be from reqwest or some other network related package maybe?

That's what I've mentioned 😄  _It's probably from one of blockifier's request/network-related dependencies_

crates/forge/tests/data/contracts/gas_checker.cairo
franciszekjob: It's this one
image

Cargo.toml
franciszekjob: Yup, it's quite old.

ddoktorski (changes requested):

crates/data-transformer/src/shared/parsing.rs
franciszekjob: From what Karol said, it's for macros and LS
cc @ksew1

crates/forge/tests/integration/gas.rs
franciszekjob: Right, reported consumed sierra gas should be`496254`, thanks for catching 👁️ .

crates/forge/tests/integration/gas.rs
franciszekjob: reported consumed sierra gas should be `1169195`

crates/data-transformer/src/shared/parsing.rs
ksew1: Yep, this should be set to `false`

crates/forge-runner/src/build_trace_data.rs
franciszekjob: Actually, I've added it as a variant of `ProfilerDeprecatedSyscallSelector` in `cairo-aoontations`, so should be fine now.

Cargo.toml
cptartur: Let's upgrade it once there's something not rc released.

crates/forge/tests/data/contracts/gas_checker.cairo
cptartur: Right, make sense. Steps are more expensive without these much repetitions of `pedersen`?

cptartur (approved):
Didn't

crates/forge-runner/src/build_trace_data.rs
ksew1: Please don't deleted this it will be used in https://github.com/foundry-rs/starknet-foundry/pull/3472

crates/forge-runner/src/build_trace_data.rs
```diff
@@ -118,13 +116,8 @@ fn build_profiler_call_trace_node(
 #[must_use]
 pub fn build_profiler_execution_resources(
     execution_resources: ExecutionResources,
-    syscall_usage: SyscallUsageMap,
     gas_consumed: u64,
 ) -> ProfilerExecutionResources {
-    let mut profiler_syscall_counter = HashMap::new();
```
ksew1: Please don't delete this it will be used in https://github.com/foundry-rs/starknet-foundry/pull/3472

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
```diff
@@ -568,6 +565,71 @@ fn handle_declare_deploy_result<T: CairoSerialize>(
...
+        .gas_costs
+        .base
+        .memory_hole_gas_cost;
+    vm_resource_to_gas_amount(n_memory_holes, gas_per_memory_hole, "memory_holes")
+}
+
+// TODO: Should be taken from blockifier
+// Copied blockifier code
+// https://github.com/starkware-libs/sequencer/blob/5bc3098ebd04facfc3868fd09d818392748a81b4/crates/blockifier/src/bouncer.rs#L649
+fn vm_resources_to_sierra_gas(
```
cptartur: Why do we have to copy it? It seems to be publish now.

crates/forge/tests/integration/gas.rs
```diff
@@ -1714,167 +1714,15 @@ fn events_contract_cost_sierra_gas() {
...
             l1_gas: GasAmount(0),
             l1_data_gas: GasAmount(96),
-            l2_gas: GasAmount(1_240_510),
-        },
-    );
-}
-
-#[test]
-#[cfg_attr(not(feature = "scarb_since_2_10"), ignore)]
-fn nested_call_cost_sierra_gas() {
```
cptartur: Why was these removed?

cptartur (changes requested):

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
franciszekjob: Wasn't present at the time when I was commiting, now it is. Changed.

crates/forge/tests/integration/gas.rs
franciszekjob: Restored, deleted by accident while rebasing 🥶

cptartur (approved):

Cargo.toml
```diff
@@ -29,22 +29,22 @@ license = "MIT"
 license-file = "LICENSE"
 
 [workspace.dependencies]
-blockifier = { git = "https://github.com/starkware-libs/sequencer.git", branch = "main-v0.13.6", default-features = false, features = ["testing", "tracing"] }
```
franciszekjob, line 32: note: `testing` is needed for cloning some structures, e.g. https://github.com/starkware-libs/sequencer/blob/1fdedd1d82efff086382db12f887defb1e4e2cdf/crates/blockifier/src/execution/call_info.rs#L34
`tracing` is needed for [`ExecutionRunnerMode::Tracing`](https://github.com/starkware-libs/sequencer/blob/1fdedd1d82efff086382db12f887defb1e4e2cdf/crates/blockifier/src/execution/entry_point_execution.rs#L58-l59)

cptartur (approved):

Cargo.toml
```diff
@@ -29,22 +29,22 @@ license = "MIT"
...
+cairo-lang-starknet = "2.12.0-rc.0"
+cairo-lang-filesystem = "2.12.0-rc.0"
+cairo-lang-diagnostics = "2.12.0-rc.0"
+cairo-lang-sierra-type-size = "2.12.0-rc.0"
+cairo-lang-syntax = "2.12.0-rc.0"
+cairo-lang-test-plugin = "2.12.0-rc.0"
+cairo-lang-starknet-classes = "2.12.0-rc.0"
+cairo-lang-parser = "2.12.0-rc.0"
+cairo-vm = "2.2.0"
+cairo-annotations = { git = "https://github.com/software-mansion/cairo-annotations.git", branch = "bump-cairo-to-2.12" }
```
ddoktorski: Why is it not `0.4.0`?

crates/runtime/src/starknet/context.rs
```diff
@@ -43,8 +42,10 @@ pub fn build_block_context(block_info: &BlockInfo, chain_id: Option<ChainId>) ->
         ChainInfo {
             chain_id: chain_id.unwrap_or_else(default_chain_id),
             fee_token_addresses: FeeTokenAddresses {
-                strk_fee_token_address: contract_address!(ERC20_CONTRACT_ADDRESS),
-                eth_fee_token_address: contract_address!(ERC20_CONTRACT_ADDRESS),
+                strk_fee_token_address: ContractAddress::try_from_hex_str(ERC20_CONTRACT_ADDRESS)
+                    .unwrap(),
+                eth_fee_token_address: ContractAddress::try_from_hex_str(ERC20_CONTRACT_ADDRESS)
+                    .unwrap(),
```
ddoktorski: Not related to this PR, but what is actually this setting and why does it use the same value for both tokens?

Cargo.toml
cptartur: Maybe let's push an RC or normal release of cairo-annotations so we can use it directly?

franciszekjob: > Why is it not 0.4.0?

0.4.0 doesn't use Cairo 2.12

> Maybe let's push an RC or normal release of cairo-annotations so we can use it directly?

Yup, cc @ksew1

crates/runtime/src/starknet/context.rs
franciszekjob: Based on field names, it should obviously set addresses for fee tokens.

> why does it use the same value for both tokens

Tbh idk xd I've changed it so it's correct now, not sure why it has been like that.

Cargo.toml
ksew1: The problem is that we use `rc` cairo deps. I will realese `0.5.0-rc.1` as it depends on `rc` version of cairo

franciszekjob: Yeah rc is completely fine 👍

ksew1: `0.5.0-rc.1` just landed on crates.io 😄

franciszekjob: Thanks!

ddoktorski (approved):

PR #3438, opened 2025-06-04
Use devnet from .tool-versoins

### LINKED ISSUE #3440: Unify devnet config

kkawula:
Closes #3440

We will lose possiblity to install devnet from ref, but probably we can get along with them to release rc if needed

## Introduced changes

-

## Checklist

- [ ] Linked relevant issue
- [ ] Added relevant tests


.github/workflows/ci.yml
```diff
@@ -264,22 +264,7 @@ jobs:
...
-        uses: actions/cache@v4
-        with:
-          path: ${{ github.workspace }}\crates\sncast\tests\utils\devnet
-          key: ${{ runner.os }}-devnet-${{ env.DEVNET_REV }}
-      - name: Install devnet
-        if: runner.os == 'Windows' && steps.windows-devnet-cache.outputs.cache-hit != 'true'
-        run: |
-          $DEVNET_INSTALL_DIR = "${{ github.workspace }}\crates\sncast\tests\utils\devnet"
-          cargo install --git https://github.com/0xSpaceShard/starknet-devnet.git --locked --rev ${{ env.DEVNET_REV }} --root $DEVNET_INSTALL_DIR
+      - uses: asdf-vm/actions/install@05e0d2ed97b598bfce82fd30daf324ae0c4570e6
```
cptartur, line 267: This breaks installation on Windows, which is fine because we don't care about it anymore. But please first create an PR that removes any logic for running tests on windows and building binaries for it.

You can create a separate task for that and assign yourself to it.

.tool-versions
```diff
@@ -1 +1,2 @@
 scarb 2.9.4
+starknet-devnet 0.4.2
```
cptartur, line 2: I like this approach however it becomes problematic when we want to use devnet from a branch or other unreleased source. Is there some override mechanism in asdf (I don't think so)?

We should communicate with shardlabs on slack and check if they will be fine with pushing pre-releases from WIP branches in case we need them.

kkawula:
Closing as duplicate #3455

Issue #3355, opened 2025-05-18
Running multiple tests with `trace-verbosity` results in a panic


amanusk:
### Select Tool

snforge

### Foundry Version

0.43.0

### What operating system are you using?

Linux

### What system architecture are you using?

x86

### Issue Description

Running multiple tests with `--trace-verbosity detailed`, results in an error

Repo:
https://github.com/amanusk/starknet-foundry-template

Error:
```thread 'tokio-runtime-worker' panicked at crates/debugging/src/trace/collect.rs:106:14:
contract name should be present in `ContractsData`
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
[ERROR] task 17 panicked with message "contract name should be present in `ContractsData`"
```

Renning without traces doesn't result in an error

Backtrace
```
thread 'tokio-runtime-worker' panicked at crates/debugging/src/trace/collect.rs:106:14:
contract name should be present in `ContractsData`
stack backtrace:
   0: rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::option::expect_failed
   3: core::ops::function::impls::<impl core::ops::function::FnOnce<A> for &mut F>::call_once
   4: <alloc::vec::Vec<T> as alloc::vec::spec_from_iter::SpecFromIter<T,I>>::from_iter
   5: debugging::trace::types::Trace::new
   6: forge_runner::test_case_summary::TestCaseSummary<forge_runner::test_case_summary::Single>::from_run_completed
   7: forge_runner::running::extract_test_case_summary
   8: forge_runner::running::run_test::{{closure}}
   9: tokio::runtime::task::core::Core<T,S>::poll
  10: tokio::runtime::task::harness::Harness<T,S>::poll
  11: tokio::runtime::blocking::pool::Inner::run
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
[ERROR] task 17 panicked with message "contract name should be present in `ContractsData`"
```

PR #3446, opened 2025-06-08
Add sync client

ksew1:
Towards #3355

commit-id:7816c922

---

**Stack**:
- #3451
- #3449
- #3448
- #3447
- #3446 ⬅

### DIFF
--- a/crates/cheatnet/src/forking/state.rs
+++ b/crates/cheatnet/src/forking/state.rs
@@ -1,5 +1,6 @@
 use crate::forking::cache::ForkCache;
 use crate::state::BlockInfoReader;
+use crate::sync_client::SyncClient;
 use anyhow::{Context, Result};
 use blockifier::execution::contract_class::{
     CompiledClassV0, CompiledClassV0Inner, CompiledClassV1, RunnableCompiledClass,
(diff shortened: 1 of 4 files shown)


crates/shared/src/sync_client.rs
```diff
@@ -0,0 +1,59 @@
...
+use starknet::providers::{JsonRpcClient, Provider, ProviderError};
+use starknet_api::block::BlockNumber;
+use starknet_types_core::felt::Felt;
+use tokio::runtime::Runtime;
+use url::Url;
+
+#[derive(Debug)]
+pub struct SyncClient {
+    client: JsonRpcClient<HttpTransport>,
+    block_id: BlockId,
```
franciszekjob: What's the idea behind holding block id here? Why isn't it a param in respective methods?

crates/shared/src/sync_client.rs
franciszekjob, line 1: shared package is an anti-pattern so I think we should omit to enlarge it 😅

ksew1: Where would you put this then?

crates/shared/src/sync_client.rs
ksew1: Throughout our codebase, we have always used a fixed block ID, as it is set in the ForkConfig.

crates/shared/src/sync_client.rs
franciszekjob: cheatnet?

ksew1: Yeah, we can put this in cheatnet and make it public

franciszekjob: Let's do that then

crates/shared/src/sync_client.rs
franciszekjob: Ok I see, let's leave it like that then

ddoktorski (approved):

franciszekjob (approved):

PR #3447, opened 2025-06-08
Add debugging `ContractData`

ksew1:
Towards #3355

commit-id:8a2d36a8

---

**Stack**:
- #3451
- #3449
- #3448
- #3447 ⬅

### DIFF
--- a/crates/debugging/src/contracts_data.rs
+++ b/crates/debugging/src/contracts_data.rs
@@ -1,0 +1,67 @@
+use crate::trace::types::{ContractName, Selector};
+use cheatnet::runtime_extensions::forge_runtime_extension::contracts_data::ContractsData as CheatnetContractsData;
+use rayon::iter::IntoParallelRefIterator;
+use rayon::iter::ParallelIterator;
+use starknet::core::types::contract::{AbiEntry, SierraClass};
+use starknet_api::core::{ClassHash, EntryPointSelector};
+use std::collections::HashMap;
+
+/// Data structure containing information about contracts,
+/// including their ABI, names, and selectors that will be used to create a [`Trace`](crate::Trace).
+pub struct ContractsData {
+    abi: HashMap<ClassHash, Vec<AbiEntry>>,
+    contract_names: HashMap<ClassHash, ContractName>,
+    selectors: HashMap<EntryPointSelector, Selector>,
+}
+
+impl ContractsData {
+    /// Creates a new instance of [`ContractsData`] from a similar structure from `cheatnet`: [`CheatnetContractsData`].
+    #[must_use]
+    pub fn new(cheatnet_contracts_data: &CheatnetContractsData) -> Self {
+        let contract_names = cheatnet_contracts_data
+            .class_hashes
+            .iter()
+            .map(|(name, class_hash)| (*class_hash, ContractName(name.clone())))
+            .collect();
+
+        let selectors = cheatnet_contracts_data
+            .selectors
+            .iter()
+            .map(|(selector, function_name)| (*selector, Selector(function_name.clone())))
+            .collect();
+
+        let abi = cheatnet_contracts_data
+            .contracts
+            .par_iter()
+            .map(|(_, contract_data)| {
+                let sierra = serde_json::from_str::<SierraClass>(&contract_data.artifacts.sierra)
+                    .expect("this should be valid `SierraClass`");
+                (contract_data.class_hash, sierra.abi)
+            })
+            .collect();
+
+        Self {
+            abi,
+            contract_names,
+            selectors,
+        }
+    }
+
+    /// Gets the [`ContractName`] for a given contract [`ClassHash`].
+    #[must_use]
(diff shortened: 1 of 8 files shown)


crates/debugging/src/contracts_data.rs
```diff
@@ -0,0 +1,67 @@
...
+use starknet::core::types::contract::{AbiEntry, SierraClass};
+use starknet_api::core::{ClassHash, EntryPointSelector};
+use std::collections::HashMap;
+
+/// Data structure containing information about contracts,
+/// including their ABI, names, and selectors that will be used to create a [`Trace`](crate::Trace).
+pub struct ContractsData {
+    abi: HashMap<ClassHash, Vec<AbiEntry>>,
+    contract_name: HashMap<ClassHash, ContractName>,
+    selectors: HashMap<EntryPointSelector, Selector>,
```
franciszekjob: These names are mix of singular and plural, let's unify it (most likely plural).

crates/debugging/src/contracts_data.rs
```diff
@@ -0,0 +1,67 @@
...
+}
+
+impl ContractsData {
+    /// Creates a new instance of [`ContractsData`] from a similar structure from `cheatnet`: [`CheatnetContractsData`].
+    #[must_use]
+    pub fn new(cheatnet_contracts_data: &CheatnetContractsData) -> Self {
+        let contract_name = cheatnet_contracts_data
```
franciszekjob: Same remark as above

crates/debugging/src/contracts_data.rs
```diff
@@ -0,0 +1,67 @@
...
+            .iter()
+            .map(|(selector, function_name)| (*selector, Selector(function_name.clone())))
+            .collect();
+
+        let abi = cheatnet_contracts_data
+            .contracts
+            .par_iter()
+            .map(|(_, contract_data)| {
+                let sierra = serde_json::from_str::<SierraClass>(&contract_data.artifacts.sierra)
+                    .expect("this should be valid SierraClass");
```
franciszekjob: nit:
```suggestion
                    .expect("this should be valid `SierraClass`");
```

crates/debugging/src/contracts_data.rs
ksew1: Went with plural except `abi`, because `abis`seems weird

crates/debugging/src/contracts_data.rs
```diff
@@ -0,0 +1,67 @@
...
+        }
+    }
+
+    /// Gets the [`ContractName`] for a given contract [`ClassHash`].
+    #[must_use]
+    pub fn get_contract_name(&self, class_hash: &ClassHash) -> Option<&ContractName> {
+        self.contract_names.get(class_hash)
+    }
+
+    /// Gets the `abi` for a given contract [`ClassHash`].
```
franciszekjob, line 56:
```suggestion
    /// Gets the [`AbiEntry`] for a given contract [`ClassHash`].
```

franciszekjob (approved):

crates/debugging/src/contracts_data.rs
ksew1: This is actually `&[AbiEntry]` but this won't work as a link so that's why i made this  `abi`

crates/debugging/src/lib.rs
```diff
@@ -1,11 +1,13 @@
 //! Crate with debugging utilities in forge.
 //!
-//! Currently, the only option that this crate gives is displaying pretty traces.
+//! Currently, the main purpose of this carate is displaying pretty traces.
```
ddoktorski:
```suggestion
//! Currently, the main purpose of this crate is displaying pretty traces.
```

crates/debugging/src/trace/collect.rs
```diff
@@ -159,6 +144,14 @@ impl<'a> Collector<'a> {
             },
         })
     }
+
+    fn class_hash(&self) -> &ClassHash {
+        self.call_trace
+            .entry_point
+            .class_hash
+            .as_ref()
+            .expect("class_hash should be set in `fn execute_call_entry_point` in cheatnet")
```
ddoktorski, line 153: This message doesn't seem to be fully accurate as `class_hash` is set in `execute_call_entry_point` after updating the call trace. But in general, is it ever possible for `class_hash` to be `None`?

ksew1: > This message doesn't seem to be fully accurate as class_hash is set in execute_call_entry_point after updating the call trace.

What do you mean by that?

> But in general, is it ever possible for class_hash to be None?

Afaik for mock calls it is not set

ddoktorski: > What do you mean by that?

Alright, I missed that part that it is indeed set there. I thought we only set this not directly in `enter_nested_call`, nevermind then.

ddoktorski (approved):

PR #3470, opened 2025-06-13
Runtime checks for `_std` packages

cptartur:
**Stack**:
- #3573
- #3559
- #3553
- #3503
- #3502
- #3470 ⬅
- #3466
- #3465
- #3464
- #3424
- #3421
- #3420
- #3419
- #3415

⚠️ *Part of a stack created by [spr](https://github.com/ejoffe/spr). Do not merge manually using the UI - doing so may have unexpected results.*

PR #3466, opened 2025-06-13
Use `snforge_std` or `snforge_std_compatibility` in `snforge` new

cptartur:
Closes #3461

---

**Stack**:
- #3573
- #3559
- #3553
- #3503
- #3502
- #3470
- #3466 ⬅
- #3465
- #3464
- #3424
- #3421
- #3420
- #3419
- #3415

⚠️ *Part of a stack created by [spr](https://github.com/ejoffe/spr). Do not merge manually using the UI - doing so may have unexpected results.*

PR #3465, opened 2025-06-13
Update release workflows to publish `_compatibility` plugin

cptartur:
Closes #3144

---

**Stack**:
- #3573
- #3559
- #3553
- #3503
- #3502
- #3470
- #3466
- #3465 ⬅
- #3464
- #3424
- #3421
- #3420
- #3419
- #3415

⚠️ *Part of a stack created by [spr](https://github.com/ejoffe/spr). Do not merge manually using the UI - doing so may have unexpected results.*

PR #3424, opened 2025-05-28
Add tests for diagnostics

cptartur:
**Stack**:
- #3573
- #3559
- #3553
- #3503
- #3502
- #3470
- #3466
- #3465
- #3464
- #3424 ⬅
- #3421
- #3420
- #3419
- #3415

⚠️ *Part of a stack created by [spr](https://github.com/ejoffe/spr). Do not merge manually using the UI - doing so may have unexpected results.*

PR #3471, opened 2025-06-13
Replacing `snforge_new` with `snforge_std_compatibility` in tests

cptartur:
**Stack**:
- #3471 ⬅
- #3470
- #3466
- #3465
- #3464
- #3424
- #3421
- #3420
- #3419
- #3415


cptartur:
Not sure about this solution. It replaces `snforge_std` with `snforge_std_compatibility` in runtime. It might not cover some cases. 

Alternative solution is to have scripts that replace file contents in CI and we run as always.

cptartur:
Important, **this PR causes most (or all) tests to use old plugin** (v1 syntax) so it should not be treated as a test reference of macros. See PRs before this one.

PR #3491, opened 2025-06-22
Get gas and vm resources directly from `CallInfo`

### LINKED ISSUE #3292: Get gas and vm resources directly from `CallInfo`
### Current State

Currently we calculate used resources and gas based on data we collect during execution, from various sources.

### Objective

Get gas and vm resources directly from `CallInfo`

### Additional Context

Except for some edge cases like artificial gas from deploy cheatcode it should be mostly straightforward to collect and use this data.

### LINKED ISSUE #3473: Resolve problem with duplicated resources for syscalls in nested calls
### Current State

```rust
...
...

fn xyz() {
...
...
}
```

### Objective

### Additional Context

_No response_

franciszekjob:
Closes #3473 
Closes #3292 

## Introduced changes

This PR changes the way we get gas and vm resources - now we get them directly from `CallInfo`.
At the same time, it fixes the bug with duplicated builtins from syscalls in nested calls.

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
```diff
@@ -691,102 +620,27 @@ pub fn update_top_call_vm_trace(runtime: &mut ForgeRuntime, cairo_runner: &mut C
...
-                &execution_resources,
-                &top_call_syscalls,
-            );
-        }
-        TrackedResource::SierraGas => {
-            sierra_gas_consumed +=
-                get_syscalls_gas_consumed(&top_call_syscalls, versioned_constants);
-        }
+    if tracked_resource == TrackedResource::SierraGas {
+        gas_consumed += get_syscalls_gas_consumed(&total_syscall_usage, versioned_constants);
```
franciszekjob, line 640: note: It seems that consumed gas returned in call info doesn't include the one spent on syscalls.
For example, `contract_keccak_cost_sierra_gas` reports:
- `67110` without added gas from syscalls (this is completely too low, as 1 deploy, 5 keccak and 1 call contract syscalls already cost `142810 + 50000 + 87650 = 280460`
- `347570` with added gas from syscalls

crates/forge/tests/integration/gas.rs
```diff
@@ -1075,7 +1074,7 @@ fn nested_call_cost_cairo_steps() {
         GasVector {
             l1_gas: GasAmount(0),
             l1_data_gas: GasAmount(288),
-            l2_gas: GasAmount(14_375_360),
+            l2_gas: GasAmount(7_215_360),
```
franciszekjob, line 1077: note: Here is "proof" for the #3473 fix - the value is reduced ~twice.

franciszekjob:
note: Comparison of resources before and after, in the package from #3473 
image

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
ddoktorski: Something is not right here and we need to understand exactly what is going on. I would suggest starting by confirming whether `vm_resources` and `gas_consumed` in the `CallInfo` struct are cumulative values that include the execution of inner calls or not.

ddoktorski (changes requested):

PR #3496, opened 2025-06-25
Fix duplicated syscall resources in nested calls

### LINKED ISSUE #3399: Exclude syscall resources when creating trace data to save
### Current State

Currently syscall resources are removed after each call: https://github.com/foundry-rs/starknet-foundry/blob/995e7f8db85736727006e943e2ed673c7687ef98/crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/entry_point.rs#L299

This entire logic can be moved to the function [build_profiler_call_trace](https://github.com/foundry-rs/starknet-foundry/blob/995e7f8db85736727006e943e2ed673c7687ef98/crates/forge-runner/src/build_trace_data.rs#L36).

### Objective

Exclude syscall resources when creating trace data to save

### Additional Context

* It will require changes in the [get_all_used_resources](https://github.com/foundry-rs/starknet-foundry/blob/995e7f8db85736727006e943e2ed673c7687ef98/crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs#L729) function
* It can be done as a follow up to #3292

### LINKED ISSUE #3473: Resolve problem with duplicated resources for syscalls in nested calls

franciszekjob:
Closes #3473
Closes #3399

## Introduced changes

This PR changes the way we subtract syscall resources. Before, we subtracted them in `execute_call_entry_point`. Now, we do it directly before passing to trace data.

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Forge
+
+#### Fixed
+
+- bug where syscall execution resources from nested calls were being calculated twice
+
 ### Cast
 
 #### Removed
--- a/crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
+++ b/crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
@@ -603,8 +603,7 @@ pub fn add_resources_to_top_call(
     }
 }
 
-pub fn update_top_call_resources(runtime: &mut ForgeRuntime) {
-    // call representing the test code
+pub fn calculate_total_syscall_usage(runtime: &mut ForgeRuntime) -> SyscallUsageMap {
     let top_call = runtime
         .extended_runtime
         .extended_runtime
@@ -614,13 +613,7 @@ pub fn update_top_call_resources(runtime: &mut ForgeRuntime) {
         .current_call_stack
         .top();
 
-    let all_execution_resources = add_execution_resources(top_call.clone());
-    let all_sierra_gas_consumed = add_sierra_gas_resources(&top_call);
-
-    let mut top_call = top_call.borrow_mut();
-    top_call.used_execution_resources = all_execution_resources;
-    top_call.gas_consumed = all_sierra_gas_consumed;
-
+    let top_call = top_call.borrow_mut();
     let top_call_syscalls = runtime
         .extended_runtime
         .extended_runtime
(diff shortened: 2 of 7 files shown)


franciszekjob:
note: Comparison of resources before (left) and after (right), in the package from #3473 

image

and trace

image

crates/forge-runner/src/build_trace_data.rs
```diff
@@ -33,7 +36,56 @@ pub const TRACE_DIR: &str = "snfoundry_trace";
 pub const TEST_CODE_CONTRACT_NAME: &str = "SNFORGE_TEST_CODE";
 pub const TEST_CODE_FUNCTION_NAME: &str = "SNFORGE_TEST_CODE_FUNCTION";
 
-pub fn build_profiler_call_trace(
+fn remove_syscall_resources_from_call_trace(
```
ddoktorski, line 39: This logic requires an adjustment, as there are scenarios where we track sierra gas, but some nested calls track VM resources instead (if contracts were compiled with Sierra <1.7.0). This situation can occur on forks. Perhaps we can take information about the tracked resources from `CallInfo`, save it in the call trace and then use it here.

ddoktorski (changes requested):

crates/forge-runner/src/build_trace_data.rs
franciszekjob: Yeah, you're right. Will update it.

CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Forge
+
+#### Fixed
+
+- bug where syscall execution resources from nested calls were being calculated twice
```
cptartur, line 14: Nit, maybe use uppercase here

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
```diff
@@ -629,6 +636,25 @@ pub fn update_top_call_resources(runtime: &mut ForgeRuntime) {
         .syscalls_usage
         .clone();
 
+    // Execution resources from nested calls syscalls were already added, so we only
+    // need to add resources from top call syscalls.
+    match tracked_resource {
```
cptartur, line 641: What about a situation where the test used mixed resources. Is converting them to the top level tracked resource when they are included valid compared to what is done on starknet?

crates/forge-runner/src/build_trace_data.rs
```diff
@@ -33,7 +36,56 @@ pub const TRACE_DIR: &str = "snfoundry_trace";
...
+
+pub fn build_profiler_call_trace_with_adjusted_resources(
+    value: &Rc<RefCell<CallTrace>>,
+    contracts_data: &ContractsData,
+    versioned_program_path: &Utf8Path,
+    tracked_resource: TrackedResource,
+) -> ProfilerCallTrace {
+    let versioned_constants = VersionedConstants::latest_constants();
+    remove_syscall_resources_from_call_trace(value, tracked_resource, versioned_constants);
+    build_profiler_call_trace(value, contracts_data, versioned_program_path)
```
cptartur, line 85: I think we can just inline `build_profiler_call_trace` here, there's not much benefit having it as a separate function.

crates/forge-runner/src/build_trace_data.rs
```diff
@@ -33,7 +36,56 @@ pub const TRACE_DIR: &str = "snfoundry_trace";
...
+                    tracked_resource,
+                    versioned_constants,
+                );
+            }
+            CallTraceNode::DeployWithoutConstructor => {}
+        }
+    }
+}
+
+pub fn build_profiler_call_trace_with_adjusted_resources(
```
cptartur, line 77: I'd just have this named `build_profiler_call_trace`, it's exposing (quite important, but still) implementation details to the caller.

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
franciszekjob: You're referring to same situation as @ddoktorski in https://github.com/foundry-rs/starknet-foundry/pull/3496#discussion_r2169375193 - current approach needs to be fixed, that's right.

cptartur: I think the logic I've commented on itself is fine (you can verify though). It's just that removing syscall resources needs to be fixed, as @ddoktorski mentioned.

PR #3421, opened 2025-05-28
Fix gas tests

cptartur:
Closes #3078

---

**Stack**:
- #3573
- #3559
- #3553
- #3503
- #3502
- #3470
- #3466
- #3465
- #3464
- #3424
- #3421 ⬅
- #3420
- #3419
- #3415

⚠️ *Part of a stack created by [spr](https://github.com/ejoffe/spr). Do not merge manually using the UI - doing so may have unexpected results.*

PR #3420, opened 2025-05-28
Disable scarb 2.10.0 specific tests

cptartur:
Closes #3096

---

**Stack**:
- #3573
- #3559
- #3553
- #3503
- #3502
- #3470
- #3466
- #3465
- #3464
- #3424
- #3421
- #3420 ⬅
- #3419
- #3415

⚠️ *Part of a stack created by [spr](https://github.com/ejoffe/spr). Do not merge manually using the UI - doing so may have unexpected results.*

PR #3415, opened 2025-05-28
Migrate to new macro syntax

cptartur:
**Stack**:
- #3573
- #3559
- #3553
- #3503
- #3502
- #3470
- #3466
- #3465
- #3464
- #3424
- #3421
- #3420
- #3419
- #3415 ⬅

⚠️ *Part of a stack created by [spr](https://github.com/ejoffe/spr). Do not merge manually using the UI - doing so may have unexpected results.*

PR #3502, opened 2025-06-27
Run tests on a new and compatibility plugin

cptartur:
**Stack**:
- #3573
- #3559
- #3553
- #3503
- #3502 ⬅
- #3470
- #3466
- #3465
- #3464
- #3424
- #3421
- #3420
- #3419
- #3415


.github/workflows/ci.yml
```diff
@@ -272,7 +284,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+        # Override scarb version because formatter from 2.12 changes `while` loops to not compile on older Scarbs
       - uses: software-mansion/setup-scarb@v1
+        with:
+          scarb-version: 2.10.1
```
ddoktorski: This will be annoying during local development, it will format with Scarb `2.12`, but CI will check formatting with version `2.10.1`. It would be better to simply omit std compatibility in this check

.tool-versions
```diff
@@ -1,2 +1,2 @@
-scarb nightly-2025-06-18
+scarb nightly-2025-06-18 2.10.1
```
ddoktorski: What happened here?

cptartur: It's on purpose, the new tests need both scarb 2.10.1 and 2.12 (pre-release). It turns out you can have multiple versions in `.tool-versions` and `asdf install` will install all of them (but use the first).

.github/workflows/ci.yml
cptartur: Will change that

franciszekjob (approved):

cptartur:

PR #3503, opened 2025-06-27
Remove `supports-panic-backtrace` test condition

cptartur:
**Stack**:
- #3573
- #3559
- #3553
- #3503 ⬅
- #3502
- #3470
- #3466
- #3465
- #3464
- #3424
- #3421
- #3420
- #3419
- #3415

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -206,20 +206,6 @@ jobs:
       - name: Run Forge Debugging tests
         run: cargo test --release -p forge --features debugging --test main e2e::debugging
 
-  test-forge-backtrace:
-    name: Test Forge Backtrace - support for `panic-backtrace` in Scarb.toml
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v4
-      - uses: dtolnay/rust-toolchain@stable
-      - uses: Swatinem/rust-cache@9d47c6ad4b02e050fd481d890b2ea34778fd09d6 # v2.7.8
-      - uses: software-mansion/setup-scarb@v1
-        with:
-          scarb-version: "nightly-2025-03-27"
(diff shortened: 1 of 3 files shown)


ksew1 (commented):
Would it be better to have feature for scarb `2.12.*` instead of running scarb to check version?

cptartur:
> Would it be better to have feature for scarb `2.12.*` instead of running scarb to check version?

We'll use Scarb 2.12 as a default version (after it is released), check for scarb version is just to enable compatibility with scheduled tests running on older scarb. After 2.14 the check will be obsolete.

ksew1:
> > Would it be better to have feature for scarb `2.12.*` instead of running scarb to check version?
> 
> We'll use Scarb 2.12 as a default version (after it is released), check for scarb version is just to enable compatibility with scheduled tests running on older scarb. After 2.14 the check will be obsolete.

So maybe let's add feature for that. I would say this is our standard procedure and when you delete it cargo automatically shows places, where you referenced this

cptartur:
> > > Would it be better to have feature for scarb `2.12.*` instead of running scarb to check version?
> > 
> > 
> > We'll use Scarb 2.12 as a default version (after it is released), check for scarb version is just to enable compatibility with scheduled tests running on older scarb. After 2.14 the check will be obsolete.
> 
> So maybe let's add feature for that. I would say this is our standard procedure and when you delete it cargo automatically shows places, where you referenced this

Since a lot of tests depends on runtime checks for scarb version (they need to pick either compatibility or standard package) I think it's easier to just do the same for backtrace.
Otherwise we still will have runtime checks for versions before 2.11 just not for backtrace.

ksew1 (approved):

cptartur:

PR #3504, opened 2025-06-28
Rename argent to ready

### LINKED ISSUE #3501: Rename Argent to Ready
### Current State

Argent was rebranded to Ready https://x.com/ready_co

### Objective

Update docs and code to reflect these changes

### Additional Context

The logic in cast has to change, ideally in a backwards compatible manner. Additionally, our documentation has to be updated. Please search for other places to change it as well.

franciszekjob:
Closes #3501 

## Introduced changes

Rename argent to ready, due to their rebranding.

Add `ready` option and deprecates existing `argent`. Fore sncast accounts json, we use `ready`, but still accept `argent` for backward compatibility. When saving account to starkli format, `ready` is mapped to `argent` (because starkli doesn't support `ready` yet).

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/sncast/tests/e2e/account/deploy.rs
```diff
@@ -23,7 +23,7 @@ use test_case::test_case;
 
 #[test_case(DEVNET_OZ_CLASS_HASH_CAIRO_0, "oz"; "cairo_0_class_hash")]
 #[test_case(&OZ_CLASS_HASH.into_hex_string(), "oz"; "cairo_1_class_hash")]
-#[test_case(&ARGENT_CLASS_HASH.into_hex_string(), "argent"; "argent_class_hash")]
+#[test_case(&READY_CLASS_HASH.into_hex_string(), "ready"; "READY_CLASS_HASH")]
```
ddoktorski: Also in other places
```suggestion
#[test_case(&READY_CLASS_HASH.into_hex_string(), "ready"; "ready_class_hash")]
```

crates/sncast/src/starknet_commands/account/create.rs
```diff
@@ -268,11 +268,11 @@ fn create_to_keystore(
                 }
             })
         }
-        AccountType::Argent => {
+        AccountType::Ready => {
             json!({
                 "version": 1,
                 "variant": {
-                    "type": AccountType::Argent,
+                    "type": AccountType::Ready,
```
ddoktorski: By doing it this way, we may potentially break compatibility with `starkli`. Could you confirm whether an account created this way would be usable in `starkli`? I guess it won't be possible to create an account with `sncast` and then deploy it using `starkli`, but that’s very unusal case, so the question is only about using a deployed account.

franciszekjob: > I guess it won't be possible to create an account with sncast and then deploy it using starkli

That's right but I think starkli should also apply changes in regards to the rebranding, so once they do it, this scenario will eventually work.

> so the question is only about using a deployed account.

Could you clarify what kind of scenario (I mean cast-starkli interaction) you mean?

ddoktorski: > Could you clarify what kind of scenario (I mean cast-starkli interaction) you mean?

Using account created by `sncast` in `starkli`

franciszekjob: Yes, then it won't be possible, but as I said - starkli should eventually also change it. We can deprecate `argent` option for `account create` and keep it until starkli applies the change. Wdyt?

ddoktorski: I mean the point is to have the starkli format align with the actual starkli format, so from a user facing perspective we can allow passing `ready` but save it as `argent` as long as `argent` remains the only value accepted by `starkli`

franciszekjob: > we can allow passing `ready` but save it as `argent` as long as `argent` remains the only value accepted by starkli

Right, generally it makes sense, but it will be a bit weird if someone uses `account list` and will get displayed `type: argent` even though our flag uses `ready`, don't you think so?
Imo deprecation + mentioning that the "new" accounts may not work with starkli is ok.
@cptartur ?

crates/sncast/src/lib.rs
```diff
@@ -65,7 +65,8 @@ impl FromStr for AccountType {
     fn from_str(s: &str) -> Result<Self, Self::Err> {
         match s {
             "open_zeppelin" | "open-zeppelin" | "oz" => Ok(AccountType::OpenZeppelin),
-            "argent" => Ok(AccountType::Argent),
+            // We still accept "argent" for backward compatibility
+            "argent" | "ready" => Ok(AccountType::Ready),
```
cptartur: I think we should show some deprecation warning if user is using `argent`

crates/sncast/src/starknet_commands/account/create.rs
cptartur: Aren't starkli (keystore) accounts saved separately from normal sncast accounts? Realistically, we don't know if and how starkli will upgrade their implementation, not sure if we should rely on that.

docs/src/appendix/sncast/account/import.md
```diff
@@ -17,7 +17,7 @@ Address of the account.
 ## `--type, -t <ACCOUNT_TYPE>`
 Required.
 
-Type of the account. Possible values: oz, argent, braavos.
+Type of the account. Possible values: `oz`, `ready`, `braavos`.
```
cptartur: Maybe it'd make sense to say `formerly bravos` here as well? And in the code `--help`.

crates/sncast/src/starknet_commands/account/create.rs
cptartur: Also, i assume these changes are still backward compatible with account files using `argent` if use created it using older version of sncast?

franciszekjob: > Also, i assume these changes are still backward compatible with account files using `argent` if use created it using older version of sncast?

Yes, when resolving account type, both `argent` and `ready` lead to `ready`.

crates/sncast/src/lib.rs
franciszekjob: Yes, that's what I've mentioned in https://github.com/foundry-rs/starknet-foundry/pull/3504#discussion_r2177543272

docs/src/appendix/sncast/account/import.md
franciszekjob: `formerly argent`, then yes 😆

crates/sncast/src/lib.rs
ddoktorski: > I think we should show some deprecation warning if user is using argent

In which scenarios would this apply? Since users shouldn't really modify account files manually, such deprecation warning seems somewhat pointless to me

franciszekjob: They should face warning when creating/importing account.

cptartur: I think the warning should show up if they pass these as command line arguments. Don't think we should be showing warnings about file contents.

franciszekjob: Yup, that's exactly what I mean.

ddoktorski: Ok I see now, that sounds good

0xEniotna:
Hello, Antoine from Ready here. 
I reached out to the Starkli guy, no answer yet... 
Let's say you merge as it is right now. Then starkli implements the name change. How big would be the work to make both tools "compatible" again as they were before the name change?

franciszekjob:
> Hello, Antoine from Ready here. I reached out to the Starkli guy, no answer yet... Let's say you merge as it is right now. 
Then starkli implements the name change. How big would be the work to make both tools "compatible" again as they were before the name change?

Hey @0xEniotna !
This change doesn't impact compatibility - when sncast saves account to stakrli format, it still uses `argent` under the hood. This will perhaphs in future, once starkli will support ready name as well. In all user-facing commands we use `ready`.

CHANGELOG.md
```diff
@@ -13,6 +13,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 
+### Cast
+
+#### Added
+
+- `--ready` flag to `account create` and `account import` commands (Argent wallet has rebranded as Ready)
+
+#### Deprecated
+
+- `--argent` flag in `account create` and `account import` commands. Use `--ready` instead
```
ddoktorski: These are not flags, but options to the flag `--type`

franciszekjob: Yup typo 😅

CHANGELOG.md
```diff
@@ -13,6 +13,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Passing a cheatcode span of 0 was incorrectly treated as `CheatSpan::Indefinite`. This is now resolved by making `CheatSpan::TargetCalls` accept `NonZero<usize>` instead of just `usize` in `snforge_std`.
 
+### Cast
+
+#### Added
+
+- `ready` option `--type` flag for `account create` and `account import` commands (Argent wallet has rebranded as Ready)
```
```suggestion
- `ready` option for `--type` flag for `account create` and `account import` commands (Argent wallet has rebranded as Ready)
```

CHANGELOG.md
```diff
@@ -13,6 +13,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+
+#### Deprecated
+
+- `argent` option for `--type` flag for `account create` and `account import` commands. Use `--ready` instead
```
```suggestion
- `argent` option for `--type` flag for `account create` and `account import` commands. Use `ready` instead
```

crates/sncast/src/starknet_commands/account/create.rs
```diff
@@ -63,7 +65,16 @@ pub async fn create(
     provider: &JsonRpcClient<HttpTransport>,
     chain_id: Felt,
     create: &Create,
+    ui: &UI,
 ) -> Result<AccountCreateResponse> {
+    // TODO(#3556): Remove this warning once we drop Argent account type
+    if create.account_type == AccountType::Argent {
+        ui.println(&WarningMessage::new(
+            "Argent has rebranded as Ready. The `--argent` option for `account create` is deprecated, please use `--ready` instead.",
```
ddoktorski:
```suggestion
            "Argent has rebranded as Ready. The `argent` option for the `--type` flag in `account create` is deprecated, please use `ready` instead.",
```

crates/sncast/src/starknet_commands/account/import.rs
```diff
@@ -68,7 +70,16 @@ pub async fn import(
     accounts_file: &Utf8PathBuf,
     provider: &JsonRpcClient<HttpTransport>,
     import: &Import,
+    ui: &UI,
 ) -> Result<AccountImportResponse> {
+    // TODO(#3556): Remove this warning once we drop Argent account type
+    if import.account_type == AccountType::Argent {
+        ui.println(&WarningMessage::new(
+                "Argent has rebranded as Ready. The `--argent` option for `account import` is deprecated, please use `--ready` instead.",
```
ddoktorski:
```suggestion
                "Argent has rebranded as Ready. The `argent` option for the `--type` flag in `account import` is deprecated, please use `ready` instead.",
```

docs/src/appendix/sncast/account/create.md
```diff
@@ -28,14 +28,21 @@ Possible values: `mainnet`, `sepolia`.
...
 Optional. Required if `--class-hash` is passed.
 
-Type of the account. Possible values: oz, argent, braavos. Defaults to oz.
+<!-- TODO(#3556): Remove `argent` option once we drop Argent account type. -->                              |
+Type of the account. Possible values: `oz`, `argent`, `ready`, `braavos`. Defaults to oz.
+
+<!-- TODO(#3556): Remove warning once we drop Argent account type. -->                              |
+> ⚠️ **Warning**
+>
+> Argent has rebranded as Ready. The `--argent` option is deprecated, please use `--ready` instead.
```
ddoktorski: The same comment as above

crates/sncast/src/starknet_commands/account/create.rs
```diff
@@ -268,11 +279,11 @@ fn create_to_keystore(
                 }
             })
         }
-        AccountType::Argent => {
+        AccountType::Argent | AccountType::Ready => {
             json!({
                 "version": 1,
                 "variant": {
-                    "type": AccountType::Argent,
+                    "type": format!("{account_type}"),
```
ddoktorski:
```suggestion
                    "type": "argent",
```
Maybe let's hardcode it here with the TODO comment and change the `fmt` implementation to always display as `ready`?

MKowalski8 (approved):

crates/sncast/src/starknet_commands/account/mod.rs
```diff
@@ -57,11 +57,16 @@ pub fn prepare_account_json(
     class_hash: Option<Felt>,
     salt: Option<Felt>,
 ) -> serde_json::Value {
+    // TODO(#3556): Once `Argent` variant is deleted, use `account_type` directly
+    let saved_account_type = match account_type {
+        AccountType::Ready => AccountType::Argent,
+        _ => account_type,
+    };
```
ddoktorski: As we're saving in our own format here, I think it makes sense to always use `ready`

ddoktorski (approved):

PR #3506, opened 2025-06-30
Fix resources from syscalls in nested calls

### LINKED ISSUE #3473: Resolve problem with duplicated resources for syscalls in nested calls

franciszekjob:
Closes #3473 

## Introduced changes

Fix bug calculation of resources from nested calls syscalls. The reason of malformed calculations was that during subtracting syscalls' resources in `remove_syscall_resources_and_exit_non_error_call`, they were only from current call and didn't include the ones from nested calls.

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests

### DIFF
--- a/crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/cairo1_execution.rs
+++ b/crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/cairo1_execution.rs
@@ -127,13 +127,10 @@ pub(crate) fn execute_entry_point_call_cairo1(
             .register_error(class_hash, pcs);
     }
 
-    let mut syscall_usage_vm_resources = SyscallUsageMap::default();
-    let mut syscall_usage_sierra_gas = SyscallUsageMap::default();
-
-    match tracked_resource {
-        TrackedResource::CairoSteps => syscall_usage_vm_resources.clone_from(&syscall_usage),
-        TrackedResource::SierraGas => syscall_usage_sierra_gas.clone_from(&syscall_usage),
-    }
+    let (syscall_usage_vm_resources, syscall_usage_sierra_gas) = match tracked_resource {
+        TrackedResource::CairoSteps => (syscall_usage, SyscallUsageMap::default()),
+        TrackedResource::SierraGas => (SyscallUsageMap::default(), syscall_usage),
+    };
 
     Ok(CallInfoWithExecutionData {
         call_info,
--- a/crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/cairo1_execution.rs
+++ b/crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/cairo1_execution.rs
@@ -5,12 +5,13 @@ use crate::runtime_extensions::call_to_blockifier_runtime_extension::execution::
 };
 use crate::runtime_extensions::cheatable_starknet_runtime_extension::CheatableStarknetRuntimeExtension;
 use crate::runtime_extensions::common::get_relocated_vm_trace;
-use blockifier::execution::contract_class::CompiledClassV1;
+use blockifier::execution::contract_class::{CompiledClassV1, TrackedResource};
 use blockifier::execution::entry_point::ExecutableCallEntryPoint;
 use blockifier::execution::entry_point_execution::{
     ExecutionRunnerMode, VmExecutionContext, finalize_execution,
     initialize_execution_context_with_runner_mode, prepare_call_arguments,
 };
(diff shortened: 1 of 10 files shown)


crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/cairo1_execution.rs
```diff
@@ -124,9 +127,18 @@ pub(crate) fn execute_entry_point_call_cairo1(
             .register_error(class_hash, pcs);
     }
 
+    let mut syscall_usage_vm_resources = SyscallUsageMap::default();
+    let mut syscall_usage_sierra_gas = SyscallUsageMap::default();
+
+    match tracked_resource {
+        TrackedResource::CairoSteps => syscall_usage_vm_resources.clone_from(&syscall_usage),
+        TrackedResource::SierraGas => syscall_usage_sierra_gas.clone_from(&syscall_usage),
+    }
```
cptartur: It could be rewritten so that variables don't have to be mutable

```rust
    let (syscall_usage_vm_resources, syscall_usage_sierra_gas) = match tracked_resource {
        TrackedResource::CairoSteps => (syscall_usage, SyscallUsageMap::default()),
        TrackedResource::SierraGas => (SyscallUsageMap::default(), syscall_usage),
    };
```

Though I'm not sure if that's actually better or not.

cptartur (approved):
Logic wise looks good, see me comment about possible refactoring

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/cairo1_execution.rs
franciszekjob: Applied 👍

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/deprecated/cairo0_execution.rs
```diff
@@ -80,9 +81,16 @@ pub(crate) fn execute_entry_point_call_cairo0(
         n_total_args,
     )?;
 
+    let (syscall_usage_vm_resources, syscall_usage_sierra_gas) =
+        match execution_result.tracked_resource {
+            TrackedResource::CairoSteps => (syscall_usage, SyscallUsageMap::default()),
+            TrackedResource::SierraGas => (SyscallUsageMap::default(), syscall_usage),
+        };
```
ddoktorski: It is not possible to track Sierra gas for Cairo 0, so we don't need this logic here

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/entry_point.rs
```diff
@@ -309,22 +315,43 @@ fn remove_syscall_resources_and_exit_non_error_call(
     let mut resources = call_info.resources.clone();
     let mut gas_consumed = call_info.execution.gas_consumed;
 
+    // Remove resources consumed by syscalls from the current call
+    // `syscall_usage_vm_resources` and `syscall_usage_sierra_gas` are flat, meaning they only include syscalls from the specific call
     match current_tracked_resource {
```
ddoktorski: nit: One of the values `syscall_usage_vm_resources` or `syscall_usage_sierra_gas` is always empty, so we can remove the `match` and it should work the same

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
```diff
@@ -769,7 +810,7 @@ pub fn get_all_used_resources(
 
     let mut execution_resources = top_call.borrow().used_execution_resources.clone();
     let mut sierra_gas_consumed = top_call.borrow().gas_consumed;
-    let top_call_syscalls = top_call.borrow().used_syscalls.clone();
+    let top_call_syscalls = top_call.borrow().get_total_used_syscalls();
```
ddoktorski, line 813: This seems incorrect, instead we should rather simply add syscall resources to the respective resource type without making a distinction between different tracked resource, something along the lines of:
```rust
execution_resources = add_syscall_execution_resources(
    versioned_constants,
    &execution_resources,
    &top_call.borrow().used_syscalls_vm_resources,
);
sierra_gas_consumed += get_syscalls_gas_consumed(
    &top_call.borrow().used_syscalls_sierra_gas,
    versioned_constants,
);
```

ddoktorski (changes requested):

ddoktorski (approved):
🙏

PR #3514, opened 2025-07-02
Always pass `function_name` to profiler call trace

### LINKED ISSUE #180: Improve way of presenting information for forks in profiler
Image

As seen at attached image information presented for forks is not readable.

We can improve following parts of output:
- selectors and addresses as hex
- get function name from abi (should be done in forge)

Related #85

ksew1:
Closes https://github.com/software-mansion/cairo-profiler/issues/180

---

**Stack**:
- #3514 ⬅
- #3513

### DIFF
--- a/crates/forge-runner/src/build_trace_data.rs
+++ b/crates/forge-runner/src/build_trace_data.rs
@@ -18,6 +18,7 @@ use cairo_annotations::trace_data::{
 use cairo_vm::vm::runners::cairo_runner::ExecutionResources;
 use cairo_vm::vm::trace::trace_entry::RelocatedTraceEntry;
 use camino::{Utf8Path, Utf8PathBuf};
+use cheatnet::forking::data::ForkData;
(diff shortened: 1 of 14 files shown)


crates/forge-runner/src/build_trace_data.rs
```diff
@@ -191,18 +198,19 @@ fn get_contract_name(
...
-    } else {
+            .into_())
+    .then(|| TEST_CODE_FUNCTION_NAME.to_string())
+    .or_else(|| {
         contracts_data
             .get_function_name(entry_point_selector)
             .cloned()
-    }
+    })
+    .or_else(|| fork_data.selectors.get(entry_point_selector).cloned())
```
cptartur: If only we could have normal if/else instead of these functional chains

cptartur (approved):

franciszekjob (approved):

crates/forge-runner/src/build_trace_data.rs
ksew1: As you wish

PR #3539, opened 2025-07-08
Add `interact_with_state` documentation

### LINKED ISSUE #3331: Closure cheatcode for interacting with `contract_state_for_testing`
### Current State

Currently, `contract_state_for_testing` doesn't play nicely with normal flow of tests - changes to the storage made through it aren't reflected in the standard test execution (using dispatchers).

The state not being reflected is caused by `ContractState` returned by `contract_state_for_testing` using  different storage addresses than the contract used by dispatcher (details to be researched, supposedly it assumes the contract address to be 0).

By cheating the contract address, we can make the `ContractState` work with the normal fow

### Objective

Implement the closure cheatcode for interacting with `contract_state_for_testing`

### Additional Context

Example interface should be something along the lines of this

```cairo
fn modify_state<
    F, 
    +Drop<F>,
    impl func: core::ops::Fn<F, ()>,
>(contract_address: ContractAddress, f: F) -> func::Output {
    // start cheat
    f()
    // end cheat
}
```

ddoktorski:
Closes #3331

commit-id:43b80241


docs/src/appendix/cheatcodes/interact_with_state.md
```diff
@@ -0,0 +1,48 @@
+# `interact_with_state`
+
+> `pub fn interact_with_state<F, +Drop<F>, impl func: core::ops::FnOnce<F, ()>, +Drop<func::Output>>(contract_address: ContractAddress, f: F,) -> func::Output`
+
+Allows to use `contract_state_for_testing` for a deployed contract, enabling interaction with its state in tests.
+
+To make it possible to use this cheatcode, it is necessary to take care of the following:
+- The contract must be visible in the test context
+- Storage variables that you want to access must be public
+- If testing internal contract functions, the respective trait must be imported
```
franciszekjob: Could we add an example for that one as well?

docs/src/appendix/cheatcodes/interact_with_state.md
```diff
@@ -0,0 +1,48 @@
...
+        let mut state = Counter::contract_state_for_testing();
+
+        // 4. Read from storage
+        let current_count = state.count.read();
+
+        // 5. Write to storage
+        state.count.write(current_count + 1);
+    }
+);
+```
```
franciszekjob: Please add necessary imports (e.g. storage related traits)

docs/src/appendix/cheatcodes/interact_with_state.md
```diff
@@ -0,0 +1,48 @@
...
+### Accessing and modifying contract state
+
+You can use `interact_with_state` to access and modify a deployed contract's state. After deploying a contract, you can pass its address to `interact_with_state` along with a closure. Inside the closure, you can use `contract_state_for_testing()` to get access to the contract's state and then read from or write to its storage variables.
+
+```rust
+#[starknet::contract]
+mod Counter {
+    #[storage]
+    struct Storage {
+        count: u64,
```
franciszekjob: This variable should be public (as per above requirements).

franciszekjob (approved):
Looks good. Maybe I'm a bit too picky but having example with code directly in this page would be more attractive. Leaving this up to you.

image

docs/src/appendix/cheatcodes/interact_with_state.md
```diff
@@ -0,0 +1,19 @@
+# `interact_with_state`
+
+> `pub fn interact_with_state<F, +Drop<F>, impl func: core::ops::FnOnce<F, ()>, +Drop<func::Output>>(contract_address: ContractAddress, f: F,) -> func::Output`
+
+Allows to use `contract_state_for_testing` for a deployed contract, enabling interaction with its state in tests.
+
+To make it possible to use this cheatcode, it is necessary to take care of the following:
+- The contract must be visible in the test context
```
cptartur:
```suggestion
- The contract implementation must be visible in the test context
```

docs/src/appendix/cheatcodes/interact_with_state.md
```diff
@@ -0,0 +1,19 @@
...
+- Storage struct along with the variables that you want to access must be public
+- If testing internal contract functions, the respective trait or specific function must be imported
+- Storage related traits must be imported, such as `StoragePointerReadAccess` and `StoragePointerWriteAccess`
+
+Please note that to use `interact_with_state` with a forked contract, it is required to have the contract's implementation.
```
cptartur: So maybe what we write above about contract being visible is a bit misleading? How does it work for forks if it doesn't require full implementation?
Maybe we should write that implementation is needed for full functionality?

docs/src/appendix/cheatcodes/interact_with_state.md
```diff
@@ -0,0 +1,19 @@
...
+
+For example usage, please refer to the [testing contract internals](../../testing/testing-contract-internals.md#modifying-the-state-of-an-existing-contract) documentation.
+
+> ⚠️ **Warning**
+>
+> Full support for closures was introduced in Cairo `2.11`. Therefore, this cheatcode may produce unexpected results or fail in earlier versions.
```
cptartur: I'd put it on top of the section and maybe just write it as more like "The feature might not work correctly in Cairo versions before 2.11 due to problems with the support for closures".

docs/src/snforge-advanced-features/storage-cheatcodes.md
```diff
@@ -9,6 +9,10 @@ In order to obtain the variable address that you'd like to write to, or read fro
 - `map_entry_address` function in tandem with `selector!` - for key-value pair of a map variable
 - `starknet::storage_access::storage_address_from_base`
 
+> 📝 **Note**
+>
+> For more user friendly access to storage, consider using the [`interact_with_state`](../appendix/cheatcodes/interact_with_state.md) cheatcode.
+
```
cptartur, line 15: Maybe we should just change this and start promoting `interact_with_state` as a default option and keep store load as an alternative option.

I know that I said previously that `interact_with_state` shouldn't be promoted that much but it provides much better api than store load.

cptartur: Though people might start overusing it so I'm not sure. Maybe we should promote it more as a function just for "mocking" the state in context of store/load functionality.

docs/src/testing/testing-contract-internals.md
```diff
@@ -1,31 +1,45 @@
...
 3. This function will always return the struct keeping track of the state of the test. It means that within one test every result of `contract_state_for_testing` actually points to the same state.
 
+#### Modifying the state of an existing contract
+
+There is a special [`interact_with_state`](../appendix/cheatcodes/interact_with_state.md) cheatcode dedicated for using `contract_state_for_testing` with a deployed contract.
+By using this function, the state will be modified for the provided contract address.
+
+```rust
+{{#include ../../listings/testing_contract_internals/src/using_interact_with_state.cairo}}
+```
```
cptartur: We are importing the whole big snippet here, is that good for the readability of the docs? Maybe we should at least split it by `#[test]` functions used in the snippet?

docs/src/appendix/cheatcodes/interact_with_state.md
ddoktorski: >So maybe what we write above about contract being visible is a bit misleading? 

I'm not sure if I understand. Why is it misleading?

docs/src/snforge-advanced-features/storage-cheatcodes.md
ddoktorski: Given that `interact_with_state` is explained in the section about testing internals, I think we don’t currently specify what the default is. Both options are described in different places, which seems reasonable to me

docs/src/testing/testing-contract-internals.md
ddoktorski: Added the same contract for both examples for `contract_state_for_testing` separately, so main snippets contain only test files. Let me know if that works for you

docs/src/appendix/cheatcodes/interact_with_state.md
ddoktorski: Done

ddoktorski:
> Looks good. Maybe I'm a bit too picky but having example with code directly in this page would be more attractive. Leaving this up to you.

I've added a minimal example on this page

docs/src/appendix/cheatcodes/interact_with_state.md
```diff
@@ -0,0 +1,38 @@
+# `interact_with_state`
+
+> ⚠️ **Warning**
+>
+> The feature might not work correctly if you're using Cairo version earlier than `2.11`, because closures weren’t fully supported before then.
```
cptartur:
```suggestion
> The feature might not work correctly if you're using Cairo version earlier than `2.11`, because closures weren’t fully supported before it.
```

docs/src/appendix/cheatcodes/interact_with_state.md
```diff
@@ -0,0 +1,38 @@
...
+
+> `pub fn interact_with_state<F, +Drop<F>, impl func: core::ops::FnOnce<F, ()>, +Drop<func::Output>>(contract_address: ContractAddress, f: F,) -> func::Output`
+
+Allows to use `contract_state_for_testing` for a deployed contract, enabling interaction with its state in tests.
+
+To make it possible to use this cheatcode, it is necessary to take care of the following:
```
cptartur:
```suggestion
To use this cheatcode, it is necessary to take care of the following:
```

docs/src/appendix/cheatcodes/interact_with_state.md
```diff
@@ -0,0 +1,38 @@
...
+- The contract implementation must be visible in the test context
+- Storage struct along with the variables that you want to access must be public
+- If testing internal contract functions, the respective trait or specific function must be imported
+- Storage related traits must be imported, such as `StoragePointerReadAccess` and `StoragePointerWriteAccess`
```
cptartur, line 16: Out of curiosity, does the compiler complain with something meaningful if these traits are missing?

docs/src/testing/testing-contract-internals.md
```diff
@@ -1,31 +1,51 @@
...
+> When using `contract_state_for_testing` without the `interact_with_state` cheatcode, the storage is modified in the context of the [`test_address`](#snforge_stdtest_address---address-of-test-contract) contract.
+
+In the examples below, we will use the following contract:
 
 ```rust
-{{#include ../../listings/testing_contract_internals/src/basic_example.cairo}}
+{{#include ../../listings/testing_contract_internals/src/contract.cairo}}
+```
+
+#### Testing state without a deployed contract
```
cptartur: I'd consider just removing this section, don't see many valid use cases for using `contract_state_for_testing` like that. Or at least move it somewhere below and as standard show usage of `contract_state_for_testing`.

docs/src/appendix/cheatcodes/interact_with_state.md
ddoktorski: In basic examples it usually displays accurate errors, but in more complex cases it depends. For example when we want to modify `Map` element we need to import `StorageMapWriteAccess`. If we don't do this, but at the same time `StoragePointerWriteAccess` is in scope, the compiler considers that trait the natural candidate when using the `write()` method instead of the correct `StorageMapWriteAccess`. In this case we get a long error message, which might be confusing for some users.

cptartur: Okay, I'd assume users using this functionality are advanced enough that they will know relevant traits are required.

docs/src/testing/testing-contract-internals.md
ddoktorski: Removed

docs/src/appendix/cheatcodes/interact_with_state.md
```diff
@@ -0,0 +1,39 @@
...
+Allows to use `contract_state_for_testing` for a deployed contract, enabling interaction with its state in tests.
+
+To use this cheatcode, it is necessary to take care of the following:
+
+- The contract implementation must be visible in the test context
+- Storage struct along with the variables that you want to access must be public
+- If testing internal contract functions, the respective trait or specific function must be imported
+- Storage related traits must be imported, such as `StoragePointerReadAccess` and `StoragePointerWriteAccess`
+
+Please note that to use `interact_with_state` with a forked contract, it is required to have the contract's implementation.
```
cptartur: Maybe let's make it a Note block

docs/src/appendix/cheatcodes/interact_with_state.md
```diff
@@ -0,0 +1,39 @@
...
+
+### Usage
+
+To use this cheatcode, follow these steps:
+
+1. Provide the contract address as the first argument
+2. Define a closure that modifies the contract's state and pass it as the second argument
+3. Inside the closure, define a mutable variable for the contract's state using `contract_state_for_testing`
+4. Use this state variable to read from or write to the contract’s storage
```
cptartur: Can we do anything with state beside these? If so, let's just write something like "Use this state variable to interact with state".

docs/src/testing/testing-contract-internals.md
```diff
@@ -1,30 +1,39 @@
 # Testing Contracts' Internals
 
-Sometimes, you want to test a function which uses Starknet context (like block number, timestamp, storage access)
-without deploying the actual contract.
+Sometimes, you want to test a function that uses Starknet context such as block number, timestamp or storage access.
 
-Since every test is treated like a contract, using the aforementioned pattern you can test:
+Since every test is treated like a contract (with the address [`test_address`](#snforge_stdtest_address---address-of-test-contract)), you can use the aforementioned pattern to test:
```
cptartur, line 5: I wonder if it's worth mentioning the test's address. It's not that it is a secret but it's kind of an implementation detail.

cptartur: Though it makes sense in the context of the warning below.

cptartur (approved):

PR #3543, opened 2025-07-09
Change type of `CheatSpan::TargetCalls` to `NonZeroUsize` 

### LINKED ISSUE #2927: Add check preventing setting `CheatSpan` to `0`

MKowalski8:
Closes #2927 

## Introduced changes

- Change type of `CheatSpan::TargetCalls` from `usize` to `NonZeroUsize`. This prevents the user from setting the value to 0, which would cause an underflow in CheatStatus::decrement_cheat_span and make the cheat persist indefinitely instead of being canceled.

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


snforge_std/src/cheatcodes.cairo
```diff
@@ -53,7 +53,8 @@ pub fn mock_call<T, impl TSerde: core::serde::Serde<T>, impl TDestruct: Destruct
     let contract_address_felt: felt252 = contract_address.into();
     let mut inputs = array![contract_address_felt, function_selector];
 
-    CheatSpan::TargetCalls(n_times).serialize(ref inputs);
+    CheatSpan::TargetCalls(n_times.try_into().expect('n_times must be greater than 0'))
```
franciszekjob: nit:
```suggestion
    CheatSpan::TargetCalls(n_times.try_into().expect('`n_times` must be greater than 0'))
```

crates/cheatnet/src/state.rs
```diff
@@ -180,9 +181,13 @@ pub enum CheatStatus<T> {
...
         if let CheatStatus::Cheated(_, CheatSpan::TargetCalls(n)) = self {
-            *n -= 1;
-            if *n == 0 {
+            let calls_number = n.get() - 1;
+
+            if calls_number == 0 {
                 *self = CheatStatus::Uncheated;
+            } else {
+                *n = NonZeroUsize::new(calls_number)
+                    .expect("NonZeroUsize should not be zero after decrement");
```
franciszekjob: nit:
```suggestion
                    .expect("`NonZeroUsize` should not be zero after decrement");
```

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/cheatcodes/cheat_block_hash.rs
```diff
@@ -90,14 +91,21 @@ impl CheatnetState {
...
                 }
                 CheatSpan::TargetCalls(num) => {
+                    let calls_number = num.get() - 1;
                     self.block_hash_contracts.insert(
                         (contract_address, block_number),
-                        (CheatSpan::TargetCalls(num - 1), block_hash),
+                        (
+                            CheatSpan::TargetCalls(
+                                NonZeroUsize::new(calls_number)
+                                    .expect("NonZeroUsize should not be zero after decrement"),
```
franciszekjob: nit:
```suggestion
                                    .expect("`NonZeroUsize` should not be zero after decrement"),
```

CHANGELOG.md
```diff
@@ -13,6 +13,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Bug where syscall execution resources from nested calls were being calculated twice
 
+#### Changed
+
+- Use of `NonZero<usize>` type instead of `usize` in `CheatSpan::TargetCalls`
```
```suggestion
- `CheatSpan::TargetCalls` accepts now `NonZero<usize>` instead of `usize`, in `snforge_std`
```

franciszekjob (approved):
👌 , please see nits

CHANGELOG.md
ddoktorski: What about moving this entry under the `Fixed` section and rephrasing it to be more focused on what it fixes from users perspective? Something along the lines of: "A bug that allowed passing a cheatcode span of 0, which incorrectly behaved like `CheatSpan::Indefinite`"

ddoktorski (commented):
After this change passing `CheatSpan::TargetCalls(0)` seems to result in most cases in the following error:
Screenshot 2025-07-14 at 12 18 43
Could we consider making the error message more descriptive, if it’s not too much work?

snforge_std/src/cheatcodes.cairo
MKowalski8: I got error `The value does not fit within the range of type core::felt252.`, after suggested change. 

I can go either with:
``` cairo
CheatSpan::TargetCalls(n_times.try_into().expect('`n_times` must be > 0))
```
Or something like that:
``` cairo
CheatSpan::TargetCalls(n_times.try_into().expect('`n_times` must be positive'))
```

franciszekjob: let's use 1st one then

MKowalski8:
> After this change passing `CheatSpan::TargetCalls(0)` seems to result in most cases in the following error: Screenshot 2025-07-14 at 12 18 43 Could we consider making the error message more descriptive, if it’s not too much work?

I can’t reproduce the error. Could you describe the exact cases you’re referring to? When I try to use `CheatSpan::TargetCalls(0)`, I get the following compile-time error:
obraz

ddoktorski:
Ahh probably I missed updating std dependency when checking. In that case no problem, the error is good.

ddoktorski (approved):

PR #3547, opened 2025-07-11
Support RPC `0.9.0`

### LINKED ISSUE #3515: Support RPC 0.9.0
### Current State

Integrate starknet-rs compatible with RPC 0.9.0

### Objective

Support RPC 0.9.0

### Additional Context

_No response_

ddoktorski:
Closes #3515


crates/sncast/src/lib.rs
```diff
@@ -482,8 +485,16 @@ pub async fn check_if_legacy_contract(
...
+        Some(class_hash) => {
+            provider
+                .get_class(BlockId::Tag(PreConfirmed), class_hash)
+                .await
+        }
+        None => {
+            provider
+                .get_class_at(BlockId::Tag(PreConfirmed), address)
+                .await
+        }
```
franciszekjob, line 491: nit: Unrelated to this PR but could be simplified:
```rust
    let contract_class = provider
        .get_class(BlockId::Tag(PreConfirmed), class_hash.unwrap_or(address))
        .await
        .map_err(handle_rpc_error)?;
```

sncast_std/src/lib.cairo
```diff
@@ -161,7 +161,6 @@ pub enum ProviderError {
 
 #[derive(Drop, Serde, PartialEq, Debug)]
 pub enum TransactionError {
-    Rejected,
     Reverted: ErrorData,
 }
```
franciszekjob, line 166: We can change it to struct, cause now it's an enum with just one variant:
```rust
#[derive(Error, Debug, CairoSerialize)]
#[error("Transaction has been reverted = {data}")]
pub struct RevertedTransactionError {
    pub data: ByteArray,
}

File: crates/sncast/src/lib.rs [hunk 3]
```diff
@@ -534,13 +545,13 @@ fn map_encoding(legacy: bool) -> ExecutionEncoding {
 
 pub fn get_block_id(value: &str) -> Result<BlockId> {
     match value {
-        "pending" => Ok(BlockId::Tag(Pending)),
+        "pre_confirmed" => Ok(BlockId::Tag(PreConfirmed)),
```
cptartur, line 542: This is a bit inconsistent with the syntax we use in toml and some other places, maybe we should always use `-` or `_` in pre-confirmed?

cptartur (commented):
Mostly looks good, see my comment about pre-confirmed spelling

sncast_std/src/lib.cairo
ddoktorski: This requires a breaking change in `sncast_std` so I would prefer not to include it in the scope of this PR

crates/sncast/src/lib.rs
ddoktorski: Replaced with using underscore in all places except `SncastMessage`, it is more consistent with other outputs to omit it there

crates/sncast/src/lib.rs
ddoktorski: It is not the same since one arm uses `get_class` and the other `get_class_at`

franciszekjob: Oh, right, disregard it.

sncast_std/src/lib.cairo
franciszekjob: We can include it in cast scripts refactor, you can link a todo here.

franciszekjob (approved):

cptartur (approved):

PR #3553, opened 2025-07-15
Bump deps, use scarb in nightly

cptartur:
**Stack**:
- #3573
- #3559
- #3553 ⬅
- #3503
- #3502
- #3470
- #3466
- #3465
- #3464
- #3424
- #3421
- #3420
- #3419
- #3415

### DIFF
--- a/.github/workflows/_publish-plugin.yml
+++ b/.github/workflows/_publish-plugin.yml
@@ -31,7 +31,7 @@ jobs:
           
           plugin_name_underscores=${{ inputs.plugin_name }}
           plugin_name_underscores=${plugin_name_underscores//-/_}
-          plugin_version=$(get_version "${{ inputs.overridden_plugin_version }}")
+          plugin_version=$(get_version "${{ inputs.overridden_plugin_version }}" "crates/${{ inputs.plugin_name }}/Scarb.toml")
           
           registry_url=${{ inputs.prod_registry == true && 'https://scarbs.xyz' || 'https://scarbs.dev' }}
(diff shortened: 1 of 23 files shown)


.github/workflows/_publish-plugin.yml
```diff
@@ -79,7 +79,7 @@ jobs:
           fi
 
       - name: Publish ${{ inputs.plugin_name }}
-        if: needs.check-uploaded.outputs.plugin_uploaded == 'false' || github.event_name == 'workflow_dispatch'
```
kkawula, line 81: `github.event_name == 'workflow_dispatch'`
this line is responsible for publishing plugin manually
As you can see, the only way to trigger `_publish-plugin.yml` is `workflow_call`, but there is no such event_name as `workflow_call`, event_name is always inherited from calling workflow, in that scenerio it will be always, `workflow_dispatch` from [`publish-plugin.yml`](https://github.com/foundry-rs/starknet-foundry/blob/52ff01d7ea6ace83480e31d8bcef1871fde07c3a/.github/workflows/publish-plugin.yml#L24), that can be triggered only in that way https://github.com/foundry-rs/starknet-foundry/blob/52ff01d7ea6ace83480e31d8bcef1871fde07c3a/.github/workflows/publish-plugin.yml#L4

cptartur: Right, but even in case of manually running a workflow, we should still check if the plugin was not published already. Because if it was, the workflow will just fail, unless that was the intended behavior?

The problem is, `_publish-plugin` workflow is also reused in nightly workflow which is also triggered manually. If any step of nightly workflow fails after uploading plugins, then due to this line, you cannot re-run it because on subsequent runs it will fail as it will try to upload plugin again (which registry doesn't allow).

cptartur:

PR #3561, opened 2025-07-21
Rename `completion` to `completions` 

MKowalski8:
## Introduced changes

- Rename `completion` command to `completions`

## Checklist

- [ ] Linked relevant issue
- [x] Added relevant tests


CHANGELOG.md
```diff
@@ -9,6 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Forge
 
+#### Deprecated
+
+- The `snforge completion` command. Go with `snforge completions` instead
```
cptartur:
```suggestion
- The `snforge completion` command. Use `snforge completions` instead
```

CHANGELOG.md
```diff
@@ -26,6 +30,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 #### Deprecated
 
 - `argent` option for `--type` flag in `account create` and `account import` commands. Use `ready` instead
+- The `sncast completion` command. Go with `sncast completions` instead
```
cptartur: Same here

crates/shared/src/auto_completions.rs
```diff
@@ -1,24 +1,33 @@
...
+        anyhow::bail!("Unsupported shell")
+    };
+
+    generate_completions_to_stdout(shell, cmd);
+
+    // TODO(#3560): Remove this warning when the `completion` alias is removed
+    if std::env::args().nth(1).as_deref() == Some("completion") {
+        let message = &WarningMessage::new(
+            "Command `completion` is deprecated and will be removed in the future. Please use `completions` instead.",
+        );
```
cptartur: I'd handle it somewhere outside of the function to generate completions, maybe at a caller site (you can extract the logic to function if needed).

crates/shared/src/auto_completions.rs
```diff
@@ -1,24 +1,33 @@
...
+
+        ui.println(&format!("# {}", message.text()));
```
cptartur: Why `#` in the output?

docs/src/getting-started/installation.md
cptartur, line 1: I assume these commands were copied from Scarb repo? Can't really tell if they are good or not.

CHANGELOG.md
MKowalski8: I did not want to repeat the same message from above `Use ready instead`, but if this is not a problem I can change it.

crates/shared/src/auto_completions.rs
MKowalski8: The `#` is necessary because the `completions` command generates a script whose output is used directly, so this line must be a comment.

docs/src/getting-started/installation.md
MKowalski8: They are based on `scarb` repo, but I tested every of them locally on `linux` and `macos`.

CHANGELOG.md
cptartur: It's not a problem

docs/src/getting-started/installation.md
cptartur: Great, maybe let's remove the section on powershell entirely, since we aren't supporting windows without WSL2 anyway.

cptartur (approved):

ddoktorski (approved):

PR #3574, opened 2025-07-24
Increase default `max_steps` limit to `usize::MAX`

### LINKED ISSUE #3571: Increase defualt max-n-steps to maxiumu value
### Current State

Right now it mimicks the starknet limit of 10M steps

### Objective

Increase defualt max-n-steps to maxiumu value

### Additional Context

Use usize::max or something and allow overriding it (lowering it) using `--max-n-steps`.

MKowalski8:
Closes #3571 

## Introduced changes

- Increase default value for `max-n-steps` to `usize::MAX` from 10 million

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/forge/tests/e2e/running.rs
```diff
@@ -773,7 +773,11 @@ fn with_exit_first() {
         ))
         .unwrap();
 
-    let output = test_runner(&temp).assert().code(1);
+    let output = test_runner(&temp)
+        .args(["--max-n-steps", "10000000"])
```
ddoktorski, line 777: Do we need to update it here?

crates/forge/tests/e2e/steps.rs
ddoktorski, line 1: Let's rename test names in this file, for example the name `should_allow_less_than_default` is misleading now

crates/forge/tests/e2e/running.rs
MKowalski8: Yes, we need to update it because this test expects a failure caused by performing too many steps. However, usize::MAX is far greater than 10 million, so the test would either never finish or take an extremely long time to complete.

crates/forge/tests/e2e/steps.rs
MKowalski8: What exactly would you propose? If our new default is `usize::MAX`, then doesn’t the `should_allow_less_than_default` test accurately describe the situation where we manually lower the limit using `--max-n-steps`? 

One test I could rename is `should_allow_more_than_10m` to something like `should_allow_more_than_blockifier_limit`, as that better describes what it actually does.

ddoktorski: Now it's more like just a test for this flag as every value is less than the default, but fair enough we can leave it as it is

ddoktorski (approved):

franciszekjob (approved):

cptartur (approved):
Let's wait with merging it though

PR #3578, opened 2025-07-27
feat: Improve error message when trying to redeclare contract using sncast declare #3253

### LINKED ISSUE #3253: Improve error message when trying to redeclare contract using `sncast declare`
### Current State

Currently, we show something along the lines of 

```shell
command: declare
error: Transaction execution error = TransactionExecutionErrorData { transaction_index: 0, execution_error: "Class with hash 0x0273a597e2dbb18296c788c471af35c99315bb738be1c8c854ca83b85a75bc98 is already declared." }
```

### Objective

Show a specific error expaling that contract was already declared instead of dumping the RPC error directly

### Additional Context

The error returned by an RPC is a specif one we can safely match by its type and display better message, that is nicely formatted (see the other tool for reference).

JoshdfG:
Closes #3253

## Introduced changes

- Added a more descriptive error message instead of dumping the whole rpc error directly 
- Added a new custom error in sncast/error.rs called `ClassAlreadyDeclared(Felt252)`

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


CHANGELOG.md
cptartur, line 1: Please don't reformat this file and add an entry about your changes here

cptartur (changes requested):
Some of cast tests are failing, probably asserts for outputs need to be updated.

Also run `cargo fmt`.

JoshdfG:
@cptartur I'll check those out I ran cargo fmt already that was an oversight

crates/sncast/src/response/errors.rs
```diff
@@ -1,5 +1,6 @@
 use crate::{ErrorData, WaitForTransactionError, handle_rpc_error};
 use anyhow::anyhow;
+use cairo_vm::Felt252;
```
cptartur:
```suggestion
use starknet_types_core::felt::Felt;
```

Please replace that, `cairo_vm::Felt252` is just an alias for that anyway.

CHANGELOG.md
```diff
@@ -1335,3 +1335,13 @@ snforge_std = { git = "https://github.com/foundry-rs/starknet-foundry", tag = "v
...
+
+### Cast
+
+#### Added
+
+- ClassAlreadyDeclared(Felt252) to the custom errors in `sncast/error.rs`.
+
+#### Changed
+
+- Match arm to include the new error message instead of the RPC provider error, for a more descriptive error.
```
cptartur: I think this kind of entry is leaking implementation detail. I'd add a `Changed` entry saying that error in case of redeclaration was improved.

cptartur (commented):
Looks good, let's wait for the tests to pass

crates/sncast/src/starknet_commands/declare.rs
```diff
@@ -137,7 +137,9 @@ pub async fn declare(
                 class_hash: class_hash.into_(),
             }))
         }
-        Err(Provider(error)) => Err(StarknetCommandError::ProviderError(error.into())),
+        Err(Provider(ProviderError::StarknetError(StarknetError::ClassAlreadyDeclared))) => Err(
+            StarknetCommandError::ClassAlreadyDeclared(class_hash.into_()),
+        ),
```
cptartur, line 142: The error here is duplicating an handling of `AlreadyDeclared` we have couple of lines above

cptartur (commented):
Thanks for contributing, this will however not address the problem. 
The added error handling duplicates the one we already have but the error still persists.

The problem is not due to missing error handling but from the fact that error comes from `estimateFee` endpoint and not `declare`.

cptartur:
> Thanks for contributing, this will however not address the problem. The added error handling duplicates the one we already have but the error still persists.
> 
> The problem is not due to missing error handling but from the fact that error comes from `estimateFee` endpoint and not `declare`.

I'll be investigating the problem with estimateFee

PR #3600, opened 2025-08-04
Move to trace components instead of verbosity

ksew1:
commit-id:022e6e71

---

**Stack**:
- #3612
- #3609
- #3605
- #3601
- #3600 ⬅


crates/debugging/src/components.rs
```diff
@@ -0,0 +1,124 @@
...
+    /// Checks if a specific [`Component`] is included in the set.
+    #[must_use]
+    pub fn contains(&self, component: &Component) -> bool {
+        self.set.contains(component)
+    }
+}
+
+/// Components that will be included in the trace.
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub enum Component {
```
franciszekjob: Same as above?
```suggestion
pub enum TraceComponent {
```

crates/debugging/src/components.rs
```diff
@@ -0,0 +1,124 @@
+use crate::trace::types::{
+    CallerAddress, ContractAddress, ContractName, TransformedCallResult, TransformedCalldata,
+};
+use blockifier::execution::entry_point::CallType;
+use paste::paste;
+use starknet_api::contract_class::EntryPointType;
+use std::collections::HashSet;
+
+/// Represents a set of [`Component`] that will be included in a trace.
+pub struct Components {
```
franciszekjob: More specific?
```suggestion
pub struct TraceComponents {
```

crates/debugging/src/components.rs
```diff
@@ -0,0 +1,124 @@
...
+                "` is included in the [`Components`]."
+            )]
+            #[derive(Debug, Clone)]
+            pub struct [<$variant Container>] {
+                value: Option<$ty>,
+            }
+
+            impl [<$variant Container>] {
+                #[doc = concat!(
+                    "Creates a new [`", stringify!($variant), "Container`] from a concrete value.\n\n",
```
franciszekjob:
```suggestion
                    "Creates a new [`", stringify!($variant), "Container`] from a specific value.\n\n",
```

crates/debugging/src/components.rs
franciszekjob, line 1: Why don't we put this file under `trace`?

ksew1: Good idea! done.

crates/debugging/src/components.rs
ksew1: I would prefer not to use bloated names. Since it is currently a module called trace, I think it can stay that way.

franciszekjob: Yeah if it's in module now then fine.

franciszekjob (approved):

cptartur (approved):

PR #3605, opened 2025-08-04
Add tests for `--trace-components` flag

ksew1:
commit-id:0a3ba8ce

---

**Stack**:
- #3612
- #3609
- #3605 ⬅

### DIFF
--- a/crates/forge/tests/e2e/debugging.rs
+++ b/crates/forge/tests/e2e/debugging.rs
@@ -1,6 +1,21 @@
 use super::common::runner::{setup_package, test_runner};
-use indoc::formatdoc;
+use indoc::{formatdoc, indoc};
 use shared::test_utils::output_assert::assert_stdout_contains;
+
+#[test]
+fn debugging_trace_custom_components() {
+    let temp = setup_package("debugging");
+
+    let output = test_runner(&temp)
+        .arg("--trace-components")
+        .arg("contract-name")
+        .arg("call-result")
+        .arg("call-type")
+        .assert()
+        .code(1);
+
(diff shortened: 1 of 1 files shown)


franciszekjob (approved):

crates/forge/tests/e2e/debugging.rs
```diff
@@ -311,3 +346,33 @@ fn minimal_debugging_trace_message_fork(test_name: &str) -> String {
...
+        │  └─ [selector] execute_calls
+        │     ├─ [contract name] SimpleContract
+        │     ├─ [call type] Call
+        │     └─ [call result] success: array![]
+        └─ [selector] fail
+           ├─ [contract name] SimpleContract
+           ├─ [call type] Call
+           └─ [call result] panic: (0x1, 0x2, 0x3, 0x4, 0x5)
+        "}
+}
```
franciszekjob, line 378: nit: I know it's consistent with other tests but I wonder what's the actual benefit of having the expected output in separate function? Imo it's more comfortable when I don't have to jump to another function in order to see the output. Leaving this one up to you.

franciszekjob (approved):

crates/forge/tests/e2e/debugging.rs
ksew1: I think it depends on the workflow of the person writing or maintaining the tests. In my case, I prefer doing it this way. These separate functions encourage reuse of the output and make the tests clearer. From my experience, when tests are messy or too long, nobody wants to write them.

You’re right that you just need to jump to another function, but once the tests are written, I rarely find myself needing to change anything in the output.

Because of that, I would prefer to keep it this way.

cptartur (approved):

PR #3621, opened 2025-08-08
Add casm debug info to ContractsDataStore

ksew1:
commit-id:6dc6b503

---

**Stack**:
- #3634
- #3633
- #3621 ⬅


crates/debugging/src/contracts_data_store.rs
```diff
@@ -99,4 +116,36 @@ impl ContractsDataStore {
...
+{
+    let mut map1 = HashMap::new();
+    let mut map2 = HashMap::new();
+
+    for (key, (v1, v2)) in iter {
+        map1.insert(key, v1);
+        map2.insert(key, v2);
+    }
+
+    (map1, map2)
```
cptartur, line 150: Let's just not start reusing this between files 😅

cptartur (approved):

crates/debugging/src/contracts_data_store.rs
```diff
@@ -99,4 +116,36 @@ impl ContractsDataStore {
...
     }
+
+    /// Gets the [`CairoProgramDebugInfo`] for a given contract [`ClassHash`].
+    #[must_use]
+    pub fn get_casm_debug_info(&self, class_hash: &ClassHash) -> Option<&CairoProgramDebugInfo> {
+        self.casm_debug_infos.get(class_hash)
+    }
+}
+
+/// Compile the given [`ContractClass`] to `casm` and return [`CairoProgramDebugInfo`]
```
franciszekjob, line 127:
```suggestion
/// Compile the given [`ContractClass`] to [`CairoProgramDebugInfo `] and return [`CairoProgramDebugInfo`]
```

franciszekjob (approved):

crates/debugging/src/contracts_data_store.rs
ksew1: No plans for that 😄

crates/debugging/src/contracts_data_store.rs
ksew1: I would say that`casm` debug info is a byproduct of compilation, not its target. So the current doc better aligns with this perspective.

franciszekjob: Hmm ok let's leave it like that

PR #3633, opened 2025-08-12
Add function level trace

ksew1:
**Stack**:
- #3657
- #3634
- #3633 ⬅

### DIFF
--- a/crates/debugging/Cargo.toml
+++ b/crates/debugging/Cargo.toml
@@ -8,12 +8,14 @@ blockifier.workspace = true
 cairo-lang-sierra.workspace = true
 cairo-lang-sierra-to-casm.workspace = true
 cairo-lang-starknet-classes.workspace = true
+cairo-annotations.workspace = true
 cheatnet = { path = "../cheatnet" }
 console.workspace = true
 data-transformer = { path = "../data-transformer" }
 paste.workspace = true
 ptree.workspace = true
 rayon.workspace = true
+regex.workspace = true
 serde_json.workspace = true
 starknet-types-core.workspace = true
 starknet.workspace = true
--- a/crates/debugging/src/trace/function/name.rs
+++ b/crates/debugging/src/trace/function/name.rs
@@ -1,0 +1,49 @@
+use cairo_lang_sierra::program::{Program, StatementIdx};
+use regex::Regex;
+use std::sync::LazyLock;
+
+/// Represents a function name in the Sierra program.
+#[derive(Debug, Clone, Hash, Eq, PartialEq)]
+pub enum FunctionName {
+    NonInlined(String),
+    // TODO: Add inlined variant in next PR.
+}
+
+impl FunctionName {
+    /// Creates a [`FunctionName`] from a [`StatementIdx`] index and a [`Program`].
+    pub fn from_program(statement_idx: StatementIdx, program: &Program) -> Self {
+        let function_idx = program
+            .funcs
+            .partition_point(|f| f.entry_point.0 <= statement_idx.0)
+            - 1;
+        let function_name = program.funcs[function_idx].id.to_string();
+        let function_name = remove_loop_suffix(&function_name);
+        let function_name = remove_monomorphization_suffix(&function_name);
+        FunctionName::NonInlined(function_name)
+    }
+
+    /// Returns the function name as a [`&str`].
+    pub fn function_name(&self) -> &str {
+        match self {
+            FunctionName::NonInlined(name) => name,
+        }
+    }
+}
+
+/// Remove suffix in case of loop function e.g. `[expr36]`.
+fn remove_loop_suffix(function_name: &str) -> String {
+    static RE_LOOP_FUNC: LazyLock<Regex> = LazyLock::new(|| {
+        Regex::new(r"\[expr\d*]")
+            .expect("Failed to create regex for normalizing loop function names")
+    });
+    RE_LOOP_FUNC.replace(function_name, "").to_string()
+}
+
+/// Remove parameters from monomorphised Cairo generics e.g. `<felt252>`.
+fn remove_monomorphization_suffix(function_name: &str) -> String {
+    static RE_MONOMORPHIZATION: LazyLock<Regex> = LazyLock::new(|| {
+        Regex::new(r"::<.*>")
+            .expect("Failed to create regex for normalizing monomorphized generic function names")
+    });
+    RE_MONOMORPHIZATION.replace(function_name, "").to_string()
+}
--- a/crates/debugging/src/trace/function/node.rs
+++ b/crates/debugging/src/trace/function/node.rs
@@ -1,0 +1,36 @@
+use crate::trace::function::name::FunctionName;
+
+/// Represents a node in the function call tree, where each node corresponds to a function name
+/// and can have child nodes representing nested function calls.
+#[derive(Debug, Clone)]
+pub struct FunctionNode {
+    pub value: FunctionName,
+    pub children: Vec<FunctionNode>,
+}
+
+impl FunctionNode {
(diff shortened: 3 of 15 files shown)


crates/debugging/src/trace/function/name.rs
```diff
@@ -0,0 +1,49 @@
+use cairo_lang_sierra::program::{Program, StatementIdx};
+use regex::Regex;
+use std::sync::LazyLock;
+
+/// Represents a function name in the Sierra program.
+#[derive(Debug, Clone, Hash, Eq, PartialEq)]
+pub enum FunctionName {
+    NonInlined(String),
+    // TODO: Add inlined variant in next PR.
```
cptartur, line 9: Gotta love stacks 🙏 🙏 🙏

crates/debugging/src/trace/function/name.rs
```diff
@@ -0,0 +1,49 @@
...
+            FunctionName::NonInlined(name) => name,
+        }
+    }
+}
+
+/// Remove suffix in case of loop function e.g. `[expr36]`.
+fn remove_loop_suffix(function_name: &str) -> String {
+    static RE_LOOP_FUNC: LazyLock<Regex> = LazyLock::new(|| {
+        Regex::new(r"\[expr\d*]")
+            .expect("Failed to create regex for normalizing loop function names")
```
cptartur, line 37: Shouldn't we have the regex as a const/static and reuse it instead of creating it multiple times?

crates/debugging/src/trace/function/name.rs
```diff
@@ -0,0 +1,49 @@
...
+    });
+    RE_LOOP_FUNC.replace(function_name, "").to_string()
+}
+
+/// Remove parameters from monomorphised Cairo generics e.g. `<felt252>`.
+fn remove_monomorphization_suffix(function_name: &str) -> String {
+    static RE_MONOMORPHIZATION: LazyLock<Regex> = LazyLock::new(|| {
+        Regex::new(r"::<.*>")
```
cptartur, line 45: Same here

crates/debugging/src/trace/function/node.rs
```diff
@@ -0,0 +1,36 @@
...
+
+    /// Adds a path of function names to the current node, creating child nodes as necessary.
+    pub fn add_path(&mut self, path: Vec<FunctionName>) {
+        self.add_path_recursive(&mut path.into_iter());
+    }
+
+    fn add_path_recursive(&mut self, iter: &mut impl Iterator<Item = FunctionName>) {
+        if let Some(next) = iter.next() {
+            if let Some(child) = self.children.iter_mut().find(|c| c.value == next) {
+                child.add_path_recursive(iter);
```
cptartur, line 28: Not sure I get it, why do we have this logic?

crates/debugging/src/trace/function/stack.rs
```diff
@@ -0,0 +1,50 @@
...
+        Self {
+            stack: Vec::new(),
+            previous_stack_lengths: Vec::new(),
+        }
+    }
+
+    /// Enters a new function call by updating the stack with the new call stack.
+    /// It saves the current stack length to allow returning to it later.
+    ///
+    /// The New call stack is expected to be a prefix of the current stack.
```
cptartur, line 21: What's ensuring that? Only the user contract? 
Would checking that be too expensive?

crates/debugging/src/trace/function/stack.rs
```diff
@@ -0,0 +1,50 @@
...
+    pub fn enter_function_call(&mut self, new_call_stack: Vec<FunctionName>) {
+        self.previous_stack_lengths.push(self.stack.len());
+
+        self.stack = new_call_stack;
+    }
+
+    /// Exits the current function call by truncating the stack to the previous length.
+    /// If there is no previous length, it does nothing.
```
cptartur, line 29: Why is that? Shouldn't it truncate the stack entirely in this case?

crates/debugging/src/trace/function/stack.rs
```diff
@@ -0,0 +1,50 @@
...
+    pub fn new_stack(&self, function_name: FunctionName) -> Vec<FunctionName> {
+        let mut stack = self.stack.clone();
+
+        let empty_or_different_function = self.stack.last().is_none_or(|current_function| {
+            current_function.function_name() != function_name.function_name()
+        });
+
+        if empty_or_different_function {
+            stack.push(function_name);
+        }
```
cptartur, line 46: So basically it appends to new function to the stack if it's not already present as last?

crates/debugging/src/trace/function/node.rs
```diff
@@ -0,0 +1,36 @@
+use crate::trace::function::name::FunctionName;
+
+/// Represents a node in the function call tree, where each node corresponds to a function name
+/// and can have child nodes representing nested function calls.
+#[derive(Debug, Clone)]
+pub struct FunctionNode {
+    pub value: FunctionName,
+    pub children: Vec<FunctionNode>,
```
franciszekjob, line 8: Imo it's reasonable to make it private and expose with getters if needed.

crates/debugging/Cargo.toml
```diff
@@ -8,14 +8,16 @@ blockifier.workspace = true
...
 paste.workspace = true
 ptree.workspace = true
 rayon.workspace = true
+regex.workspace = true
 serde_json.workspace = true
 starknet-types-core.workspace = true
 starknet.workspace = true
 starknet_api.workspace = true
-thiserror.workspace = true
+thiserror.workspace = true
```
franciszekjob: Newline 😅

crates/debugging/src/trace/function/trace.rs
```diff
@@ -0,0 +1,131 @@
...
+
+    CasmLevelInfo {
+        run_with_call_header: false,
+        vm_trace,
+    }
+}
+
+/// Builds a [`FunctionTrace`] from the given stacks of function names.
+fn build_function_trace(stacks: Vec<Vec<FunctionName>>) -> FunctionTrace {
+    let mut dummy_root = FunctionNode::new(FunctionName::NonInlined("root".to_string()));
```
franciszekjob: nit: `dummy_root` works, but `placeholder_root` might better convey that it's a temporary node for building the tree:
```suggestion
    let mut placeholder_root = FunctionNode::new(FunctionName::NonInlined("root".to_string()));
```

crates/debugging/src/trace/function/trace.rs
```diff
@@ -0,0 +1,131 @@
...
+        if contracts_data_store.is_fork(&class_hash) {
+            return Err(FunctionTraceError::ForkContract);
+        }
+        let program_artifact = contracts_data_store
+            .get_program_artifact(&class_hash)
+            .expect("program artifact should be present for local contracts");
+
+        let program = &program_artifact.program;
+
+        let sierra_program_registry = ProgramRegistry::<CoreType, CoreLibfunc>::new(program)
```
cptartur, line 42: Some of these objects, like `ProgramRegistry` we already create during forge runtime. I wonder if there would be any significant benefit in trying to reuse them?

cptartur: Same goes for constructing trace

crates/debugging/src/tree/ui/display.rs
```diff
@@ -84,6 +89,27 @@ impl NodeDisplay for TransformedCallResult {
...
+
+impl NodeDisplay for FunctionTraceError {
+    const TAG: &'static str = "function trace error";
+    fn string_pretty(&self) -> String {
+        self.to_string()
+    }
+}
+
+impl NodeDisplay for FunctionNode {
+    const TAG: &'static str = "inlined";
```
cptartur: Are they always inlined?

PR #3635, opened 2025-08-12
Add `--tip` flag

### LINKED ISSUE #3516: Support tip mechanism
### Current State

Starknet 0.14.0 will support tips. Transaction with a higher tip will likely be placed earlier in the block. Read full pre-release notes [here](https://docs.google.com/document/d/1wgqtk9L_12trHBJ5SFSWwxiB4u0duureGiBncrmQiv8/edit?tab=t.0#heading=h.68jtk4thzoqb)

In the scope of this task, we need to allow passing tip in sncast

### Objective

Support tip mechanism

### Additional Context

_No response_

ddoktorski:
Closes #3516

## Introduced changes

- Add a `--tip` flag to relevant commands. If not provided set it to 0 (passing `None` will result in tip estimation on the starknet-rs side)


crates/sncast/src/helpers/fee.rs
```diff
@@ -74,8 +79,12 @@ impl FeeArgs {
                 Felt::from(max_fee)
             );
 
-            let fee_settings = FeeSettings::try_from(fee_estimate.clone())
+            let mut fee_settings = FeeSettings::try_from(fee_estimate.clone())
                 .expect("Failed to convert FeeEstimate to FeeSettings");
+
+            // If a tip is not provided, set it to 0
+            fee_settings.tip = Some(self.tip.unwrap_or(0));
```
cptartur: Maybe setting a tip could be a method on `FeeSettings` that takes `self` and returns a new instance? So we don't make it mutable.

crates/sncast/src/helpers/fee.rs
```diff
@@ -117,6 +127,7 @@ impl TryFrom<FeeEstimate> for FeeSettings {
             l2_gas_price: Some(fee_estimate.l2_gas_price),
             l1_data_gas: Some(fee_estimate.l1_data_gas_consumed),
             l1_data_gas_price: Some(fee_estimate.l1_data_gas_price),
+            tip: Some(0),
```
cptartur: Why do we use a default value here instead of None? Not sure if that's 100% idiomatic.

crates/sncast/tests/e2e/declare.rs
```diff
@@ -176,14 +180,27 @@ async fn test_happy_case_different_fees(fee_args: FeeArgs) {
...
+    assert_eq!(
+        receipt.execution_result.status(),
+        TransactionExecutionStatus::Succeeded
+    );
 
-    assert!(matches!(receipt, Declare(_)));
+    assert_eq!(
+        receipt.execution_result.status(),
+        TransactionExecutionStatus::Succeeded
+    );
```
cptartur: Isn't this duplicated?

franciszekjob (approved):
Pls update `account deploy` appendix as well

docs/src/appendix/sncast/multicall/run.md
```diff
@@ -60,6 +60,11 @@ Optional.
 
 Maximum L1 data gas unit price for the `invoke` transaction. When not used, defaults to auto-estimation.
 
+## `--tip <TIP>`
+Optional.
+
+Tip for the transaction. When not used, defaults to 0.
```
cptartur:
```suggestion
Tip for the transaction. When not provided, defaults to 0.
```

crates/sncast/src/helpers/fee.rs
```diff
@@ -105,6 +112,17 @@ pub struct FeeSettings {
     pub l2_gas_price: Option<u128>,
     pub l1_data_gas: Option<u64>,
     pub l1_data_gas_price: Option<u128>,
+    pub tip: Option<u64>,
+}
+
+impl FeeSettings {
+    #[must_use]
+    pub fn update_tip(&self, tip: u64) -> FeeSettings {
```
cptartur, line 120: I think it would make sense for it to take `self` so it's clearer that we are converting the existing instance to a new one with tip. Don't have a strong opinion on it though.

cptartur (approved):

crates/sncast/src/helpers/fee.rs
ddoktorski: Good point, I'll update it in the next PR with tip estimation

PR #3636, opened 2025-08-12
Add `--estimate-tip` flag

### LINKED ISSUE #3554: Suggested tip functionality
### Select Tool

sncast

### Requested Feature

Add functionality which allows to estimate tip. Estimation logic should be as follows:
- Get the latest block
- Take median of tips paid by all txs

ddoktorski:
Closes #3554


crates/sncast/src/helpers/fee.rs
```diff
@@ -82,8 +87,11 @@ impl FeeArgs {
             let mut fee_settings = FeeSettings::try_from(fee_estimate.clone())
                 .expect("Failed to convert FeeEstimate to FeeSettings");
 
-            // If a tip is not provided, set it to 0
-            fee_settings.tip = Some(self.tip.unwrap_or(0));
+            fee_settings.tip = if self.estimate_tip {
+                None // If we leave it as None, the tip will be estimated before sending the transaction
+            } else {
+                Some(self.tip.unwrap_or(0))
+            };
```
franciszekjob: nit: Maybe we can put it into separate method (e.g. `resolve_tip`) so it will be reused below as well.

franciszekjob (commented):
Could we add a test in which we use `--estimate-tip` flag and later check the tip of transaction (that it's not `0`)?

docs/src/appendix/sncast/multicall/run.md
```diff
@@ -63,7 +63,13 @@ Maximum L1 data gas unit price for the `invoke` transaction. When not used, defa
 ## `--tip <TIP>`
 Optional.
 
-Tip for the transaction. When not used, defaults to 0.
+Tip for the transaction. When not used, defaults 0 to unless [`--estimate-tip`](#--estimate-tip-estimate_tip) is used.
+
+## `--estimate-tip <ESTIMATE_TIP>`
+Optional.
+Conflicts with: [`--tip`](#--tip-tip)
```
cptartur, line 71: Let's add this to `--tip` as well?

docs/src/appendix/sncast/multicall/run.md
```diff
@@ -63,7 +63,13 @@ Maximum L1 data gas unit price for the `invoke` transaction. When not used, defa
...
+
+If passed, an estimated tip will be added to pay for the transaction. The tip is estimated based on the current network conditions and added to the transaction fee.
```
cptartur, line 73: Do we have any more details on how is it estimated? I think starknet rs probably uses block median, as we do. If so, maybe we should provide this information here?

crates/sncast/src/helpers/fee.rs
```diff
@@ -82,8 +87,11 @@ impl FeeArgs {
             let mut fee_settings = FeeSettings::try_from(fee_estimate.clone())
                 .expect("Failed to convert FeeEstimate to FeeSettings");
 
-            // If a tip is not provided, set it to 0
-            fee_settings.tip = Some(self.tip.unwrap_or(0));
+            fee_settings.tip = if self.estimate_tip {
+                None // If we leave it as None, the tip will be estimated before sending the transaction
```
cptartur: That's just handled by starknet.rs?

cptartur: If so, then please resolve

cptartur (commented):
Left some comments, but otherwise looks good

cptartur (commented):
Maybe some test with mocks validating that tip estimation is actually doing anything would be good?

docs/src/appendix/sncast/multicall/run.md
franciszekjob: We can but imo we imo aren't obliged to. This is SDK's responsibility, cast is higher level tool.

docs/src/appendix/sncast/multicall/run.md
franciszekjob: It was added in preceding PR

docs/src/appendix/sncast/multicall/run.md
ddoktorski: It just takes a median from a given block. We can include this information here, but I don't think this is critical for users

ddoktorski:
> Maybe some test with mocks validating that tip estimation is actually doing anything would be good?

Agree, but it's a bit problematic to mock a single call while serving the rest through the actual running devnet. I'll update if I manage to add it.

crates/sncast/src/helpers/fee.rs
```diff
@@ -117,11 +122,14 @@ pub struct FeeSettings {
 
 impl FeeSettings {
     #[must_use]
-    pub fn update_tip(&self, tip: u64) -> FeeSettings {
-        FeeSettings {
-            tip: Some(tip),
-            ..*self
-        }
+    pub fn resolve_tip(&self, tip: Option<u64>, estimate_tip: bool) -> FeeSettings {
```
franciszekjob: This makes the API more builder-like:
```suggestion
    pub fn with_resolved_tip(&self, tip: Option<u64>, estimate_tip: bool) -> FeeSettings {
```

CHANGELOG.md
```diff
@@ -25,6 +25,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 #### Added
 
 - `--tip` flag to `invoke`, `declare`, `deploy`, `multicall run` and `account deploy` commands to set the transaction tip
+- `--estimate-tip` flag to enable adding an estimate tip to the transaction. The tip is estimated based on the current network conditions and added to the transaction fee
```
```suggestion
- `--estimate-tip` flag which automatically adds an estimated tip to the transaction. The tip is calculated based on the current network conditions and added to the transaction fee
```

franciszekjob (approved):

docs/src/appendix/sncast/account/deploy.md
```diff
@@ -58,7 +58,13 @@ Maximum L1 data gas unit price for the `deploy_account` transaction. When not us
 ## `--tip <TIP>`
 Optional.
```
cptartur, line 61: Let's say it conflicts with estimate-tip

cptartur: In other docs as well

cptartur (approved):

PR #3638, opened 2025-08-12
ci: bump actions/checkout to v5

rejected-l:
GitHub-hosted runners now use Node 24, so actions/checkout@v5 is required. Minimum runner version v2.327.1. Workflows only updated—no functional changes.

See: https://github.com/actions/checkout/releases/tag/v5.0.0


cptartur (approved):

franciszekjob (approved):

cptartur:
Seems that we'll need to change how scheduled tests are run on PRs that modify their files. Unless you remove any modification to `scheduled.yml` file, this PR won't pass any tests for now (you'd need access to our private token we use to publish to scarbs, which we cannot share with external contributors).

ksew1 (approved):

PR #3639, opened 2025-08-12
Fix fuzzer attribute for generic types

ddoktorski:
Reported: https://t.me/starknet_foundry_support/6223

- After migrating to macros V2 the `fuzzer` attribute may generate code that contains syntax errors when using generic types

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Forge
+
+#### Added
+
+- A bug that caused `#[fuzzer]` attribute to fail when used with generic structs
+
 ## [0.48.0] - 2025-08-05
 
 ### Forge
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Forge
 
-#### Added
+#### Fixed
 
 - A bug that caused `#[fuzzer]` attribute to fail when used with generic structs
 
--- a/crates/forge/tests/e2e/fuzzing.rs
+++ b/crates/forge/tests/e2e/fuzzing.rs
@@ -36,7 +36,7 @@ fn fuzzing() {
         [PASS] fuzzing::tests::uint128_arg (runs: 256, [..]
         [PASS] fuzzing::tests::uint256_arg (runs: 256, [..]
         Running 0 test(s) from tests/
-        Tests: 12 passed, 1 failed, 0 ignored, 11 filtered out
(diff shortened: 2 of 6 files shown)


crates/snforge-scarb-plugin/src/attributes/fuzzer/wrapper.rs
```diff
@@ -76,11 +77,14 @@ fn fuzzer_wrapper_internal(
         let name = param.name(db).as_syntax_node();
         let name = SyntaxNodeWithDb::new(&name, db);
 
-        let name_type = param.type_clause(db).as_syntax_node();
+        let name_type = match param.type_clause(db) {
+            TypeClause(type_clause) => type_clause.ty(db).as_syntax_node(),
+            Empty(option_type_clause) => option_type_clause.as_syntax_node(),
```
piotmag769, line 82: unreachable even though it is allowed on type level xD

piotmag769 (approved):

crates/snforge-scarb-plugin/src/attributes/fuzzer/wrapper.rs
ddoktorski: Agree, but it's kind of safer to leave it this way. In fact passing an empty `::<>::` in the path below is still correct (assuming the type can be inferred somehow)

CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Forge
+
+#### Added
+
+- A bug that caused `#[fuzzer]` attribute to fail when used with generic structs
```
franciszekjob: Shouldn't it be under `Fixed`?

crates/forge/tests/e2e/fuzzing.rs
```diff
@@ -36,7 +36,7 @@ fn fuzzing() {
         [PASS] fuzzing::tests::uint128_arg (runs: 256, [..]
         [PASS] fuzzing::tests::uint256_arg (runs: 256, [..]
         Running 0 test(s) from tests/
-        Tests: 12 passed, 1 failed, 0 ignored, 11 filtered out
+        Tests: 12 passed, 1 failed, 0 ignored, 12 filtered out
```
franciszekjob, line 39: nit: Not related to this PR, but I think we can safely (?) change it to `Tests: 12 passed, 1 failed, 0 ignored, [..] filtered out` so adding a new test case won't require increasing the number of filtered tests.

franciszekjob (approved):

crates/forge/tests/e2e/fuzzing.rs
ddoktorski: Yeah, I think it can be safely replaced. If these tests need updating again, we can replace it then

PR #3645, opened 2025-08-14
Add events summary and signature lengths to trace file

THenry14:
Related to https://github.com/software-mansion/cairo-profiler/issues/91

## Introduced changes

- wite info about signatures and events into the trace file; required for l2 gas estimations in profiler

## Checklist

- [X] Linked relevant issue
- [ ] Added relevant tests

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,7 +45,7 @@ cairo-lang-starknet-classes = "2.12.0"
 cairo-lang-parser = "2.12.0"
 cairo-lang-sierra-to-casm = "2.12.0"
 cairo-vm = "2.2.0"
-cairo-annotations = "=0.5.0"
+cairo-annotations = "0.5.1"
 dirs = "6.0.0"
 dialoguer = "0.11.0"
 starknet-types-core = { version = "0.1.7", features = ["hash", "prime-bigint"] }
@@ -117,6 +117,3 @@ paste = "1.0.15"
 strum = "0.27"
 strum_macros = "0.27"
 scarb-oracle-hint-service = { git = "https://github.com/software-mansion/scarb", rev = "210da8dfd0b370f0f1970b33a373f1a7afe6ae33" }
-
-[patch.crates-io]
-cairo-annotations = { git = "https://github.com/software-mansion/cairo-annotations.git", branch = "szymczyk/l2-gas" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -117,3 +117,6 @@ paste = "1.0.15"
 strum = "0.27"
 strum_macros = "0.27"
 scarb-oracle-hint-service = { git = "https://github.com/software-mansion/scarb", rev = "210da8dfd0b370f0f1970b33a373f1a7afe6ae33" }
+
+[patch.crates-io]
+cairo-annotations = { git = "https://github.com/software-mansion/cairo-annotations.git", branch = "szymczyk/l2-gas" }
--- a/crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/entry_point.rs
+++ b/crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/entry_point.rs
@@ -77,7 +77,7 @@ pub fn execute_call_entry_point(
     // We skip recursion depth validation here.
     cheatnet_state
         .trace_data
-        .enter_nested_call(entry_point.clone(), cheated_data);
+        .enter_nested_call(entry_point.clone(), cheated_data.clone());
 
     if let Some(cheat_status) = get_mocked_function_cheat_status(entry_point, cheatnet_state)
         && let CheatStatus::Cheated(ret_data, _) = (*cheat_status).clone()
@@ -94,6 +94,7 @@ pub fn execute_call_entry_point(
             },
             &[],
             None,
+            vec![],
             vec![],
         );
         let tracked_resource = *context
@@ -179,14 +180,6 @@ pub fn execute_call_entry_point(
         .pop()
         .expect("Unexpected empty tracked resource.");
 
-    let entrypoint_address = entry_point.code_address.unwrap_or_default();
-    let signature = cheatnet_state
-        .get_cheated_execution_info_for_contract(entrypoint_address)
-        .tx_info
-        .signature
-        .as_value()
-        .unwrap_or_default();
-
     // region: Modified blockifier code
     match evaluate_execution_result(
         result,
@@ -208,7 +201,7 @@ pub fn execute_call_entry_point(
                 context,
                 cheatnet_state,
                 vm_trace,
-                signature,
+                cheated_data.tx_info.signature.unwrap_or_default(),
             );
             Ok(call_info)
         }
@@ -353,6 +346,7 @@ fn remove_syscall_resources_and_exit_non_error_call(
         &call_info.execution.l2_to_l1_messages,
         vm_trace,
         signature,
+        call_info.execution.events.clone(),
     );
 }
 
@@ -374,6 +368,7 @@ fn exit_error_call(
         CallResult::from_err(error, &identifier),
         &[],
         vm_trace,
+        vec![],
         vec![],
     );
 }
--- a/crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/entry_point.rs
+++ b/crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/entry_point.rs
@@ -94,6 +94,7 @@ pub fn execute_call_entry_point(
             },
             &[],
             None,
+            vec![],
         );
         let tracked_resource = *context
             .tracked_resource_stack
@@ -178,6 +179,14 @@ pub fn execute_call_entry_point(
         .pop()
         .expect("Unexpected empty tracked resource.");
 
+    let entrypoint_address = entry_point.code_address.unwrap_or_default();
+    let signature = cheatnet_state
+        .get_cheated_execution_info_for_contract(entrypoint_address)
+        .tx_info
+        .signature
+        .as_value()
+        .unwrap_or_default();
+
     // region: Modified blockifier code
     match evaluate_execution_result(
         result,
@@ -199,6 +208,7 @@ pub fn execute_call_entry_point(
                 context,
                 cheatnet_state,
                 vm_trace,
+                signature,
             );
             Ok(call_info)
         }
@@ -302,6 +312,7 @@ fn remove_syscall_resources_and_exit_non_error_call(
     context: &mut EntryPointExecutionContext,
     cheatnet_state: &mut CheatnetState,
     vm_trace: Option<Vec<RelocatedTraceEntry>>,
+    signature: Vec<Felt>,
 ) {
     let versioned_constants = context.tx_context.block_context.versioned_constants();
     // We don't want the syscall resources to pollute the results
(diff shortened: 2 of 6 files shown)


Cargo.toml
```diff
@@ -115,3 +115,6 @@ derive_more = { version = "2.0.1", features = ["display"] }
 paste = "1.0.15"
 strum = "0.27"
 strum_macros = "0.27"
+
+[patch.crates-io]
+cairo-annotations = { git = "https://github.com/software-mansion/cairo-annotations.git", branch = "szymczyk/l2-gas" }
```
THenry14: will remove after new version is released
https://github.com/software-mansion/cairo-annotations/pull/95

THenry14: https://github.com/foundry-rs/starknet-foundry/pull/3645/commits/bba8d464db4d5c9c67c277e036a5f18e49fa0ed0

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/entry_point.rs
```diff
@@ -178,6 +179,14 @@ pub fn execute_call_entry_point(
         .pop()
         .expect("Unexpected empty tracked resource.");
 
+    let entrypoint_address = entry_point.code_address.unwrap_or_default();
```
ddoktorski: I'm not familiar with the `code_address` field as we usually use `storage_address` to get the contract address. Do you happen to know the difference between these two fields and what the purpose of `code_address` is?

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/entry_point.rs
```diff
@@ -178,6 +179,14 @@ pub fn execute_call_entry_point(
...
+    let signature = cheatnet_state
+        .get_cheated_execution_info_for_contract(entrypoint_address)
+        .tx_info
+        .signature
+        .as_value()
+        .unwrap_or_default();
```
ddoktorski: Couldn't we get this information from the `cheated_data` variable?

crates/forge-runner/src/running.rs
```diff
@@ -363,19 +365,22 @@ pub fn run_test_case(
...
-            },
-            call_trace: call_trace_ref,
-            gas_used,
-            used_resources,
-            encountered_errors,
-            fuzzer_args,
-            fork_data,
-        })),
+        Ok(result) => {
+            collect_and_assign_events(&result, &call_trace_ref);
```
ddoktorski: Would it be possible to save events data in the call trace during the execution of `remove_syscall_resources_and_exit_non_error_call`, similar to how other necessary data is handled? It would be better to avoid adding profiler specific logic to the core forge test runner function

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/entry_point.rs
THenry14: AFAIK these two are the same in most cases - the exceptions being a library call (there is just no address in such case) and for calls like deploy (duh), then code_address is being set to None, where storage_address is not.

Why is that exactly I can only guess (maybe storage address just stores an address for calls?)

That being said, since it does not matter to me that much (I only need it for external calls) and as you correctly pointed out in the other comment I can just use cheated_data to get current signature, this is no longer in the codebase.

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/entry_point.rs
THenry14: Woah, I don't know how I missed it, it's way less convoluted that way - thank you. Changed.

crates/forge-runner/src/running.rs
THenry14: Hmm, it is possible, but I loose events emitted by top level entrypoint (snforge). 

And now thinking about it, it maybe is not that bad thing either, as per our conversation a few days back, we should estimate nested_calls l2 gas only, so I won't need it either way. It also simplifies things and assures the segragation you mention. 

Changed in the newest.

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/entry_point.rs
```diff
@@ -199,6 +201,7 @@ pub fn execute_call_entry_point(
                 context,
                 cheatnet_state,
                 vm_trace,
+                cheated_data.tx_info.signature.unwrap_or_default(),
```
ddoktorski, line 204: I'm not sure it's a good idea to consider the cheated signature in gas calculation. In forge it serves a slightly different purpose and is never used for gas computation. I'd expect this approach would incorrectly count the signature for inner calls if a cheated signature exists for the contract, whereas it should only apply (if at all) to the outer call. Personally I recommend not using this cheated data for gas calculation for the following reasons:
* The signature has a minimal impact on gas costs
* It's not calculated on forge's side, leading to misalignment between profiler and forge results
* Cheated data generally shouldn't be part of gas calculation, for instance should a mock call be treated the same way as a call contract syscall?

THenry14: The problem is, that snforge does not support transactions at all (for snforge, the test is a transaction, and that's it). Because of that, you will never be able to include signature to any gas calculations (because you do not sign the test using the cheatcode - and that is correct, because why would you?). So what happens, is snforge just sends an empty signature to blockifier to estimate l2 gas.

This behaviour was discussed with starkware, and the decision was to include cheated signatures (if provided; it kind of simulates the skip validate feature by default, if the cheat is not used).

> I'd expect this approach would incorrectly count the signature for inner calls if a cheated signature exists for the contract, whereas it should only apply (if at all) to the outer call.

This actually depends on which cheat you're using - if you're cheating globally then yes, but this is somewhat expected. If you're cheating per address, some inner calls can also have the signature (because you're cheating the address, and they can be called from within that address - this is basically what the library call is). This behaviour is not a problem in my case.

to you points:

- signature costs 5120 * signature length of gas, so it may not be that insignificant
- it is not, but this behaviour distorts the l2 gas estimations; either way it also could be included if you wanted to (not sure waht was the rationale originally to not include it)
- mock call is something rather different; you cannot get signature any other way in forge, as I described in a paragraphs above.

ddoktorski: > This actually depends on which cheat you're using - if you're cheating globally then yes, but this is somewhat expected. If you're cheating per address, some inner calls can also have the signature (because you're cheating the address, and they can be called from within that address - this is basically what the library call is). This behaviour is not a problem in my case.

I believe inner calls shouldn't include a signature, as it should be present only once per transaction. Could you confirm that?

> signature costs 5120 * signature length of gas, so it may not be that insignificant

I think this only confirms that the cost is negligible, given that the signature is usually of length 2

> not sure waht was the rationale originally to not include it

These cheats are meant only to simulate what execution info returns. They shouldn't be used for anything else

THenry14: > I believe inner calls shouldn't include a signature, as it should be present only once per transaction. Could you confirm that?

Yes - I did not see any inner call that has the signature set (BUT - as said previously - the exception here being a library call, but it's not a problem for me)

> I think this only confirms that the cost is negligible, given that the signature is usually of length 2

Depends on the test, in my opinion. Sometimes it might be negligible, sometimes not.

> These cheats are meant only to simulate what execution info returns. They shouldn't be used for anything else

I see. Well, I still think it makes sense to include this in the profiler estimations if it's present, as there is no other way in snforge to have this info

cptartur: My concern is that these estimations will be mostly misleading. On the real network, afaik the only signature is on the top level call (the one we don't really have in foundry), all internal calls do not incur any cost of signatures, validations etc.
Trying to create a logic that will include these where we don't really have this concept in our execution will be weird, especially in the cases where different nested calls have different signatures cheated. Will we include costs of all of these in such case?

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/entry_point.rs
```diff
@@ -77,7 +77,7 @@ pub fn execute_call_entry_point(
     // We skip recursion depth validation here.
     cheatnet_state
         .trace_data
-        .enter_nested_call(entry_point.clone(), cheated_data);
+        .enter_nested_call(entry_point.clone(), cheated_data.clone());
```
cptartur, line 80: Why do we clone here?

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/entry_point.rs
THenry14: The convention (and I think this was also agreed to be on your side as well) was that each nested call should be treated as separate transaction (by nested call I mean children of top level SNFORGE_TEST_CODE entrypoint, not children of those children), so I think signature per nested call fits in perfectly in this convention.

> we don't really have this concept in our execution will be weird

wait, I don't think I follow - you do have a concept of signature, you allow to set it per contract address/per specific invokation/globally. or did you have something else in mind?

> especially in the cases where different nested calls have different signatures cheated. Will we include costs of all of these in such case?

yes, and in profiler each nested call will have its l2 gas cost that includes its specific signature

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/entry_point.rs
THenry14: Because otherwise this moves the `cheated_data`, and I need it later on to get the signature

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/entry_point.rs
ddoktorski: > Yes - I did not see any inner call that has the signature set (BUT - as said previously - the exception here being a library call, but it's not a problem for me)

What I was referring to is that, for example in our template project if you add a cheat signature for the contract, it also gets applied to the `get_balance` call, which seems clearly unintended in that context. Of course, you have full control on the profiler side, so we can capture everything and leave it to the profiler to decide how to handle it, but it would be better to avoid cases that are clearly incorrect by definition to prevent unnecessary problems.

ddoktorski (approved):

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/entry_point.rs
cptartur: Cheated data is not that small, but the performance/resources cost of cloning is probably acceptable here.

crates/forge-runner/src/build_trace_data.rs
```diff
@@ -344,3 +356,13 @@ pub fn save_trace_data(
         .context("Failed to write call trace to a file")?;
     Ok(dir_to_save_trace.join(&filename))
 }
+
+fn to_summed_up_events(events: &[OrderedEvent]) -> Vec<SummedUpEvent> {
```
cptartur, line 360: Could this be implemented as `From` trait somewhere?

THenry14: cairo annotations would have to have a dependency on blockifier, which is an overkill to be honest, and it cannot be done on forge side

cptartur (approved):

PR #3649, opened 2025-08-18
Support cheats in `meta_tx_v0`

### LINKED ISSUE #3575: Support cheats in `meta_v0_tx` syscall
### Current State

Currently it executes on blockifier directly so it doesn't support cheating.

### Objective

Support cheats in `meta_v0_tx` syscall

### Additional Context

It needs to be moved to our cheatable runtime and handled there explicitly.

MKowalski8:
Closes #3575 

## Introduced changes

- Implement handling of` meta_tx_v0 `syscall in `snforge`, ensuring that cheat codes are properly supported and can be executed during tests.

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/cheatnet/tests/cheatcodes/meta_tx_v0.rs
```diff
@@ -0,0 +1,169 @@
...
+
+    let contract_address = test_env.deploy("CheatCallerAddressCheckerMetaTxV0", &[]);
+
+    test_env
+        .cheatnet_state
+        .start_cheat_caller_address(contract_address, ContractAddress::from(123_u32));
+
+    let meta_contract = test_env.deploy("MetaTxV0Test", &[]);
+
+    let signature: Vec<Felt> = vec![];
```
franciszekjob: nit: (same for other occurrences below)
```suggestion
    let signature = vec![];
```

crates/cheatnet/tests/cheatcodes/meta_tx_v0.rs
franciszekjob, line 1: Not sure if it's 100% needed, but we could check it with other cheatcodes for the sake of consistency?

crates/cheatnet/tests/contracts/src/meta_tx_v0/checker.cairo
```diff
@@ -0,0 +1,51 @@
...
+
+#[starknet::contract]
+mod MetaTxV0Test {
+    use starknet::syscalls::meta_tx_v0_syscall;
+
+    #[storage]
+    struct Storage {}
+
+    #[constructor]
+    fn constructor(ref self: ContractState) {}
```
franciszekjob: nit: Could be removed
```suggestion
```

crates/forge/tests/data/contracts/meta_tx_v0_test.cairo
```diff
@@ -0,0 +1,51 @@
...
+
+#[starknet::contract]
+mod MetaTxV0Test {
+    use starknet::syscalls::meta_tx_v0_syscall;
+
+    #[storage]
+    struct Storage {}
+
+    #[constructor]
+    fn constructor(ref self: ContractState) {}
```
franciszekjob: nit: Same as above
```suggestion
```

crates/forge/tests/integration/meta_tx_v0.rs
```diff
@@ -0,0 +1,479 @@
...
+                let signature = ArrayTrait::new();
+
+                let old_block_hash = meta_dispatcher.execute_meta_tx_v0_get_block_hash(checker_address, block_number, signature.span());
+
+                let cheated_hash = 555;
+                start_cheat_block_hash(checker_address, block_number, cheated_hash);
+
+                let meta_result = meta_dispatcher.execute_meta_tx_v0_get_block_hash(checker_address, block_number, signature.span());
+
+                assert(meta_result == 555, 'Should see cheated hash');
```
franciszekjob: nit:
```suggestion
                assert(meta_result == cheated_hash, 'Should see cheated hash');
```

crates/forge/tests/integration/meta_tx_v0.rs
```diff
@@ -0,0 +1,479 @@
...
+                let signature = ArrayTrait::new();
+
+                let old_block_number = meta_dispatcher.execute_meta_tx_v0(checker_address, signature.span());
+
+                let cheated_block_number = 999;
+                start_cheat_block_number(checker_address, cheated_block_number);
+
+                let meta_result = meta_dispatcher.execute_meta_tx_v0(checker_address, signature.span());
+
+                assert(meta_result == 999, 'Should see cheated block');
```
franciszekjob: nit:
```suggestion
                assert(meta_result == cheated_block_number, 'Should see cheated block');
```

crates/cheatnet/tests/contracts/src/meta_tx_v0/checker.cairo
```diff
@@ -0,0 +1,51 @@
...
+        }
+
+        fn execute_meta_tx_v0_get_block_hash(
+            self: @ContractState,
+            target: starknet::ContractAddress,
+            block_number: u64,
+            signature: Span<felt252>,
+        ) -> felt252 {
+            let selector = selector!("__execute__");
+            let calldata: Array<felt252> = array![block_number.into()];
```
franciszekjob: nit: Same for other occurrences below
```suggestion
            let calldata = array![block_number.into()];
```

franciszekjob (commented):
Looking good, left some nits

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/cheated_syscalls.rs
```diff
@@ -228,6 +243,149 @@ pub fn call_contract_syscall(
...
+) -> SyscallResult<ReadOnlySegment> {
+    if syscall_handler.base.context.execution_mode == ExecutionMode::Validate {
+        //region: Modified blockifier code
+        return Err(SyscallExecutorBaseError::InvalidSyscallInExecutionMode {
+            syscall_name: "meta_tx_v0".to_string(),
+            execution_mode: ExecutionMode::Validate,
+        }
+        .into());
+        // endregion
+    }
```
ddoktorski: This if statement should never occur as the execution mode is hardcoded to [`Execute`](https://github.com/foundry-rs/starknet-foundry/blob/0bab7179b49e3c6460307c0fcc70340f9ad68e0c/crates/runtime/src/starknet/context.rs#L110). Although I'm not sure whether to remove it or not, maybe it makes sense to keep it for the sake of consistency

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/cheated_syscalls.rs
```diff
@@ -228,6 +243,149 @@ pub fn call_contract_syscall(
...
+        &mut entry_point,
+        vm,
+        syscall_handler,
+        cheatnet_state,
+        remaining_gas,
+    )
+    .map_err(|error| {
+        SyscallExecutionError::from_self_or_revert(error.try_extract_revert().map_original(
+            |error| {
+                // TODO(lior): Change to meta-tx specific error.
```
ddoktorski: Let's omit this comment

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/cheated_syscalls.rs
```diff
@@ -228,6 +243,149 @@ pub fn call_contract_syscall(
...
+    // region: Modified blockifier code
+    // Call the cheatcode-aware meta_tx_v0 implementation
+    let retdata_segment = meta_tx_v0(
+        syscall_handler,
+        vm,
+        cheatnet_state,
+        request.contract_address,
+        request.entry_point_selector,
+        request.calldata,
+        request.signature,
```
ddoktorski, line 263: We should somehow highlight that this signature is not validated in any way

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/cheated_syscalls.rs
```diff
@@ -228,6 +243,149 @@ pub fn call_contract_syscall(
     // endregion
 }
 
+// blockifier/src/execution/syscalls/hint_processor.rs:645 (meta_tx_v0)
```
cptartur: I've checked and this was probably copied from `main` branch of blockifier and we should have probably used `main-v0.14.0` so it's inline with the version we are using.

The differences are minimal, but please verify/update it.

crates/forge/tests/data/contracts/meta_tx_v0_checkers.cairo
```diff
@@ -0,0 +1,127 @@
...
+        fn __execute__(ref self: ContractState) -> felt252 {
+            starknet::get_caller_address().into()
+        }
+    }
+}
+
+#[starknet::interface]
+trait ICheatBlockNumberCheckerMetaTxV0<TContractState> {
+    fn __execute__(ref self: TContractState) -> felt252;
+}
```
ddoktorski: This interface can be defined only once with some common name

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/cheated_syscalls.rs
```diff
@@ -228,6 +243,149 @@ pub fn call_contract_syscall(
...
+    remaining_gas: &mut u64,
+) -> SyscallResult<ReadOnlySegment> {
+    if syscall_handler.base.context.execution_mode == ExecutionMode::Validate {
+        //region: Modified blockifier code
+        return Err(SyscallExecutorBaseError::InvalidSyscallInExecutionMode {
+            syscall_name: "meta_tx_v0".to_string(),
+            execution_mode: ExecutionMode::Validate,
+        }
+        .into());
+        // endregion
```
cptartur: Why did we have to modify this region?

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/cheated_syscalls.rs
```diff
@@ -228,6 +243,149 @@ pub fn call_contract_syscall(
...
+            only_query,
+        },
+        max_fee: Fee(0),
+    });
+    syscall_handler.base.context.tx_context = Arc::new(TransactionContext {
+        block_context: old_tx_context.block_context.clone(),
+        tx_info: new_tx_info,
+    });
+
+    // region: Modified blockifier code
```
cptartur, line 361: What did we modify here?

crates/forge/tests/integration/meta_tx_v0.rs
ddoktorski, line 1: I'm not sure we need so many tests for this feature. Since it's already well covered in the cheatnet crate I would prefer to limit the number of integration tests here

crates/cheatnet/tests/cheatcodes/meta_tx_v0.rs
```diff
@@ -0,0 +1,169 @@
...
+        "execute_meta_tx_v0",
+        &[contract_address.into_(), signature.len().into()]
+            .into_iter()
+            .chain(signature.clone())
+            .collect::<Vec<_>>(),
+    );
+
+    assert_success(meta_result, &[Felt::ZERO]);
+
+    assert_success(direct_execute_result, &[Felt::from(3_u32)]);
```
cptartur, line 168: Why do we assert that?

crates/forge/tests/integration/meta_tx_v0.rs
MKowalski8: Okay, then I can leave two or three tests here

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/cheated_syscalls.rs
MKowalski8: The orginal code use `self` of syscall handler which have private method with this `Err`. I chosed the easiest way, to just copy this one `return`. So basically, there is no change in the logic.

crates/cheatnet/tests/cheatcodes/meta_tx_v0.rs
MKowalski8: I added two `asserts` here, one for the version called with `meta_v0_tx` and one direct, because I wanted to make sure that this zero is not due to some error, and not necessarily due to substitution by our `syscall`

MKowalski8: My first idea was to get two values, version and caller address (to check if this was called correctly), but there were some problems with serialization and deserialization etc. 
This is just an easier aproach.

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/cheated_syscalls.rs
MKowalski8: I use our own `execute_inner_call` with `cheatnet_state`

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/cheated_syscalls.rs
MKowalski8: I see, but as you said differences are minimal. I'll update it

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/cheated_syscalls.rs
MKowalski8: Do you have any idea how to do it?

crates/cheatnet/tests/cheatcodes/meta_tx_v0.rs
cptartur: Okay, makes sense.

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/cheated_syscalls.rs
MKowalski8: Since we are based on the original code from `blockifier`, I would leave it. But @cptartur or @franciszekjob can say what they think.

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/cheated_syscalls.rs
ddoktorski: It wouldn't be ideal to always display a warning when this syscall is used, so at a minimum we can mention it somewhere in the documentation cc @cptartur

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/cheated_syscalls.rs
cptartur: We can technically just have `unreachable!(...)` in the if branch so it shows that blockifier supports it but it cannot occur for us. And we'll get a nice error if it is ever reached somehow.

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/cheated_syscalls.rs
cptartur: If someone called an (account) contract that actually validates this signature, I think it would get verified, no? I think it happens at contract level, not protocol so we wouldn't be verifying it anyway.

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/cheated_syscalls.rs
franciszekjob: Same take, let's use `unreachable!(...)`.

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/cheated_syscalls.rs
ddoktorski: > If someone called an (account) contract that actually validates this signature, I think it would get verified, no?

I don't think so, but that's a good question what actually happens with the signature in this syscall. I would expect that it's used for validation

ddoktorski: It seems that Cairo 0 accounts didn't have a separate entry point for validate, but instead performed validation as part of the execute call, which I wasn't aware of. If that's the case, no further action is required and this thread can be resolved

ddoktorski (approved):
As discussed previously let's confirm with the OZ team whether everything works as expected with this syscall on their end

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/cheated_syscalls.rs
cptartur: I think at least some of them didn't have that. Not sure if `__validate__` was introduced for cairo0 at some point or did it only happen in cairo 1.

cptartur (approved):

PR #3664, opened 2025-08-22
Update gas pointer in `CallToBlockifierRuntime`

### LINKED ISSUE #3658: Update remaining gas in `CallToBlockifierRuntime`
### Current State

Not updated atm

### Objective

Update remaining gas in `CallToBlockifierRuntime`

### Additional Context

_No response_

ddoktorski:
Closes #3658

commit-id:5ba0cf9f

---

**Stack**:
- #3672
- #3664 ⬅

### DIFF
--- a/crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/mod.rs
+++ b/crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/mod.rs
@@ -1,24 +1,24 @@
 use std::marker::PhantomData;
 
+use crate::state::CheatnetState;
 use blockifier::execution::entry_point::{CallEntryPoint, CallType};
-use blockifier::execution::execution_utils::felt_from_ptr;
-use blockifier::execution::syscalls::hint_processor::SyscallHintProcessor;
+use blockifier::execution::syscalls::hint_processor::{OUT_OF_GAS_ERROR, SyscallHintProcessor};
 use blockifier::execution::syscalls::syscall_executor::SyscallExecutor;
 use blockifier::execution::syscalls::vm_syscall_utils::{
     CallContractRequest, LibraryCallRequest, RevertData, SingleSegmentResponse,
-    SyscallRequestWrapper, SyscallSelector,
+    SyscallExecutorBaseError, SyscallRequestWrapper, SyscallSelector,
 };
 use blockifier::execution::{
     execution_utils::ReadOnlySegment,
     syscalls::vm_syscall_utils::{SyscallRequest, SyscallResponse, SyscallResponseWrapper},
 };
+use blockifier::utils::u64_from_usize;
 use cairo_vm::types::relocatable::MaybeRelocatable;
 use cairo_vm::vm::{errors::hint_errors::HintError, vm_core::VirtualMachine};
-use runtime::{ExtendedRuntime, ExtensionLogic, SyscallHandlingResult, SyscallPtrAccess};
+use runtime::{ExtendedRuntime, ExtensionLogic, SyscallHandlingResult};
 use starknet_api::contract_class::EntryPointType;
 use starknet_api::core::ContractAddress;
-
-use crate::state::CheatnetState;
+use starknet_types_core::felt::Felt;
 
 use crate::runtime_extensions::call_to_blockifier_runtime_extension::rpc::{
     AddressOrClassHash, call_entry_point,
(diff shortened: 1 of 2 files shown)


ddoktorski:
Note: This change hasn’t been tested in detail at this point, aside from a basic sanity check comparing `gas_consumed` in the trace for the same test before and after the change. The comparison was not thorough, it only verified whether the gas value increased significantly, which it did. This is expected as the cost of the `call_contract_syscall` is now included, whereas it was not tracked previously.

Proper testing will be done as part of #3292. It is expected to work as intended if the current gas values in tests do not require changes or only minor adjustments.

franciszekjob (approved):

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/mod.rs
```diff
@@ -151,24 +142,55 @@ impl ExecuteCall for LibraryCallRequest {
 }
 
 fn execute_syscall<Request: ExecuteCall + SyscallRequest>(
+    selector: SyscallSelector,
     vm: &mut VirtualMachine,
     cheatable_starknet_runtime: &mut CheatableStarknetRuntime,
 ) -> Result<(), HintError> {
-    let _selector = felt_from_ptr(vm, cheatable_starknet_runtime.get_mut_syscall_ptr())?;
+    let syscall_handler = &mut cheatable_starknet_runtime.extended_runtime.hint_handler;
```
cptartur, line 151: Was the new code copied from somewhere? If so we might want to link to some kind of source.

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/mod.rs
```diff
@@ -151,24 +142,55 @@ impl ExecuteCall for LibraryCallRequest {
...
+    let syscall_base_cost = syscall_handler.get_syscall_base_gas_cost();
+
+    // Sanity check for preventing underflow.
+    assert!(
+        syscall_gas_cost >= syscall_base_cost,
+        "Syscall gas cost must be greater than base syscall gas cost"
+    );
+
+    // Refund `SYSCALL_BASE_GAS_COST` as it was pre-charged.
+    let required_gas = syscall_gas_cost - syscall_base_cost;
```
cptartur, line 179: Do we have the logic for precharging?

ddoktorski: This should be handled by compiler: https://github.com/starkware-libs/sequencer/blob/v0.15.0-rc.2/crates/blockifier/src/blockifier_versioned_constants.rs#L1057. Note that this is also the minimal possible value

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/mod.rs
ddoktorski: This is basically [execute_syscall](https://github.com/starkware-libs/sequencer/blob/v0.15.0-rc.2/crates/blockifier/src/execution/syscalls/vm_syscall_utils.rs#L677C1-L677C19) function with modifications. We have some duplicated code with `CheatableStarknetRuntime`, but I believe it's easier long term to maintain it as separate functions

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/mod.rs
cptartur: You mean the precharging is included at sierra level by the compiler?

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/mod.rs
cptartur: The linked function definitely looks copied from blockifier so if you can pinpoint some kind of source for this and link it it would be nice. But if that would take more than a super quick search then disregard my comment.

cptartur (commented):
Replied to comments

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/mod.rs
ddoktorski: Yes, that's how I understand it

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/mod.rs
ddoktorski: Done

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/mod.rs
cptartur: Maybe let's include that in the comment

cptartur (approved):
Left one reply, but otherwise looks good

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/mod.rs
ddoktorski: Will add it in the following PR

PR #3666, opened 2025-08-22
Implement basic Cairo Native integration

gabrielbosio:
See https://github.com/lambdaclass/starknet-foundry/pull/4 for further details.

PR #3667, opened 2025-08-22
Add initial structure for a Native compatible runtime

gabrielbosio:
See https://github.com/lambdaclass/starknet-foundry/pull/5 for further details.
Depends on #3666


cptartur:
Currently only migrating contract execution to native. Will be reopened once it is relevant.

PR #3668, opened 2025-08-22
Add CLI argument to run tests with Native

gabrielbosio:
See https://github.com/lambdaclass/starknet-foundry/pull/8 for further details.
Depends on #3667


cptartur:
Currently only migrating contract execution to native. Will be reopened once it is relevant.

Issue #1431, opened 2023-12-22
Parametrized tests MVP


Arcticae:

Arcticae:
Postponing because of other priorities

cptartur:
A new or updated design doc might be required to tackle this since the previous one is quite old.

PR #3669, opened 2025-08-22
Parametrized tests

franciszekjob:
Towards #1431

**Stack**:
- #3697
- #3669 ⬅
- #3671
- #3670

## Introduced changes

Introduce parametrized tests by addin `#[test_case]` attribute.

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/forge/tests/e2e/test_case.rs
```diff
@@ -0,0 +1,102 @@
...
+        [PASS] test_case_integrationtest::single_attribute::simple_addition_1_2 [..]
+        [PASS] test_case_integrationtest::single_attribute::simple_addition_3_4 [..]
+        [PASS] test_case_integrationtest::single_attribute::simple_addition_5_6 [..]
+        Running 0 test(s) from src/
+        Tests: 3 passed, 0 failed, 0 ignored, [..] filtered out
+        "},
+    );
+}
+
+// TODO: Use separate process (with oracle) instead of heavy fib calculation.
```
franciszekjob: note: Currently this tests uses same approach as the one in fuzzer - [`fuzzing_exit_first`](https://github.com/foundry-rs/starknet-foundry/blob/6c261a0652a3aff60b61368a8d2d56dca97bb3f0/crates/forge/tests/e2e/fuzzing.rs#L177). However, this is technically non-deterministic and ultimately we should use an external process (oracles) to make one of the tests fail before another.
If you agree with above, I will create and link an issue.

crates/forge/tests/data/test_case/tests/multiple_attributes.cairo
cptartur, line 1: I think we should also have an e2e contract test that uses `name:`.

crates/forge/tests/e2e/test_case.rs
cptartur: We can create one for that

crates/snforge-scarb-plugin/src/args/named.rs
```diff
@@ -24,6 +24,10 @@ impl DerefMut for NamedArgs {
 }
 
 impl NamedArgs {
+    pub fn new(map: &HashMap<SmolStr, Vec<Expr>>) -> Self {
+        Self(map.clone())
+    }
```
cptartur: IMO it's bit weird to hide a clone like this, I'd prefer an explicit clone at caller side.

crates/snforge-scarb-plugin/src/attributes/test_case/name.rs
```diff
@@ -0,0 +1,71 @@
...
+use crate::attributes::test_case::TestCaseCollector;
+use cairo_lang_macro::Diagnostics;
+use cairo_lang_parser::utils::SimpleParserDatabase;
+use cairo_lang_syntax::node::TypedSyntaxNode;
+use cairo_lang_syntax::node::ast::Expr;
+use regex::Regex;
+use std::sync::LazyLock;
+
+static RE_SANITIZE: LazyLock<Regex> =
+    LazyLock::new(|| Regex::new(r"[^a-zA-Z0-9]+").expect("Failed to create regex"));
```
cptartur, line 13: Can we even have this kind of invalid characters in the test name itself?

crates/snforge-scarb-plugin/src/attributes/test_case/name.rs
```diff
@@ -0,0 +1,71 @@
...
+
+fn sanitize_expr(expr: &Expr, db: &SimpleParserDatabase) -> String {
+    let expr_text = &expr.as_syntax_node().get_text(db);
+    let expr_sanitized = RE_SANITIZE
+        .replace_all(expr_text, "_")
```
cptartur, line 18: How does the replacement work exactly? If I have multiple consecutive invalid characters, will they be replaced with multiple `_` or a single one?

crates/snforge-scarb-plugin/src/attributes/test_case/name.rs
```diff
@@ -0,0 +1,71 @@
...
+
+    let exprs = unnamed_args
+        .iter()
+        .map(|(_, expr)| sanitize_expr(expr, db))
+        .collect::<Vec<_>>();
+
+    Ok(exprs.join("_"))
+}
+
+pub fn resolve_test_case_name(
```
cptartur: Nit but maybe just this? 

```suggestion
pub fn test_case_name(
```

crates/snforge-scarb-plugin/src/attributes/test_case/name.rs
```diff
@@ -0,0 +1,71 @@
...
+
+    let suffix = if let Some(expr) = test_case_name {
+        match expr {
+            Expr::ShortString(_) | Expr::String(_) => {}
+            _ => {
+                return Err(Diagnostics::from(TestCaseCollector::error(
+                    "Only string literals are allowed for 'name' argument.",
+                )));
+            }
+        }
```
cptartur, line 59: Maybe `assert_matches!` would be better than match with empty arms?

crates/snforge-scarb-plugin/src/attributes/test.rs
```diff
@@ -37,7 +37,21 @@ fn test_internal(
...
+
+    // If the function has #[test] attribute and does not have #[fuzzer], we can
+    // safely skip #[test].
+    if has_test_case && !has_fuzzer {
+        let func_item = func.as_syntax_node();
+        let func_item = SyntaxNodeWithDb::new(&func_item, db);
+
+        return Ok(quote!(
+            #func_item
+        ));
```
cptartur: What about generating the code required for blockifier entry point? 
What's the benefit of skipping this here?

crates/snforge-scarb-plugin/src/attributes/test_case.rs
```diff
@@ -0,0 +1,162 @@
...
+impl AttributeInfo for TestCaseCollector {
+    const ATTR_NAME: &'static str = "test_case";
+}
+
+#[must_use]
+pub fn test_case(args: TokenStream, item: TokenStream) -> ProcMacroResult {
+    into_proc_macro_result(args, item, test_case_handler)
+}
+
+fn test_case_handler(
```
cptartur: Nit but we usually named these `<something>_internal` or `internal_<something>`

crates/snforge-scarb-plugin/src/attributes/test_case.rs
```diff
@@ -0,0 +1,162 @@
...
+                #filtered_fn_attrs
+                fn #func_name #signature
+                #func_body
+            );
+
+            let call_args = unnamed_args
+                .clone()
+                .into_iter()
+                .map(|(_, expr)| expr.as_syntax_node().get_text(args_db))
+                .collect::<Vec<_>>()
```
cptartur: Is that safe to use in regards to macro diagnostics?

cc @piotmag769 @Draggu

crates/snforge-scarb-plugin/src/attributes/test_case.rs
```diff
@@ -0,0 +1,162 @@
...
+                .join(", ")
+                .to_string();
+            let call_args = format_ident!("({})", call_args);
+
+            let case_fn_name = format_ident!("{}", case_fn_name);
+
+            let out_of_gas = create_single_token("'Out of gas'");
+
+            Ok(quote!(
+                #[implicit_precedence(core::pedersen::Pedersen, core::RangeCheck, core::integer::Bitwise, core::ec::EcOp, core::poseidon::Poseidon, core::SegmentArena, core::circuit::RangeCheck96, core::circuit::AddMod, core::circuit::MulMod, core::gas::GasBuiltin, System)]
```
cptartur: We are basically reimplementing most of the logic form `#[test]` and we need them to be the same.
This should be shared somehow.

crates/snforge-scarb-plugin/src/attributes/test_case.rs
```diff
@@ -0,0 +1,162 @@
...
+                #[snforge_internal_test_executable]
```
cptartur: What if `#[test]` managed to execute already before? Won't that duplicate the executables?

crates/snforge-scarb-plugin/src/attributes/test_case.rs
```diff
@@ -0,0 +1,162 @@
...
+    func: &FunctionWithBody,
+    unnamed_args: &UnnamedArgs,
+    db: &SimpleParserDatabase,
+) -> Result<(), Diagnostics> {
+    let param_count = func
+        .declaration(db)
+        .signature(db)
+        .parameters(db)
+        .elements(db)
+        .len();
```
cptartur: This query could be reused, not sure how expensive it is though.

crates/snforge-scarb-plugin/src/attributes/test_case.rs
```diff
@@ -0,0 +1,162 @@
...
+        return Err(Diagnostics::from(TestCaseCollector::error(format!(
+            "Expected {} parameters, but got {}",
+            param_count,
+            unnamed_args.len()
+        ))));
+    }
+    Ok(())
+}
+
+fn get_filtered_func_attributes(
```
cptartur: If it's filtering out just test attr in case there is no fuzzer then maybe it can be named something more explicit.

crates/snforge-scarb-plugin/tests/integration/single_attributes/test_case.rs
cptartur, line 1: Can we also add some multiple attributes tests cases with at least fuzzer, test and test_case present? 

And please explicitly test the expansion of this order:
```
#[test]
#[fuzzer]
#[test_case(...)]
```

crates/forge/tests/e2e/test_case.rs
cptartur, line 1: Please also add/expand `e2e/plugin_diagnostics.rs` tests to check for diagnostics with `#[test_case]` attribute.

cptartur (changes requested):

crates/forge/tests/data/test_case/Scarb.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "test_case"
+version = "0.1.0"
+edition = "2024_07"
+
+# See more keys and their definitions at https://docs.swmansion.com/scarb/docs/reference/manifest.html
+
```
ddoktorski: nit:
```suggestion
```

crates/forge/tests/e2e/test_case.rs
ddoktorski: Given that it behaves deterministically in practice, could you clarify why it's a problem?

crates/snforge-scarb-plugin/src/attributes/test_case.rs
```diff
@@ -0,0 +1,162 @@
...
+pub struct TestCaseCollector;
+
+impl AttributeInfo for TestCaseCollector {
+    const ATTR_NAME: &'static str = "test_case";
+}
+
+#[must_use]
+pub fn test_case(args: TokenStream, item: TokenStream) -> ProcMacroResult {
+    into_proc_macro_result(args, item, test_case_handler)
+}
```
ddoktorski, line 28: Please refactor this code to make it consistent with the other attributes, such that the wrapper calls `into_proc_macro_result` passing `with_parsed_values`, which in turn calls the internal function 🙏

crates/snforge-scarb-plugin/src/attributes/test_case.rs
```diff
@@ -0,0 +1,162 @@
...
+) -> Result<TokenStream, Diagnostics> {
+    with_parsed_values::<TestCaseCollector>(
+        args,
+        item,
+        warns,
+        |func_db, func, args_db, arguments, _warns| {
+            let unnamed_args = arguments.unnamed();
+
+            ensure_params_present(func, func_db)?;
+            ensure_args_count_valid(func, &unnamed_args, func_db)?;
```
ddoktorski: Can we merge these checks into one function?

crates/forge/tests/data/test_case/tests/multiple_attributes.cairo
```diff
@@ -0,0 +1,44 @@
+use test_case::add;
+
+#[test]
+#[test_case(1, 2)]
+#[test_case(3, 4)]
+#[available_gas(l2_gas: 40000000)]
```
ddoktorski, line 6: Let's mix the order of attributes in these tests

ddoktorski (changes requested):

crates/forge/tests/data/test_case/tests/multiple_attributes.cairo
franciszekjob: Added in https://github.com/foundry-rs/starknet-foundry/pull/3669/commits/cd77408690e504d0bb8d5fda87d9af29165ae5e5

crates/snforge-scarb-plugin/src/attributes/test_case/name.rs
franciszekjob: In the test function name itself not, but keep in mind that test case params are included in the generated name. Then, if params contain e.g. structs, hence the textual representation of such expression will include, for instance, `{`.

crates/snforge-scarb-plugin/src/attributes/test_case/name.rs
franciszekjob: The regex uses `+`, so consecutive invalid characters are matched as a single group. That means they will be replaced with exactly one `_`, not multiple

crates/snforge-scarb-plugin/src/attributes/test_case/name.rs
franciszekjob: Could be, https://github.com/foundry-rs/starknet-foundry/pull/3669/commits/bd4af1f5971f3aa5c558b26d47a37799a404d621

crates/snforge-scarb-plugin/src/attributes/test_case.rs
franciszekjob: Good idea, https://github.com/foundry-rs/starknet-foundry/pull/3669/commits/7ac594f5aa4e2d9910212e27bd421ce46b2fa902

crates/snforge-scarb-plugin/src/attributes/test_case/name.rs
franciszekjob: Generally yes, but `assert_matches!` is unstable yet
https://doc.rust-lang.org/std/assert_matches/macro.assert_matches.html

crates/snforge-scarb-plugin/src/attributes/test_case.rs
franciszekjob: Not relevant anymore, as it's done only once thanks to refactored params validation.

crates/snforge-scarb-plugin/src/attributes/test_case.rs
franciszekjob: Maybe `collect_attrs_excluding_test_without_fuzzer`
https://github.com/foundry-rs/starknet-foundry/pull/3669/commits/fca165c816cfd92f4b7574d21d571f97d20445f8

crates/snforge-scarb-plugin/src/attributes/test.rs
franciszekjob: If `#[test]`precedes `#[test_case]`, then we simply get rid of `#[test]`. Later, the code required for blockifier entry point will be added later during `#[test_case`] resolution. Hope that explains it for you.

crates/snforge-scarb-plugin/src/attributes/test_case.rs
franciszekjob: See https://github.com/foundry-rs/starknet-foundry/pull/3669#discussion_r2352432290.
Current logic prevents from such situation.

crates/forge/tests/e2e/test_case.rs
franciszekjob: Added

crates/forge/tests/e2e/test_case.rs
franciszekjob: It's just something I noticed while writing tests and added a remark - it's not a problem, nor it's important.

crates/snforge-scarb-plugin/src/attributes/test_case.rs
franciszekjob: https://github.com/foundry-rs/starknet-foundry/pull/3669/commits/55fcf2cdf62c945b14959b4b19eb8092df1581c7

crates/forge/tests/data/test_case/tests/multiple_attributes.cairo
franciszekjob: Mixed, hope it's fine now.
Sidenote: it would be best if we created a permutation of attributes, because then we're sure that it would work for any order.

crates/snforge-scarb-plugin/tests/integration/single_attributes/test_case.rs
franciszekjob: https://github.com/foundry-rs/starknet-foundry/pull/3669/commits/fc7ea914c34ef228ef50bb8262a6ff13322b854d

crates/snforge-scarb-plugin/src/attributes/test_case/name.rs
cptartur: Interesting, I though we were already using it. Disregard then.

crates/snforge-scarb-plugin/src/attributes/test.rs
cptartur: Maybe let's improve the comment so it says that it is handled by `test_case` then? Because the current one makes it look like an optimization and not a part of the logic.

crates/snforge-scarb-plugin/src/attributes/test_case/name.rs
franciszekjob: We're using `assert_matches` from `snapbox`.

crates/snforge-scarb-plugin/src/attributes/test.rs
franciszekjob: https://github.com/foundry-rs/starknet-foundry/pull/3669/commits/c9194e75620e2ba7a5db415e6539c069249bcbf2

crates/snforge-scarb-plugin/src/args/named.rs
```diff
@@ -24,6 +24,10 @@ impl DerefMut for NamedArgs {
 }
 
 impl NamedArgs {
+    // pub fn new(map: &HashMap<SmolStr, Vec<Expr>>) -> Self {
+    //     Self(map.clone())
+    // }
```
cptartur: Maybe just remove it?

crates/snforge-scarb-plugin/src/attributes/test_case.rs
```diff
@@ -0,0 +1,130 @@
...
+
+    if param_count == 0 {
+        return Err(Diagnostics::from(TestCaseCollector::error(
+            "The function must have at least one parameter to use #[test_case] attribute",
+        )));
+    }
+
+    if param_count != unnamed_args.len() {
+        return Err(Diagnostics::from(TestCaseCollector::error(format!(
+            "Expected {} parameters, but got {}",
```
cptartur: Shouldn't it be like this?

```suggestion
            "Expected {} arguments, but got {}",
```

What's invalid is number of arguments user provided to `#[test_case]`, not arguments of a function. At least that's what makes more sense to me.

crates/snforge-scarb-plugin/tests/integration/multiple_attributes.rs
```diff
@@ -451,3 +451,73 @@ fn works_with_fuzzer_config_wrapper() {
...
+                    core::gas::withdraw_gas_all(core::gas::get_builtin_costs()), 'Out of gas',
+                );
+                test_add(1, 2, 3);
+                let mut arr = ArrayTrait::new();
+                core::array::ArrayTrait::span(@arr)
+            }
+
+            #[__fuzzer_config]
+            #[__fuzzer_wrapper]
+            fn test_add(x: i128, y: i128, expected: i128) {}
```
cptartur, line 547: Can we add steps for fuzzer_config and fuzzer_wrapper so we have a complete example?

crates/forge/tests/e2e/plugin_diagnostics.rs
cptartur, line 1: Could we modify the scarb projects used for some already existing tests in this file so they include `#[test_case]` attribute as well? If relevant to any.

crates/forge/tests/e2e/test_case.rs
cptartur: Let's link an issue before we merge

crates/snforge-scarb-plugin/src/args/named.rs
franciszekjob: Yes, leftover which I didn't notice

crates/forge/tests/e2e/test_case.rs
franciszekjob: I wonder if it's actually worth the effort. Will either create an issue or remove todo.

crates/snforge-scarb-plugin/tests/integration/multiple_attributes.rs
franciszekjob: https://github.com/foundry-rs/starknet-foundry/pull/3669/commits/642d77f5ffdfbb1ac5288fd09531af6b52209307

crates/snforge-scarb-plugin/src/attributes/test.rs
```diff
@@ -37,7 +37,21 @@ fn test_internal(
 ) -> Result<TokenStream, Diagnostics> {
     assert_is_used_once::<TestCollector>(db, func)?;
     args.assert_is_empty::<TestCollector>()?;
-    ensure_parameters_only_with_fuzzer_attribute(db, func)?;
+    ensure_parameters_only_with_fuzzer_or_test_case_attribute(db, func)?;
+
+    let has_test_case = has_test_case_attribute(db, func);
+    let has_fuzzer = has_fuzzer_attribute(db, func);
+
+    // If the function has `#[test]` attribute and does not have `#[fuzzer]`, we can
```
ddoktorski: Shouldn't it be test_case?
```suggestion
    // If the function has `#[test_case]` attribute and does not have `#[fuzzer]`, we can
```

crates/snforge-scarb-plugin/src/attributes/test_case.rs
```diff
@@ -0,0 +1,130 @@
...
+    let param_count = func
+        .declaration(db)
+        .signature(db)
+        .parameters(db)
+        .elements(db)
+        .len();
+
+    if param_count == 0 {
+        return Err(Diagnostics::from(TestCaseCollector::error(
+            "The function must have at least one parameter to use #[test_case] attribute",
```
cptartur, line 101:
```suggestion
            "The function must have at least one argument to use #[test_case] attribute",
```

To be consistent with below

cptartur (approved):

crates/snforge-scarb-plugin/src/attributes/test_case.rs
franciszekjob: Function (in the meaning of function definition) has parameters, not arguments

crates/snforge-scarb-plugin/src/attributes/test.rs
```diff
@@ -134,3 +139,32 @@ fn has_parameters(db: &SimpleParserDatabase, func: &FunctionWithBody) -> bool {
         .len()
         != 0
 }
+
+#[must_use]
+pub fn test_func_with_attrs(
+    test_fn_name_ident: &TokenTree,
+    fn_name_ident: &TokenTree,
+    call_args: &TokenTree,
```
ddoktorski: It would be better if `call_args` were a `TokenStream`. It could be also passed without parentheses

crates/snforge-scarb-plugin/src/attributes/test_case.rs
```diff
@@ -0,0 +1,130 @@
...
+    let case_fn_name = test_case_name(&func_name, &args, args_db)?;
+    let filtered_fn_attrs = collect_attrs_excluding_test_without_fuzzer(func, db);
+
+    let signature = func.declaration(db).signature(db).as_syntax_node();
+    let signature = SyntaxNodeWithDb::new(&signature, db);
+
+    let func_body = func.body(db).as_syntax_node();
+    let func_body = SyntaxNodeWithDb::new(&func_body, db);
+
+    let func_name = format_ident!("{}", func_name);
```
ddoktorski: Please use `format_ident!` only when introducing a new code, in this case we can make it a `SyntaxNode` as for other elements

crates/snforge-scarb-plugin/src/attributes/test_case.rs
```diff
@@ -0,0 +1,130 @@
...
+    );
+
+    let call_args = unnamed_args
+        .clone()
+        .into_iter()
+        .map(|(_, expr)| expr.as_syntax_node().get_text(args_db))
+        .collect::<Vec<_>>()
+        .join(", ")
+        .to_string();
+    let call_args = format_ident!("({})", call_args);
```
ddoktorski: Let's make it a `TokenStream` and omit parentheses  as mentioned in the other comment

crates/snforge-scarb-plugin/src/attributes/test_case.rs
```diff
@@ -0,0 +1,130 @@
...
+    let attr_list = func.attributes(func_db);
+    let has_fuzzer = has_fuzzer_attribute(func_db, func);
+
+    // We do not want to copy the `#[test]` attribute if there is no `#[fuzzer]`
+    attr_list
+        .elements(func_db)
+        .filter(|attr| {
+            let test_attr_text = format!("#[{}]", TestCollector::ATTR_NAME);
+            let attr_text = attr.as_syntax_node().get_text(func_db);
+            let attr_text = attr_text.trim();
```
ddoktorski, line 129: nit:
```suggestion
            let attr_text = attr.as_syntax_node().get_text(func_db).trim();
```

crates/snforge-scarb-plugin/src/attributes/test_case/name.rs
```diff
@@ -0,0 +1,66 @@
...
+fn sanitize_expr(expr: &Expr, db: &SimpleParserDatabase) -> String {
+    let expr_text = &expr.as_syntax_node().get_text(db);
+    let expr_sanitized = RE_SANITIZE
+        .replace_all(expr_text, "_")
+        .to_lowercase()
+        .trim_matches('_')
+        .to_string();
+
+    if expr_sanitized.is_empty() {
+        "_empty".into()
```
ddoktorski, line 24: Can we have multiple names with `_empty`?

crates/snforge-scarb-plugin/src/attributes/test.rs
franciszekjob: Good catch 👁️

crates/snforge-scarb-plugin/src/attributes/test_case.rs
franciszekjob: We need binding here, otherwise compiler complains

crates/snforge-scarb-plugin/src/attributes/test_case/name.rs
franciszekjob: Such situation could occur only in a scenario where multiple variables of same type and different values procude same sanitized expr.

For instance, following Cairo code
```rust
#[test_case("")]
#[test_case("!!!")]
fn test_str(x: ByteArray) {}
```

gives such diagnostic

image


However, this is consistent with the original Rust `test_case`, example below:
image

Considering that, I think we can leave it as it is. If someone has such problem, they can simply name the case.

ddoktorski: I guess it's fine, @cptartur any thoughts on this?

cptartur: I think it's fine, maybe we could add some note on duplication names to docs.

crates/forge/tests/data/diagnostics/test_case_attr/Scarb.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "test_case"
+version = "0.1.0"
+edition = "2024_07"
+
+[dependencies]
+starknet = "2.9.3"
```
ddoktorski, line 7: nit:
```suggestion
starknet = "2.12.0"
```

crates/snforge-scarb-plugin/src/attributes/test_case.rs
```diff
@@ -0,0 +1,136 @@
...
+}
+
+fn args_to_token_stream(args: &UnnamedArgs, db: &SimpleParserDatabase) -> TokenStream {
+    let args_str = args
+        .iter()
+        .map(|(_, expr)| expr.as_syntax_node().get_text(db))
+        .collect::<Vec<_>>()
+        .join(", ")
+        .to_string();
+    TokenStream::new(vec![format_ident!("{}", args_str)])
```
ddoktorski: nit:
```suggestion
    args.iter()
        .map(|(_, expr)| {
            let expr = expr.as_syntax_node();
            let expr = SyntaxNodeWithDb::new(&expr, db);
            quote! { #expr, }
        })
        .fold(TokenStream::empty(), |mut acc, token| {
            acc.extend(token);
            acc
        })
```

crates/snforge-scarb-plugin/src/attributes/test_case/name.rs
franciszekjob: https://github.com/foundry-rs/starknet-foundry/pull/3697/commits/05541bd9df5f13d009891ecd32ceea94f9e0eb24

crates/forge/tests/e2e/plugin_diagnostics.rs
ddoktorski, line 1: Where are these tests run on ci?

ddoktorski (commented):
Does it work for this order of attributes?
```
#[test_case]
#[test]
#[fuzzer]
```

crates/forge/tests/data/diagnostics/test_case_attr/Scarb.toml
franciszekjob: Will bump this in separate PR with other packages.

crates/snforge-scarb-plugin/src/attributes/test.rs
```diff
@@ -73,28 +88,16 @@ fn test_internal(
...
     let attributes = SyntaxNodeWithDb::new(&attributes, db);
 
-    let test_func_ident = format_ident!("{}__test_generated", name);
+    let test_func = TokenStream::new(vec![format_ident!("{}__test_generated", name)]);
     let func_ident = format_ident!("{}", name);
 
-    let out_of_gas = create_single_token("'Out of gas'");
-
     if should_run_test {
+        let call_args = TokenStream::new(vec![format_ident!("")]);
```
ddoktorski:
```suggestion
        let call_args = TokenStream::empty();
```

ddoktorski (approved):
Please confirm with the LS team whether running tests from VS Code works as expected after this change
Screenshot 2025-09-22 at 09 32 35

franciszekjob:
> Please confirm with the LS team whether running tests from VS Code works as expected after this change Screenshot 2025-09-22 at 09 32 35

https://github.com/software-mansion/cairols/issues/1000

crates/snforge-scarb-plugin/src/attributes/test_case.rs
```diff
@@ -0,0 +1,139 @@
...
+    let has_fuzzer = has_fuzzer_attribute(func_db, func);
+
+    // We do not want to copy the `#[test]` attribute if there is no `#[fuzzer]`
+    attr_list
+        .elements(func_db)
+        .filter(|attr| {
+            let test_attr_text = format!("#[{}]", TestCollector::ATTR_NAME);
+            let attr_text = attr.as_syntax_node().get_text(func_db);
+            let attr_text = attr_text.trim();
+            let is_test_attr = attr_text == test_attr_text;
```
Draggu, line 130: will not work with `#[test()]`

crates/snforge-scarb-plugin/src/attributes/test_case/name.rs
```diff
@@ -0,0 +1,66 @@
...
+    func_name: &str,
+    arguments: &Arguments,
+    db: &SimpleParserDatabase,
+) -> Result<String, Diagnostics> {
+    let named_args = arguments.named();
+    let test_case_name = named_args.as_once_optional("name")?;
+
+    let suffix = if let Some(expr) = test_case_name {
+        match expr {
+            Expr::ShortString(_) | Expr::String(_) => {}
```
Draggu, line 53: We could use a regular identifier instead, which would be more readable

PR #3671, opened 2025-08-24
Refactor function names generated by plugin

franciszekjob:
Towards #1431

**Stack**:
- #3697
- #3669
- #3671 ⬅
- #3670

## Introduced changes

This change makes the name of the test function remain the same. Previously, we modified its name by adding `_return_wrapper` and `_actual_body` suffixes. However, while implementing `#[test_case]`, it turned out its tricky to determine the name of the function to be called. 

## Checklist

- [ ] Linked relevant issue
- [ ] Added relevant tests


crates/snforge-scarb-plugin/src/attributes/test.rs
```diff
@@ -68,11 +73,9 @@ fn test_internal(
     let attributes = func.attributes(db).as_syntax_node();
     let attributes = SyntaxNodeWithDb::new(&attributes, db);
 
-    let name_return_wrapper =
-        format_ident!("{}_return_wrapper", func.declaration(db).name(db).text(db));
-
-    let mut return_wrapper = TokenStream::new(vec![name_return_wrapper.clone()]);
-    return_wrapper.extend(signature);
+    let name = format_ident!("{}__test_generated", func.declaration(db).name(db).text(db));
```
cptartur: Can't we reuse it?

```suggestion
    let name = format_ident!("{}__test_generated", func_name);
```

crates/snforge-scarb-plugin/src/attributes/test.rs
```diff
@@ -55,8 +53,15 @@ fn test_internal(
...
-    let name = SyntaxNodeWithDb::new(&name, db);
+    let func_name = func.declaration(db).name(db).text(db);
+
+    let has_fuzzer = has_fuzzer_attribute(db, func);
+    let name = if has_fuzzer {
+        format!("{func_name}__fuzzer_generated")
+    } else {
+        func_name.to_string().clone()
+    };
+    let name_return_wrapper = format_ident!("{}", name);
```
cptartur: Wouldn't something along this work? 

```suggestion
    let name_return_wrapper = if has_fuzzer {
        format_ident!("{func_name}__fuzzer_generated")
    } else {
        format_ident!("{}", name)
    };
```

crates/snforge-scarb-plugin/src/common.rs
```diff
@@ -58,3 +61,24 @@ where
...
+fn has_attributes(db: &SimpleParserDatabase, func: &FunctionWithBody, attr_names: &[&str]) -> bool {
+    func.attributes(db).elements(db).any(|attr| {
+        attr_names.contains(
+            &attr
+                .attr(db)
+                .as_syntax_node()
+                .get_text_without_trivia(db)
+                .as_str(),
+        )
+    })
```
cptartur, line 78: This will evaluate to `true` if function attributes contain any single attribute from `attr_names`, is that intended?

If so let's rename it to `has_any_attribute` or similar, because the current name suggest that this is checking if `func` contains all of the attributes passed as `attr_names`.

crates/snforge-scarb-plugin/src/attributes/fuzzer/wrapper.rs
```diff
@@ -104,8 +106,7 @@ fn fuzzer_wrapper_internal(
         }
     });
 
-    let actual_body_fn_name =
-        format_ident!("{}_actual_body", func.declaration(db).name(db).text(db));
+    let actual_body_fn_name = format_ident!("{}", func.declaration(db).name(db).text(db));
```
cptartur: Let's rename the variables as well to match the changes to generated names

crates/snforge-scarb-plugin/src/attributes/test.rs
cptartur, line 1: Let's rename variables name here and in the quote macro itself so they actually match what code is generated. E.g. `name` was a plain function name before but it no longer is, making it hard to follow the generated code.

crates/snforge-scarb-plugin/tests/integration/multiple_attributes.rs
cptartur, line 1: I think we should also add a test case that follows the macro expansions if `#[fuzzer]` is used before `#[test]` since `#[test]` has special handling of `#[fuzzer]` and we should test that.

cptartur (changes requested):

crates/snforge-scarb-plugin/tests/integration/multiple_attributes.rs
franciszekjob: Added `works_with_fuzzer_before_test`

ksew1 (approved):

crates/snforge-scarb-plugin/tests/integration/multiple_attributes.rs
```diff
@@ -216,7 +216,55 @@ fn works_with_fuzzer() {
...
+                );
+                empty_fn__fuzzer_generated();
+
+                let mut arr = ArrayTrait::new();
+                core::array::ArrayTrait::span(@arr)
+            }
+
+            #[__fuzzer_config(runs: 123, seed: 321)]
+            #[__fuzzer_wrapper]
+            #[__internal_config_statement]
```
cptartur, line 265: Can we extend this tests so we later call `fuzzer_wrapper` on the `empty_func`? So we have complete flow of generation tested for this order.

franciszekjob: https://github.com/foundry-rs/starknet-foundry/pull/3671/commits/6bd35bdebac8e969b9643d8cf5a0f77bd598b303

cptartur (approved):

crates/snforge-scarb-plugin/src/attributes/fuzzer/wrapper.rs
```diff
@@ -104,8 +106,7 @@ fn fuzzer_wrapper_internal(
         }
     });
 
-    let actual_body_fn_name =
-        format_ident!("{}_actual_body", func.declaration(db).name(db).text(db));
+    let func_ident = format_ident!("{}", func.declaration(db).name(db).text(db));
```
ddoktorski: Given that we don't modify this name, it would be better to keep the original token stream so that we're not losing mappings
```suggestion
    let func_name = func.declaration(db).name(db).as_syntax_node();
    let func_name = SyntaxNodeWithDb::new(&func_name, db);
```

crates/snforge-scarb-plugin/src/attributes/test.rs
```diff
@@ -55,8 +53,14 @@ fn test_internal(
...
-    let name = func.declaration(db).name(db).as_syntax_node();
-    let name = SyntaxNodeWithDb::new(&name, db);
+    let func_name = func.declaration(db).name(db).text(db);
+
+    let has_fuzzer = has_fuzzer_attribute(db, func);
+    let called_func_name_ident = if has_fuzzer {
+        format_ident!("{func_name}__fuzzer_generated")
+    } else {
+        format_ident!("{}", name)
+    };
```
ddoktorski: Could you add a comment explaining this logic?

crates/snforge-scarb-plugin/src/attributes/test.rs
```diff
@@ -68,35 +72,33 @@ fn test_internal(
...
+            fn #test_func_name_ident(mut _data: Span<felt252>) -> Span::<felt252> {
                 core::internal::require_implicit::<System>();
                 core::internal::revoke_ap_tracking();
                 core::option::OptionTraitImpl::expect(core::gas::withdraw_gas(), #out_of_gas);
 
                 core::option::OptionTraitImpl::expect(
                     core::gas::withdraw_gas_all(core::gas::get_builtin_costs()), #out_of_gas
                 );
-                #name_return_wrapper();
+                #called_func_name_ident();
```
ddoktorski: Can we split this into 2 token streams, similarly as in the fuzzer wrapper?

crates/snforge-scarb-plugin/src/attributes/test.rs
```diff
@@ -55,8 +53,14 @@ fn test_internal(
         None => true,
     };
 
-    let name = func.declaration(db).name(db).as_syntax_node();
-    let name = SyntaxNodeWithDb::new(&name, db);
+    let func_name = func.declaration(db).name(db).text(db);
```
ddoktorski: Can we use the above `name` variable instead?

crates/snforge-scarb-plugin/src/attributes/test.rs
```diff
@@ -68,35 +72,33 @@ fn test_internal(
...
+                #called_func_name_ident();
 
                 let mut arr = ArrayTrait::new();
                 core::array::ArrayTrait::span(@arr)
             }
 
             #attributes
             #[#internal_config]
-            fn #return_wrapper
+            fn #func_ident
```
ddoktorski: Also here, to have something like `#func_ident #signature`

ddoktorski (approved):

Issue #3659, opened 2025-08-21
Remove `deploy` cheatcode


ddoktorski:
### Current State

Remove it as a cheatcode and make it a wrapper around `deploy_syscall` to avoid introducing a breaking change. `deploy_at` should have its own separate cheatcode that accepts a contract address provided by the user.

### Objective

Remove `deploy` cheatcode

ddoktorski:
Merged in #3699

Issue #3428, opened 2025-05-30
Add tests comparing gas to values from real network


cptartur:
### Current State

Currently we only compare gas against values from equation from fee docs. This can catch some errors but not all.

### Objective

Add tests comparing gas to values from real network using a real node

### Additional Context

Devnet should not be used for that because there is a risk of it having incorrect values as well. The tests should run some tests in snforge, then we should run similar operations on a real node (make requests to Starknet) and compare if the values are withing some threshold.

PR #3674, opened 2025-08-25
Remove `deploy` as cheatcode

ddoktorski:
Towards #3659

commit-id:012afbae

---

**Stack**:
- #3700
- #3699
- #3695
- #3676
- #3675
- #3674 ⬅

### DIFF
--- a/crates/forge/tests/integration/gas.rs
+++ b/crates/forge/tests/integration/gas.rs
@@ -724,7 +724,7 @@ fn multiple_storage_writes_cost_cairo_steps() {
         GasVector {
             l1_gas: GasAmount(0),
             l1_data_gas: GasAmount(192),
-            l2_gas: GasAmount(400_000),
+            l2_gas: GasAmount(360_000),
         },
     );
 }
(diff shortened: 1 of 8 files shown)


crates/forge/tests/integration/gas.rs
ddoktorski, line 1: Note:
1. The values in these tests had to be adjusted because we no longer need to serialize the input for the `execute_cheatcode_and_deserialize` function and deserialization now only involves a single felt. The updated values reflect the reduced cost from removing input serialization and simplifying deserialization, as confirmed by the profiler.
2. The comments containing cost calculations have not been updated for a while. Updating them is time consuming and unmaintainable long term as it requires changes with every new version of Cairo and Blockifier. We should rethink these tests in the scope of #3428

franciszekjob, line 1: Maybe let's add a todo on top of this file, linking the issue. It may take some time once we fully address it so we should somehow mark that they aren't consistent with calculations to prevent potential confusion.

franciszekjob (approved):

crates/forge/tests/integration/gas.rs
cptartur: Re 2. I know that maintaining these is cumbersome, but until we have #3428, I'd prefer the comments are maintained.

cptartur (commented):
Replied to a comment

crates/forge/tests/integration/gas.rs
ddoktorski: Done in #3699

cptartur (approved):

PR #3675, opened 2025-08-25
Remove `deploy` cheatnet tests

ddoktorski:
Towards #3659

commit-id:a6d72598

---

**Stack**:
- #3700
- #3699
- #3695
- #3676
- #3675 ⬅

### DIFF
--- a/crates/cheatnet/tests/common/mod.rs
+++ b/crates/cheatnet/tests/common/mod.rs
@@ -2,13 +2,15 @@ use assertions::ClassHashAssert;
 use blockifier::execution::call_info::CallInfo;
 use blockifier::execution::contract_class::TrackedResource;
 use blockifier::execution::entry_point::{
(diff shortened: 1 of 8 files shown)


franciszekjob (approved):

crates/cheatnet/tests/common/mod.rs
```diff
@@ -189,6 +192,41 @@ pub fn deploy_at_wrapper(
     Ok(contract_address)
 }
 
+fn deploy_helper(
```
cptartur, line 195: Why do we need this helper, doesn't this mean we are effectively testing the logic of the helper instead?

ddoktorski: At this point it is kind of in between, but in #3695 it is used only as a setup for cheatnet tests

cptartur: Let's make sure it's only used for setup then.

cptartur (approved):

PR #3698, opened 2025-08-28
Bundle LLVM

cptartur:
Closes #

## Introduced changes

-

## Checklist

- [ ] Linked relevant issue
- [ ] Added relevant tests

### DIFF
--- a/NATIVE_INSTALLATION.md
+++ b/NATIVE_INSTALLATION.md
@@ -1,0 +1,21 @@
+# Installing Starknet Foundry With Cairo Native Support
+
+Cairo Native introduces additional dependencies outside of the Rust ecosystem.
+
+## LLVM
+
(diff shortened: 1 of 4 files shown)


ksew1 (commented):
This works?

cptartur:
> This works?

@ksew1  It does, for the targets that weren't commented out in build workflow.

NATIVE_INSTALLATION.md
```diff
@@ -0,0 +1,21 @@
+# Installing Starknet Foundry With Cairo Native Support
```
ksew1, line 1: We should probably link this somewhere in documentation. Also maybe we could create short doc page for naitive

cptartur: After have a proper release, we can do that. But initially it will just be available from a branch anway.

PR #3703, opened 2025-09-01
Ignore `meta_tx_v0` integration tests for scarb < `2.11.0`

MKowalski8:
Closes #

## Introduced changes

- Fix scheduled tests

## Checklist

- [ ] Linked relevant issue
- [ ] Added relevant tests

### DIFF
--- a/crates/forge/tests/integration/meta_tx_v0.rs
+++ b/crates/forge/tests/integration/meta_tx_v0.rs
@@ -10,16 +10,10 @@ use test_utils::test_case;
 // TODO(#3704) Remove scarb version check
 fn skip_if_scarb_lt_2_11_0() -> Option<String> {
     match ScarbCommand::version().run() {
-        Ok(version_info) => {
-            if version_info.scarb < Version::new(2, 11, 0) {
-                Some(
-                    "[IGNORED] `meta_tx_v0` syscall is not supported in Scarb < 2.11.0".to_string(),
-                )
-            } else {
-                None
-            }
+        Ok(version_info) if version_info.scarb < Version::new(2, 11, 0) => {
+            Some("[IGNORED] `meta_tx_v0` syscall is not supported in Scarb < 2.11.0".to_string())
         }
-        Err(_) => None,
+        _ => None,
     }
 }
 
--- a/crates/forge/tests/integration/meta_tx_v0.rs
+++ b/crates/forge/tests/integration/meta_tx_v0.rs
@@ -7,11 +7,14 @@ use test_utils::runner::{Contract, assert_passed};
 use test_utils::running_tests::run_test_case;
 use test_utils::test_case;
 
-fn should_skip_for_scarb_2_10_1() -> Option<String> {
+// TODO(#3704) Remove scarb version check
+fn skip_if_scarb_lt_2_11_0() -> Option<String> {
     match ScarbCommand::version().run() {
         Ok(version_info) => {
-            if version_info.scarb == Version::new(2, 10, 1) {
-                Some("[IGNORED] `meta_tx_v0` syscall is not supported in Scarb 2.10.1".to_string())
+            if version_info.scarb < Version::new(2, 11, 0) {
+                Some(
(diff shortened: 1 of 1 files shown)


crates/forge/tests/integration/meta_tx_v0.rs
```diff
@@ -1,12 +1,32 @@
...
+    }
+}
+
 #[test]
 fn meta_tx_v0_with_cheat_caller_address() {
+    if let Some(skip_message) = should_skip_for_scarb_2_10_1() {
+        eprintln!("{skip_message}");
+        return;
+    }
+
```
MKowalski8: note:  I didn’t use `#[ignore]` because all tests run with the same command, so I can’t add a feature flag for that.

crates/forge/tests/integration/meta_tx_v0.rs
```diff
@@ -1,12 +1,32 @@
...
+        Ok(version_info) => {
+            if version_info.scarb == Version::new(2, 10, 1) {
+                Some("[IGNORED] `meta_tx_v0` syscall is not supported in Scarb 2.10.1".to_string())
+            } else {
+                None
+            }
+        }
+        Err(_) => None,
+    }
+}
```
cptartur: Maybe let's use `<=` for version instead asserting for a specific one?

MKowalski8: Okay, I can do it that way

crates/forge/tests/integration/meta_tx_v0.rs
```diff
@@ -1,12 +1,36 @@
...
 use indoc::indoc;
+use scarb_api::ScarbCommand;
+use semver::Version;
 use std::path::Path;
 use test_utils::runner::{Contract, assert_passed};
 use test_utils::running_tests::run_test_case;
 use test_utils::test_case;
 
+// TODO(#3704) Remove scarb version check
+fn skip_if_scarb_lt_2_11_0() -> Option<String> {
```
ksew1: Maybe this:

```rs
fn skip_if_scarb_lt_2_11_0() -> Option<String> {
    match ScarbCommand::version().run() {
        Ok(version_info) if version_info.scarb < Version::new(2, 11, 0) => {
            Some("[IGNORED] `meta_tx_v0` syscall is not supported in Scarb < 2.11.0".to_string())
        }
        _ => None,
    }
}
```

ksew1 (approved):

crates/forge/tests/integration/meta_tx_v0.rs
```diff
@@ -1,12 +1,30 @@
...
+    match ScarbCommand::version().run() {
+        Ok(version_info) if version_info.scarb < Version::new(2, 11, 0) => {
+            Some("[IGNORED] `meta_tx_v0` syscall is not supported in Scarb < 2.11.0".to_string())
+        }
+        _ => None,
```
cptartur: I'd just panic in case getting scarb version failed i.e. unwrap the result. If we are unable to get version from scarb we are in some kind of incorrect tests setup/state anyway.

crates/forge/tests/integration/meta_tx_v0.rs
```diff
@@ -1,12 +1,30 @@
...
+    }
+}
+
 #[test]
 fn meta_tx_v0_with_cheat_caller_address() {
+    // TODO(#3704) Remove scarb version check
+    if let Some(skip_message) = skip_if_scarb_lt_2_11_0() {
+        eprintln!("{skip_message}");
+        return;
+    }
```
cptartur: Maybe let's keep the message printing in `skip_if_scarb_lt_2_11_0`, this way you could just do this


```suggestion
    skip_if_scarb_lt_2_11_0();
```

MKowalski8: I can move printing, but I still need to do the `return` part

cptartur: Ah right, disregard that then

cptartur (approved):

PR #3709, opened 2025-09-02
Refactor native compilation

ksew1:
**Stack**:
- #3711
- #3710
- #3709 ⬅
- #3713

### DIFF
--- a/crates/native-api/src/lib.rs
+++ b/crates/native-api/src/lib.rs
@@ -1,0 +1,40 @@
+//! This module provides functionality related to `cairo-native`.
+//!
+//! Currently, it includes:
+//!  - Compiling Sierra contract classes into `cairo-native` executors.
+
+use cairo_lang_starknet_classes::contract_class::ContractClass;
+use cairo_native::executor::AotContractExecutor;
+use starknet_api::contract_class::SierraVersion;
+
+/// Compiles a given Sierra [`ContractClass`] into an [`AotContractExecutor`] for `cairo-native` execution.
+#[must_use]
+pub fn compile_contract_class(contract_class: &ContractClass) -> AotContractExecutor {
+    let sierra_program = contract_class
+        .extract_sierra_program()
+        .expect("extraction should succeed");
+
+    let sierra_version = extract_sierra_version(contract_class);
+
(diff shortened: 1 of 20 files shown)


crates/native-api/src/lib.rs
```diff
@@ -0,0 +1,40 @@
...
+    let sierra_program = contract_class
+        .extract_sierra_program()
+        .expect("extraction should succeed");
+
+    let sierra_version = extract_sierra_version(contract_class);
+
+    AotContractExecutor::new(
+        &sierra_program,
+        &contract_class.entry_points_by_type,
+        sierra_version.clone().into(),
```
cptartur, line 22: I don't think we officially support any scarb version that produces incompatible sierra, but as a sanity check, maybe we should validate the version somewhere.

I think native only supports sierra from 1.7.0 upwards.

Maybe minimal scarb version in foundry should be bumped to lowest scarb that supports that version? (Still there's a problem of contracts from network in forking, we should ensure we don't compile incompatible ones).

crates/scarb-api/src/artifacts.rs
```diff
@@ -12,34 +13,53 @@ mod deserialized;
...
+#[derive(Debug, Clone)]
 pub struct StarknetContractArtifacts {
     /// Compiled sierra code
     pub sierra: String,
     /// Compiled casm code
     pub casm: String,
+
+    /// Optional AOT compiled native executor
+    pub executor: Option<AotContractExecutor>,
+}
```
cptartur, line 25: To be 100% idiomatic, I think this struct should be moved to forge crate or somewhere else so that that impl is owned by the user. It doesn't make much sense for Scarb api to own a struct for `StarknetContractArtifacts` with details about native compilation if scarb knows nothing about that.

cptartur: And while we are at it, having casm here if compilation is done by USC not scarb is also weird.

cptartur: Unless refactoring this is an extremely big lift, I think now might be a good moment to finally untangle that.

cptartur (approved):
Probably disregard bigger refactoring for now

PR #3715, opened 2025-09-03
Implement fingerprint callback for snforge-scarb-plugin

maciektr:


ksew1 (approved):

crates/snforge-scarb-plugin/src/attributes/test.rs
```diff
@@ -109,6 +110,20 @@ fn get_forge_test_filter() -> Result<String, VarError> {
...
+/// whether Cairo code depending on this macro should be recompiled.
+/// The callback is concerned with informing Scarb about changes to inputs that don't come from Scarb directly,
+/// like the `SNFORGE_TEST_FILTER` environmental variable.
+///
+/// Warning: Removing this callback can break incremental compilation with this macro!
+#[fingerprint]
+fn test_filter_fingerprint() -> u64 {
+    get_forge_test_filter()
+        .ok()
+        .map(|filter| xxh3_64(filter.as_bytes()))
```
ksew1: Maybe we could add comment that hasher needs to be deterministic across runs so because of the we can't use rust `DefaultHasher`

maciektr: Sure, whatever helps!

cptartur (commented):
Is the fingerprint already used by Scarb 2.12.0 or is it ignored and will be handled in future versions?

maciektr:
@cptartur no, it's not released yet.

cptartur:
> @cptartur no, it's not released yet.

I assume current scarbs will just ignore this?

cptartur:
> @cptartur no, it's not released yet.

@maciektr so current scarb versions just ignore the fingerprint?

maciektr:
@cptartur `0.2.1` of `cairo-lang-macro` exposes a new symbol in built macro, that Scarb will use in future to run `fingerprint` callbacks that macro may define. Currently, fingerprints calculation for proc macros is defined internally in Scarb.

franciszekjob (approved):

crates/snforge-scarb-plugin/src/attributes/test.rs
```diff
@@ -109,6 +110,22 @@ fn get_forge_test_filter() -> Result<String, VarError> {
...
 
+/// This function implements a callback, that will be used by Scarb to determine,
+/// whether Cairo code depending on this macro should be recompiled.
+/// The callback is concerned with informing Scarb about changes to inputs that don't come from Scarb directly,
+/// like the `SNFORGE_TEST_FILTER` environmental variable.
+///
+/// Warning: Removing this callback can break incremental compilation with this macro!
+#[fingerprint]
+fn test_filter_fingerprint() -> u64 {
+    get_forge_test_filter()
```
Draggu: Because it is crucial to have this code correct, maybe do it more like
```rs
fn get_external_input() -> Externalnput {
 ....
}

#[derive(Hash)] // this way we will NOT forget to add new fields to hash value
struct Externalnput {
    forge_test_filter: String
}
```
So it will be clearly visible when adding new external input source that we MUST have it hashed as well.
WDYT?

We should also ban modules like `std::fs`,`std::env` etc. in clippy and explicitly allow these in `get_external_input()` function only.

cptartur (commented):
Please re-request after addressing @Draggu comment

maciektr:
@cptartur done

Draggu (approved):

crates/snforge-scarb-plugin/src/lib.rs
```diff
@@ -1,3 +1,5 @@
+#![deny(clippy::disallowed_methods)]
```
cptartur, line 3: Maybe let's add a comment about why was it added, so someone doesn't remove it in the future.

cptartur (approved):

Issue #3587, opened 2025-07-28
Allow declaring from precompiled contracts in `sncast`


cptartur:
### Current State

Currently you can only declare contracts you have source code available. There are use cases where an already compiled contract needs to be declared e.g. to avoid issues with different compiler versions producing different results.

### Objective

Allow declaring from precompiled contracts in `sncast`

PR #3727, opened 2025-09-15
Implement `declare-from`

franciszekjob:
Towards #3587

**Stack**:
- #3729
- #3727 ⬅
- #3726

## Introduced changes

Add `sncast declare-from` command

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/sncast/src/starknet_commands/declare_from.rs
```diff
@@ -0,0 +1,133 @@
...
+
+    #[command(flatten)]
+    pub source_rpc: SourceRpcArgs,
+
+    #[command(flatten)]
+    pub rpc: RpcArgs,
+
+    /// Block identifier on which call should be performed.
+    /// Possible values: `pre_confirmed`, `latest`, block hash (0x prefixed string)
+    /// and block number (u64)
```
ddoktorski: This description doesn't seem to be accurate. It's the block from which the contract will be pulled, right?

crates/conversions/src/sierra_class.rs
```diff
@@ -0,0 +1,26 @@
+use crate::TryFromConv;
+use anyhow::Result;
+use starknet::core::types::{
+    FlattenedSierraClass,
+    contract::{AbiEntry, SierraClass, SierraClassDebugInfo},
+};
+
+use std::vec;
+
+impl TryFromConv<FlattenedSierraClass> for SierraClass {
```
ddoktorski: I'd consider retrieving a raw Sierra from a network without using starknet-rs structs, so we avoid any manipulations with the returned code, wdyt?

crates/sncast/src/starknet_commands/declare_from.rs
franciszekjob: Absolutely, missed it while copying description.

crates/conversions/src/sierra_class.rs
franciszekjob: > I'd consider retrieving a raw Sierra from a network without using starknet-rs 

All requests are done through starknet-rs, so how actually would you do that?

> so we avoid any manipulations with the returned code, wdyt?

We need to make this conversion because `declare_with_artifacts` needs `SierraClass` instance, but before we receive  `FlattenedSierraClass`. The only manipulation is that we add `sierra_program_debug_info` with empty fields (i.e. empty arrays), not sure it it's a serious modification. Maybe it would be better to have it in a helper function instead a conversion, wdyt?

crates/sncast/src/starknet_commands/declare_from.rs
```diff
@@ -0,0 +1,133 @@
...
+        Ok(provider)
+    }
+
+    #[must_use]
+    fn get_url(&self) -> Option<String> {
+        if let Some(network) = self.source_network {
+            let free_provider = FreeProvider::semi_random();
+            Some(network.url(&free_provider))
+        } else {
+            self.source_url
```
cptartur, line 79: Should we add an assert that it is not empty here as well? We have one above but not sure if it is necessary at all.

crates/sncast/src/starknet_commands/declare_from.rs
```diff
@@ -0,0 +1,133 @@
...
+            return Err(StarknetCommandError::UnknownError(anyhow::anyhow!(
+                "Declaring from Cairo 0 (legacy) contracts is not supported"
+            )));
+        }
+    };
+    let sierra: SierraClass = flattened_sierra
+        .try_into_()
+        .expect("Failed to parse flattened sierra class");
+
+    let casm_json: String = compile_sierra(
```
cptartur, line 112: Technically there's an endpoint to fetch casm from network, but not sure how reliable it is. Most likely it makes sense to use `compile_sierra` still but you can take a look at it.

crates/sncast/tests/e2e/declare_from.rs
```diff
@@ -0,0 +1,219 @@
...
+        .env("SNCAST_FORCE_SHOW_EXPLORER_LINKS", "1")
+        .current_dir(temp_dir.path());
+    let output = snapbox.assert().success();
+
+    assert_stdout_contains(
+        output,
+        indoc! {r"
+        Success: Declaration completed
+
+        Class Hash:       0x[..]
```
cptartur: We should assert the class hash here. The intended behavior is that the declared-from contract has the same class hash as the one of the source address.

crates/sncast/src/starknet_commands/declare_from.rs
franciszekjob: > Technically there's an endpoint to fetch casm from network

Yes, but starknet-rs doesn't support it.

Even if it was available, I think compiling is better approach, since it guarantees the CASM is valid and ensures a single source of truth.

crates/sncast/tests/e2e/declare_from.rs
franciszekjob: https://github.com/foundry-rs/starknet-foundry/pull/3727/commits/b26c0d7306116a66ac4adaa00899cd4d70105fc3

crates/conversions/src/sierra_class.rs
cptartur: Maybe let's make it a function so someone doesn't try to reuse it in the future as a proper conversion. But I think logic wise it's fine - we don't want to have any debug info on network anyway, if that's even possible.

cptartur (approved):

ddoktorski (approved):

PR #3728, opened 2025-09-15
Fix duplicate packages

cptartur:
Closes #

## Introduced changes

- Renamed packages with duplicate names (otherwise scarb would produced warnings if using snforge_std from git dependency)
- Added test verifying using snforge_std from git doesn't produce any warnings

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/forge/tests/data/hello_workspaces2/Scarb.toml
```diff
@@ -16,7 +16,7 @@ snforge_std = { path = "../../../../../snforge_std" }
 version = "0.1.0"
 
 [package]
-name = "hello_workspaces"
+name = "hello_workspaces2"
```
ddoktorski: nit: Maybe instead of appending a number to the name, could we add a crate name suffix instead, given that there can't be duplicates within the scope of a single package? For example `hello_workspaces_docs`. This would make it easier to immediately see where the package is used.

ddoktorski (approved):

crates/forge/tests/data/hello_workspaces2/Scarb.toml
cptartur: Renamed

franciszekjob (approved):

ddoktorski (approved):

PR #3736, opened 2025-09-18
Get gas resources from `CallInfo`

### LINKED ISSUE #2960: Verify correctness of asserted gas value in `l1_handler_cost` test

### LINKED ISSUE #3292: Get gas and vm resources directly from `CallInfo`

ddoktorski:
Closes #3292
Closes #2960

commit-id:7967ca03

---

**Stack**:
- #3756
- #3755
- #3752
- #3751
- #3736 ⬅

### DIFF
--- a/crates/forge/tests/integration/gas.rs
+++ b/crates/forge/tests/integration/gas.rs
@@ -13,6 +13,9 @@ use test_utils::test_case;
 // 1 cairo step = 0.0025 L1 gas = 100 L2 gas
 // 1 sierra gas = 1 l2 gas
 // Costs of syscalls (if provided) are taken from versioned_constants (blockifier)
+// In Sierra gas tests, only the most significant costs are considered
+// Asserted values should be slightly greater than the computed values,
+// but must remain within a reasonable range
 
 #[test]
 fn declare_cost_is_omitted_cairo_steps() {
@@ -894,19 +897,17 @@ fn l1_handler_cost_cairo_steps() {
 
     let result = run_test_case(&test, ForgeTrackedResource::CairoSteps);
     assert_passed(&result);
-    // TODO(#2960): These calculations are based on code from blockifier 0.14.0-rc.1
-    // ATM, they are a bit different from the formula in the docs
-    // We should verify correctness of this (docs may be outdated)
-
     // 96 = gas cost of onchain data (deploy cost)
     // int(5.12 * 4) = 21 = keccak cost from l1 handler
     // in this test, l1_handler_payload_size = 6
     // 15923 = 12251 (gas used for processing L1<>L2 messages on L1) + 3672 (SHARP gas, 6 * 612)
+    // https://github.com/starkware-libs/sequencer/blob/028db0341378147037b5e7236d8e136e4ca7c30d/crates/blockifier/src/fee/resources.rs#L338-L340
     // 12251 = 3072 (6 * 512, 512 is gas per memory word) +
-    //         + 4179 (result of get_consumed_message_to_l2_emissions_cost(6) which is get_event_emission_cost(3, 3 + 6) = 375 + (3 + 1) * 375 + 9 * 256) +
-    //         + 0 +
-    //         + 5000 (1 * 5000, 5000 is gas per counter decrease, ref: https://github.com/starkware-libs/sequencer/blob/main/crates/blockifier/src/fee/resources.rs#L364-L368)
-    //
+    //         + 4179 (cost of `ConsumedMessageToL2` event for payload with length 6:
+    //              -> 375 const opcode cost
+    //              -> 4 * 375 topics cost (fromAddress, toAddress, selector and 1 default)
+    //              -> 9 * 256 data array cost (payload length, nonce and 2 required solidity params for array)
+    //         + 5000 (gas per counter decrease, fixed cost of L1 -> L2 message)
     //
     // 15923 l1_gas + 96 l1_data_gas + 21 * (100 / 0.0025) l2 gas
     assert_gas(
@@ -1153,6 +1154,29 @@ fn nested_call_cost_in_forked_contract_cairo_steps() {
 }
 
 #[test]
+fn empty_test_sierra_gas() {
+    let test = test_case!(indoc!(
+        r"
+            #[test]
+            fn empty_test() {}
+        "
+    ));
+
+    let result = run_test_case(&test, ForgeTrackedResource::SierraGas);
+    assert_passed(&result);
+
+    assert_gas(
+        &result,
+        "empty_test",
+        GasVector {
+            l1_gas: GasAmount(0),
+            l1_data_gas: GasAmount(0),
+            l2_gas: GasAmount(13_840),
+        },
+    );
+}
+
+#[test]
 fn declare_cost_is_omitted_sierra_gas() {
     let test = test_case!(
         indoc!(
@@ -1174,15 +1198,14 @@ fn declare_cost_is_omitted_sierra_gas() {
     let result = run_test_case(&test, ForgeTrackedResource::SierraGas);
 
     assert_passed(&result);
-    // 17_980 = reported consumed sierra gas
-    // 0 l1_gas + 0 l1_data_gas + 17_980 l2 gas = 17_980 total gas
+
     assert_gas(
         &result,
         "declare_cost_is_omitted",
         GasVector {
             l1_gas: GasAmount(0),
             l1_data_gas: GasAmount(0),
-            l2_gas: GasAmount(17980),
+            l2_gas: GasAmount(25_850),
         },
     );
 }
@@ -1197,7 +1220,7 @@ fn deploy_syscall_cost_sierra_gas() {
             #[test]
             fn deploy_syscall_cost() {
                 let contract = declare("GasConstructorChecker").unwrap().contract_class().clone();
-                let (address, _) = deploy_syscall(contract.class_hash, 0, array![1].span(), false).unwrap();
+                let (address, _) = deploy_syscall(contract.class_hash, 0, array![0].span(), false).unwrap();
                 assert(address != 0.try_into().unwrap(), 'wrong deployed addr');
             }
         "#
@@ -1216,21 +1239,20 @@ fn deploy_syscall_cost_sierra_gas() {
     // n = 1 (unique contracts updated - in this case it's the new contract address)
     // ( l + n * 2 ) * felt_size_in_bytes(32) = 96 (total l1 data cost)
     //
-    // 20_000 = cost of 2 keccak syscall (because 2 * 100 * 100) (from constructor)
-    //      -> 1 keccak syscall costs 100 cairo steps
     // 151_970 = cost of 1 deploy syscall (because 1 * (1173 + 8) * 100 + (7 + 1) * 4050 + 21 * 70)
     //      -> 1 deploy syscall costs 1132 cairo steps, 7 pedersen and 18 range check builtins
     //      -> 1 calldata element costs 8 cairo steps and 1 pedersen
     //      -> 1 pedersen costs 4050, 1 range check costs 70
-    // 430_424 = reported consumed sierra gas
-    // 0 l1_gas + 96 l1_data_gas + (20_000 + 151_970 + 430_424) l2 gas
+    //
+    // 96 l1_data_gas
+    // l2 gas > 151_970
     assert_gas(
         &result,
         "deploy_syscall_cost",
         GasVector {
             l1_gas: GasAmount(0),
             l1_data_gas: GasAmount(96),
-            l2_gas: GasAmount(602_394),
+            l2_gas: GasAmount(186_210),
         },
     );
 }
@@ -1244,7 +1266,7 @@ fn snforge_std_deploy_cost_sierra_gas() {
             #[test]
             fn deploy_cost() {
                 let contract = declare("GasConstructorChecker").unwrap().contract_class();
-                let (address, _) = contract.deploy(@array![1]).unwrap();
+                let (address, _) = contract.deploy(@array![0]).unwrap();
                 assert(address != 0.try_into().unwrap(), 'wrong deployed addr');
             }
         "#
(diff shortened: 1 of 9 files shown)


crates/forge/tests/integration/gas.rs
```diff
@@ -894,19 +897,16 @@ fn l1_handler_cost_cairo_steps() {
...
     // 12251 = 3072 (6 * 512, 512 is gas per memory word) +
-    //         + 4179 (result of get_consumed_message_to_l2_emissions_cost(6) which is get_event_emission_cost(3, 3 + 6) = 375 + (3 + 1) * 375 + 9 * 256) +
-    //         + 0 +
-    //         + 5000 (1 * 5000, 5000 is gas per counter decrease, ref: https://github.com/starkware-libs/sequencer/blob/main/crates/blockifier/src/fee/resources.rs#L364-L368)
-    //
+    //         + 4179 (cost of `ConsumedMessageToL2` event for payload with length 6:
+    //              -> 375 const opcode cost
+    //              -> 4 * 375 topics cost (fromAddress, toAddress, selector and 1 default)
+    //              -> 9 * 256 data array cost (payload length, nonce and 2 required solidity params for array)
+    //         + 5000 (gas per counter decrease, fixed cost of L1 -> L2 message)
```
cptartur, line 910: Where do we got these calculations from? Seems similar to these docs but not 100% exact https://docs.starknet.io/learn/protocol/fees#l2%E2%86%92l1-messages

crates/forge/tests/integration/gas.rs
```diff
@@ -1292,21 +1312,24 @@ fn keccak_cost_sierra_gas() {
     assert_passed(&result);
     // 10_000 = cost of 1 keccak syscall (1 * 100 * 100)
     //      -> 1 keccak syscall costs 100 cairo steps
-    // 37_290 = reported consumed sierra gas
-    // 0 l1_gas + 0 l1_data_gas + (10_000 + 37_290) l2 gas
+    // 171_707 = cost of 1 keccak round syscall (136_189 + 3498 + 3980 + 28_100)
+    //      -> 1 keccak builtin costs 136_189
+    //      -> 6 bitwise builtin cost 6 * 583 = 3498
+    //      -> 56 range check builtins cost 56 * 70 = 3980
```
cptartur, line 1321: `keccak_u256s_le_inputs` consists of these builtins?

crates/forge/tests/integration/gas.rs
```diff
@@ -1338,65 +1361,20 @@ fn contract_keccak_cost_sierra_gas() {
     assert_passed(&result);
     // 96 = gas cost of onchain data (see `deploy_syscall_cost_sierra_gas` test)
     // 147_120 = cost of 1 deploy syscall (see `deploy_syscall_cost_sierra_gas` test)
-    // 50_000 = cost of 5 keccak syscall (5 * 100 * 100)
-    //      -> 1 keccak syscall costs 100 cairo steps
+    // 908_535 = 5 * 181_707 = cost of 5 keccak syscall (see `keccak_cost_sierra_gas` test)
```
cptartur, line 1367: Should it be that or 5 * the total cost of `keccak_cost_sierra_gas`

crates/forge/tests/integration/gas.rs
```diff
@@ -1338,65 +1361,20 @@ fn contract_keccak_cost_sierra_gas() {
...
         GasVector {
             l1_gas: GasAmount(0),
             l1_data_gas: GasAmount(96),
-            l2_gas: GasAmount(1_342_425),
-        },
-    );
-}
-
-#[test]
-fn contract_range_check_cost_sierra_gas() {
```
cptartur, line 1360: Why was this removed?

crates/forge/tests/integration/gas.rs
```diff
@@ -1539,23 +1514,29 @@ fn l1_message_cost_sierra_gas() {
...
 
     assert_passed(&result);
-    // todo(2960): verify l2 -> l1 message cost
-    // 29_524 = gas cost of l2 -> l1 message
+    // 29_524 = gas cost of l2 -> l1 message (payload length 3)
+    //      -> 6 * 1124 =  state update costs
+    //      -> 375 const opcode cost
+    //      -> 3 * 375 = 1225 topics cost of `LogMessageToL1` event (fromAddress, toAddress and 1 default)
+    //      -> 5 * 256 = 1280 data array cost (payload length + 2 required solidity params for array)
+    //      -> 20_000 l1 storage write cost
```
cptartur, line 1527: Same as above, this is mostly fine but seem to doesn't 100% match the equation from here https://docs.starknet.io/learn/protocol/fees#l2%E2%86%92l1-messages

crates/forge/tests/integration/gas.rs
```diff
@@ -1244,7 +1265,7 @@ fn snforge_std_deploy_cost_sierra_gas() {
             #[test]
             fn deploy_cost() {
                 let contract = declare("GasConstructorChecker").unwrap().contract_class();
-                let (address, _) = contract.deploy(@array![1]).unwrap();
+                let (address, _) = contract.deploy(@array![0]).unwrap();
```
franciszekjob, line 1269: Why is this changed?

crates/forge/tests/integration/gas.rs
franciszekjob: I remember from our discussion that the asserted value differs here in just 60, but we couldn't find the reason of that, correct @ddoktorski ?

crates/forge/tests/integration/gas.rs
ddoktorski: The documentation is not exhaustive on this topic, it only covers a basic L2->L1 case, but it does align with what we have here. It's helpful to take a look at the [core contract structure events](https://github.com/starkware-libs/cairo-lang/blob/66355d7d99f1962ff9ccba8d0dbacbce3bd79bf8/src/starkware/starknet/solidity/IStarknetMessagingEvents.sol#L4) and it also requires additional knowledge, such as how arrays are serialized on Ethereum (e.g. requiring two extra elements in the data array). Additionally, reviewing the blockifier code can be helpful, so overall I'd say the calculation is a mix of information from multiple resources, the documentation alone is not sufficient.

crates/forge/tests/integration/gas.rs
ddoktorski: Responded in the other comment

crates/forge/tests/integration/gas.rs
ddoktorski: Keccak syscall in fact uses KeccakRound resources and it does consist of these builtins:
* https://github.com/starkware-libs/sequencer/blob/v0.15.0-rc.3/crates/blockifier/resources/blockifier_versioned_constants_0_14_0.json#L267
* https://github.com/starkware-libs/sequencer/blob/v0.15.0-rc.3/crates/blockifier/src/execution/syscalls/syscall_executor.rs#L190

crates/forge/tests/integration/gas.rs
ddoktorski: That could work, but I’d prefer to stick to what we can calculate precisely. The full `keccak_cost_sierra_gas` also includes a setup cost, which would be counted multiple times if we use it

crates/forge/tests/integration/gas.rs
ddoktorski: This test has been added specifically for cairo steps calculation. It executes range check builtin mutliple times, so with cairo steps it is easy to calculate used gas as the formula uses the maximum from all builtins. However, with sierra gas the cost of the range check builtin represents only a part of the overall execution cost, so this test doesn't provide much relevant insight and it doesn't make much sense to keep it

crates/forge/tests/integration/gas.rs
```diff
@@ -1539,23 +1514,29 @@ fn l1_message_cost_sierra_gas() {
     let result = run_test_case(&test, ForgeTrackedResource::SierraGas);
 
     assert_passed(&result);
-    // todo(2960): verify l2 -> l1 message cost
-    // 29_524 = gas cost of l2 -> l1 message
+    // 29_524 = gas cost of l2 -> l1 message (payload length 3)
+    //      -> 6 * 1124 =  state update costs
```
franciszekjob: very nit (so we know it's base 3 + our payload length):
```suggestion
    //      -> (3 + 3) * 1124 =  state update costs
```

crates/forge/tests/integration/gas.rs
```diff
@@ -1539,23 +1514,29 @@ fn l1_message_cost_sierra_gas() {
...
+    //      -> 375 const opcode cost
+    //      -> 3 * 375 = 1225 topics cost of `LogMessageToL1` event (fromAddress, toAddress and 1 default)
+    //      -> 5 * 256 = 1280 data array cost (payload length + 2 required solidity params for array)
```
franciszekjob, line 1526: IIUC, these 2 points reflect below points

Image

Could our calculations be broken down into more detailed way so they explicitly match the formula?

crates/forge/tests/integration/gas.rs
cptartur: Okay, makes sense, but let's have a brief mention of that in the comment, so someone in the future doesn't have to figure it out again.

crates/forge/tests/integration/gas.rs
cptartur: Okay, we can remove it then

ddoktorski: > I remember from our discussion that the asserted value differs here in just 60, but we couldn't find the reason of that, correct @ddoktorski ?

No, here the point is that you cannot really calculate gas usage precisely (unless you look at sierra code and compute it based on libfunc used)

crates/forge/tests/integration/gas.rs
cptartur: I think it would be good to have it broken up more

crates/forge/tests/integration/gas.rs
ddoktorski: I don't want to modify cairo steps test, so I've added an if to still make additional computation in cairo steps, but not for sierra gas. At the same time deploy cost remains the same

crates/forge/tests/integration/gas.rs
cptartur: Two different things:
1. If you did your own investigation on how this is calculated, then let's include these details (with some links?) in the comments. This tests are hard enough to maintain, it'd be unwise to lose all this knowledge eventually.
2. Maybe you can follow-up with Starkware about their docs so that they can improve them?

cptartur (commented):
Responded to comments, please re-request

crates/forge/tests/integration/gas.rs
ddoktorski: It does match the formula, the example is for 2 topics and 2 words data array, in our case it is 3 topics and 5 word data array

crates/forge/tests/integration/gas.rs
ddoktorski: 1. I mainly looked into this topic because there was an issue linked for investigation, otherwise I wouldn't have analyzed it in such depth. In my opinion, the current comment already explains well enough how the calculation was done
2. I mentioned that the documentation alone isn't sufficient, but to be fair it does contain all the crucial information. There's definitely room for improvement, but @franciszekjob requested that some time ago and I think it hasn’t been updated since then

crates/forge/tests/integration/gas.rs
franciszekjob: Ok my bad, thx for clarification

crates/forge/tests/integration/gas.rs
franciszekjob: Imo currently it's sufficient but let's request Starkware to improve/clarify their docs, so we can later make these calculations more detailed.

cptartur: > I mainly looked into this topic because there was an issue linked for investigation, otherwise I wouldn't have analyzed it in such depth. In my opinion, the current comment already explains well enough how the calculation was done

I'd prefer to have it written down somewhere, since it might become relevant again in the future if anything changes

crates/forge/tests/integration/gas.rs
ddoktorski: Added a related comment under the `keccak_cost_sierra_gas`

crates/forge/tests/integration/gas.rs
ddoktorski: If you have any suggestions for what could be added to the current description to make it clearer, I’d really appreciate it

franciszekjob (approved):

crates/forge/tests/integration/gas.rs
```diff
@@ -1290,23 +1310,28 @@ fn keccak_cost_sierra_gas() {
     let result = run_test_case(&test, ForgeTrackedResource::SierraGas);
 
     assert_passed(&result);
+    // Note: To calculate the gas cost of the keccak syscall, we need to include the cost of keccak round
+    // https://github.com/starkware-libs/sequencer/blob/v0.15.0-rc.3/crates/blockifier/src/execution/syscalls/syscall_executor.rs#L190
```
cptartur: Tags are basically immutable but it'd be safer to include permalinks.

crates/forge/tests/integration/gas.rs
```diff
@@ -1600,24 +1584,25 @@ fn l1_message_cost_for_proxy_sierra_gas() {
...
-    // todo(2960): verify l2 -> l1 message cost
-    // 29_524 = gas cost of l2 -> l1 message
+    // 29_524 = gas cost of l2 -> l1 message (see `l1_message_cost_sierra_gas` test)
     // 128 = n(2) * 2 * 32
     // 64 = l(2) * 32
     //      -> l = number of class hash updates
     //      -> n = unique contracts updated
-    // 294_240 = cost of 2 deploy syscall (see `deploy_syscall_cost_sierra_gas` test)
-    // 183_120 = cost of 2 call contract syscalls (see `multiple_storage_writes_cost_sierra_gas` test)
+    // 294_240 = 2 * 147_120 = cost of 2 deploy syscall (see `deploy_syscall_cost_sierra_gas` test)
```
cptartur, line 1594: Is that correct? The `deploy_syscall_cost_sierra_gas` asserts the gas to be greater than `151_970`.

crates/forge/tests/integration/gas.rs
cptartur: Let's link to relevant part of blockifier code if there is any.

crates/forge/tests/integration/gas.rs
cptartur: Same here, let's link relevant part of blockifer code if there is any

cptartur (approved):

crates/forge/tests/integration/gas.rs
ddoktorski: Yes, that's correct, `deploy_syscall_cost_sierra_gas` has a higher cost due to calldata. In this test case, the constructor doesn't have any parameters

crates/forge/tests/integration/gas.rs
ddoktorski: Done

crates/forge/tests/integration/gas.rs
ddoktorski: Done

crates/forge/tests/integration/gas.rs
ddoktorski: Done

PR #3739, opened 2025-09-19
Add basic tracing

cptartur:
Closes #

## Introduced changes

-

## Checklist

- [ ] Linked relevant issue
- [ ] Added relevant tests


crates/forge/src/main.rs
```diff
@@ -14,3 +16,83 @@ fn main() {
...
+    let profile_layer = if tracing_profile {
+        let mut path = PathBuf::from(format!(
+            "./snforge-profile-{}.json",
+            Local::now().to_rfc3339()
+        ));
+
+        // Create the file now, so that we early panic, and `fs::canonicalize` will work.
+        let profile_file = fs::File::create(&path).expect("failed to create profile file");
+
+        // Try to canonicalise the path so that it is easier to find the file from logs.
```
franciszekjob:
```suggestion
        // Try to canonicalize the path so that it is easier to find the file from logs.
```

crates/forge/src/main.rs
```diff
@@ -14,3 +16,83 @@ fn main() {
...
+        if let Ok(canonical) = fs::canonicalize(&path) {
+            path = canonical;
+        }
+
+        eprintln!(
+            "this snforge run will output tracing profile to: {}",
```
franciszekjob: nit: Imo it sounds better without "this":
```suggestion
            "`snforge` run will output tracing profile to: {}",
```

crates/forge/src/main.rs
```diff
@@ -14,3 +16,83 @@ fn main() {
...
+            path.display()
+        );
+        eprintln!(
+            "open that file with https://ui.perfetto.dev (or chrome://tracing) to analyze it"
```
franciszekjob:
```suggestion
            "Open that file with https://ui.perfetto.dev (or chrome://tracing) to analyze it"
```

crates/forge/src/main.rs
```diff
@@ -14,3 +16,83 @@ fn main() {
...
+        if let Ok(canonical) = fs::canonicalize(&path) {
+            path = canonical;
+        }
+
+        eprintln!(
+            "`snforge` run will output tracing profile to: {}",
+            path.display()
+        );
+        eprintln!(
+            "Open that file with https://ui.perfetto.dev (or chrome://tracing) to analyze it"
```
franciszekjob, line 68: nit: I think for Chrome we could we provide a ready-to-open link? Might be handy

crates/forge/src/main.rs
```diff
@@ -1,9 +1,11 @@
-use std::sync::Arc;
-
 use forge::{ExitStatus, main_execution};
 use foundry_ui::{UI, components::error::ErrorMessage};
+use std::io::IsTerminal;
+use std::sync::Arc;
+use std::{env, io};
 
 fn main() {
+    let _guard = init_logging();
```
ddoktorski, line 8: Why is it always executed instead of checking the env first?

ddoktorski (approved):

crates/forge/src/main.rs
cptartur: I think `ui.perfetto.dev` is better anyway, don't think the link is that useful.

crates/forge/src/main.rs
cptartur: We want the logic to print debug/trace/etc info to always run. The env just controls the generation of the file.

franciszekjob (approved):

PR #3742, opened 2025-09-22
Don't fetch block numbers nor check rpc version for ignored tests

### LINKED ISSUE #1637: Don't fetch block numbers nor check rpc version for ignored tests when no `--ignored` nor `--include-ignored` flag is specified

snforge

### Description

Currently when no `--ignored` or `--include-ignored` flag is passed we let tests marked with `#[ignore]` pass the filtering here 
https://github.com/foundry-rs/starknet-foundry/blob/f92814d162c99f1e99f31ebb1554fd7fcbf22e80/crates/forge/src/test_filter.rs#L105-L109

https://github.com/foundry-rs/starknet-foundry/blob/f92814d162c99f1e99f31ebb1554fd7fcbf22e80/crates/forge/src/lib.rs#L104-L107

to later print tests as ignored (matches cargo behaviour).

This makes us collect block numbers and check rpc version for tests that won't be executed (in case they are fork tests)

https://github.com/foundry-rs/starknet-foundry/blob/f92814d162c99f1e99f31ebb1554fd7fcbf22e80/crates/forge/src/lib.rs#L126-L127

Additionally, https://github.com/foundry-rs/starknet-foundry/blob/f92814d162c99f1e99f31ebb1554fd7fcbf22e80/crates/forge/src/test_filter.rs#L118-L126 should be changed to:

```rs
impl TestCaseFilter for TestsFilter {
    fn should_be_run(&self, test_case: &TestCaseRunnable) -> bool {
        match self.ignored_filter {
            IgnoredFilter::All => true,
            IgnoredFilter::Ignored => {
                assert!(test_case.ignored); // we filter all not ignored test before
                true
            }
            IgnoredFilter::NotIgnored => !test_case.ignored,
        }
    }
}
```

Abeeujah:
Closes #1637 

## Introduced changes

Block number and RPC version for ignored tests is no longer being checked

- `fork_config` returns `None` for ignored tests
- `should_be_run` `IgnoredFilter::Ignored` match arm asserts that test case is ignored. 

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/forge/src/test_filter.rs
```diff
@@ -123,7 +123,10 @@ impl TestCaseFilter for TestsFilter {
 
         match self.ignored_filter {
             IgnoredFilter::All => true,
-            IgnoredFilter::Ignored => ignored,
+            IgnoredFilter::Ignored => {
+                assert!(ignored);
+                true
+            }
```
cptartur: Don't think we should be including this change, it's technically true given our current logic, but I think it's safer to still check if the test is ignored here since it's not very expensive.

crates/forge/src/run_tests/resolve_config.rs
```diff
@@ -32,12 +32,12 @@ pub async fn resolve_config(
...
-                    block_number_map,
-                    fork_targets,
-                )
-                .await?,
+                fork_config: if case.config.ignored {
+                    None
+                } else {
+                    resolve_fork_config(case.config.fork_config, block_number_map, fork_targets)
+                        .await?
+                },
```
cptartur: This isn't correct, in case forge is run with `--ignored` or `--include-ignored`, they will not have their configs resolved. 

We should probably split separate the steps `test_target_with_config` and `resolve_config` and call them as separate methods, not as a part of `test_package_with_config_resolved`.

Then we can pass `TestsFilter` to `resolve_config` and call `resolve_fork_config` based on if `should_be_run` is true for a test case.

cptartur (changes requested):

crates/forge/src/run_tests/resolve_config.rs
cptartur: We could probably make do without splitting from `test_package_with_config_resolved` and just passing `TestFilter` to the method, but it seems a bit weird to me to use test filter without we even did any kind of filtering with it.

Abeeujah: Passing `tests_filter` to `test_package_with_config_resolved` would require marking the `ignored_filter` field `pub`, that looks like a change that would be needing approval?

Filtering is being done there, tests which the fork config should be resolved or not is going to be dependent on the `tests_filter` argument being passed there, if it wasn't for that, then we'll be running into the issue described above, unless there's another way to resolve that.

crates/forge/src/run_tests/resolve_config.rs
```diff
@@ -19,25 +21,27 @@ pub async fn resolve_config(
...
-                    block_number_map,
-                    fork_targets,
-                )
-                .await?,
+                fork_config: if ignored && matches!(ignored_filter, IgnoredFilter::NotIgnored) {
+                    None
+                } else {
+                    resolve_fork_config(case.config.fork_config, block_number_map, fork_targets)
+                        .await?
+                },
```
cptartur: We are duplicating the filtering logic here, why not use `TestsFilter.should_be_run` instead? 
If the test isn't going to be run for whatever reason (including being ignored), then we don't need to resolve fork config for it.

cptartur (changes requested):

crates/forge/src/run_tests/resolve_config.rs
Abeeujah: I tried using `TestsFilter.should_be_run` but it expects `test_case: &TestCaseWithResolvedConfig` while in this scenario, the test case being filtered is `TestCase<TestCaseConfig>`  that's why I did that, also worth mentioning that `TestsFilter.should_be_run` would require me sharing the whole `TestsFilter` instance instead of just the needed `ignored_filter` field.

cptartur: > I tried using TestsFilter.should_be_run but it expects test_case: &TestCaseWithResolvedConfig while in this scenario, the test case being filtered is TestCase<TestCaseConfig>

What could be done in that case is we should introduce a trait that both `TestCaseConfig` and `TestCaseResolvedConfig` that contains one method like `is_ignored`. Then `should_be_run` and relevant `TestCaseFilter` can be made generic, so they take `TestCase<T>` where T implements the introduced trait.

This way it can be reused between both test_case types and logic can be shared.

> also worth mentioning that TestsFilter.should_be_run would require me sharing the whole TestsFilter instance instead of just the needed ignored_filter field.

Passing a reference to it to the function doesn't incur any significant cost

Abeeujah: Hi @cptartur I did implement this, and requested a review from you, looking forward to getting it reviewed at your earliest convenience

crates/forge/src/run_tests/resolve_config.rs
```diff
@@ -19,28 +24,30 @@ pub async fn resolve_config(
...
                 available_gas: case.config.available_gas,
-                ignored: case.config.ignored
-                    || (env_ignore_fork_tests && case.config.fork_config.is_some()),
+                ignored,
+                fork_config: if ignored && tests_filter.should_be_run(&case) {
+                    None
+                } else {
+                    resolve_fork_config(case.config.fork_config, block_number_map, fork_targets)
+                        .await?
+                },
```
cptartur: I think the condition should be flipped. Also, just checking for if the tests should be run should suffice, we don't need to also check if it was ignored.


```suggestion
                fork_config: tests_filter.should_be_run(&case) {
                    resolve_fork_config(case.config.fork_config, block_number_map, fork_targets)
                        .await?
                } else {
                    None
                },
```

crates/forge/src/test_filter.rs
```diff
@@ -117,9 +118,24 @@ impl TestsFilter {
...
+    fn default() -> Self {
+        Self {
+            name_filter: NameFilter::All,
+            ignored_filter: IgnoredFilter::All,
+            last_failed_filter: false,
+            skip_filter: Vec::new(),
+            failed_tests_cache: FailedTestsCache::default(),
+        }
+    }
+}
```
cptartur: If the default is just for tests, let's implement it behind `#[cfg(tests)]`. I don't think it's a good idea to have a default for this kind of structure.

cptartur (commented):
We should also add some tests checking if fork configs get properly resolved (or not resolved) based on the ignored filter.

crates/forge/src/run_tests/resolve_config.rs
Abeeujah: Flipped, thanks for mentioning, this sounds more intuitive actually.

crates/forge/src/test_filter.rs
Abeeujah: I've put it behind `#[cfg(test)]`

crates/forge/src/run_tests/resolve_config.rs
```diff
@@ -19,30 +24,31 @@ pub async fn resolve_config(
...
-            fuzzer_config: case.config.fuzzer_config,
-            disable_predeployed_contracts: case.config.disable_predeployed_contracts,
-        };
-        test_cases.push(TestCaseWithResolvedConfig::new(
-            &case.name,
-            case.test_details,
-            config,
-        ));
+        let ignored =
+            case.config.ignored || (env_ignore_fork_tests && case.config.fork_config.is_some());
```
cptartur: This could be inlined again

crates/forge/src/run_tests/resolve_config.rs
```diff
@@ -19,30 +24,31 @@ pub async fn resolve_config(
...
+                } else {
+                    None
+                },
+                expected_result: case.config.expected_result,
+                fuzzer_config: case.config.fuzzer_config,
+                disable_predeployed_contracts: case.config.disable_predeployed_contracts,
+            },
+            name: case.name,
+            test_details: case.test_details,
+        });
```
cptartur: Let's try not modify code that is not relevant to the PR changes

crates/forge/src/run_tests/resolve_config.rs
```diff
@@ -161,44 +197,463 @@ mod tests {
...
+
+        let test_target = create_test_target_with_cases(vec![test_case]);
+
+        let fork_targets = vec![create_fork_target(
+            "valid_fork",
+            "https://example.com",
+            BlockId::BlockNumber(100),
+        )];
+
+        let tests_filter = TestsFilter::default();
```
cptartur: I think it would be clearer to create one directly here otherwise you have to know the implementation of default to know if this test makes sense.

crates/forge/src/test_filter.rs
cptartur: I think it will be easier to just create this in tests directly, as I wrote in the comment above.

cptartur (commented):
I see a bunch of tests is failing, not sure if due to changes from this PR. Maybe try updating from master

crates/forge/src/run_tests/resolve_config.rs
Abeeujah: This right here was due to Rust reporting a partial move, the `name`, `test_details`, `expected_result`, `fuzzer_config`, `disable_predeployed_contracts` don't implement copy, so they get partially moved, before I try to borrow `&case` in `should_be_run`. This is why I modified it to be that way.

Abeeujah:
The failing tests seem to look like they all expect std contains some text... `assert_stdout_contains`

cptartur:
> The failing tests seem to look like they all expect std contains some text... `assert_stdout_contains`

I've double checked, the tests are failing because `TestCaseWithResolvedConfig::new` was replaced with direct creation of that struct. We cannot do that because `new` contains logic that sanitizes tests names – a necessary step for all test cases, otherwise they will contain `_test_generated` and `_fuzzer_generated` from our proc macros.

Abeeujah:
> > The failing tests seem to look like they all expect std contains some text... `assert_stdout_contains`
> 
> I've double checked, the tests are failing because `TestCaseWithResolvedConfig::new` was replaced with direct creation of that struct. We cannot do that because `new` contains logic that sanitizes tests names – a necessary step for all test cases, otherwise they will contain `_test_generated` and `_fuzzer_generated` from our proc macros.

You're right, Instantiating with `new` solves it.

crates/forge/src/test_filter.rs
cptartur, line 1: Don't think we need to change any tests in this file.

crates/forge/src/run_tests/resolve_config.rs
```diff
@@ -161,44 +196,497 @@ mod tests {
...
+            })),
+        );
+
+        let test_target = create_test_target_with_cases(vec![test_case]);
+        let tests_filter = TestsFilter::from_flags(
+            None,
+            false,
+            Vec::new(),
+            false,
+            true,
```
cptartur: This could be `false`

crates/forge/src/run_tests/resolve_config.rs
```diff
@@ -161,44 +196,497 @@ mod tests {
...
+            "https://base-fork.com",
+            BlockId::BlockNumber(100),
+        )];
+
+        let tests_filter = TestsFilter::from_flags(
+            None,
+            false,
+            Vec::new(),
+            false,
+            true,
```
cptartur: This could be `false`

crates/forge/src/run_tests/resolve_config.rs
cptartur, line 1: Logic looks good, left some nitpicks for tests

crates/forge/src/run_tests/resolve_config.rs
```diff
@@ -161,44 +196,497 @@ mod tests {
...
+            false,
+            Some(RawForkConfig::Named("non_existent".into())),
+        )]);
+
+        let tests_filter = TestsFilter::from_flags(
+            None,
+            false,
+            Vec::new(),
+            false,
+            true,
```
cptartur:
```suggestion
            false,
```

cptartur (approved):
Looks good, left one more comment.

Let's wait for someone else from the team to review it as well.

ddoktorski (approved):

PR #3748, opened 2025-09-24
Update `scarb-oracle-hint-service` to bring final `shell` protocol implementation

mkaput:


mkaput:
💀 starknet-type-core issue here. LMK when this will be unlocked on `main`

mkaput:
I will recreate this PR soonish

PR #3749, opened 2025-09-24
Update to starknet-devnet 0.6.0

FabijanC:
## Introduced changes

- Update Devnet to 0.6.0
  - This skips version 0.5.1
- Update `mint_token` to use RPC
  - Let me know if there are similar places that require attention.
- Fix expected type of deployment receipt.

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,2 +1,2 @@
 scarb 2.12.1 2.10.1
-starknet-devnet 0.5.0
+starknet-devnet 0.6.0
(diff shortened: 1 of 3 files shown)


cptartur (changes requested):
Some deploy tests are failing, because they incorrectly asserted transaction type to be `Deploy` instead of `Invoke` when using UDC.

Simply changing the enum variant in tests should be enough

franciszekjob (approved):
Thanks 🙏

cptartur (approved):

PR #3751, opened 2025-09-25
Add test for `get_available_gas`

### LINKED ISSUE #3499: Pass gas correctly between runtimes
### Current State

In cheatnet, we currently sometimes reinitialize the gas counter as `i64:: MAX` instead of passing it correctly between runtimes. This doesn't permit users to use `core::testing::get_available_gas` method to assert gas costs of parts of code.

### Objective

Pass gas correctly between runtimes

### Additional Context

Seems that blockifer has logic to do this correctly and it might mostly be possible to mimic it.

ddoktorski:
Closes #3499

commit-id:c7694535

---

**Stack**:
- #3756
- #3755
- #3752
- #3751 ⬅

### DIFF
--- a/crates/forge/tests/integration/get_available_gas.rs
+++ b/crates/forge/tests/integration/get_available_gas.rs
@@ -1,0 +1,54 @@
+use forge_runner::forge_config::ForgeTrackedResource;
+use indoc::indoc;
+use std::path::Path;
+use test_utils::runner::{Contract, assert_passed};
+use test_utils::running_tests::run_test_case;
+use test_utils::test_case;
+
+#[test]
+fn test_get_available_gas() {
+    let test = test_case!(
+        indoc!(
+            r#"
+            use snforge_std::{declare, DeclareResultTrait, ContractClassTrait};
+
+            #[test]
+            fn check_available_gas() {
+                let contract = declare("HelloStarknet").unwrap().contract_class().clone();
+                let (contract_address, _) = contract.deploy(@ArrayTrait::new()).unwrap();
+
+                let gas_before = core::testing::get_available_gas();
+                core::gas::withdraw_gas().unwrap();
+
+                starknet::syscalls::call_contract_syscall(
+                    contract_address, selector!("increase_balance"), array![10].span(),
+                ).unwrap();
+
+                let gas_after = core::testing::get_available_gas();
+                core::gas::withdraw_gas().unwrap();
+
+                let gas_diff = gas_before - gas_after;
+
+                // call_contract syscall: 91_560 gas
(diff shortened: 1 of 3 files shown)


crates/forge/tests/integration/get_available_gas.rs
```diff
@@ -0,0 +1,54 @@
...
+                let gas_diff = gas_before - gas_after;
+
+                // call_contract syscall: 91_560 gas
+                // storage_write and storage_read syscalls: 10_000 gas each
+                let min_expected_gas = 91_560 + 10_000 * 2;
+
+                // Check that gas used is above the expected minimum
+                assert(gas_diff > min_expected_gas, 'Incorrect gas');
+
+                // Allow an arbitrary margin of 10_000 gas
```
franciszekjob, line 39: Let's add a small comment on why we chose 10000

franciszekjob (approved):

crates/forge/tests/integration/get_available_gas.rs
ddoktorski: If there were a good reason, I don’t think it would be called _arbitrary_. It’s simply a small value to cover the remaining computation costs

crates/forge/tests/integration/get_available_gas.rs
```diff
@@ -0,0 +1,54 @@
...
+            r#"
+            use snforge_std::{declare, DeclareResultTrait, ContractClassTrait};
+
+            #[test]
+            fn check_available_gas() {
+                let contract = declare("HelloStarknet").unwrap().contract_class().clone();
+                let (contract_address, _) = contract.deploy(@ArrayTrait::new()).unwrap();
+
+                let gas_before = core::testing::get_available_gas();
+                core::gas::withdraw_gas().unwrap();
```
cptartur, line 21: Why do we need to have explicit calls to `withdraw_gas` in this test?

cptartur (approved):

crates/forge/tests/integration/get_available_gas.rs
ddoktorski: My understanding is that gas is only updated through `withdraw_gas` or `redeposit_gas` calls. Ideally the gas should be updated after each libfunc call, but since `withdraw_gas` also incurs a cost, doing so would be not optimal.

In this case to get an accurate value we need to call `withdraw_gas` right before `get_available_gas`. This is confirmed in the documentation: https://github.com/starkware-libs/cairo/blob/a6ea471b34bd95b24232f6c924368b21a2c11e8d/corelib/src/testing.cairo#L13-L14.

PR #3752, opened 2025-09-25
Exclude syscall resources when building trace data to save

### LINKED ISSUE #3399: Exclude syscall resources when creating trace data to save

ddoktorski:
Closes #3399
Related #3744

commit-id:40786103

---

**Stack**:
- #3756
- #3755
- #3752 ⬅

### DIFF
--- a/crates/forge-runner/src/build_trace_data.rs
+++ b/crates/forge-runner/src/build_trace_data.rs
@@ -5,6 +5,7 @@ use blockifier::execution::syscalls::vm_syscall_utils::{
     SyscallSelector, SyscallUsage, SyscallUsageMap,
 };
 
+use blockifier::blockifier_versioned_constants::VersionedConstants;
 use blockifier::execution::call_info::OrderedEvent;
 use cairo_annotations::trace_data::{
     CairoExecutionInfo, CallEntryPoint as ProfilerCallEntryPoint,
@@ -20,6 +21,7 @@ use cairo_vm::vm::runners::cairo_runner::ExecutionResources;
 use cairo_vm::vm::trace::trace_entry::RelocatedTraceEntry;
 use camino::{Utf8Path, Utf8PathBuf};
 use cheatnet::forking::data::ForkData;
+use cheatnet::runtime_extensions::common::{get_syscalls_gas_consumed, sum_syscall_usage};
 use cheatnet::runtime_extensions::forge_runtime_extension::contracts_data::ContractsData;
 use cheatnet::state::{CallTrace, CallTraceNode};
 use conversions::IntoConv;
@@ -68,8 +70,9 @@ pub fn build_profiler_call_trace(
     ProfilerCallTrace {
         entry_point,
         cumulative_resources: build_profiler_execution_resources(
-            value.used_execution_resources.clone(),
-            value.get_total_used_syscalls(),
+            &value.used_execution_resources,
+            &value.used_syscalls_vm_resources,
+            &value.used_syscalls_sierra_gas,
             value.gas_consumed,
         ),
         used_l1_resources: value.used_l1_resources.clone(),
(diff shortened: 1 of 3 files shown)


crates/forge-runner/src/build_trace_data.rs
```diff
@@ -133,10 +136,21 @@ fn build_profiler_call_trace_node(
...
+    syscall_usage_sierra_gas: &SyscallUsageMap,
     gas_consumed: u64,
 ) -> ProfilerExecutionResources {
+    // Subtract syscall related resources to get the values expected by the profiler.
+    // The profiler operates on resources excluding syscall overhead.
+    let versioned_constants = VersionedConstants::latest_constants();
+    let execution_resources = execution_resources
+        - &versioned_constants.get_additional_os_syscall_resources(syscall_usage_vm_resources);
+    let gas_consumed =
+        gas_consumed - get_syscalls_gas_consumed(syscall_usage_sierra_gas, versioned_constants);
```
THenry14, line 150: I think we should be able to just remove this, as profiler does not depend on these fields with its calculations - we should be able to just pass `execution_resources` and `gas_consumed` to the `ProfilerExecutionResources` as is, as the costs are calculated from the trace. The only caveat being that either one of `gas_consumed` and `execution_resources` is 0 while the other is not, but I think this is already being handled elsewhere (in a place when tracked resource is being set)?

ddoktorski: Do you mean that the entire logic for removing syscall resources might not be necessary? If so, why is it included in the trace at all?

THenry14: From profiler point of view - yes. Why this was introduced in the first place, sadly I don't really know.

ddoktorski: What would you suggest, should we leave the default values or use the actual returned resources as is?
Also, are there any other fields we’re currently including that you think might not be needed?

THenry14: By default values you probably mean 0 for gas_consumed and empty ExecutionResources, right? in that case it's a no-no, as one of them must reflect the tracked resource stack. I'd say we can just use actual returned resources here (but again - let's verify the logic where only one can be set at a time is still true here).

Looking through it briefly it seems these two are the only ones that do not have to be precise, the rest (like syscall usage map or used_l1_resources) are still needed, because these are the things that cannot be calculated from the trace.

cptartur (approved):

crates/forge-runner/src/build_trace_data.rs
ddoktorski: Thanks for the explanation! As mentioned offline, it would be easier to keep this PR as a refactor only, since it's still beneficial to move this logic out of the main execution. Changing the values to ones that include syscall resources would break some of our tests, so that will be done in a separate PR under the scope of #3744

THenry14 (approved):

crates/forge-runner/src/build_trace_data.rs
piotmag769: >as profiler does not depend on these fields with its calculations

It does for cases where it cannot figure out resources by itself - so basically for builtins (in cases where vm resources are used)

>Why this was introduced in the first place, sadly I don't really know

^ that's why

THenry14: ahh, is this for forks?

piotmag769: Not only, for regular entrypoints where you want to profile builtins (which we cannot figure out from current mappings) too. I mean for the case with builtins it could be with or without builtins from syscalls, but if it was with them, profiler has to know the syscall cost map that snforge uses, subtract them and use syscall cost map provided by the user (most of the time these maps are the same ofc)

THenry14: yes, that was my reasoning, as right now we get syscalls (and their resources) from maps either way. 
but you're right that we cannot guarantee using the same maps (this is up to a user), so indeed maybe safer would be to just leave the previous logic. Sorry for the confusion @ddoktorski 🙏

ddoktorski: Thanks for checking @piotmag769 I changed the comment to the initial version

Issue #3603, opened 2025-08-04
Easier devnet account importing


cptartur:
### Current State

Currently accounts have to be imported to sncast using `sncast account import` so that they can be used. These mix with normal accounts, require using consistent devnet seed etc.

### Objective

Design and create an easier devnet account importing functionality

### Additional Context

For example they could be somehow auto-imported, perhaps a dedicated endpoint for that could be created in devnet.

PR #3760, opened 2025-09-26
Add `DevnetProvider`

franciszekjob:
## Introduced changes

Towards #3603 

**Stack**:
- #3762
- #3761
- #3760 ⬅

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/sncast/tests/helpers/constants.rs
```diff
@@ -9,6 +9,7 @@ pub const SEPOLIA_RPC_URL: &str = "http://188.34.188.184:7070/rpc/v0_9";
 pub const URL: &str = "http://127.0.0.1:5055/rpc";
 pub const NETWORK: &str = "testnet";
 pub const SEED: u32 = 1_053_545_548;
+pub const ACCOUNTS_NUMBER: u8 = 20;
```
cptartur: Do we need to have as many devnet accounts as default?

cptartur:
```suggestion
pub const DEVNET_ACCOUNTS_NUMBER: u8 = 20;
```

crates/sncast/tests/helpers/devnet_provider.rs
```diff
@@ -0,0 +1,42 @@
...
+        .expect("Failed to get predeployed accounts");
+
+    assert!(predeployed_accounts.len().to_u8().unwrap() == ACCOUNTS_NUMBER);
+
+    let first_account = &predeployed_accounts[0];
+    let expected_first_account = PredeployedAccount {
+        address: felt!("0x06f4621e7ad43707b3f69f9df49425c3d94fdc5ab2e444bfa0e7e4edeff7992d"),
+        private_key: felt!("0x0000000000000000000000000000000056c12e097e49ea382ca8eadec0839401"),
+        public_key: felt!("0x048234b9bc6c1e749f4b908d310d8c53dae6564110b05ccf79016dca8ce7dfac"),
+    };
```
cptartur: Won't that test fail ever time they bump something in devnet? I wonder if we wouldn't be fine just asserting this is a valid address (method succeeds).

crates/sncast/tests/helpers/constants.rs
franciszekjob: (We had this number already, I didn't add it)
Most likely it was an arbitrary decision, doesn't matter much.

crates/sncast/tests/helpers/devnet_provider.rs
franciszekjob: > Won't that test fail ever time they bump something in devnet?

No, these number will remain the same. 

> if we wouldn't be fine just asserting this is a valid address (method succeeds).

We can, but considering these account details doesn't change, we it's not problematic to keep this check. I'm good with both approaches though.

CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [0.50.0] - 2025-09-29
+
```
ddoktorski: Should be removed

crates/sncast/src/helpers/devnet_provider.rs
```diff
@@ -0,0 +1,116 @@
...
+    #[serde(rename = "devnet_getConfig")]
+    GetConfig,
+
+    /// The `devnet_getPredeployedAccounts` method.
+    #[serde(rename = "devnet_getPredeployedAccounts")]
+    GetPredeployedAccounts,
+}
+
+impl DevnetProvider {
+    /// Constructs a new [`DevnetProvider`] from given url.
```
ddoktorski: The comments in the highlighted section seem a bit redundant

crates/sncast/src/helpers/devnet_provider.rs
```diff
@@ -0,0 +1,116 @@
...
+            .header("Content-Type", "application/json")
+            .json(&json!({
+                "jsonrpc": "2.0",
+                "method": method,
+                "params": params,
+                "id": 1,
+            }))
+            .send()
+            .await
+            .expect("Error occurred during request")
```
ddoktorski:
```suggestion
            .context("Error occurred during request")?
```

crates/sncast/src/helpers/devnet_provider.rs
franciszekjob: Removed

ddoktorski (approved):

crates/sncast/src/helpers/devnet_provider.rs
```diff
@@ -0,0 +1,113 @@
...
+            .await;
+
+        match res {
+            Ok(res_body) => {
+                if let Some(error) = res_body.get("error") {
+                    Err(anyhow::anyhow!(error.to_string()))
+                } else if let Some(result) = res_body.get("result") {
+                    serde_json::from_value(result.clone()).map_err(anyhow::Error::from)
+                } else {
+                    Err(anyhow::anyhow!("Malformed RPC response: {res_body}"))
```
ksew1: i think this should be panic

crates/sncast/src/helpers/devnet_provider.rs
```diff
@@ -0,0 +1,113 @@
...
+        let url = Url::parse(url).expect("Invalid URL");
+        Self {
+            client: Client::new(),
+            url,
+        }
+    }
+}
+
+impl DevnetProvider {
+    async fn send_request<P, R>(&self, method: DevnetProviderMethod, params: P) -> Result<R, Error>
```
ksew1: Lets instead of using `Result<R, anyhow::Error>` do `anyhow::Result<T>`

crates/sncast/src/helpers/devnet_provider.rs
```diff
@@ -0,0 +1,113 @@
...
+            .json(&json!({
+                "jsonrpc": "2.0",
+                "method": method,
+                "params": params,
+                "id": 1,
+            }))
+            .send()
+            .await
+            .context("Error occurred during request")?
+            .json::<serde_json::Value>()
```
ksew1, line 56: `.context("Failed to parse response")`

crates/sncast/src/helpers/devnet_provider.rs
```diff
@@ -0,0 +1,113 @@
...
+    #[serde(rename = "devnet_getConfig")]
+    GetConfig,
+
+    #[serde(rename = "devnet_getPredeployedAccounts")]
+    GetPredeployedAccounts,
+}
+
+impl DevnetProvider {
+    #[must_use]
+    pub fn new(url: &str) -> Self {
```
ksew1, line 28: Since Devnet instances run on localhost, the requests should be fast. We can verify whether the URL actually points to a Devnet instance to maintain this invariant.

franciszekjob: The idea is good, but placing it in constructor will make us end up with async constructor + a side effect, which create not the best pattern.

Maybe let's add a separate `ensure_alive` method which will fail if devnet is not alive. Wdyt?

ksew1: Let's do that

crates/sncast/src/helpers/devnet_provider.rs
```diff
@@ -0,0 +1,113 @@
...
+            .header("Content-Type", "application/json")
+            .json(&json!({
+                "jsonrpc": "2.0",
+                "method": method,
+                "params": params,
+                "id": 1,
+            }))
+            .send()
+            .await
+            .context("Failed to parse response")?
```
ksew1: Sorry i put suggestion in the wrong place 😅. I mean this context to be after ` .json::<serde_json::Value>()`

franciszekjob: Right 😅 , changed

crates/sncast/src/helpers/devnet_provider.rs
```diff
@@ -81,6 +81,27 @@ impl DevnetProvider {
         self.send_request(DevnetProviderMethod::GetPredeployedAccounts, json!({}))
             .await
     }
+
+    /// Ensures the Devnet instance is alive.
+    pub async fn ensure_alive(&self) -> Result<(), Error> {
```
ksew1, line 82: Does it makes sure it is devnet instance or can i be other rpc node? Because it should ensure it is devnet instance

ksew1 (approved):

crates/sncast/src/helpers/devnet_provider.rs
ksew1: Discussed offline

crates/sncast/tests/helpers/devnet_provider.rs
cptartur: Won't them bumping oz version (or other accounts) that are used internally, cause it to be changed?

franciszekjob: Right, oz bump will make it change. Let's remove these assertions then.

cptartur (approved):

PR #3761, opened 2025-09-26
Support inbuilt devnet accounts

franciszekjob:
Towards #3603 

**Stack**:
- #3762
- #3761 ⬅
- #3760

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/sncast/src/lib.rs
```diff
@@ -245,6 +247,32 @@ pub async fn get_nonce(
 }
 
 pub async fn get_account<'a>(
+    account: &str,
+    accounts_file: &Utf8PathBuf,
+    provider: &'a JsonRpcClient<HttpTransport>,
+    url: &str,
+    keystore: Option<&Utf8PathBuf>,
+    ui: &UI,
+) -> Result<SingleOwnerAccount<&'a JsonRpcClient<HttpTransport>, LocalWallet>> {
```
franciszekjob, line 257: note: I pass url and provider instead of the entire rpc args because this avoids lifetime issues. It would be great if provider exposed getter however that's not the case.

crates/sncast/src/helpers/account.rs
```diff
@@ -48,3 +55,56 @@ pub fn load_accounts(accounts_file: &Utf8PathBuf) -> Result<Value> {
...
+}
+
+pub async fn get_account_from_devnet<'a>(
+    account: &str,
+    provider: &'a JsonRpcClient<HttpTransport>,
+    url: &str,
+) -> Result<SingleOwnerAccount<&'a JsonRpcClient<HttpTransport>, LocalWallet>> {
+    let account_number: u8 = account
+        .strip_prefix("devnet-")
+        .map(|s| s.parse::<u8>().expect("Invalid devnet account number"))
```
cptartur, line 89: I think `expect` you have here will just panic with the message in case parsing failed. `context` from below applies to cases where map resolves to `None`.

crates/sncast/src/starknet_commands/multicall/mod.rs
```diff
@@ -50,12 +50,15 @@ pub async fn multicall(
         }
         starknet_commands::multicall::Commands::Run(run) => {
             let provider = run.rpc.get_provider(&config, ui).await?;
+            let url = run.rpc.get_url(&config.url).expect("Failed to get url");
```
cptartur: Is it safe to use `expect` here? Shouldn't we handle the error?

crates/sncast/tests/e2e/devnet_accounts.rs
```diff
@@ -0,0 +1,151 @@
...
+    ];
+
+    let snapbox = runner(&args).current_dir(temp_dir.path());
+    let output = snapbox.assert().failure();
+
+    assert_stderr_contains(
+        output,
+        indoc! {
+            "
+            Error: {\"code\":-32601,\"message\":\"Method not found\"}
```
cptartur: It would be hard for the user to figure out he's using incorrect account just from this error. 
Is it coming from the logic for fetching the devnet account or is the contract call just failing?

crates/sncast/src/helpers/account.rs
franciszekjob: Changed to `context`

crates/sncast/src/starknet_commands/multicall/mod.rs
franciszekjob: Changed to `context`

> Shouldn't we handle the error?

We mostly handle errors coming from provider/chain

crates/sncast/tests/e2e/devnet_accounts.rs
franciszekjob: Right, in case when someone uses devnet account with network not being devnet, they'll get such error

`The provided network is not a Devnet instance. Devnet accounts can only be used with Devnet.`

crates/sncast/src/helpers/account.rs
```diff
@@ -48,3 +55,67 @@ pub fn load_accounts(accounts_file: &Utf8PathBuf) -> Result<Value> {
 
     Ok(accounts)
 }
+
+pub fn check_account_exists(
```
ksew1, line 59: I guess we are aware that this doesn't check keystore accounts?

crates/sncast/src/helpers/account.rs
```diff
@@ -48,3 +55,64 @@ pub fn load_accounts(accounts_file: &Utf8PathBuf) -> Result<Value> {
...
+    account_name: &str,
+    network_name: &str,
+    accounts_file: &Utf8PathBuf,
+) -> Result<bool> {
+    check_account_file_exists(accounts_file)?;
+
+    let accounts: HashMap<String, HashMap<String, AccountData>> =
+        read_and_parse_json_file(accounts_file)?;
+
+    if let Some(network_accounts) = accounts.get(network_name) {
```
ksew1: What do you think about this?
```rs
    accounts
        .get(network_name)
        .map(|network_accounts| network_accounts.contains_key(account_name))
        .ok_or_else(|| anyhow!("Network with name {network_name} does not exist in accounts file"))
```

crates/sncast/src/helpers/account.rs
```diff
@@ -48,3 +55,64 @@ pub fn load_accounts(accounts_file: &Utf8PathBuf) -> Result<Value> {
...
+
+    let devnet_config = devnet_provider.get_config().await;
+    let devnet_config = match devnet_config {
+        Ok(config) => config,
+        Err(err) => {
+            return Err(err);
+        }
+    };
+
+    if account_number > devnet_config.total_accounts || account_number == 0 {
```
ksew1: We can use `ensure!`

crates/sncast/src/lib.rs
```diff
@@ -245,6 +247,32 @@ pub async fn get_nonce(
...
+    let chain_id = get_chain_id(provider).await?;
+    let network_name = chain_id_to_network_name(chain_id);
+    let is_devnet_account = is_devnet_account(account);
+    let exists_in_accounts_file = check_account_exists(account, &network_name, accounts_file)?;
+
+    if is_devnet_account && exists_in_accounts_file {
```
ksew1: What do you think about `match`? I find it more readable 

```rs
    match (is_devnet, exists_in_file) {
        (true, true) => {
            ui.println(&WarningMessage::new(format!(
                "Using account {account} from accounts file {accounts_file}. \
                 To use an inbuilt devnet account, please rename your existing account."
            )));
            ui.print_blank_line();
            get_account_from_accounts_file(account, accounts_file, provider, keystore).await
        }
        (true, false) => get_account_from_devnet(account, provider, url).await,
        _ => get_account_from_accounts_file(account, accounts_file, provider, keystore).await,
    }
```

ksew1 (approved):
Minor suggestions. Otherwise LGTM 😎

crates/sncast/src/helpers/account.rs
franciszekjob: Yes, but imo holding a devnet account in keystore is a very rare case, that's why I didn't handle it.

crates/sncast/src/lib.rs
ddoktorski: Would it be possible to define `DevnetProvider` as a trait and then implement it for `JsonRpcClient`?

crates/sncast/src/lib.rs
```diff
@@ -245,6 +247,37 @@ pub async fn get_nonce(
...
+    let chain_id = get_chain_id(provider).await?;
+    let network_name = chain_id_to_network_name(chain_id);
+    let is_devnet_account = is_devnet_account(account);
+    let exists_in_accounts_file = check_account_exists(account, &network_name, accounts_file)?;
+
+    match (is_devnet_account, exists_in_accounts_file) {
+        (true, true) => {
+            ui.println(&WarningMessage::new(format!(
+                "Using account {account} from accounts file {accounts_file}. \
+                 To use an inbuilt devnet account, please rename your existing account."
```
cptartur:
```suggestion
                 To use an inbuilt devnet account, please rename your existing account or use a devnet account with a different number."
```

Maybe this will be better from the UX standpoint than having users renaming accounts immediately.

crates/sncast/tests/e2e/devnet_accounts.rs
```diff
@@ -0,0 +1,148 @@
...
+    ];
+
+    let snapbox = runner(&args).current_dir(temp_dir.path());
+    let output = snapbox.assert().failure();
+
+    assert_stderr_contains(
+        output,
+        format! {"Error: Node at {SEPOLIA_RPC_URL} is not responding to the Devnet health check (GET `/is_alive`). It may not be a Devnet instance or it may be down."
+        },
+    );
```
cptartur, line 147: I think we should also have a logic that if user uses `--network` flag with sepolia or mainnet then using inbuilt devnet accounts fails immediately, what do you think?

One possible problem is if someone has an account named `devnet-<number>` for mainnet/sepolia, but maybe in this case the check could be skippable by some env var or something?

crates/sncast/src/lib.rs
franciszekjob: Tbh I would stay with currently implementation, don't want to mix devnet and regular RPC API. 
If you seriously opt to change it let me know cc @cptartur @ksew1 .

crates/sncast/tests/e2e/devnet_accounts.rs
franciszekjob: Hmm, generally this makes sense, but ultimately it sums up to (imo) a bit too complicated logic, both in the codebase and for user (considering these env etc.), don't you think so 🤔 ?

cptartur: Problem with not adding that is if user, for whatever reason, tries to do something like `sncast --account devnet-1 ... --network sepolia`, they will get an error like `Node at https://starknet-sepolia.public.blastapi.io/rpc/0_9 is not responding to the Devnet ...`

This is not the end of the world but it might be slightly confusing since the user never really sees or inputs this URL.

cc @ksew1 for another opinion

franciszekjob: This is a valid point - I agree.
Let me rework this.

ksew1: I agree with @cptartur. But i would skip logic for env vars. Let's just disallow devnet-<number> mainnet/sepolia.

crates/sncast/src/lib.rs
cptartur: I think it's fine to have it separate

crates/sncast/tests/e2e/devnet_accounts.rs
cptartur: I think not having env vars is fine as well, let's just add the disallowing logic.

cptartur (commented):
Please re-request once addressed

crates/sncast/tests/e2e/devnet_accounts.rs
franciszekjob: Done

crates/sncast/src/lib.rs
```diff
@@ -245,6 +249,51 @@ pub async fn get_nonce(
...
+    let network_name = chain_id_to_network_name(chain_id);
+    let account = &config.account;
+    let is_devnet_account = is_devnet_account(account);
+
+    if is_devnet_account
+        && let Some(network) = rpc_args.network
+        && (network == Network::Mainnet || network == Network::Sepolia)
+    {
+        bail!(format!(
+            "Devnet accounts cannot be used with `{network}` for `--network` flag."
```
cptartur:
```suggestion
            "Devnet accounts cannot be used with `--network {network}` flag."
```

cptartur (approved):

PR #3766, opened 2025-09-29
Bump deps, starknet types core, rust

cptartur:
Closes #

## Introduced changes

- This is using a blockifer fork, needed to comment out the code using `size-of`, because types from new versions of starknet-types-core didn't have it and it didn't compile. 

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/sncast/tests/e2e/account/import.rs
```diff
@@ -600,7 +600,7 @@ pub async fn test_invalid_private_key_in_file() {
         output,
         indoc! {r"
         Command: account import
-        Error: Failed to obtain private key from the file my_private_key: Failed to create Felt from string
+        Error: Failed to obtain private key from the file my_private_key: failed to create Felt from string: invalid dec string
```
cptartur, line 603: So this is technically a change that could be included in the changelog. Should we do that? 

It's hard to say what exactly changed though, since there might have been different changes in starnet-types-core that our tests don't detect.

MKowalski8: When we bumped `starknet-types-core` to 0.1.9 we didn't inform about that (as far as I know), also when we reverted `starknet-types-core` to previous version it had changed and there was no mention. So if we hadn't done it before, I wouldn't be doing it now.

MKowalski8 (approved):

Cargo.toml
```diff
@@ -33,26 +33,26 @@ license = "MIT"
 license-file = "LICENSE"
 
 [workspace.dependencies]
-blockifier = { version = "0.15.0-rc.3", features = ["testing", "tracing"] }
+blockifier = { git = "https://github.com/cptartur/sequencer.git", branch = "main-v0.14.0-fork", features = ["testing", "tracing"] }
```
ddoktorski: Can we use the fork from https://github.com/software-mansion-labs/sequencer instead?

cptartur: Changed

ddoktorski (approved):

PR #3771, opened 2025-10-01
Display deploy hint command after `sncast declare`

### LINKED ISSUE #2758: Display constructor arguments in deploy command hint

sncast

### Description

When contract is declared, a hint with deploy command is displayed. However, when contract's constructor takes some arguments, they should be also displayed. In the scope of this issue, we should:
- implement a mechanism which fetches contract's constructor
- if constructor has params -> display them in hint deploy command

franciszekjob:
Closes #2758

## Introduced changes

`sncast declare` command now outputs a ready-to-use deployment command after successful declaration.

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/sncast/src/response/declare.rs
```diff
@@ -129,3 +134,98 @@ impl OutputLink for DeclareTransactionResponse {
...
+        formatdoc!(
+            "
+            To deploy a contract of this class, run:
+            {command}
+            "
+        )
+    }
+
+    fn json(&self) -> Value {
+        json!(self)
```
ksew1: Does it even make sense in json? I think i would omit this message there

crates/sncast/src/response/declare.rs
```diff
@@ -129,3 +134,98 @@ impl OutputLink for DeclareTransactionResponse {
...
+impl DeployCommandMessage {
+    pub fn new(
+        abi: &[AbiEntry],
+        response: &DeclareTransactionResponse,
+        account: &str,
+        accounts_file: &Utf8PathBuf,
+        rpc_url: Option<&str>,
+        network: Option<&Network>,
+    ) -> Result<Self, Error> {
+        let mut constructor_entries = abi
```
ksew1: Maybe something like this to be more streamlined

```rs
        let arguments = abi.iter().find_map(|entry| {
            if let AbiEntry::Constructor(constructor) = entry {
                let args = generate_constructor_placeholder_args(constructor.clone());
                (!args.is_empty()).then_some(args)
            } else {
                None
            }
        });

        let network_flag = generate_network_flag(rpc_url, network);
        let accounts_file_str = accounts_file.to_string();
        let accounts_file = (!accounts_file_str
            .contains("starknet_accounts/starknet_open_zeppelin_accounts.json"))
        .then_some(accounts_file_str);

        Ok(Self {
            account: account.to_string(),
            accounts_file,
            class_hash: response.class_hash,
            arguments,
            network_flag,
        })
```

crates/sncast/src/response/declare.rs
```diff
@@ -129,3 +134,98 @@ impl OutputLink for DeclareTransactionResponse {
...
+            class_hash: response.class_hash,
+            arguments,
+            network_flag,
+        })
+    }
+}
+
+impl Message for DeployCommandMessage {
+    fn text(&self) -> String {
+        let command = "sncast";
```
ksew1: Maybe this?

```rs
 let mut command = String::from("sncast");

        if let Some(file) = &self.accounts_file {
            write!(command, " --accounts-file {file}").unwrap();
        }

        write!(command, " --account {}", self.account).unwrap();
        write!(
            command,
            " deploy --class-hash {}",
            self.class_hash.into_hex_string()
        )
        .unwrap();

        if let Some(arguments) = &self.arguments {
            write!(command, " --arguments '{arguments}'").unwrap();
        }

        write!(command, " {}", self.network_flag).unwrap();

        formatdoc!(
            "
            To deploy a contract of this class, run:
            {command}
            "
        )
```

crates/sncast/src/response/declare.rs
franciszekjob: It's required by `Message` trait

ksew1: Yeah, I know but this message is only helpful in human mode. Let's not dump everything into machine mode just because it is needed in human

ksew1: It makes json mode less readable and less usable in scripts

franciszekjob: Ok, makes sense. Now it returns `Value:Null`

ksew1 (approved):

crates/sncast/tests/data/contracts/contract_with_constructor_params/Scarb.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "contract_with_constructor_params"
+version = "0.1.0"
+edition = "2024_07"
+
+[dependencies]
+starknet = ">=2.0.2"
```
ddoktorski: nit:
```suggestion
starknet = ">=2.10.1"
```

crates/sncast/src/main.rs
```diff
@@ -291,7 +292,7 @@ async fn run_async_command(cli: Cli, config: CastConfig, ui: &UI) -> Result<()>
             .expect("Failed to build contract");
 
             let result = starknet_commands::declare::declare(
-                declare,
+                declare.clone(),
```
ddoktorski: Can't we pass it by reference?

crates/sncast/src/main.rs
```diff
@@ -311,8 +312,35 @@ async fn run_async_command(cli: Cli, config: CastConfig, ui: &UI) -> Result<()>
 
             let block_explorer_link =
                 block_explorer_link_if_allowed(&result, provider.chain_id().await?, &rpc, &config);
+
+            let deploy_command_message = if let Ok(response) = &result {
+                let contract_artifacts = artifacts.get(&declare.contract.clone()).ok_or(
+                    StarknetCommandError::ContractArtifactsNotFound(ErrorData {
+                        data: ByteArray::from(declare.contract.as_str()),
+                    }),
+                )?;
```
ddoktorski: What about changing the signature of the `declare` function to take `StarknetContractArtifacts` instead of `&HashMap<String, StarknetContractArtifacts>` so there is no need to repeat this operation?

crates/sncast/src/response/declare.rs
```diff
@@ -129,3 +135,97 @@ impl OutputLink for DeclareTransactionResponse {
...
+
+        if let Some(arguments) = &self.arguments {
+            write!(command, " --arguments '{arguments}'").unwrap();
+        }
+
+        write!(command, " {}", self.network_flag).unwrap();
+
+        formatdoc!(
+            "
+            To deploy a contract of this class, run:
```
ddoktorski: I'd somehow include information that the arguments are only placeholders, because right now it suggests that it's a ready to copy and run command, which is not the case if there are any arguments

crates/sncast/src/main.rs
franciszekjob: Changed

crates/sncast/src/response/declare.rs
franciszekjob:
```
Note: replace the placeholders in `--arguments` with your actual values.
```
Is something like this enough?

franciszekjob: Changed

crates/sncast/src/main.rs
franciszekjob: Good idea. However, this results in a bunch of changes (mostly in sncast scripts and their tests), so let's resolve this in https://github.com/foundry-rs/starknet-foundry/issues/3785. Linked as todo in code.

crates/sncast/src/main.rs
```diff
@@ -309,8 +310,35 @@ async fn run_async_command(cli: Cli, config: CastConfig, ui: &UI) -> Result<()>
...
```
cptartur: It's a bit weird to me to have any error handling here at all. If the declare succeed then mustn't the contract exist? 
I think a simple unwrap/expect should do for all of this.

crates/sncast/src/response/declare.rs
```diff
@@ -129,3 +135,103 @@ impl OutputLink for DeclareTransactionResponse {
...
+    constructor
+        .inputs
+        .into_iter()
+        .map(|input| {
+            let input_type = input
+                .r#type
+                .split("::")
+                .last()
+                .expect("Failed to get last part of input type");
+            format!("<{} ({})>", input.name, input_type)
```
cptartur: Bit of a nitpick, but wouldn't a format like `name: Type` or `<name: Type>` be more familiar for cairo devs?

crates/sncast/src/response/declare.rs
```diff
@@ -129,3 +135,103 @@ impl OutputLink for DeclareTransactionResponse {
...
+
+        Ok(Self {
+            account: account.to_string(),
+            accounts_file,
+            class_hash: response.class_hash,
+            arguments,
+            network_flag,
+        })
+    }
+}
```
cptartur: Nit, but some of these are using a helper functions and some not, I think it would be a bit more readable if logic for generating `arguments` and `accounts_file` was moved to a helper as well.

crates/sncast/src/main.rs
```diff
@@ -309,8 +310,35 @@ async fn run_async_command(cli: Cli, config: CastConfig, ui: &UI) -> Result<()>
...
+                    serde_json::from_str(&contract_artifacts.sierra)
+                        .context("Failed to parse sierra artifact")?;
+
+                Some(DeployCommandMessage::new(
+                    &contract_definition.abi,
+                    response,
+                    &config.account,
+                    &config.accounts_file,
+                    rpc.get_url(&config.url).as_deref(),
+                    rpc.network.as_ref(),
```
ddoktorski: nit: Can we pass simply `rpc` instead of these 2 arguments?

crates/sncast/src/main.rs
ddoktorski: I think TODO should be linked also in this place. Could you make this change as follow up PR?

ddoktorski (approved):

crates/sncast/src/response/declare.rs
franciszekjob: Yeah, definitely, changed.

crates/sncast/src/main.rs
franciszekjob: Hmm, it's a small thing, but maybe `expect` or `unreachable!` would be better, considering that the contract is compiled earlier. Will change it.

crates/sncast/src/main.rs
franciszekjob: Will do

crates/sncast/src/main.rs
franciszekjob: Changed so it takes `network_flag` param

crates/sncast/src/response/declare.rs
franciszekjob: Changed

PR #3780, opened 2025-10-03
Remove unused response structs

cptartur:
commit-id:fcef8826

# Conflicts:
#	crates/sncast/src/response/declare.rs

---

**Stack**:
- #3783
- #3782
- #3781
- #3780 ⬅

⚠️ *Part of a stack created by [spr](https://github.com/ejoffe/spr). Do not merge manually using the UI - doing so may have unexpected results.*

PR #3779, opened 2025-10-03
Handle redeclaration in estimate fee, closes #3253

cptartur:
**Stack**:
- #3783
- #3782
- #3781
- #3780
- #3779 ⬅

⚠️ *Part of a stack created by [spr](https://github.com/ejoffe/spr). Do not merge manually using the UI - doing so may have unexpected results.*

PR #3781, opened 2025-10-03
Change declare arguments

cptartur:
**Stack**:
- #3783
- #3782
- #3781 ⬅
- #3780
- #3779

### DIFF
--- a/crates/sncast/src/starknet_commands/declare.rs
+++ b/crates/sncast/src/starknet_commands/declare.rs
@@ -31,8 +31,8 @@ use std::sync::Arc;
 #[command(about = "Declare a contract to starknet", long_about = None)]
 pub struct Declare {
     /// Contract name
-    #[arg(short = 'c', long = "contract-name")]
-    pub contract: String,
+    #[arg(short = 'c', long)]
+    pub contract_name: String,
 
     #[command(flatten)]
     pub fee_args: FeeArgs,
@@ -50,8 +50,11 @@ pub struct Declare {
 }
 
 // TODO(#3785)
+#[expect(clippy::too_many_arguments)]
 pub async fn declare(
-    declare: &Declare,
+    contract_name: String,
+    fee_args: FeeArgs,
+    nonce: Option<Felt>,
     account: &SingleOwnerAccount<&JsonRpcClient<HttpTransport>, LocalWallet>,
     artifacts: &HashMap<String, StarknetContractArtifacts>,
     wait_config: WaitForTx,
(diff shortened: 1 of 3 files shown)


crates/sncast/src/starknet_commands/declare.rs
```diff
@@ -49,8 +49,11 @@ pub struct Declare {
     pub rpc: RpcArgs,
 }
 
+#[expect(clippy::too_many_arguments)]
 pub async fn declare(
-    declare: Declare,
+    contract: String,
```
franciszekjob: nit: In consistency with flag name:
```suggestion
    contract_name: String,
```

franciszekjob (approved):

crates/sncast/src/starknet_commands/declare.rs
```diff
@@ -50,8 +50,11 @@ pub struct Declare {
 }
 
 // TODO(#3785)
+#[expect(clippy::too_many_arguments)]
 pub async fn declare(
-    declare: &Declare,
+    contract: String,
+    fee_args: FeeArgs,
+    nonce: Option<Felt>,
```
ddoktorski: What is the benefit of making this change?

cptartur: I don't need to construct `Declare` struct to pass it to the function, especially since some of it's contents are ignored by `declare`. 

It made sense to have it when declare was only called from main once, but now it is used by deploy as well.

ddoktorski: Wouldn't it be better to still have something like `DeclareArgs`, so we don't need to branch lint checks?

crates/sncast/src/starknet_commands/declare.rs
ddoktorski: Why was it resolved?

crates/sncast/src/starknet_commands/declare.rs
cptartur: This way we'd have a separate `Declare` struct with CLI args and second for declare transaction args, which we would have to construct from the first one.

Not sure if this gives any benefit in terms of maintaining the code or performance.

ddoktorski (approved):

crates/sncast/src/starknet_commands/declare.rs
cptartur: I've changed the name in the struct but forgot abut this one. Thanks for checking this

cptartur:
✓ Commit merged in pull request [#3783](https://github.com/foundry-rs/starknet-foundry/pull/3783)

PR #3782, opened 2025-10-03
Allow disabling printing in `wait_for_tx`

cptartur:
**Stack**:
- #3783
- #3782 ⬅
- #3781
- #3780
- #3779

⚠️ *Part of a stack created by [spr](https://github.com/ejoffe/spr). Do not merge manually using the UI - doing so may have unexpected results.*

PR #3783, opened 2025-10-03
Allow declaring with deployment

### LINKED ISSUE #3253: Improve error message when trying to redeclare contract using `sncast declare`

### LINKED ISSUE #3606: Implement `declare-deploy` command in cast
### Current State

Currently you need to declare and deploy the contract each time something changes in it. Each time it requires first declaring contracts by their names and then deploying by their class hashes.

### Objective

Desing and implement flow simplyfing declaring and deploying for testing

### Additional Context

There are multiple approaches we could take

1. Restart the effort to implement `sncast declare-deploy` command that wraps declaration and deployment into a single step. This would make deployment for testings bit less cumbersome.
2. Introduce a functionality strictly for declaring and deploying contracts on devnet that would work in a more automatic manner.

cptartur:
commit-id:82e786a7

# Conflicts:
#	crates/sncast/src/main.rs

---

**Stack**:
- #3783 ⬅
- #3782
- #3781
- #3780

### DIFF
--- a/crates/sncast/src/main.rs
+++ b/crates/sncast/src/main.rs
@@ -260,6 +260,7 @@ async fn run_async_command(cli: Cli, config: CastConfig, ui: &UI) -> Result<()> 
     let wait_config = WaitForTx {
         wait: cli.wait,
         wait_params: config.wait_params,
+        show_ui_outputs: true,
     };
 
     match cli.command {
--- a/crates/sncast/src/main.rs
+++ b/crates/sncast/src/main.rs
@@ -291,7 +291,9 @@ async fn run_async_command(cli: Cli, config: CastConfig, ui: &UI) -> Result<()> 
             .expect("Failed to build contract");
 
             let result = starknet_commands::declare::declare(
-                &declare,
+                declare.contract_name.clone(),
+                declare.fee_args,
+                declare.nonce,
                 &account,
                 &artifacts,
                 wait_config,
@@ -315,7 +317,7 @@ async fn run_async_command(cli: Cli, config: CastConfig, ui: &UI) -> Result<()> 
             let deploy_command_message = if let Ok(response) = &result {
                 // TODO(#3785)
                 let contract_artifacts = artifacts
-                    .get(&declare.contract.clone())
+                    .get(&declare.contract_name)
                     .expect("Failed to get contract artifacts");
                 let contract_definition: SierraClass =
                     serde_json::from_str(&contract_artifacts.sierra)
--- a/crates/sncast/src/main.rs
+++ b/crates/sncast/src/main.rs
@@ -1,3 +1,4 @@
+use crate::starknet_commands::declare::declare;
 use crate::starknet_commands::declare_from::DeclareFrom;
 use crate::starknet_commands::deploy::DeployArguments;
 use crate::starknet_commands::multicall;
@@ -11,6 +12,7 @@ use anyhow::{Context, Result, bail};
 use camino::Utf8PathBuf;
 use clap::{CommandFactory, Parser, Subcommand};
 use configuration::load_config;
+use conversions::IntoConv;
 use data_transformer::transform;
 use foundry_ui::components::warning::WarningMessage;
 use foundry_ui::{Message, UI};
@@ -24,7 +26,10 @@ use sncast::helpers::rpc::generate_network_flag;
 use sncast::helpers::scarb_utils::{
     BuildConfig, assert_manifest_path_exists, build_and_load_artifacts, get_package_metadata,
 };
-use sncast::response::declare::{DeclareResponse, DeployCommandMessage};
+use sncast::response::declare::{
+    AlreadyDeclaredResponse, DeclareResponse, DeclareTransactionResponse, DeployCommandMessage,
+};
+use sncast::response::deploy::{DeployResponse, DeployResponseWithDeclare};
 use sncast::response::errors::handle_starknet_command_error;
 use sncast::response::explorer_link::block_explorer_link_if_allowed;
 use sncast::response::transformed_call::transform_response;
@@ -393,9 +398,12 @@ async fn run_async_command(cli: Cli, config: CastConfig, ui: &UI) -> Result<()> 
 
         Commands::Deploy(deploy) => {
             let Deploy {
+                contract_identifier: identifier,
                 arguments,
                 fee_args,
                 rpc,
+                mut nonce,
+                package,
                 ..
             } = deploy;
 
@@ -404,27 +412,98 @@ async fn run_async_command(cli: Cli, config: CastConfig, ui: &UI) -> Result<()> 
             let account =
                 get_account(&config, &provider, &rpc, config.keystore.as_ref(), ui).await?;
 
+            let (class_hash, declare_response) = if let Some(class_hash) = identifier.class_hash {
+                (class_hash, None)
+            } else if let Some(contract_name) = identifier.contract_name {
+                let manifest_path = assert_manifest_path_exists()?;
+                let package_metadata = get_package_metadata(&manifest_path, &package)?;
+                let artifacts = build_and_load_artifacts(
+                    &package_metadata,
+                    &BuildConfig {
+                        scarb_toml_path: manifest_path,
+                        json: cli.json,
+                        profile: cli.profile.unwrap_or("release".to_string()),
+                    },
+                    false,
+                    ui,
+                )
+                .expect("Failed to build contract");
+
+                let declare_result = declare(
+                    contract_name,
+                    fee_args.clone(),
+                    nonce,
+                    &account,
+                    &artifacts,
+                    WaitForTx {
+                        wait: true,
+                        wait_params: wait_config.wait_params,
+                        // Only show outputs if user explicitly provides `--wait` flag
+                        show_ui_outputs: wait_config.wait,
+                    },
+                    true,
+                    ui,
+                )
+                .await
+                .map_err(handle_starknet_command_error);
+
+                // Increment nonce after successful declare if it was explicitly provided
+                nonce = nonce.map(|n| n + Felt::ONE);
+
+                match declare_result {
+                    Ok(DeclareResponse::AlreadyDeclared(AlreadyDeclaredResponse {
+                        class_hash,
+                    })) => (class_hash.into_(), None),
+                    Ok(DeclareResponse::Success(declare_transaction_response)) => (
+                        declare_transaction_response.class_hash.into_(),
+                        Some(declare_transaction_response),
+                    ),
+                    Err(err) => {
+                        process_command_result::<DeclareTransactionResponse>(
+                            "deploy",
+                            Err(err),
+                            ui,
+                            None,
+                        );
+                        return Ok(());
+                    }
+                }
+            } else {
+                unreachable!("Either `--class_hash` or `--contract_name` must be provided");
+            };
+
             // safe to unwrap because "constructor" is a standardized name
             let selector = get_selector_from_name("constructor").unwrap();
 
-            let contract_class = get_contract_class(deploy.class_hash, &provider).await?;
+            let contract_class = get_contract_class(class_hash, &provider).await?;
 
             let arguments: Arguments = arguments.into();
             let calldata = arguments.try_into_calldata(contract_class, &selector)?;
 
             let result = starknet_commands::deploy::deploy(
-                deploy.class_hash,
+                class_hash,
                 &calldata,
                 deploy.salt,
                 deploy.unique,
                 fee_args,
-                deploy.nonce,
+                nonce,
                 &account,
                 wait_config,
                 ui,
             )
             .await
             .map_err(handle_starknet_command_error);
+
+            let result = if let Some(declare_response) = declare_response {
+                result.map(|r| {
+                    DeployResponse::WithDeclare(DeployResponseWithDeclare::from_responses(
+                        &r,
+                        &declare_response,
+                    ))
+                })
+            } else {
+                result.map(DeployResponse::Standard)
+            };
 
             let block_explorer_link =
                 block_explorer_link_if_allowed(&result, provider.chain_id().await?, &rpc, &config);
--- a/crates/sncast/src/response/deploy.rs
+++ b/crates/sncast/src/response/deploy.rs
@@ -1,38 +1,31 @@
+use super::command::CommandResponse;
 use crate::helpers::block_explorer::LinkProvider;
+use crate::response::cast_message::SncastMessage;
+use crate::response::declare::DeclareTransactionResponse;
+use crate::response::explorer_link::OutputLink;
 use conversions::string::IntoPaddedHexStr;
 use conversions::{padded_felt::PaddedFelt, serde::serialize::CairoSerialize};
 use foundry_ui::Message;
 use foundry_ui::styling;
 use indoc::formatdoc;
 use serde::{Deserialize, Serialize};
-use serde_json::Value;
-use serde_json::json;
-
-use super::{command::CommandResponse, explorer_link::OutputLink};
-use crate::response::cast_message::SncastMessage;
+use serde_json::{Value, json};
 
 #[derive(Clone, Serialize, Deserialize, CairoSerialize, Debug, PartialEq)]
-pub struct DeployResponse {
-    pub contract_address: PaddedFelt,
-    pub transaction_hash: PaddedFelt,
+#[serde(untagged)]
+pub enum DeployResponse {
+    Standard(StandardDeployResponse),
+    WithDeclare(DeployResponseWithDeclare),
 }
 
 impl CommandResponse for DeployResponse {}
 
 impl Message for SncastMessage<DeployResponse> {
     fn text(&self) -> String {
-        styling::OutputBuilder::new()
-            .success_message("Deployment completed")
-            .blank_line()
-            .field(
-                "Contract Address",
-                &self.command_response.contract_address.into_padded_hex_str(),
-            )
-            .field(
-                "Transaction Hash",
-                &self.command_response.transaction_hash.into_padded_hex_str(),
-            )
-            .build()
+        match &self.command_response {
+            DeployResponse::Standard(response) => response.text(),
+            DeployResponse::WithDeclare(response) => response.text(),
+        }
     }
 
     fn json(&self) -> Value {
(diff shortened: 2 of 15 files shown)


crates/sncast/src/response/deploy.rs
```diff
@@ -1,62 +1,138 @@
...
-                "details": err.to_string()
-            })
-        })
-    }
+#[derive(Clone, Serialize, Deserialize, CairoSerialize, Debug, PartialEq)]
+pub struct DeployResponseWithDeclare {
+    pub contract_address: PaddedFelt,
+    pub class_hash: PaddedFelt,
+    pub deploy_transaction_hash: PaddedFelt,
+    pub declare_transaction_hash: PaddedFelt,
```
franciszekjob: nit:
```suggestion
    contract_address: PaddedFelt,
    class_hash: PaddedFelt,
    deploy_transaction_hash: PaddedFelt,
    declare_transaction_hash: PaddedFelt,
```

crates/sncast/src/starknet_commands/deploy.rs
```diff
@@ -17,12 +17,22 @@ use starknet::providers::jsonrpc::HttpTransport;
...
 
+#[derive(Args, Debug, Clone)]
+#[group(required = true, multiple = false)]
+pub struct DeployIdentifier {
+    /// Class hash of contract to deploy
+    #[arg(short = 'g', long)]
+    pub class_hash: Option<Felt>,
+
+    #[arg(long, conflicts_with = "class_hash")]
+    pub contract_name: Option<String>,
```
franciszekjob: nit: For consistency:
```suggestion
    /// Contract name
    #[arg(long, conflicts_with = "class_hash")]
    pub contract_name: Option<String>,
```

crates/sncast/src/starknet_commands/deploy.rs
```diff
@@ -17,12 +17,22 @@ use starknet::providers::jsonrpc::HttpTransport;
 use starknet::signers::LocalWallet;
 use starknet_types_core::felt::Felt;
 
+#[derive(Args, Debug, Clone)]
+#[group(required = true, multiple = false)]
+pub struct DeployIdentifier {
```
franciszekjob: A bit clearer:
```suggestion
pub struct Contractdentifier {
```

crates/sncast/src/starknet_commands/deploy.rs
```diff
@@ -17,12 +17,22 @@ use starknet::providers::jsonrpc::HttpTransport;
...
+}
+
 #[derive(Args)]
 #[command(about = "Deploy a contract on Starknet")]
 pub struct Deploy {
-    /// Class hash of contract to deploy
-    #[arg(short = 'g', long)]
-    pub class_hash: Felt,
+    #[command(flatten)]
+    pub identifier: DeployIdentifier,
```
franciszekjob:
```suggestion
    pub contract_identifier: ContractIdentifier,
```

crates/sncast/src/main.rs
franciszekjob, line 1: There's lots of new code and logic here, which makes our branches even more overgrown 😅 . We should either move it to `deploy` function, or have separate wrapper (e.g. `deploy_with_declare`), which later calls `deploy`.

crates/sncast/src/response/deploy.rs
```diff
@@ -1,62 +1,138 @@
...
+use serde_json::{Value, json};
 
-use super::{command::CommandResponse, explorer_link::OutputLink};
-use crate::response::cast_message::SncastMessage;
+#[derive(Clone, Serialize, Deserialize, CairoSerialize, Debug, PartialEq)]
+#[serde(untagged)]
+pub enum DeployResponseKind {
+    Deploy(DeployResponse),
+    DeployWithDeclare(DeployResponseWithDeclare),
+}
```
franciszekjob: This seems cleaner:

```rust
pub enum DeployResponse {
    Direct(DeployResponse),
    WithDeclare(DeployResponseWithDeclare),
}
```

or alternatively

```rust
pub enum DeployResult {
    Deploy(DeployResponse),
    DeployWithDeclare(DeployResponseWithDeclare),
}
```

crates/sncast/tests/e2e/deploy.rs
```diff
@@ -405,3 +410,251 @@ async fn test_json_output_format() {
...
+            "declare",
+            "--url",
+            URL,
+            "--contract-name",
+            "Map",
+        ];
+        let args = apply_test_resource_bounds_flags(args);
+
+        runner(&args).current_dir(tempdir.path()).assert().success();
+    }
```
franciszekjob, line 520: We don't need to put it inside block:
```suggestion
    // Declare the contract first
    let args = vec![
        "--accounts-file",
        "accounts.json",
        "--account",
        "my_account",
        "declare",
        "--url",
        URL,
        "--contract-name",
        "Map",
    ];
    let args = apply_test_resource_bounds_flags(args);

    runner(&args).current_dir(tempdir.path()).assert().success();
```

crates/sncast/tests/e2e/deploy.rs
```diff
@@ -405,3 +410,251 @@ async fn test_json_output_format() {
...
+        .env("SNCAST_FORCE_SHOW_EXPLORER_LINKS", "1")
+        .current_dir(tempdir.path());
+    let output = snapbox.assert().success();
+
+    assert_stderr_contains(
+        output,
+        indoc! {
+            "
+            Command: deploy
+            Error: Invalid transaction nonce
```
franciszekjob, line 683: This error should be improved. User should see something like 
```
Error occurred during contract declaration: ...
```
Currently it's unclear whether this error comes from declaration or deployment.

docs/src/starknet/deploy.md
```diff
@@ -82,6 +82,45 @@ transaction: https://sepolia.starkscan.co/tx/[..]
 > It is important to know how types are serialized because all values passed as constructor calldata are
 > interpreted as a field elements (felt252).
 
+### Deploying and Declaring at the Same Time
+
+You can deploy and declare a contract at the same time by passing `--contract-name` flag instead of `--class-hash`.
+Under the hood, if the passed contract was never declared to starknet, it will run the [declare](../starknet/declare.md)
+command first and then execute the contract deployment.
```
franciszekjob: nit: I think we should present this as deployment by contract name rather than deploying and declaring at the same time. The automatic declare is just a side effect, which allows for a simpler way to deploy without needing the class hash.

crates/sncast/src/response/deploy.rs
cptartur: I'll rename this somehow but it cannot be neither `DeployResponse`, because that conflicts with other `DeployResponse` nor `DeployResult` because these structs only are used to represent the messages output by functions, all of these are named `...Response` in other files.

crates/sncast/tests/e2e/deploy.rs
cptartur: I know but it's done on purpose, so that nothing used only for the initial declare is available in the test scope itself.

crates/sncast/tests/e2e/deploy.rs
cptartur: That's an error all of our functions throw already if the nonce is invalid. It wasn't introduced by these changes.

docs/src/starknet/deploy.md
cptartur: I don't think we should be hiding the fact that this makes 2 separate transactions. This is meant to be used as an convenience function, not magic to abstract how starknet works from users.

crates/sncast/src/main.rs
cptartur: The way I wanted this to be implemented is that deploying with declaring is composed only using top level APIs, but they remain two entirely separate functions.

We could do a separate refactor of main and move out things that don't have to be there to separate modules. But I don't think hiding top level logic for functions is beneficial.

cptartur: This logic is copied from declare and could technically be shared, but otherwise logic for declaring in deploy and normal declare are slightly different

```
            let manifest_path = assert_manifest_path_exists()?;
            let package_metadata = get_package_metadata(&manifest_path, &declare.package)?;
            let artifacts = build_and_load_artifacts(
                &package_metadata,
                &BuildConfig {
                    scarb_toml_path: manifest_path,
                    json: cli.json,
                    profile: cli.profile.unwrap_or("release".to_string()),
                },
                false,
                ui,
            )
```

crates/sncast/tests/e2e/deploy.rs
franciszekjob: Ah ok

docs/src/starknet/deploy.md
franciszekjob: > I don't think we should be hiding the fact that this makes 2 separate transactions.

Definitely,  that isn't my intention 😄 

What I'm saying is that it's more suitable to promote this command as deployment by contract name. Also, title alone "Deploying and Declaring at the Same Time" may even suggest that this happens simultaneously (I know it's described below).

How about this small change?

```
### Deploying By Contract Name

You can deploy contract by passing `--contract-name` flag instead of `--class-hash`.
Under the hood, if the passed contract was never declared to starknet, it will run the [declare](../starknet/declare.md)
command first and then execute the contract deployment.
```

cptartur: Sounds good, will change it

crates/sncast/src/main.rs
franciszekjob: Why can't we have `deploy_with_declare` and `deploy` wrappers, which will share some common logic (most likely extracted into another function)?

franciszekjob: Why can't we have `deploy_with_declare` and `deploy` wrappers, which will share some common logic (most likely extracted into another function)?

franciszekjob (commented):
Wdyt about mentioning this in sncast 101?

docs/src/starknet/deploy.md
```diff
@@ -82,6 +87,46 @@ transaction: https://sepolia.starkscan.co/tx/[..]
...
+Contract Address:         0x0[..]
+Class Hash:               0x0[..]
+Declare Transaction Hash: 0x0[..]
+Deploy Transaction Hash:  0x0[..]
+
+To see deployment details, visit:
+contract: [..]
+class: [..]
+declare transaction: [..]
+deploy transaction: [..]
```
franciszekjob, line 128: Maybe I'm a bit too picky, but after deployment details we have link to declare transaction. Maybe it can be moved above/separated?

crates/sncast/src/response/deploy.rs
```diff
@@ -1,38 +1,31 @@
...
+use serde_json::{Value, json};
 
 #[derive(Clone, Serialize, Deserialize, CairoSerialize, Debug, PartialEq)]
-pub struct DeployResponse {
-    pub contract_address: PaddedFelt,
-    pub transaction_hash: PaddedFelt,
+#[serde(untagged)]
+pub enum DeployResponse {
+    Deploy(StandardDeployResponse),
+    DeployWithDeclare(DeployResponseWithDeclare),
```
franciszekjob: nit:
```suggestion
    Standard(StandardDeployResponse),
    WithDeclare(DeployResponseWithDeclare),
```

crates/sncast/src/main.rs
```diff
@@ -404,28 +411,98 @@ async fn run_async_command(cli: Cli, config: CastConfig, ui: &UI) -> Result<()>
...
+                            "deploy",
+                            Err(err),
+                            ui,
+                            None,
+                        );
+                        return Ok(());
+                    }
+                }
+            } else {
+                unreachable!("One of class_hash or contract_name must be provided");
```
franciszekjob: nit:
```suggestion
                unreachable!("One of `--class-hash` or `--contract-name` must be provided");
```

crates/sncast/src/starknet_commands/deploy.rs
```diff
@@ -17,12 +17,23 @@ use starknet::providers::jsonrpc::HttpTransport;
...
+pub struct ContractIdentifier {
+    /// Class hash of contract to deploy
+    #[arg(short = 'g', long)]
+    pub class_hash: Option<Felt>,
+
+    /// Contract name
+    #[arg(long, conflicts_with = "class_hash")]
+    pub contract_name: Option<String>,
+}
+
```
franciszekjob, line 31: nit: Usage of `id` and `group` seems suitable here:
```suggestion
#[derive(Args, Debug, Clone)]
#[group(id = "contract_identifier", required = true, multiple = false)]
pub struct ContractIdentifier {
    /// Class hash of contract to deploy
    #[arg(short = 'g', long, group = "contract_identifier")]
    pub class_hash: Option<Felt>,

    /// Contract name
    #[arg(long, group = "contract_identifier")]
    pub contract_name: Option<String>,
}

```

franciszekjob (approved):

crates/sncast/src/main.rs
```diff
@@ -404,28 +411,98 @@ async fn run_async_command(cli: Cli, config: CastConfig, ui: &UI) -> Result<()>
             let account =
                 get_account(&config, &provider, &rpc, config.keystore.as_ref(), ui).await?;
 
+            let (class_hash, declare_response) = if let Some(class_hash) = identifier.class_hash {
+                (class_hash, None)
+            } else if let Some(contract_name) = identifier.contract_name {
+                let manifest_path = assert_manifest_path_exists()?;
+                let package_metadata = get_package_metadata(&manifest_path, &None)?;
```
ddoktorski: Does this feature work for workspaces?

crates/sncast/src/main.rs
```diff
@@ -404,28 +411,98 @@ async fn run_async_command(cli: Cli, config: CastConfig, ui: &UI) -> Result<()>
...
+                let declare_result = declare(
+                    contract_name,
+                    fee_args.clone(),
+                    nonce,
+                    &account,
+                    &artifacts,
+                    WaitForTx {
+                        wait: true,
+                        wait_params: wait_config.wait_params,
+                        show_ui_outputs: true,
```
ddoktorski: What if someone passes the `--wait` flag?

crates/sncast/tests/e2e/deploy.rs
```diff
@@ -405,3 +410,251 @@ async fn test_json_output_format() {
...
+            class: [..]
+            declare transaction: [..]
+            deploy transaction: [..]
+            "
+        },
+    );
+}
+
+#[tokio::test]
+async fn test_deploy_with_declare_nonexistent_contract() {
```
ddoktorski: What's the rationale behind this test given that its name refers to `nonexistent_contract`, but the asserted error is `Invalid transaction nonce`?

docs/src/starknet/deploy.md
```diff
@@ -82,6 +87,46 @@ transaction: https://sepolia.starkscan.co/tx/[..]
...
+contract from a Scarb project by providing the `--contract-name` flag.
+Under the hood, if the passed contract was never declared to starknet, it will run the [declare](../starknet/declare.md)
+command first and then execute the contract deployment.
+
+> 📝 **Note**
+> When passing `--contract-name` flag, `sncast` must wait for the declare transaction to be completed first.
+> The contract might wait for a few seconds before executing the deployment.
+
+> 📝 **Note**
+> If fee arguments are provided to the method, they will be shared between declare and deploy transactions.
```
ddoktorski: It might suggest that the provided fee will will try to cover both transactions, which is not the case. I think it should be rephrased to clarify that the same fee arguments are used separately for each transaction

crates/sncast/src/main.rs
cptartur: These errors don't really matter since they will never appear in runtime (unless we messed something up seriously)

crates/sncast/src/starknet_commands/deploy.rs
cptartur: I'll remove the `conflicts_with` because it's unnecessary since we have `multiple = false`. What's the benefit of using group id here?

crates/sncast/src/main.rs
cptartur: The logic was copied from `declare`, so if it doesn't then declare doesn't work for them as well. I'll check

crates/sncast/src/main.rs
cptartur: We'll wait for the deploy tx that is done later as well. For declare tx we always need to wait, because otherwise the subsequent deploy tx could fail due to contract not being declared yet.

crates/sncast/tests/e2e/deploy.rs
cptartur: Error on my side, changing

crates/sncast/src/main.rs
ddoktorski: Yes, but my point is that when someone passes the `--wait` flag, they would likely expect to see the outputs, but in this case that won't happen

crates/sncast/src/main.rs
ddoktorski: `declare` command has a flag for `--package`

cptartur: Actually, disregard. This doesn't work for workspaces. I could introduce another `--package` argument that is mutually exclusive with using `--class-hash`, though I wonder if it's not cluttering the interface too much.

crates/sncast/src/main.rs
cptartur: Ah, right. I can add that logic

crates/sncast/src/main.rs
```diff
@@ -404,28 +412,99 @@ async fn run_async_command(cli: Cli, config: CastConfig, ui: &UI) -> Result<()>
...
+                            Err(err),
+                            ui,
+                            None,
+                        );
+                        return Ok(());
+                    }
+                }
+            } else {
+                unreachable!("Either `--class_hash` or `--contract_name` must be provided");
+            };
```
ddoktorski, line 473: Can we move this logic to a separate function?

crates/sncast/src/main.rs
```diff
@@ -404,28 +412,99 @@ async fn run_async_command(cli: Cli, config: CastConfig, ui: &UI) -> Result<()>
...
+            let result = if let Some(declare_response) = declare_response {
+                result.map(|r| {
+                    DeployResponse::WithDeclare(DeployResponseWithDeclare::from_responses(
+                        &r,
+                        &declare_response,
+                    ))
+                })
+            } else {
+                result.map(DeployResponse::Standard)
+            };
```
ddoktorski, line 506:
```suggestion
            let result = result.map(|r| match declare_response {
                Some(declare_response) => DeployResponse::WithDeclare(
                    DeployResponseWithDeclare::from_responses(&r, &declare_response),
                ),
                None => DeployResponse::Standard(r),
            });
```

crates/sncast/src/starknet_commands/deploy.rs
```diff
@@ -44,6 +55,10 @@ pub struct Deploy {
 
     #[command(flatten)]
     pub rpc: RpcArgs,
+
+    /// Specifies scarb package to be used. Only possible to use with --class-hash.
```
ddoktorski: This should be the opposite

ddoktorski (commented):
Changelog should be also updated

crates/sncast/src/main.rs
cptartur: Functionally it's the same, I don't really like matching on `Option`, `Result`, `bool`, etc.

crates/sncast/src/main.rs
cptartur: I've actually discussed this with @franciszekjob, because he wanted me to moved it as well. I think just making it a function and moving it will not make maintenance of sncast any easier – the code will not be shared anyway and we're just "hiding" the bad logic.

What I want to do eventually is refactor the flow of handling in cast that should actually address the very long main. But until then I think we would just be hiding the fact that it desperately needs refactoring.

crates/sncast/src/main.rs
cptartur: Added `--package`

crates/sncast/src/starknet_commands/deploy.rs
```diff
@@ -44,6 +55,10 @@ pub struct Deploy {
 
     #[command(flatten)]
     pub rpc: RpcArgs,
+
+    /// Specifies scarb package to be used. Only possible to use with --contract-name.
```
ddoktorski: nit:
```suggestion
    /// Specifies scarb package to be used. Only possible to use with `--contract-name`.
```

crates/sncast/tests/e2e/deploy.rs
```diff
@@ -270,6 +275,33 @@ fn test_wrong_calldata() {
...
+        "my_package",
+    ];
+
+    let snapbox = runner(&args);
+    let output = snapbox.assert().failure();
+
+    assert_stderr_contains(
+        output,
+        indoc! {r"
+        error: the argument '--class-hash <CLASS_HASH>' cannot be used with '--package <PACKAGE>'
```
ddoktorski, line 300: This mostly tests clap logic. There were some discussions in the past with conclusion to rather avoid adding such tests

crates/sncast/src/main.rs
ddoktorski: The fact that the code isn’t in the best shape right now doesn’t mean it has to get worse. Making this a separate function obviously won’t fix everything, but it would be slightly better than extending an already very long function. I'll leave this decision up to you

crates/sncast/src/main.rs
ddoktorski: I think this is more concise and idiomatic, but this is a nitpick so again your call

docs/src/starknet/deploy.md
```diff
@@ -4,7 +4,12 @@
 
 Starknet Foundry `sncast` supports deploying smart contracts to a given network with the `sncast deploy` command.
 
-It works by invoking a [Universal Deployer Contract](https://docs.openzeppelin.com/contracts-cairo/0.19.0/udc), which deploys the contract with the given class hash and constructor arguments.
+It works by invoking a [Universal Deployer Contract](https://docs.openzeppelin.com/contracts-cairo/0.19.0/udc), which
```
ddoktorski: The link is outdated
```suggestion
It works by invoking a [Universal Deployer Contract](https://docs.openzeppelin.com/contracts-cairo/2.x/udc), which
```

crates/sncast/tests/e2e/deploy.rs
cptartur: I treat it mostly as sanity check, so someone doesn't remove the `conflicts_with` by accident. The errors on clap side are quite stable so it's not much maintenance effort.

crates/sncast/src/main.rs
cptartur: My problem with just moving this to another helper is that we are ending with a big chain of helpers calling each other without any benefits for maintaining them other than files having less lines of code (and they aren't super long at the moment). 
It's not about the effort required, I can move the logic into another function right away but I think it's just hiding the problems we already have. I'd rather keep this and try refactoring the code properly soon.

ddoktorski (approved):

PR #3786, opened 2025-10-03
Support devnet in `--network` flag

### LINKED ISSUE #3764: Support `devnet` in `--network` flag
### Objective

Support `devnet` in `--network` flag

### Additional Context

It should use some heuristic to detect common devnet addresses or even a fixed address

MKowalski8:
Closes #3764 

## Introduced changes

- Implement `devnet` detection for `--network` flag

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/sncast/src/helpers/devnet_detection.rs
```diff
@@ -0,0 +1,312 @@
+use std::process::Command;
+
+#[derive(Debug, Clone)]
+struct DevnetInfo {
```
franciszekjob:
```suggestion
struct DevnetProcessInfo {
```

crates/sncast/src/helpers/devnet_detection.rs
```diff
@@ -0,0 +1,312 @@
...
+    host: String,
+    port: u16,
+}
+
+#[derive(Debug)]
+enum FindDevnetError {
+    None,
+    Multiple,
+    CommandFailed,
+}
```
franciszekjob:
```suggestion
enum DevnetDetectionError {
    NoInstance,
    MultipleInstances,
    CommandFailed,
}
```

crates/sncast/src/helpers/devnet_detection.rs
```diff
@@ -0,0 +1,312 @@
...
+#[must_use]
+pub fn is_devnet_running() -> bool {
+    detect_devnet_from_processes().is_ok()
+}
+
+fn detect_devnet_from_processes() -> Result<String, String> {
+    match find_devnet_process_info() {
+        Ok(info) => Ok(format!("http://{}:{}", info.host, info.port)),
+        Err(FindDevnetError::Multiple) => {
+            Err("Multiple starknet-devnet instances found. Please use --url to specify which one to use.".to_string())
```
franciszekjob: nit:
```suggestion
            Err("Multiple starknet-devnet instances found. Please use `--url <URL>` to specify which one to use.".to_string())
```

crates/sncast/src/helpers/devnet_detection.rs
```diff
@@ -0,0 +1,312 @@
...
+        }
+        Err(FindDevnetError::None | FindDevnetError::CommandFailed) => {
+            // Fallback to default starknet-devnet URL if reachable
+            if is_port_reachable("127.0.0.1", 5050) {
+                Ok("http://127.0.0.1:5050".to_string())
+            } else {
+                Err(
+                    "Could not detect running starknet-devnet instance. Please use --url instead."
```
franciszekjob: nit:
```suggestion
                    "Could not detect running starknet-devnet instance. Please use `--url<URL>` instead or start devnet."
```

crates/sncast/src/helpers/devnet_detection.rs
```diff
@@ -0,0 +1,312 @@
...
+        }
+    }
+    None
+}
+
+fn extract_port_from_flag(cmdline: &str, flag: &str) -> Option<u16> {
+    if let Some(port_str) = extract_string_from_flag(cmdline, flag)
+        && let Ok(p) = port_str.parse::<u16>()
+        && p > 1024
+        && p < 65535
```
franciszekjob: 1024 and 65535 are also included in ephemeral ports:
```suggestion
        && p >= 1024
        && p <= 65535
```

franciszekjob: nit: This could be moved into helper function `is_port_ephermal` (because it's used in another placed as well).

crates/sncast/src/helpers/devnet_detection.rs
```diff
@@ -0,0 +1,312 @@
...
+
+fn detect_devnet_from_processes() -> Result<String, String> {
+    match find_devnet_process_info() {
+        Ok(info) => Ok(format!("http://{}:{}", info.host, info.port)),
+        Err(FindDevnetError::Multiple) => {
+            Err("Multiple starknet-devnet instances found. Please use --url to specify which one to use.".to_string())
+        }
+        Err(FindDevnetError::None | FindDevnetError::CommandFailed) => {
+            // Fallback to default starknet-devnet URL if reachable
+            if is_port_reachable("127.0.0.1", 5050) {
```
franciszekjob: Maybe let's move this host and port to consts (e.g. `DEFAULT_DEVNET_HOST` and `DEFAULT_DEVNET_PORT`) as they are used in other places too?

crates/sncast/src/helpers/devnet_detection.rs
```diff
@@ -0,0 +1,312 @@
...
+    DevnetInfo {
+        host: final_host,
+        port: final_port,
+    }
+}
+
+fn is_port_reachable(host: &str, port: u16) -> bool {
+    let url = format!("http://{host}:{port}/is_alive");
+
+    std::process::Command::new("curl")
```
franciszekjob: nit:
```suggestion
    Command::new("curl")
```

crates/sncast/src/helpers/devnet_detection.rs
```diff
@@ -0,0 +1,312 @@
...
+fn find_devnet_process_info() -> Result<DevnetInfo, FindDevnetError> {
+    let output = Command::new("ps")
+        .args(["aux"])
+        .output()
+        .map_err(|_| FindDevnetError::CommandFailed)?;
+    let ps_output = String::from_utf8_lossy(&output.stdout);
+
+    let devnet_processes: Vec<DevnetInfo> = ps_output
+        .lines()
+        .filter(|line| line.contains("starknet-devnet"))
```
franciszekjob: We could use `grep` or `awk` to filter lines containing "starknet-devnet", but in practice this won’t make a noticeable difference - a typical system has only about 500–1000 processes (checked on my machine with `ps aux | wc -l` 😅, was 532 ), so the ps aux output is small.
If the number of processes were significantly higher (e.g., in the millions), then offloading filtering to awk could provide a measurable benefit.

crates/sncast/src/starknet_commands/verify/voyager.rs
```diff
@@ -392,6 +392,7 @@ impl<'a> VerificationInterface<'a> for Voyager<'a> {
             Err(_) => match self.network {
                 Network::Mainnet => "https://api.voyager.online/beta".to_string(),
                 Network::Sepolia => "https://sepolia-api.voyager.online/beta".to_string(),
+                Network::Devnet => String::new(),
```
franciszekjob: I don't think we should return empty string in such case. Let's raise an error informing that verification can't be used with devnet.

crates/sncast/src/starknet_commands/verify/walnut.rs
```diff
@@ -114,6 +114,7 @@ impl VerificationInterface<'_> for WalnutVerificationInterface {
         let path = match self.network {
             Network::Mainnet => "/v1/sn_main/verify",
             Network::Sepolia => "/v1/sn_sepolia/verify",
+            Network::Devnet => "",
```
franciszekjob: Same as above.

crates/sncast/src/lib.rs
```diff
@@ -89,17 +89,22 @@ pub const MAINNET: Felt =
 pub const SEPOLIA: Felt =
     Felt::from_hex_unchecked(const_hex::const_encode::<10, true>(b"SN_SEPOLIA").as_str());
 
+pub const DEVNET: Felt =
+    Felt::from_hex_unchecked(const_hex::const_encode::<6, true>(b"SN_DEV").as_str());
```
franciszekjob: It looks like devnet uses `SN_SEPOLIA` as its chain id, have a look:
https://github.com/0xSpaceShard/starknet-devnet/blob/b1f0252829c9e61815bf642c792e96a501027f21/website/docs/api.md?plain=1#L65

franciszekjob (changes requested):
Once [`3603-3-devnet-interactions-docs`](https://github.com/foundry-rs/starknet-foundry/tree/3603-3-devnet-interactions-docs) is merged, let's mention ability to use `--network devnet` in the docs 👍 .

crates/sncast/src/lib.rs
MKowalski8: Yeah, you are right. I misunderstood this.

crates/sncast/src/starknet_commands/verify/voyager.rs
MKowalski8: I agree, that will be better. Firstly I thought that it can never be executed, but I found a case for that. I'll take care of this

crates/sncast/src/helpers/devnet_detection.rs
MKowalski8: I was considering it. I abandoned the idea, because this solution was "cleaner". However if you think that would be better I am into it.

crates/sncast/src/helpers/devnet_detection.rs
MKowalski8: Actually, the upper bond was not necessary, due to `u16` range

docs/src/starknet/sncast-overview.md
```diff
@@ -56,8 +56,16 @@ Response: [0x0, 0x0, 0x43686172697a617264, 0x9, 0x0, 0x0, 0x41a78e741e5af2fec34b
...
+The `--network` flag supports the following networks:
+
+- **mainnet** - Connects to Starknet mainnet using a free RPC provider
+- **sepolia** - Connects to Starknet Sepolia testnet using a free RPC provider
+- **devnet** - Attempts to auto-detect running starknet-devnet instances
+
+When using **mainnet** or **sepolia**, `sncast` will randomly select one of the free RPC providers.
+When using free providers you may experience rate limits and other unexpected behavior.
+
+For **devnet**, `sncast` will try to detect running `starknet-devnet` instance and connect to it, but it may fail. If detection fails, then use the `--url` flag instead.
```
MKowalski8: > Once [3603-3-devnet-interactions-docs](https://github.com/foundry-rs/starknet-foundry/tree/3603-3-devnet-interactions-docs) is merged, let's mention ability to use --network devnet in the docs 👍 .

Since we have [interaction-docs](https://github.com/foundry-rs/starknet-foundry/blob/master/docs/src/starknet/integration_with_devnet.md), after adding info about `--network devnet` there. Should I remove this changes in `overview`? I am aware that it is not the best place and we may want to preserve only short messages here

crates/sncast/src/helpers/devnet_detection.rs
```diff
@@ -0,0 +1,308 @@
...
+}
+
+#[must_use]
+pub fn is_devnet_running() -> bool {
+    detect_devnet_from_processes().is_ok()
+}
+
+fn detect_devnet_from_processes() -> Result<String, String> {
+    match find_devnet_process_info() {
+        Ok(info) => Ok(format!("http://{}:{}", info.host, info.port)),
```
cptartur: You could probably do this

```suggestion
        Ok(DevnetProcessInfo { host, port }) => Ok(format!("http://{}:{}", host, port)),
```

docs/src/starknet/sncast-overview.md
franciszekjob: It's fine, leave it there too.

crates/sncast/src/helpers/devnet/detection.rs
```diff
@@ -0,0 +1,308 @@
...
+
+    if port.is_none() {
+        port = std::env::var("PORT")
+            .ok()
+            .and_then(|port_env| parse_valid_port(&port_env));
+    }
+
+    if host.is_none()
+        && let Ok(host_env) = std::env::var("HOST")
+        && !host_env.is_empty()
```
cptartur: We validate port but take host directly, to be consistent we should probably either validate both or none.

What was the motivation for only validating one of them?

crates/sncast/src/helpers/devnet_detection.rs
```diff
@@ -0,0 +1,308 @@
...
+            .unwrap_or("")
+            .trim_start_matches('=')
+            .trim_start_matches(':');
+
+        if !value_str.is_empty() {
+            return Some(value_str.to_string());
+        }
+    }
+    None
+}
```
franciszekjob: Wouldn't regex be better here?

crates/sncast/src/helpers/devnet_detection.rs
```diff
@@ -0,0 +1,308 @@
...
+
+fn extract_devnet_info_from_cmdline(cmdline: &str) -> DevnetProcessInfo {
+    let mut port = extract_port_from_flag(cmdline, "--port");
+    let mut host = extract_string_from_flag(cmdline, "--host");
+
+    if port.is_none() {
+        port = std::env::var("PORT")
+            .ok()
+            .and_then(|port_env| parse_valid_port(&port_env));
```
cptartur: This way we silently ignore user settings if they are invalid. If we aren't doing any nice error handling for invalid values I'd rather we panic than silently ignore them.

crates/sncast/src/helpers/devnet_detection.rs
```diff
@@ -0,0 +1,308 @@
...
+}
+
+fn is_port_reachable(host: &str, port: u16) -> bool {
+    let url = format!("http://{host}:{port}/is_alive");
+
+    Command::new("curl")
+        .args(["-s", "-f", "--max-time", "1", &url])
+        .output()
+        .map(|output| output.status.success())
+        .unwrap_or(false)
```
cptartur: Rather than rely on curl being installed, couldn't we use some library for HTTP requests?

cptartur: We already have a dependency on `reqwest`

crates/sncast/src/helpers/devnet_detection.rs
cptartur: This goes for all places we do so, not only here

crates/sncast/src/helpers/devnet_detection.rs
```diff
@@ -0,0 +1,308 @@
...
+    }
+}
+
+fn extract_string_from_flag(cmdline: &str, flag: &str) -> Option<String> {
+    if let Some(pos) = cmdline.find(flag) {
+        let after_pattern = &cmdline[pos + flag.len()..];
+        let value_str = after_pattern
+            .split_whitespace()
+            .next()
+            .unwrap_or("")
```
cptartur: If the iterator returns nothing (we unwrap with default), doesn't that mean we won't find anything anyway?

Also a nit, but probably `unwrap_or_default` could do here.

crates/sncast/src/helpers/devnet_detection.rs
```diff
@@ -0,0 +1,308 @@
...
+            .trim_start_matches('=')
+            .trim_start_matches(':');
```
cptartur: What's the purpose if these?

crates/sncast/src/helpers/devnet/detection.rs
```diff
@@ -0,0 +1,308 @@
...
+        .collect();
+
+    match devnet_processes.as_slice() {
+        [] => Err(DevnetDetectionError::NoInstance),
+        [single] => Ok(single.clone()),
+        _ => Err(DevnetDetectionError::MultipleInstances),
+    }
+}
+
+fn extract_string_from_flag(cmdline: &str, flag: &str) -> Option<String> {
```
cptartur: Some helpers could go into a submodule within `devnet_detection`, in a separate file.

Generally this file could use being split into submodules.

crates/sncast/src/helpers/rpc.rs
```diff
@@ -15,7 +15,8 @@ pub struct RpcArgs {
     #[arg(short, long)]
     pub url: Option<String>,
 
-    /// Use predefined network with a public provider. Note that this option may result in rate limits or other unexpected behavior
+    /// Use predefined network with a public provider. Note that this option may result in rate limits or other unexpected behavior.
+    /// For devnet, attempts to auto-detect running starknet-devnet instance. If auto-detection fails, use --url instead.
```
cptartur:
```suggestion
    /// For devnet, attempts to auto-detect running starknet-devnet instance.
```

The error will show that to the user anyways, no?

crates/sncast/src/helpers/rpc.rs
```diff
@@ -44,29 +43,17 @@ impl RpcArgs {
...
-                }
-            })
+    pub fn get_url(&self, config_url: &str) -> Result<String> {
+        match (&self.network, &self.url, config_url.is_empty()) {
+            (Some(network), None, _) => {
+                let free_provider = FreeProvider::semi_random();
+                network.url(&free_provider)
+            }
+            (None, Some(url), _) => Ok(url.clone()),
+            (None, None, false) => Ok(config_url.to_string()),
```
cptartur: This is probably slightly safer

```suggestion
        match (&self.network, &self.url, config_url)) {
            (Some(network), None, _) => {
                let free_provider = FreeProvider::semi_random();
                network.url(&free_provider)
            }
            (None, Some(url), _) => Ok(url.clone()),
            (None, None, config_url) if !config_url.is_empty() => Ok(config_url.to_string()),
```

crates/sncast/src/response/explorer_link.rs
```diff
@@ -48,29 +49,30 @@ pub enum ExplorerError {
...
 pub fn block_explorer_link_if_allowed<T>(
     result: &anyhow::Result<T>,
     chain_id: Felt,
-    rpc: &RpcArgs,
     config: &CastConfig,
 ) -> Option<ExplorerLinksMessage>
 where
     T: OutputLink + Clone,
 {
-    if (!config.show_explorer_links || rpc.is_localhost(&config.url))
```
cptartur, line 62: Why do we no longer check for localhost?

docs/src/starknet/sncast-overview.md
```diff
@@ -56,8 +56,16 @@ Response: [0x0, 0x0, 0x43686172697a617264, 0x9, 0x0, 0x0, 0x41a78e741e5af2fec34b
 
 ### Network and RPC Providers
 
-When providing `--network` flag, `sncast` will randomly select on of the free RPC providers.
-When using free provider you may experience rate limits and other unexpected behavior.
+The `--network` flag supports the following networks:
+
+- **mainnet** - Connects to Starknet mainnet using a free RPC provider
+- **sepolia** - Connects to Starknet Sepolia testnet using a free RPC provider
+- **devnet** - Attempts to auto-detect running starknet-devnet instances
```
cptartur: We fail on multiple 

```suggestion 
- **devnet** - Attempts to auto-detect running starknet-devnet instance
```

docs/src/starknet/sncast-overview.md
```diff
@@ -56,8 +56,16 @@ Response: [0x0, 0x0, 0x43686172697a617264, 0x9, 0x0, 0x0, 0x41a78e741e5af2fec34b
...
+
+When using **mainnet** or **sepolia**, `sncast` will randomly select one of the free RPC providers.
+When using free providers you may experience rate limits and other unexpected behavior.
```
cptartur: This could be `> ` note

docs/src/starknet/sncast-overview.md
cptartur:
```suggestion
For **devnet**, `sncast` will try to detect running `starknet-devnet` instance and connect to it, but detection is not guaranteed. If `sncast` is unable to detect your devnet instance, provide the `--url` directly.
```

crates/sncast/src/helpers/devnet/detection.rs
```diff
@@ -0,0 +1,308 @@
...
+        .unwrap_or(false)
+}
+
+fn parse_valid_port(port_str: &str) -> Option<u16> {
+    // Ports below 1024 typically require elevated permissions
+    port_str.parse::<u16>().ok().filter(|&p| p >= 1024)
+}
+
+#[cfg(test)]
+mod tests {
```
cptartur, line 103: Didn't review the tests yet

crates/sncast/src/helpers/devnet_detection.rs
```diff
@@ -0,0 +1,308 @@
...
+            } else {
+                extract_devnet_info_from_cmdline(line)
+            }
+        })
+        .collect();
+
+    match devnet_processes.as_slice() {
+        [] => Err(DevnetDetectionError::NoInstance),
+        [single] => Ok(single.clone()),
+        _ => Err(DevnetDetectionError::MultipleInstances),
```
franciszekjob: `Ok` at the top:
```suggestion
        [single] => Ok(single.clone()),
        [] => Err(DevnetDetectionError::NoInstance),
        _ => Err(DevnetDetectionError::MultipleInstances),
```

crates/sncast/src/helpers/devnet/detection.rs
franciszekjob, line 1: We can make a separate module `devnet`, place this file and the one containing `DevnetProvider`.

crates/sncast/src/helpers/devnet/detection.rs
```diff
@@ -0,0 +1,308 @@
...
+            } else if parts.len() == 2
+                && let Ok(host_port) = parts[0].parse::<u16>()
+            {
+                return Some((DEFAULT_DEVNET_HOST.to_string(), host_port));
+            }
+        }
+    }
+
+    None
+}
```
franciszekjob: Same remark about regex usage

crates/sncast/src/helpers/devnet/detection.rs
MKowalski8: The idea was to check if the port is in the right range (basically if it is not below 1024, e.g 443, 80, 22). However we can assume that if the process is running it was run with correct arguments, so we can as well do not check and just take it as it is.

crates/sncast/src/helpers/devnet_detection.rs
MKowalski8: I didn't think it was necessary, but I can add more messages if something goes wrong

crates/sncast/src/helpers/devnet_detection.rs
MKowalski8: One problem was that, afaik I need to change this func's to async, and because of that some other func's as well (I mean whole `Network`, then `RpcArgs` with every usage of `get_url`, and so on). Curl was just cleaner option, considering that this is only for fallback and double check, as well it might not be there.

crates/sncast/src/helpers/devnet_detection.rs
MKowalski8: The `=` is for cases like --port=5050, but I see that trim `:` is a leftover and is not necessary

crates/sncast/src/response/explorer_link.rs
MKowalski8: I changed it, because there was an issue with checking for localhost that we might ignore some local nodes of other networks. So if the new heuristic is better in detecting devnet and there was an issue with the previous one, I deleted this check

crates/sncast/src/helpers/devnet_detection.rs
MKowalski8: Since from `ps aux` we get a formated output, this seemed to me easier to understand. But I will try to change it to regex.

crates/sncast/src/helpers/devnet_detection.rs
MKowalski8: I found one case where I need to use it in default checking. So if it worth to change some logic to async, I can change it.

cptartur: Most of cast functions are async anyway so it's fine by me

crates/sncast/src/helpers/devnet/detection.rs
cptartur: I wouldn't do any more validations that just checking if we can connect to the address we got. If not, then fail. We could also probably say what address was tested if you think that's beneficial to the users.

MKowalski8: I tested it that way, but for me it may be confusing for the user. I would limit it just to info, that we detected running instance, but we couldn't connect

crates/sncast/src/helpers/devnet_detection.rs
cptartur: I'd rather we panic without any message at all than silently ignore errors. What I mean is that we should be failing for these errors somehow rather than handle them like this.

crates/sncast/src/helpers/devnet/detection.rs
cptartur: That's fine as well, I'd just not do a specific handling of these ports etc.. Only say that "we tried to connect to X but we failed".

crates/sncast/src/helpers/devnet/detection.rs
MKowalski8: Do you mean that some function could be extracted and reused in the future? I can see only two. `extract_string_from_flag` and `extract_port_from_flag`. If I unserstand you correctly I can move it in other file in `helpers` or for example into [command.rs](https://github.com/foundry-rs/starknet-foundry/blob/master/crates/sncast/src/helpers/command.rs) (if you agree it is a good place).

crates/sncast/src/helpers/devnet/detection.rs
MKowalski8: With test's I am not quite sure, where they should be. I could move them into a test folder, but then I have to make some funcs public, that imo don't have to be public, also I saw that it is a common practice in helpers to put test inside, so I left them there

crates/sncast/src/helpers/devnet/provider.rs
```diff
@@ -84,7 +84,10 @@ impl DevnetProvider {
             .client
             .get(format!(
                 "{}/is_alive",
-                self.url.to_string().replace("/rpc", "")
+                self.url
+                    .to_string()
+                    .trim_end_matches('/')
+                    .replace("/rpc", "")
```
MKowalski8, line 90: Note: The change was made here, because after parsing `http://127.0.0.1:5050` (not `http://127.0.0.1:5050/rpc`), the `/` was added in the end, so it ended up with `http:/127.0.0.1:5050//is_alive`

crates/sncast/src/helpers/devnet/detection.rs
```diff
@@ -0,0 +1,306 @@
+use regex::Regex;
+use std::process::Command;
+
+use crate::helpers::devnet::provider::DevnetProvider;
+
+const DEFAULT_DEVNET_HOST: &str = "127.0.0.1";
+const DEFAULT_DEVNET_PORT: u16 = 5050;
+
+#[derive(Debug, Clone)]
+struct DevnetProcessInfo {
```
franciszekjob: nit: This can be any process info:

```suggestion
struct ProcessInfo {
```

crates/sncast/src/helpers/devnet/detection.rs
```diff
@@ -0,0 +1,306 @@
...
+        assert_eq!(info3.host, "127.0.0.1");
+    }
+
+    fn test_extract_devnet_info_from_docker_line() {
+        let cmdline1 = "docker run -p 127.0.0.1:5055:5050 shardlabs/starknet-devnet-rs";
+        let info1 = extract_devnet_info_from_docker_line(cmdline1).unwrap();
+        assert_eq!(info1.port, 5055);
+        assert_eq!(info1.host, "127.0.0.1");
+
+        let cmdline2 = "docker run --publish     8080:5050 shardlabs/starknet-devnet-rs";
```
franciszekjob: nit:
```suggestion
        let cmdline2 = "docker run --publish 8080:5050 shardlabs/starknet-devnet-rs";
```

crates/sncast/src/helpers/devnet/detection.rs
```diff
@@ -0,0 +1,306 @@
...
+#[derive(Debug, Clone)]
+struct DevnetProcessInfo {
+    host: String,
+    port: u16,
+}
+
+#[derive(Debug, thiserror::Error)]
+enum DevnetDetectionError {
+    #[error(
+        "Could not detect running starknet-devnet instance. Please use `--url <URL>` instead or start devnet if it is not running."
```
franciszekjob: nit: Assuming devnet was not detected, it's most likely not running:
```suggestion
        "Could not detect running starknet-devnet instance. Please use `--url <URL>` instead or start devnet."
```

franciszekjob (approved):

crates/sncast/src/helpers/devnet/detection.rs
MKowalski8: It is intentional, to check if something goes wrong with withespaces

franciszekjob: Right, resolving

crates/sncast/src/helpers/devnet/detection.rs
cptartur: I mean some of more basic helpers could be moved to a separate file within a new `detection` module. They could probably still be private.

crates/sncast/src/helpers/devnet/detection.rs
```diff
@@ -0,0 +1,306 @@
...
+    MultipleInstances,
+    #[error("Failed to execute process detection command.")]
+    CommandFailed,
+    #[error(
+        "Found starknet-devnet process, but could not reach it. Please use `--url <URL>` to specify the correct URL."
+    )]
+    ProcessNotReachable,
+}
+
+pub async fn detect_devnet_url() -> Result<String, String> {
```
cptartur: Couldn't these methods just return `Result<String, DevnetDetectionError>`? And make it the callers responsibility to handle converting errors to string etc?

crates/sncast/src/helpers/devnet/detection.rs
```diff
@@ -0,0 +1,306 @@
...
+        }
+        Err(DevnetDetectionError::NoInstance | DevnetDetectionError::CommandFailed) => {
+            // Fallback to default starknet-devnet URL if reachable
+            if is_devnet_url_reachable(DEFAULT_DEVNET_HOST, DEFAULT_DEVNET_PORT).await {
+                Ok(format!(
+                    "http://{DEFAULT_DEVNET_HOST}:{DEFAULT_DEVNET_PORT}"
+                ))
+            } else {
+                Err(DevnetDetectionError::NoInstance.to_string())
+            }
```
cptartur, line 62: Shouldn't we be checking if the default devnet instance is available first, before we do any other checking? I.e. before `match find_devnet_process_info`. I'd assume that in most cases this will be the devnet instance used anyways.

crates/sncast/src/helpers/devnet/detection.rs
```diff
@@ -0,0 +1,304 @@
...
+    let mut port = extract_port_from_flag(cmdline, "--port");
+    let mut host = extract_string_from_flag(cmdline, "--host");
+
+    if port.is_none()
+        && let Ok(port_env) = std::env::var("PORT")
+    {
+        port = Some(
+            port_env
+                .parse()
+                .map_err(|_| DevnetDetectionError::ProcessNotReachable)?,
```
cptartur: One problem with using this general error here is that it will be hard to debug it. We'll just get a report that sncast failed to connect but without any details on what failed exactly.

It maybe would be helpful to have an inner error (just as a string maybe) inside `ProcessNotReachable` where we could write some more details like `failed to parse port {}` etc.

cptartur: On the other hand this is creating a risk of writing some details from `ps aux` that might be sensitive. So maybe let's just stick to what you have.

crates/sncast/src/helpers/devnet/detection.rs
cptartur: These are fine to keep here. Let's add some tests for error cases however.

crates/sncast/tests/helpers/devnet_detection.rs
```diff
@@ -0,0 +1,10 @@
+use sncast::helpers::devnet::detection::detect_devnet_url;
+
+#[tokio::test]
+async fn test_detect_devnet_url() {
+    let result = detect_devnet_url()
+        .await
+        .expect("Failed to detect devnet URL");
+
+    assert_eq!(result, "http://127.0.0.1:5055");
```
cptartur: You could maybe check it against url from `URL` const.

cptartur (commented):
Looks good, left some additional comments

crates/sncast/src/helpers/devnet/detection.rs
MKowalski8: We can, but for example we wouldn't detect if someone uses multiple instances then, because we will always send to default.

MKowalski8: But yeah, in most cases it will be default. However it doesn't take long to process, so I don't think if this is something that user can notice

crates/sncast/src/helpers/devnet/detection.rs
MKowalski8: I wanted to add info about the port first, but then there was a case that if I do it for port, I also should do it for host, and in the end, does user really needs it? If something goes wrong he just use `--url` as recommended in this type of situation. 
Another thing is that, only case when it should happen is when someone set `PORT = "abcd"` in one terminal and then open another terminal and run `starknet-devnet` with default port. Is this helpful for him that we couldn't parse `abcd`?

crates/sncast/src/helpers/devnet/detection.rs
cptartur: > We can, but for example we wouldn't detect if someone uses multiple instances then, because we will always send to default.

Ah right, let's keep the logic you have then.

crates/sncast/src/helpers/devnet/detection.rs
cptartur: Probably not, let's keep what we have

crates/sncast/src/helpers/devnet/detection.rs
MKowalski8: With the current implementation it can. I'll try to change it

crates/sncast/src/helpers/devnet/detection.rs
cptartur: Let's move `extract_string_from_flag`, `extract_port_from_flag`, `extract_docker_mapping`, `extract_devnet_info_from_docker_line`, `extract_devnet_info_from_cmdline` to a submodule for detection.rs

We could structure it like 

```
devnet/
  detection/
    docker.rs
    cmd.rs (name up to you)
    common.rs (name up to you)
  detection.rs
```

cptartur (commented):
See my comment about splitting up the file, otherwise looks good

crates/sncast/src/helpers/devnet/detection.rs
MKowalski8: Okay, I see what you mean now. Sure do

cptartur (approved):

Issue #3660, opened 2025-08-21
Gas report


ddoktorski:
### Select Tool

snforge

### Requested Feature

Introduce a new `--gas-report` flag that outputs a breakdown table with gas usage details for each contract and entrypoint. The report should also include declare costs for each contract.

It should look similar to the gas report in Eth Foundry: https://getfoundry.sh/forge/gas-tracking/gas-reports/

PR #3794, opened 2025-10-09
Store execution summary in trace data

ddoktorski:
Related #3660

### DIFF
--- a/crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
+++ b/crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
@@ -742,14 +742,22 @@ pub fn update_top_call_vm_trace(runtime: &mut ForgeRuntime, cairo_runner: &mut C
     }
 }
 
-pub fn compute_and_store_execution_summary(trace: &Rc<RefCell<CallTrace>>) -> ExecutionSummary {
+pub fn compute_and_store_execution_summary(trace: &Rc<RefCell<CallTrace>>) {
     let execution_summary = if trace.borrow().nested_calls.is_empty() {
         get_call_execution_summary(trace)
     } else {
         let mut nested_calls_summaries = vec![];
         for nested_call in &trace.borrow().nested_calls {
             if let CallTraceNode::EntryPointCall(nested_call) = nested_call {
-                nested_calls_summaries.push(compute_and_store_execution_summary(nested_call));
+                compute_and_store_execution_summary(nested_call);
+                nested_calls_summaries.push(
+                    nested_call
+                        .borrow()
+                        .gas_report_data
+                        .as_ref()
+                        .expect("Gas report data must be set after calling `compute_and_store_execution_summary`")
+                        .execution_summary
+                        .clone());
             }
         }
         let mut current_call_summary =
@@ -764,8 +772,6 @@ pub fn compute_and_store_execution_summary(trace: &Rc<RefCell<CallTrace>>) -> Ex
     };
 
     trace.borrow_mut().gas_report_data = Some(GasReportData::new(execution_summary.clone()));
-
-    execution_summary
 }
 
 // Based on blockifier/src/execution/call_info.rs (summarize)
--- a/crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
+++ b/crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
@@ -17,14 +17,17 @@ use crate::runtime_extensions::{
         storage::{calculate_variable_address, load, store},
     },
 };
-use crate::state::{CallTrace, CallTraceNode};
+use crate::state::{CallTrace, CallTraceNode, GasReportData};
 use anyhow::{Context, Result, anyhow};
 use blockifier::bouncer::vm_resources_to_sierra_gas;
 use blockifier::context::TransactionContext;
-use blockifier::execution::call_info::CallInfo;
+use blockifier::execution::call_info::{
+    CallInfo, CallSummary, ChargedResources, EventSummary, ExecutionSummary, OrderedEvent,
+};
 use blockifier::execution::contract_class::TrackedResource;
 use blockifier::execution::syscalls::vm_syscall_utils::{SyscallSelector, SyscallUsageMap};
 use blockifier::state::errors::StateError;
+use blockifier::utils::u64_from_usize;
 use cairo_vm::vm::runners::cairo_runner::CairoRunner;
 use cairo_vm::vm::{
     errors::hint_errors::HintError, runners::cairo_runner::ExecutionResources,
@@ -42,10 +45,11 @@ use runtime::{
 };
 use scarb_oracle_hint_service::OracleHintService;
 use starknet::signers::SigningKey;
+use starknet_api::execution_resources::GasAmount;
 use starknet_api::{contract_class::EntryPointType::L1Handler, core::ClassHash};
 use starknet_types_core::felt::Felt;
 use std::cell::RefCell;
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 use std::rc::Rc;
 use std::sync::{Arc, Mutex};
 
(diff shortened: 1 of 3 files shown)


ddoktorski:

- #3794 ◀
    - #3795
        - #3796
            - #3798
                - #3801
                    - #3823

crates/cheatnet/src/state.rs
```diff
@@ -203,6 +203,18 @@ impl<T> CheatStatus<T> {
     }
 }
 
+#[derive(Clone, Debug)]
+pub struct GasReportData {
+    pub execution_summary: ExecutionSummary,
```
franciszekjob, line 208:
```suggestion
    execution_summary: ExecutionSummary,
```

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
```diff
@@ -738,6 +742,59 @@ pub fn update_top_call_vm_trace(runtime: &mut ForgeRuntime, cairo_runner: &mut C
...
+            trace.borrow().used_execution_resources.clone();
+        current_call_summary.charged_resources.gas_consumed =
+            GasAmount(trace.borrow().gas_consumed);
+        current_call_summary
+    };
+
+    trace.borrow_mut().gas_report_data = Some(GasReportData::new(execution_summary.clone()));
+
+    execution_summary
+}
```
franciszekjob: Why don't we add it as method of `CallTrace`?

ddoktorski: There are mainly two reasons:
- This is consistent with the other code e.g functions `update_top_call_resources` and `update_top_call_l1_resources`
- Personally I find it a bit easier to see that we don't mess anything with the pointers due to more explicit borrow calls instead of passing self mutable reference 

However, if you think it's worth making this a method on `CallTrace` it could be changed.

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
```diff
@@ -738,6 +742,59 @@ pub fn update_top_call_vm_trace(runtime: &mut ForgeRuntime, cairo_runner: &mut C
...
+        }
+        let mut current_call_summary =
+            get_call_execution_summary(trace) + nested_calls_summaries.into_iter().sum();
+
+        // vm_resources and gas_consumed of a call already contain the resources of its inner calls.
+        current_call_summary.charged_resources.vm_resources =
+            trace.borrow().used_execution_resources.clone();
+        current_call_summary.charged_resources.gas_consumed =
+            GasAmount(trace.borrow().gas_consumed);
+        current_call_summary
```
cptartur: Just so I understand these steps. We first add all of the summaries but replace gas and vm resources in the result with the ones from `trace` because they were already cumulative? 

Otherwise they would be too high?

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
```diff
@@ -738,6 +742,59 @@ pub fn update_top_call_vm_trace(runtime: &mut ForgeRuntime, cairo_runner: &mut C
...
+pub fn compute_and_store_execution_summary(trace: &Rc<RefCell<CallTrace>>) -> ExecutionSummary {
+    let execution_summary = if trace.borrow().nested_calls.is_empty() {
+        get_call_execution_summary(trace)
+    } else {
+        let mut nested_calls_summaries = vec![];
+        for nested_call in &trace.borrow().nested_calls {
+            if let CallTraceNode::EntryPointCall(nested_call) = nested_call {
+                nested_calls_summaries.push(compute_and_store_execution_summary(nested_call));
+            }
+        }
```
cptartur, line 762: Do we need separate handling for empty nested calls and not? Wouldn't the logic work fine without this?

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
```diff
@@ -738,6 +742,59 @@ pub fn update_top_call_vm_trace(runtime: &mut ForgeRuntime, cairo_runner: &mut C
...
+
+        // vm_resources and gas_consumed of a call already contain the resources of its inner calls.
+        current_call_summary.charged_resources.vm_resources =
+            trace.borrow().used_execution_resources.clone();
+        current_call_summary.charged_resources.gas_consumed =
+            GasAmount(trace.borrow().gas_consumed);
+        current_call_summary
+    };
+
+    trace.borrow_mut().gas_report_data = Some(GasReportData::new(execution_summary.clone()));
```
cptartur, line 774: Why are we modifying the trace here?

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
```diff
@@ -738,6 +742,59 @@ pub fn update_top_call_vm_trace(runtime: &mut ForgeRuntime, cairo_runner: &mut C
...
+
+    execution_summary
+}
+
+// Based on blockifier/src/execution/call_info.rs (summarize)
+fn get_call_execution_summary(trace: &Rc<RefCell<CallTrace>>) -> ExecutionSummary {
```
cptartur: Maybe this could be implemented as `From<CallTrace>` for `ExecutionSummary`?

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
ddoktorski: That's the point of this function to update trace data with the cumulative resources data, so we can later use it for gas calculation

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
ddoktorski: `sum` for the `ExecutionSummary` works in a way that it extends a default value, so if the vector is empty we will get the default

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
ddoktorski: Yes, we want to have cumulative values for each field, gas consumed and vm resources are already cumulative, so we keep the initial values

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
ddoktorski: I'm not sure this is conceptually a good idea. Implementing `From` would suggest that we get the actual, complete execution summary from the given call trace, which is not what this function does. This is a simple helper to prepare an initial struct that we can later update with the nested calls data. Perhaps the name should be improved to something like `get_execution_summary_without_nested_calls`?

cptartur: Renaming is fine as well

cptartur (approved):

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
franciszekjob: I'm more a fan of object-oriented style, but considering that other functions/code is procedural, we can leave it, disregard then.

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
```diff
@@ -738,6 +742,59 @@ pub fn update_top_call_vm_trace(runtime: &mut ForgeRuntime, cairo_runner: &mut C
     }
 }
 
+pub fn compute_and_store_execution_summary(trace: &Rc<RefCell<CallTrace>>) -> ExecutionSummary {
```
franciszekjob: Currently, the method both returns a value and performs a side effect by modifying state. Imo, we should split this into two separate methods - one that returns the `ExecutionSummary`, and another that updates `gas_report_data`.

ddoktorski: We can modify the function to avoid returning the computed value and instead retrieve it from the nested call. The logic in the loop would then be updated to something like:
```rust
compute_and_store_execution_summary(nested_call);
nested_calls_summaries.push(
    nested_call
        .borrow()
        .gas_report_data
        .expect("Gas report data must be set")
        .execution_summary
        .clone(),
);
```
Does this address your comment or do you have another suggestion in mind?

franciszekjob: Ok, this seems enough, let's update as above.

franciszekjob (approved):

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
ddoktorski: https://github.com/foundry-rs/starknet-foundry/pull/3794/commits/ea4328cd001e5c332e5c5c97474bd4e0fce3352e

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
ddoktorski: https://github.com/foundry-rs/starknet-foundry/pull/3794/commits/e2675f3d94af5f11f81e6b7c2e144adc94a35012

PR #3795, opened 2025-10-09
Keep resources of failed calls in trace

ddoktorski:
Related #3660, #3624

PR #3796, opened 2025-10-09
Add gas component to debugging trace

### LINKED ISSUE #3624: Add gas breakdown
### Select Tool

snforge

### Requested Feature

Add gas vector as new component

ddoktorski:
Closes #3624

### DIFF
--- a/crates/cheatnet/src/state.rs
+++ b/crates/cheatnet/src/state.rs
@@ -10,6 +10,7 @@ use crate::runtime_extensions::forge_runtime_extension::cheatcodes::cheat_execut
 };
 use crate::runtime_extensions::forge_runtime_extension::cheatcodes::spy_events::Event;
 use crate::runtime_extensions::forge_runtime_extension::cheatcodes::spy_messages_to_l1::MessageToL1;
+use blockifier::blockifier_versioned_constants::VersionedConstants;
 use blockifier::execution::call_info::{ExecutionSummary, OrderedEvent, OrderedL2ToL1Message};
 use blockifier::execution::contract_class::RunnableCompiledClass;
 use blockifier::execution::entry_point::CallEntryPoint;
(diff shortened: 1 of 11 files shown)


ddoktorski:

- #3794
    - #3795
        - #3796 ◀
            - #3798
                - #3801
                    - #3823

crates/cheatnet/src/state.rs
```diff
@@ -206,12 +209,25 @@ impl<T> CheatStatus<T> {
 #[derive(Clone, Debug)]
 pub struct GasReportData {
     pub execution_summary: ExecutionSummary,
+    pub gas: OnceCell<GasVector>,
```
franciszekjob:
```suggestion
    gas: OnceCell<GasVector>,
```

franciszekjob: Why don't we call it `l2_gas`?

crates/debugging/src/tree/ui/display.rs
```diff
@@ -84,6 +84,13 @@ impl NodeDisplay for TransformedCallResult {
     }
 }
 
+impl NodeDisplay for Gas {
+    const TAG: &'static str = "L2 gas";
+    fn string_pretty(&self) -> String {
+        self.0.0.to_string()
```
ksew1:
```suggestion
        self.0.to_string()
```

ksew1 (approved):

crates/cheatnet/src/state.rs
cptartur: Do we ever want to expand this for other gas types? I'd rename it then.

cc @ksew1

cptartur (commented):
Left a comment

crates/cheatnet/src/state.rs
franciszekjob: > Do we ever want to expand this for other gas types? I'd rename it then.

As far as I remember l2 gas should be in initial implementation, and in future we want to add rest as well.

ksew1: I would call it L2 gas. It is not user facing so not a big deal

ddoktorski: This field can also contain L1 gas from L2->L1 messages. We can either rename it to `l2_gas` and change the type to `GasAmount` or keep the current struct and rename it to something like `partial_gas_usage`. It seems more convenient to me to keep the `GasVector` here so it can be easily extended in the future if necessary

cptartur: Didn't realize it's a GasVector. If that's for future proofing, lets make sure it's obvious from code (either rename somehow or add a comment).

cptartur (approved):
See my response to a comment about GasVector, otherwise looks good

crates/cheatnet/src/state.rs
franciszekjob: > Didn't realize it's a GasVector.

Same here 😅 , resolving then

franciszekjob (approved):

Issue #2319, opened 2024-07-17
Implement balance command


ksew1:
### Which component is your feature related to?

Cast

### Feature Request

Add command for checking balance of account, we should currently support:
- ETH
- STRK

PR #3817, opened 2025-10-14
Add `sncast balance` command

franciszekjob:
Towards #2319 

**Stack**:
- #3818
- #3817 ⬅

## Introduced changes

Add `sncast balance` command

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/sncast/src/helpers/token.rs
```diff
@@ -0,0 +1,61 @@
...
+impl fmt::Display for Token {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Token::Strk => write!(f, "strk"),
+            Token::Eth => write!(f, "eth"),
+        }
+    }
+}
+
+impl CairoSerialize for Token {
```
cptartur: Why do we need `CairoSerialize` for that?

crates/sncast/src/response/balance.rs
```diff
@@ -0,0 +1,49 @@
...
+        } else {
+            balance
+        };
+
+        styling::OutputBuilder::new()
+            .field(
+                "Account Address",
+                &self.command_response.account_address.into_padded_hex_str(),
+            )
+            .field("Balance", &balance_str)
```
cptartur, line 29: Do we need it it include account address? User has to provide account anyways.

crates/sncast/src/starknet_commands/utils/balance.rs
```diff
@@ -0,0 +1,106 @@
...
+use sncast::response::balance::BalanceResponse;
+use sncast::response::errors::StarknetCommandError;
+use starknet::{
+    core::{types::FunctionCall, utils::get_selector_from_name},
+    providers::{JsonRpcClient, Provider, jsonrpc::HttpTransport},
+};
+use starknet_types_core::felt::Felt;
+
+#[derive(Args, Debug, Clone)]
+#[group(id = "TokenIdentifier", multiple = false)]
```
cptartur: Why do we need to provide group id explicitly?

Also, shouldn't we set `required = true`?

crates/sncast/src/starknet_commands/utils/balance.rs
```diff
@@ -0,0 +1,106 @@
...
+        entry_point_selector: get_selector_from_name("balance_of").expect("Failed to get selector"),
+        calldata: vec![account_address],
+    };
+    let block_id = get_block_id(&balance.block_id)?;
+
+    let res = provider.call(call, block_id).await.map_err(|err| {
+        StarknetCommandError::ProviderError(
+            sncast::response::errors::SNCastProviderError::UnknownError(err.into()),
+        )
+    })?;
```
cptartur: Why are we mapping this into unknown error if `provider.call` actually returns a proper error?

crates/sncast/src/starknet_commands/utils/balance.rs
```diff
@@ -0,0 +1,106 @@
...
+    let res = res
+        .iter()
+        .map(|val| u128::from_str_radix(&val.to_string(), 16).expect("Failed to parse u128"))
```
cptartur: Will the response always be u256? If it implements the ERC-20 interface yes, but otherwise I'm not sure anything is enforcing that. Shouldn't we return an error about parsing of balance failed instead of panicking?

crates/sncast/src/starknet_commands/balance.rs
```diff
@@ -0,0 +1,106 @@
...
+        .iter()
+        .map(|val| u128::from_str_radix(&val.to_string(), 16).expect("Failed to parse u128"))
+        .collect::<Vec<u128>>();
+
+    let (low, high) = match res.as_slice() {
+        [low, high] => (*low, *high),
+        _ => {
+            return Err(StarknetCommandError::UnknownError(anyhow!(
+                "Balance response should contain exactly two u128 values"
+            )));
```
cptartur: Cannot we either create a new enum variant for the error or return `anyway::Result` from the function instead of the `Result<BalanceResponse, StarknetCommandError>`? 

We just `into()` this error to `anyway::Error` when we call `handle_starknet_command_error` anyways.

crates/sncast/src/starknet_commands/utils/balance.rs
```diff
@@ -0,0 +1,106 @@
...
+                "Balance response should contain exactly two u128 values"
+            )));
+        }
+    };
+
+    let displayed_token = match (
+        balance.token_identifier.token,
+        balance.token_identifier.token_address,
+    ) {
+        (Some(tok), None) => Some(tok),
```
cptartur: We aren't saving anything with abbreviations

```suggestion
        (Some(token), None) => Some(token),
```

crates/sncast/src/starknet_commands/utils/balance.rs
```diff
@@ -0,0 +1,106 @@
...
+            return Err(StarknetCommandError::UnknownError(anyhow!(
+                "Balance response should contain exactly two u128 values"
+            )));
+        }
+    };
+
+    let displayed_token = match (
+        balance.token_identifier.token,
+        balance.token_identifier.token_address,
+    ) {
```
cptartur: This could be a method on `token_identifier`

cptartur: Also why the preference for matches like this instead of if else?

crates/sncast/src/starknet_commands/utils/mod.rs
```diff
@@ -74,6 +80,25 @@ pub async fn utils(
 
             process_command_result("class-hash", Ok(result), ui, None);
         }
+
+        Commands::Balance(balance) => {
```
cptartur: I don't think this fits in `utils`. Right now, utils contains offline, helper commands.

Maybe let's just have it in top level as `sncast balance`, or we could even consider `sncast account balance`, though not sure about that.

crates/sncast/tests/e2e/balance.rs
```diff
@@ -0,0 +1,167 @@
...
+use crate::helpers::runner::runner;
+use indoc::{formatdoc, indoc};
+use shared::test_utils::output_assert::AsOutput;
+use sncast::helpers::token::Token;
+use tempfile::tempdir;
+use test_case::test_case;
+
+#[tokio::test]
+pub async fn happy_case() {
+    let tempdir = tempdir().expect("Unable to create a temporary directory");
```
cptartur: Nit but we can just unwrap in tests, IDE will show you the error location.

crates/sncast/tests/e2e/balance.rs
```diff
@@ -0,0 +1,167 @@
...
+        For more information, try '--help'.
+    "});
+}
+
+#[tokio::test]
+pub async fn happy_case_with_token_address() {
+    let tempdir = tempdir().expect("Unable to create a temporary directory");
+    let accounts_json_path = get_accounts_path("tests/data/accounts/accounts.json");
+
+    let strk_address = "0x04718f5a0fc34cc1af16a1cdee98ffb20c31f5cd61d6ab07201858f4287c938d";
```
cptartur, line 138: I'd check it with some popular custom token instead of a one we support out of the box.

crates/sncast/src/helpers/token.rs
franciszekjob: `BalanceResponse` derives `CairoSerialize` so all it's fields must derive it too (there's field of type `Token`).

crates/sncast/src/response/balance.rs
franciszekjob: We can remove it, maybe it's a bit redundant.

crates/sncast/src/starknet_commands/utils/balance.rs
franciszekjob: > Why do we need to provide group id explicitly?

Actually no need to hardcode it.

> Also, shouldn't we set required = true?

No, we have default (strk) if none of these flags are passed.

crates/sncast/src/starknet_commands/utils/balance.rs
franciszekjob: Mistake, change to relevant error.

crates/sncast/src/starknet_commands/utils/balance.rs
franciszekjob: We only support ERC-20 tokens (in fact I should mention this in command help + docs) - yes.

> Shouldn't we return an error about parsing of balance failed instead of panicking?

Changed.

crates/sncast/src/starknet_commands/utils/balance.rs
franciszekjob: > This could be a method on `token_identifier`

Applied

> Also why the preference for matches like this instead of if else?

Easier to read for 2 or more values + it's a bit a matter of taste.

crates/sncast/src/starknet_commands/utils/mod.rs
franciszekjob: > Right now, utils contains offline, helper commands.

No, `sncast utils serialize` is not offline if `--class-hash` is passed.

> Maybe let's just have it in top level as `sncast balance`, or we could even consider `sncast account balance`, though not sure about that.

Initially I wanted to place it as subcommand of `sncast account` but then usage is like `sncast --account xyz account ...` which seems like a bad structure UX-wise (we use account both as flag and subcommand). 

Placing it as standalone command is fine, I even kinda like it because of simplicity, let's see what others say cc @ksew1 @ddoktorski @MKowalski8 .

ksew1: I would go with `sncast balance`; I would say utils is for niche use cases, and this seems like a standard action to perform.

crates/sncast/src/helpers/token.rs
cptartur: Do we need to derive it for `BalanceResponse`?

crates/sncast/src/response/balance.rs
cptartur: I'd remove it

crates/sncast/src/starknet_commands/utils/mod.rs
cptartur: > > Right now, utils contains offline, helper commands.
> 
> No, `sncast utils serialize` is not offline if `--class-hash` is passed.
> 

Fair point, but I think we should treat `utils` as more of a helpers methods for less standard use cases.

> Initially I wanted to place it as subcommand of sncast account but then usage is like sncast --account xyz account ... which seems like a bad structure UX-wise (we use account both as flag and subcommand).

Let's simply have it as `sncast account` then if others agree

ddoktorski: > Let's simply have it as sncast account then if others agree

What if someone wants to check the balance of a non-account contract?

franciszekjob: > Let's simply have it as sncast account then if others agree

@cptartur didn't you mean `sncast balance`?

crates/sncast/tests/e2e/balance.rs
franciszekjob: It doesn't change anything if we want to test logic with `--token-address`, resolving.

crates/sncast/src/starknet_commands/utils/mod.rs
MKowalski8: > Placing it as standalone command is fine, I even kinda like it because of simplicity, let's see what others say cc 

For me, we can go just with `sncast balance`

crates/sncast/src/helpers/token.rs
ksew1: I don't see reason why we need to derive it on `BalanceResponse`

crates/sncast/src/helpers/token.rs
```diff
@@ -0,0 +1,61 @@
...
+use starknet_types_core::felt::Felt;
+use std::{fmt, str::FromStr};
+
+const STRK_CONTRACT_ADDRESS: Felt =
+    felt!("0x04718f5a0fc34cc1af16a1cdee98ffb20c31f5cd61d6ab07201858f4287c938d");
+
+const ETH_CONTRACT_ADDRESS: Felt =
+    felt!("0x049d36570d4e46f48e99674bd3fcc84644ddd6b96f7c741b1562b82f9e004dc7");
+
+#[derive(Default, Serialize, Deserialize, Clone, Copy, Debug, PartialEq, Eq, ValueEnum)]
```
ksew1:
```suggestion
#[derive(Default, Serialize, Clone, Copy, Debug, ValueEnum)]
```

It is enough. This is internal code, if somebody will need them they will just derive it

crates/sncast/src/response/balance.rs
```diff
@@ -0,0 +1,43 @@
...
+    helpers::token::Token,
+    response::{cast_message::SncastMessage, command::CommandResponse},
+};
+use conversions::serde::serialize::CairoSerialize;
+use foundry_ui::{Message, styling};
+use serde::{Deserialize, Serialize};
+use serde_json::json;
+use starknet::core::types::U256;
+
+#[derive(Clone, Serialize, Deserialize, CairoSerialize, Debug, PartialEq)]
```
ksew1:
```suggestion
#[derive(Serialize, Debug)]
```

crates/sncast/src/response/balance.rs
```diff
@@ -0,0 +1,43 @@
...
+            balance
+        };
+
+        styling::OutputBuilder::new()
+            .field("Balance", &balance_str)
+            .build()
+    }
+
+    fn json(&self) -> serde_json::Value {
+        serde_json::to_value(&self.command_response).unwrap_or_else(|err| {
```
ksew1: Pretty much always we doing this code. Why not just create one function that will be used in that cases, like:

```rs
pub fn serialize_json(sncast_message: &SncastMessage<impl CommandResponse>) -> serde_json::Value {
    serde_json::to_value(&sncast_message.command_response).unwrap_or_else(|err| {
        serde_json::json!({
            "error": "Failed to serialize response",
            "command": sncast_message.command,
            "details": err.to_string()
        })
    })
}
```

crates/sncast/src/helpers/token.rs
```diff
@@ -0,0 +1,61 @@
...
+    #[must_use]
+    pub fn contract_address(self) -> Felt {
+        match self {
+            Token::Strk => STRK_CONTRACT_ADDRESS,
+            Token::Eth => ETH_CONTRACT_ADDRESS,
+        }
+    }
+}
+
+impl FromStr for Token {
```
ksew1: This is not needed

crates/sncast/src/helpers/token.rs
```diff
@@ -0,0 +1,61 @@
...
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s {
+            "strk" => Ok(Token::Strk),
+            "eth" => Ok(Token::Eth),
+            token => Err(anyhow!("Invalid token: {token}")),
+        }
+    }
+}
+
+impl fmt::Display for Token {
```
ksew1: let's just use `strum`

crates/sncast/src/helpers/token.rs
```diff
@@ -0,0 +1,61 @@
...
+#[serde(rename_all = "lowercase")]
+pub enum Token {
+    #[default]
+    Strk,
+    Eth,
+}
+
+impl Token {
+    #[must_use]
+    pub fn contract_address(self) -> Felt {
```
ksew1:
```rs
impl Token {
    #[must_use]
    pub fn as_contract_address(&self) -> Felt {
        match self {
            Token::Strk => STRK_CONTRACT_ADDRESS,
            Token::Eth => ETH_CONTRACT_ADDRESS,
        }
    }
}
```

No need to take by value

crates/sncast/src/response/balance.rs
```diff
@@ -0,0 +1,43 @@
...
+};
+use conversions::serde::serialize::CairoSerialize;
+use foundry_ui::{Message, styling};
+use serde::{Deserialize, Serialize};
+use serde_json::json;
+use starknet::core::types::U256;
+
+#[derive(Clone, Serialize, Deserialize, CairoSerialize, Debug, PartialEq)]
+pub struct BalanceResponse {
+    pub balance: (u128, u128),
```
ksew1: I know this is the trick as `U256` doesn't implement Serialize. Maybe we should use `CairoU256` from datatransformer

crates/sncast/src/starknet_commands/balance.rs
```diff
@@ -0,0 +1,106 @@
...
+impl TokenIdentifier {
+    pub fn contract_address(&self) -> Felt {
+        if let Some(addr) = self.token_address {
+            addr
+        } else if let Some(tok) = self.token {
+            tok.contract_address()
+        } else {
+            // Both token and token address are optional, hence we cannot have
+            // default value for token at clap level.
+            Token::Strk.contract_address()
```
ksew1:
```suggestion
            Token::default().as_contract_address()
```

crates/sncast/src/starknet_commands/balance.rs
franciszekjob: https://github.com/foundry-rs/starknet-foundry/pull/3817/commits/ede212d0a2849f0a93c87023357e7d444ca32494

crates/sncast/src/helpers/token.rs
franciszekjob: FIne, but I would stay with `contract_address()`

crates/sncast/src/response/balance.rs
franciszekjob: > I know this is the trick as U256 doesn't implement Serialize.

Exactly

> Maybe we should use CairoU256 from datatransformer

It doesn't implement `Serialize` so we can't use it without additional changes. Imo not really worth the effort

crates/sncast/src/response/balance.rs
franciszekjob: We can, but let's address this in separate PR, not to mix changes.

crates/sncast/src/starknet_commands/balance.rs
franciszekjob: Changed error type to `anyhow::Error`
https://github.com/foundry-rs/starknet-foundry/pull/3817/commits/7b0111c1052a20870a78a0d4b98f4b40bdfd7785

crates/sncast/src/response/balance.rs
franciszekjob: https://github.com/foundry-rs/starknet-foundry/pull/3822

crates/sncast/src/helpers/token.rs
```diff
@@ -0,0 +1,29 @@
...
+const STRK_CONTRACT_ADDRESS: Felt =
+    felt!("0x04718f5a0fc34cc1af16a1cdee98ffb20c31f5cd61d6ab07201858f4287c938d");
+
+const ETH_CONTRACT_ADDRESS: Felt =
+    felt!("0x049d36570d4e46f48e99674bd3fcc84644ddd6b96f7c741b1562b82f9e004dc7");
+
+#[derive(Default, Serialize, Clone, Copy, Debug, ValueEnum, strum_macros::Display)]
+pub enum Token {
+    #[default]
+    #[strum(serialize = "strk")]
```
cptartur: Can't we use do this with plain serde? Maybe this is relevant here https://serde.rs/container-attrs.html#rename_all

crates/sncast/src/response/balance.rs
cptartur: How will this be displayed to the user using `--json` output? I'd assume as two u128, which is not ideal.

There's `U256` in crate `primitive-types` which is already a dependency in our workspace, just need to add it to cast. You'll need to change it in top level Cargo.toml so it derives serde.

`primitive-types = {  version =  "0.13.1", features = ["serde"] }`

Not sure about constructing it from two u128 but maybe it can be achieved.

crates/sncast/src/starknet_commands/balance.rs
```diff
@@ -0,0 +1,109 @@
...
+        } else if let Some(tok) = self.token {
+            tok.contract_address()
+        } else {
+            // Both token and token address are optional, hence we cannot have
+            // default value for token at clap level.
+            Token::default().contract_address()
+        }
+    }
+
+    pub fn displayed_token(&self) -> Option<Token> {
```
cptartur:
```suggestion
    pub fn token_suffix(&self) -> Option<Token> {
```

crates/sncast/src/starknet_commands/balance.rs
```diff
@@ -0,0 +1,109 @@
...
+                .map_err(|_| anyhow::anyhow!("Failed to parse balance as u128"))
+        })
+        .collect();
+
+    let (low, high) = match res?.as_slice() {
+        [low, high] => (*low, *high),
+        _ => {
+            // return Err(StarknetCommandError::UnknownError(anyhow!(
+            //     "Balance response should contain exactly two u128 values"
+            // )));
```
cptartur:
```suggestion
```

crates/sncast/src/helpers/token.rs
```diff
@@ -0,0 +1,29 @@
...
+    Strk,
+    #[strum(serialize = "eth")]
```
ksew1:
```suggestion
     #[strum(to_string = "lowercase")]
```

crates/sncast/src/helpers/token.rs
ksew1:
```suggestion
    #[strum(to_string = "lowercase")]
```

crates/sncast/src/helpers/token.rs
franciszekjob: Almost correct, `#[strum(serialize_all = "lowercase")]` should be used to whole `Token` enum

crates/sncast/src/helpers/token.rs
franciszekjob: `#[serde(rename_all = "lowercase")]` will make us accept lowercase values, but when converted to string, they will remain as they are. `#[strum(serialize_all = "lowercase")]` achieves what you want.

Buut, we also need `#[serde(rename_all = "lowercase")]` because with `--json` flag we want to keep the token lowercase.

crates/sncast/src/response/balance.rs
franciszekjob: Yeah it's possible. Needed a custom serialization for `BalanceResponse` though, because default `U256` serialies to hex string.

crates/sncast/src/response/balance.rs
```diff
@@ -0,0 +1,54 @@
...
+            self.command_response.balance.to_string()
+        };
+
+        styling::OutputBuilder::new()
+            .field("Balance", &balance_str)
+            .build()
+    }
+
+    fn json(&self) -> serde_json::Value {
+        serde_json::to_value(&self.command_response).unwrap_or_else(|err| {
```
ksew1: I guess we can rebase and use new method

ksew1 (approved):

crates/sncast/src/helpers/token.rs
```diff
@@ -0,0 +1,29 @@
...
+
+const STRK_CONTRACT_ADDRESS: Felt =
+    felt!("0x04718f5a0fc34cc1af16a1cdee98ffb20c31f5cd61d6ab07201858f4287c938d");
+
+const ETH_CONTRACT_ADDRESS: Felt =
+    felt!("0x049d36570d4e46f48e99674bd3fcc84644ddd6b96f7c741b1562b82f9e004dc7");
+
+#[derive(Default, Serialize, Clone, Copy, Debug, ValueEnum, strum_macros::Display)]
+#[strum(serialize_all = "lowercase")]
+#[serde(rename_all = "lowercase")]
```
cptartur, line 16: Let's add a comment in code about why we need them both so someone doesn't remove it by accident.

crates/sncast/src/starknet_commands/balance.rs
```diff
@@ -0,0 +1,113 @@
...
+
+    #[command(flatten)]
+    pub rpc: RpcArgs,
+}
+
+pub async fn balance(
+    account_address: Felt,
+    provider: &JsonRpcClient<HttpTransport>,
+    balance: &Balance,
+) -> Result<BalanceResponse, anyhow::Error> {
```
cptartur: Nit but you can do `use anyhow::Result` at the top of the file and then the signature will just be this

```suggestion
) -> Result<BalanceResponse> {
```

crates/sncast/src/starknet_commands/balance.rs
```diff
@@ -0,0 +1,113 @@
...
+    let token = &balance.token_identifier.token_suffix();
+    let balance = erc20_balance_to_u256(&res)?;
+
+    Ok(BalanceResponse {
+        balance,
+        token: *token,
+    })
+}
+
+fn erc20_balance_to_u256(balance: &[Felt]) -> Result<U256, Error> {
```
cptartur, line 96: Maybe some basic unit tests could be added for that

crates/sncast/tests/e2e/balance.rs
```diff
@@ -0,0 +1,203 @@
...
+        "user1",
+        "balance",
+        "--url",
+        URL,
+    ];
+
+    let snapbox = runner(&args).current_dir(tempdir.path());
+
+    snapbox.assert().stdout_matches(indoc! {r#"
+        {"balance":"[..]"}
```
cptartur, line 181: It would be nice to assert it is actually a valid decimal string, because we are doing custom serialization, if that's possible to do easily.

cptartur (approved):
Left some comments, but otherwise looks good. Feel free to merge after they are resolved.

PR #3821, opened 2025-10-15
Replaced uses: actions/setup-node@v5 → uses: actions/setup-node@v6

MamunC0der:
Release notes: https://github.com/actions/setup-node/releases/tag/v6.0.0


cptartur:
@MamunC0der Hello, thanks for contributing, however we are not accepting PRs that only bump dependencies as these are handled by dependabot.

PR #3822, opened 2025-10-15
Refactor sncast messages json

franciszekjob:
Closes #

## Introduced changes

-

## Checklist

- [ ] Linked relevant issue
- [ ] Added relevant tests


crates/sncast/src/response/balance.rs
cptartur, line 1: Is this relevant to this PR?

crates/sncast/src/response/helpers.rs
cptartur, line 1: I'd name this file something more specific so it doesn't became a collection of unrelated functions. E.g. even `json.rs` would work for me.

crates/sncast/src/response/helpers.rs
```diff
@@ -0,0 +1,11 @@
...
+
+pub fn serialize_json(sncast_message: &SncastMessage<impl CommandResponse>) -> serde_json::Value {
+    serde_json::to_value(&sncast_message.command_response).unwrap_or_else(|err| {
+        serde_json::json!({
+            "error": "Failed to serialize response",
+            "command": sncast_message.command,
+            "details": err.to_string()
+        })
+    })
+}
```
cptartur: If we want to use shared logic for jsons for all (or most) cast messages, we can create a trait with a default impl like that

```rust
pub trait SncastTextMessage {
    fn text(&self) -> String;
}

impl<T: CommandResponse> Message for SncastMessage<T>
where
    SncastMessage<T>: SncastTextMessage,
{
    fn text(&self) -> String {
        SncastTextMessage::text(self)
    }

    fn json(&self) -> Value {
        serialize_json(self)
    }
}
```

and implement it (`SncastTextMessage`) for relevant responses instead of implementing `Message` directly.

This way you won't have to repeat the impl with call to `serialize_json` and we can still have plain `Message` impls where needed.

crates/sncast/src/response/balance.rs
franciszekjob: Removed

crates/sncast/src/response/helpers.rs
franciszekjob: It's inlined now so there's no separate file.

crates/sncast/src/response/helpers.rs
franciszekjob: `SncastCommandMessage` seems better, applied.

cptartur (approved):

ksew1 (approved):

PR #3826, opened 2025-10-16
micro opt: stop emitting __internal_config_statement from test attr

maciektr:
The test attr is a hot path, so it may make sense to avoid forcing the expansion to happen twice for every test. The time penalty is not big, but seems easy enough to avoid.

Events breakdown for OZ before / after:

image

image

### DIFF
--- a/crates/snforge-scarb-plugin/src/attributes/test.rs
+++ b/crates/snforge-scarb-plugin/src/attributes/test.rs
@@ -2,7 +2,7 @@ use super::{AttributeInfo, ErrorExt, internal_config_statement::InternalConfigSt
 use crate::asserts::assert_is_used_once;
 use crate::common::{has_fuzzer_attribute, has_test_case_attribute};
 use crate::external_inputs::ExternalInput;
-use crate::utils::{create_single_token, process_statements};
+use crate::utils::{create_single_token, get_statements};
 use crate::{
     args::Arguments,
     common::{into_proc_macro_result, with_parsed_values},
@@ -95,12 +95,7 @@ fn test_internal(
 
         let test_func_with_attrs = test_func_with_attrs(&test_func, &called_func, &call_args);
 
-        let statements = func
-            .body(db)
-            .statements(db)
-            .elements(db)
-            .collect::<Vec<_>>();
-        let (statements, if_content) = process_statements(db, &statements);
+        let (statements, if_content) = get_statements(db, func);
 
         Ok(quote!(
             #test_func_with_attrs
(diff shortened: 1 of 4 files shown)


crates/snforge-scarb-plugin/src/attributes/test.rs
```diff
@@ -98,13 +95,27 @@ fn test_internal(
...
 
             #attributes
-            #[#internal_config]
             fn #func_ident #signature
-            #body
+            {
+               if snforge_std::_internals::is_config_run() {
+                   #if_content
+
+                   return;
```
cptartur, line 109: I think we should be able to do the same in fuzzer wrapper in `wrapper.rs` file. 

Replace appending `#[#internal_config_attr]` there with simply creating an another if branch like you did here.

Then we could get rid of the internal config statement attribute entirely.

cptartur: cc @ddoktorski @franciszekjob please check if I'm not making things up

ddoktorski: Yup, I think that should be possible to do

ddoktorski (approved):

crates/snforge-scarb-plugin/src/attributes/test.rs
maciektr: @cptartur This sounds like a good refactoring idea, but tbh I don't think this matters at all apart from code cleanliness - you would normally expect a lot more `test` attrs than any other. 

Especially as I would need to have some equivalent of https://github.com/foundry-rs/starknet-foundry/blob/0618a5d717c5bc41a83b99742e3ce0de13af40ab/crates/snforge-scarb-plugin-deprecated/src/utils.rs#L72 that accepts TokenStream instead of FunctionWithBody to keep the logic exactly the same.

cptartur: We can do that refactor at a later date then.

cptartur (approved):

crates/snforge-scarb-plugin/src/attributes/test.rs
```diff
@@ -98,13 +95,27 @@ fn test_internal(
 
         let test_func_with_attrs = test_func_with_attrs(&test_func, &called_func, &call_args);
 
+        let statements = func
+            .body(db)
+            .statements(db)
+            .elements(db)
+            .collect::<Vec<_>>();
+        let (statements, if_content) = process_statements(db, &statements);
```
cptartur: Why did you need to extract `process_statements` here? Couldn't existing `get_statements` be used by passing the `func` reference we already have?

maciektr: guess we can

cptartur (approved):

PR #3827, opened 2025-10-17
Do not instantiate Program for each test case run separately when fuzzing

maciektr:
Correct me if I am wrong, but it seems we recreate the same test case program multiple times when fuzzing. 

When run with `snforge test --workspace --features fuzzing --fuzzer-runs 500 ` on OZ:
Before:  272.79s
After: 208.78s

### DIFF
--- a/crates/forge-runner/src/lib.rs
+++ b/crates/forge-runner/src/lib.rs
@@ -176,9 +176,12 @@ fn run_with_fuzzing(
 
         let mut tasks = FuturesUnordered::new();
 
+        let program = case.try_into_program(&casm_program)?;
+
         for _ in 1..=fuzzer_runs.get() {
             tasks.push(run_fuzz_test(
                 case.clone(),
+                program.clone(),
                 casm_program.clone(),
                 forge_config.clone(),
                 versioned_program_path.clone(),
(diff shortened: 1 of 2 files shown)


cptartur (commented):
I don't think creating program does any compilation, no? We compile sierra to casm here https://github.com/foundry-rs/starknet-foundry/blob/291ef23b8df49165d08da0394608f83d24a54f31/crates/forge-runner/src/running/with_config.rs#L45

I was under impression that creating a `Program` simply performs some transformations on the stuff we already have. 

But I'm all for merging this if it improves performance, please just address the title or correct me if I don't understand.

maciektr:
@cptartur yeah, my lingo is off. 
The point is, this does a lot of copying / allocations and can actually take roughly as much as the test runtime itself. 
Before: 
image
After: 
image

maciektr:
Test runs: 
Before: https://github.com/maciektr/cairo-contracts/actions/runs/18588786535/job/52998742316
After: https://github.com/maciektr/cairo-contracts/actions/runs/18588112713/job/52996649469

cptartur (approved):

ddoktorski (approved):
Thanks!

PR #3836, opened 2025-10-20
Create nightly release on schedule

### LINKED ISSUE #3829: Run Maat raports as part of scheduled tests
### Current State

Currently we only generate Maat reports for stable foundry releases. It would be good to have some intermediate report during the release cycle. 

### Objective

Run Maat raports as part of scheduled tests

### Additional Context

_No response_

cptartur:
Closes #3829 

## Introduced changes

-

## Checklist

- [ ] Linked relevant issue
- [ ] Added relevant tests

### DIFF
--- a/.github/workflows/scheduled.yml
+++ b/.github/workflows/scheduled.yml
@@ -6,17 +6,12 @@ on:
       - scripts/get_scarb_versions.sh
       - .github/workflows/scheduled.yml
   schedule:
-    - cron: '0 0 * * 3,0'
+    # Two schedules needed so we can distinguish between nightly and non-nightly runs
+    - cron: '0 0 * * 0'
+    - cron: '0 0 * * 3'
   workflow_dispatch:
 
 jobs:
-  create-nightly-release:
-    # Creating a nightly release will trigger a Maat experiment for it as well
-    if: ${{ (github.event_name == 'schedule') || (github.event_name == 'workflow_dispatch') }}
-    secrets: inherit
-    uses:
-      ./.github/workflows/nightly.yml
-
   get-scarb-versions:
     if: github.event.repository.fork == false
     name: Get Scarb versions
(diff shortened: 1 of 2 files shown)


.github/workflows/scheduled.yml
```diff
@@ -10,6 +10,13 @@ on:
   workflow_dispatch:
 
 jobs:
+  create-nightly-release:
```
ddoktorski: Are we sure we want to run nightly builds twice a week?

.github/workflows/scheduled.yml
```diff
@@ -10,6 +10,13 @@ on:
...
+    # Creating a nightly release will trigger a Maat experiment for it as well
+    if: ${{ (github.event_name == 'schedule') || (github.event_name == 'workflow_dispatch') }}
```
ddoktorski: Why for both events?

cptartur: I added this mostly for testing the workflow

.github/workflows/scheduled.yml
cptartur: Don't think it's problematic to do so, especially since they are called that

ddoktorski: It seems too frequent to me. My concern is that with that frequency the reports either won’t be checked thoroughly or will unnecessarily take too much time. But if it works for you I’m fine with that

.github/workflows/scheduled.yml
ddoktorski, line 1: nit: Shouldn't we start the nightly build at the end only if all tests passed?

ddoktorski (commented):
Why have some checks been canceled?

.github/workflows/scheduled.yml
cptartur: Fair point, I've updated the workflow so that nightly is only built on Sundays

.github/workflows/scheduled.yml
cptartur: Removed

.github/workflows/scheduled.yml
cptartur: Changed

ddoktorski (approved):

franciszekjob (approved):

PR #3838, opened 2025-10-21
Release 0.51.0

cptartur:
Closes #

## Introduced changes

- Reverted removing `--experimental-oracles` flag until all issues with oracles are resolved

## Checklist

- [ ] Linked relevant issue
- [ ] Added relevant tests


CHANGELOG.md
```diff
@@ -33,6 +34,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Support for automatically declaring the contract when running `sncast deploy`, by providing `--contract-name` flag instead of `--class-hash`. 
 - `sncast balance` command to fetch the balance of an account for a specified token.
```
ddoktorski, line 35: Shouldn't it be "balance of a contract" rather than "balance of an account"?

franciszekjob (approved):

CHANGELOG.md
ddoktorski, line 1: Entry about sncast logging should have `SNCAST_LOG` instead of `CAST_LOG`

MKowalski8 (approved):

CHANGELOG.md
cptartur: It works based on an `--account` argument you provide to it, so the wording is correct

CHANGELOG.md
cptartur: Good catch

CHANGELOG.md
franciszekjob: Why? Account is indeed a contract, but check balance with `--account` specified so current entry in changelog seems accurate.

ddoktorski: Is there a specific reason we limit its use to accounts and not to all contracts?

franciszekjob: Ok I see what you mean. We can ofc extend it so it will also accept pure address.

PR #3801, opened 2025-10-13
Change default tracked resource to sierra gas

ddoktorski:
Related #3660

PR #3798, opened 2025-10-10
Move trace data code to separate file

ddoktorski:
[Refactor only]

PR #3841, opened 2025-10-21
Add gas report logic

ddoktorski:
Towards #3660

### DIFF
--- a/crates/debugging/src/lib.rs
+++ b/crates/debugging/src/lib.rs
@@ -7,5 +7,6 @@ mod contracts_data_store;
 mod trace;
 mod tree;
 
+pub use contracts_data_store::ContractsDataStore;
 pub use trace::components::{Component, Components};
-pub use trace::{context::Context, types::Trace};
+pub use trace::{context::Context, types::ContractName, types::Trace};
--- a/crates/forge-runner/src/gas/report.rs
+++ b/crates/forge-runner/src/gas/report.rs
@@ -1,6 +1,5 @@
 use crate::gas::stats::GasStats;
 use cheatnet::trace_data::{CallTrace, CallTraceNode};
-use debugging::ContractName as DebuggingContractName;
 use debugging::ContractsDataStore;
 use starknet_api::core::{ClassHash, EntryPointSelector};
 use starknet_api::execution_resources::GasVector;
@@ -105,10 +104,8 @@ impl SingleTestGasInfo {
 fn get_contract_name(contracts_data: &ContractsDataStore, class_hash: ClassHash) -> ContractName {
     contracts_data
         .get_contract_name(&class_hash)
-        .cloned()
-        .unwrap_or_else(|| DebuggingContractName("forked contract".to_string()))
-        .0
-        .clone()
+        .map_or("forked contract", |name| name.0.as_str())
+        .to_string()
 }
 
 fn get_selector(contracts_data: &ContractsDataStore, selector: EntryPointSelector) -> Selector {
--- a/crates/forge-runner/src/gas/report.rs
+++ b/crates/forge-runner/src/gas/report.rs
@@ -83,9 +83,10 @@ impl GasSingleTestInfo {
     ) {
         let contract_info = self.report_data.0.entry(contract_name).or_default();
 
-        if let Some(gas) = contract_info.gas_used.checked_add(gas_used) {
-            contract_info.gas_used = gas;
-        }
+        let current_gas = contract_info.gas_used;
+        contract_info.gas_used = current_gas.checked_add(gas_used).unwrap_or_else(|| {
+            panic!("Gas addition overflow when adding {gas_used:?} to {current_gas:?}.")
+        });
 
         let entry = contract_info.functions.entry(selector).or_default();
         entry.records.push(gas_used.l2_gas.0);
--- a/crates/forge-runner/src/gas/report.rs
+++ b/crates/forge-runner/src/gas/report.rs
@@ -10,12 +10,12 @@ type ContractName = String;
 type Selector = String;
 
 #[derive(Debug, Clone)]
-pub struct GasSingleTestInfo {
+pub struct SingleTestGasInfo {
     pub gas_used: GasVector,
     pub report_data: ReportData,
 }
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Default)]
 pub struct ReportData(BTreeMap<ContractName, ContractInfo>);
 
 #[derive(Debug, Clone, Default)]
@@ -31,12 +31,12 @@ pub struct SelectorReportData {
     pub records: Vec<u64>,
 }
 
-impl GasSingleTestInfo {
+impl SingleTestGasInfo {
     #[must_use]
     pub fn new(gas_used: GasVector) -> Self {
         Self {
             gas_used,
-            report_data: ReportData(BTreeMap::new()),
+            report_data: ReportData::default(),
         }
     }
 
--- a/crates/forge-runner/src/gas/report.rs
+++ b/crates/forge-runner/src/gas/report.rs
@@ -1,0 +1,119 @@
+use crate::gas::stats::GasStats;
+use cheatnet::trace_data::{CallTrace, CallTraceNode};
+use debugging::ContractName as DebuggingContractName;
+use debugging::ContractsDataStore;
+use starknet_api::core::{ClassHash, EntryPointSelector};
+use starknet_api::execution_resources::GasVector;
+use std::collections::BTreeMap;
+
+type ContractName = String;
+type Selector = String;
+
+#[derive(Debug, Clone)]
+pub struct GasSingleTestInfo {
+    pub gas_used: GasVector,
+    pub report_data: ReportData,
+}
+
+#[derive(Debug, Clone)]
+pub struct ReportData(BTreeMap<ContractName, ContractInfo>);
+
+#[derive(Debug, Clone, Default)]
+pub struct ContractInfo {
+    pub gas_used: GasVector,
+    pub functions: BTreeMap<Selector, SelectorReportData>,
+}
+
+#[derive(Debug, Clone, Default)]
+pub struct SelectorReportData {
+    pub gas_stats: GasStats,
+    pub n_calls: u64,
+    pub records: Vec<u64>,
+}
+
+impl GasSingleTestInfo {
+    #[must_use]
+    pub fn new(gas_used: GasVector) -> Self {
+        Self {
+            gas_used,
+            report_data: ReportData(BTreeMap::new()),
+        }
+    }
+
+    #[must_use]
+    pub fn new_with_report(
+        gas_used: GasVector,
+        call_trace: &CallTrace,
+        contracts_data: &ContractsDataStore,
+    ) -> Self {
+        Self::new(gas_used).collect_gas_data(call_trace, contracts_data)
+    }
+
+    fn collect_gas_data(mut self, trace: &CallTrace, contracts_data: &ContractsDataStore) -> Self {
+        let mut stack = trace.nested_calls.clone();
+
+        while let Some(call_trace_node) = stack.pop() {
+            if let CallTraceNode::EntryPointCall(call) = call_trace_node {
+                let call = call.borrow();
+                let class_hash = call.entry_point.class_hash.expect(
+                    "class_hash should be set in `fn execute_call_entry_point` in cheatnet",
+                );
+
+                let contract_name = get_contract_name(contracts_data, class_hash);
+                let selector = get_selector(contracts_data, call.entry_point.entry_point_selector);
+                let gas = call
+                    .gas_report_data
+                    .as_ref()
+                    .expect("Gas report data must be updated after test execution")
+                    .get_gas();
+
+                self.update_entry(contract_name, selector, gas);
+                stack.extend(call.nested_calls.clone());
+            }
+        }
+        self.finalize();
+        self
+    }
+
+    fn update_entry(
+        &mut self,
+        contract_name: ContractName,
+        selector: Selector,
+        gas_used: GasVector,
+    ) {
+        let contract_info = self.report_data.0.entry(contract_name).or_default();
+
+        if let Some(gas) = contract_info.gas_used.checked_add(gas_used) {
+            contract_info.gas_used = gas;
+        }
+
+        let entry = contract_info.functions.entry(selector).or_default();
+        entry.records.push(gas_used.l2_gas.0);
+        entry.n_calls += 1;
+    }
+
+    fn finalize(&mut self) {
+        for contract_info in self.report_data.0.values_mut() {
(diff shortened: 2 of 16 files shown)


ddoktorski:

- #3794
    - #3795
        - #3796
            - #3798
                - #3801
                    - #3823
                        - #3840
                            - #3841 ◀
                                - #3856
                                    - #3842
                                        - #3876
                                            - #3879
                                            - #3890

crates/forge-runner/src/gas/report.rs
```diff
@@ -0,0 +1,119 @@
+use crate::gas::stats::GasStats;
+use cheatnet::trace_data::{CallTrace, CallTraceNode};
+use debugging::ContractName as DebuggingContractName;
+use debugging::ContractsDataStore;
+use starknet_api::core::{ClassHash, EntryPointSelector};
+use starknet_api::execution_resources::GasVector;
+use std::collections::BTreeMap;
+
+type ContractName = String;
+type Selector = String;
```
franciszekjob, line 9: nit: I see that our codebase already has a few places in which we have such types - in [cheatnet](https://github.com/foundry-rs/starknet-foundry/blob/e206db9bca74d437e8d7e4e63383ed238d8bfb5b/crates/cheatnet/src/runtime_extensions/forge_runtime_extension/contracts_data.rs#L15) and in [debugging](https://github.com/foundry-rs/starknet-foundry/blob/e206db9bca74d437e8d7e4e63383ed238d8bfb5b/crates/debugging/src/trace/types.rs#L46). Maybe we could somehow share these types?

crates/forge-runner/src/gas/report.rs
```diff
@@ -0,0 +1,119 @@
...
+            }
+        }
+    }
+}
+
+fn get_contract_name(contracts_data: &ContractsDataStore, class_hash: ClassHash) -> ContractName {
+    contracts_data
+        .get_contract_name(&class_hash)
+        .cloned()
+        .unwrap_or_else(|| DebuggingContractName("forked contract".to_string()))
```
franciszekjob: nit:
```suggestion
        .unwrap_or(&DebuggingContractName("forked contract".to_string()))
```

crates/forge-runner/src/gas/report.rs
```diff
@@ -0,0 +1,119 @@
...
+    pub n_calls: u64,
+    pub records: Vec<u64>,
+}
+
+impl GasSingleTestInfo {
+    #[must_use]
+    pub fn new(gas_used: GasVector) -> Self {
+        Self {
+            gas_used,
+            report_data: ReportData(BTreeMap::new()),
```
franciszekjob: nit: Seems more idiomatic (need to derive `Default` on `ReportData`):
```suggestion
            report_data: ReportData::default(),
```

crates/forge-runner/src/gas/report.rs
```diff
@@ -0,0 +1,119 @@
...
+        &mut self,
+        contract_name: ContractName,
+        selector: Selector,
+        gas_used: GasVector,
+    ) {
+        let contract_info = self.report_data.0.entry(contract_name).or_default();
+
+        if let Some(gas) = contract_info.gas_used.checked_add(gas_used) {
+            contract_info.gas_used = gas;
+        }
```
franciszekjob: Potential overflow here is "ignored" - is this intended?

crates/forge/src/scarb/config.rs
```diff
@@ -59,6 +59,9 @@ pub struct ForgeConfigFromScarb {
     /// Generate a coverage report for the executed tests which have passed and are not fuzz tests
     #[serde(default)]
     pub coverage: bool,
+    /// Display a table of L2 gas breakdown for each contract and selector
```
franciszekjob, line 62: nit: If we plan to add l1 gas in future, we can link todo here so we update the helper text.

crates/forge-runner/src/gas/report.rs
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    #[must_use]
+    pub fn new_with_report(
+        gas_used: GasVector,
+        call_trace: &CallTrace,
+        contracts_data: &ContractsDataStore,
+    ) -> Self {
+        Self::new(gas_used).collect_gas_data(call_trace, contracts_data)
+    }
```
franciszekjob: Wdyt about a little refactor here? Seems better to me:

```rust
    #[must_use]
    pub fn new_with_report(
        gas_used: GasVector,
        call_trace: &CallTrace,
        contracts_data: &ContractsDataStore,
    ) -> Self {
        let mut info = Self::new(gas_used);
        info.collect_gas_data(call_trace, contracts_data);
        info.finalize();
        info
    }
```

and we don't call `finalize` in `collect_gas_data`

```rust
fn collect_gas_data(&mut self, trace: &CallTrace, contracts_data: &ContractsDataStore) {
        let mut stack = trace.nested_calls.clone();

        while let Some(call_trace_node) = stack.pop() {
            if let CallTraceNode::EntryPointCall(call) = call_trace_node {
                let call = call.borrow();
                let class_hash = call.entry_point.class_hash.expect(
                    "class_hash should be set in `fn execute_call_entry_point` in cheatnet",
                );

                let contract_name = get_contract_name(contracts_data, class_hash);
                let selector = get_selector(contracts_data, call.entry_point.entry_point_selector);
                let gas = call
                    .gas_report_data
                    .as_ref()
                    .expect("Gas report data must be updated after test execution")
                    .get_gas();

                self.update_entry(contract_name, selector, gas);
                stack.extend(call.nested_calls.clone());
            }
        }
    }
```

crates/forge-runner/src/gas/report.rs
```diff
@@ -0,0 +1,119 @@
...
+use debugging::ContractsDataStore;
+use starknet_api::core::{ClassHash, EntryPointSelector};
+use starknet_api::execution_resources::GasVector;
+use std::collections::BTreeMap;
+
+type ContractName = String;
+type Selector = String;
+
+#[derive(Debug, Clone)]
+pub struct GasSingleTestInfo {
```
cptartur: Super nit

```suggestion
pub struct SingleTesGasInfo {
```

crates/forge-runner/src/gas/report.rs
```diff
@@ -0,0 +1,119 @@
...
+    pub gas_used: GasVector,
+    pub report_data: ReportData,
+}
+
+#[derive(Debug, Clone)]
+pub struct ReportData(BTreeMap<ContractName, ContractInfo>);
```
cptartur, line 18: Do we need ordering for this?

crates/forge-runner/src/gas/report.rs
cptartur: Bit weird to have a public constructor that doesn't have a gas report when `collect_gas_data` is private.

crates/forge-runner/src/gas/report.rs
cptartur, line 18: Is there a point to this wrapper? It's used in a public field, but the inner `BTreeMap<ContractName, ContractInfo>);` is not public. So I understand it cannot be accessed by external modules anyway.

Why don't we just make the fields of `GasSingleTestInfo` not pub?

crates/forge-runner/src/gas/report.rs
cptartur: Maybe it would be better to just panic here? Are we expecting gas overflow in standard use cases?

crates/forge-runner/src/gas/report.rs
```diff
@@ -0,0 +1,119 @@
...
+        entry.n_calls += 1;
+    }
+
+    fn finalize(&mut self) {
+        for contract_info in self.report_data.0.values_mut() {
+            for gas_info in contract_info.functions.values_mut() {
+                gas_info.gas_stats = GasStats::new(&gas_info.records);
+            }
+        }
+    }
```
cptartur, line 103: I'd consider structuring this a bit different. Instead of having this struct technically accessible in non-final state, before finalized is called, I'd make it's internals private and changed the `finalize` method to return a new, different struct (some fields can probably be reused). 

This way it is clear that this one is a "in progress" one and there's a separate struct representing a final result.

crates/forge-runner/src/gas/report.rs
cptartur: This way we always construct `DebuggingContractName`, even if the unwrap succeeded. Here it probably has negligible performance cost, but it's a slight optimization anway, because the closure in `unwrap_or_else` will only execute if unwrap fails.

crates/forge-runner/src/gas/report.rs
ksew1: In debugging we have new types, so it is quite different

crates/forge-runner/src/gas/report.rs
ksew1: You can even do this on the same struct using type state pattern

```rs
pub struct NoReport;
pub struct WithReport;

#[derive(Debug, Clone)]
pub struct GasSingleTestInfo<State> {
    pub gas_used: GasVector,
    pub report_data: ReportData,
    _state: PhantomData<State>,
}


pub fn finalize(self) -> GasSingleTestInfo<WithReport>
```

crates/forge-runner/src/gas/report.rs
```diff
@@ -0,0 +1,119 @@
...
+}
+
+fn get_contract_name(contracts_data: &ContractsDataStore, class_hash: ClassHash) -> ContractName {
+    contracts_data
+        .get_contract_name(&class_hash)
+        .cloned()
+        .unwrap_or_else(|| DebuggingContractName("forked contract".to_string()))
+        .0
+        .clone()
```
ksew1: No need for clone

ksew1: Also here is shorter version without needing `DebuggingContractName` in scope:

```rs
    contracts_data
        .get_contract_name(&class_hash)
        .map_or("forked contract", |name| name.0.as_str())
        .to_string()
```

crates/forge-runner/src/gas/report.rs
ksew1: Deterministic sorted iteration for display i guess

crates/debugging/src/lib.rs
```diff
@@ -7,5 +7,6 @@ mod contracts_data_store;
 mod trace;
 mod tree;
 
+pub use contracts_data_store::ContractsDataStore;
```
ksew1, line 10: It is not a biggie right now, but we are beginning to leak implementation details, and this crate is becoming entangled with something unrelated—in this case, the gas report. Maybe `ContractsData` should contain better information so that we can use it here, as it is already used as the base in `ContractsDataStore`.

ksew1 (commented):
I do rest of review tomorrow as I am getting sleepy 😴

crates/forge-runner/src/gas/report.rs
franciszekjob: Ok, disregard

crates/forge-runner/src/gas/report.rs
ddoktorski: Agree, it should panic here (although it should never happen at this point). Added in https://github.com/foundry-rs/starknet-foundry/pull/3841/commits/300114dec4328d9f9406db1c9dbda1555ba095b8

crates/forge/src/scarb/config.rs
ddoktorski: Not sure if we’ll be adding it at all. Also I don’t think adding a TODO comment at this place is really necessary

crates/forge-runner/src/gas/report.rs
ddoktorski: >I'd consider structuring this a bit different. Instead of having this struct technically accessible in non-final state, before finalized is called

What do you mean by technically accessible in a non-final state? It's possible to modify the fields since they are public, but there's a distinction between `new` and `new_with_report`, both return "finalized" versions

crates/debugging/src/lib.rs
ddoktorski: Would it make sense to move `ContractsDataStore` out of the debugging crate?

ksew1: It is one of the options, but now we would have three structures: `ContractsData`, `ForkData`, and `ContractsDataStore`. `ContractsDataStore` is most complete so maybe we could delete other 2?

crates/forge-runner/src/gas/report.rs
cptartur: I mean it's bit weird for me to share the same struct for gas data with and without raport, where `report_data` is typed as non-optional, and `finalize` method is private, so depending how the struct was constructed, there's no way to move it to a state with gas report non private.

I'd either represent these different states more explicitly, use separate structs or follow the pattern @ksew1 shared.

cptartur (commented):
Looks good, let's agree on and resolve the open comments

crates/forge-runner/src/gas/report.rs
ddoktorski: Main reason of this wrapper is to be able to implement `Display` on it in the next PRs. We also need to have access to `GasSingleTestInfo` fields for example in tests

crates/forge-runner/src/gas/report.rs
ddoktorski: Fair point. However, I think the type state pattern makes this code unnecessarily overcomplicated, so I made the following changes:
* `report_data` in `SingeGasTestInfo` is now an `Option`
* `SingeGasTestInfo` has only two public methods `new` and `get_with_report_data` (which consumes `self`)
* other helper methods like `finalize` are implemented on `ReportData`

Let me know what you think.

crates/forge-runner/src/gas/report.rs
ddoktorski: Addressed in https://github.com/foundry-rs/starknet-foundry/pull/3841/commits/959f1e4ec788f561d7597bce68d4a995d3e86cf0

crates/forge-runner/src/gas/report.rs
ddoktorski: It was refactored slightly different in https://github.com/foundry-rs/starknet-foundry/pull/3841/commits/959f1e4ec788f561d7597bce68d4a995d3e86cf0

crates/forge-runner/src/gas/report.rs
cptartur: Okay, I see the point of this.

crates/forge-runner/src/gas/report.rs
cptartur: Works for me

cptartur (approved):

franciszekjob (approved):

crates/debugging/src/lib.rs
ddoktorski: https://github.com/foundry-rs/starknet-foundry/issues/3905 🙏

ksew1 (approved):

PR #3848, opened 2025-10-22
Resolve #2074

cptartur:
Closes #2074

---

**Stack**:
- #3926
- #3925
- #3850
- #3849
- #3848 ⬅
- #3847

⚠️ *Part of a stack created by [spr](https://github.com/ejoffe/spr). Do not merge manually using the UI - doing so may have unexpected results.*

PR #3847, opened 2025-10-22
Add oracles test

cptartur:
Closes #3825

---

**Stack**:
- #3926
- #3925
- #3850
- #3849
- #3848
- #3847 ⬅

⚠️ *Part of a stack created by [spr](https://github.com/ejoffe/spr). Do not merge manually using the UI - doing so may have unexpected results.*

PR #3849, opened 2025-10-22
Refactor `RawFeltVec`

cptartur:
**Stack**:
- #3926
- #3925
- #3850
- #3849 ⬅
- #3848
- #3847


crates/conversions/src/serde/no_length_felt_vec.rs
```diff
@@ -0,0 +1,41 @@
+use conversions::serde::deserialize::{BufferReadResult, BufferReader, CairoDeserialize};
+use conversions::serde::serialize::{BufferWriter, CairoSerialize};
+use starknet_types_core::felt::Felt;
+
+/// use this wrapper to NOT add extra length felt
+/// useful e.g. when you need to pass an already serialized value
+#[derive(Debug)]
+pub struct NoLengthFeltVec<T>(pub Vec<T>)
```
ksew1: Maybe we could name it `SerializedValue`? As this i think is only usefull if value is serialized

cptartur: Renaming

crates/conversions/src/serde/deserialize/deserialize_impl.rs
```diff
@@ -53,6 +53,19 @@ where
     }
 }
 
+impl<T: CairoDeserialize, E: CairoDeserialize> CairoDeserialize for Result<T, E> {
+    fn deserialize(reader: &mut BufferReader<'_>) -> BufferReadResult<Self> {
+        let variant: Felt = reader.read()?;
+        let variant: usize = variant.to_usize().ok_or(BufferReadError::ParseFailed)?;
```
franciszekjob: nit:
```suggestion
        let variant = variant.to_usize().ok_or(BufferReadError::ParseFailed)?;
```

franciszekjob (approved):

crates/conversions/src/serde/serialized_value.rs
```diff
@@ -0,0 +1,42 @@
+use conversions::serde::deserialize::{BufferReadResult, BufferReader, CairoDeserialize};
+use conversions::serde::serialize::{BufferWriter, CairoSerialize};
+use starknet_types_core::felt::Felt;
+
+/// Represents an already serialized Vec of values.
+///
+/// Use this to omit adding extra felt for the length of the vector during serialization.
+#[derive(Debug)]
+pub struct SerializedValue<T>(pub Vec<T>)
```
ksew1, line 9: I would also drop generic. Only valid representation i would say is `pub struct SerializedValue(pub Vec<Felt>)`

crates/conversions/src/serde/serialized_value.rs
```diff
@@ -0,0 +1,42 @@
...
+    T: CairoSerialize,
+{
+    fn serialize(&self, output: &mut BufferWriter) {
+        for e in &self.0 {
+            e.serialize(output);
+        }
+    }
+}
+
+impl CairoDeserialize for SerializedValue<Felt> {
```
ksew1, line 34: Is this needed?

crates/conversions/src/serde/serialized_value.rs
cptartur: There's one use case where we want to write serialized array of `ResourceBounds` structs into memory in serialized format and we do it as so 

```rust
&SerializedValue::new(resource_bounds).serialize_to_vec(),
```

We can change serialized value to be a `Vec<Felt>` but we would either have to provide an constructor that allows creating it from other values that implement serialize (we would have to do the serialization at construction time though) or change the usage to something along the lines of

```rust
&resource_bounds
                .iter()
                .flat_map(conversions::serde::serialize::SerializeToFeltVec::serialize_to_vec)
                .collect::<Vec<_>>(),
```

which kind of duplicates the logic.

crates/conversions/src/serde/serialized_value.rs
cptartur: Yes, we need it to `.read()` the already serialized data that oracle returns.

ksew1 (approved):

cptartur:

PR #3850, opened 2025-10-22
Integrate oracles result handling

cptartur:
**Stack**:
- #3926
- #3925
- #3850 ⬅
- #3849
- #3848
- #3847

### DIFF
--- a/crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
+++ b/crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
@@ -35,8 +35,9 @@ use cairo_vm::vm::{
 };
 use conversions::byte_array::ByteArray;
 use conversions::felt::{ToShortString, TryInferFormat};
+use conversions::serde::SerializedValue;
 use conversions::serde::deserialize::BufferReader;
-use conversions::serde::serialize::CairoSerialize;
+use conversions::serde::serialize::{CairoSerialize, SerializeToFeltVec};
 use data_transformer::cairo_types::CairoU256;
 use rand::prelude::StdRng;
 use runtime::{
@@ -97,7 +98,13 @@ impl<'a> ExtensionLogic for ForgeExtension<'a> {
             let output = self
                 .oracle_hint_service
                 .execute_cheatcode(oracle_selector, input_reader.into_remaining());
-            return Ok(CheatcodeHandlingResult::Handled(output));
+            let mut reader = BufferReader::new(&output);
+            let deserialized: Result<SerializedValue<Felt>, ByteArray> = reader.read()?;
+            let converted = deserialized
+                .map_err(|error| EnhancedHintError::OracleError { error })
+                .map(|r| r.serialize_to_vec())?;
+
+            return Ok(CheatcodeHandlingResult::Handled(converted));
         }
 
         match selector {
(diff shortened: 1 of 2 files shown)


crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
```diff
@@ -97,7 +98,13 @@ impl<'a> ExtensionLogic for ForgeExtension<'a> {
             let output = self
                 .oracle_hint_service
                 .execute_cheatcode(oracle_selector, input_reader.into_remaining());
-            return Ok(CheatcodeHandlingResult::Handled(output));
+            let mut reader = BufferReader::new(&output);
+            let deserialized: Result<NoLengthFeltVec<Felt>, ByteArray> = reader.read()?;
+            let converted = deserialized
+                .map_err(|e| anyhow!("Oracle error: {e}"))
```
piotmag769: I don't think you should change the error of the oracle - this goes into the Cairo memory

cptartur: Changed that

crates/runtime/src/lib.rs
```diff
@@ -470,6 +470,8 @@ pub enum EnhancedHintError {
     StarknetApi(#[from] StarknetApiError),
     #[error("Failed to parse {path} file")]
     FileParsing { path: String },
+    #[error("{error}")]
```
piotmag769, line 473: Is the conversion chain `ByteArray -> String -> ByteArray` a noop? It would be better to just have `impl From<EnhancedHintError> for ByteArray` at this point imo since this is what happens [here](https://github.com/foundry-rs/starknet-foundry/blob/cb6778adf662c910b9db19f6bd38bbc0229b7cec/crates/runtime/src/lib.rs#L290)

cptartur: It's noop, or at least it should be. I'll try what you've suggested

cptartur: The `Error` (from thiserror and std) trait requires us to implement `Display` so we need to have this. The conversion we do not is a bit indirect. As an alternative we could get rid of the `Error` trait here and have it as a standard enum but not sure if that would be more maintainable if what we have now. And we would have to handle constructing from inner errors ourselves.

franciszekjob (approved):

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
```diff
@@ -97,7 +98,13 @@ impl<'a> ExtensionLogic for ForgeExtension<'a> {
             let output = self
                 .oracle_hint_service
                 .execute_cheatcode(oracle_selector, input_reader.into_remaining());
-            return Ok(CheatcodeHandlingResult::Handled(output));
+            let mut reader = BufferReader::new(&output);
+            let deserialized: Result<SerializedValue<Felt>, ByteArray> = reader.read()?;
+            let converted = deserialized
+                .map_err(|error| EnhancedHintError::OracleError { error })
+                .map(|r| r.serialize_to_vec());
```
ksew1: Nit:
```suggestion
                .map(|r| r.serialize_to_vec())?;
```

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
```diff
@@ -97,7 +98,13 @@ impl<'a> ExtensionLogic for ForgeExtension<'a> {
...
+
+            return Ok(CheatcodeHandlingResult::Handled(converted?));
```
ksew1:
```suggestion
            return Ok(CheatcodeHandlingResult::Handled(converted));
```

ksew1 (approved):

cptartur:

Issue #3548, opened 2025-07-11
Support test partitioning


cptartur:
### Current State

For users running tests in CI it is possible for them to run for a very long time i.e. when using a lot of fuzz tests. Currently there is no straightforward approach for partitioning tests between different processes. 

### Objective

Support partitioning tests and running them through separate processes

### Additional Context

Create some function/new tool similar to cargo nextest https://nexte.st/

PR #3852, opened 2025-10-22
Add `--partition` flag

franciszekjob:
Towards #3548

**Stack**:
- #3855
- #3854
- #3853
- #3852 ⬅

## Introduced changes

Add `--partition` flag to `snforge`

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/forge/src/partition.rs
```diff
@@ -0,0 +1,111 @@
+use serde::Serialize;
+use std::{collections::HashMap, str::FromStr};
+
+#[derive(Debug, Clone, Copy, Serialize)]
+pub struct Partition {
+    index: usize,
+    total: usize,
+}
+
+#[allow(dead_code)] // TODO: Removed in later PRs
```
cptartur: Doesn't really matter here but except is safer than allow because it will fail if the condition (dead_code in this case) is not met

```suggestion
#[except(dead_code)] // TODO: Removed in later PRs
```

crates/forge/src/partition.rs
```diff
@@ -0,0 +1,111 @@
...
+    #[must_use]
+    pub fn index(&self) -> usize {
+        self.index
+    }
+
+    #[must_use]
+    pub fn total(&self) -> usize {
+        self.total
+    }
+}
```
cptartur: We could allow just accessing this by making these pub.

You can disallow creating struct instances with pub fields by adding this attribute to it https://doc.rust-lang.org/reference/attributes/type_system.html#the-non_exhaustive-attribute if that's what you wanted to prevent.

cptartur (approved):

crates/forge/src/partition.rs
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    #[test]
+    fn test_invalid_format() {
+        let err = "2-5".parse::<Partition>().unwrap_err();
+        assert_eq!(err, "Partition must be in the format <INDEX>/<TOTAL>");
+    }
+
+    // #[test]
+    #[test_case("-1/5", "INDEX")]
```
MKowalski8, line 66: Leftover?
```suggestion
    #[test_case("-1/5", "INDEX")]
```

MKowalski8 (approved):

PR #3859, opened 2025-10-23
Refactor universal-sierra-compiler-api

ksew1:
commit-id:e0444c34

---

**Stack**:
- #3861
- #3860
- #3859 ⬅


crates/universal-sierra-compiler-api/src/command.rs
```diff
@@ -1,88 +1,71 @@
...
-            .context("Cannot find `universal-sierra-compiler` binary. \
-                      Make sure you have USC installed https://github.com/software-mansion/universal-sierra-compiler \
-                      and added to PATH (or set at UNIVERSAL_SIERRA_COMPILER env var)"
-            )?;
-        Ok(())
-    }
+    #[error(
+        "Error while compiling Sierra. \
+             Make sure you have the latest universal-sierra-compiler binary installed. \
+             Contact us if it doesn't help."
```
cptartur:
```suggestion
             Contact Starknet Foundry team through Github or Telegram if it doesn't help."
```

crates/universal-sierra-compiler-api/src/command.rs
```diff
@@ -1,88 +1,71 @@
...
-    #[must_use]
-    pub fn new() -> Self {
-        Self::default()
-    }
+/// Errors that can occur while working with `universal-sierra-compiler` command.
+#[derive(Debug, Error)]
+pub enum USCError {
+    #[error(
+        "`universal-sierra-compiler` binary not available. \
+             Make sure it is installed and available in PATH or set via UNIVERSAL_SIERRA_COMPILER."
```
cptartur: I think it was actually good to have a repo url in the error, let's maybe restore it.

crates/universal-sierra-compiler-api/src/compile.rs
```diff
@@ -0,0 +1,61 @@
...
+use thiserror::Error;
+
+/// Errors that can occur during Sierra compilation.
+#[derive(Debug, Error)]
+pub enum CompilationError {
+    #[error("Failed to write Sierra JSON to temp file: {0}")]
+    TempFileWrite(#[from] io::Error),
+
+    #[error("Could not serialize Sierra JSON")]
+    Serialization(serde_json::Error),
```
cptartur, line 17: Why aren't we generating `From` for serde_json errors?

crates/universal-sierra-compiler-api/src/compile.rs
```diff
@@ -0,0 +1,61 @@
...
+    let utf8_path =
+        Utf8Path::from_path(temp_sierra_file.path()).expect("temp file path is valid UTF-8");
+
+    compile_sierra_at_path(utf8_path, sierra_type)
+}
+
+/// Compiles the Sierra file at the given path into the specified type using the `universal-sierra-compiler`.
+#[tracing::instrument(skip_all, level = "debug")]
+pub fn compile_sierra_at_path(
+    sierra_file_path: &Utf8Path,
```
cptartur: Do we need this to be `Utf8Path`? We aren't printing it to the user, so it could just be `Path` and we could use `to_os_str()` instead of `to_str()`.

This way we won't have to create utf8 path in `compile_sierra` just to pass it here.

Or we could even make it more general, make sierra_file_path be `T: AsRef<Path>`

crates/universal-sierra-compiler-api/src/representation.rs
```diff
@@ -0,0 +1,25 @@
...
+    /// `StatementIdx(i)` (i-th index in Sierra statements vector):
+    /// - code offset in the CASM bytecode
+    /// - index in CASM instructions vector
+    ///
+    /// Those 2 values are usually not equal since the instruction sizes in CASM may vary
+    pub debug_info: Vec<(CasmCodeOffset, CasmInstructionIdx)>,
+}
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
+pub struct AssembledCairoProgramWithSerde {
```
cptartur: Couldn't it be renamed? I'm not sure what `WithSerde` was supposed to represent.

```suggestion
pub struct AssembledCairoProgram {
```

crates/universal-sierra-compiler-api/src/compile.rs
ksew1: Cause you can't as serialization and de serialization base on the same error

crates/universal-sierra-compiler-api/src/compile.rs
ksew1: Went with `Path` as i don't what overhead of monomorphization

cptartur (approved):

franciszekjob (approved):

PR #3860, opened 2025-10-23
Refactor metadata module

ksew1:
commit-id:a2fe82e0

---

**Stack**:
- #3861
- #3860 ⬅

### DIFF
--- a/crates/scarb-api/src/metadata.rs
+++ b/crates/scarb-api/src/metadata.rs
@@ -1,14 +1,72 @@
-use anyhow::{Context, Result};
-pub use scarb_metadata::{Metadata, MetadataCommand, MetadataCommandError, PackageMetadata};
+//! Functionality for fetching `scarb` metadata.
+//!
+//! # Note:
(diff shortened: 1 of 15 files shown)


ddoktorski (approved):

franciszekjob (approved):

crates/scarb-api/src/metadata.rs
cptartur, line 1: I like these improvements but I wonder if some kind of builder for metadata command wouldn't be more extensible? 
The current approach is purpose built for 3 different use-cases and maybe that's enough improvements, but I wonder what you think.

cc @franciszekjob @ddoktorski

ksew1: The idea behind this is to discourage the use of too many options. The issue with the previous solution was that too much freedom resulted in metadata calls that were not consistent.

cptartur: Okay, if the rest also disagrees with generalizing this, just disregard my comment.

PR #3861, opened 2025-10-23
Refactor version module

ksew1:
commit-id:f673da00

### DIFF
--- a/crates/scarb-api/src/version.rs
+++ b/crates/scarb-api/src/version.rs
@@ -1,71 +1,168 @@
-use crate::ScarbCommand;
-use anyhow::{Context, Result};
+//! Functionality for fetching and parsing `scarb` version information.
+
+use crate::{ScarbCommand, ScarbUnavailableError, ensure_scarb_available};
 use regex::Regex;
 use semver::Version;
-use shared::command::CommandExt;
-use std::path::PathBuf;
-use std::str::from_utf8;
+use shared::command::{CommandError, CommandExt};
+use std::collections::HashMap;
+use std::path::Path;
+use std::process::Output;
+use std::sync::LazyLock;
+use thiserror::Error;
 
+#[derive(Debug)]
 pub struct ScarbVersionOutput {
     pub scarb: Version,
     pub cairo: Version,
+    pub sierra: Version,
 }
 
-#[derive(Clone, Debug, Default)]
-pub struct VersionCommand {
-    current_dir: Option<PathBuf>,
+/// Errors that can occur when fetching `scarb` version information.
+#[derive(Error, Debug)]
+pub enum VersionError {
+    #[error(transparent)]
+    ScarbNotFound(#[from] ScarbUnavailableError),
+
+    #[error("Failed to execute `scarb --version`: {0}")]
+    CommandExecutionError(#[from] CommandError),
+
+    #[error("Could not parse version for tool `{0}`: {1}")]
+    VersionParseError(String, #[source] semver::Error),
+
+    #[error("Missing version entry for `{0}`")]
+    MissingToolVersion(String),
 }
 
-impl VersionCommand {
-    #[must_use]
-    pub fn new() -> Self {
-        Self { current_dir: None }
+/// Fetches `scarb` version information for the current directory.
+pub fn scarb_version() -> Result<ScarbVersionOutput, VersionError> {
+    scarb_version_internal(None)
+}
+
+/// Fetches `scarb` version information for a specific directory.
+pub fn scarb_version_for_dir(dir: impl AsRef<Path>) -> Result<ScarbVersionOutput, VersionError> {
+    scarb_version_internal(Some(dir.as_ref()))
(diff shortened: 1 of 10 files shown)


crates/scarb-api/src/version.rs
```diff
@@ -1,71 +1,118 @@
...
-        Ok(ScarbVersionOutput {
-            scarb: Self::extract_version(version_output, "scarb")?,
-            cairo: Self::extract_version(version_output, "cairo:")?,
-        })
-    }
+/// Extracts tool versions from the version output string.
+fn extract_versions(version_output: &str) -> Result<HashMap<String, Version>, VersionError> {
+    // https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string
+    static VERSION_REGEX: LazyLock<Regex> = LazyLock::new(|| {
+        Regex::new(r"(?P<tool>\w+)[\s:]+(?P<ver>(?:0|[1-9]\d*)\.(?:0|[1-9]\d*)\.(?:0|[1-9]\d*)(?:-(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*)?(?:\+[0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*)?)").expect("this should be a valid regex")
```
cptartur: We don't need to mach unlimited number of whitespace between tool name and it's version, just an optional `:` and whitespace

```suggestion
        Regex::new(r"(?P<tool>\w+):?\s(?P<ver>(?:0|[1-9]\d*)\.(?:0|[1-9]\d*)\.(?:0|[1-9]\d*)(?:-(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*)?(?:\+[0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*)?)").expect("this should be a valid regex")
```

Also I think if you are already matching for digits, you don't need extra matches for numbers

```suggestion
        Regex::new(r"(?P<tool>\w+)[\s:]+(?P<ver>(?:\d*)\.(?:\d*)\.(?:\d*)(?:-(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*)?(?:\+[0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*)?)").expect("this should be a valid regex")
```

cptartur: The other groups could be simplified as well

```suggestion
        Regex::new(r"(?P<tool>\w+):?\s(?P<ver>(?:\d*)\.(?:\d*)\.(?:\d*)(?:-(?:[0-9a-zA-Z-]+)(?:\.(?:[0-9a-zA-Z-]*))*)?(?:\+[0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*)?)").expect("this should be a valid regex")
```

crates/scarb-api/src/version.rs
```diff
@@ -1,71 +1,118 @@
...
-            .context(format!("Could not find {tool} version"))?
-            .as_str();
+            Version::parse(ver_str)
+                .map_err(|e| VersionError::VersionParseError(tool.clone(), e))
+                .map(|version| (tool, version))
+        })
+        .collect()
+}
+#[cfg(test)]
+mod tests {
```
cptartur, line 111: Shouldn't we add some more tests (even generate them through some LLM) for more versions combinations? As a sanity check for regexes.

ksew1: Added

crates/scarb-api/src/version.rs
```diff
@@ -1,71 +1,168 @@
...
+
+/// Fetches `scarb` version information for a specific directory.
+pub fn scarb_version_for_dir(dir: impl AsRef<Path>) -> Result<ScarbVersionOutput, VersionError> {
+    scarb_version_internal(Some(dir.as_ref()))
+}
+
+/// Internal function to fetch `scarb` version information.
+fn scarb_version_internal(dir: Option<&Path>) -> Result<ScarbVersionOutput, VersionError> {
+    let output = run_command(dir)?;
+    parse_output(&output)
```
MKowalski8: nit: What would you say to name it like this, or simillar:
```suggestion
    let output = run_version_command(dir)?;
    parse_version_output(&output)
```
I know that it is a private function, but it would take away this few seconds to compute, that it is not some generic `run_command` function 😅

cptartur: Sounds good to me

cptartur (approved):

MKowalski8 (approved):

PR #3868, opened 2025-10-23
Add `get_current_vm_step`

### LINKED ISSUE #3824: Function allowing to assert steps
### Select Tool

snforge

### Requested Feature

ATM, snforge allows to configure number of maximum steps for an entire test, thanks to `--max-n-steps` flag. However, this limit applies globally to the whole test execution, making it impossible to assert or constrain the number of steps taken during specific parts of the test (e.g., after setup or between critical operations).

Example usage
```cairo
fn setup() {
    // ...
}

#[test]
fn test_xyz() {
    setup();

    assert_max_n_steps(10_000);

    // ...
}
```

franciszekjob:
Closes #3824

## Introduced changes

Introduce `get_current_vm_step` as part of `snforge_std::testing`

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
```diff
@@ -552,6 +553,14 @@ impl<'a> ExtensionLogic for ForgeExtension<'a> {
                     .cheat_block_hash(block_number, operation);
                 Ok(CheatcodeHandlingResult::from_serializable(()))
             }
+            "get_current_step" => {
+                let current_step: u32 = vm
```
ksew1: Why not just use `usize` instead of casting?

crates/forge/tests/data/contracts/meta_tx_v0_checkers.cairo
ksew1, line 60: No rebase? It think this doesn't belong to this PR

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
```diff
@@ -81,6 +81,7 @@ impl<'a> ExtensionLogic for ForgeExtension<'a> {
         selector: &str,
         mut input_reader: BufferReader<'_>,
         extended_runtime: &mut Self::Runtime,
+        vm: &mut VirtualMachine,
```
ksew1: We don't need this mut

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
franciszekjob: I used `u32` cause our flag `--max-n-steps` assumes that number of steps cannot exceed this type. However, `get_current_step` from vm returns `usize` so maybe it's better to use this type. Changed.

ksew1 (approved):

ddoktorski (commented):
In the current implementation this doesn't actually work for contracts since each contract call is executed in a separate VM. Was it intended to support contract testing at all?

cptartur (commented):
Please re-request after changes from slack discussion are addressed

franciszekjob:
> In the current implementation this doesn't actually work for contracts since each contract call is executed in a separate VM. Was it intended to support contract testing at all?

Good catch! I've changed the code so now it also adds steps from calls's resource, should be fine. Please take a look.

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
```diff
@@ -552,6 +554,25 @@ impl<'a> ExtensionLogic for ForgeExtension<'a> {
...
+                    .cheatnet_state
+                    .used_resources;
+                let execution_resources_from_used_syscalls =
+                    &VersionedConstants::latest_constants().get_additional_os_syscall_resources(
+                        &extended_runtime
+                            .extended_runtime
+                            .extension
+                            .cheatnet_state
+                            .used_syscalls,
+                    );
```
ddoktorski: Would it be possible to use `extended_runtime.extended_runtime.extended_runtime.hint_handler.base.syscalls_usage` minus the syscalls used in the `CallTrace` instead of adding a new field to the cheatnet state?

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
```diff
@@ -552,6 +554,25 @@ impl<'a> ExtensionLogic for ForgeExtension<'a> {
                     .cheat_block_hash(block_number, operation);
                 Ok(CheatcodeHandlingResult::from_serializable(()))
             }
+            "get_current_step" => {
+                let used_resources = &extended_runtime
+                    .extended_runtime
+                    .extension
+                    .cheatnet_state
+                    .used_resources;
```
ddoktorski: I think as suggested by @cptartur during the call, we should rely on the sum of inner call resources from the `CallTrace` to get this value

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
franciszekjob: `extended_runtime.extended_runtime.extended_runtime.hint_handler.base.syscalls_usage` includes **only** syscalls from top call (i.e. from test level), so subtracting here won't be valid.

However, we can get rid of extra field in cheatnet state and calculate syscalls as the sum of `extended_runtime.extended_runtime.extended_runtime.hint_handler.base.syscalls_usage` and syscalls from `CallTrace` inner calls.

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
franciszekjob: Good idea - no new field in cheatnet state.

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/cheatcodes/testing.rs
```diff
@@ -0,0 +1,52 @@
...
+            CallTraceNode::DeployWithoutConstructor => acc,
+        },
+    );
+
+    let total_syscalls = sum_syscall_usage(inner_calls_syscalls.clone(), top_call_syscalls);
+
+    let total_syscalls_exeucution_resources = &VersionedConstants::latest_constants()
+        .get_additional_os_syscall_resources(&total_syscalls);
+
+    let resources_from_calls = used_resources + total_syscalls_exeucution_resources;
```
cptartur: I don't think this step is required. `used_execution_resources` for inner calls should already be finalized and contain syscall costs there. 

I think adding the costs of inner_calls_syscalls will inflate the cost over the real one.

So we should just be able to add resources from syscalls in the top call and that's it, no?

cc @ddoktorski 


```suggestion
    let used_resources =
        &top_call
            .borrow()
            .nested_calls
            .iter()
            .fold(ExecutionResources::default(), |acc, node| match node {
                CallTraceNode::EntryPointCall(call_trace) => {
                    &acc + &call_trace.borrow().used_execution_resources
                }
                CallTraceNode::DeployWithoutConstructor => acc,
            });

    let total_syscalls_exeucution_resources = &VersionedConstants::latest_constants()
        .get_additional_os_syscall_resources(&top_call_syscalls);

    let resources_from_calls = used_resources + total_syscalls_exeucution_resources;

    resources_from_calls.n_steps
```

cptartur: Side note, if this is true, I'd rather move this logic directly to the cheatcode handling (or have it as a free function in module there) so someone doesn't try to reuse this logic. It is specific for this one cheatcode.

franciszekjob: > `used_execution_resources` from inner calls should already be finalized and contain syscall costs there.

Ok, you're right. Changed.

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
```diff
@@ -552,6 +554,27 @@ impl<'a> ExtensionLogic for ForgeExtension<'a> {
...
+
+                let top_call_syscalls = &extended_runtime
+                    .extended_runtime
+                    .extended_runtime
+                    .hint_handler
+                    .base
+                    .syscalls_usage;
+
+                let steps_from_calls = calculate_steps_from_calls(&top_call, top_call_syscalls);
+                let total_steps = steps_from_calls + vm.get_current_step();
```
ddoktorski: It would be better to avoid mixing syscall steps with contract calls. Let's structure the code in a way that clearly separates these calculations for better clarity
```suggestion
                let top_call = extended_runtime
                    .extended_runtime
                    .extension
                    .cheatnet_state
                    .trace_data
                    .current_call_stack
                    .top();
                let steps_from_calls = get_steps_from_calls(&top_call);

                let top_call_syscalls = &extended_runtime
                    .extended_runtime
                    .extended_runtime
                    .hint_handler
                    .base
                    .syscalls_usage;
                let syscall_steps = &VersionedConstants::latest_constants()
                    .get_additional_os_syscall_resources(top_call_syscalls)
                    .n_steps;

                let total_steps = steps_from_calls + syscall_steps + vm.get_current_step();
```

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
```diff
@@ -552,6 +554,27 @@ impl<'a> ExtensionLogic for ForgeExtension<'a> {
                     .cheat_block_hash(block_number, operation);
                 Ok(CheatcodeHandlingResult::from_serializable(()))
             }
+            "get_current_step" => {
```
ddoktorski:
```suggestion
            "get_current_vm_step" => {
```

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
ddoktorski: > extended_runtime.extended_runtime.extended_runtime.hint_handler.base.syscalls_usage includes only syscalls from top call (i.e. from test level)

That's even better as this is the value that we want to get

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
```diff
@@ -858,3 +883,26 @@ pub fn get_all_used_resources(
         l1_handler_payload_lengths,
     }
 }
+
+fn calculate_vm_steps_from_calls(
+    top_call: &Rc<RefCell<CallTrace>>,
+    // top_call_syscalls: &SyscallUsageMap,
```
ddoktorski:
```suggestion
```

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
```diff
@@ -858,3 +883,26 @@ pub fn get_all_used_resources(
...
+            .iter()
+            .fold(ExecutionResources::default(), |acc, node| match node {
+                CallTraceNode::EntryPointCall(call_trace) => {
+                    &acc + &call_trace.borrow().used_execution_resources
+                }
+                CallTraceNode::DeployWithoutConstructor => acc,
+            });
+    // let total_syscalls_exeucution_resources = &VersionedConstants::latest_constants()
+    //     .get_additional_os_syscall_resources(&top_call_syscalls);
+    // let resources_from_calls = used_resources + total_syscalls_exeucution_resources;
```
ddoktorski:
```suggestion
```

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
```diff
@@ -552,6 +554,29 @@ impl<'a> ExtensionLogic for ForgeExtension<'a> {
                     .cheat_block_hash(block_number, operation);
                 Ok(CheatcodeHandlingResult::from_serializable(()))
             }
+            "get_current_vm_step" => {
+                let top_call = extended_runtime
```
ddoktorski, line 559: nit: We can consider adding a short comment explaining why it's necessary to add these extra steps

ddoktorski (approved):

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
```diff
@@ -552,6 +554,33 @@ impl<'a> ExtensionLogic for ForgeExtension<'a> {
...
+                    .extended_runtime
+                    .extended_runtime
+                    .hint_handler
+                    .base
+                    .syscalls_usage;
+                let vm_steps_from_syscalls = &VersionedConstants::latest_constants()
+                    .get_additional_os_syscall_resources(top_call_syscalls)
+                    .n_steps;
+                let total_vm_steps =
+                    vm_steps_from_calls + vm_steps_from_syscalls + vm.get_current_step();
```
cptartur: Or `nested`

```suggestion
                    vm_steps_from_inner_calls + vm_steps_from_syscalls + vm.get_current_step();
```

crates/forge/tests/integration/get_current_vm_step.rs
cptartur, line 1: Can we add some kind of sanity check where we assert specific value of used steps (or a delta of steps)? So we detect eventually regressions if these change significantly.

cptartur: I see that we have a test below that does this for contracts but let's have one for top level calls as well.

cptartur, line 1: Also, let's add a test where we have inner calls and these inner calls do some expensive computation (not just increment some counter, but call some more expensive syscalls). Check if steps value is "large enough" there.

cptartur: We should be able to roughly estimate the steps value from fee mechanism docs https://docs.starknet.io/learn/protocol/fees#vm-resources similarly as we do in our gas integration tests and check if values in the test are roughly withing the expected range.

docs/src/appendix/snforge-library/testing/get_current_vm_step.md
cptartur, line 1: I'd reword this so it's more of a reference than a tutorial, explanations like that fit more in the `snforge Overview` section and we can assume users looking at the reference are familiar with these docs.

Take a look at fuzzable docs for example, let's show an reference for what `get_current_vm_step` shows and does and keep a simple example without too much explanations.

crates/forge/tests/integration/get_current_vm_step.rs
franciszekjob: I've extended `test_get_current_vm_step` so now it checks steps after two possible scenarios: syscalls only in top call + syscalls both in top call and inner calls. It should cover this and the comment below, pls TAL.

franciszekjob: Responded in comment above.

docs/src/appendix/snforge-library/testing/get_current_vm_step.md
franciszekjob: > I'd reword this so it's more of a reference than a tutorial, explanations like that fit more in the `snforge Overview` section and we can assume users looking at the reference are familiar with these docs.
> 
> Take a look at fuzzable docs for example, let's show an reference for what `get_current_vm_step` shows and does and keep a simple example without too much explanations.

It's not shown in `snforge Overview`, hence the example here has more detailed explanation.

I've reduced the textual description, but kept contract and test code (https://github.com/foundry-rs/starknet-foundry/pull/3868/commits/13f1af4f02ad6fb2d399b62818b0e6b7b4aad7bc).

We can move this little tutorial under `snforge Advanced Features` as `Testing Utils`?

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
```diff
@@ -858,3 +887,19 @@ pub fn get_all_used_resources(
...
+        &top_call
+            .borrow()
+            .nested_calls
+            .iter()
+            .fold(ExecutionResources::default(), |acc, node| match node {
+                CallTraceNode::EntryPointCall(call_trace) => {
+                    &acc + &call_trace.borrow().used_execution_resources
+                }
+                CallTraceNode::DeployWithoutConstructor => acc,
+            });
```
cptartur: I'd just use a `usize` as accumulator here, don't want to rely on the semantics of adding execution resources if we don't need that here.

crates/forge/tests/integration/get_current_vm_step.rs
```diff
@@ -0,0 +1,94 @@
...
+
+            const STEPS_MARGIN: u32 = 100;
+
+            // 1173 = cost of 1 deploy syscall without calldata
+            const DEPLOY_SYSCALL_STEPS: u32 = 1173;
+
+            // 903 = steps of 1 call contract syscall
+            const CALL_CONTRACT_SYSCALL_STEPS: u32 = 903;
+
+            // 90 = steps of 1 call contract syscall
```
cptartur: how did we obtain these values? I'd consider including a reference here.

crates/forge/tests/integration/get_current_vm_step.rs
```diff
@@ -0,0 +1,94 @@
...
+                    );
+
+                // Sycalls between step_b and step_c:
+                // top call: 1 x call contract syscall
+                // inner calls: 1 x storage read syscall, 1 x call contract syscall
+                let step_c = get_current_vm_step();
+
+                let expected_steps_taken = 2 * CALL_CONTRACT_SYSCALL_STEPS
+                    + 1 * STORAGE_READ_SYSCALL_STEPS
+                    + 277; // 277 are steps from VM
```
cptartur, line 66: These are likely to change on each cairo bump, right?

docs/src/appendix/snforge-library/testing/get_current_vm_step.md
cptartur: I'd prefer if we just kept a simpler reference in the appendix, and I think we can showcase the usage of `get_current_vm_step` sufficiently with just a snippet, without showing a whole contract code.

We already have a tutorial on using cheatcodes in general, don't think we need a whole tutorial for each cheatcode, especially one that is quite straightforward in terms of the usage.

snforge_std/src/testing.cairo
```diff
@@ -0,0 +1,6 @@
+use crate::cheatcode::execute_cheatcode_and_deserialize;
+
+/// Gets the current step from Cairo VM during test execution
```
cptartur: Maybe let's omit "from Cairo VM" since technically we are combining values from multiple places (steps from nested vms/nested calls as well). I'd just say that it gets current step during test execution.

crates/forge/tests/integration/get_current_vm_step.rs
franciszekjob: > how did we obtain these values?

From versioned constants

> I'd consider including a reference here.

Added permalinks - https://github.com/foundry-rs/starknet-foundry/pull/3868/commits/eeaf9582c4abb1152f1806bf4fe902b608fbc597

crates/forge/tests/integration/get_current_vm_step.rs
franciszekjob: Probably, that's why I've added margin.

snforge_std/src/testing.cairo
franciszekjob: https://github.com/foundry-rs/starknet-foundry/pull/3868/commits/73f6048a52a0d7b56eff2c0281f1e99849c019e8

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
franciszekjob: Right, that's better - https://github.com/foundry-rs/starknet-foundry/pull/3868/commits/a04e162ac0ea750cc013eaa6da354d266af6e606

docs/src/appendix/snforge-library/testing/get_current_vm_step.md
franciszekjob: Ok, I removed contract code and simplified test code, should be better.

docs/src/appendix/snforge-library/testing/get_current_vm_step.md
```diff
@@ -0,0 +1,36 @@
...
+Collected 1 test(s) from testing_reference package
+Running 1 test(s) from tests/
+[PASS] testing_reference_integrationtest::tests::test_setup_steps ([..])
+Running 0 test(s) from src/
+Tests: 1 passed, 0 failed, 0 ignored, 0 filtered out
+```
+</details>
+<br>
+
+The test fails because the `setup` function exceeded the allowed number of steps (100 in this case).
```
cptartur: The output doesn't show any failure

cptartur (approved):
Looks good, just make sure that the comment I've added is addressed

PR #3869, opened 2025-10-24
Validate snfoundry.toml

### LINKED ISSUE #3811: Error handling for `snfoundry.toml` parsing
### Current State

Currently, if `snfoundry.toml` parsing fails, it is silently ignored.

### Objective

When `snfoundry.toml` parsing fails, error should be shown and execution of `sncast` should fail instead of silently ignoring the errors

### Additional Context

_No response_

franciszekjob:
Closes #3811

## Introduced changes

- Add basic validation for snfoundry.toml

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/configuration/src/validation.rs
cptartur, line 1: TOML crate already does parsing of the document and enforces it is valid toml with proper errors. With these checks we will never allow user too see that information.

E.g. they will see 

```
Section [{section_name}] must be a table (e.g. key-value pairs inside [sncast.default])
```

instead of a specific error about table entry.

I'd rather restructure this handling that it tries to do all of the conversions without error handling, silently failing if they do and only check if the user provided keys are valid. Leave the document validity checking to toml crate as it does currently.

The situation we want to avoid is for example if someone provides invalid argument `urls` instead of `url`, so they get an error.

cptartur: I think it can be achieved with that in the config structs like `CastConfig` etc. https://serde.rs/container-attrs.html#deny_unknown_fields
This way we don't have to do any handling of extra args ourselves, maybe just add some context to the errors.

Not sure if we can enable it globally in config somewhere or we need to add these attributes everywhere.

cptartur (changes requested):

crates/configuration/src/validation.rs
franciszekjob: Comment about `deny_unknown_fields` is 100% right, we don't have to implement this manually.

But apart from checking unknown fields, we want to prevent user from a situation, in which their fields won't be under any profile (it can happen simply by commenting out), e.g.

```toml
# [sncast.default]
account = "my_account"
accounts-file = "accounts.json"
url = "http://127.0.0.1:5055/rpc"
```

ATM, no error is raised in such situation (we've experienced this during internal hackaton).
So I would leave sections validation, we can modify error though.

cptartur: What if we created an internal struct to be used as schema for config file (similarly to what we do in starknet.py), could we define the structe of that so we can load the whole config file in type safe way (with serde validation) and then get profiles from that?

Maybe something along the lines of this would work?

```rust
struct Config {
  profiles: Vec<CastConfig>,
}
```

franciszekjob: It could work but I'm wondering if it's not an overkill. It's simpler to validate the toml by checking whether it only contains table, nothing else.

cptartur: I don't think it's simpler than having a struct, this way we don't need to own a logic for any validation that could break.

crates/configuration/tests/data/stubtool_complex_snfoundry.toml
franciszekjob, line 1: note: We assume that config file contains profiles of the same type so we have to move these profiles (`StubComplexConfig` and `StubConfig` are different).

crates/configuration/src/validation.rs
franciszekjob: Ok, changed, TAL pls.

cptartur (approved):

crates/configuration/src/lib.rs
cptartur, line 1: What about adding [deny_unknown_fields](https://serde.rs/container-attrs.html#deny_unknown_fields) to cast config?

cptartur (changes requested):

crates/configuration/src/lib.rs
franciszekjob: https://github.com/foundry-rs/starknet-foundry/pull/3869/commits/43e9bcab0cbef4a0cb79a25426a465375d953aca

crates/configuration/src/lib.rs
```diff
@@ -31,32 +45,38 @@ pub fn resolve_config_file() -> Utf8PathBuf {
...
-                .context("Failed to read snfoundry.toml config file")?
-                .parse::<Value>()
-                .context("Failed to parse snfoundry.toml config file")?;
-
-            let raw_config_json = serde_json::to_value(raw_config_toml)
-                .context("Conversion from TOML value to JSON value should not fail.")?;
-
-            core::load_config(
-                raw_config_json,
-                profile.map_or_else(|| Profile::Default, |p| Profile::Some(p.to_string())),
```
cptartur, line 55: This code was specifically refactored so that `load_config` can actually be shared between this top level function and forge which uses it directly.

If we cannot share logic between cast and forge at all, let's get rid of `core::load_config` entirely and use the primitives from `core` directly in forge logic.

crates/configuration/src/lib.rs
```diff
@@ -31,32 +45,38 @@ pub fn resolve_config_file() -> Utf8PathBuf {
     })
 }
 
-pub fn load_config<T: Config + Default>(
-    path: Option<&Utf8PathBuf>,
-    profile: Option<&str>,
-) -> Result<T> {
-    let config_path = path
+pub fn load_config<T>(path: Option<&Utf8PathBuf>, profile: Option<&str>) -> Result<T>
```
cptartur, line 48: Maybe rename this something it suggests more it's loading form snfoundry_toml?

crates/configuration/src/core.rs
```diff
@@ -3,7 +3,7 @@ use anyhow::anyhow;
 use serde_json::Number;
 use std::env;
 
-fn resolve_env_variables(config: serde_json::Value) -> anyhow::Result<serde_json::Value> {
+pub fn resolve_env_variables(config: serde_json::Value) -> anyhow::Result<serde_json::Value> {
```
cptartur, line 6:
```suggestion
pub(crate) fn resolve_env_variables(config: serde_json::Value) -> anyhow::Result<serde_json::Value> {
```

If you need this in other modules `pub` is okay, but if it remains just for uses in this crate, let's not publish it.

PR #3871, opened 2025-10-24
Ignore cheats test of `meta_tx_v0` on `cairo-native`

MKowalski8:
Closes #

## Introduced changes

- Ignore cheats test of `meta_tx_v0` on `ciaro-native`, since it is not supported

## Checklist

- [ ] Linked relevant issue
- [x] Added relevant tests

### DIFF
--- a/crates/forge/tests/integration/meta_tx_v0.rs
+++ b/crates/forge/tests/integration/meta_tx_v0.rs
@@ -95,6 +95,10 @@ fn meta_tx_v0_with_cheat_caller_address() {
     assert_passed(&result);
 }
 
+#[cfg_attr(
+    feature = "cairo-native",
+    ignore = "Cheats in `meta_tx_v0` are not supported on `cairo-native`"
+)]
 #[test]
 fn meta_tx_v0_with_cheat_block_hash() {
     // TODO(#3704) Remove scarb version check
--- a/crates/forge/tests/integration/meta_tx_v0.rs
+++ b/crates/forge/tests/integration/meta_tx_v0.rs
@@ -21,6 +21,68 @@ fn skip_scarb_lt_2_11_0() -> bool {
     }
 }
 
+#[cfg(feature = "cairo-native")]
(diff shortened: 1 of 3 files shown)


crates/forge/tests/integration/meta_tx_v0.rs
```diff
@@ -21,6 +21,68 @@ fn skip_scarb_lt_2_11_0() -> bool {
     }
 }
 
+#[cfg(feature = "cairo-native")]
+#[test]
+fn check_meta_tx_v0_syscall_work_on_native() {
```
MKowalski8: Note: I added this test, to make sure that `meta_tx_v0` syscall works on `cairo-native`, but if you think it is unnecessary I can remove it.

cptartur: It's okay to keep it, but maybe let's add `_without_cheats` or something to it.

Also, can't we run it on vm as well? Is there anything native specific in it?

crates/forge/tests/integration/meta_tx_v0.rs
```diff
@@ -95,6 +157,10 @@ fn meta_tx_v0_with_cheat_caller_address() {
     assert_passed(&result);
 }
 
+#[cfg_attr(
+    feature = "cairo-native",
+    ignore = "Cheats in `meta_tx_v0` are not supported on `cairo-native`"
+)]
```
cptartur: We are ignoring only this test, do you have an idea why is this test failing while other meta tx v0 tests are working on native?

cptartur (approved):

crates/forge/tests/integration/meta_tx_v0.rs
MKowalski8: We can, but I thought that it is not necessary, since other tests somehow check if this `syscall` work, so it was kind of redundant.

franciszekjob (approved):

PR #3874, opened 2025-10-27
Configure `--network` providers in `snfoundry.toml`

### LINKED ISSUE #3765: Configurable network address
### Objective

Allow configuring addresses uses by `--network` flag in sncast in the toml

### Additional Context

_No response_

MKowalski8:
Closes #3765 

Stack:
- #3874 ⬅
- #3875

## Introduced changes

- Add possibility to configure used providers via `--network` flag in `snfoundry.toml`

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/sncast/src/helpers/config.rs
```diff
@@ -50,6 +50,12 @@ fn build_default_manifest() -> String {
         # accounts-file = "{default_accounts_file}"
         # account = ""
         # keystore = ""
+        
+        # Configure custom network addresses
+        # [sncast.default.networks]
+        # mainnet = "https://custom-starknet.example.com/rpc/mainnet/{{RPC_URL_VERSION}}?api_key=YOUR-API-KEY"
+        # sepolia = "https://custom-starknet.example.com/rpc/sepolia/{{RPC_URL_VERSION}}?api_key=YOUR-API-KEY"
```
franciszekjob: nit: Actually, I think we can reduce these links as they are just placeholders + the file is not bloated:
```suggestion
        # mainnet = "https://mainnet.your-node.com"
        # sepolia = "https://sepolia.your-node.com"
```

crates/sncast/src/helpers/configuration.rs
```diff
@@ -10,6 +10,41 @@ pub const fn show_explorer_links_default() -> bool {
...
+impl NetworksConfig {
+    #[must_use]
+    pub fn get(&self, network: &str) -> Option<&String> {
+        match network {
+            "mainnet" => self.mainnet.as_ref(),
+            "sepolia" => self.sepolia.as_ref(),
+            "devnet" => self.devnet.as_ref(),
+            _ => None,
+        }
+    }
```
franciszekjob, line 29: This should be named as `get_url` and take `&Network` instead of `&str`

crates/sncast/src/helpers/configuration.rs
```diff
@@ -46,6 +81,10 @@ pub struct CastConfig {
     )]
     /// Print links pointing to pages with transaction details in the chosen block explorer
     pub show_explorer_links: bool,
+
+    #[serde(default)]
+    /// Custom network URLs - mainnet, sepolia, and devnet are supported
```
franciszekjob: nit: This may be misunderstood as urls of custom networks 😅 
```suggestion
    /// Configurable urls of predefined networks - mainnet, sepolia, and devnet are supported
```

crates/sncast/src/helpers/configuration.rs
```diff
@@ -71,3 +111,72 @@ impl Config for CastConfig {
...
+            sepolia: Some("https://global-sepolia.example.com".to_string()),
+            devnet: None,
+        };
+        let local = NetworksConfig {
+            mainnet: Some("https://local-mainnet.example.com".to_string()),
+            sepolia: None,
+            devnet: None,
+        };
+
+        global.merge(&local);
```
franciszekjob: Having a `merge` method I would assume that both 

```rust
local.merge(global)
```

and

```rust
global.merge(local)
```

will give the same result, while currently they don't.

We have 2 possible options here:
1. Add a static method `merged` accepting `local` and `global` (this assumes local will always override global)
e.g.
```rust
NetworksConfig::merged(local, global)
```

2. Add `override_with` method, which will always override fields of given instance with values from `other` (assuming they're not `None`).
e.g.
```rust
let mut global = ...;
let local = ...;`

global.override_with(local);
```

Wdyt?

franciszekjob (changes requested):

crates/sncast/src/helpers/config.rs
MKowalski8: I am okay with that

crates/sncast/src/helpers/configuration.rs
MKowalski8: Yeah, good take I didn't notice that, much better

crates/sncast/src/helpers/configuration.rs
MKowalski8: Okay, I will go with second, since it was the main idea behind it, and it is quite self describing

crates/sncast/src/helpers/rpc.rs
```diff
@@ -122,4 +127,37 @@ mod tests {
...
+    async fn test_fallback_to_default_network_url() {
+        let config = CastConfig::default();
+
+        let rpc_args = RpcArgs {
+            url: None,
+            network: Some(Network::Mainnet),
+        };
+
+        let url = rpc_args.get_url(&config).await.unwrap();
+        assert!(url.contains("starknet-mainnet"));
```
franciszekjob: Let's assert equality with the result of this https://github.com/foundry-rs/starknet-foundry/blob/56232a01fe67f74bdea36055d9a548fe5dda5332/crates/sncast/src/helpers/rpc.rs#L80

crates/sncast/src/helpers/configuration.rs
```diff
@@ -71,3 +110,71 @@ impl Config for CastConfig {
...
+        );
+    }
+
+    #[test]
+    fn test_networks_config_rejects_unknown_fields_and_typos() {
+        // Unknown fields should cause an error
+        let toml_str = r#"
+            mainnet = "https://mainnet.example.com"
+            custom = "https://custom.example.com"
+            sepoila = "https://sepolia.example.com"
```
franciszekjob: nit: Obviously incorrect at first glance:
```suggestion
            wrong_key = "https://sepolia.example.com"
```

franciszekjob (approved):
LGTM

crates/sncast/src/helpers/configuration.rs
```diff
@@ -10,6 +10,40 @@ pub const fn show_explorer_links_default() -> bool {
...
 
+#[derive(Deserialize, Serialize, Clone, Debug, PartialEq, Default)]
+#[serde(deny_unknown_fields)]
+pub struct NetworksConfig {
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub mainnet: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub sepolia: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub devnet: Option<String>,
```
cptartur: Why do we need `skip_serializing_if` here?

crates/sncast/src/helpers/configuration.rs
```diff
@@ -71,3 +110,71 @@ impl Config for CastConfig {
...
+        let toml_str = r#"
+            mainnet = "https://mainnet.example.com"
+            custom = "https://custom.example.com"
+            wrong_key = "https://sepolia.example.com"
+        "#;
+
+        let result: Result<NetworksConfig, _> = toml::from_str(toml_str);
+        assert!(result.is_err());
+        assert!(result.unwrap_err().to_string().contains("unknown field"));
+    }
```
cptartur, line 176: This is just testing serde logic, do we need that?

crates/sncast/src/helpers/rpc.rs
```diff
@@ -122,4 +127,40 @@ mod tests {
...
+
+        let rpc_args = RpcArgs {
+            url: None,
+            network: Some(Network::Mainnet),
+        };
+
+        let url = rpc_args.get_url(&config).await.unwrap();
+        assert_eq!(
+            url,
+            format!("https://starknet-mainnet.public.blastapi.io/rpc/{RPC_URL_VERSION}")
```
cptartur: The free provider could be extracted to some const and shared between logic and tests. Otherwise we'll need to update these tests any time we change the provider.

crates/sncast/src/main.rs
```diff
@@ -742,11 +742,11 @@ fn get_cast_config(cli: &Cli, ui: &UI) -> Result<CastConfig> {
...
         load_config::<CastConfig>(Some(&global_config_path.clone()), cli.profile.as_deref())
-            .unwrap_or_else(|_| {
-                load_config::<CastConfig>(Some(&global_config_path), None).unwrap()
-            });
+            .or_else(|_| load_config::<CastConfig>(Some(&global_config_path), None))
+            .map_err(|err| anyhow::anyhow!(format!("Failed to load config: {err}")))?;
 
-    let local_config = load_config::<CastConfig>(None, cli.profile.as_deref())?;
+    let local_config = load_config::<CastConfig>(None, cli.profile.as_deref())
+        .map_err(|err| anyhow::anyhow!(format!("Failed to load config: {err}")))?;
```
cptartur, line 749: What's the motivation for these changes?

MKowalski8: For me the message error of `#[serde(deny_unknown_fields)]` was not enough, so I expanded it this way.

crates/sncast/src/helpers/configuration.rs
MKowalski8: It seems to not affect the logic, so I'll remove it. I thought it is needed

crates/sncast/src/helpers/configuration.rs
MKowalski8: If something were to change in the future, this is a test to check whether the error is actually displayed, so I think it is good to have

crates/sncast/src/helpers/rpc.rs
MKowalski8: I changed it to `Network::free_mainnet_rpc(&FreeProvider::Alchemy)` (which is much better). 
This is only one test, so I think it is not such a big deal, to specially extract it for `const` here

cptartur (approved):

PR #3856, opened 2025-10-22
Create `ContractsDataStore` only once

ddoktorski:
[Refactor only]

PR #3842, opened 2025-10-21
Display gas report table

ddoktorski:
Towards #3660

PR #3876, opened 2025-10-28
Display gas report table for fork contracts

ddoktorski:
Towards #3660


ddoktorski:

- #3794
    - #3795
        - #3796
            - #3798
                - #3801
                    - #3823
                        - #3840
                            - #3841
                                - #3856
                                    - #3842
                                        - #3876 ◀
                                            - #3879
                                            - #3890

crates/forge-runner/src/gas/report.rs
```diff
@@ -12,14 +13,20 @@ use std::fmt::Display;
 type ContractName = String;
 type Selector = String;
 
+#[derive(Debug, Clone, PartialOrd, PartialEq, Ord, Eq)]
+pub enum ContractId {
+    TestContract(ContractName),
```
franciszekjob: nit: wdyt?
```suggestion
    LocalContract(ContractName),
```

crates/forge-runner/src/gas/report.rs
```diff
@@ -119,11 +121,26 @@ impl Display for ReportData {
...
-        .get_contract_name(&class_hash)
-        .map_or("forked contract", |name| name.0.as_str())
-        .to_string()
+impl Display for ContractId {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let name = match self {
+            ContractId::TestContract(name) => format!("{name} Contract"),
+            ContractId::ForkedContract(class_hash) => {
+                format!(
+                    "forked contract\n(class hash: {})",
```
franciszekjob: In many cases people will use addresses in order to interact with contract from forks, so imo it would be handy here too. Displaying only class hash adds an extra step because people have to look it up. Wdyt?

franciszekjob (approved):

crates/forge-runner/src/gas/report.rs
cptartur: I prefer local

crates/forge-runner/src/gas/report.rs
```diff
@@ -119,11 +121,26 @@ impl Display for ReportData {
...
+                    shorten_felt(class_hash.0)
+                )
```
cptartur, line 137: Capitalization is inconsistent between these

crates/forge-runner/src/gas/report.rs
```diff
@@ -119,11 +121,26 @@ impl Display for ReportData {
...
-        .map_or("forked contract", |name| name.0.as_str())
-        .to_string()
+impl Display for ContractId {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let name = match self {
+            ContractId::TestContract(name) => format!("{name} Contract"),
+            ContractId::ForkedContract(class_hash) => {
+                format!(
+                    "forked contract\n(class hash: {})",
+                    shorten_felt(class_hash.0)
```
cptartur, line 136: I wonder if it should be configurable to show full contract urls, maybe let's just add an issue for that, maybe it can be integrated if we ever add a `-v` flag to foundry.

crates/forge-runner/src/gas/utils.rs
```diff
@@ -0,0 +1,27 @@
+use starknet_types_core::felt::Felt;
+
+pub(super) fn shorten_felt(felt: Felt) -> String {
+    let padded = format!("{felt:#066x}");
+    let first = &padded[..4];
+    let last = &padded[padded.len() - 4..];
+    format!("{first}…{last}")
+}
```
cptartur, line 8: The first 3 characters area always `0x0`, maybe we should also show 4 after that? It's not very useful to show that beginning otherwise.

ddoktorski: Changed to first 6 characters

crates/forge-runner/src/gas/report.rs
ddoktorski: That's actually on purpose, normally it is `<capitalized contract name> Contract`, so it would suggest that `Forked` is a contract name

crates/forge-runner/src/gas/report.rs
ddoktorski: Showing full addresses would make the table formatting look messy. I wouldn't provide such an option, this is only a helper to make identifying the contract easier

cptartur (approved):

PR #3877, opened 2025-10-28
Bump blockifier

### LINKED ISSUE #3872: Bump blockifier
### Current State

`blockifier` and `starknet_api` crates had a new pre-release published, 0.16.0-rc.1.

### Objective

Start working on bumping the blockifer

### Additional Context

Currently we use a fork so it will need to be updated.

ksew1:
Closes #3872

PR with fixes in blockifier: https://github.com/starkware-libs/sequencer/pull/10405

### DIFF
--- a/crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/deprecated/cairo0_execution.rs
+++ b/crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/deprecated/cairo0_execution.rs
@@ -20,7 +20,6 @@ use cairo_vm::hint_processor::hint_processor_definition::HintProcessor;
 use cairo_vm::vm::runners::cairo_runner::{CairoArg, CairoRunner};
 
 // blockifier/src/execution/deprecated_execution.rs:36 (execute_entry_point_call)
-#[expect(clippy::result_large_err)]
 pub(crate) fn execute_entry_point_call_cairo0(
     call: ExecutableCallEntryPoint,
     compiled_class_v0: CompiledClassV0,
@@ -90,7 +89,6 @@ pub(crate) fn execute_entry_point_call_cairo0(
 }
 
 // blockifier/src/execution/deprecated_execution.rs:192 (run_entry_point)
-#[expect(clippy::result_large_err)]
 pub fn cheatable_run_entry_point(
     runner: &mut CairoRunner,
     hint_processor: &mut dyn HintProcessor,
@@ -104,13 +102,15 @@ pub fn cheatable_run_entry_point(
     let program_segment_size = None; // Infer size from program.
     let args: Vec<&CairoArg> = args.iter().collect();
 
-    runner.run_from_entrypoint(
-        entry_point_pc,
-        &args,
-        verify_secure,
-        program_segment_size,
-        hint_processor,
-    )?;
+    runner
+        .run_from_entrypoint(
+            entry_point_pc,
+            &args,
+            verify_secure,
+            program_segment_size,
+            hint_processor,
+        )
+        .map_err(Box::new)?;
 
     Ok(())
 }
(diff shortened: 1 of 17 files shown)


crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/deprecated/cairo0_execution.rs
```diff
@@ -104,13 +102,15 @@ pub fn cheatable_run_entry_point(
...
-    )?;
+    runner
+        .run_from_entrypoint(
+            entry_point_pc,
+            &args,
+            verify_secure,
+            program_segment_size,
+            hint_processor,
+        )
+        .map_err(Box::new)?;
```
cptartur, line 113: Why do w need this now? Just a signature change on runner side?

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
```diff
@@ -617,8 +617,9 @@ pub fn add_resources_to_top_call(
     match tracked_resource {
         TrackedResource::CairoSteps => top_call.used_execution_resources += resources,
         TrackedResource::SierraGas => {
+            let builtin_gas_cost = versioned_constants.os_constants.gas_costs.builtins;
             top_call.gas_consumed +=
-                vm_resources_to_sierra_gas(&resources.clone(), versioned_constants).0;
+                vm_resources_to_gas(&resources.clone(), &builtin_gas_cost, versioned_constants).0;
```
cptartur: Why was this changed? Do we also need to update the subtracting logic in `build_profiler_execution_resources` in some way?

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/deprecated/cairo0_execution.rs
ksew1: Signature change from `EntryPointExecutionError::CairoRunError(CairoRunError)` --> `EntryPointExecutionError::CairoRunError(Box<CairoRunError>)`

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
ksew1: This function was just removed: https://github.com/starkware-libs/sequencer/commit/e3f7c0a2b8ddca51a36355cbfaa1ac3e2873e863

Current code does the same thing

crates/runtime/src/starknet/context.rs
```diff
@@ -51,6 +51,7 @@ pub fn build_block_context(block_info: &BlockInfo, chain_id: Option<ChainId>) ->
                 eth_fee_token_address: ContractAddress::try_from_hex_str(ETH_CONTRACT_ADDRESS)
                     .unwrap(),
             },
+            is_l3: false,
```
franciszekjob, line 54: nit: Let's rely on `ChainInfo`’s defined defaults:
```suggestion
            is_l3: Default::default(),
```

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
```diff
@@ -617,8 +617,9 @@ pub fn add_resources_to_top_call(
     match tracked_resource {
         TrackedResource::CairoSteps => top_call.used_execution_resources += resources,
         TrackedResource::SierraGas => {
+            let builtin_gas_cost = versioned_constants.os_constants.gas_costs.builtins;
```
franciszekjob: super nit: Should be plural 😅 :
```suggestion
            let builtin_gas_costs = versioned_constants.os_constants.gas_costs.builtins;
```

franciszekjob (approved):

crates/runtime/src/starknet/context.rs
cptartur: I think it's safer to set false here explicitly, there rather won't be a case where we want this to be `true`.

cptartur (approved):

crates/runtime/src/starknet/context.rs
ksew1: What you probably wanted to do is `..ChainInfo::default()` but i would still prefer to be explicit

franciszekjob: > What you probably wanted to do is `..ChainInfo::default()`

Yeah, my bad.

>  but i would still prefer to be explicit

Let's leave it then

Issue #3594, opened 2025-07-30
Try using `insta` package for test outputs


cptartur:
### Current State

Currently we have to update asserted outputs in tests manually. It might turn out that using `insta` will be easier/faster.

### Objective

Try using `insta` package for test outputs

### Additional Context

We should probably migrate just one or a couple of test modules and see how it goes.

PR #3890, opened 2025-10-31
Migrate gas report tests to use `insta`

ddoktorski:
Related #3594

## Introduced changes

- Use insta snapshots for gas report tests


ddoktorski:

- #3794
    - #3795
        - #3796
            - #3798
                - #3801
                    - #3823
                        - #3840
                            - #3841
                                - #3856
                                    - #3842
                                        - #3876
                                            - #3890 ◀

crates/forge/tests/e2e/common/output.rs
```diff
@@ -0,0 +1,25 @@
+/// Asserts a cleaned `stdout` snapshot using `insta`, filtered for non-deterministic lines.
+/// Uses the current Scarb version as a snapshot suffix.
+#[macro_export]
+macro_rules! assert_cleaned_output {
```
cptartur, line 4: This doesn't really need to be be a macro

crates/forge/tests/e2e/common/output.rs
```diff
@@ -0,0 +1,25 @@
...
+            .run()
+            .expect("Failed to get scarb version")
+            .scarb;
+
+        insta::with_settings!({
+            snapshot_suffix => scarb_version.to_string(),
+            filters => vec![
+                (r"\x1B\[[0-?]*[ -/]*[@-~]", ""), // ANSI escape regex - needed for CI
+                (r"(?m)^\s*(Compiling|Finished|Blocking).*", ""), // scarb output
+                (r"(?m)^\s*(Collected|Running|Tests:|Latest block number).*", ""), // snforge output
```
cptartur, line 17: We are reconstructing these regexes multiple times. Performance doesn't matter that much since this is in tests but if we eventually move the rest of the code to insta it might become impactful.

Can we use lazy or something like that and extract regexes to a static block?

crates/forge/tests/e2e/common/output.rs
ddoktorski: Actually there's a reason why this is implemented as a macro. Insta works in such a way that it saves the snapshot using the name of the function where `assert_snapshot!` is called. So if we were to use a regular function instead of a macro the snapshot would be saved under `assert_cleaned_output@<version>` rather than the actual test name

crates/forge/tests/e2e/common/output.rs
ddoktorski: Regexes are passed as `&str` here. I'm not sure how we could improve performance in this case

crates/forge/tests/e2e/common/output.rs
cptartur: Okay, let's keep it as that then.

crates/forge/tests/e2e/common/output.rs
cptartur: Right, never mind if it's just strings

cptartur (approved):

franciszekjob (approved):

PR #3895, opened 2025-11-03
Release 0.51.2

cptartur:
Closes #

## Introduced changes

- Forced Cairo 2.12.3 for all packages
- Set up required envs in scheduled tests and build jobs
- Replaced blast api in generated code with placeholder URL

## Checklist

- [ ] Linked relevant issue
- [ ] Added relevant tests


.github/workflows/_build-binaries.yml
```diff
@@ -11,6 +11,9 @@ on:
         required: false
         type: string
 
+env:
+  ALCHEMY_API_KEY: ${{ secrets.ALCHEMY_API_KEY }}
```
ddoktorski, line 15: nit: Is it necessary here given that we make `secrets: inherit` in workflow calls?

crates/forge/tests/e2e/new.rs
```diff
@@ -24,7 +23,8 @@ static RE_NEWLINES: LazyLock<Regex> = LazyLock::new(|| Regex::new(r"\n{3,}").unw
 
 #[test_case(&Template::CairoProgram; "cairo-program")]
 #[test_case(&Template::BalanceContract; "balance-contract")]
-#[test_case(&Template::Erc20Contract; "erc20-contract")]
+// TODO restore this test case
+// #[test_case(&Template::Erc20Contract; "erc20-contract")]
```
ddoktorski, line 27: Why is it broken?

ddoktorski (approved):

crates/forge/tests/e2e/new.rs
cptartur: ERC-20 template has fork testing which need RPC provider and we no longer provide any in foundry.

franciszekjob (approved):

crates/forge/tests/e2e/new.rs
cptartur: Linking #3896 there

.github/workflows/_build-binaries.yml
cptartur: Not sure how it is actually resolved, but from my understanding just inheriting the secretes doesn't mean envs are also set.

cptartur: Shouldn't be problematic to keep it even if it's not 100% necessary and it's not practical to test it.

PR #3906, opened 2025-11-04
Fix snforge new

ksew1:
commit-id:13eded66

---

**Stack**:
- #3907
- #3906 ⬅

### DIFF
--- a/crates/forge/src/new.rs
+++ b/crates/forge/src/new.rs
@@ -17,6 +17,7 @@ static TEMPLATES_DIR: Dir = include_dir!("snforge_templates");
 
 const DEFAULT_ASSERT_MACROS: Version = Version::new(0, 1, 0);
 const MINIMAL_SCARB_FOR_CORRESPONDING_ASSERT_MACROS: Version = Version::new(2, 8, 0);
+const SCARB_WITHOUT_CAIRO_TEST_TEMPLATE: Version = Version::new(2, 13, 0);
 
 struct Dependency {
     name: String,
@@ -263,8 +264,9 @@ fn add_assert_macros(document: &mut DocumentMut) -> Result<()> {
     };
 
     document
-        .get_mut("dev-dependencies")
-        .and_then(|dep| dep.as_table_mut())
+        .entry("dev-dependencies")
+        .or_insert(Item::Table(Table::new()))
+        .as_table_mut()
         .context("Failed to get dev-dependencies from Scarb.toml")?
         .insert("assert_macros", value(version.to_string()));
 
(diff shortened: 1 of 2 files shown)


crates/forge/tests/e2e/new.rs
```diff
@@ -331,3 +332,19 @@ fn create_new_project_and_check_gitignore() {
...
+
+/// Asserts that two manifest contents match, ignoring the order of lines.
+///
+/// # Why Line Order Can Vary
+///
+/// Starting from Scarb version 2.13, `scarb init` no longer inserts the `[dev-dependencies]` section.
+/// When tools like `forge new` generate a new project, they insert this section later
+/// in the manifest. As a result, the relative placement of `[dev-dependencies]` and other sections
+/// might differ depending on the version of Scarb used.
+fn assert_manifest_matches(expected: &str, actual: &str) {
```
cptartur, line 344: This wouldn't detect completely incorrect order of the lines, can we sort it by sections somehow?

ksew1: I don't think it is worth way more complicated code.

For example toml_edit parser doesn't preserve comments.
So we would need to use regex etc

ksew1: But i you really want we can do it

ksew1: We can also force `dev-dependencies` section to be below dependencies if you want

cptartur: >  So we would need to use regex etc

Let's not do it then 😅

cptartur (approved):

crates/forge/src/new.rs
```diff
@@ -377,19 +379,27 @@ pub fn new(
             cmd.arg("--no-vcs");
         }
 
-        cmd.env("SCARB_INIT_TEST_RUNNER", "cairo-test")
+        let test_runner = if scarb_version < SCARB_WITHOUT_CAIRO_TEST_TEMPLATE {
+            "cairo-test"
+        } else {
+            "none"
+        };
```
MKowalski8, line 387: Do we want to add some kind of `// TODO` here for the future? In the other hand during some refactoring this  const `SCARB_WITHOUT_CAIRO_TEST_TEMPLATE` might be deleted on its own, so up to you.

MKowalski8 (approved):

crates/forge/src/new.rs
ksew1: Added

franciszekjob (approved):

PR #3911, opened 2025-11-05
Bump `MINIMAL_SCARB_VERSION` to `2.11.4`

ksew1:


crates/forge/src/lib.rs
```diff
@@ -34,8 +34,8 @@ mod warn;
 
 pub const CAIRO_EDITION: &str = "2024_07";
 
-const MINIMAL_SCARB_VERSION: Version = Version::new(2, 10, 0);
-const MINIMAL_RECOMMENDED_SCARB_VERSION: Version = Version::new(2, 10, 1);
+const MINIMAL_SCARB_VERSION: Version = Version::new(2, 11, 0);
+const MINIMAL_RECOMMENDED_SCARB_VERSION: Version = Version::new(2, 11, 4);
```
cptartur, line 38: I think we should just bump the `MINIMAL_RECOMMENDED_SCARB_VERSION` unless there have been any known breaking changes for older scarbs.

Historically we kept it below `MINIMAL_RECOMMENDED_SCARB_VERSION`, we don't want to block users from using older scarbs if they really want to unless we know it won't work for sure.

cptartur (changes requested):

crates/forge/src/lib.rs
ksew1: Reverted in https://github.com/foundry-rs/starknet-foundry/pull/3912

ksew1: spr isssue

PR #3912, opened 2025-11-05
Bump `MINIMAL_SCARB_VERSION` to `2.11.4`

ksew1:
commit-id:41b271a1

---

**Stack**:
- #3914
- #3912 ⬅

### DIFF
--- a/crates/forge/src/lib.rs
+++ b/crates/forge/src/lib.rs
@@ -35,7 +35,7 @@ mod warn;
 pub const CAIRO_EDITION: &str = "2024_07";
 
 const MINIMAL_SCARB_VERSION: Version = Version::new(2, 10, 0);
-const MINIMAL_RECOMMENDED_SCARB_VERSION: Version = Version::new(2, 10, 1);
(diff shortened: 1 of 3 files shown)


cptartur (approved):

crates/forge/src/lib.rs
```diff
@@ -35,7 +35,7 @@ mod warn;
 pub const CAIRO_EDITION: &str = "2024_07";
 
 const MINIMAL_SCARB_VERSION: Version = Version::new(2, 10, 0);
```
MKowalski8, line 37: Shouldn't it change to `2.11.0`? Since we had before minimal recommended `2.10.1` and minimal `2.10.0`?

MKowalski8 (approved):

crates/forge/src/lib.rs
cptartur: @MKowalski8 we only want to bump minimal scarb version if we know support for versions before that is 100% broken. If not, users can still keep using that if they want, they'll just see a warning.

PR #3925, opened 2025-11-07
Run oracles tests on CI

cptartur:
**Stack**:
- #3926
- #3925 ⬅
- #3850
- #3849
- #3848
- #3847

### DIFF
--- a/.github/actions/setup-tools/action.yml
+++ b/.github/actions/setup-tools/action.yml
@@ -10,6 +10,10 @@ inputs:
     description: 'Whether to setup scarb'
     required: false
     default: 'true'
+  scarb-version:
+    description: 'Optional: Scarb version to install. If not set, uses version from .tool-versions'
+    required: false
+    default: ''
   setup-usc:
     description: 'Whether to setup universal-sierra-compiler'
     required: false
@@ -24,6 +28,8 @@ runs:
 
     - uses: software-mansion/setup-scarb@v1
       if: ${{ inputs.setup-scarb == 'true' }}
+      with:
+        scarb-version: ${{ inputs.scarb-version }}
 
     - uses: software-mansion/setup-universal-sierra-compiler@v1
       if: ${{ inputs.setup-usc == 'true' }}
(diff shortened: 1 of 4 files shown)


.github/actions/setup-tools/action.yml
```diff
@@ -10,6 +10,10 @@ inputs:
     description: 'Whether to setup scarb'
     required: false
     default: 'true'
+  scarb-version:
+    description: 'Optional: Scarb version to install. If not set, uses version from .tool-versions'
+    required: false
+    default: ''
```
franciszekjob, line 16: Why default value isn't a proper scarb version? Perhaps we could use a minimal supported version here.

crates/forge/Cargo.toml
```diff
@@ -11,6 +11,7 @@ no_scarb_installed = []
 non_exact_gas_assertions = []
 skip_test_for_only_latest_scarb = []
 skip_test_for_scarb_since_2_11 = []
+run_test_for_scarb_since_2_13 = []
```
franciszekjob: Patch version indeed matters in this case, as oracles don't work with 2.13.0:
```suggestion
run_test_for_scarb_since_2_13_1 = []
```

.github/actions/setup-tools/action.yml
cptartur: The default is `''` so we use version from `.tool-versions` if not provided. If there was a default value here then having we would need to always update it when bumping scarb in tool-versions otherwise this setting would override it anyways.

franciszekjob: So we don't need it - should be removed.

franciszekjob (approved):

ksew1 (approved):

.github/actions/setup-tools/action.yml
cptartur: It's so case where input is not provided at all (it's not required) can be handled. I use it to distinguish if it was provided or not.

cptartur:

PR #3949, opened 2025-11-24
Bump starknet packages, integrate new hash function selection logic

cptartur:
Closes #

## Introduced changes

-

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/shared/src/rpc.rs
```diff
@@ -27,3 +28,10 @@ pub async fn get_rpc_version(client: &JsonRpcClient<HttpTransport>) -> Result<Ve
         .parse::<Version>()
         .context("Failed to parse RPC spec version")
 }
+
+pub async fn get_starknet_version(client: &JsonRpcClient<HttpTransport>) -> Result<String> {
+    client
+        .starknet_version(BlockId::Tag(BlockTag::PreConfirmed))
+        .await
+        .context("Error while calling RPC method spec_version")
```
MKowalski8: Is it right error message? It is not `spec_version` if I am right

MKowalski8 (approved):

crates/shared/src/rpc.rs
cptartur: It is not, good catch

franciszekjob (approved):

PR #3955, opened 2025-11-26
ci: bump actions/checkout to v6

rejected-l:
Routine update of checkout action to the latest stable version (v6).

https://github.com/actions/checkout/releases/tag/v6.0.0


cptartur:
Hello, thanks for contributing, however we are not accepting PRs that only bump dependencies as these are handled by dependabot.

PR #3962, opened 2025-11-28
Don't require `Serialize` in `Message`

cptartur:
**Stack**:
- #3965
- #3964
- #3963
- #3962 ⬅
- #3961

⚠️ *Part of a stack created by [spr](https://github.com/ejoffe/spr). Do not merge manually using the UI - doing so may have unexpected results.*

PR #3964, opened 2025-11-28
Implement UI specific for Cast

cptartur:
**Stack**:
- #3965
- #3964 ⬅
- #3963
- #3962
- #3961

### DIFF
--- a/crates/sncast/src/response/ui.rs
+++ b/crates/sncast/src/response/ui.rs
@@ -1,0 +1,197 @@
+use foundry_ui::{Message, OutputFormat, UI as BaseUI};
+use serde::Serialize;
+use serde_json::Value;
+
+#[derive(Default)]
+pub struct UI {
+    ui: BaseUI,
+}
+
+struct MessageWrapper<T: Message> {
+    command: String,
+    message: T,
+}
+struct ErrorWrapper<T: Message> {
+    command: String,
+    error: T,
+}
+struct NotificationWrapper<T: Message> {
+    notification: T,
+}
+struct WarningWrapper<T: Message> {
+    warning: T,
+}
+
+#[derive(Serialize)]
+struct OutputWithCommand<'a, T> {
+    r#type: String,
+    command: String,
+    #[serde(flatten)]
+    data: &'a T,
+}
+
+#[derive(Serialize)]
+struct Output<'a, T> {
+    r#type: String,
+    #[serde(flatten)]
+    data: &'a T,
+}
+
+impl<T> Message for MessageWrapper<T>
+where
+    T: Message,
+{
+    fn text(&self) -> String {
+        self.message.text()
+    }
+
+    fn json(&self) -> Value {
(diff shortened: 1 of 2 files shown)


crates/sncast/src/response/ui.rs
```diff
@@ -0,0 +1,197 @@
...
+struct ErrorWrapper<T: Message> {
+    command: String,
+    error: T,
+}
+struct NotificationWrapper<T: Message> {
+    notification: T,
+}
+struct WarningWrapper<T: Message> {
+    warning: T,
+}
```
franciszekjob, line 23: Wdyt - would it make sense to add `command` to notification and wrapper as well?

crates/sncast/src/response/ui.rs
```diff
@@ -0,0 +1,197 @@
...
+    }
+
+    pub fn print_message<T>(&self, command: &str, message: T)
+    where
+        T: Message,
+    {
+        // TODO(#3960) Add better handling for no JSON output
+        if is_empty_json(&message.json()) {
+            return;
+        }
```
franciszekjob, line 136: Returning silently on an empty JSON may mask unexpected behavior.
If an empty JSON is not a valid case for this message type, wouldn’t it be more appropriate to treat this as an error (e.g., panic or explicit error return) instead of ignoring it?

crates/sncast/src/response/ui.rs
cptartur: It can be done but we would either have to pass `command` to some functions which doesn't necessarily need it (take a look at usages of `print_notification` for example), or we could adjust an abstraction for UI so that it takes a `command ` as an argument in `new` and adds to automatically to all outputs.

Either way, I wouldn't do it in this stack because I think it took long enough compared to the payoff.

crates/sncast/src/response/ui.rs
cptartur: We have cases where we explicitly return `Value::Null` like in `DeployCommandMessage` an it gets returned to the user as `null` in the console which isn't great (it was already like this before my changes in this PR).

I though a simplest approach for now would be to add this if, so we just stop printing `null`s. 

We'd either have to implement json output in all these commands that return `Value::Null` now or change the logic somehow (e.g. change `json` signature to `Option<Value>`) so we can represent the cases where we don't want to print anything in json mode.

franciszekjob: Ok, I see

> We have cases where we explicitly return Value::Null like in DeployCommandMessage an it gets returned to the user as null in the console which isn't great (it was already like this before my changes in this PR).

This is reasonable explanation, let's leave it then.

franciszekjob (approved):

MKowalski8 (approved):

cptartur:
✓ Commit merged in pull request [#3965](https://github.com/foundry-rs/starknet-foundry/pull/3965)

PR #3965, opened 2025-11-28
Restore `command` field in sncast responses

cptartur:
commit-id:bea0b208

---

**Stack**:
- #3965 ⬅
- #3964
- #3963
- #3962
- #3961


crates/sncast/src/response/declare.rs
```diff
@@ -150,6 +143,7 @@ impl Message for DeployCommandMessage {
     }
 
     fn json(&self) -> Value {
+        // TODO Change this
```
franciszekjob: Issue not linked

crates/sncast/src/response/explorer_link.rs
```diff
@@ -2,13 +2,13 @@ use crate::Network;
...
 use serde::Serialize;
-use serde_json::{Value, json};
+use serde_json::Value;
 use starknet_types_core::felt::Felt;
 
 const SNCAST_FORCE_SHOW_EXPLORER_LINKS_ENV: &str = "SNCAST_FORCE_SHOW_EXPLORER_LINKS";
 
 // TODO(#3391): This code should be refactored to either use common `Message` trait or be directly
-// included in `sncast` output messages.
+//  included in `sncast` output messages.
```
franciszekjob, line 11: Any specific reason for this change?

crates/sncast/src/main.rs
```diff
@@ -670,16 +700,36 @@ async fn run_async_command(cli: Cli, config: CastConfig, ui: &UI) -> Result<()>
...
+
+            impl Message for CompletionsMessage {
+                fn text(&self) -> String {
+                    self.completions.clone()
+                }
+
+                fn json(&self) -> Value {
+                    json!(self)
+                }
+            }
```
franciszekjob: Let's move this to separate file

crates/sncast/src/main.rs
```diff
@@ -670,16 +700,36 @@ async fn run_async_command(cli: Cli, config: CastConfig, ui: &UI) -> Result<()>
...
                 );
 
                 // `#` is required since `sncast completion` generates a script and the output is used directly
-                ui.println(&format!("# {}", message.text()));
+                ui.print_message(
+                    "completion",
+                    CompletionsMessage {
+                        completions: format!("# {}", message.text()),
+                    },
+                );
```
franciszekjob, line 716: It seems a bit artificial to separate a command from its message type. Shouldn't a command always be limited to its message type?

franciszekjob (commented):
Presence of new fields in JSON responses should be mentioned in changelog

crates/sncast/src/response/explorer_link.rs
cptartur: If it's indented, it treats it as an continuation of TODO in Intellij and colors the whole string.

obraz

obraz

crates/sncast/src/main.rs
cptartur: Not sure I get it, what change you'd like to have here?

crates/sncast/src/main.rs
cptartur: Moved

crates/sncast/src/response/declare.rs
cptartur: Linked now

crates/sncast/src/response/explorer_link.rs
franciszekjob: Oh nice, TIL

crates/sncast/src/main.rs
franciszekjob: I mean that we pass command (in this case `completion`) and the message instance separately, to `print_message`. Why can't the command be placed directly in the message type (as getter or smth)?

crates/sncast/src/starknet_commands/account/import.rs
```diff
@@ -73,7 +73,7 @@ pub async fn import(
 ) -> Result<AccountImportResponse> {
     // TODO(#3556): Remove this warning once we drop Argent account type
     if import.account_type == AccountType::Argent {
-        ui.println(&WarningMessage::new(
+        ui.print_warning(WarningMessage::new(
```
MKowalski8, line 76: Should `print_warning` take `WarningMessage`? Shouldn't it wrap itself?

crates/sncast/src/main.rs
```diff
@@ -121,6 +123,28 @@ struct Cli {
     command: Commands,
 }
 
+impl Cli {
```
MKowalski8, line 126: Is it intended to be `Cli` not `CLI`? I usually come across the `CLI` version

crates/sncast/src/main.rs
cptartur: This doesn't allow for same message to be reused between different commands. Not that we have cases like this now, but it seemed better for me to support these.

crates/sncast/src/starknet_commands/account/import.rs
cptartur: It takes anything that implements `Message`, and `WarningMessage` is from foundry-ui and has warning specific formatting. 

It seemed easier to do it this way than to change every usage of `WarningMessage` in cast and duplicating the formatting there.

crates/sncast/src/main.rs
cptartur: It's been that way forever in cast, I just added an impl to that. I think it's not worth changing it here (and it technically follows the naming scheme I think).

crates/sncast/src/starknet_commands/account/import.rs
MKowalski8: So that I understand correctly

There would be difference bettwen `ui.print_warning(ErrorMessage::new(...))` and `ui.print_warning(WarningMessage::new(...)),`?

For me it should have the same output, since it is `print_warning` not `print_error`, but if I am correct, `foundry-ui` will add `Error` and `Warning` itself

Edit: I've read how it works in the second PR, so I see there would be no difference (correct me if I am wrong)

cptartur: > There would be difference bettwen ui.print_warning(ErrorMessage::new(...)) and ui.print_warning(WarningMessage::new(...)),?

In the current setup, there would be no difference on the printing level, the only difference in the output would come from how `text` and `json` are implemented in `ErrorMessage` and `WarningMessage`.

I considered an approach where I would introduce new traits specifically for `print_*` methods that would need to be implemented but it was much bigger change to cast logic. Currently the only difference is that using `print_warning` adds `"type": "warning"` to the json output so it can be distingushed easily.

MKowalski8 (approved):

crates/sncast/src/starknet_commands/account/import.rs
MKowalski8: > would come from how text and json are implemented in ErrorMessage and WarningMessage

But in `print_warning` you do not use `text` method, you only take `message`, so there would be no difference, right?
``` rust
let internal_message = WarningWrapper { warning: message };
self.ui.println(&internal_message);
```

cptartur: Take a look at Take a look at how `Message` is then implemented for `WarningWrapper`:

https://github.com/foundry-rs/starknet-foundry/blob/d0aa1579d15bf53e91820f17973eb4b6f1e8f1a9/crates/sncast/src/response/ui.rs#L80

It uses the underlying warning/message `.text()` method.

franciszekjob (approved):

PR #3966, opened 2025-11-28
Don't ask for password in declare two times

### LINKED ISSUE #3799: Fix keystores support
### Current State

When creating a keystore account, the address saved in the file doesn't match the one of the generated one. It is impossible to use that account with sncast.

### Objective

Fix generated file so it can be used again

### Additional Context

_No response_

ksew1:
Closes #3799

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 #### Fixed
 
 - Restored missing `command` missing from some responses when using `--json` output format.
+- When deploying account with `--keystore` flag `sncast` now longer ask two times for password
 
 ## [0.53.0] - 2025-11-24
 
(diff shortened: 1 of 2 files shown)


cptartur (approved):

CHANGELOG.md
```diff
@@ -15,6 +15,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Hash function used when declaring contracts is selected based on the Starknet version
 
+#### Fixed
+
+- When deploying account with `--keystore` flag cast now longer ask two times for password
```
MKowalski8: Nit:
```suggestion
- When deploying account with `--keystore` flag `sncast` now longer ask two times for password
```

From my research in this form, `sncast` appeared seven times, while `cast` appeared once as a cast script.

MKowalski8: Also shouldn't it be under unreleased? Now it is under `0.53.0`

MKowalski8 (approved):

Issue #3983, opened 2025-12-03
Fuzzed ByteArrays that contain non-printable ascii bytes cause rust to panic


bengineer42:
### Select Tool

snforge

### Foundry Version

0.52.0

### What operating system are you using?

Linux

### What system architecture are you using?

x86

### Issue Description

When running test with `#[fuzzer]` if an input item has a member of `ByteArray` that contain bytes outside of the range of printable ascii characters (32 to 127) the thread panics.  The bytes are displayable by cairo fmt as escaped bytes and dont panic at the cairo level, only at the rust level.

Ascii characters allowed: 9, 10, 12, 13 and  32 to 126 (inclusive).

### Command Line Output

"\0\x01\x02\x03\x04\x05\x06\x07\x08
\x0b
\x0e\x0f\x10\x11\x12\x13\x14\x15\x16\x17\x18\x19\x1a\x1b\x1c\x1d\x1e\x1f !"#$%&'()*+,-./0123456789:;<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[\]^_`abcdefghijklmnopqrstuvwxyz{|}~"

thread 'tokio-runtime-worker' (1853670) panicked at /rust/lib/rustlib/src/rust/library/alloc/src/string.rs:2821:14:
a Display implementation returned an error unexpectedly: Error
stack backtrace:
   0: __rustc::rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::result::unwrap_failed
   3: <cheatnet::runtime_extensions::forge_runtime_extension::ForgeExtension as runtime::ExtensionLogic>::handle_cheatcode
   4: <runtime::ExtendedRuntime<Extension> as cairo_vm::hint_processor::hint_processor_definition::HintProcessorLogic>::execute_hint
   5: cairo_vm::vm::vm_core::VirtualMachine::step
   6: cairo_vm::vm::runners::cairo_runner::CairoRunner::run_until_pc
   7: cairo_vm::vm::runners::cairo_runner::CairoRunner::run_from_entrypoint
   8: forge_runner::running::run_test_case
   9: forge_runner::running::run_fuzz_test::{{closure}}
  10: tokio::runtime::task::core::Core<T,S>::poll
  11: tokio::runtime::task::harness::Harness<T,S>::poll
  12: tokio::runtime::blocking::pool::Inner::run
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
[...]

PR #3984, opened 2025-12-03
ByteArrays bytes that cannot be parsed as ascii printable now fallback to \x00

bengineer42:
ByteArrays bytes that cannot be parsed as ascii printable now fallback to \x00

Closes #

#3983 

## Introduced changes

ByteArrays now are displayed with `String::from_ut8_lossy` and will not panic if contain non ascii printable bytes, also improved handling to check number of pending bytes and not relying on \x0  

## Checklist

- [/] Linked relevant issue
- [ ] Added relevant tests
- [/] Performed self-review of the code


franciszekjob:
Hey @bengineer42 , closing it as the issue was resolved in https://github.com/foundry-rs/starknet-foundry/pull/4034. Thanks for the fix anyway 🙌 !

PR #4009, opened 2025-12-10
Bump recommended scarb to 2.12.2

MKowalski8:
Closes #3678 

Stack:

  #4009
  #4010
  #4011 
  #4012 

## Introduced changes

- Bump minimal recommended scarb to `2.12.2` 

## Checklist

- [x] Linked relevant issue
- [ ] Updated relevant documentation
- [ ] Added relevant tests
- [x] Performed self-review of the code
- [x] Added changes to `CHANGELOG.md`

PR #4012, opened 2025-12-10
Remove `--argent` option

### LINKED ISSUE #3556: Remove `--argent` option
### Current State

Due to rebranding (from Argent to Ready), `--argent` option is deprecated. In future, we should remove it entirely. This is also related with starkli, which currently doesn't operate on "ready" name.

### Objective

Remove `--argent` option

### Additional Context

_No response_

MKowalski8:
Closes #3556 

Stack:

  #4009
  #4010
  #4011 
  #4012 

## Introduced changes

- Remove depracated argent account type

## Checklist

- [x] Linked relevant issue
- [ ] Added relevant tests

### DIFF
--- a/crates/sncast/src/starknet_commands/account/create.rs
+++ b/crates/sncast/src/starknet_commands/account/create.rs
@@ -7,7 +7,6 @@ use camino::Utf8PathBuf;
 use clap::Args;
 use console::style;
 use conversions::IntoConv;
-use foundry_ui::components::warning::WarningMessage;
 use serde_json::json;
 use sncast::helpers::braavos::BraavosAccountFactory;
 use sncast::helpers::constants::{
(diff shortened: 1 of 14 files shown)


crates/sncast/src/starknet_commands/account/create.rs
```diff
@@ -273,12 +262,11 @@ fn create_to_keystore(
                 }
             })
         }
-        AccountType::Argent | AccountType::Ready => {
+        AccountType::Ready => {
             json!({
                 "version": 1,
                 "variant": {
-                    // TODO(#3556): Remove hardcoded "argent" and use format! with `AccountType::Ready`
```
cptartur, line 280: What about preexisting argent accounts, I assume the logic to use them is still existing?

franciszekjob (changes requested):
This breaks backwards compatibility -  accounts with `type: "argent"` should still be recognized and allowed to use.

Image

crates/sncast/src/starknet_commands/account/create.rs
MKowalski8: Yes my bad. I didn't think about that

PR #4013, opened 2025-12-11
Allow using `url` from config for `account create` and `account import`

franciszekjob:
**Stack**
- #4020
- #4019
- #4017
- #4013 ⬅

Closes #

## Introduced changes

When using e.g. `account create` without explicitly passing `--url` or `--network` and having `url` defined in `snfoundry.toml`, we should use the one from config. This PR fixes that.

## Checklist

- [ ] Linked relevant issue
- [x] Added relevant tests


crates/sncast/src/helpers/rpc.rs
```diff
@@ -27,12 +27,6 @@ impl RpcArgs {
         config: &CastConfig,
         ui: &UI,
     ) -> Result<JsonRpcClient<HttpTransport>> {
-        if self.network.is_some() && !config.url.is_empty() {
-            bail!(
-                "The argument '--network' cannot be used when `url` is defined in `snfoundry.toml` for the active profile"
-            )
-        }
-
```
franciszekjob, line 35: note: This doesn't make much sense. Arguments passed explicitly should have priority so both `--network` and `--url` should override `url` from config.

cptartur: I think the motivation for that was to avoid a situation where user uses `--network` flag when they don't understand it's overriding url settings.

I'd consider keeping a warning saying something along the lines of "network used when url is configured in the active profile".

crates/sncast/src/helpers/rpc.rs
```diff
@@ -116,13 +110,18 @@ impl Network {
 }
 
 #[must_use]
-pub fn generate_network_flag(rpc_url: Option<&str>, network: Option<&Network>) -> String {
-    if let Some(network) = network {
+pub fn generate_network_flag(rpc_args: &RpcArgs, config_url: &str) -> String {
```
cptartur, line 113: If it's taking whole `RpcArgs`, it should be implemented for the struct itself.

crates/sncast/tests/e2e/account/create.rs
```diff
@@ -1045,3 +1045,21 @@ pub async fn test_no_explorer_links_on_localhost() {
...
+        temp_dir.path().join(accounts_file),
+    );
+
+    let args = vec!["--accounts-file", accounts_file, "account", "create"];
+
+    let snapbox = runner(&args)
+        .env("SNCAST_FORCE_SHOW_EXPLORER_LINKS", "1")
+        .current_dir(temp_dir.path());
+
+    snapbox.assert().success();
```
cptartur, line 1064: Do we need a test like this specifically for account create? If we don't have one using account from config file already, I'd consider testing it as part of a more standard use case, like `sncast invoke`.

crates/sncast/src/helpers/rpc.rs
franciszekjob: > I'd consider keeping a warning saying something along the lines of "network used when url is configured in the active profile".

But this approach applies to more flags which can be configured in snfoundry.toml (`--account`, `--url`, etc.) and it doesn't make much sense if someone uses configs in a conscious way.

crates/sncast/src/helpers/rpc.rs
franciszekjob: I have to disagree - this function is not behavior intrinsic to `RpcArgs`. It builds CLI flags in a specific context, and we don’t want to couple `RpcArgs` with that logic, so keeping it as a free function is more appropriate here.

crates/sncast/tests/e2e/account/create.rs
franciszekjob: > If we don't have one using account from config file already

I assume it's a typo and you refer to url, right?

I added this case for `create` and `import`, as `url` from config didn't work specifically for these commands.

crates/sncast/src/helpers/rpc.rs
cptartur: It's still closely related to `RpcArgs` and the logic would need to change in case it's structure changed. Also, `RpcArgs` doesn't represent anything complex beyond being a wrapper for rpc arguments user passes through the CLI. Having these methods there makes more sense to me.

crates/sncast/tests/e2e/account/create.rs
cptartur: What I meant is that there's nothing special about these methods now so we could have a test for something more general, like `sncast invoke`. But it doesn't hurt us to have more tests.

cptartur (approved):

crates/sncast/src/helpers/rpc.rs
```diff
@@ -116,13 +110,18 @@ impl Network {
...
+    } else if let Some(rpc_url) = &rpc_args.url {
         format!("--url {rpc_url}")
+    } else if !config_url.is_empty() {
+        // Since url is defined in config, no need to pass any flag
+        String::new()
     } else {
-        unreachable!("Either `--rpc_url` or `--network` must be provided.")
+        unreachable!(
+            "`--rpc_url` or `--network` must be provided or url must be set in snfoundry.toml"
+        );
```
MKowalski8, line 124: Why is this `--rpc_url` not just `--url`? Since we don't have `--rpc_url` flag

MKowalski8 (approved):

crates/sncast/src/helpers/rpc.rs
franciszekjob: Good catch! Changed.

PR #4017, opened 2025-12-12
Use `url::Url` instead of string across sncast

franciszekjob:
**Stack**
- #4020
- #4019
- #4017 ⬅
- #4013

Closes #

## Introduced changes

-

## Checklist

- [ ] Linked relevant issue
- [x] Added relevant tests


crates/configuration/src/lib.rs
```diff
@@ -162,7 +161,8 @@ mod tests {
         }
 
         fn from_raw(config: serde_json::Value) -> Result<Self> {
-            Ok(serde_json::from_value::<StubConfig>(config)?)
+            serde_json::from_value::<Self>(config)
+                .map_err(|e| anyhow!("Failed to parse stubtool config: {e}"))
```
franciszekjob: note: This needs to be changed because it ensures that validation (e.g. for `Url`) will be applied.

cptartur: Not sure I understand this, previously we bubbled the error that came from `from_value`, how mapping it changes the behavior?

crates/shared/src/rpc.rs
```diff
@@ -7,9 +7,8 @@ use starknet_rust::providers::{JsonRpcClient, Provider};
 use std::str::FromStr;
 use url::Url;
 
-pub fn create_rpc_client(url: &str) -> Result<JsonRpcClient<HttpTransport>> {
-    let parsed_url = Url::parse(url).with_context(|| format!("Failed to parse URL: {url}"))?;
-    let client = JsonRpcClient::new(HttpTransport::new(parsed_url));
+pub fn create_rpc_client(url: Url) -> Result<JsonRpcClient<HttpTransport>> {
```
cptartur: Why not continue accepting a reference here?

crates/sncast/src/helpers/devnet/detection.rs
```diff
@@ -31,9 +31,14 @@ pub enum DevnetDetectionError {
         "Found starknet-devnet process, but could not reach it. Please use `--url <URL>` to specify the correct URL."
     )]
     ProcessNotReachable,
+    #[error("Failed to parse devnet URL.")]
+    InvalidUrl {
+        #[source]
+        source: url::ParseError,
+    },
```
cptartur: I think this implementation won't show the underlying error you have, string from `#[error(...)]` is used in displaying, but a test checking how it is shown would be useful.

Also, in addition to printing, I wonder if just using `InvalidUrl(#[from] url::ParseError)` wouldn't be a bit simpler here.

crates/sncast/src/helpers/configuration.rs
```diff
@@ -104,7 +106,8 @@ impl Config for CastConfig {
     }
 
     fn from_raw(config: serde_json::Value) -> Result<Self> {
-        Ok(serde_json::from_value::<CastConfig>(config)?)
+        serde_json::from_value::<Self>(config)
+            .map_err(|e| anyhow!("Failed to parse cast config: {e}"))
```
cptartur: Same here, why do we need this here?

crates/sncast/src/starknet_commands/account/create.rs
```diff
@@ -334,7 +339,7 @@ fn write_account_to_file(
 fn generate_deploy_command(
     accounts_file: &Utf8PathBuf,
     rpc_args: &RpcArgs,
-    config_url: &str,
+    config_url: Option<&Url>,
```
cptartur, line 342: Did it previously depend on config_url being and empty string if not set?

crates/sncast/src/starknet_commands/show_config.rs
```diff
@@ -28,7 +28,7 @@ pub async fn show_config(
         None
     };
 
-    let rpc_url = Some(show.rpc.url.clone().unwrap_or(cast_config.url)).filter(|p| !p.is_empty());
+    let rpc_url = show.rpc.url.clone().or(cast_config.url);
```
cptartur, line 31: This chain returns an `Option`, is a case where both rpc url and config_url are none valid?

crates/sncast/src/lib.rs
```diff
@@ -194,9 +195,8 @@ impl Default for ValidatedWaitParams {
     }
 }
 
-pub fn get_provider(url: &str) -> Result<JsonRpcClient<HttpTransport>> {
-    raise_if_empty(url, "RPC url")?;
-    create_rpc_client(url)
+pub fn get_provider(url: &Url) -> Result<JsonRpcClient<HttpTransport>> {
+    create_rpc_client(url.clone())
 }
```
cptartur, line 200: Why did this change?

crates/sncast/tests/helpers/fixtures.rs
```diff
@@ -90,7 +90,7 @@ pub async fn deploy_latest_oz_account() {
     .await;
 }
 pub async fn deploy_ready_account() {
-    let provider = get_provider(URL).expect("Failed to get the provider");
+    let provider = get_provider(&Url::parse(URL).unwrap()).expect("Failed to get the provider");
```
cptartur, line 93: Cannot we parse `URL` somewhere globally so it can just be used in tests? It's quite annoying to have to always do `Url::parse(URL).unwrap()`.

crates/sncast/src/starknet_commands/account/create.rs
franciszekjob: Yeah but that's kind of weird - it should either be either defined as correct url or not present. In practice, this is a breaking change so we should mention this, wdyt?

crates/sncast/src/lib.rs
franciszekjob: `url` is not a string anymore, so we don't have to check whether it's empty.

crates/sncast/tests/helpers/fixtures.rs
franciszekjob: Doing 5 unwraps in this code is not really painful, compared to over 150 usages of `URL` as string (in sncast e2e tests) 😅

crates/sncast/src/starknet_commands/show_config.rs
franciszekjob: Yes

crates/shared/src/rpc.rs
franciszekjob: Changed

crates/sncast/src/helpers/devnet/detection.rs
franciszekjob: Included the source from parsing error

crates/sncast/src/starknet_commands/account/create.rs
cptartur: We can mention it but it's kind of an weird case.

crates/sncast/tests/helpers/fixtures.rs
cptartur: Okay, if we also use it as a string, we can keep it.

crates/sncast/src/starknet_commands/account/create.rs
franciszekjob: Let's not mention it then

crates/configuration/src/lib.rs
```diff
@@ -231,7 +250,8 @@ mod tests {
         }
 
         fn from_raw(config: serde_json::Value) -> Result<Self> {
-            Ok(serde_json::from_value::<StubComplexConfig>(config)?)
+            serde_json::from_value::<Self>(config)
+                .map_err(|e| anyhow!("Failed to parse stubtool config: {e}"))
```
cptartur: Don't seem it was answered (please redirect me to existing comment if it did), why do we need this mapping of errors?

crates/sncast/src/helpers/devnet/detection.rs
```diff
@@ -31,9 +31,14 @@ pub enum DevnetDetectionError {
         "Found starknet-devnet process, but could not reach it. Please use `--url <URL>` to specify the correct URL."
     )]
     ProcessNotReachable,
+    #[error("Failed to parse devnet URL: {source}")]
+    InvalidUrl {
+        #[source]
+        source: url::ParseError,
+    },
```
cptartur: If you did this instead


```suggestion
    #[error("Failed to parse devnet URL: {0}")]
    InvalidUrl(#[from] url::ParseError),
```

you could then write the code below without any mapping, changing this

```
Ok(Url::parse(&format!("http://{}:{}", info.host, info.port))
                    .map_err(|e| DevnetDetectionError::InvalidUrl { source: e })?)
```

to this

```
Ok(Url::parse(&format!("http://{}:{}", info.host, info.port))?)
```

as automatic conversion is generated

crates/configuration/src/lib.rs
franciszekjob: We don't need it, forgot to remove it from complex config too. Removed.

crates/sncast/src/helpers/devnet/detection.rs
franciszekjob: Applied 👍

cptartur (approved):

MKowalski8 (approved):

PR #4019, opened 2025-12-12
Allow saving `network` in snfoundry.toml

### LINKED ISSUE #3250: `sncast account create` doesn't allow using `--network` with `--add-profile` flag
### Current State

It should be allowed to create account when using a predefined network. Currently it is not possibe

### Objective

Allow using `--network` with `--add-profile` flag

### Additional Context

Creating profile will require saving the free provider URL in `snfoundry.toml` which is not ideal. Some mechanism for that should be proposed (a mini design doc) and then implemented. Maybe it makes sense to extend the provider url syntax, so it either accepts an url or some predefined string, so we can represent this nicely in the toml

franciszekjob:
**Stack**
- #4020
- #4019 ⬅
- #4017
- #4013

Closes #3250

## Introduced changes

-

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/configuration/src/lib.rs
```diff
@@ -161,8 +161,7 @@ mod tests {
         }
 
         fn from_raw(config: serde_json::Value) -> Result<Self> {
-            serde_json::from_value::<Self>(config)
-                .map_err(|e| anyhow!("Failed to parse stubtool config: {e}"))
+            serde_json::from_value::<Self>(config).map_err(anyhow::Error::from)
```
cptartur: Why this mapping?

crates/sncast/src/helpers/config.rs
```diff
@@ -1,12 +1,69 @@
...
+}
+
+impl RpcConfig {
+    pub async fn url(&self) -> Result<Url> {
+        match self {
+            RpcConfig::Url(url) => Ok(url.clone()),
+            RpcConfig::Network(network) => Ok(network.url(&FreeProvider::semi_random()).await?),
+        }
+    }
+}
```
cptartur: I'm not sure about this, this is kind of mixing the logic of loading configs and selecting the RPC provider and we have to duplicate it with the code we already have.

I'd rather have plain url field that is typed `Option<Url>` (or `Option<String>`) and another field for `network` also typed as option something. We can than have a validation step after loading config or something that ensures that both aren't present and panics if they are.

cptartur: Eventually what should probably happen in cast is that we introduce another config that represents a functional config that is created as combination of CLI and file. This could probably simplify the logic we have in other places. And `Config` in current form could be kept as a simple schema for loading tomls.

cptartur (changes requested):

crates/configuration/src/lib.rs
franciszekjob: It's not present after pulling changes from upstream branch

crates/sncast/src/helpers/config.rs
franciszekjob: Sure, we can apply some RPC-related refactor here, but I would stay with current approach of having an enum for url/network in config. It ensures correctness on type level, while separate fields enforce us to make validation in runtime.

> Eventually what should probably happen in cast is that we introduce another config that represents a functional config that is created as combination of CLI and file

Yes, I agree.

cptartur: In my opinion, it's introducing more logic to maintain that is currently beneficial. I'd prefer we have to fields and consider doing some checking of them not being duplicated later. Right now we have a custom deserialize, which while not super complex, increases the surface for things to break

cptartur (commented):
Left a comment

crates/sncast/src/helpers/config.rs
franciszekjob: > it's introducing more logic to maintain that is currently beneficial

It may seem it's introducing some amount of logic, but it's beneficial later as we don't have to operate on two fields, that's my opinion.

@MKowalski8 @ddoktorski @ksew1 wdyt about this?

MKowalski8: When we do this:

> Eventually what should probably happen in cast is that we introduce another config that represents a functional config that is created as combination of CLI and file

Will we need `RpcConfig`? Because if we set all these values ​​at the very beginning and make one config, then we won't need `RpcConfig` anymore, am I right? 

In that case, I think I agree with Artur that at this stage (loading from toml)  it can be kept as simple as possible and it should be handled later

franciszekjob: Ok guys, updated, pls TAL

docs/src/appendix/snfoundry-toml.md
MKowalski8, line 114: From what i checked it doesn't work with this. Could you verify? 
``` toml
[sncast.default]                                         
network = "devnet"

[sncast.default.networks]
"devnet" = "http://127.0.0.1:6000"
```

When I run devnet on 5050 it should try to call 6000 no matter what. I checked also sepolia and effect is the same, it call our default provider and omit this configuration.

franciszekjob: Good catch, fixed in https://github.com/foundry-rs/starknet-foundry/pull/4019/changes/08d108d26dd4da7f262953d994b165bf861b2ce6

MKowalski8 (approved):

crates/sncast/src/helpers/configuration.rs
```diff
@@ -84,10 +88,22 @@ pub struct CastConfig {
...
 }
 
+impl CastConfig {
+    pub fn validate(&self) -> anyhow::Result<()> {
+        match (&self.url, &self.network) {
+            (Some(_), Some(_)) => {
+                anyhow::bail!("Only one of `url` or `network` may be specified")
+            }
+            _ => Ok(()),
+        }
```
cptartur, line 98: Maybe let's add an todo for your issue somewhere in code? https://github.com/foundry-rs/starknet-foundry/issues/4027

crates/sncast/src/helpers/rpc.rs
```diff
@@ -110,17 +116,17 @@ impl Network {
...
         format!("--url {rpc_url}")
-    } else if config_url.is_some() {
-        // Since url is defined in config, no need to pass any flag
+    } else if config.url.is_some() || config.network.is_some() {
+        // Since url or network is defined in config, no need to pass any flag
         String::new()
     } else {
         unreachable!(
-            "`--url` or `--network` must be provided or url must be set in snfoundry.toml"
+            "`--url` or `--network` must be provided or url or network fields must be set in snfoundry.toml"
```
cptartur:
```suggestion
            "`--url` or `--network` must be provided or one of url or network field must be set in snfoundry.toml"
```

crates/sncast/src/lib.rs
```diff
@@ -90,7 +90,8 @@ pub const MAINNET: Felt =
 pub const SEPOLIA: Felt =
     Felt::from_hex_unchecked(const_hex::const_encode::<10, true>(b"SN_SEPOLIA").as_str());
 
-#[derive(ValueEnum, Clone, Copy, Debug, PartialEq)]
+#[derive(ValueEnum, Clone, Copy, Debug, PartialEq, Serialize, Deserialize)]
+#[serde(rename_all = "lowercase")]
```
cptartur, line 94: Why do we need this change now?

crates/sncast/tests/data/files/correct_snfoundry.toml
```diff
@@ -33,6 +33,17 @@ account = "user1"
...
+accounts-file = "/path/to/account.json"
+
 [sncast.no_url]
 accounts-file = "../account-file"
 account = "user1"
+
+[sncast.url_and_network]
+account = "user1"
+url = "http://some.url"
+network = "sepolia"
```
cptartur: Is having both correct? The validation we have for configs says it shouldn't be.

crates/sncast/src/starknet_commands/account/mod.rs
```diff
@@ -174,21 +176,28 @@ pub fn add_created_profile_to_configuration(
...
+            (config.url.clone(), config.network)
+        };
         let config = CastConfig {
-            url: Some(url),
+            url,
+            network,
             account: account_name.into(),
             accounts_file: accounts_file.into(),
-            keystore,
+            keystore: keystore.map(std::convert::Into::into),
```
cptartur: Why do we need mapping here?

crates/sncast/src/starknet_commands/account/mod.rs
```diff
@@ -174,21 +176,28 @@ pub fn add_created_profile_to_configuration(
...
-            .expect("the argument '--network' should not be used with '--add-profile' argument");
+        if rpc_args.network.is_some() {
+            bail!("the argument '--network' should not be used with '--add-profile' argument");
+        }
+
+        let (url, network) = if rpc_args.url.is_some() || rpc_args.network.is_some() {
+            (rpc_args.url.clone(), rpc_args.network)
+        } else {
+            (config.url.clone(), config.network)
+        };
```
cptartur: Is the case where network is in config and url in rpc_args and vice versa not possible?

crates/sncast/src/lib.rs
franciszekjob: `CastConfig` derives `Serialize` and `Deserialize`, so since `Network` is used there, we also need these traits. Without `#[serde(rename_all = "lowercase")]` variants would remain same as in code when mapping to JSON value, so we also need it.

crates/sncast/tests/data/files/correct_snfoundry.toml
franciszekjob: No, it isn't. I moved it to a new file `invalid_snfoundry.toml` so it's not confusing anymore, good catch.

crates/sncast/src/starknet_commands/account/mod.rs
franciszekjob: We can use `keystore: keystore.cloned()`, changed.

crates/sncast/src/starknet_commands/account/mod.rs
franciszekjob: Yes, such case is possible but I don't see a reason that it should be handled separately (if that's what you meant?).
This code simply resolves url and network from rpc args and config (args have priority) so they can be later passed to saved cast config.

docs/src/appendix/snfoundry-toml.md
```diff
@@ -18,13 +18,24 @@ If it is not found, default values will be used.
...
+The `url` field specifies the address of RPC provider. It's mutually exclusive with the `network` field.
 
 ```toml
 [sncast.myprofile]
 url = "http://example.com"
 ```
 
+#### `network`
+
+The `network` field specifies the predefined or custom network to use. It's mutually exclusive with the `url` field.
```
ddoktorski: What do you mean by `custom network`?

crates/sncast/src/starknet_commands/account/mod.rs
```diff
@@ -174,21 +176,28 @@ pub fn add_created_profile_to_configuration(
...
         let config = CastConfig {
-            url: Some(url),
+            url,
+            network,
             account: account_name.into(),
             accounts_file: accounts_file.into(),
-            keystore,
+            keystore: keystore.cloned(),
```
ddoktorski: Why does it need to be cloned now?

crates/sncast/src/helpers/rpc.rs
```diff
@@ -38,9 +38,15 @@ impl RpcArgs {
...
+            (Some(url), _) => Some(url.clone()),
+            (None, Some(network)) => Some(self.resolve_network_url(network, config).await?),
+            (None, None) => None,
+        };
+        match (&self.network, &self.url, &url_from_config) {
             (Some(network), None, _) => self.resolve_network_url(network, config).await,
-            (None, Some(url), _) | (None, None, Some(url)) => Ok(url.clone()),
+            (None, Some(url), _) => Ok(url.clone()),
+            (None, None, Some(url_from_config)) => Ok(url_from_config.clone()),
             _ => bail!("Either `--network` or `--url` must be provided."),
```
ddoktorski: Please refactor this to have only one match statement

docs/src/appendix/snfoundry-toml.md
franciszekjob: I mostly meant the aspect of customizing the address of predefined network but looking at it now, it doesn't make much sense + it's not necessary. Changed it to:
```
The `network` field specifies the network to use. It's mutually exclusive with the `url` field.
```

crates/sncast/src/starknet_commands/account/mod.rs
franciszekjob: `keystore` param in `generate_add_profile_message` took reference. Changed it to `Option<Utf8PathBuf>`, no need to clone now.

crates/sncast/src/helpers/configuration.rs
```diff
@@ -48,6 +48,10 @@ pub struct CastConfig {
     /// RPC url
     pub url: Option<Url>,
 
+    #[serde(default)]
+    /// RPC network
```
ddoktorski: nit: seems redundant
```suggestion
```

docs/src/appendix/sncast/account/delete.md
```diff
@@ -20,6 +20,8 @@ Use predefined network with a public provider
 
 Possible values: `mainnet`, `sepolia`, `devnet`.
 
+Overrides network from `snfoundry.toml`.
```
ddoktorski, line 23: Isn't it true for all arguments that can be set in `snfoundry.toml`?

ddoktorski (approved):

docs/src/appendix/sncast/account/delete.md
franciszekjob: No, for `--name` it's not true

PR #4021, opened 2025-12-14
Display class hash on already declared

### LINKED ISSUE #3956: Show class hash in `sncast declare` error in case contract is already declared
### Current State

Currently, we have a heuristic here https://github.com/foundry-rs/starknet-foundry/blob/c953d8695197843ed5b69705613733c4d7a1b1eb/crates/sncast/src/starknet_commands/declare.rs#L177 that tries to detect "already declared" error coming from the `estimateFee` endpoint.

We cannot have a concrete match there because RPC spec for estimateFee errors is very basic as of 0.10

### Objective

Add logic to show calculated class hash in `sncast declare` error in case contract is already declared

### Additional Context

It's better we use our own calculated class hash to show in the error (which we always have) instead of using the one from the error because technically the spec doesn't require it to be one always.

franciszekjob:
Closes #3956

## Introduced changes

-

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/sncast/src/response/errors.rs
```diff
@@ -148,7 +149,11 @@ impl From<StarknetError> for SNCastStarknetError {
             StarknetError::TransactionExecutionError(err) => {
                 SNCastStarknetError::TransactionExecutionError(err)
             }
-            StarknetError::ClassAlreadyDeclared => SNCastStarknetError::ClassAlreadyDeclared,
+            StarknetError::ClassAlreadyDeclared => {
+                unreachable!(
+                    "ClassAlreadyDeclared error requires class hash parameter which is present in StarknetError::ClassAlreadyDeclared. This conversion should not be used."
+                )
+            }
```
cptartur, line 156: How can we ensure we'll never perform this conversion? I wonder if it's not better to keep it and show message without class hash in the incorrect case than having cast fail entirely.

franciszekjob: But `SNCastStarknetError::ClassAlreadyDeclared` has an error parameter, so we would have to place a dummy value there (which isn't a good solution).

MKowalski8 (approved):

crates/sncast/src/response/errors.rs
cptartur: Okay, let's hope we don't convert it that way and users report if we somehow forget.

cptartur (approved):

PR #4036, opened 2025-12-17
Bump upload-artifact to v6

Daulox92:
Updated actions/upload-artifact from v5 to v6 across all workflows.


cptartur:
Hello, thanks for contributing, however we are not accepting PRs that only bump dependencies as these are handled by dependabot.

PR #4059, opened 2025-12-22
Exclude pre-release when comparing versions

franciszekjob:
Closes #

## Introduced changes

-

## Checklist

- [ ] Linked relevant issue
- [ ] Added relevant tests

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -33,13 +33,23 @@ jobs:
         with:
           cmd: yq -oy '.workspace.package.version' 'Cargo.toml'
 
-      - name: Get version from the latest releases
+      - name: Get version from latest stable releases (master)
         id: lookupVersionRelease
+        if: github.ref == 'refs/heads/master'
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -39,7 +39,7 @@ jobs:
         with:
           owner: foundry-rs
           repo: starknet-foundry
-          excludes: draft
+          excludes: prerelease, draft
```
MKowalski8, line 52: We deleted it with previous releases, becasue If i remember correctly it causes new release workflow with every commit that was pushed to main (when the current version was prelease) https://github.com/foundry-rs/starknet-foundry/commit/70b32aa33dddccc5c54d77cf525dc5acedcba346

MKowalski8 (changes requested):

.github/workflows/release.yml
franciszekjob: Good catch. I've changed it so on master there's still `excludes: draft` while on different branches there is `excludes: prerelease, draft`

PR #4061, opened 2025-12-22
Update release workflow

franciszekjob:
Closes #

## Introduced changes

-

## Checklist

- [ ] Linked relevant issue
- [ ] Added relevant tests


.github/workflows/release.yml
```diff
@@ -43,6 +43,8 @@ jobs:
 
       - name: Compare versions
         id: validVersion
+        # Check if triggered by branch push
+        if: startsWith(github.ref, 'refs/heads/')
```
cptartur: We can use `github.ref_type` and explicitly check if it was branch or tag, instead of this. Take a look at docs https://docs.github.com/en/actions/reference/workflows-and-actions/contexts

cptartur (changes requested):

cptartur (approved):
Looks good 🙏

cptartur (approved):

.github/workflows/release.yml
```diff
@@ -52,6 +53,22 @@ jobs:
           echo "versionIsValid=$IS_VALID" >> "$GITHUB_OUTPUT"
           echo "version=$COMMIT_VERSION" >> "$GITHUB_OUTPUT"
 
+      - name: Verify release does not already exist
+        if: github.ref_type == 'tag'
+        run: |
+          TAG_NAME="${GITHUB_REF#refs/tags/}"
```
ddoktorski: Wouldn't it be enough to use `GITHUB_REF_NAME`?

.github/workflows/release.yml
```diff
@@ -52,6 +53,22 @@ jobs:
...
+            echo "Release $TAG_NAME already exists – aborting"
+            exit 1
+          fi
+
+          echo "versionIsValid=true" >> "$GITHUB_OUTPUT"
+          echo "version=${TAG_NAME#v}" >> "$GITHUB_OUTPUT"
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+
```
ddoktorski: Let's update result from the jon `Output job skipped` below to reflect these changes

.github/workflows/release.yml
```diff
@@ -52,6 +53,22 @@ jobs:
...
+        run: |
+          TAG_NAME="${GITHUB_REF#refs/tags/}"
+
+          if gh release view "$TAG_NAME" >/dev/null 2>&1; then
+            echo "Release $TAG_NAME already exists – aborting"
+            exit 1
+          fi
+
+          echo "versionIsValid=true" >> "$GITHUB_OUTPUT"
+          echo "version=${TAG_NAME#v}" >> "$GITHUB_OUTPUT"
```
ddoktorski: We should also verify that steps.lookupVersion.outputs.result equals tag version

.github/workflows/release.yml
franciszekjob: Should be, changed

.github/workflows/release.yml
```diff
@@ -42,19 +42,56 @@ jobs:
...
+            echo "Release $TAG_NAME already exists - aborting"
+            exit 1
+          fi
+
+          echo "versionIsValid=true" >> "$GITHUB_OUTPUT"
+          echo "version=$TAG_VERSION" >> "$GITHUB_OUTPUT"
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Set validation result
```
ddoktorski: Actually can we remove this step and `Output job skipped` at all?

ddoktorski (approved):

.github/workflows/release.yml
franciszekjob: ~~Yeah, it's not necessary to have it.~~ It's needed, because of the definition of `verify-version` outputs. Fixed in #4064

PR #4067, opened 2025-12-23
Restore `range_check` as most costly factor in the corresponding gas test

### LINKED ISSUE #4031: Make range check most costly factor in `contract_range_check_cost_cairo_steps`
### Current State

In new scarb the `contract_range_check_cost_cairo_steps` changed his most costly factor to steps, not range check. It should be rewritten to maintain it original establishment

### Objective

Rewrite test to restore range check as most costly factor

### Additional Context

_No response_

MKowalski8:
Closes #4031 

## Introduced changes

- Change test that it uses `range_check` as most costly factor again, not `cairo_steps`

## Checklist

- [ ] Linked relevant issue
- [ ] Added relevant tests

### DIFF
--- a/crates/forge/tests/data/contracts/gas_checker.cairo
+++ b/crates/forge/tests/data/contracts/gas_checker.cairo
@@ -34,68 +34,9 @@ mod GasChecker {
 
         fn range_check(self: @ContractState) {
             // Felt into ContractAddress conversion uses RangeCheck as implicit argument
-            let _x: ContractAddress = 1234.try_into().unwrap();
-            let _x: ContractAddress = 1234.try_into().unwrap();
-            let _x: ContractAddress = 1234.try_into().unwrap();
-            let _x: ContractAddress = 1234.try_into().unwrap();
-            let _x: ContractAddress = 1234.try_into().unwrap();
-            let _x: ContractAddress = 1234.try_into().unwrap();
-            let _x: ContractAddress = 1234.try_into().unwrap();
-            let _x: ContractAddress = 1234.try_into().unwrap();
-            let _x: ContractAddress = 1234.try_into().unwrap();
-            let _x: ContractAddress = 1234.try_into().unwrap();
-            let _x: ContractAddress = 1234.try_into().unwrap();
-            let _x: ContractAddress = 1234.try_into().unwrap();
-            let _x: ContractAddress = 1234.try_into().unwrap();
-            let _x: ContractAddress = 1234.try_into().unwrap();
-            let _x: ContractAddress = 1234.try_into().unwrap();
-            let _x: ContractAddress = 1234.try_into().unwrap();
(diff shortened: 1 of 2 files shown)


crates/forge/tests/data/contracts/gas_checker.cairo
```diff
@@ -34,68 +34,9 @@ mod GasChecker {
...
-            let _x: ContractAddress = 1234.try_into().unwrap();
-            let _x: ContractAddress = 1234.try_into().unwrap();
-            let _x: ContractAddress = 1234.try_into().unwrap();
-            let _x: ContractAddress = 1234.try_into().unwrap();
-            let _x: ContractAddress = 1234.try_into().unwrap();
-            let _x: ContractAddress = 1234.try_into().unwrap();
-            let _x: ContractAddress = 1234.try_into().unwrap();
-            let _x: ContractAddress = 1234.try_into().unwrap();
-            let _x: ContractAddress = 1234.try_into().unwrap();
+            for _ in 0..1000_u16 {
```
MKowalski8, line 37: Note: the smallest value that meets the conditions is around `250`. I added more so that it wouldn't have to be updated with minor optimizations. I haven't noticed that it would have any real impact on execution time.

crates/forge/tests/integration/gas.rs
```diff
@@ -270,18 +270,17 @@ fn contract_range_check_cost_cairo_steps() {
 
     let result = run_test_case(&test, ForgeTrackedResource::CairoSteps);
 
-    // TODO: (#4031) Change test that the range check will be most costly factor, not steps
     assert_passed(&result);
     // 96 = cost of deploy (see snforge_std_deploy_cost test)
-    // 6 = cost of 2344 cairo steps (because int(0.0025 * 2344) = 6)
-    // 0 l1_gas + 96 l1_data_gas + 6 * (100 / 0.0025) l2 gas
+    // 43 = cost of 1052 range check builtins (because int(0.04 * 1052) = 43)
```
cptartur, line 272: Why 1052 if there are 1000 iterations? The 52 rest are coming from other parts of the execution?

MKowalski8: With exactly one iteration, there are 53 range checks, so yes, 52 are coming from other parts.

ddoktorski: Please update the comment above the test function at lines 239-241

MKowalski8: I see this comment has not been updated since the file creation. Even with actual data it is misleading. 
I will just get rid of it, because every necessary data is already in the second comment. 

If you want I can add somewhere note that `range_check` function uses 1000 range checks.

ddoktorski (approved):

franciszekjob (approved):

Issue #3896, opened 2025-11-03
Restore tests and logic using free RPC provider in forge once there's one


cptartur:
### Current State

Currently, we don't provide any free RPC provider in forge because there isn't one (after blast was deprecated).

### Objective

Restore the logic for providing RPC api in forge, e.g. when using erc-20 contract template and relevant tests

ddoktorski:
We can consider predefining `sepolia` and `mainnet` for fork usages. Similarly as it is possible in `Scarb.toml` right now: https://foundry-rs.github.io/starknet-foundry/snforge-advanced-features/fork-testing.html#configure-fork-in-scarbtoml.
Then we could easily fix erc template by changing it to `#[fork("SEPOLIA_LATEST")]`

PR #4073, opened 2026-01-07
`forge new` template: bump OZ version to `3.0.0`

### LINKED ISSUE #3908: Update openzeppelin version used in ERC-20 forge new template to 3.0.0
### Current State

We currently use 1.0.0, and 2.0.0 exists.

### Objective

Update openzeppelin version used in ERC-20 forge new template to 2.0.0

### Additional Context

_No response_

DelevoXDG:
Closes #3908 

## Introduced changes

- bump `openzeppelin_token` to `3.0.0`
- bump `openzeppelin_intefaces` to `2.1.0`
- bump `openzeppelin_utils` to `2.1.0`
- update template

## Checklist

- [x] Linked relevant issue
- [ ] Added relevant tests

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,14 +67,7 @@ jobs:
       - name: Determine test features
         id: test_features
         run: |
-          # Get Scarb version from .tool-versions
-          SCARB_VERSION=$(grep "^scarb " .tool-versions | awk '{print $2}')
-
-          if [ "$(printf '%s\n' "2.15.1" "$SCARB_VERSION" | sort -V | head -n1)" = "2.15.1" ]; then
-            FEATURES="cairo-native,run_test_for_scarb_since_2_15_1"
-          else
-            FEATURES="cairo-native"
-          fi
+          FEATURES="cairo-native,run_test_for_scarb_since_2_15_1"
           echo "features=$FEATURES" >> "$GITHUB_OUTPUT"
       - name: Build and archive tests
         run: cargo nextest archive --cargo-profile ci -p forge --features "${{ steps.test_features.outputs.features }}" --archive-file 'nextest-archive-${{ runner.os }}-native.tar.zst'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,14 +36,7 @@ jobs:
       - name: Determine test features
         id: test_features
         run: |
-          # Get Scarb version from .tool-versions
-          SCARB_VERSION=$(grep "^scarb " .tool-versions | awk '{print $2}')
-          
-          if [ "$(printf '%s\n' "2.15.1" "$SCARB_VERSION" | sort -V | head -n1)" = "2.15.1" ]; then
-            FEATURES="run_test_for_scarb_since_2_15_1"
-          else
-            FEATURES=""
-          fi
+          FEATURES="run_test_for_scarb_since_2_15_1"
           echo "features=$FEATURES" >> "$GITHUB_OUTPUT"
       - name: Build and archive tests
         run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,8 +33,25 @@ jobs:
         uses: taiki-e/install-action@v2
         with:
           tool: nextest@0.9.98
+      - name: Determine test features
+        id: test_features
+        run: |
+          # Get Scarb version from .tool-versions
+          SCARB_VERSION=$(grep "^scarb " .tool-versions | awk '{print $2}')
+          
+          if [ "$(printf '%s\n' "2.15.1" "$SCARB_VERSION" | sort -V | head -n1)" = "2.15.1" ]; then
+            FEATURES="run_test_for_scarb_since_2_15_1"
+          else
+            FEATURES=""
+          fi
+          echo "features=$FEATURES" >> "$GITHUB_OUTPUT"
       - name: Build and archive tests
-        run: cargo nextest archive --cargo-profile ci -p forge --archive-file 'nextest-archive-${{ runner.os }}.tar.zst'
+        run: |
+          if [ -n "${{ steps.test_features.outputs.features }}" ]; then
+            cargo nextest archive --cargo-profile ci -p forge --features "${{ steps.test_features.outputs.features }}" --archive-file 'nextest-archive-${{ runner.os }}.tar.zst'
+          else
+            cargo nextest archive --cargo-profile ci -p forge --archive-file 'nextest-archive-${{ runner.os }}.tar.zst'
+          fi
       - name: Upload archive to workflow
         uses: actions/upload-artifact@v5
         with:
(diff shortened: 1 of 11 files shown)


franciszekjob (changes requested):
I tried to create a new package with erc20-template and it turns out that there is a dependency conflict:

Image

No surprise that CI didn't fail because the test with this template is commented out 😅 
https://github.com/foundry-rs/starknet-foundry/blob/c26a21c929763341d05b16c79aa074484cd04f91/crates/forge/tests/e2e/new.rs#L28

First, let's restore this test (it should be possible) and then let's resolve the deps issue

DelevoXDG:
This is a bug on the Scarb's resolver side: https://swmansion.slack.com/archives/C05LGECFSCR/p1767959068837409?thread_ts=1767958432.422929&cid=C05LGECFSCR

It should be fixed in the next version of Scarb (2.15.1) 👍🏻

DelevoXDG:
On the side note:

> No surprise that CI didn't fail because the test with this template is commented out 😅

There's actually another test that I relied on https://github.com/foundry-rs/starknet-foundry/blob/6a369e47d574a8fc35f4cb503d08213e159cf267/crates/forge/tests/e2e/templates.rs#L16-L16

that is testing this template as well.

Both tests don't fail neither on CI nor locally, because we have `DEV_DISABLE_SNFORGE_STD_DEPENDENCY` set to `true`, which is masking the Scarb's resolver bug.

The question is whether keeping `DEV_DISABLE_SNFORGE_STD_DEPENDENCY` is worth the risk of potentially hiding certain errors? 

I see there's previously been discussion on removing it: https://swmansion.slack.com/archives/C05K10167C3/p1740758248886209?thread_ts=1740757438.730109&cid=C05K10167C3

Is there a particular reason it was decided not to do so?

@franciszekjob

crates/forge/tests/e2e/new.rs
```diff
@@ -25,8 +25,7 @@ static RE_NEWLINES: LazyLock<Regex> = LazyLock::new(|| Regex::new(r"\n{3,}").unw
 
 #[test_case(&Template::CairoProgram; "cairo-program")]
 #[test_case(&Template::BalanceContract; "balance-contract")]
-// TODO(#3896) restore this test case
```
DelevoXDG, line 28: Note: 

#3896 appears to be unrelated?

ddoktorski:
> The question is whether keeping DEV_DISABLE_SNFORGE_STD_DEPENDENCY is worth the risk of potentially hiding certain errors?
> Is there a particular reason it was decided not to do so?

I'd be happy to remove it, but unfortunately it might cause errors in releases, since it would reference a version that doesn't exist in the registry yet. If you have any ideas on how we can rework this, feel free to change it 🙏

franciszekjob:
> On the side note:
> 
> > No surprise that CI didn't fail because the test with this template is commented out 😅
> 
> There's actually another test that I relied on
> 
> https://github.com/foundry-rs/starknet-foundry/blob/6a369e47d574a8fc35f4cb503d08213e159cf267/crates/forge/tests/e2e/templates.rs#L16-L16
> 
> that is testing this template as well.
> 
> Both tests don't fail neither on CI nor locally, because we have `DEV_DISABLE_SNFORGE_STD_DEPENDENCY` set to `true`, which is masking the Scarb's resolver bug.
> 
> The question is whether keeping `DEV_DISABLE_SNFORGE_STD_DEPENDENCY` is worth the risk of potentially hiding certain errors?
> 
> I see there's previously been discussion on removing it: https://swmansion.slack.com/archives/C05K10167C3/p1740758248886209?thread_ts=1740757438.730109&cid=C05K10167C3
> 
> Is there a particular reason it was decided not to do so?
> 
> @franciszekjob

I think we can spend some time and potentially figure out a different solution to this problem, hopefully a better one, wdyt @DelevoXDG ?

CHANGELOG.md
```diff
@@ -37,6 +37,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 #### Changed
 
 - Minimal recommended `Scarb` version is now `2.12.2` (updated from `2.11.4`)
+- Template created by `snforge new --template erc20-contract` now uses `openzeppelin_token` version `3.0.0` (updated from `1.0.0`)
```
franciszekjob: pls move it under `[Unreleased]`

franciszekjob (approved):

.github/workflows/ci.yml
```diff
@@ -33,8 +33,25 @@ jobs:
...
         with:
           tool: nextest@0.9.98
+      - name: Determine test features
+        id: test_features
+        run: |
+          # Get Scarb version from .tool-versions
+          SCARB_VERSION=$(grep "^scarb " .tool-versions | awk '{print $2}')
+          
+          if [ "$(printf '%s\n' "2.15.1" "$SCARB_VERSION" | sort -V | head -n1)" = "2.15.1" ]; then
+            FEATURES="run_test_for_scarb_since_2_15_1"
```
MKowalski8: Do we need to do this on CI? Since `CI` will be run only for scarb >= 2.15.1?

DelevoXDG: We theoretically can change what is in `.tool_versions`. But I guess we shouldn't ever revert this to past version? 🤔

ddoktorski: > But I guess we shouldn't ever revert this to past version?

Yup, that's correct. Let's remove it

CHANGELOG.md
```diff
@@ -13,6 +13,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - `--detailed-resources` output now includes all gas-related resources
 - `deploy` and `deploy_at` methods on `ContractClass` instance now allow constructor errors to be caught instead of failing immediately. This change diverges from the behavior of the `deploy_syscall` on the network, where such errors cannot be caught.
+- Template created by `snforge new --template erc20-contract` now uses `openzeppelin_token` version `3.0.0` (updated from `1.0.0`)
```
ddoktorski: We can mention that it requires Scarb `2.15.1` to work

crates/forge/tests/e2e/new.rs
```diff
@@ -25,8 +25,11 @@ static RE_NEWLINES: LazyLock<Regex> = LazyLock::new(|| Regex::new(r"\n{3,}").unw
...
-// #[test_case(&Template::Erc20Contract; "erc20-contract")]
+#[test_case(&Template::Erc20Contract; "erc20-contract")]
+#[cfg_attr(
+    not(feature = "run_test_for_scarb_since_2_15_1"),
+    ignore = "Skipping test because feature run_test_for_scarb_since_2_15_1 is not enabled"
```
ddoktorski: nit:
```suggestion
    ignore = "Skipping because ERC-20 template works only with Scarb >=2.15.1"
```

DelevoXDG: Hmm, not sure I agree, we can have Scarb >= 2.15.1 installed locally, but because feature is not specified, it will be ignored.

This makes it clear what to do.

That's also how we did it before:
https://github.com/foundry-rs/starknet-foundry/blob/6924af1a7a16d32dd522cbbd9c0bc46fef2f7fc2/crates/forge/tests/e2e/oracles.rs#L9-L9

ddoktorski: That's fine if you prefer it that way. One other (and more important) thing I just realized is that we should skip only the ERC test case, not all of them

DelevoXDG: > One other (and more important) thing I just realized is that we should skip only the ERC test case, not all of them

You're right! Fixed

ddoktorski (approved):

PR #4075, opened 2026-01-08
Fix unexpected plugin attributes

### LINKED ISSUE #2484: Fix handling incorrect arguments in test attributes

snforge

### Description

For example this snippet would work.
```rust
#[test]
#[fork(tomato: "red", url: "http://your.rpc.url", block_number: 123)]
fn test_using_forked_state() {
    // ...
}
```
But in this scenario `snforge` should throw an error because `tomato` isn't a proper argument.

DelevoXDG:
Closes https://github.com/foundry-rs/starknet-foundry/issues/2484

## Introduced changes

- error out on unexpected args in macros: `fork`, `available_gas`, `fuzzer`, `should_panic`, `test_case`
- refactor `test_case_name` to clarify `name` access at `test_case_internal` level

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests

### DIFF
--- a/crates/snforge-scarb-plugin/src/args/named.rs
+++ b/crates/snforge-scarb-plugin/src/args/named.rs
@@ -35,7 +35,7 @@ impl NamedArgs {
 
         if unexpected.is_empty() {
             return Ok(());
-        };
+        }
 
         // Sort for deterministic output
         unexpected.sort_by_key(|k| k.as_str());
--- a/crates/snforge-scarb-plugin/src/args/named.rs
+++ b/crates/snforge-scarb-plugin/src/args/named.rs
@@ -1,5 +1,7 @@
+use crate::attributes::{AttributeInfo, ErrorExt};
 use cairo_lang_macro::Diagnostic;
 use cairo_lang_syntax::node::ast::Expr;
+use indoc::formatdoc;
 use smol_str::SmolStr;
 use std::{
     collections::HashMap,
@@ -24,6 +26,26 @@ impl DerefMut for NamedArgs {
 }
 
 impl NamedArgs {
+    pub fn allow_only<T: AttributeInfo>(&self, expected: &[&str]) -> Result<(), Diagnostic> {
+        let unexpected = self
+            .0
+            .keys()
+            .filter(|k| !expected.contains(&k.as_str()))
+            .collect::<Vec<_>>();
+
+        if unexpected.is_empty() {
+            return Ok(());
+        };
+        Err(T::error(formatdoc!(
+            "unexpected argument(s): {}",
+            unexpected
+                .iter()
+                .map(|k| format!("<{}>", k.as_str()))
+                .collect::<Vec<_>>()
+                .join(", ")
+        )))
+    }
+
     pub fn as_once(&self, arg: &str) -> Result<&Expr, Diagnostic> {
         let exprs = self
             .0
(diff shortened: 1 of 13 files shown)


crates/snforge-scarb-plugin/src/args/named.rs
```diff
@@ -24,6 +26,30 @@ impl DerefMut for NamedArgs {
 }
 
 impl NamedArgs {
+    pub fn allow_only<T: AttributeInfo>(&self, expected: &[&str]) -> Result<(), Diagnostic> {
+        let mut unexpected = self
+            .0
+            .keys()
+            .filter(|k| !expected.contains(&k.as_str()))
+            .collect::<Vec<_>>();
```
franciszekjob, line 34: Very minor nit: Complexity likely doesn’t matter here because we only ever call `allow_only` with a small, constant-sized expected list, so the current approach is effectively `O(|keys|)`.
Buuut, if we still want to make it asymptotically better in the general case, we could precompute a set:

```rust
let expected: HashSet<&str> = expected.iter().copied().collect();
```

Then the check becomes O(1) average per key, making the overall complexity `O(|keys| + |expected|)` on average instead of `O(|keys| * |expected|)` 😄 .

franciszekjob (approved):
LGTM

crates/snforge-scarb-plugin/src/args/named.rs
DelevoXDG: Yes, I've considered the complexity aspect as well, but I think we should consider first and foremost actual performance rather then asymptotical one.

The constant factor hidden behind `O(1)` lookup and `O(n)` creation of `HashSet` is likely significant, as it involves allocations and costs of computing a hash.

So, it seems probable to me that for small number of elements using `HashSet` will likely to be worse than better here in terms of real-life performance.

I didn't benchmark this, but that's something that seems to be true for Rust as well: https://stackoverflow.com/questions/72877598/hashset-is-slower-than-bruteforce-in-rust

So, unless we plan to make attributes with something like 15+ supported args, I think it makes sense to stick to current approach

WDYT?

franciszekjob: > So, it seems probable to me that for small number of elements using HashSet will likely to be worse than better here in terms of real-life performance.

> So, unless we plan to make attributes with something like 15+ supported args, I think it makes sense to stick to current approach

Yes, I agree with both, and in this case we should leave current implementation, I meant broader aspect  😄

franciszekjob (approved):

crates/snforge-scarb-plugin/src/args/named.rs
```diff
@@ -24,6 +26,30 @@ impl DerefMut for NamedArgs {
 }
 
 impl NamedArgs {
+    pub fn allow_only<T: AttributeInfo>(&self, expected: &[&str]) -> Result<(), Diagnostic> {
```
ddoktorski, line 29: What do you think about adding expected names as a const in `AttributeInfo` (e.g. `ALLOWED_NAMED_ARGS`) instead of passing it as an argument to this function?

DelevoXDG: I suppose that can be done. 

What I had in mind though is this being hard to miss:
```rs
    named_args.allow_only::<AvailableGasCollector>(&["l1_gas", "l1_data_gas", "l2_gas"])?;

    let max = u64::MAX;
    let l1_gas = named_args
        .as_once_optional("l1_gas")?
        .map(|arg| Number::parse_from_expr::<AvailableGasCollector>(db, arg, "l1_gas"))
        .transpose()?
        .unwrap_or(Number(max.into()));
        ...
```

Meaning if one adds / removes `".as_once_optional("new_argument")`, it's immediately clear where one should allow / disallow `"new_argument"` 🤔

ddoktorski: Both options work for me, leaving this up to you

ddoktorski (approved):

PR #4080, opened 2026-01-09
Fix release script to update from rc versions

### LINKED ISSUE #3952: Release script doesn't support upgrading from non-standard version
### Current State

When running the release script from non-standard version i.e. one that has additional information after patch number (e.g. an `-rc.0`), upgrade script will fail at the step which moves files in `.snfoundry_cache` using `mv`

### Objective

Fix the release script so it supports upgrading from non-standard versions

### Additional Context

_No response_

ddoktorski:
Closes #3952

## Introduced changes

- Make script failing on errors
- Modify regex to catch non standard versions

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -1,4 +1,5 @@
 #!/bin/bash
+set -euxo pipefail
 
 VERSION=$1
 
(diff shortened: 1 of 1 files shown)


DelevoXDG (approved):

franciszekjob (approved):
🙏

PR #4081, opened 2026-01-09
Document `skip-optimizations` option for backtrace

### LINKED ISSUE #4022: Investigate backtrace `inlined` regression
### Current State

`(inlined)` disappeared between 2.13.1 and 2.14.0

### Objective

Investigate

### Additional Context

_No response_

MKowalski8:
Closes #4022

## Introduced changes

- Add information about `skip-optimizations` in docs and add test which uses this option 

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests

### DIFF
--- a/docs/src/snforge-advanced-features/debugging.md
+++ b/docs/src/snforge-advanced-features/debugging.md
@@ -148,16 +148,11 @@ to have:
 1. [Scarb](https://github.com/software-mansion/scarb) version `2.8.0` or higher
 2. `Scarb.toml` file with the following Cairo compiler configuration:
 
-> 📝 **Note**
->
-> If you are using scarb 2.12 or later there is a way to improve backtrace for panic in contracts if
-> you set `panic-backtrace = true`
-
 ```toml
 [profile.dev.cairo]
 unstable-add-statements-code-locations-debug-info = true
 unstable-add-statements-functions-debug-info = true
-panic-backtrace = true # only for scarb 2.12 or later
+panic-backtrace = true
 ```
 
 > 📝 **Note**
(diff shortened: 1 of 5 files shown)


docs/src/snforge-advanced-features/debugging.md
```diff
@@ -163,10 +163,25 @@ panic-backtrace = true # only for scarb 2.12 or later
 > 📝 **Note**
 >
 > That `unstable-add-statements-code-locations-debug-info = true` and
-`unstable-add-statements-functions-debug-info = true` will slow down the compilation and cause it to use more system
+> `unstable-add-statements-functions-debug-info = true` will slow down the compilation and cause it to use more system
 > memory. It will also make the compilation artifacts larger. So it is only recommended to add these flags when you need
 > their functionality.
 
+> 📝 **Note**
```
ddoktorski: We have now stacked 3 notes and 1 warning blocks in this section, which makes it a bit unreadable. My suggestions are:
* Remove the note about `panic-backtrace`, make it point 3. or normal text
* Now we have `Prerequisites` and `Usage`. Let's add something like "Additional configuration" section (try to find better name) and move information about `skip-optimization` there, leave the warning
* Focus more on the fact that setting `skip-optimization` will provide users as much information as possible, because it is relevant from the snforge perspective

ddoktorski (changes requested):

docs/src/snforge-advanced-features/debugging.md
MKowalski8: > Remove the note about panic-backtrace, make it point 3. or normal text

Actually this note is outdated, because `panic-backtrace` is now required to run backtrace (we added it with bump minimal recommended scarb to 2.12). 

The question is whether we still want prerequisite on scarb 2.8.0, since it was a long time ago?

```
1. [Scarb](https://github.com/software-mansion/scarb) version `2.8.0` or higher
```

ddoktorski: What about changing this prerequisite to Scarb `2.12.0` then?

MKowalski8: We can do that. However, backtrace for older versions will work as before, you need to add the `panic-backtrace` option, but it simply won't affect anything. This doesn't mean that with scarb 2.11.4 you can no longer get a backtrace. Or at least I think it won't affect anything, I can check it out

MKowalski8: I checked it and it works as I wrote, although we had the following warning before:

```
[WARNING] Scarb version should be 2.12 or higher and `Scarb.toml` should have the following Cairo compiler configuration to get accurate backtrace results:

[profile.dev.cairo]
unstable-add-statements-functions-debug-info = true
unstable-add-statements-code-locations-debug-info = true
panic-backtrace = true # only for scarb 2.12 or higher
... other entries ...
```

So I think we can simply change prerequisite to Scarb 2.12.0 then

ddoktorski (approved):
LGTM

docs/src/snforge-advanced-features/debugging.md
```diff
@@ -234,3 +229,18 @@ stack backtrace:
 </details>
 <br>
 
+### Advanced Configuration
+
+For the most detailed backtrace information, you can add `skip-optimizations = true` to your `Scarb.toml` (requires Scarb >= 2.14.0).
+This skips as much compiler optimization as possible, keeping the compiled code closer to the original and allowing `snforge` to provide more complete and accurate backtrace information.
```
franciszekjob: nit:
```suggestion
This skips as much compiler optimizations as possible, keeping the compiled code closer to the original and allowing `snforge` to provide more complete and accurate backtrace information.
```

franciszekjob (approved):

PR #4083, opened 2026-01-09
Refactor gas resources logic

DelevoXDG:
## Stack
- #4083 
- #4092

## Introduced changes

Add common `GasCalculationResources` that is used both for calculations and displaying resources in `--detailed-resources` output. This will make it easier to track the affected resources and avoid missing the any resources in output.

## Checklist

- [ ] Linked relevant issue
- [ ] Added relevant tests


crates/forge-runner/src/gas/resources.rs
```diff
@@ -0,0 +1,164 @@
...
+                    "{vm_resources_output}
+        syscalls: ({syscalls})
+        "
+                )
+            }
+            ForgeTrackedResource::SierraGas => {
+                let mut output = format!(
+                    "
+        sierra gas: {}
+        syscalls: ({syscalls})",
```
MKowalski8, line 105: I was just wondering if there is a better approach to this (I mean this two `format!`), but I couldn't figure it out. If you have sth in mind, you can try, but I am not pushing.

MKowalski8 (approved):
LGTM

crates/forge-runner/src/gas/resources.rs
MKowalski8: Oh, I see it is changed in #4092, never mind 😅

ddoktorski (approved):

DelevoXDG:
I sent the re-review request mistakenly, meant to do so for the subsequent PR 😅

PR #4085, opened 2026-01-12
Fix nightly build in scheduled tests

ddoktorski:
- Overwrite built binaries within a workflow run if already exist
- Make explicit calls with `prod_registry: false` when publishing to dev registry


.github/workflows/_build-binaries.yml
```diff
@@ -92,3 +92,4 @@ jobs:
         with:
           name: build-${{ matrix.target }}
           path: ${{ env.PKG_FULL_NAME }}.*
+          overwrite: true
```
MKowalski8, line 95: Wouldn't this also allow overwriting the official release? I see that it is used [here in release workflow](https://github.com/foundry-rs/starknet-foundry/blob/6b021bdf2988bd0c81cfba70baa6d1bbafe7f6ba/.github/workflows/release.yml#L97) 
Should we create such a possibility?

ddoktorski: If `build-binaries` is run only once which is the case in the release workflow, then there is nothing to overwrite so it should work as expected

MKowalski8: Yeah I know, but we had the situation that release workflow was run twice (of course it shouldn't have happened, but it happened), but it only overwrote the release notes, the published binary itself remained the same (if I understand it correctly). In this case would it also change the published binary?

MKowalski8: Although maybe I don't fully understand what it does, correct me if I'm wrong.

ddoktorski: It only overwrites artifacts within the scope of a given workflow run. It won't modify any release assets etc.

MKowalski8 (approved):

franciszekjob (approved):

PR #4093, opened 2026-01-14
Fix minor bugs in release flow

ddoktorski:
- Exit without error if tag already exists to avoid misleading failures in standard release flow (release workflow is triggered again after publishing release and creating a tag)
- Fork cache filenames include two versions (rpc and forge) which was overlooked in the last change. It should now correctly modify the path by stripping only the last `v_*` segment

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -80,9 +80,12 @@ jobs:
             exit 1
           fi
 
+          # Tag is usually created when publishing release from GitHub UI.
+          # Then it triggers this workflow again and we want to exit without error.
           if gh release view "$TAG_NAME" >/dev/null 2>&1; then
             echo "Release $TAG_NAME already exists - aborting"
-            exit 1
+            echo "versionIsValid=false" >> "$GITHUB_OUTPUT"
+            exit 0
           fi
 
           echo "versionIsValid=true" >> "$GITHUB_OUTPUT"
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
MKowalski8, line 159: I know that `test-binary` depends on `build-binaries` which won't run without `versionIsValid == 'true'`.
However, the `verify-version` will succeed now (it exit without error), don't we also want to add: 

``` yml
    needs: [ test-binary, verify-version ]
    if: ${{ needs.verify-version.outputs.versionIsValid == 'true' }}
```

As double check?

MKowalski8: Of course it is not needed now, but with some changes in the future, someone might overlook it.

ddoktorski: Actually it seems that for the cleanest approach we should rely on cascade behavior of workflow jobs. Therefore, version validity check can be done only in `build-*` jobs, in all other places `needs: verify-version` and `if: ${{ needs.verify-version.outputs.versionIsValid == 'true' }}` are redundant. What do you think?

MKowalski8: It seems cleaner, so I am okay with that. 

My only concern is whether it’s obvious enough that `build_*` handles this check, but I might be overthinking it, especially since it should be easier to understand if it's done only there. 😅

ddoktorski: Let's change it in a separate PR though

MKowalski8 (approved):

franciszekjob (approved):

PR #4116, opened 2026-01-21
Refactor: add re-usable `clear_events_and_messages_from_reverted_call`

DelevoXDG:
## Stack
- #4116 
- #4117
## Introduced changes

De-duplicates some logic via adding re-usable func (re-used once again in the following PR)

## Checklist

- [ ] Linked relevant issue
- [ ] Updated relevant documentation
- [ ] Added relevant tests
- [x] Performed self-review of the code
- [ ] Added changes to `CHANGELOG.md`

PR #4117, opened 2026-01-21
Make errors from `deploy` and `deploy_at` of `ContractClass` catchable

### LINKED ISSUE #3992: Enable catching deploy panics from top level call
### Issue Description

ATM, when we make a deploy syscalls from nested call and constructor panics, we allow the error to be catched (i.e. it works with `#[should_panic]`. This is inconsistent with the network behavior, described [here](https://community.starknet.io/t/starknet-v0-13-4-pre-release-notes/115257#p-2358763-catching-errors-12)

We should enable catching deploy panics by handling them in `CallToBlockifierRuntime`. Deploy panics from nested calls can throw hard errors as on the real network.

DelevoXDG:
## Stack
- #4116 
- #4117

Closes #3992

## Introduced changes

When `deploy_syscall` is executed in the context of `ContractClass::deploy` `ContractClass::deploy_at` cheatcodes, deployment will be **reverted** instead of resulting in hard error.

This diverges with how `deploy_syscall` behaves on real network: https://community.starknet.io/t/starknet-v0-13-4-pre-release-notes/115257#p-2358763-catching-errors-12 

This is done intentionally to allow catching and testing deployment errors.
 

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,8 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 #### Changed
 
 - `--detailed-resources` output now includes all gas-related resources
-- `deploy` and `deploy_at` methods on `ContractClass` instance now allow errors to be caught instead of failing immediately. This change diverges from the behavior of the `deploy_syscall` on the network, where this errors are not revertible and cannot be caught.
-
+- `deploy` and `deploy_at` methods on `ContractClass` instance now allow errors to be caught instead of failing immediately. This change diverges from the behavior of the `deploy_syscall` on the network, where such errors cannot be caught.
 ## [0.55.0] - 2026-01-13
 
 ## Forge
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 #### Changed
 
 - `--detailed-resources` output now includes all gas-related resources
+- `deploy` method on `ContractClass` instance now allow errors to be caught instead of failing immediately. This change diverges from the behavior of the `deploy_syscall` on the network, where this errors are not revertible and cannot be caught.
 
 ## [0.55.0] - 2026-01-13
 
@@ -21,7 +22,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Minimal recommended `Scarb` version is now `2.13.1` (updated from `2.12.2`)
 - `snforge_scarb_plugin` now emits an error if unexpected named args are passed to `#[available_gas]`, `#[fork]`, `#[fuzzer]`, `#[should_panic]` and `#[test_case]` attributes
-- `--detailed-resources` output now includes all gas-related resources
 
 #### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 #### Changed
 
 - `--detailed-resources` output now includes all gas-related resources
-- `deploy` method on `ContractClass` instance now allow errors to be caught instead of failing immediately. This change diverges from the behavior of the `deploy_syscall` on the network, where this errors are not revertible and cannot be caught.
+- `deploy` and `deploy_at` methods on `ContractClass` instance now allow errors to be caught instead of failing immediately. This change diverges from the behavior of the `deploy_syscall` on the network, where this errors are not revertible and cannot be caught.
 
 ## [0.55.0] - 2026-01-13
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 #### Changed
 
 - `--detailed-resources` output now includes all gas-related resources
-- `deploy` and `deploy_at` methods on `ContractClass` instance now allow errors to be caught instead of failing immediately. This change diverges from the behavior of the `deploy_syscall` on the network, where such errors cannot be caught.
+- `deploy` and `deploy_at` methods on `ContractClass` instance now allow constructor errors to be caught instead of failing immediately. This change diverges from the behavior of the `deploy_syscall` on the network, where such errors cannot be caught.
 ## [0.55.0] - 2026-01-13
 
 ## Forge
--- a/crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/cheated_syscalls.rs
+++ b/crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/cheated_syscalls.rs
@@ -156,9 +156,9 @@ pub fn deploy_syscall(
     let raw_retdata = call_info.execution.retdata.0.clone();
     syscall_handler.base.inner_calls.push(call_info);
 
-    // This should be unreachable with the current implementation of `execute_deployment`
-    // This check is kept in case of future changes + to be aligned with revert handling elsewhere.
     if failed && from_cheatcode {
+        // This should be unreachable with the current implementation of `execute_deployment`
+        // This check is kept in case of future changes + to be aligned with revert handling elsewhere.
         return convert_deploy_failure_to_revert(syscall_handler, revert_idx, raw_retdata);
     }
 
--- a/crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/cheated_syscalls.rs
+++ b/crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/cheated_syscalls.rs
@@ -35,12 +35,12 @@ use cairo_vm::vm::vm_core::VirtualMachine;
 use conversions::string::TryFromHexStr;
 use runtime::starknet::constants::TEST_ADDRESS;
 use starknet_api::abi::abi_utils::selector_from_name;
+use starknet_api::contract_class::EntryPointType;
 use starknet_api::core::EntryPointSelector;
 use starknet_api::transaction::constants::EXECUTE_ENTRY_POINT_NAME;
 use starknet_api::transaction::fields::TransactionSignature;
 use starknet_api::transaction::{TransactionHasher, TransactionOptions, signed_tx_version};
 use starknet_api::{
-    contract_class::EntryPointType,
     core::{ClassHash, ContractAddress, Nonce, calculate_contract_address},
     transaction::{
         InvokeTransactionV0, TransactionVersion,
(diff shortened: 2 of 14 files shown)


crates/forge/tests/e2e/snapshots/backtrace/main__e2e__backtrace__handled_error_not_display@2.15.1.snap
```diff
@@ -2,7 +2,7 @@
 source: crates/forge/tests/e2e/backtrace.rs
 expression: stdout
 ---
-[PASS] dispatchers_integrationtest::test::test_handle_and_panic (l1_gas: ~0, l1_data_gas: ~288, l2_gas: ~576580)
+[PASS] dispatchers_integrationtest::test::test_handle_and_panic (l1_gas: ~0, l1_data_gas: ~288, l2_gas: ~583520)
```
DelevoXDG, line 5: This seemingly is caused by extra `start_cheatcode_deploy` call, is this something we can ignore or is it problematic? 🤔

CHANGELOG.md
```diff
@@ -21,7 +22,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Minimal recommended `Scarb` version is now `2.13.1` (updated from `2.12.2`)
 - `snforge_scarb_plugin` now emits an error if unexpected named args are passed to `#[available_gas]`, `#[fork]`, `#[fuzzer]`, `#[should_panic]` and `#[test_case]` attributes
-- `--detailed-resources` output now includes all gas-related resources
```
DelevoXDG, line 24: Apparently I mistakenly left this here as well 😬

crates/forge/tests/e2e/snapshots/backtrace/main__e2e__backtrace__handled_error_not_display@2.15.1.snap
ddoktorski: We can ignore it, but please note that it probably affects all gas related tests so they need to be updated

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 #### Changed
 
 - `--detailed-resources` output now includes all gas-related resources
+- `deploy` and `deploy_at` methods on `ContractClass` instance now allow errors to be caught instead of failing immediately. This change diverges from the behavior of the `deploy_syscall` on the network, where this errors are not revertible and cannot be caught.
```
```suggestion
- `deploy` and `deploy_at` methods on `ContractClass` instance now allow errors to be caught instead of failing immediately. This change diverges from the behavior of the `deploy_syscall` on the network, where such errors cannot be caught.
```

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
```diff
@@ -198,6 +198,13 @@ impl<'a> ExtensionLogic for ForgeExtension<'a> {
 
                 Ok(CheatcodeHandlingResult::from_serializable(()))
             }
+            // Internal cheatcode to mark next `deploy_syscall` as coming from cheatcode.
+            "start_cheatcode_deploy" => {
```
ddoktorski: nit: Could we rename it to something that is less similar to our standard cheatcode functions? e.g. `set_next_deploy_from_cheatcode`

DelevoXDG: Yes, make sense.

Perhaps `set_next_deploy_syscall_from_cheatcode` to be max-explicit?

CHANGELOG.md
DelevoXDG: I'd rather add more info than less here, tbh 🤔 

Perhaps even linking the 0.13.4 release notes, to answer the questions of whoever might be interested what are we referring here too?

Or do you think it's out of scope for the changelog?

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
DelevoXDG: On the second thought, there's no need to overcomplicate. Updated as you requested.

CHANGELOG.md
ddoktorski: That's fine, but could we rephrase it? It's not about errors being reverted, but rather about the changes that occurred before the error

ddoktorski: > Perhaps even linking the 0.13.4 release notes, to answer the questions of whoever might be interested what are we referring here too?

We could actually add a section to the docs as well. I can try to create something like that as part of the work on revert changes

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/cheated_syscalls.rs
```diff
@@ -107,19 +130,37 @@ pub fn deploy_syscall(
...
+        Ok(info) => info,
+        Err(EntryPointExecutionError::ExecutionFailed { error_trace }) if from_cheatcode => {
+            let panic_data = error_trace.last_retdata.0.clone();
+            return convert_deploy_failure_to_revert(syscall_handler, revert_idx, panic_data);
+        }
+        Err(err) => return Err(err.into()),
+    };
+    let failed = call_info.execution.failed;
+    let raw_retdata = call_info.execution.retdata.0.clone();
+    syscall_handler.base.inner_calls.push(call_info);
```
ddoktorski, line 164: Does it need to be called now at this place rather than at the end?

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/cheated_syscalls.rs
```diff
@@ -107,19 +130,37 @@ pub fn deploy_syscall(
...
 
-    let constructor_retdata =
-        create_retdata_segment(vm, syscall_handler, &call_info.execution.retdata.0)?;
+    if failed && from_cheatcode {
```
ddoktorski, line 166: Is it reachable at all? Given that `execute_deployment` calls `non_reverting_execute_call_entry_point`, all failed calls should be converted to errors then

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/cheated_syscalls.rs
```diff
@@ -107,19 +130,37 @@ pub fn deploy_syscall(
...
         syscall_handler.base.state,
         cheatnet_state,
         syscall_handler.base.context,
         &ctor_context,
         request.constructor_calldata,
         remaining_gas,
-    )?;
+    ) {
+        Ok(info) => info,
+        Err(EntryPointExecutionError::ExecutionFailed { error_trace }) if from_cheatcode => {
```
ddoktorski, line 151: Note that our code hasn't been updated and `execute_deployment` should return `ConstructorEntryPointExecutionError` as here: https://github.com/software-mansion-labs/sequencer/blob/v0.16.0-rc.1/crates/blockifier/src/execution/entry_point.rs#L567. I think it's fine to leave it as is for now, but we should eventually adjust it to align with Blockifier. Could you crate an issue for that?

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/cheated_syscalls.rs
DelevoXDG: I based this off how `execute_inner_call` ordered this.

The order didn't matter before since we didn't revert deploy.

But now that we do, the approach used by `execute_inner_call` makes sense, since we want to clear messages and events for the *last, just reverted* call.
```rs
    syscall_handler.base.inner_calls.push(call_info);

    if failed {
        ...
        let reverted_call = syscall_handler.base.inner_calls.last_mut().unwrap();
        clear_events_and_messages_from_reverted_call(reverted_call);
        ...
    }
```

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/cheated_syscalls.rs
DelevoXDG: It shouldn't be reachable.

But since we have `call_info.execution.failed` handling elsewhere, it just made sense to me to have it here just in case, + to have this aligned with other reverts e.g. `execute_inner_call`, and to work regardless of changes to `non_reverting_execute_call_entry_point`. 

But if you think it's more explicit, I can mark this branch as unreachable.

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/cheated_syscalls.rs
DelevoXDG: https://github.com/foundry-rs/starknet-foundry/issues/4126

crates/cheatnet/src/state.rs
franciszekjob, line 1: I think we could potentially generalize `next_deploy_from_cheatcode` into a reusable syscall-origin mechanism. A per-syscall boolean solves the immediate need, but it does not scale once we want to distinguish origins for other syscalls. I know that we have fields like `deploy_salt_base` / `next_deploy_at_address` but note that they're  inherently deploy-specific, while the case of syscall origin is at a higher abstraction level.

A more extensible approach could be to introduce something like:

```rust
...
enum SyscallOrigin {
    Cheatcode,
    Direct
}
```

and then store a one-shot “pending origin” in the state (similar to how `take_next_*` works now), but in a generic form - e.g. `next_syscall_origin: Option<SyscallOrigin>` or smth similar.

WDYT?

DelevoXDG: We can make it re-usable. 

Correct me if I'm wrong, but it should be even simpler to accomplish - by means of renaming `next_deploy_from_cheatcode` to `next_syscall_from_cheatcode`. If I understand correctly, it should still be able to accomplish the same function as the `SyscallOrigin` enum you suggested.

This way we can both keep it simple, and, if anyone wants to re-use this logic in other cheatcode, it's clear what to use.

Thoughts?

franciszekjob: seems good 👍

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/cheated_syscalls.rs
ddoktorski: Are we sure `inner_calls` are always empty in case of the error flow and this block never runs?
```rust
if let Some(reverted_call) = syscall_handler.base.inner_calls.last_mut() {
    clear_events_and_messages_from_reverted_call(reverted_call);
}
```

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/cheated_syscalls.rs
ddoktorski: We can consider adding a note mentioning that with the current implementation this should be unreachable, but I’ll leave it up to you to decide if that makes sense

DelevoXDG: Added note

franciszekjob (approved):

ddoktorski (approved):
LGTM

PR #4123, opened 2026-01-22
Remove deprecated plugin; Drop support for Scarb < 2.12.0

DelevoXDG:
Closes #4057 

## Stack
- #4123
- #4124 

## Introduced changes

- drop support for Scarb < `2.12.0` (set minimal required version to `2.12.0`)
- remove deprecated components:
  - `snforge_scarb_plugin_deprecated`
  - `snforge_std_deprecated`
- remove workflow steps for publishing deprecated plugin and package

## Checklist

- [x] Linked relevant issue
- [ ] Updated relevant documentation
- [ ] Added relevant tests
- [x] Performed self-review of the code
- [x] Added changes to `CHANGELOG.md`

PR #4124, opened 2026-01-22
Update migration guide (0.56)

DelevoXDG:
## Stack
- #4123
- #4124 

## Introduced changes

-

## Checklist

- [ ] Linked relevant issue
- [ ] Added relevant tests


docs/src/appendix/0-56-0-migration-guide.md
```diff
@@ -0,0 +1,53 @@
...
+No further action is required.
+
+## Scarb Versions < 2.12.0
+
+> ⚠️ **Warning**
+>
+> **Support for Scarb versions < 2.12.0 has been removed in `snforge` 0.56.0.**
+>
+> If you are using Scarb < 2.12.0, you must upgrade to Scarb 2.12.0 or newer.
+> The deprecated `snforge_std_deprecated` package and `snforge_scarb_plugin_deprecated` plugin will no longer be supported, and no new versions of these will be published.
```
franciszekjob:
```suggestion
> The deprecated `snforge_std_deprecated` package and `snforge_scarb_plugin_deprecated` plugin are no longer be supported, and no new versions of these will be published.
```

docs/src/appendix/0-56-0-migration-guide.md
```diff
@@ -0,0 +1,53 @@
+# `snforge` 0.56.0 Migration Guide
+
+Starting from version 0.56.0, `snforge` requires Scarb version 2.12.0 or newer.
+This is due to the migration to the Scarb V2 version of procedural macros, which are used to handle arguments like `#[test]`
+in `snforge`.
+Thanks to this migration, tools like the Cairo plugin for VSCode shows better, more descriptive error.
```
franciszekjob:
```suggestion
Thanks to this migration, tools like the Cairo plugin for VSCode shows better, more descriptive errors.
```

franciszekjob (approved):
side note: I wonder if people will actually notice it, as it's on the very bottom 😆

docs/src/appendix/0-56-0-migration-guide.md
DelevoXDG:
```suggestion
> The deprecated `snforge_std_deprecated` package and `snforge_scarb_plugin_deprecated` plugin are no longer supported, and no new versions of these will be published.
```

docs/src/appendix/0-56-0-migration-guide.md
```diff
@@ -0,0 +1,53 @@
+# `snforge` 0.56.0 Migration Guide
```
MKowalski8, line 1: Do we want to name it `0.56.0` migration guide? It would only affect people who didn't migrate at `0.48.0`, which was half a year ago. Also scarb `2.12.0` is not officially supported. I think it is a good idea to have some info in our docs that we deleted it (`snforge_std_deprecated`), but I would communicated it differently. WDYT?

DelevoXDG: > It would only affect people who didn't migrate at `0.48.0`, which was half a year ago.

What do you mean? 

I think it would affect the people who **did migrate** at `0.48.0`, but didn't update Scarb, in line with what the `0.48.0` migration guide suggested.

https://github.com/foundry-rs/starknet-foundry/blob/8cf124f91e1100749b58f567ca6da3e63f2b8bb1/docs/src/appendix/0-48-0-migration-guide.md#L25-L25

We have here the path for those who:
1. Followed migration guide + have updated Scarb (bump dep)
2. Followed migration guide + chose to use older Scarb (undo 0.48 migration guide changes)
3. Didn't follow migration guide

Which to me seems like it should affect everybody 🤔  

> Also scarb 2.12.0 is not officially supported. 

WDYM? 

Afair, from our last offline discussion, the new `MINIMAL_SCARB_VERSION` should be `2.12.0`?

 
>  I think it is a good idea to have some info in our docs that we deleted it (snforge_std_deprecated), but I would communicated it differently. WDYT?

Can you please clarify what you mean? 

Do you want to remove this migration guide and use some entirely different approach to inform users about the change? 

To my mind, "migration guide" seems like the right place to include this info + I suppose it's something users would expect as well

ddoktorski: > Afair, from our last offline discussion, the new MINIMAL_SCARB_VERSION should be 2.12.0?

We have a somewhat vague distinction between the minimum supported and minimum recommended versions. Officially we only support the last three Scarb versions (so currently `>=2.13`). However, as long as it works with `2.12` it's fine to keep that version mentioned in the documentation. Initially I thought this migration guide wasn’t necessary as these changes won't affect many users, but in fact it’s worth having it for completeness. I'm also fine with keeping it at the bottom of the docs.

ddoktorski (approved):

docs/src/appendix/0-56-0-migration-guide.md
MKowalski8: > We have here the path for those who:
> 
>     Followed migration guide + have updated Scarb (bump dep)
>     Followed migration guide + chose to use older Scarb (undo 0.48 migration guide changes)
>     Didn't follow migration guide
> 
> Which to me seems like it should affect everybody 🤔

I mean it only affect people who still uses `snforge_std_deprecated`. 
People already using `snforge_std` won't notice any changes, and this probably applies mainly to our users.

> To my mind, "migration guide" seems like the right place to include this info + I suppose it's something users would expect as well

I agree, but since it won't affect that many users (at least I don't think so), I thought it wasn't a suitable name. 
However, as long as it's in appendix and others are okay with it, we can leave it as it is.

DelevoXDG: > I mean it only affect people who still uses `snforge_std_deprecated`.
> People already using `snforge_std` won't notice any changes, and this probably applies mainly to our users.

Well, explicitly changing `snforge_std`->`snforge_std_deprecated` is something we advised in the previous migration guide, so for everyone who did so, I think it makes sense to have this instruction that resolves this

Besides, it's still being used:

<a href="https://github.com/user-attachments/assets/ea195de8-b806-4cce-b219-786c68cd5d36" target="_blank"></a>

> it won't affect that many users (at least I don't think so),

IMO this guide will be checked mainly if someone has problem with an update, which will happen for those using deprecated dep.

And if someone else checks it, it's just bump dep + "No further action is required"

MKowalski8: > Well, explicitly changing snforge_std->snforge_std_deprecated is something we advised in the previous migration guide, so for everyone who did so, I think it makes sense to have this instruction that resolves this

We advised it for people who didn't want to upgrade their scarb version, but I would call it more like a "possibility" than "advise". We recommended to upgrade scarb on the first place and it was called `deprecated` to avoid encouraging its use. However it's not important at this point, as I said, we can leave it as it is 😅

> IMO this guide will be checked mainly if someone has problem with an update, which will happen for those using deprecated dep.

100% agree with that

MKowalski8 (approved):

PR #4127, opened 2026-01-23
Allow skiping backtrace validation

ksew1:


crates/forge/tests/e2e/snapshots/backtrace/main__e2e__backtrace__wrong_scarb_toml_configuration_with_skip_check@2.15.0.snap
ddoktorski, line 1: Please add snapshots also for other Scarb versions (all supported in scheduled tests)

crates/forge/src/profile_validation/mod.rs
```diff
@@ -59,3 +59,9 @@ fn contains_entry_with_value(table: &Table, key: &str, value: &str) -> bool {
         }
     })
 }
+
+/// Check if the `SNFORGE_SKIP_BACKTRACE_CHECK` environment variable is set to `true`.
+fn skip_backtrace_check() -> bool {
+    const SKIP_BACKTRACE_ENV: &str = "SNFORGE_SKIP_BACKTRACE_CHECK";
+    env::var(SKIP_BACKTRACE_ENV).is_ok_and(|value| value == "true")
```
ddoktorski, line 66: nit: Do we want to allow both `true` and `1`?

ksew1: I don't have a strong opinion on this. Whatever suits you best 😄

piotmag769 (commented):
Isn't it better to add an env to disable ALL checks (literally all of them)? I mean an env that says "I know what I am doing, trust me"

crates/forge/src/profile_validation/mod.rs
DelevoXDG: Perhaps we should add universal helper similarly to Scarb: 

https://github.com/software-mansion/scarb/blob/e546395b1e5c4b2f3dde0bd2257382ed5848e28c/scarb/src/process.rs#L213

We do check `1` here for example:
https://github.com/foundry-rs/starknet-foundry/blob/d05db9673239332b4b8d12188891045db25c92c7/crates/forge/src/main.rs#L47-L47

So ig it makes sense do to it here as well

DelevoXDG:
> Allow skiping backtrace validation

Can you pls provide some context on why this change exists? 😅

ksew1:
> > Allow skiping backtrace validation
> 
> Can you pls provide some context on why this change exists? 😅

Dm'ed you on slack

ksew1:
> Isn't it better to add an env to disable ALL checks (literally all of them)? I mean an env that says "I know what I am doing, trust me"

It could be useful in these cases. We can go with that if the other members of snfoundry agree. We would lose granularity, but we wouldn't need additional environment variables.

crates/forge/src/profile_validation/mod.rs
ksew1: Nice idea we can go with that

piotmag769:
>We would lose granularity

I believe we don't care about that, since it is a setting for power users saying "let me do what I want"

Issue #3837, opened 2025-10-20
Possible storage writes not reverted bug


cptartur:
### Current State

Starkware reported issue with storage writes not being reported.

### Objective

Investigate the possible bug

PR #4120, opened 2026-01-22
Make top call state revertible

ddoktorski:
Closes #3837

## Introduced changes

- Revert state for all failed calls

PR #4113, opened 2026-01-20
Refactor `CallToBlockifierRuntime` and related structs

ddoktorski:
Towards #3307 & #3837 

## Introduced changes

- Remove our custom logic for writing to VM memory
- Rename `Panic`, `Error` to `Recoverable` and `Unrecoverable`
- Make `CallToBlockifierRuntime` flow similar to other runtimes and Blockifier (e.g. `execute_syscall` function)

PR #4129, opened 2026-01-26
Handle generic error dynamically in should panic attribute

ddoktorski:
Towards #3837

### DIFF
--- a/crates/forge-runner/src/test_case_summary.rs
+++ b/crates/forge-runner/src/test_case_summary.rs
@@ -543,4 +543,31 @@ mod tests {
         assert!((stats.l1_data_gas.std_deviation - 16.33).abs() < FLOAT_ERROR);
         assert!((stats.l2_gas.std_deviation - 24.49).abs() < FLOAT_ERROR);
     }
+
+    #[test]
+    fn test_is_matching_should_panic_data_entrypoint_failed() {
+        let data = vec![
+            Felt::from(11_u8),
+            Felt::from(22_u8),
+            Felt::from(33_u8),
+            ENTRYPOINT_FAILED_ERROR_FELT,
+            ENTRYPOINT_FAILED_ERROR_FELT,
+        ];
+
+        assert!(is_matching_should_panic_data(&data, &data));
+
+        let non_matching_pattern = vec![
+            Felt::from(11_u8),
+            Felt::from(22_u8),
+            Felt::from(33_u8),
+            ENTRYPOINT_FAILED_ERROR_FELT,
+        ];
+        assert!(!is_matching_should_panic_data(&data, &non_matching_pattern));
+
+        let pattern = vec![Felt::from(11_u8), Felt::from(22_u8), Felt::from(33_u8)];
+        assert!(is_matching_should_panic_data(&data, &pattern));
+
+        let non_matching_pattern = vec![Felt::from(11_u8), Felt::from(22_u8)];
+        assert!(!is_matching_should_panic_data(&data, &non_matching_pattern));
+    }
 }
--- a/crates/forge-runner/src/test_case_summary.rs
+++ b/crates/forge-runner/src/test_case_summary.rs
@@ -7,9 +7,7 @@ use crate::gas::report::SingleTestGasInfo;
 use crate::gas::stats::GasStats;
 use crate::package_tests::with_config_resolved::TestCaseWithResolvedConfig;
 use crate::running::{RunCompleted, RunStatus};
-use blockifier::execution::syscalls::hint_processor::{
-    ENTRYPOINT_FAILED_ERROR_FELT, ENTRYPOINT_NOT_FOUND_ERROR_FELT,
-};
+use blockifier::execution::syscalls::hint_processor::ENTRYPOINT_FAILED_ERROR_FELT;
 use cairo_annotations::trace_data::VersionedCallTrace as VersionedProfilerCallTrace;
 use camino::Utf8Path;
 use cheatnet::runtime_extensions::call_to_blockifier_runtime_extension::rpc::UsedResources;
@@ -22,11 +20,6 @@ use starknet_api::execution_resources::GasVector;
 use starknet_types_core::felt::Felt;
 use std::fmt;
 use std::option::Option;
-
-const GENERIC_ERRORS: [Felt; 2] = [
-    ENTRYPOINT_FAILED_ERROR_FELT,
-    ENTRYPOINT_NOT_FOUND_ERROR_FELT,
-];
 
 #[derive(Debug, PartialEq, Clone, Default)]
 pub struct GasFuzzingInfo {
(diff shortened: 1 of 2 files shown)


ddoktorski:

- #4113
    - #4129 ◀
        - #4120

crates/forge-runner/src/test_case_summary.rs
```diff
@@ -391,14 +401,20 @@ fn join_short_strings(data: &[Felt]) -> String {
...
         data.contains(&pattern) // If both data and pattern are byte arrays, pattern should be a substring of data
     } else {
-        data == pattern // Otherwise, data should be equal to pattern
+        // Before comparing actual with expected data, remove generic errors from the data.
+        let filtered: Vec<Felt> = data
+            .iter()
+            .copied()
+            .filter(|e| !GENERIC_ERRORS.contains(e))
+            .collect();
+        filtered.as_slice() == pattern
```
DelevoXDG: I wonder if we should make it dynamic - so that these are allowed and validated *if included* in `should_panic`, but ignored otherwise?

The reason I'm asking is:
1. This will break a lot of tests. Sort of minor, but still an annoyance
2. Sometimes, it makes sense to expect for say `ENTRYPOINT_NOT_FOUND` and not some custom message, e.g.

https://github.com/OpenZeppelin/cairo-contracts/blame/4c1c8f26e1798f3f662dc8877efc59417b59a0b6/packages/presets/src/tests/test_erc20.cairo#L434C50-L434C50

ddoktorski: > This will break a lot of tests. Sort of minor, but still an annoyance

Generally this shouldn't break any existing tests. With the current logic we return only the actual error in the top level call (by using the `extract_trailing_cairo1_revert_trace` function). Changes in the subsequent PR would break most of the existing `#[should_panic]` tests without this modification. Given the example you provided I think we should limit this logic to remove only `ENTRYPOINT_FAILED` without `ENTRYPOINT_NOT_FOUND`.

DelevoXDG: > Changes in the subsequent PR would break most of the existing `#[should_panic]` tests without this modification.

Can you please clarify?

> With the current logic we return only the actual error in the top level call (by using the `extract_trailing_cairo1_revert_trace` function). 

Afaik `extract_trailing_cairo1_revert_trace` is not currently used to extract panic data for `should_panic`? 🤔 

https://github.com/foundry-rs/starknet-foundry/blob/7e455da7d9ba5d22a3f1ef05d803d8994088c7fb/crates/forge-runner/src/running.rs#L370-L370
 
The current logic (`filtered.as_slice() == pattern`) seems to strongly suggest that something like this will fail:
https://github.com/caddyfinance/Options-vault-contracts/blob/e38390c791c4435bebe0ec90196de605f32e66c1/src/Token.cairo#L331 

Perhaps we can add a test to check this?

ddoktorski: > Can you please clarify?

After making top level calls revertible, we'll get the actual error + `ENTRYPOINT_FAILED` for each level of the call stack above the one where the error occurred. Basically, as the stack unwinds this error gets added at each level here: https://github.com/foundry-rs/starknet-foundry/blob/0358d1950115e5d65a3ddab644581469ae7ce4ce/crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/calls.rs#L58

If we don't clean these errors each `should_panic` test that use expected array will fail after reverting top call e.g. before: `['panic']`, now: `['panic', 'ENTRYPOINT_FAILED', 'ENTRYPOINT_FAILED']`

> Afaik extract_trailing_cairo1_revert_trace is not currently used to extract panic data for should_panic? 🤔

It is used in `non_reverting_execute_call_entry_point`, so result uses for `should_panic` is already "extracted" :
https://github.com/foundry-rs/starknet-foundry/blob/0358d1950115e5d65a3ddab644581469ae7ce4ce/crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/entry_point.rs#L208

> The current logic (filtered.as_slice() == pattern) seems to strongly suggest that something like this will fail:
https://github.com/caddyfinance/Options-vault-contracts/blob/e38390c791c4435bebe0ec90196de605f32e66c1/src/Token.cairo#L331
Perhaps we can add a test to check this?

Checking 👀

ddoktorski: > The current logic (filtered.as_slice() == pattern) seems to strongly suggest that something like this will fail:
https://github.com/caddyfinance/Options-vault-contracts/blob/e38390c791c4435bebe0ec90196de605f32e66c1/src/Token.cairo#L331

Actually in this example they might be using `cairo-test` for testing

ddoktorski:
> I wonder if we should make it dynamic - so that these are allowed and validated if included in should_panic, but ignored otherwise?

As discussed during the call this approach has been implemented. Also only the `ENTRYPOINT_FAILED` error is being checked.

crates/forge-runner/src/test_case_summary.rs
```diff
@@ -391,14 +394,26 @@ fn join_short_strings(data: &[Felt]) -> String {
         .join(", ")
 }
 
-fn is_matching(data: &[Felt], pattern: &[Felt]) -> bool {
+fn is_matching_should_panic_data(data: &[Felt], pattern: &[Felt]) -> bool {
```
franciszekjob, line 402: I know we didn't have it but would be nice to add some unit tests tbh.

docs/src/testing/test-attributes.md
```diff
@@ -67,6 +67,11 @@ Asserting that the function panics (any with any panic data):
 #[should_panic]
 ```
 
+> 📝 **Note**
```
franciszekjob, line 70:
```suggestion
> 📝 **Note**
>
```

docs/src/testing/test-attributes.md
```diff
@@ -67,6 +67,11 @@ Asserting that the function panics (any with any panic data):
...
+> `#[should_panic]` dynamically validates the generic `ENTRYPOINT_FAILED` error.
+> If it's included in the `expected` data, the full panic data (including all occurrences of `ENTRYPOINT_FAILED`) will be asserted.
+> Otherwise, these generic errors will be skipped and only the actual, user-defined error will be checked.
```
franciszekjob: nit: Seems a bit more readable:
```suggestion
> By default, `#[should_panic]` ignores generic `ENTRYPOINT_FAILED` errors and validates only the user-defined error.
> If `ENTRYPOINT_FAILED` is explicitly listed in the expected data, the full panic payload is asserted instead.
```

crates/forge-runner/src/test_case_summary.rs
```diff
@@ -391,14 +394,26 @@ fn join_short_strings(data: &[Felt]) -> String {
...
     let data_str = convert_felts_to_byte_array_string(data);
     let pattern_str = convert_felts_to_byte_array_string(pattern);
 
     if let (Some(data), Some(pattern)) = (data_str, pattern_str) {
         data.contains(&pattern) // If both data and pattern are byte arrays, pattern should be a substring of data
     } else {
-        data == pattern // Otherwise, data should be equal to pattern
+        // Compare logic depends on the presence of `ENTRYPOINT_FAILED_ERROR` in the expected data.
+        if pattern.contains(&ENTRYPOINT_FAILED_ERROR_FELT) {
```
DelevoXDG, line 410: Do you think it's possible this list is to be expanded anytime in the future with any of the other errors?

https://github.com/software-mansion-labs/sequencer/blob/c2e2dff39f7105ef3adb622e7e38cf9a1ebbd996/crates/blockifier/src/execution/syscalls/hint_processor.rs#L222

Or is it more like `ENTRYPOINT_NOT_FOUND` case? 

I just wonder if it makes sense to generalize this approach so that we're checkin and filtering the entire list 🤔

docs/src/testing/test-attributes.md
DelevoXDG: I'm not sure which one is more readable and whether order matters here. 

But, regardless of that, I'd includek snippets for both so it's immediately clear what are we referring to
```rust
#[should_panic(expected: ('panic message', 'ENTRYPOINT_FAILED')]
```
```rust
#[should_panic(expected: ('panic message')]
```

DelevoXDG (approved):

crates/forge-runner/src/test_case_summary.rs
ddoktorski: It’s hard to say, probably more like `ENTRYPOINT_NOT_FOUND`, but who knows. Anyway I'd only look into it if it actually occurs, not now

DelevoXDG (approved):

franciszekjob (approved):

PR #4109, opened 2026-01-19
Partitioning logic

franciszekjob:
Towards #3548

**Stack**
- #4150
- #4110
- #4112
- #4109 ⬅️ 
- #4108
- #4107
- #4106

## Introduced changes

- Add partitioning-related logic
- Tests

Benchmarks (on OpenZeppelin package [v3.0.0](https://github.com/OpenZeppelin/cairo-contracts/releases/tag/v3.0.0)) when running `snforge test --workspace --features fuzzing --fuzzer-runs 200`:

| Scenario | Cache | Job 1 | Job 2 | Job 3 | Overall CI wall-clock (approx.) | Sum of job times | Average job time |
|---|---:|---:|---:|---:|---:|---:|---:|
| **Single run (no partitioning)** | **No** | — | — | — | **10m 34s** | **10m 34s** | **10m 34s** |
| **Single run (no partitioning)** | **Yes** | — | — | — | **7m 53s** | **7m 53s** | **7m 53s** |
| **Partitioned (3 jobs)** | **No** | 7m 54s | 8m 17s | 7m 58s | **8m 17s** | **24m 09s** | **8m 03s** |
| **Partitioned (3 jobs)** | **Yes** | 5m 34s | 5m 34s | 5m 46s | **5m 46s** | **16m 54s** | **5m 38s** |

## Checklist

- [x] Linked relevant issue
- [ ] Updated relevant documentation
- [x] Added relevant tests
- [x] Performed self-review of the code
- [ ] Added changes to `CHANGELOG.md`

PR #4134, opened 2026-01-27
Run schedule_compile_raw_sierra_at_path preemptively on another thread

maciektr:
On cairo-contracts:

Before:
```
Time (mean ± σ):     46.300 s ±  1.312 s    [User: 275.481 s, System: 22.427 s]  Range (min … max):   45.293 s … 48.290 s    5 runs
```
image

After: 
```
Time (mean ± σ):     36.926 s ±  2.170 s    [User: 279.394 s, System: 28.798 s]  Range (min … max):   35.337 s … 40.697 s    5 runs
```
image

### DIFF
--- a/crates/forge/src/run_tests/workspace.rs
+++ b/crates/forge/src/run_tests/workspace.rs
@@ -15,6 +15,8 @@ use anyhow::{Context, Result};
 use forge_runner::test_case_summary::AnyTestCaseSummary;
 use forge_runner::{CACHE_DIR, test_target_summary::TestTargetSummary};
 use foundry_ui::UI;
+use rayon::iter::IntoParallelRefIterator;
+use rayon::iter::ParallelIterator;
 use scarb_api::metadata::{MetadataOpts, metadata_with_opts};
 use scarb_api::{
     metadata::{Metadata, PackageMetadata},
@@ -24,6 +26,7 @@ use scarb_ui::args::PackagesFilter;
 use shared::consts::SNFORGE_TEST_FILTER;
 use std::env;
 use std::sync::Arc;
+use universal_sierra_compiler_api::schedule_compile_raw_sierra_at_path;
 
 #[tracing::instrument(skip_all, level = "debug")]
 pub async fn run_for_workspace(args: TestArgs, ui: Arc<UI>) -> Result<ExitStatus> {
@@ -80,17 +83,29 @@ pub async fn run_for_workspace(args: TestArgs, ui: Arc<UI>) -> Result<ExitStatus
     let cache_dir = workspace_root.join(CACHE_DIR);
     let packages_len = packages.len();
 
-    for package in packages {
-        env::set_current_dir(&package.root)?;
+    let packages_with_args = packages
+        .par_iter()
+        .map(|package: &PackageMetadata| {
+            let args = RunForPackageArgs::build(
+                package.clone(),
+                &scarb_metadata,
+                &args,
+                &cache_dir,
+                &artifacts_dir_path,
+                &ui,
+            )?;
+            Ok((package, args))
+        })
+        .collect::<Result<Vec<_>>>()?;
 
-        let args = RunForPackageArgs::build(
-            package,
-            &scarb_metadata,
-            &args,
-            &cache_dir,
-            &artifacts_dir_path,
-            &ui,
-        )?;
+    for (_, args) in &packages_with_args {
+        for test_target in &args.test_targets {
+            schedule_compile_raw_sierra_at_path(test_target.sierra_program_path.as_ref())?;
+        }
+    }
+
+    for (package, args) in packages_with_args {
+        env::set_current_dir(&package.root)?;
 
         let result = run_for_package(args, &mut block_number_map, ui.clone()).await?;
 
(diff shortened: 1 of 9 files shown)


crates/forge/src/run_tests/workspace.rs
```diff
@@ -91,17 +94,29 @@ pub async fn run_for_workspace(args: TestArgs, ui: Arc<UI>) -> Result<ExitStatus
     let cache_dir = workspace_root.join(CACHE_DIR);
     let packages_len = packages.len();
 
-    for package in packages {
-        env::set_current_dir(&package.root)?;
+    let packages_with_args = packages
+        .par_iter()
+        .map(|package: &PackageMetadata| {
+            let args = RunForPackageArgs::build(
```
DelevoXDG, line 89: There was an issue in #4109 that I think was caused by similar logic:
```rs
let mut test_full_paths: Vec<String> = packages
            .par_iter()
            .map(|package| -> Result<Vec<String>> {
                let raw_test_targets = load_test_artifacts(artifacts_dir, package)?;
                ...
```
Trying to load multiple artifacts in parallel caused CI on OZ `cairo-contracts` repeatedly failed for larger packages (e.g. `openzeppelin_governance` with their 649 tests):

https://github.com/franciszekjob/cairo-contracts/actions/runs/21204513806/job/60997671365

```
/home/runner/work/_temp/a551415d-0a83-41d7-940e-d0be312184c0.sh: line 1:  2362 Killed                  snforge test --partition '2/3' --workspace --features fuzzing --fuzzer-runs 200
Error: The operation was canceled.
```
Note that this was working fine locally, but always crashed on CI, likely due to system limitations of runners.

I suspect this problem is very likely to be present here as well, as `RunForPackageArgs::build` uses `load_test_artifacts` as well.

With than in mind, I think it would make sense to:
- Build a nightly and test whether this works with OZ workflow on github
- Consider some other approach here (it was just decided to [remove](https://github.com/foundry-rs/starknet-foundry/pull/4109/changes/e6bfa9c6811636e85a00b19f1445d3801e218264) `par_iter` in case of mentioned PR; but perhaps limiting number of threads could work as well)

crates/universal-sierra-compiler-api/src/lib.rs
```diff
@@ -37,12 +41,63 @@ pub fn compile_raw_sierra(sierra_json: &Value) -> Result<RawCasmProgram, Compila
...
-    let json = compile_sierra_at_path(sierra_file_path, SierraType::Raw)?;
-    serde_json::from_str(&json).map_err(CompilationError::Deserialization)
+) -> anyhow::Result<RawCasmProgram> {
+    let cell = SIERRA_RAW_COMPILATION_DATA_LOADER.get_or_init(OnceMap::default);
+    let path = sierra_file_path.to_path_buf();
+    let span = trace_span!("waiting_for_compiled_raw_sierra_at_path");
+    let result = {
+        let _g = span.enter();
+        cell.wait_blocking(&path)
+            .expect("schedule_compile_raw_sierra_at_path must be called first")
```
DelevoXDG, line 109: 1. I wonder if we should just try to compile inline if `schedule_compile_raw_sierra_at_path` was not run before?
2. Why not `?`

crates/universal-sierra-compiler-api/src/lib.rs
```diff
@@ -37,12 +41,63 @@ pub fn compile_raw_sierra(sierra_json: &Value) -> Result<RawCasmProgram, Compila
...
 
-/// Compiles Sierra JSON file at the given path of a raw program into [`RawCasmProgram`].
-pub fn compile_raw_sierra_at_path(
+static SIERRA_RAW_COMPILATION_DATA_LOADER: OnceLock<
+    OnceMap<PathBuf, Result<RawCasmProgram, CompilationErrorString>>,
+> = OnceLock::new();
+
+// We need a cloneable error type for the `OnceMap` to work.
+#[derive(Clone)]
+struct CompilationErrorString(String);
```
DelevoXDG, line 52: This can work, but perhaps using cloneable error would be better?
https://github.com/software-mansion/scarb/blob/22357ecabd43c06595f439b3e75e3032edf1dfc7/scarb/src/internal/cloneable_error.rs#L7-L7

crates/universal-sierra-compiler-api/src/lib.rs
maciektr: We could do that, but asserting on explicit scheduling forces the dev to make a conscious decision on when the compilation should be started. I worry, that implicit scheduling would become a norm over time if allowed - you need to do some profiling rounds and focus on this part of the whole flow to catch the difference.

crates/universal-sierra-compiler-api/src/lib.rs
maciektr: Forgot about this one, thanks!

crates/universal-sierra-compiler-api/src/lib.rs
DelevoXDG: > asserting on explicit scheduling forces the dev to make a conscious decision on when the compilation should be started.

I guess that's a valid point. 

I'd still propagate the error rather than panicking here though 🤔

crates/forge/src/run_tests/workspace.rs
franciszekjob: Confirming, as @DelevoXDG says - running `load_test_artifacts` in parallel worked without any problems locally, but it was too heavy for Github runner (at least on my personal fork) and ultimately caused it to crash.

PR #4148, opened 2026-02-04
Snap tests CI: run in merge queue

DelevoXDG:
Closes #

## Stack
- #4148 
- #4158
- #4159

## Introduced changes

- add `snap_` prefix to snap tests
- run snap tests when adding to merge queue
- ✅ Test run: https://github.com/foundry-rs/starknet-foundry/actions/runs/21675768894/job/62495624725

## Checklist

- [ ] Linked relevant issue
- [ ] Added relevant tests


.github/workflows/ci.yml
```diff
@@ -213,6 +233,51 @@ jobs:
...
+  test-forge-e2e-snap:
+    if: github.event_name == 'merge_group'
+    name: Test Forge / E2E Snap (Scarb ${{ matrix.version }})
+    runs-on: ubuntu-latest
+    needs: get-scarb-versions
+    strategy:
+      fail-fast: false
+      matrix:
+        version: ${{ fromJSON(needs.get-scarb-versions.outputs.versions) }}
+
```
MKowalski8, line 245: Is tnere a way to not run it for scarb in `tool_version?` I mean those test are run normaly in `e2e` tests and it would just repeat them. It is not a big deal, but if it can be done with one `if` or so. It would be okay to do this I think. 

Also I'm wondering if we should run this for version newer than `.tool_version`, because on the one hand it is okay to know that we don't have snaps for new scarb, but on the other it can block us with PR's that are not releated to this and will force us to update them as soon as possible. Like now for example. there is already `scarb 2.16.0` and we don't have snaps on main yet, so this PR couldn't be merged in that case.

DelevoXDG: Added `--previous` flag to `get_scarb_versions.sh`, will now only run for older then current

.github/workflows/ci.yml
```diff
@@ -13,6 +13,26 @@ concurrency:
...
+      versions: ${{ steps.get_versions.outputs.versions }}
+    steps:
+      - uses: actions/checkout@v6
+      - uses: asdf-vm/actions/install@05e0d2ed97b598bfce82fd30daf324ae0c4570e6
+        with:
+          tool_versions: |
+            scarb latest
+
+      - name: Get versions
+        id: get_versions
```
franciszekjob, line 30: nit: Maybe a bit more precisely?
```suggestion
        id: get_scarb_versions
```

.github/workflows/ci.yml
```diff
@@ -213,6 +233,51 @@ jobs:
...
+    strategy:
+      fail-fast: false
+      matrix:
+        version: ${{ fromJSON(needs.get-scarb-versions.outputs.versions) }}
+
+    steps:
+      - name: Extract branch name
+        if: github.event_name != 'pull_request'
+        run: echo "BRANCH_NAME=$(echo ${GITHUB_REF#refs/heads/})" >> $GITHUB_ENV
+        shell: bash
```
franciszekjob, line 250: nit: If bash is found it will be used by default, so can get rid of it (similarly below).
```suggestion
```

scripts/get_scarb_versions.sh
```diff
@@ -13,13 +16,47 @@ function get_latest_patch_version() {
   get_all_patch_versions "$1" | sort -uV | tail -1
 }
 
+function version_less_than() {
+  [[ "$1" != "$current" ]] && [[ "$(printf '%s\n%s' "$1" "$current" | sort -V | head -1)" == "$1" ]]
```
franciszekjob: Wouldn't it be better if `current` was an explicit param?

.github/workflows/ci.yml
DelevoXDG: I copied it from scheduled workflow.

Please note that the job is called `get-scarb-versions` already, and this is just a step in a job.

I don't know whether having same job and step ID would cause error or not, but as a step in a job - I think the name is descriptive enough.

WDYT? 🤔

.github/workflows/ci.yml
DelevoXDG: Hmmm, I generally agree, but `shell: bash` is set in plenty of other places 🤔 

I don't know whether there's some intention behind that or not. If there isn't, I'd rather drop it everywhere at once as a separate issue.

WDYT?

scripts/get_scarb_versions.sh
DelevoXDG: Sure, updated

.github/workflows/ci.yml
franciszekjob: Ok, disregard.

.github/workflows/ci.yml
franciszekjob: > I don't know whether there's some intention behind that or not.

Maybe there is a scenario in which it's actually needed, let's keep it then, doesn't make any harm.

franciszekjob (approved):

MKowalski8 (approved):

PR #4112, opened 2026-01-20
Add partitioning messages

franciszekjob:
Towards #3548

**Stack**
- #4150
- #4110
- #4112 ⬅️ 
- #4109
- #4108
- #4107
- #4106

## Introduced changes

Include messages about start and end of partitioned run during tests

## Checklist

- [x] Linked relevant issue
- [ ] Updated relevant documentation
- [x] Added relevant tests
- [x] Performed self-review of the code
- [ ] Added changes to `CHANGELOG.md`

PR #4110, opened 2026-01-19
Partitioning docs

franciszekjob:
Towards #3548

**Stack**
- #4150
- #4110 ⬅️ 
- #4112
- #4109
- #4108
- #4107
- #4106

## Introduced changes

Add docs about partitioning

## Checklist

- [x] Linked relevant issue
- [x] Updated relevant documentation
- [x] Added relevant tests
- [x] Performed self-review of the code
- [x] Added changes to `CHANGELOG.md`

PR #4108, opened 2026-01-19
Add partition config

franciszekjob:
Towards #3548

**Stack**
- #4150
- #4110
- #4112
- #4109
- #4108 ⬅️ 
- #4107
- #4106

## Introduced changes

Add new partition-related data types and logic for collecting test case full paths

## Checklist

- [x] Linked relevant issue
- [ ] Updated relevant documentation
- [ ] Added relevant tests
- [x] Performed self-review of the code
- [ ] Added changes to `CHANGELOG.md`

PR #4107, opened 2026-01-19
Refactor tests filtering

franciszekjob:
Towards #3548

**Stack**
- #4150
- #4110
- #4112
- #4109
- #4108
- #4107 ⬅️ 
- #4106
## Introduced changes

Little refactor for filtering

## Checklist

- [x] Linked relevant issue
- [ ] Updated relevant documentation
- [ ] Added relevant tests
- [x] Performed self-review of the code
- [ ] Added changes to `CHANGELOG.md`

PR #4150, opened 2026-02-04
Display partition tests count

### LINKED ISSUE #3548: Support test partitioning

franciszekjob:
Closes #3548

**Stack**
- #4150 ⬅️
- #4110 
- #4112
- #4109
- #4108
- #4107
- #4106

## Introduced changes

Display number of tests included in currently run partition

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/forge-runner/src/tests_summary.rs
```diff
@@ -43,9 +49,16 @@ impl TestsSummary {
...
-        format!(
+        let summary = format!(
             "{} passed, {} failed, {} ignored, {filtered} filtered out{interrupted}",
             self.passed, self.failed, self.ignored,
-        )
+        );
+
+        if let Some(skipped) = self.skipped {
+            let skipped_str = format!(", {skipped} skipped");
+            format!("{summary}{skipped_str}")
```
ddoktorski: It won't be well formatted if `interrupted` is not `None`

ddoktorski (commented):
* I'm not sure if distinction between `filtered out` and `skipped` makes sense to me in this context
* It's definitely unintuitive in the code that `skipped` is used for partition, but not for the `--skip` flag

franciszekjob:
> * I'm not sure if distinction between `filtered out` and `skipped` makes sense to me in this context
> * It's definitely unintuitive in the code that `skipped` is used for partition, but not for the `--skip` flag

We could add an explicit status "excluded"/"excluded from partition" - should be clearer. Wdyt @ddoktorski ?

crates/forge-runner/src/tests_summary.rs
franciszekjob: Fixed

ddoktorski:
>We could add an explicit status "excluded"/"excluded from partition" - should be clearer. 

Can we include this information directly in the partition message instead? Something like:
```
Finished partition run: 1/3, executed 5 out of total 14 test.
```

franciszekjob:
> > We could add an explicit status "excluded"/"excluded from partition" - should be clearer.
> 
> Can we include this information directly in the partition message instead? Something like:
> 
> ```
> Finished partition run: 1/3, executed 5 out of total 14 test.
> ```

Ok, makes sens. Slight adjustment:

```
Finished partition run: 1/3, included 5 out of total 14 tests
```

crates/forge-runner/src/partition.rs
```diff
@@ -145,6 +145,19 @@ impl PartitionMap {
...
+    pub fn count_partition_tests(&self, run_partition: NonZeroUsize) -> (usize, usize) {
+        let included = self
+            .0
+            .values()
+            .filter(|assigned_partition| **assigned_partition == run_partition)
+            .count();
+        let excluded = self.0.len() - included;
+
+        (included, excluded)
+    }
```
DelevoXDG: Nit:
I'd wrap that into helper struct that keeps `included` and `excluded` fields - would be slightly more readable.

crates/forge/src/run_tests/messages/partition.rs
```diff
@@ -4,23 +4,35 @@ use foundry_ui::{Message, components::labeled::LabeledMessage};
...
     }
 
     fn summary(&self) -> String {
-        partition_summary(&self.partition)
+        format!(
+            "{}/{}, included {} out of total {} tests",
+            self.partition.index(),
+            self.partition.total(),
+            self.included_tests_count,
+            self.excluded_tests_count + self.included_tests_count
```
DelevoXDG: Nit:

it's fine as is, but I noticed we do:
```rs
let excluded = self.0.len() - included;
```
only to later
```rs
self.excluded_tests_count + self.included_tests_count
```

Since we're not actually using `excluded_tests_count` anywhere, we could replace `PartitionMap::count_partition_tests` with `PartitionMap::included_tests_count(partition) -> usize` and `PartitionMap::total_tests_count() -> usize`

It seems like these would be a bit easier to reason about, and we wouldn't be computing anything we don't need.

DelevoXDG (approved):

MKowalski8 (approved):

PR #4154, opened 2026-02-09
Remove starkscan option as block explorer

franciszekjob:
Closes #

## Introduced changes

httos://starkscan.co has been terminated, hence we need to remove it as option for `block-explorer` field in `snfoundry.toml`

## Checklist

- [ ] Linked relevant issue
- [ ] Added relevant tests


design_documents/contract_verification.md
```diff
@@ -48,7 +48,7 @@ Specifies the Blockchain Explorer to verify with.
 
 Options as of writing of this document: 
  - voyager
- - starkscan
+ - walnut
```
franciszekjob, line 51: note: Typo, `starkscan` shouldn't be listed here.

crates/sncast/tests/e2e/show_config.rs
```diff
@@ -66,7 +66,7 @@ async fn test_show_config_from_cli_and_snfoundry_toml() {
         Wait Timeout:        300s
         Wait Retry Interval: 5s
         Show Explorer Links: true
-        Block Explorer:      Voyager
+        Block Explorer:      ViewBlock
```
franciszekjob, line 69: note: When `block-explorer` is set to a non-default value, it gets displayed in `show-config` command.  Otherwise, this field is omitted, hence this change.

MKowalski8 (approved):

crates/sncast/src/helpers/block_explorer.rs
DelevoXDG, line 1: I think the error could be a bit surprising to the users since there was no deprecation period for the service.

I wonder whether we should have one test case that checks behavior with `starkscan` with in `snfoundry.toml`?

If the error is already reasonable, we can keep it as it.

If it's not, I think it would be useful for the user to see some message, e.g.:
```
starkscan.co was terminated.
please set `block-explorer` to `"Voyager"` or other explorer of your choice.
```

WDYT?

franciszekjob: Sure, added.

DelevoXDG (approved):

PR #4155, opened 2026-02-11
Add `--sierra-file` to `declare-from` (declare contract from sierra path)

### LINKED ISSUE #4077: Allow declaring contracts from path
### Select Tool

sncast

### Requested Feature

Currently it is possible to declare contracts from local Scarb project (with `sncast declare`) or redeclaring existing contract from one network on another (with `sncast declare-from`). We can also enable declaring a contract by providing a path to compiled Sierra. It can be done as a separate command or as an extension of `declare-from`. The default flow with `declare` command should remain reserved only for Scarb projects.

DelevoXDG:
## Stack

- #4162
- #4155

Closes #4077 

## Introduced changes

Update `declare-from` command to accept `--sierra-file`.
This flag conflicts with all the flags used in "declare from network" mode.  

## Note on `starkli` parity

`starkli` also allowed to optionally specify the casm file, or even the casm hash itself, directly.

This is problematic because we now have: 
- network-version-based hash computations
- casm produced by USC that may differ from the one Scarb produces

Having similar options in `sncast` highly increases the likelihood of users shooting themselves in the foot (get `Compiled class hash mismatch` errors), and I don't see good reasons to let this to happen 🤔  LMK if you have alternative thoughts on that

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/sncast/src/starknet_commands/declare.rs
```diff
@@ -87,6 +88,24 @@ pub async fn declare(
     .await
 }
 
+#[allow(clippy::result_large_err)]
+pub fn compile_casm_from_sierra(
```
franciszekjob: nit:
```suggestion
pub fn compile_sierra_to_casm(
```

crates/sncast/src/main.rs
```diff
@@ -150,6 +152,9 @@ enum Commands {
     /// Declare a contract by fetching it from a different Starknet instance
     DeclareFrom(DeclareFrom),
 
+    /// Declare a contract from a compiled Sierra file
+    DeclareFile(DeclareFile),
+
```
franciszekjob: How about consolidating this into a `declare-from` command rather than introducing a separate `declare-file` command? My points:
1. From a pragmatic standpoint, `declare-from` feels more natural because it focuses on the action (declaring) while remaining flexible about the source - whether it's file with Sierra, a class hash from another chain, etc.
2. Using flags to specify the source is generally more ergonomic than bloating the CLI with more top-level commands that serve same purpose. We didn't do this while introducing `declare-from` as ultimately `declare` would be too complicated, but I think in this case it may be applicable.

WDYT?

CHANGELOG.md
```diff
@@ -12,6 +12,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 #### Added
 
 - Partitioned test execution with `--partition <INDEX>/<TOTAL>` flag. Read more [here](https://foundry-rs.github.io/starknet-foundry/snforge-advanced-features/tests-partitioning.html)
+- `sncast declare-file` command to declare a contract from compiled Sierra contract file
+
```
franciszekjob: very nit:
```suggestion
- `sncast declare-file` command to declare a contract from compiled Sierra contract file
```

docs/src/appendix/sncast/declare_from.md
```diff
@@ -26,7 +39,7 @@ Possible values: `mainnet`, `sepolia`, `devnet`.
 
 Overrides network from `snfoundry.toml`.
 
-## `--source-url, -u <RPC_URL>`
+## `--source-url, -s <RPC_URL>`
```
DelevoXDG, line 42: note: unrelated typo fix (here and few changes below)

crates/sncast/src/main.rs
DelevoXDG: Discussed offline as the way to go

CHANGELOG.md
```diff
@@ -15,6 +15,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Cast
 
+#### Added
+- `--sierra-file` flag to `sncast declare-from` to allow declaring contract from a local Sierra file.
```
franciszekjob, line 20: very nit:
```suggestion
#### Added

- `--sierra-file` flag to `sncast declare-from` to allow declaring contract from a local Sierra file.
```

crates/sncast/tests/e2e/declare_from.rs
```diff
@@ -217,3 +225,236 @@ async fn test_invalid_block_id() {
         "},
     );
 }
```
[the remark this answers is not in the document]
DelevoXDG: Doesn't appear to be related to the changes introduced in this PR

docs/src/appendix/sncast/declare_from.md
```diff
@@ -1,12 +1,25 @@
...
-Declare a contract by fetching it from a different Starknet instance.
+Declare a contract either: 
+- from a compiled Sierra file
+- by fetching it from another Starknet instance.
+
+The allowed args depend on the chosen contract source:
+- **File:** `--sierra-file` (required)
+- **Network:** `--class-hash` (required), `--block-id` (optional),  `--source-url` (optional), `--source-network` (optional)
+
+Note: **file** and **network** args are mutually exclusive.
```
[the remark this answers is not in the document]
DelevoXDG: That's how it was previously described in the docs under the individual flags. We can specify it there, but I'd rather keep it simple at top level

@franciszekjob should we correct this or will users just have to find out through interacting with the command?

crates/sncast/src/starknet_commands/declare.rs
```diff
@@ -94,11 +95,29 @@ pub async fn declare(
...
+    sierra_class: &SierraClass,
+) -> Result<CompiledClass, StarknetCommandError> {
+    let casm_json: String = serde_json::to_string(
+        &compile_contract_sierra(
+            &serde_json::to_value(sierra_class)
+                .with_context(|| "Failed to convert sierra to JSON value".to_string())?,
+        )
+        .with_context(|| "Failed to compile sierra to casm".to_string())?,
+    )
+    .expect("serialization should succeed");
```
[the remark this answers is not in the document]
DelevoXDG: That's how it was done before and it's not really error that we should be able to handle 🤔

crates/sncast/src/starknet_commands/declare_from.rs
```diff
@@ -76,56 +89,73 @@ impl SourceRpcArgs {
...
+            let flattened_sierra = match class {
+                ContractClass::Sierra(c) => c,
+                ContractClass::Legacy(_) => {
+                    return Err(StarknetCommandError::UnknownError(anyhow::anyhow!(
+                        "Declaring from Cairo 0 (legacy) contracts is not supported"
+                    )));
+                }
+            };
+            let sierra: SierraClass = flattened_sierra_to_sierra(flattened_sierra)
+                .expect("Failed to parse flattened sierra class");
```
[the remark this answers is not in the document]
DelevoXDG: As above, unrelated to the current changes

franciszekjob (approved):
We can mention `--sierra-file` in [docs](https://foundry-rs.github.io/starknet-foundry/starknet/declare.html) too.
Otherwise👌

docs/src/appendix/sncast/declare_from.md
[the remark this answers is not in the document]
franciszekjob: I think we can leave it as it was. Message from CLI about lack of one of them should be enough.

ddoktorski (approved):

PR #4160, opened 2026-02-13
Add basic debugger integration

### LINKED ISSUE #3: Integrate debugger with snforge - only tests with no contract calls
The scope is to enable snforge to be run in debug mode e.g. via `snforge test --exact pkg::my_test --debug`. Instead of starting the execution right away and finishing it on its own, snforge should open a tcp port and listen to DAP requests from the client - just like [here](https://github.com/software-mansion-labs/cairo-debugger/blob/dd6bb8b091408410eb886e0796d3ae9f1865446b/src/lib.rs#L15).

Requests can stay unhandled for now.

The architectural approach to this is TBD:
- one idea is to wrap `CairoRunner` with `CairoDebugger` and implement a common trait - may be hard to maintain when cairo-vm changes `CairoRunner` API
- other one is to add a function that will execute after each step of the cairo-vm - it would check for the DAP requests, handle them etc. This is the approach that @mkaput proposed - easier to maintain, but the execution can be much slower (we don't know how much slower though and there are prob optimizations we can make like "skip checking the tcp port for now since we just handled `Continue` request, so we have to continue until we hit a breakpoint anyways")

piotmag769:
Closes https://github.com/software-mansion/cairo-debugger/issues/3

## Introduced changes

Enables launching a test execution in a debug mode, which makes snforge act as debug adapter that can communicate via DAP with e.g. VSCode using https://github.com/software-mansion/vscode-cairo/pull/152. Enables debugging only test that don't contain contract calls - all the contract calls parts will be skipped, since they are executed on separate VMs.

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/forge/src/combine_configs.rs
```diff
@@ -1,280 +1,53 @@
...
             execution_data_to_save,
-            gas_report: gas_report || forge_config_from_scarb.gas_report,
+            gas_report: args.gas_report || forge_config_from_scarb.gas_report,
         }),
     }
 }
-
-#[cfg(test)]
-mod tests {
-    use super::*;
```
piotmag769, line 69: I removed these since they are not really useful

crates/forge/src/lib.rs
```diff
@@ -225,6 +225,13 @@ pub struct TestArgs {
 
     #[command(flatten)]
     scarb_args: ScarbArgs,
+
+    /// Launch the given test in debug mode using cairo-debugger.
+    ///
+    /// It makes snforge act as a debug adapter, enabling communication with an editor/IDE
+    /// such as VSCode over DAP protocol.
```
piotmag769: Idk if it isn't too technical, wdyt?

Cargo.toml
```diff
@@ -131,3 +131,7 @@ tracing-chrome = "0.7.2"
 tracing-log = "0.2.0"
 comfy-table = "7"
 insta = { version = "1.46.0", features = ["filters"] }
+cairo-debugger = { git = "https://github.com/software-mansion-labs/cairo-debugger", rev = "975cb93f9e72b621f711bc746b836e551bc317d8" }
```
piotmag769: I believe pinning to git is fine - saves the trouble of releasing on crates.io, especially since we use a custom dap-rs fork in the debugger (git dep, so I cannot publish). I can make tags on git that would act as releases as well

docs/src/snforge-advanced-features/debugging.md
```diff
@@ -1,11 +1,84 @@
...
+> 📝 **Note**
+>
+> `--launch-debugger` requires `--exact` - only a single test can be debugged at a time.
+
+### Prerequisites
+
+Live debugging relies on debug information provided by Scarb. To generate the necessary debug information, you need
+to have:
+
+1. [Scarb](https://github.com/software-mansion/scarb) version `2.17.1` or higher
```
piotmag769: TBD, I need to convince Ori to release 2.17.1 with my fix

docs/book.toml
```diff
@@ -13,7 +13,8 @@ traverse-parent-directories = false
 exclude = [
     "foundry-rs.github.io/starknet-foundry/snforge_std",
     "foundry-rs.github.io/starknet-foundry/sncast_std",
-    "starknet-faucet.vercel.app"
+    "starknet-faucet.vercel.app",
+    "marketplace.visualstudio.com"
```
piotmag769, line 17: They seem to reject bot requests

Cargo.toml
```diff
@@ -53,7 +54,7 @@ cairo-lang-test-plugin = "=2.17.0"
 cairo-lang-starknet-classes = "=2.17.0"
 cairo-lang-parser = "=2.17.0"
 cairo-lang-sierra-to-casm = "=2.17.0"
-cairo-vm = "=3.2.0"
+cairo-vm = { version = "=3.2.0", features = ["test_utils"] }
```
piotmag769, line 57: I need this for `cairo-debugger`. I don't like adding whole `test_utils` since it adds `#[derive(Arbitrary)]` trait to vm, but it shouldn't be a problem behaviour-wise. Checked `snforge` binary size and it is still the same, so that's also not a problem

crates/forge/src/lib.rs
```diff
@@ -244,6 +244,13 @@ pub struct TestArgs {
 
     #[command(flatten)]
     scarb_args: ScarbArgs,
+
+    /// Launch the given test in debug mode using `cairo-debugger` crate.
+    ///
+    /// It makes snforge act as a debug adapter, enabling communication with an editor/IDE
+    /// such as `VSCode` over DAP protocol.
+    #[arg(long, requires = "exact")]
```
ddoktorski, line 252:
```suggestion
    #[arg(long, requires = "exact", conflicts_with = "run_native")]
```

crates/forge/tests/e2e/mod.rs
```diff
@@ -14,6 +14,7 @@ mod components;
 mod contract_artifacts;
 #[cfg(not(feature = "cairo-native"))]
 mod coverage;
+mod debugger;
```
ddoktorski, line 18: nit:
```suggestion
#[cfg(not(feature = "cairo-native"))]
mod debugger;
```

docs/src/snforge-advanced-features/debugging.md
```diff
@@ -1,11 +1,84 @@
...
+skip-optimizations = true
+```
+
+### Debugging in VSCode
+
+1. Open your Cairo project in VSCode.
+2. Make sure the latest [Cairo extension](https://marketplace.visualstudio.com/items?itemName=starkware.cairo1) is installed.
+3. Set breakpoints in your cairo files.
+4. Start a debugging session:
+   - Use the extension's **▶ Debug Test** code lens (displayed above each test function) to launch the debugger -
```
ddoktorski, line 43: Does it make sense to include a screenshot here?

crates/forge-runner/src/running/setup.rs
```diff
@@ -52,6 +54,10 @@ pub fn initialize_execution_context<'a>(
         disable_trace_padding,
     )?;
 
+    if let Some(debugger) = debugger {
+        runner.set_vm_hooks(Box::new(debugger));
+    }
```
ddoktorski: Technically it makes sense to set it here, but given that `initialize_execution_context` is basically a copied function from the blockifier I'd prefer to move this `if` to `run_test_case` body in order to minimize changes in blockifier functions on our end as in long term it makes code maintenance more difficult

crates/forge/src/combine_configs.rs
ddoktorski: Why do you think so? Some of them seem to be quite useful

crates/forge/src/lib.rs
ddoktorski, line 252: It'd be good to somehow also validate that passed test is not a fuzz test

docs/src/snforge-advanced-features/debugging.md
piotmag769: Good idea!

crates/forge/src/combine_configs.rs
piotmag769: I can rewrite them, but tbh the logic of this function is so simple that I don't think it needs to be tested

crates/forge/src/lib.rs
piotmag769: Done

crates/forge/src/combine_configs.rs
ddoktorski: Still good sanity check

piotmag769: During restoring of the tests with clanker it found that `ForgeConfigFromScarb.tracked_resource` is ignored, lol - #4281

docs/src/snforge-advanced-features/debugging.md
piotmag769: FYI: this section may change and become just a link debugger docs (separate website) per @Arcticae suggestion. Added the screenshot for now

ddoktorski (approved):

docs/src/snforge-advanced-features/debugging.md
piotmag769: Changed to 2.18.0

franciszekjob (approved):

PR #4167, opened 2026-02-20
Implement simple inlining optimizer

maciektr:


crates/forge/src/lib.rs
```diff
@@ -240,9 +246,29 @@ impl TestArgs {
             self.tracked_resource = ForgeTrackedResource::SierraGas;
         }
     }
+
+    pub fn validate_single_exact_test_case(&self) -> Result<()> {
```
piotmag769: 1. After #4168 it would be enough to require only `--exact`.
2. Just use the clap macro like in #4168

crates/forge/src/lib.rs
```diff
@@ -240,9 +246,29 @@ impl TestArgs {
...
+            test_filter.is_some(),
+            "optimize-inlining requires exactly one test case name"
+        );
+
+        Ok(())
+    }
 }
 
-#[derive(Parser, Debug)]
+#[derive(Parser, Debug, Clone)]
```
piotmag769: Why?

crates/forge/src/optimize_inlining/args.rs
```diff
@@ -0,0 +1,74 @@
...
+    #[arg(long, default_value = "4089446")]
+    pub max_contract_size: u64,
+
+    /// Maximum allowed contract felts count
+    #[arg(long, default_value = "81920")]
+    pub max_contract_felts: u64,
+
+    /// Dry run - don't modify Scarb.toml at the end
+    #[arg(long, default_value = "false")]
+    pub dry_run: bool,
```
piotmag769: It is quite unexpected that it modifies your Scarb.toml - I wouldn't expect that at all and it is not documented in the subcommand docs.

crates/forge/src/optimize_inlining/args.rs
```diff
@@ -0,0 +1,74 @@
...
+
+    /// Test arguments (same as `snforge test`)
```
piotmag769:
```suggestion
    /// Test arguments (same as for `snforge test`)
```

crates/forge/src/optimize_inlining/args.rs
```diff
@@ -0,0 +1,74 @@
...
+            self.max_threshold
+        );
+        ensure!(self.step > 0, "step must be greater than 0");
+        self.test_args.validate_single_exact_test_case()?;
+        Ok(())
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
piotmag769, line 68: I think they just clutter the code and don't bring anything useful

crates/forge/src/optimize_inlining/args.rs
```diff
@@ -0,0 +1,74 @@
...
+
+impl OptimizeInliningArgs {
+    pub fn validate(&self) -> Result<()> {
+        ensure!(
+            self.min_threshold <= self.max_threshold,
+            "min-threshold ({}) must be <= max-threshold ({})",
+            self.min_threshold,
+            self.max_threshold
+        );
+        ensure!(self.step > 0, "step must be greater than 0");
```
piotmag769: Can't you use `NonZeroU32` as an arg?

crates/forge/tests/e2e/optimize_inlining.rs
```diff
@@ -0,0 +1,118 @@
...
+        Testing min threshold 0...
+        [..]
+
+        Testing max threshold 10...
+        [..]
+
+        Optimization Results:
+        [..]Threshold[..]
+        Optimal threshold: [..]
+        Updated Scarb.toml with inlining-strategy = [..]
```
piotmag769: We should assert the content, no?

crates/forge/src/optimize_inlining/args.rs
```diff
@@ -0,0 +1,74 @@
...
+use clap::Parser;
+
+#[derive(Parser, Debug)]
+pub struct OptimizeInliningArgs {
+    /// Minimum inlining-strategy value to test
+    #[arg(long, default_value = "0")]
+    pub min_threshold: u32,
+
+    /// Maximum inlining-strategy value to test
+    #[arg(long, default_value = "100")]
```
Arcticae: What's this value meaning? Are there any common configurations we can suggest here?

crates/forge/src/optimize_inlining/args.rs
Arcticae: they all have to be greater than zero

Arcticae: (if they are unsigned, is that validation necessary?)

crates/forge/src/optimize_inlining/args.rs
```diff
@@ -0,0 +1,74 @@
...
+    #[arg(long, default_value = "81920")]
+    pub max_contract_felts: u64,
+
+    /// Dry run - don't modify Scarb.toml at the end
+    #[arg(long, default_value = "false")]
+    pub dry_run: bool,
+
+    /// Test arguments (same as `snforge test`)
+    #[command(flatten)]
+    pub test_args: TestArgs,
```
Arcticae, line 48: will it work with i.e. profiles?

crates/forge/src/optimize_inlining/contract_size.rs
```diff
@@ -0,0 +1,113 @@
...
+pub struct StarknetContractArtifactPaths {
+    pub sierra: Option<Utf8PathBuf>,
+    pub casm: Option<Utf8PathBuf>,
+}
+
+fn parse_starknet_artifacts(path: &Utf8Path) -> Result<StarknetArtifacts> {
+    let content = fs::read_to_string(path).with_context(|| format!("Failed to read {path}"))?;
+    serde_json::from_str(&content)
+        .with_context(|| format!("Failed to parse starknet artifacts from {path}"))
+}
```
Arcticae: Isn't this duplicated logic?

Arcticae: (we're loading those in snforge)

crates/forge/src/optimize_inlining/contract_size.rs
```diff
@@ -0,0 +1,113 @@
...
+#[derive(Debug)]
+pub struct ContractSizeInfo {
+    pub contract_id: String,
+    pub artifact_type: ContractArtifactType,
+    pub size: u64,
+    pub felts_count: u64,
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[allow(dead_code)]
```
Arcticae: ?

crates/forge/src/optimize_inlining/contract_size.rs
```diff
@@ -0,0 +1,113 @@
...
+) -> Result<(bool, Vec<ContractSizeInfo>)> {
+    let mut sizes = Vec::new();
+    let mut all_valid = true;
+
+    let get_contract_size = |contract_path: &Utf8PathBuf| -> Result<u64> {
+        let size = fs::metadata(contract_path)
+            .with_context(|| format!("Failed to read {contract_path}"))?
+            .len();
+        Ok(size)
+    };
```
Arcticae: this can be a regular function

crates/forge/src/optimize_inlining/contract_size.rs
```diff
@@ -0,0 +1,113 @@
...
+
+    for starknet_artifacts_path in starknet_artifacts_paths {
+        let artifacts = parse_starknet_artifacts(starknet_artifacts_path)?;
+        let artifacts_dir = starknet_artifacts_path
+            .parent()
+            .expect("Starknet artifacts path must have a parent");
```
Arcticae, line 43: This expectation would be met already by this point, wouldnt it? (we parse the artifacts above)

crates/forge/src/optimize_inlining/contract_size.rs
```diff
@@ -0,0 +1,113 @@
...
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[allow(dead_code)]
+pub enum ContractArtifactType {
+    Sierra,
+    Casm,
+}
+
+pub fn check_contract_sizes(
```
Arcticae:
```suggestion
pub fn check_and_validate_contract_sizes(
```

crates/forge/src/optimize_inlining/runner.rs
```diff
@@ -0,0 +1,325 @@
...
+    let artifacts_dir = target_dir_for_workspace(scarb_metadata).join(profile);
+
+    let starknet_artifacts_paths =
+        find_test_target_starknet_artifacts(&artifacts_dir, scarb_metadata)?;
+    if starknet_artifacts_paths.is_empty() {
+        return Err(anyhow::anyhow!(
+            "No starknet_artifacts.json found. Only projects with contracts can be optimized."
+        ));
+    }
+    let (sizes_valid, sizes) = check_contract_sizes(
```
Arcticae: The check function should probably return a typed error instead with an indication which limit was exceeded, and short circuit the validation. 

Also: When this exceeding of this validation will occur? What the user can do about it? Perhaps we can suggest a solution here if we know what to do.

crates/scarb-api/src/manifest.rs
```diff
@@ -0,0 +1,75 @@
...
+    };
+    let Item::Table(target_table) = target_item else {
+        return false;
+    };
+    let Some(starknet_contract_item) = target_table.get_mut("starknet-contract") else {
+        return false;
+    };
+    let Item::ArrayOfTables(starknet_contract_tables) = starknet_contract_item else {
+        return false;
+    };
```
Arcticae, line 61: `let chain` can be used here

crates/forge/src/optimize_inlining/mod.rs
```diff
@@ -0,0 +1,198 @@
...
+    let original_metadata = metadata_with_opts(MetadataOpts {
+        profile: profile.clone(),
+        ..MetadataOpts::default()
+    })?;
+
+    let workspace_root = &original_metadata.workspace.root;
+    let target_dir = &original_metadata
+        .target_dir
+        .unwrap_or(workspace_root.join("target"));
+    ui.println(&"Copying project to temporary directory...".to_string());
```
Arcticae: Is this something the user needs to know?

crates/forge/src/optimize_inlining/mod.rs
```diff
@@ -0,0 +1,198 @@
...
+    let manifest_editor = ManifestEditor::new(&original_metadata.runtime_manifest);
+
+    let mut optimizer = Optimizer::new(args, &scarb_metadata);
+    let optimization_result = optimizer.optimize(args, cores, ui);
+
+    ui.print_blank_line();
+    ui.println(&"Optimization Results:".to_string());
+    optimizer.print_results_table(ui);
+
+    let graph_path = target_dir.join("optimization_results.png");
```
Arcticae: we should probably parametrize the name so it's easier to decode - or add a metadata file alongside for later use

crates/forge/src/optimize_inlining/optimizer.rs
```diff
@@ -0,0 +1,389 @@
...
+use anyhow::{Result, anyhow};
+use camino::Utf8Path;
+use foundry_ui::UI;
+use plotters::prelude::*;
+use plotters::style::{FontStyle, register_font};
+use scarb_api::metadata::Metadata;
+use std::sync::Arc;
+
+const ROBOTO_REGULAR: &[u8] = include_bytes!("../../assets/fonts/Roboto-Regular.ttf");
+const ROBOTO_FAMILY: &str = "roboto";
```
Arcticae, line 18: do we have to ship that?

crates/forge/src/optimize_inlining/optimizer.rs
```diff
@@ -0,0 +1,389 @@
...
+                "  ✓ gas: {:.0}, contract bytecode L2 gas: {}",
+                max_result.total_gas.l2(),
+                max_result.contract_code_l2_gas
+            ));
+        } else {
+            ui.println(&format!(
+                "  ✗ {}",
+                max_result.error.as_deref().unwrap_or("failed")
+            ));
+        }
```
Arcticae: its the same thing as above with `min`

crates/forge/src/optimize_inlining/optimizer.rs
```diff
@@ -0,0 +1,389 @@
...
+            iteration += 1;
+        }
+
+        self.find_best_result()
+    }
+
+    fn find_best_result(&self) -> Result<OptimalResult> {
+        self.results
+            .iter()
+            .filter(|r| r.tests_passed && r.error.is_none())
```
Arcticae: this can be extracted, it's used a lot

crates/forge/src/optimize_inlining/optimizer.rs
```diff
@@ -0,0 +1,389 @@
...
+            .min_by(|a, b| {
+                a.total_gas
+                    .l2()
+                    .total_cmp(&b.total_gas.l2())
+                    .then(a.contract_code_l2_gas.cmp(&b.contract_code_l2_gas))
```
Arcticae, line 63: This defines a priority of gas optimization over total contract size - why are we doing that? At the very least we should allow the user to choose which one would be best

crates/forge/src/optimize_inlining/optimizer.rs
```diff
@@ -0,0 +1,389 @@
...
+                    .then(a.threshold.cmp(&b.threshold))
+            })
+            .map(|r| OptimalResult {
+                threshold: r.threshold,
+                total_gas: r.total_gas.clone(),
+                max_contract_size: r.max_contract_size,
+                contract_code_l2_gas: r.contract_code_l2_gas,
+            })
+            .ok_or_else(|| anyhow!("No valid optimization results found"))
```
Arcticae: We should probably validate this and fail earlier with a more specific error message

crates/forge/src/optimize_inlining/optimizer.rs
```diff
@@ -0,0 +1,389 @@
...
+        Ok(())
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn bundled_font_renders_text_into_bitmap() {
```
Arcticae: i would prefer a snapshot with a genrated example image, and test if it generates the same one. Tests more than just the font

crates/forge/src/optimize_inlining/runner.rs
```diff
@@ -0,0 +1,325 @@
...
+}
+
+impl TotalGas {
+    pub fn l2(&self) -> f64 {
+        self.l2
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct OptimizationResult {
```
Arcticae:
```suggestion
pub struct OptimizationIterationResult {
```

crates/forge/src/optimize_inlining/runner.rs
```diff
@@ -0,0 +1,325 @@
...
+
+    let starknet_artifacts_paths =
+        find_test_target_starknet_artifacts(&artifacts_dir, scarb_metadata)?;
+    if starknet_artifacts_paths.is_empty() {
+        return Err(anyhow::anyhow!(
+            "No starknet_artifacts.json found. Only projects with contracts can be optimized."
+        ));
+    }
+    let (sizes_valid, sizes) = check_contract_sizes(
+        &starknet_artifacts_paths,
```
Arcticae, line 127: What if we build 2 different contracts in tests and one interact with the other? i.e. erc20 + our contract and they are both deployed in tests

crates/forge/src/optimize_inlining/runner.rs
```diff
@@ -0,0 +1,325 @@
...
+            Some(
+                test_result
+                    .error
+                    .unwrap_or_else(|| "Some tests failed".to_string()),
+            )
+        },
+    })
+}
+
+fn contract_code_l2_gas(sizes: &[ContractSizeInfo]) -> Result<u64> {
```
Arcticae, line 195: Can we document which formula it follows? I assume it does, right?

crates/forge/src/optimize_inlining/runner.rs
Arcticae: I guess this one is not easily solvable. We probably should discuss this culprit offline

crates/forge/src/optimize_inlining/paths.rs
Arcticae, line 1: I understand it's necessary because we can't set/override this value explicitly via scarbs API? What would it take to do it that way?

Arcticae (changes requested):

crates/forge/src/optimize_inlining/paths.rs
maciektr: So the problem is, that we are copying the package to a temp directory, thus the relative paths may break. 
All relative paths from the package manifest are calculated as relative to the manifest location. https://github.com/software-mansion/scarb/blob/6808e76c9b254b3743f398bea921004eae888f01/scarb/src/core/manifest/toml_manifest.rs#L1598-L1598 
While we could introduce some magic to overwrite this behaviour inside Scarb, I don't feel like we should - rewriting them here is simpler and backward compatible.

crates/forge/src/optimize_inlining/optimizer.rs
maciektr: If we do not ship the fonts statically with the binary, those would be pulled from the os, which may break depending on what the user have installed in their system 😓

crates/forge/src/optimize_inlining/mod.rs
maciektr: Yes and no - we could skip that if you think this is too informative.

crates/scarb-api/src/manifest.rs
maciektr: Only if we rewrite to an if statement that wraps the later logic, so I don't think it's worth doing

crates/forge/src/optimize_inlining/contract_size.rs
maciektr: Done!

crates/forge/src/optimize_inlining/args.rs
maciektr: Yes, profiles are handled correctly. https://github.com/maciektr/starknet-foundry/blob/8798db74aaf97e54370bad3e610745ceed714847/crates/forge/src/optimize_inlining/mod.rs#L27

crates/forge/src/optimize_inlining/args.rs
maciektr: As per https://docs.swmansion.com/scarb/docs/reference/manifest.html#inlining-strategy , we do not provide any definition of what the threshold is, and the underlying impl can change with each compiler release, so not really. 
I've changed it to a higher value though!

crates/forge/src/optimize_inlining/args.rs
maciektr: Unsigned can accept `0`, and we would spin in place.

crates/forge/src/optimize_inlining/contract_size.rs
maciektr: Yes, we can unify them. 
Done!

crates/forge/src/optimize_inlining/contract_size.rs
maciektr: Removed!

crates/forge/src/optimize_inlining/contract_size.rs
maciektr: It can not be one :p 
Changed!

crates/forge/src/optimize_inlining/contract_size.rs
maciektr: Yes, that's why it's unwrapped instead of being handled

crates/forge/src/optimize_inlining/mod.rs
maciektr: Done!

crates/forge/src/optimize_inlining/optimizer.rs
maciektr: Changed!

crates/forge/src/optimize_inlining/optimizer.rs
maciektr: Done!

crates/forge/src/optimize_inlining/optimizer.rs
maciektr: Right, changed!

crates/forge/src/optimize_inlining/runner.rs
maciektr: Done.

crates/forge/src/optimize_inlining/runner.rs
maciektr: Not sure if I understand the question 🤔 
This will iterate over all contracts that are compiled for each test target in the workspace.

crates/forge/src/optimize_inlining/runner.rs
maciektr: > When this exceeding of this validation will occur? 

So, as you know, the inlining threshold can widely affect the sizes of resulting contracts. From the compilers POV, there is no limit on how large the generated contract code can be. The limit is only practical from starknet perspective, as the network does not want you to declare too large contracts. 

The failed validation only means, that the resulting contract is too large for you to declare on starknet. For example, we are doing too much inlining and a lot of contract code is duplicated.

> The check function should probably return a typed error instead

From the optimizers perspective, it's not a failure per se. We should advise the user to not use this particular threshold value (as the contracts are too large, also the limit can be configured as optimization parameter via CLI), but it's still a valid data point we want to show. 

> Perhaps we can suggest a solution here if we know what to do.

The solution should be just to use a different threshold from the ones we tested.

crates/forge/src/optimize_inlining/optimizer.rs
maciektr: Done!

crates/forge/src/optimize_inlining/runner.rs
maciektr: Done!

crates/forge/src/optimize_inlining/optimizer.rs
maciektr: I am not sure what you mean by earlier - we need to iterate over all results for this anyway! So that suggests the optimizer must finish before that 🤔

crates/forge/src/optimize_inlining/args.rs
ddoktorski: Why not to use `NonZeroU32` as the step type?

crates/scarb-api/src/manifest.rs
ddoktorski, line 1: Maybe we can move it to the `scarb_api` crate? It can potentially be extended and reused in other places

crates/forge/src/optimize_inlining/optimizer.rs
```diff
@@ -0,0 +1,361 @@
...
+            })
+            .map(|r| OptimalResult {
+                threshold: r.threshold,
+                total_gas: r.total_gas.clone(),
+                contract_code_l2_gas: r.contract_code_l2_gas,
+            })
+            .expect("valid_results must return at least one result"))
+    }
+
+    pub fn print_results_table(&self, ui: &UI) {
```
ddoktorski, line 92: Wouldn't it be easier to use external dependency for table formatting? We already use `comfy-table`

crates/forge/src/optimize_inlining/runner.rs
```diff
@@ -0,0 +1,334 @@
...
+    manifest_editor.set_inlining_strategy(threshold, profile)?;
+
+    let build_result = ScarbCommand::new_with_stdio()
+        .manifest_path(&scarb_metadata.runtime_manifest)
+        .arg("--profile")
+        .arg(profile)
+        .arg("build")
+        .arg("-w")
+        .arg("--test")
+        .run();
```
ddoktorski, line 97: It produces a lot of noise in the output from `Compiling...`, `Finished...`, I'd consider running it with verbosity=quiet

crates/forge/tests/data/optimize_inlining/src/lib.cairo
ddoktorski, line 1: I'm pretty sure we already have this package somewhere in the data directory. If that's the case please use the existing one

crates/forge/src/lib.rs
```diff
@@ -245,9 +251,29 @@ impl TestArgs {
             self.tracked_resource = ForgeTrackedResource::SierraGas;
         }
     }
+
+    pub fn validate_single_exact_test_case(&self) -> Result<()> {
+        ensure!(
+            self.exact,
+            "optimize-inlining requires --exact and one exact test case name"
```
ddoktorski:
```suggestion
            "optimize-inlining requires using the `--exact` flag"
```
exact allows to pass only one test anyway, so it seems redundant, but tbh I'm not sure if this is a bug or not 🤔

crates/forge/src/lib.rs
```diff
@@ -245,9 +251,29 @@ impl TestArgs {
...
+        let test_filter = self
+            .test_filter
+            .as_deref()
+            .map(str::trim)
+            .filter(|name| !name.is_empty());
+
+        ensure!(
+            test_filter.is_some(),
+            "optimize-inlining requires exactly one test case name"
+        );
```
ddoktorski: Not needed after #4168

crates/forge/src/run_tests/workspace.rs
```diff
@@ -30,6 +30,21 @@ use std::sync::Arc;
...
+        ExitStatus::Failure
+    } else {
+        ExitStatus::Success
+    })
+}
+
+#[derive(Debug)]
+pub struct WorkspaceExecutionSummary {
+    pub all_tests: Vec<TestTargetSummary>,
+}
```
ddoktorski: nit: Let's move this struct above `run_for_workspace` function

crates/forge/assets/fonts/Roboto-Regular.ttf
ddoktorski: Could you explain why this is necessary and what can happen if we don't include it?

crates/scarb-api/src/manifest.rs
```diff
@@ -0,0 +1,75 @@
...
+            target["casm"] = value(true);
+            changed = true;
+        }
+        if target.get("sierra").and_then(Item::as_bool) != Some(true) {
+            target["sierra"] = value(true);
+            changed = true;
+        }
+    }
+
+    changed
```
ddoktorski, line 74: nit:
```suggestion
    let keys = ["casm", "sierra"];
    let mut changed = false;
    
    for target in starknet_contract_tables.iter_mut() {
        for key in keys {
            if target.get(key).and_then(Item::as_bool) != Some(true) {
                target[key] = value(true);
                changed = true;
            }
        }
    }

    changed
```

crates/forge/src/optimize_inlining/args.rs
```diff
@@ -0,0 +1,80 @@
...
+    #[arg(long, default_value = "25")]
+    pub step: u32,
+
+    /// Maximum allowed contract size in bytes
+    #[arg(long, default_value = "4089446")]
+    pub max_contract_size: u64,
+
+    /// Maximum allowed contract felts count
+    #[arg(long, default_value = "81920")]
+    pub max_contract_felts: u64,
```
ddoktorski: `max_contract_felts` seems a bit ambiguous name, is it actually maximum sierra program length?

crates/forge/src/optimize_inlining/optimizer.rs
ddoktorski: What exactly may break?

crates/forge/tests/e2e/snapshots/main__e2e__optimize_inlining__optimize_inlining_dry_run.snap.png
ddoktorski, line 1: Chart legend seems a bit broken

ddoktorski (changes requested):

crates/forge/src/optimize_inlining/paths.rs
Arcticae: Ok i can live with that i guess

crates/forge/src/lib.rs
piotmag769: #4168

crates/forge/src/optimize_inlining/args.rs
```diff
@@ -0,0 +1,80 @@
...
+
+    /// Maximum inlining-strategy value to test
+    #[arg(long, default_value = "250")]
+    pub max_threshold: u32,
+
+    /// Step size for threshold search
+    #[arg(long, default_value = "25")]
+    pub step: u32,
+
+    /// Maximum allowed contract size in bytes
```
piotmag769: I would add this concerns a contract file size. It could be understood as contract bytecode size

crates/forge/src/optimize_inlining/args.rs
piotmag769: Actually, is this comment visible in the command line, considering we used flatten?

crates/scarb-api/src/artifacts/deserialized.rs
```diff
@@ -20,6 +20,7 @@ pub struct StarknetContract {
 #[derive(Deserialize, Debug, PartialEq, Clone)]
 pub struct StarknetContractArtifactPaths {
     pub sierra: Utf8PathBuf,
+    pub casm: Option<Utf8PathBuf>,
```
piotmag769, line 24: What does this field do?

crates/forge/src/optimize_inlining/contract_size.rs
```diff
@@ -0,0 +1,84 @@
...
+    pub artifact_type: ContractArtifactType,
+    pub size: u64,
+    pub felts_count: u64,
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum ContractArtifactType {
+    Sierra,
+    Casm,
+}
```
piotmag769, line 26: Please explain the purpose of this enum

crates/forge/src/optimize_inlining/args.rs
maciektr: outdated

crates/scarb-api/src/artifacts/deserialized.rs
maciektr: discussed offline

crates/forge/src/optimize_inlining/args.rs
maciektr: We can do that, the error message is a bit strange though 
`error: invalid value '0' for '--step <STEP>': number would be zero for non-zero type`

crates/forge/src/lib.rs
maciektr: execute_workspace took ownership - changed to reference! 
https://github.com/maciektr/starknet-foundry/blob/51b798aa2834968f0ba1c8cc0fc5e3d771ec7ce7/crates/forge/src/optimize_inlining/runner.rs#L248

crates/scarb-api/src/manifest.rs
maciektr: Done!

crates/forge/src/optimize_inlining/optimizer.rs
maciektr: Done!

crates/forge/src/optimize_inlining/runner.rs
maciektr: I agree it's noisy, but silencing the compilation output creates a risk of failing silently from users perspective, e.g. if the compilation fails because of some errors in the user code, or if it hangs for whatever reason. Showing the output as the compilation progresses is safer in these cases.

crates/forge/src/run_tests/workspace.rs
maciektr: Done!

crates/scarb-api/src/manifest.rs
maciektr: Done!

crates/forge/src/optimize_inlining/args.rs
maciektr: Done. 
No, it's not - by design.

crates/forge/src/optimize_inlining/args.rs
maciektr: Done.

crates/forge/src/lib.rs
maciektr: Done!

crates/forge/src/lib.rs
maciektr: Removed.

crates/forge/src/lib.rs
maciektr: 1. Done.
2. To do what? 🤔

crates/forge/src/optimize_inlining/contract_size.rs
maciektr: Discussed offline. We gather contract sizes for both casm and sierra compiled contracts. While this can be skipped, as both are priced with the same weight, I kept it for debugging purposes.

crates/forge/src/lib.rs
piotmag769: 2. To force `--exact` flag

maciektr: I don't think I can - I don't want to set arg to arg relation, but always require some arg to be provided for some parser.

crates/forge/tests/e2e/snapshots/main__e2e__optimize_inlining__optimize_inlining_dry_run.snap.png
maciektr: Fixed!

crates/forge/tests/data/optimize_inlining/src/lib.cairo
maciektr: Done.

crates/forge/tests/e2e/optimize_inlining.rs
maciektr: Done.

crates/forge/src/optimize_inlining/args.rs
maciektr: Done!

crates/forge/src/optimize_inlining/args.rs
maciektr: This is currently applied to both sierra and casm contracts, as `class.bytecode.len()` or `class.sierra_program.len()`. We could have two flags, but I am not sure if we should care? The default is same for both 🤔

crates/forge/src/optimize_inlining/args.rs
Arcticae: i don't think it's worth digging into to make it better

crates/scarb-api/src/manifest.rs
Arcticae: readability is better in those, i prefer this to the multiple `false` returns, but do as you wish

Arcticae: if nesting is problematic you can make additional function for logic below

Arcticae: this is a nitpick feel free to resolve it

crates/forge/src/optimize_inlining/mod.rs
Arcticae: i would not show that

crates/forge/src/optimize_inlining/runner.rs
```diff
@@ -0,0 +1,427 @@
...
+
+    if !sizes_valid {
+        return Ok(OptimizationIterationResult {
+            threshold,
+            total_gas: TotalGas::default(),
+            max_contract_size,
+            contract_code_l2_gas,
+            tests_passed: false,
+            error: Some(format!(
+                "Contract size {} exceeds limit {} or felts {} exceeds limit {}",
```
Arcticae: Please tell the user what he needs to (try) to do. 
He needs to decrease the threshold

Arcticae (approved):

crates/forge/src/optimize_inlining/runner.rs
ddoktorski: Ok, leaving this up to you

crates/forge/src/optimize_inlining/optimizer.rs
maciektr: So, to render fonts on the graph, we need to get the fonts file somehow. There are two approaches, as described here https://docs.rs/plotters/0.3.7/plotters/#:~:text=Font%20manipulation%20features
One would be to rely on the runtime OS to find installed fonts, via some interface like libfontconfig.so. The second one would be to bundle the fonts file somehow, and distribute it with the binary itself. 
Note, that the problem goes beyond static / dynamic linking, as even if we are linking all code required to read and render the fonts (code libraries like libfontconfig.so etc), we still need to obtain the font files themselves during the runtime - and these are data, no code, so we are not linking them anyhow. 
In case we link dynamically against libfontconfig.so, we would need to make sure it's installed on both the ci/cd runners building our releases and on end user os (that's why plotters state installing it as a prerequisite in docs https://docs.rs/plotters/0.3.7/plotters/#ubuntu-linux). Otherwise, we would either get compilation / linking error in ci/cd and "missing library" error in runtime. 
Even if we have all the code linked statically, it's possible that we fail to find font assets on runtime os (e.g. docker, github actions etc) and we would still fail with some runtime error when trying to render the graph (note that in general, when linking font renderer statically, we our the ones responsible for providing the font file path, by whatever means of discovery we choose). 
For this reason I suggest including the file into the binary, as it is just very straightforward and simple in comparison to going the os discovery way - let me know what you think!

crates/forge/src/optimize_inlining/runner.rs
maciektr: I am voting for keeping the output then 😅

crates/forge/src/optimize_inlining/mod.rs
maciektr: Removed!

crates/forge/src/optimize_inlining/runner.rs
maciektr: Done!

crates/forge/Cargo.toml
```diff
@@ -59,6 +60,11 @@ tracing-subscriber.workspace = true
 tracing.workspace = true
 tracing-chrome.workspace = true
 rayon.workspace = true
+tempfile.workspace = true
+fs_extra.workspace = true
+comfy-table.workspace = true
+plotters = { version = "0.3", default-features = false, features = ["all_series", "all_elements", "svg_backend", "bitmap_backend", "bitmap_encoder", "full_palette", "colormaps", "ab_glyph"] }
```
piotmag769: Imo put into workspace toml and inherit

crates/forge/src/optimize_inlining/args.rs
piotmag769: bump

crates/forge/src/optimize_inlining/mod.rs
```diff
@@ -0,0 +1,204 @@
...
+mod tests {
+    use super::paths::{normalize_utf8_path_lexically, rewrite_manifest_paths_to_absolute};
+    use anyhow::Result;
+    use camino::Utf8PathBuf;
+    use indoc::indoc;
+    use std::fs;
+    use toml_edit::DocumentMut;
+
+    #[test]
+    fn rewrites_relative_manifest_paths_to_absolute_with_original_manifest_as_base() -> Result<()> {
```
piotmag769: This should not be in this file but in `path.rs`

crates/forge/src/optimize_inlining/optimizer.rs
```diff
@@ -0,0 +1,335 @@
...
+                self.results.push(result);
+            }
+            current += self.step.get();
+            iteration += 1;
+        }
+
+        self.find_best_result_by_gas()
+    }
+
+    pub fn find_best_result_by_gas(&self) -> Result<OptimalResult> {
```
piotmag769, line 55: Don't need this `pub`

crates/forge/src/optimize_inlining/optimizer.rs
```diff
@@ -0,0 +1,335 @@
...
+            })
+            .map(|r| OptimalResult {
+                threshold: r.threshold,
+                total_gas: r.total_gas.clone(),
+                contract_code_l2_gas: r.contract_code_l2_gas,
+            })
+            .expect("valid_results must return at least one result"))
+    }
+
+    pub fn find_best_result_by_contract_size(&self) -> Result<OptimalResult> {
```
piotmag769, line 74: vide supra

crates/forge/src/optimize_inlining/optimizer.rs
piotmag769, line 1: Please put `pub` above private in the impl

crates/forge/src/optimize_inlining/optimizer.rs
```diff
@@ -0,0 +1,335 @@
...
+
+    fn valid_results(&self) -> Result<Vec<&OptimizationIterationResult>> {
+        let results: Vec<_> = self
+            .results
+            .iter()
+            .filter(|r| r.tests_passed && r.error.is_none())
+            .collect();
+        if results.is_empty() {
+            return Err(anyhow!("No valid optimization results found"));
+        }
```
piotmag769: Should be the first thing in this func

crates/forge/src/optimize_inlining/runner.rs
piotmag769, line 1: Should be a submodule of `optimizer`

crates/forge/src/optimize_inlining/runner.rs
```diff
@@ -0,0 +1,430 @@
...
+        .with_context(|| format!("Failed to parse {artifact_path}"))?;
+
+    Ok(json
+        .get("contracts")
+        .and_then(serde_json::Value::as_array)
+        .is_some_and(|contracts| !contracts.is_empty()))
+}
+
+#[cfg(test)]
+mod tests {
```
piotmag769: I don't think we need to test such a simple function xD

crates/forge/src/optimize_inlining/contract_size.rs
```diff
@@ -0,0 +1,120 @@
...
+
+fn get_contract_size(contract_path: &Utf8PathBuf) -> Result<u64> {
```
piotmag769: Should be lower

crates/forge/src/optimize_inlining/contract_size.rs
piotmag769, line 1: Should be a submodule of `runner`

crates/forge/src/optimize_inlining/mod.rs
```diff
@@ -0,0 +1,204 @@
...
+            Ok(ExitStatus::Success)
+        }
+        Err(e) => {
+            bail!(format!("Optimization failed: {e}"));
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
piotmag769: This belongs to `paths.rs`

crates/forge/src/optimize_inlining/optimizer.rs
maciektr: I do 
image

crates/forge/src/optimize_inlining/optimizer.rs
maciektr: image

crates/forge/src/optimize_inlining/optimizer.rs
maciektr: wdym? I need to filter the results first.

crates/forge/src/optimize_inlining/args.rs
maciektr: I don't see nothing wrong, with asserting we validate inputs

crates/forge/src/optimize_inlining/runner.rs
```diff
@@ -0,0 +1,396 @@
...
+///
+/// See <https://docs.starknet.io/learn/protocol/fees#l2-data> for details.
+fn contract_code_l2_gas(sizes: &[ContractSizeInfo]) -> Result<u64> {
+    let mut felts_by_contract: HashMap<&str, u64> = HashMap::new();
+
+    for size in sizes {
+        if matches!(
+            size.artifact_type,
+            ContractArtifactType::Sierra | ContractArtifactType::Casm
+        ) {
```
ddoktorski, line 202: Isn't it always true? 🤔

crates/forge/src/optimize_inlining/mod.rs
```diff
@@ -0,0 +1,110 @@
...
+        args.min_threshold, args.max_threshold, args.step
+    ));
+    create_output_dir::create_output_dir(target_dir.as_std_path())?;
+    if let Err(e) = optimizer.save_results_graph(&graph_path, ui) {
+        ui.eprintln(&format!("Warning: Failed to save graph: {e}"));
+    }
+
+    match optimization_result {
+        Ok(_) => {
+            let profile_name = profile.unwrap_or_else(|| "dev".to_string());
```
ddoktorski: Shouldn't we use `release` as default instead of `dev`?

crates/forge/src/optimize_inlining/args.rs
```diff
@@ -0,0 +1,100 @@
...
+    /// Maximum inlining-strategy value to test
+    #[arg(long, default_value = "250")]
+    pub max_threshold: u32,
+
+    /// Step size for threshold search
+    #[arg(long, default_value = "25")]
+    pub step: NonZeroU32,
+
+    /// Maximum allowed contract file size in bytes
+    #[arg(long, default_value = "4089446")]
```
ddoktorski, line 25: For all default values could you add a source? 🙏 either here or in the docs

crates/forge/src/optimize_inlining/optimizer.rs
ddoktorski: Thanks for the exhaustive response! Not ideal to distribute it with the binary, but taking everything into consideration I guess we can live with it

crates/forge/src/optimize_inlining/optimizer.rs
```diff
@@ -0,0 +1,335 @@
...
+            })
+            .map(|r| OptimalResult {
+                threshold: r.threshold,
+                total_gas: r.total_gas.clone(),
+                contract_code_l2_gas: r.contract_code_l2_gas,
+            })
+            .expect("valid_results must return at least one result"))
+    }
+
+    pub fn print_results_table(&self, ui: &UI) {
```
ddoktorski, line 92: Is it supposed to look like this?

Image

crates/forge/tests/e2e/optimize_inlining.rs
ddoktorski, line 1: Is it possible that this graph snapshot will require updating for newer Scarb versions in the future?

crates/forge/src/lib.rs
```diff
@@ -306,6 +312,12 @@ pub enum ExitStatus {
 #[tracing::instrument(skip_all, level = "debug")]
 pub fn main_execution(ui: Arc<UI>) -> Result<ExitStatus> {
     let cli = Cli::parse();
+    let cores = if let Ok(available_cores) = available_parallelism() {
+        available_cores.get()
+    } else {
+        ui.eprintln(&"Failed to get the number of available cores, defaulting to 1");
+        1
+    };
```
ddoktorski: Let's use `resolve_thread_count` function

crates/forge/src/run_tests/workspace.rs
```diff
@@ -27,9 +27,28 @@ use shared::consts::SNFORGE_TEST_FILTER;
...
+        ExitStatus::Failure
+    } else {
+        ExitStatus::Success
+    })
+}
+
+pub async fn execute_workspace(
+    args: &TestArgs,
+    ui: Arc<UI>,
+    skip_build: bool,
```
ddoktorski: nit: Maybe we can move build logic to the `run_for_workspace` function and omit passing `skip_build` flag here?

crates/forge/src/optimize_inlining/runner.rs
maciektr: Yes! As in https://github.com/foundry-rs/starknet-foundry/pull/4167#discussion_r2902490701, I only left the distinction for debug purposes and future reference, the gas cost is the same for both casm and sierra, so If you prefer I can remove this enum completely from the code - let me know!

crates/forge/src/optimize_inlining/mod.rs
maciektr: It shouldn't really matter for now - as the default inlining is same for both, but I can change to release if you prefer!

crates/forge/src/optimize_inlining/optimizer.rs
maciektr: Resolving then 😅

crates/forge/tests/e2e/optimize_inlining.rs
maciektr: I think yes, it's possible

crates/forge/src/lib.rs
maciektr: Fixed!

crates/forge/src/run_tests/workspace.rs
maciektr: Sure, changed!

crates/forge/src/optimize_inlining/optimizer.rs
maciektr: Hmm, what terminal is this? 🤔 Looks fine on mine 😓 
image
added assert

crates/forge/src/optimize_inlining/runner.rs
maciektr: Resolving

crates/forge/src/optimize_inlining/optimizer.rs
maciektr: Resolving

crates/forge/src/optimize_inlining/runner.rs
maciektr: Ignoring

crates/forge/src/optimize_inlining/contract_size.rs
maciektr: Ignoring

crates/forge/src/optimize_inlining/args.rs
maciektr: Done.

crates/forge/src/optimize_inlining/optimizer.rs
ddoktorski: I'm using iTerm2. When I tried to run it in the built in terminal it is formatted as expected

crates/forge/src/optimize_inlining/runner.rs
ddoktorski: If you think it makes sense to keep it let's leave it, but a short comment in the code might be helpful for the future

crates/forge/src/optimize_inlining/mod.rs
ddoktorski: Yup I'd change it to release in this case

crates/forge/tests/e2e/optimize_inlining.rs
ddoktorski: All tests that assert gas values require adding snapshots for all Scarb versions that we support as described here https://github.com/foundry-rs/starknet-foundry/blob/master/CONTRIBUTING.md#snapshot-tests. Could you add them?

ddoktorski (commented):
One more comment regarding tests

crates/forge/src/optimize_inlining/mod.rs
maciektr: Done!

crates/forge/src/optimize_inlining/runner.rs
maciektr: Done!

ddoktorski (approved):
LGTM

PR #4171, opened 2026-02-21
Add `ContractRegistry`

franciszekjob:
Towards #3810

**Stack**
- #4176
- #4174
- #4216
- #4173
- #4172
- #4171 ⬅️ 

## Introduced changes

Add `ContractRegistry` used in multicall

## Checklist

- [x] Linked relevant issue
- [ ] Updated relevant documentation
- [x] Added relevant tests
- [x] Performed self-review of the code
- [ ] Added changes to `CHANGELOG.md`

Issue #3810, opened 2025-10-13
New multicall interface


cptartur:
### Current State

Current multicall interface is very cumbersome to use – it requires creating a file just to sent a multicall. 

### Objective

Design and implement and a new multicall interface with a better UX

PR #4172, opened 2026-02-21
Refactor `Deploy` and `Invoke` structs

franciszekjob:
Towards #3810

**Stack**
- #4176
- #4174
- #4216
- #4173
- #4172 ⬅️
- #4171 

## Introduced changes

Add `DeployCommonArgs` and `InvokeCommonArgs`

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/sncast/src/starknet_commands/deploy.rs
ddoktorski, line 38: Does it make sense to list these arguments here instead of keeping separate `DeployArguments` struct?

crates/sncast/src/starknet_commands/invoke.rs
```diff
@@ -29,6 +28,13 @@ pub struct Invoke {
 
     #[command(flatten)]
     pub arguments: Arguments,
```
ddoktorski, line 30: Same question here

franciszekjob: I think it's reasonable to have `Arguments` separate struct. For instance, we have [`try_into_calldata`](https://github.com/foundry-rs/starknet-foundry/blob/60c7d522925524be41d76e67b321404fcc68b8b7/crates/sncast/src/main.rs#L203) method, which we use in a couple of places.

In regards to `DeployArguments` - we could remove it as it doesn't have any additional logic, but for sake of consistency I think we can keep it.

crates/sncast/src/starknet_commands/deploy.rs
franciszekjob: Responded in https://github.com/foundry-rs/starknet-foundry/pull/4172#discussion_r2871774509

ddoktorski (approved):

DelevoXDG (approved):
lgtm

PR #4173, opened 2026-02-21
Refactor logic of `multicall run`

franciszekjob:
Towards #3810

**Stack**
- #4176
- #4174
- #4216
- #4173 ⬅️
- #4172
- #4171 

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/sncast/src/starknet_commands/multicall/contract_registry.rs
```diff
@@ -1,16 +1,90 @@
...
+    providers::{JsonRpcClient, jsonrpc::HttpTransport},
+};
 use starknet_types_core::felt::Felt;
-use std::collections::HashMap;
+use std::{
+    collections::{HashMap, hash_map::Entry},
+    sync::Arc,
+};
+
+pub(crate) struct ContractsCache {
```
ddoktorski: It seems like `pub(crate)` modifiers are redundant. We can simplify this by making the entire module visible at this level and using standard pub markers instead, but given that sncast is not used as a library anyway we can just make everything public as in the rest of the code

crates/sncast/src/starknet_commands/multicall/contract_registry.rs
```diff
@@ -1,16 +1,90 @@
...
+    address_to_class_hash: HashMap<Felt, Felt>,
+    class_hash_to_contract_class: HashMap<Felt, Arc<ContractClass>>,
```
ddoktorski: Does it need to be `Arc`?

crates/sncast/src/main.rs
```diff
@@ -228,6 +221,17 @@ impl Arguments {
     }
 }
 
+pub fn calldata_to_felts(calldata: &[String]) -> Result<Vec<Felt>> {
+    calldata
+        .iter()
+        .map(|data| {
+            Felt::from_dec_str(data)
+                .or_else(|_| Felt::from_hex(data))
+                .context("Failed to parse to felt")
```
ddoktorski: nit: Add `data` to context message to see what failed to parse

crates/sncast/src/starknet_commands/multicall/run_calls.rs
```diff
@@ -0,0 +1,173 @@
...
+                ],
+                vec![
+                    "deploy".to_string(),
+                    "--class-hash".to_string(),
+                    "0x456".to_string()
+                ]
+            ]
+        );
+    }
+}
```
ddoktorski: newline 👀

crates/sncast/src/starknet_commands/multicall/deploy.rs
```diff
@@ -0,0 +1,92 @@
...
+            .class_hash
+            .context("Using deploy with multicall requires providing `--class-hash`")?;
+        let constructor_calldata = if let Some(raw_calldata) = &constructor_arguments.calldata {
+            calldata_to_felts(raw_calldata)?
+        } else {
+            let contract_class = contracts_registry
+                .cache
+                .get_contract_class_by_class_hash(&class_hash)
+                .await?;
+            constructor_arguments.try_into_calldata(&contract_class, &constructor_selector)?
```
ddoktorski: nit: Move `let constructor_arguments = ...` from line 35 here

crates/sncast/src/starknet_commands/multicall/run.rs
```diff
@@ -80,52 +83,57 @@ pub async fn run(
 
         match call_type.unwrap().as_str() {
             Some("deploy") => {
-                let deploy_call: DeployCall = toml::from_str(toml::to_string(&call)?.as_str())
+                let item: DeployItem = toml::from_str(toml::to_string(&call)?.as_str())
```
ddoktorski: It seems that this serialization logic can be somehow improved

crates/sncast/src/starknet_commands/multicall/run_calls.rs
```diff
@@ -0,0 +1,173 @@
...
+use crate::starknet_commands::{
+    invoke::execute_calls,
+    multicall::{
+        Multicall, contracts_registry::ContractsRegistry, deploy::MulticallDeploy,
+        invoke::MulticallInvoke,
+    },
+};
+
+pub async fn run_calls(
+    tokens: &[String],
```
ddoktorski: Would it be possible to introduce better typing than relying on strings?

ddoktorski (changes requested):

crates/sncast/src/starknet_commands/multicall/run_calls.rs
franciszekjob: Removed whole file (added accidentally).

crates/sncast/src/starknet_commands/multicall/deploy.rs
franciszekjob: `constructor_arguments` is used in lien 46 as well
```rust
...
let constructor_calldata = if let Some(raw_calldata) = &constructor_arguments.calldata {
...
```

so I cannot move it as you suggested.

crates/sncast/src/starknet_commands/multicall/run.rs
franciszekjob: Good idea - added https://github.com/foundry-rs/starknet-foundry/pull/4173/changes/84195087d5a6f6ebc2643d1b2b960e223ece7e48, take a look pls.

crates/sncast/src/starknet_commands/multicall/contract_registry.rs
franciszekjob: `ContractClass` can be a heavy structure so we want to avoid potential expensive cloning.

crates/sncast/src/starknet_commands/multicall/run_calls.rs
franciszekjob: Resolving, as this file was removed.

crates/sncast/src/starknet_commands/multicall/contract_registry.rs
ddoktorski: Why can't we return a reference in `get_contract_class_by_class_hash`? I also don't think that we use this struct in any multi threaded context right now

crates/sncast/src/starknet_commands/multicall/contract_registry.rs
```diff
@@ -1,16 +1,89 @@
...
+    core::types::ContractClass,
+    providers::{JsonRpcClient, jsonrpc::HttpTransport},
+};
 use starknet_types_core::felt::Felt;
-use std::collections::HashMap;
+use std::{
+    collections::{HashMap, hash_map::Entry},
+    sync::Arc,
+};
+pub struct ContractsCache {
```
ddoktorski:
```suggestion
pub struct ContractCache {
```
similarly as for `ContractRegistry`, plural makes the name a bit clunky

ddoktorski (changes requested):

crates/sncast/src/starknet_commands/multicall/contract_registry.rs
franciszekjob: > I also don't think that we use this struct in any multi threaded context right now

Ok, valid. Removed `Arc` usage.

crates/sncast/src/starknet_commands/multicall/contract_registry.rs
```diff
@@ -1,16 +1,89 @@
...
+    pub async fn get_class_hash_by_address(&mut self, address: &Felt) -> Result<Felt> {
+        match self.address_to_class_hash.entry(*address) {
+            Entry::Occupied(entry) => Ok(*entry.get()),
+            Entry::Vacant(entry) => {
+                let class_hash = get_class_hash_by_address(&self.provider, *address)
+                    .await
+                    .context("Failed to fetch class hash from provider")?;
+                Ok(*entry.insert(class_hash))
+            }
+        }
```
ddoktorski:
```suggestion
        if let Some(hash) = self.address_to_class_hash.get(address) {
            return Ok(*hash);
        }

        let class_hash = get_class_hash_by_address(&self.provider, *address)
            .await
            .context("Failed to fetch class hash from provider")?;
        self.address_to_class_hash.insert(*address, class_hash);
        Ok(class_hash)
```
We can simplify this logic by checking the cache first. This also avoids holding a mutable borrow on self while waiting for the async call

crates/sncast/src/starknet_commands/multicall/deploy.rs
```diff
@@ -0,0 +1,92 @@
...
+use starknet_types_core::felt::Felt;
+
+use crate::starknet_commands::deploy::DeployCommonArgs;
+use crate::starknet_commands::multicall::contract_registry::ContractRegistry;
+use crate::starknet_commands::multicall::invoke::replaced_calldata;
+use crate::{Arguments, calldata_to_felts};
+
+#[derive(Args)]
+pub struct MulticallDeploy {
+    /// Optional identifier to reference this step in later steps
```
ddoktorski: nit:
```suggestion
    /// Optional identifier to reference this contract in later steps
```

crates/sncast/src/main.rs
```diff
@@ -228,6 +221,17 @@ impl Arguments {
     }
 }
 
+pub fn calldata_to_felts(calldata: &[String]) -> Result<Vec<Felt>> {
+    calldata
+        .iter()
+        .map(|data| {
+            Felt::from_dec_str(data)
+                .or_else(|_| Felt::from_hex(data))
+                .context(format!("Failed to parse to felt. Data: {data}"))
```
ddoktorski:
```suggestion
                .context(format!("Failed to parse {data} to felt"))
```

crates/sncast/src/starknet_commands/multicall/run.rs
```diff
@@ -140,17 +146,18 @@ pub async fn run(
...
     let mut parsed_inputs = Vec::new();
     for input in inputs {
         let felt_value = match input {
-            Input::String(s) => contracts_registry
+            Input::String(s) => contract_registry
                 .get_address_by_id(s)
-                .map_or_else(|| s.parse(), Ok)?,
-            Input::Number(n) => (*n).into(),
+                .map_or_else(|| s.parse(), Ok)
+                .expect(""),
```
ddoktorski: Why is expect here?

franciszekjob: Changed, `parse_inputs` returns `Result` now.

crates/sncast/src/starknet_commands/multicall/contract_registry.rs
```diff
@@ -1,16 +1,88 @@
...
 use starknet_types_core::felt::Felt;
-use std::collections::HashMap;
+use std::collections::{HashMap, hash_map::Entry};
+pub struct ContractCache {
+    address_to_class_hash: HashMap<Felt, Felt>,
+    class_hash_to_contract_class: HashMap<Felt, ContractClass>,
+    provider: JsonRpcClient<HttpTransport>,
+}
+
+impl ContractCache {
```
ddoktorski: `ContractCache` is never used independently, so it seems reasonable to move implementation of the below functions (except `new`) to `ContractRegistry`. Then we can make cache private within `ContractRegistry`

crates/sncast/src/starknet_commands/multicall/run.rs
```diff
@@ -140,16 +148,24 @@ pub async fn run(
         .map_err(handle_starknet_command_error)
 }
 
-fn parse_inputs(inputs: &Vec<Input>, contracts_registry: &ContractRegistry) -> Result<Vec<Felt>> {
+fn parse_inputs(inputs: &Vec<Input>, contract_registry: &ContractRegistry) -> Result<Vec<String>> {
```
ddoktorski: Now it takes string input, parses it to felt and then converts it to string 🤔 What is the rationale behind it?

crates/sncast/src/starknet_commands/multicall/run.rs
```diff
@@ -60,77 +76,69 @@ struct InvokeCall {
...
-                );
-
-                if !deploy_call.id.is_empty() {
-                    contracts.insert_new_id_to_address(deploy_call.id.clone(), contract_address)?;
-                }
+    for call in multicall.calls {
+        match call {
+            CallItem::Deploy(item) => {
+                let constructor_calldata = parse_inputs(&item.inputs, &contracts)?;
+                let deploy = MulticallDeploy {
```
ddoktorski: What about adding a `new` function for `MulticallDeploy` and `MulticallInvoke` structs?

crates/sncast/src/starknet_commands/multicall/contract_registry.rs
franciszekjob: https://github.com/foundry-rs/starknet-foundry/pull/4173/changes/ab1ce12c2b87d63d88428874d64e07f62164b442

crates/sncast/src/starknet_commands/multicall/run.rs
franciszekjob: The rationale was that `calldata` in `DeployArguments `and `Arguments` expects a `Vec<String>`. However, you're right that doing this 'round-trip' conversion inside the function doesn't make sense. Refactored in https://github.com/foundry-rs/starknet-foundry/pull/4173/changes/554fcd55ed3c27163887903848082a1e737fe103 so it  now returns `Vec<Felt>` and handle the string mapping only where strictly required by the downstream structs.

crates/sncast/src/starknet_commands/multicall/run.rs
franciszekjob: Makes sense, added `new_from_item` (because we constructor it from `DeployItem`/`InvokeItem`).
https://github.com/foundry-rs/starknet-foundry/pull/4173/changes/a9acba8228de9fbf939b94f6d435bf2d25ddb106

crates/sncast/src/starknet_commands/multicall/contract_registry.rs
```diff
@@ -1,16 +1,29 @@
...
+    providers::{JsonRpcClient, jsonrpc::HttpTransport},
+};
 use starknet_types_core::felt::Felt;
-use std::collections::HashMap;
+use std::collections::{HashMap, hash_map::Entry};
 
-/// Registry for multicall execution, storing mappings from ids to contract addresses.
+/// Registry for managing contract information during multicall execution.
+/// It maintains mappings between user-defined ids, contract addresses, class hashes, and contract classes
+/// to facilitate reference and reuse of deployed contracts across multiple calls in a multicall sequence.
```
DelevoXDG: nit:
```suggestion
/// It stores the following mappings:
/// - user-defined ids to contract addresses
/// - contract addresses to class hashes
/// - class hashes to contract classes
/// to allow referencing and re-using deployed contracts across multiple calls in a multicall sequence.
```

crates/sncast/src/starknet_commands/multicall/contract_registry.rs
```diff
@@ -28,16 +41,71 @@ impl ContractRegistry {
         self.id_to_address.insert(id, address);
         Ok(())
     }
+
+    /// Retrieves the class hash associated with the given contract address.
+    /// Checks the local cache first, and fetches from the provider if not cached.
+    pub async fn get_class_hash_by_address(&mut self, address: &Felt) -> Result<Felt> {
+        if let Some(hash) = self.address_to_class_hash.get(address) {
+            return Ok(*hash);
+        }
```
DelevoXDG: Why not re-use `get_class_hash_by_address_local`? 🤔

crates/sncast/src/starknet_commands/multicall/contract_registry.rs
```diff
@@ -28,16 +41,71 @@ impl ContractRegistry {
...
+
+            self.class_hash_to_contract_class
+                .insert(*class_hash, contract_class);
+        }
+
+        Ok(self
+            .class_hash_to_contract_class
+            .get(class_hash)
+            .expect("Value was just inserted or already existed"))
+    }
```
DelevoXDG: Nit:
I notice that you use different approach here and in the function above - while they seem to do the similar thing. Any particular reason for that? Either is fine, but I'd prefer using one and keeping them aligned.

crates/sncast/src/starknet_commands/multicall/run.rs
```diff
@@ -60,77 +77,31 @@ struct InvokeCall {
...
-                parsed_calls.push(Call {
-                    to: contract_address,
-                    selector: get_selector_from_name(&invoke_call.function)?,
-                    calldata,
-                });
-            }
-            Some(unsupported) => {
-                anyhow::bail!("Unsupported call type found = {unsupported}");
+            CallItem::Invoke(item) => {
+                let invoke = MulticallInvoke::new_from_item(item, &contracts)?;
```
DelevoXDG:
```suggestion
                let invoke = MulticallInvoke::new_from_item(&item, &contracts)?;
```

crates/sncast/src/starknet_commands/multicall/run.rs
```diff
@@ -36,22 +38,37 @@ pub struct Run {
...
+}
+
+#[derive(Deserialize, Debug)]
+struct MulticallFile {
+    #[serde(rename = "call")]
+    calls: Vec<CallItem>,
+}
+
+#[derive(Deserialize, Debug, Getters)]
+#[getset(get = "pub")]
```
DelevoXDG: Nit:
Can you elaborate on your motivation for using `getset` here?

I'm not sure what specifically do you guard against here, but some minor drawbacks are:
- A bit less readable field accessing fields `*item.class_hash()` instead of `item.class_hash`
- Extra compile-time dep

There's also a bit of inconsistency - we're already using manual getters at some places:
https://github.com/foundry-rs/starknet-foundry/blob/308ef94e82a057871d4cb9fe744ee1d773462647/crates/sncast/src/lib.rs#L188-L188

Having different approach is fine I supppose, but I guess it's a bit better to keep it aligned.

Thoughts?

crates/sncast/src/starknet_commands/multicall/run.rs
```diff
@@ -140,14 +111,22 @@ pub async fn run(
...
                 .map_or_else(|| s.parse(), Ok)?,
-            Input::Number(n) => (*n).into(),
+            Input::Number(n) => {
+                if let Ok(n) = Felt::try_from_dec_str(&n.to_string()) {
+                    n
+                } else if let Ok(n) = Felt::from_hex(&n.to_string()) {
+                    n
+                } else {
+                    anyhow::bail!("Failed to parse {n} to felt")
+                }
```
DelevoXDG: this should do the same thing 🤔 wdyt?
```suggestion
                Felt::from_str(&n.to_string())
                    .with_context(|| format!("Failed to parse {} to felt", n))?
```

crates/sncast/src/starknet_commands/utils/dry_run.rs
```diff
@@ -0,0 +1 @@
+pub fn run_with_dry_run() -> () {}
```
DelevoXDG: Unrelated? 😄

crates/sncast/src/starknet_commands/multicall/deploy.rs
```diff
@@ -0,0 +1,116 @@
...
+            common: DeployCommonArgs {
+                contract_identifier: ContractIdentifier {
+                    class_hash: Some(*item.class_hash()),
+                    contract_name: None,
+                },
+                arguments: DeployArguments {
+                    constructor_calldata: Some(
+                        constructor_calldata
+                            .iter()
+                            .map(std::string::ToString::to_string)
```
DelevoXDG: Nit:

```suggestion
                            .map(ToString::to_string)
```

crates/sncast/src/starknet_commands/multicall/invoke.rs
```diff
@@ -0,0 +1,101 @@
...
+
+        Ok(Call {
+            to: self.common.contract_address,
+            selector,
+            calldata,
+        })
+    }
+}
+
+pub fn replaced_calldata(
```
DelevoXDG: It's re-used in in `deploy`, so perhaps let's move it elsewhere? `mod.rs` / new file?

crates/sncast/src/starknet_commands/multicall/invoke.rs
```diff
@@ -0,0 +1,101 @@
...
+                    calldata: Some(replaced_calldata),
+                    arguments: None,
+                }
+            }
+            (None, _) => function_arguments.clone(),
+            (Some(_), Some(_)) => anyhow::bail!(
+                "Invalid arguments: both `calldata` and `arguments` are set. Please provide only one."
+            ),
+        },
+    )
```
DelevoXDG: Nit:
A bit difficult to read. Perhaps add a description for this function for scannability? 🤔

crates/sncast/src/starknet_commands/multicall/deploy.rs
```diff
@@ -0,0 +1,116 @@
...
+            &Arguments::from(self.common.arguments.clone()),
+            contract_registry,
+        )?;
+
+        let constructor_selector = get_selector_from_name("constructor")?;
+        let class_hash = self
+            .common
+            .contract_identifier
+            .class_hash
+            .context("Using deploy with multicall requires providing `--class-hash`")?;
```
DelevoXDG: Since we're for now not deprecating file-based multicall approach - this probably will emit incorrect error messages in the file context, where the message would be for `class_hash` instead.

We should probably consider a solution for this, either some dynamic approach, or just calling it "class hash" so it's universal.

crates/sncast/src/starknet_commands/multicall/contract_registry.rs
franciszekjob: Valid point, changed.

crates/sncast/src/starknet_commands/utils/dry_run.rs
franciszekjob: Yes 🫣, removed

crates/sncast/src/starknet_commands/multicall/invoke.rs
franciszekjob: Sure, added https://github.com/foundry-rs/starknet-foundry/pull/4173/changes/9211fe893dd982b9fe778d860ab92cfa0fb8786b

crates/sncast/src/starknet_commands/multicall/deploy.rs
franciszekjob: Imo it will be enough to change it to  ```Using deploy with multicall requires providing class hash```. Is it fine for you?

crates/sncast/src/starknet_commands/multicall/run.rs
franciszekjob: > Can you elaborate on your motivation for using getset here?

Simply wanted to add auto-generated getters. But seems we can simplify it just by making fields `pub`, no special need to guard them. Changed in 15d28eb07506652f89d19f29906e8d3b58bf22b4

crates/sncast/src/starknet_commands/multicall/contract_registry.rs
franciszekjob: I've aligned both functions to use a consistent check-then-fetch pattern.
I couldn't use a single `entry().or_insert_with()` because the borrow checker doesn't allow holding a mutable reference across an `.await` point. Refactored in 34e10c04d5e83a85319b0a68c4867738a7f7c1f4, TAL please.

crates/sncast/src/starknet_commands/multicall/contract_registry.rs
```diff
@@ -1,16 +1,32 @@
...
-use std::collections::HashMap;
+use std::collections::{HashMap, hash_map::Entry};
 
-/// Registry for multicall execution, storing mappings from ids to contract addresses.
+/// Registry for managing contract information during multicall execution.
+/// It stores the following mappings:
+/// - user-defined ids to contract addresses
+/// - contract addresses to class hashes
+/// - class hashes to contract classes
+///   to allow referencing and re-using deployed contracts across multiple calls in a multicall sequence.
```
ddoktorski, line 15: nit:
```suggestion
/// to allow referencing and re-using deployed contracts across multiple calls in a multicall sequence.
```

crates/sncast/src/starknet_commands/multicall/deploy.rs
```diff
@@ -0,0 +1,116 @@
...
+    /// Optional identifier to reference this contract in later steps
+    #[arg(long)]
+    pub id: Option<String>,
+
+    #[command(flatten)]
+    pub common: DeployCommonArgs,
+}
+
+impl MulticallDeploy {
+    pub fn new_from_item(item: &DeployItem, contracts: &ContractRegistry) -> Result<Self> {
```
ddoktorski, line 30:
```suggestion
    pub fn new(item: &DeployItem, contracts: &ContractRegistry) -> Result<Self> {
```
nit: Since there is no other `new` function it seems natural to name it this way, the same for `MulticallInvoke`

crates/sncast/src/starknet_commands/multicall/run.rs
```diff
@@ -36,101 +38,68 @@ pub struct Run {
...
+            CallItem::Deploy(item) => {
+                let deploy = MulticallDeploy::new_from_item(&item, &contracts)?;
+                let call = deploy.build_call(account, &mut contracts).await?;
+                parsed_calls.push(call);
             }
-            Some(unsupported) => {
-                anyhow::bail!("Unsupported call type found = {unsupported}");
+            CallItem::Invoke(item) => {
+                let invoke = MulticallInvoke::new_from_item(&item, &contracts)?;
+                let call = invoke.build_call(&mut contracts).await?;
```
ddoktorski: nit:
```suggestion
                let call = MulticallInvoke::new_from_item(&item, &contracts)?
                    .build_call(&mut contracts)
                    .await?;
```

crates/sncast/src/starknet_commands/multicall/run.rs
```diff
@@ -36,101 +38,68 @@ pub struct Run {
...
-                parsed_calls.push(Call {
-                    to: contract_address,
-                    selector: get_selector_from_name(&invoke_call.function)?,
-                    calldata,
-                });
+    for call in multicall.calls {
+        match call {
+            CallItem::Deploy(item) => {
+                let deploy = MulticallDeploy::new_from_item(&item, &contracts)?;
+                let call = deploy.build_call(account, &mut contracts).await?;
```
ddoktorski: nit:
```suggestion
                let call = MulticallDeploy::new_from_item(&item, &contracts)?
                    .build_call(account, &mut contracts)
                    .await?;
```

crates/sncast/src/starknet_commands/multicall/mod.rs
```diff
@@ -87,3 +96,32 @@ pub async fn multicall(
...
+pub fn replaced_arguments(
+    arguments: &Arguments,
+    contract_registry: &ContractRegistry,
+) -> Result<Arguments> {
+    Ok(match (&arguments.calldata, &arguments.arguments) {
+        (Some(calldata), None) => {
+            let replaced_calldata = calldata
+                .iter()
+                .map(|input| {
+                    if let Some(address) = contract_registry.get_address_by_id(input) {
```
ddoktorski, line 110: Should it check all elements, not just the ones starting with `@`?

crates/sncast/src/starknet_commands/multicall/contract_registry.rs
DelevoXDG: why? 🤔

crates/sncast/src/starknet_commands/multicall/deploy.rs
DelevoXDG: I prefer the current naming tbh, since unlike many `new()` that just instantiate the struct from passed field values, this one does a little bit of transformation (from item)

crates/sncast/src/starknet_commands/multicall/contract_registry.rs
```diff
@@ -28,16 +44,73 @@ impl ContractRegistry {
...
+    pub async fn get_contract_class_by_class_hash(
+        &mut self,
+        class_hash: &Felt,
+    ) -> Result<&ContractClass> {
+        if self.class_hash_to_contract_class.contains_key(class_hash) {
+            return Ok(self
+                .class_hash_to_contract_class
+                .get(class_hash)
+                .expect("Contract class should exist"));
+        }
```
DelevoXDG, line 88: Why not do as you do above? 🤔 


```suggestion
        if let Some(contract_class) = self.class_hash_to_contract_class.get(class_hash) {
            return Ok(contract_class);
        }
```

crates/sncast/src/starknet_commands/multicall/contract_registry.rs
ddoktorski: Is there any purpose of these leading spaces that I'm not aware of?

crates/sncast/src/starknet_commands/multicall/mod.rs
DelevoXDG: I think the `@` syntax is only intended for cli context? 🤔

crates/sncast/src/main.rs
```diff
@@ -228,6 +221,17 @@ impl Arguments {
...
 }
 
+pub fn calldata_to_felts(calldata: &[String]) -> Result<Vec<Felt>> {
+    calldata
+        .iter()
+        .map(|data| {
+            Felt::from_dec_str(data)
+                .or_else(|_| Felt::from_hex(data))
+                .with_context(|| format!("Failed to parse {data} to felt"))
+        })
```
DelevoXDG: Probably can simplify it same way as you did in `parse_inputs`
```suggestion
        .map(|data| Felt::from_str(data).with_context(|| format!("Failed to parse {data} to felt")))
```

DelevoXDG (approved):
looks good, left few minor things to address

crates/sncast/src/starknet_commands/multicall/contract_registry.rs
DelevoXDG: sorry, my miss, i somehow thought that you suggested adding them, not removing them 😅

crates/sncast/src/starknet_commands/multicall/deploy.rs
franciszekjob: `new` is usually reserved for simple constructors, while `new_from_item` has some additional transformations. Leaving current name.

crates/sncast/src/starknet_commands/multicall/mod.rs
franciszekjob: > Should it check all elements, not just the ones starting with @?

file-based multicall doesn't use `@`, hence such logic.

crates/sncast/src/starknet_commands/multicall/contract_registry.rs
franciszekjob: Borrow checker complains, note that we return `&ContractClass`

crates/sncast/src/starknet_commands/multicall/mod.rs
ddoktorski: what if there is an overlap in id name and one of the calldata value?

crates/sncast/src/starknet_commands/multicall/contract_registry.rs
franciszekjob: Clippy complains, leaving as it is.

crates/sncast/src/starknet_commands/multicall/mod.rs
franciszekjob: I see what you mean -`replaced_calldata` will have an additional flag param, indicating the interface (file-based/CLI). Based on that, we can include (or not) `@` when resolving values. Will introduce it in https://github.com/foundry-rs/starknet-foundry/pull/4174. Does that sound well?

ddoktorski (approved):

PR #4174, opened 2026-02-21
Add CLI interface for multicall

franciszekjob:
Towards #3810

**Stack**
- #4176
- #4174 ⬅️
- #4216
- #4173
- #4172
- #4171 

## Introduced changes

Introduce `sncast multicall execute` which allows to pass calls via CLI arguments

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/sncast/src/starknet_commands/multicall/run_calls.rs
```diff
@@ -0,0 +1,173 @@
...
+
+pub async fn run_calls(
+    tokens: &[String],
+    multicall: &Multicall,
+    provider: &JsonRpcClient<HttpTransport>,
+    account: &SingleOwnerAccount<&JsonRpcClient<HttpTransport>, LocalWallet>,
+    wait_config: WaitForTx,
+    ui: &UI,
+) -> Result<MulticallRunResponse> {
+    let allowed_commands = ["deploy".to_string(), "invoke".to_string()];
```
ddoktorski: What about making a const?
```rust
const ALLOWED_MULTICALL_COMMANDS: [&str; 2] = ["deploy", "invoke"]
```

crates/sncast/src/starknet_commands/multicall/mod.rs
```diff
@@ -21,10 +24,21 @@ use sncast::{
...
+use starknet_types_core::felt::Felt;
 
 #[derive(Args)]
 #[command(about = "Execute multiple calls at once", long_about = None)]
 pub struct Multicall {
+    #[command(flatten)]
+    pub fee_args: FeeArgs,
+
+    #[command(flatten)]
+    pub rpc: RpcArgs,
```
ddoktorski: These options are now duplicated for `multicall run`

crates/sncast/src/main.rs
```diff
@@ -619,6 +619,7 @@ async fn run_async_command(cli: Cli, config: CastConfig, ui: &UI) -> Result<()>
             let selector = get_selector_from_name(&function)
                 .context("Failed to convert entry point selector to FieldElement")?;
 
+            let contract_address = contract_address.parse()?;
```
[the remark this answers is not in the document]
ddoktorski: ⬆️

crates/sncast/tests/e2e/multicall/execute.rs
ddoktorski, line 1: We can also add a test for a specific case with only 1 call

ddoktorski (changes requested):

crates/sncast/tests/e2e/multicall/execute.rs
franciszekjob: [c1cf7d4](https://github.com/foundry-rs/starknet-foundry/commit/c1cf7d486c4149a2dcf28b15ff8e8d2cde469778)

crates/sncast/src/main.rs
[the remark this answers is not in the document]
franciszekjob: [a158e9b](https://github.com/foundry-rs/starknet-foundry/commit/a158e9babbbb0dd5afbc263938bf58dd9270e8f1)

crates/sncast/src/starknet_commands/multicall/mod.rs
franciszekjob: Since these are now present in both `multicall` and `multicall run`, we could consolidate them into just `multicall`. Removing them from `multicall run` would be breaking, but perhaps it’s not too drastic to justify the cleaner structure. WDYT?

ddoktorski: What about `multicall new`? These flags are irrelevant for this command

franciszekjob: That's true.

I see two possible solutions for this matter:

1. **Top-level arguments with manual guards**: Keep args at the `multicall` level and prevent their use with `new`/`run` in code. This preserves currently proposed UX:

```
sncast multicall --network sepolia <COMMANDS>
```

2. **Subcommand-specific arguments**: Require these args to be passed within the individual calls (only in one of them). This changes the UX to:

```
multicall invoke ... --network sepolia / invoke ...
```

Imo we should keep option 1 as it's more intuitive and readable with many calls.

WDYT @ddoktorski @DelevoXDG @MKowalski8 ?

ddoktorski: Can we somehow use this? https://docs.rs/clap/latest/clap/struct.Command.html#method.args_conflicts_with_subcommands

franciszekjob: In standard case yes, but here's we have a dynamic subcommand. Look - if we add  `args_conflicts_with_subcommands` and use 

```
sncast multicall --network sepolia new example.toml
```

then the error is as follows:
```
Command: multicall
Error: Unknown multicall command: 'new'. Expected 'deploy' or 'invoke'.
```

Instead of clap error, `new example.toml` is treated as tokens (parsed later as subcommands), hence the errors about unknown command.

Imo the best way in this situatio is to add guard in code.

ddoktorski: So maybe just remove args from `run` command to have it in one place and for `new` command they'll be simply ignored?

franciszekjob: Initially I was a bit opposed to it since it's breaking. However, the potential confusion coming from the args' presence both on `multicall` and `multicall run` level seems worse overall 🤔 . Fine, will change it.

crates/sncast/src/main.rs
```diff
@@ -617,6 +617,9 @@ async fn run_async_command(cli: Cli, config: CastConfig, ui: &UI) -> Result<()>
             let selector = get_selector_from_name(&function)
                 .context("Failed to convert entry point selector to FieldElement")?;
 
+            let contract_address = contract_address
+                .parse()
+                .context("Failed to parse contract address: expected a hex or decimal string")?;
```
DelevoXDG: Nit: 
I'd prefer to be explicit: 

```suggestion
            let contract_address = Felt::from_str(&contract_address)
                .context("Failed to parse contract address: expected a hex or decimal string")?;
```

crates/sncast/src/starknet_commands/invoke.rs
```diff
@@ -20,7 +20,7 @@ use starknet_types_core::felt::Felt;
 pub struct InvokeCommonArgs {
     /// Address of contract to invoke
     #[arg(short = 'd', long)]
-    pub contract_address: Felt,
+    pub contract_address: String,
```
DelevoXDG: I understand the idea here - we need to support both felts and ids.

But it seems problematic to me that field is very detached from the logic that determines what it is and processes it.

I think it would be cleaner if we defined something like:
```rs
struct FeltOrId(String);
```
with some helpers
```rs
impl FeltOrId {
    pub fn as_felt(&self) -> Option<Felt> { .. }
    pub fn as_id(&self, mode: MulticallMode) -> Option<&str> { .. }
}
```

That way the parsing logic would be centralized, and you wouldn't have to know full context of `resolve_contract_address` to understand what this field represents.

WDYT?

crates/sncast/src/starknet_commands/multicall/mod.rs
```diff
@@ -125,3 +176,29 @@ pub fn replaced_arguments(
...
+    mode: MulticallMode,
+) -> Result<Felt> {
+    let parse_fallback = || contract_address.parse::<Felt>().map_err(Into::into);
+
+    match mode {
+        MulticallMode::File => {
+            contracts
+            .get_address_by_id(contract_address)
+            .map_or_else(parse_fallback, Ok)
+        },
```
DelevoXDG: Why do we handle file differently than the cli variant?

Correct me if I'm wrong, but wouldn't that fallback lead to us outputting something like:
```
expected a hex or decimal string
```
which is misleading, because alphanumeric id *is* the valid format type, but we'd be hiding that.

Besides, it seems to me that in almost all cases ID and Felt do not intersect:
- `0x[..]` -> Felt
- `[1-9]..` -> Felt
- `[a-z0-9]` -> ID

The only problematic part is intersection of numeric ID (is that allowed?) and decimal.

If not, we can just print "Ensure the referenced id is defined in a previous step."
If yes, we can output more generic message like:
```
Invalid contract address: [..]. Expected hex / decimal string, or id defined in the previous step.
```

Thoughts?

crates/sncast/src/starknet_commands/multicall/execute.rs
```diff
@@ -0,0 +1,171 @@
...
+                    .await?;
+                calls.push(call);
+            }
+            _ => bail!("Unknown multicall command: '{cmd_name}'. Expected 'deploy' or 'invoke'."),
+        }
+    }
+
+    if calls.is_empty() {
+        bail!("No valid multicall commands found to execute. Please check the provided commands.");
+    }
```
DelevoXDG: Let's add tests that assert this behavior

crates/sncast/src/starknet_commands/multicall/mod.rs
```diff
@@ -33,6 +49,8 @@ pub struct Multicall {
 pub enum Commands {
     Run(Box<Run>),
     New(New),
+    #[command(external_subcommand)]
+    Calls(Vec<String>),
```
DelevoXDG: Currently, the user has no way to discover that this option exists from CLI perspective:
```
Execute multiple calls

Usage: sncast multicall [OPTIONS] <COMMAND>

Commands:
  run   Execute a multicall from a .toml file
  new   Generate a template for the multicall .toml file
  help  Print this message or the help of the given subcommand(s)

Options:
  ...
  -h, --help
          Print help
```

If discoverability is of importance, here are some proposal for that:

### proposal 1
add `after_help` that informs on this logic

(below is just an example on how it looks)
```rs
#[command(
    about = "Execute multiple calls at once",
    long_about = None,
    after_help = "\
INLINE MULTICALL:
  You can also run calls directly in the cli, separated by `/`:

    sncast multicall invoke --contract-address 0x... --function foo / deploy --class-hash 0x...

  Supported commands: `deploy`, `invoke`."
)]
pub struct Multicall {
```

### proposal 2
We can define a new subcommand, e.g. something like: `sncast multicall execute`

Yes, it's longer, but allows for some separation of concerns and could perhaps be easier to reason about for the user?

We can then define full `--help` message that is contained at `sncast multicall execute` level only.

IMO having one extra keyword is negligible price for discoverability of this feature.

Thoughts?

DelevoXDG: I think proposal 2 should also fix problem discussed in this thread:
https://github.com/foundry-rs/starknet-foundry/pull/4174/changes#r2917759961 

We would have own sets of flags for `new`, `run` and `execute` commands

CHANGELOG.md
```diff
@@ -16,6 +16,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Bug with invalid function name mappings for functions with `#[test]` attribute
 
+### Cast
```
DelevoXDG, line 27: Should we also mention that we added support for CLI multicall?

crates/sncast/src/starknet_commands/multicall/mod.rs
franciszekjob: Discussed offline: will implement option 2 - mentioned aspects are more important the keeping the same command name.

crates/sncast/src/starknet_commands/multicall/execute.rs
franciszekjob: 62391aebbd020989df6eb1f39be63e2786526018

crates/sncast/src/starknet_commands/invoke.rs
franciszekjob: Makes sense - added https://github.com/foundry-rs/starknet-foundry/pull/4216 for it, not to bundle so many changes in this PR.

crates/sncast/src/starknet_commands/multicall/mod.rs
franciszekjob: Resolving, since this problem is not present anymore after introducing `multicall execute`

crates/sncast/src/starknet_commands/multicall/mod.rs
franciszekjob: After adding #4216, this problem is not present anymore. Id now always requires `@` prefix (regardless if it's file or CLI mode).

crates/sncast/src/starknet_commands/utils/felt_or_id.rs
```diff
@@ -0,0 +1,41 @@
...
+
+const ID_PREFIX: char = '@';
+
+#[derive(Deserialize, Debug, Clone)]
+pub struct FeltOrId(String);
+
+impl FeltOrId {
+    pub fn new(s: String) -> Self {
+        FeltOrId(s)
+    }
```
ddoktorski, line 14: Should be moved into the impl block below

crates/sncast/src/starknet_commands/multicall/execute.rs
```diff
@@ -0,0 +1,197 @@
...
+            "invoke".to_string(),
+            "--contract-address".to_string(),
+            "0xabc".to_string(),
+            "--function".to_string(),
+            "my_function".to_string(),
+            "/".to_string(),
+            "deploy".to_string(),
+            "--class-hash".to_string(),
+            "0x456".to_string(),
+        ];
```
ddoktorski: nit:
```suggestion
        let tokens = vec![
            "deploy",
            "--class-hash",
            "0x123",
            "/",
            "invoke",
            "--contract-address",
            "0xabc",
            "--function",
            "my_function",
            "/",
            "deploy",
            "--class-hash",
            "0x456",
        ]
        .into_iter()
        .map(String::from)
        .collect::<Vec<_>>();
```

crates/sncast/src/starknet_commands/multicall/execute.rs
```diff
@@ -0,0 +1,197 @@
...
+
+    let matches = cmd
+        .try_get_matches_from(argv)
+        .with_context(|| format!("Failed to parse args for `{command_name}`"))?;
+
+    T::from_arg_matches(&matches).map_err(Into::into)
+}
+
+#[cfg(test)]
+mod tests {
```
ddoktorski, line 175: These tests could be extended to cover various edge cases such as: leading `/`, trailing `/`, consecutive `/ /` etc.

crates/sncast/src/starknet_commands/multicall/mod.rs
```diff
@@ -94,28 +99,43 @@ pub async fn multicall(
...
+        }
     }
 }
 
 /// Replaces arguments that reference user-defined ids with their corresponding values from the contract registry.
-pub fn replaced_arguments(
-    arguments: &Arguments,
-    contract_registry: &ContractRegistry,
-) -> Result<Arguments> {
+pub fn replaced_arguments(arguments: &Arguments, conracts: &ContractRegistry) -> Result<Arguments> {
```
ddoktorski:
```suggestion
pub fn replaced_arguments(arguments: &Arguments, contracts: &ContractRegistry) -> Result<Arguments> {
```

crates/sncast/src/starknet_commands/multicall/mod.rs
ddoktorski: Isn't it unreachable?

crates/sncast/src/starknet_commands/multicall/mod.rs
```diff
@@ -94,28 +99,43 @@ pub async fn multicall(
...
                 .iter()
-                .map(|input| {
-                    if let Some(address) = contract_registry.get_address_by_id(input) {
-                        Ok(address.to_string())
-                    } else {
-                        Ok(input.clone())
-                    }
-                })
-                .collect::<Result<Vec<String>>>()?;
+                .map(|input| FeltOrId::new(input.clone()).as_id(conracts))
```
ddoktorski: We currently convert all non id values to felt just to convert them back to strings in the return below. We should modify only id strings here without touching the rest of the calldata

crates/sncast/src/starknet_commands/multicall/mod.rs
ddoktorski, line 154: Is it possible to reference id when passing `--arguments` flag?

crates/sncast/tests/e2e/multicall/execute.rs
```diff
@@ -0,0 +1,342 @@
...
+        "--account",
+        "my_account",
+        "multicall",
+        "execute",
+        "--url",
+        URL,
+    ];
+
+    let snapbox = runner(&args)
+        .env("SNCAST_FORCE_SHOW_EXPLORER_LINKS", "1")
```
ddoktorski: nit:
```suggestion
```
also in other tests

ddoktorski (changes requested):

crates/sncast/src/starknet_commands/multicall/mod.rs
franciszekjob: Correct, changed.

franciszekjob: ATM it's not possible. Do you want to extend the interface to allow so (I don't mean specifically in this PR)?

crates/sncast/src/starknet_commands/multicall/mod.rs
franciszekjob: Correct, changed it so we don't make redundant conversion.

crates/sncast/src/starknet_commands/multicall/execute.rs
franciszekjob: Added

crates/sncast/tests/e2e/multicall/execute.rs
franciszekjob: It's indeed needed to show the explorer links, why would you like to remove it 🤔 ?

crates/sncast/src/starknet_commands/multicall/mod.rs
ddoktorski: It's fine, but make it very clear in documenation and `--help` output that id can be used only with `--calldata` flag

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 #### Added
 - `--max-threads` flag to control the maximum number of threads used for test execution
+- `sncast multicall execute`, which allows to pass calls directly as CLI arguments. Read more [here](https://foundry-rs.github.io/starknet-foundry/starknet/multicall.html#multicall-with-cli-arguments)
```
ddoktorski: It's under `Forge` section

crates/sncast/src/starknet_commands/multicall/mod.rs
```diff
@@ -94,34 +98,65 @@ pub async fn multicall(
...
+                *execute.clone(),
+                &account,
+                &provider,
+                wait_config,
+                ui,
+            )
+            .await;
+            let block_explorer_link =
+                block_explorer_link_if_allowed(&result, provider.chain_id().await?, &config).await;
+            process_command_result("multicall", result, ui, block_explorer_link);
```
ddoktorski:
```suggestion
            process_command_result("multicall execute", result, ui, block_explorer_link);
```

crates/sncast/src/starknet_commands/multicall/execute.rs
```diff
@@ -0,0 +1,297 @@
...
+) -> Result<MulticallRunResponse> {
+    let command_groups = extract_commands_groups(&execute.tokens, "/", &ALLOWED_MULTICALL_COMMANDS);
+
+    let mut contract_registry = ContractRegistry::new(provider);
+    let mut calls = vec![];
+
+    for group in command_groups {
+        if group.is_empty() {
+            continue;
+        }
```
ddoktorski: Isn't it a dead check given that we omit empty vectors when appending in `extract_commands_groups`?

crates/sncast/src/starknet_commands/multicall/execute.rs
```diff
@@ -0,0 +1,297 @@
...
+    }
+
+    if calls.is_empty() {
+        bail!("No valid multicall commands found to execute. Please check the provided commands.");
+    }
+
+    execute_calls(
+        account,
+        calls,
+        execute.fee_args.clone(),
```
ddoktorski: It seems redundant
```suggestion
        execute.fee_args,
```

crates/sncast/src/starknet_commands/multicall/mod.rs
```diff
@@ -94,34 +98,65 @@ pub async fn multicall(
...
         (Some(calldata), None) => {
             let replaced_calldata = calldata
                 .iter()
-                .map(|input| {
-                    if let Some(address) = contract_registry.get_address_by_id(input) {
-                        Ok(address.to_string())
+                .map(|raw_input| {
+                    let input = FeltOrId::new(raw_input.clone());
+                    if let Some(id) = input.as_id() {
+                        contracts
```
ddoktorski, line 137:
```suggestion
                    if let Some(id) = raw_input.strip_prefix(ID_PREFIX) {
                        contracts
```
nit: In this case it'd be easier to call it directly and remove the `new` function as it's only used here

crates/sncast/tests/e2e/multicall/execute.rs
ddoktorski: I meant the tests that don't assert it, it's good now

crates/sncast/src/starknet_commands/multicall/execute.rs
```diff
@@ -0,0 +1,297 @@
...
+            "--function",
+            "my_function",
+            "/",
+            "deploy",
+            "--class-hash",
+            "0x456",
+        ]
+        .into_iter()
+        .map(String::from)
+        .collect::<Vec<_>>();
```
ddoktorski: nit: Now when there are more tests it can be extracted as util function

crates/sncast/src/starknet_commands/multicall/execute.rs
```diff
@@ -0,0 +1,297 @@
...
+}
+
+fn parse_args<T>(command_name: &str, tokens: &[String]) -> anyhow::Result<T>
+where
+    T: Args + FromArgMatches,
+{
+    let cmd = T::augment_args(Command::new(command_name.to_string()));
+
+    let argv = std::iter::once(command_name.to_string())
+        .chain(tokens.iter().cloned())
```
ddoktorski: nit:
```suggestion
    let argv = std::iter::once(command_name)
        .chain(tokens.iter().map(String::as_str))
```

CHANGELOG.md
franciszekjob: It was good before, had to break after merging 🤦‍♂️

crates/sncast/src/starknet_commands/multicall/execute.rs
franciszekjob: Removed, we shouldn't assume the group may be empty.

crates/sncast/src/starknet_commands/multicall/mod.rs
franciszekjob: Imo it's better to keep the stripping (i.e. id-related logic) in one place (so in `FeltOrId`) - leaving current code.

crates/sncast/src/starknet_commands/multicall/execute.rs
```diff
@@ -0,0 +1,261 @@
...
+    pub fee_args: FeeArgs,
+
+    #[command(flatten)]
+    pub rpc: RpcArgs,
+
+    /// Nonce of the transaction. If not provided, nonce will be set automatically
+    #[arg(short, long)]
+    pub nonce: Option<Felt>,
+
+    /// The multicall arguments. Subsequent calls should be separated by a '/' token.
```
ddoktorski, line 39: Let's add information that arguments are the same as to `invoke` and `deploy` commands

ddoktorski (approved):

crates/sncast/src/starknet_commands/multicall/execute.rs
```diff
@@ -0,0 +1,266 @@
...
+    /// Currently, `invoke` and `deploy` calls are supported. Their syntax is the same
+    /// as for `sncast invoke` and `sncast deploy` commands (with additional id argument for deploy calls).
```
DelevoXDG: I don't want to block this, so these are ideas for separate issue - but I think it would be good to eventually have a better help message here.

We can make it a bit more directional + document `--id` syntax better.

Something like:
```suggestion
Supported commands: `invoke`, `deploy`.
  Their syntax is the same as `sncast invoke` and `sncast deploy`. 
  Use `sncast  invoke --help` and `sncast deploy --help` for reference.
Additionally, `deploy` supports `--id <ID>` argument.
  Next, `@id` can be passed to `--contract-address` and `--calldata` flags to reference deploy contract address in subsequent calls. 
```

We can also consider adding a link to docs perhaps?

But, as I said, to avoid blocking release it can be done separately.

MKowalski8 (approved):

crates/sncast/src/starknet_commands/multicall/execute.rs
DelevoXDG: We can make it a bit clearer + document `--id` syntax better.

Something like:
```suggestion
    /// The multicall arguments. Subsequent calls should be separated by a '/' token.
    ///
    /// Supported commands: `invoke`, `deploy`.
    /// Their syntax is the same as `sncast invoke` and `sncast deploy`.
    /// Use `sncast  invoke --help` and `sncast deploy --help` for reference.
    ///
    /// Additionally, `deploy` supports `--id <ID>` argument to name the deployed contract in this multicall.
    /// In subsequent calls, `@<ID>` can be referenced in `--contract-address` and `--calldata` flags to reference the deployed contract address.
```

We can also consider adding a link to docs perhaps?

But we can consider this as separate issue - to avoid blocking release.

DelevoXDG (approved):

crates/sncast/src/starknet_commands/multicall/execute.rs
franciszekjob: > We can also consider adding a link to docs perhaps?

Why not, added in fda4459500b93b3f2c02f6f0792a392ab7879a55

Issue #4027, opened 2025-12-16
Research and refactor sncast config logic


franciszekjob:
### Current State

Research (whether such approach is valid) and potentially implement new logic for sncast configs. The concept is as follows:
- config from .toml file should be ready
- config (partial) from passed flags should be read
- based on these 2 configs, a functional config should be created; it would be the one actually used in the commands

### Objective

Research and refactor sncast config logic

PR #4175, opened 2026-02-23
Refactor sncast config logic

DelevoXDG:
Towards #4027

## Stack
- #4175
- #4205 
- #4249 
- #4234 

#### Outline

This is initial refactor of cast config handling. 

It aims to make the logic easier to reason about, removes tricky defaults resolution (using `None` instead), removes manual raw string approach config construction, and paves way for fixes introduced in #4205 

Subsequent PR introduces more significant changes that change how configs are actually handled in more breaking way in attempt to address various issues, while this one mainly tries to keep current logic intact.

#### Changes
- Add `PartialCastConfig` with all fields having `None` as `Default` value. 
- Create global, local, cli `PartialCastConfig`. Then combine them (add `Override` trait for config and it's subcomponents)
- Replace raw string approach to config in `add_created_profile_to_configuration`

## Checklist

- [x] Linked relevant issue
- [ ] Added relevant tests


crates/sncast/tests/e2e/account/create.rs
```diff
@@ -276,11 +276,13 @@ pub async fn test_happy_case_add_profile(rpc_flag: &str, rpc_value: &str) {
...
     let expected_lines = [
         "[sncast.my_account]",
+        &format!("{} = \"{}\"", rpc_flag.trim_start_matches("--"), rpc_value),
         "account = \"my_account\"",
         "accounts-file = \"accounts.json\"",
-        &format!("{} = \"{}\"", rpc_flag.trim_start_matches("--"), rpc_value),
     ];
     let expected_block = expected_lines.join("\n");
+    println!("Expected:\n{expected_block}");
+    println!("Contents:\n{contents}");
```
franciszekjob: Is this needed?

crates/sncast/src/helpers/configuration.rs
```diff
@@ -1,16 +1,62 @@
...
+impl Override for NetworkParams {
+    fn override_with(&self, other: NetworkParams) -> NetworkParams {
+        if other == NetworkParams::default() {
+            self.clone()
+        } else {
+            other
+        }
+    }
+}
+
```
franciszekjob, line 59: We can make invalid states unrepresentable by collapsing `url` and `network` into a single `Option<NetworkSource>` enum. This removes the need for manual `validate()` checks and simplifies `override_with`.

```rust
#[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
#[serde(untagged)]
pub enum NetworkSource {
    Url(Url),
    Network(Network),
}

pub struct NetworkParams {
    #[serde(flatten)]
    pub target: Option<NetworkSource>,
}
```

and later override becomes

```rust
fn override_with(&self, other: NetworkParams) -> NetworkParams {
    if other.target.is_some() { other } else { self.clone() }
}
```

Thoughts?

crates/sncast/src/helpers/configuration.rs
```diff
@@ -133,6 +181,90 @@ impl Config for CastConfig {
...
+            show_explorer_links: other.show_explorer_links.or(self.show_explorer_links),
+            networks: override_optional(self.networks.clone(), other.networks),
+        }
+    }
+}
+
+impl PartialCastConfig {
+    pub fn local(opts: &CliConfigOpts) -> Result<Self> {
+        let local_config = load_config::<PartialCastConfig>(None, opts.profile.as_deref())
+            .map_err(|err| anyhow::anyhow!(format!("Failed to load config: {err}")))?;
```
franciszekjob, line 212: nit: (same below):
```suggestion
            .map_err(|err| anyhow::anyhow!("Failed to load config: {err}"))?;
```

crates/sncast/src/helpers/configuration.rs
```diff
@@ -133,6 +181,90 @@ impl Config for CastConfig {
     }
 }
 
+impl PartialCastConfig {
+    pub fn validate(&self) -> anyhow::Result<()> {
+        block_explorer::Service::validate_for_config(self.block_explorer)?;
+        self.wait_params.map(ValidatedWaitParams::from);
```
franciszekjob, line 188: Why is this needed?

Cargo.toml
```diff
@@ -115,6 +115,7 @@ sha2 = "0.10.8"
 base16ct = { version = "1.0.0", features = ["alloc"] }
 async-trait = "0.1.87"
 serde_path_to_error = "0.1.20"
+serde_with = "3.16.1"
```
franciszekjob: nit: Maybe use newer version [3.18.0](https://github.com/jonasbb/serde_with/releases/tag/v3.18.0)?

crates/sncast/src/lib.rs
```diff
@@ -147,30 +148,52 @@ pub struct WaitForTx {
...
+        )
+    }
+}
+
+/// Effective wait params used at runtime.
+/// Note: Built from [`PartialWaitParams`], not (de)serialized.
+#[derive(Clone, Debug, Copy, PartialEq)]
+pub struct ValidatedWaitParams {
+    timeout: u16,
     retry_interval: u8,
```
franciszekjob, line 185: How about ensuring non-zeroness on type level 😄 ?

```suggestion
    timeout: NonZeroU16,
    retry_interval: NonZeroU8,
```

crates/sncast/src/lib.rs
franciszekjob, line 205: Unrelated to changes from this PR, but the name of this should be `elapsed_time`, since it measures how much time has passed.

crates/sncast/tests/e2e/account/create.rs
DelevoXDG: Nope, good catch, removed

Cargo.toml
DelevoXDG: Ok

crates/sncast/src/helpers/configuration.rs
DelevoXDG: Let's skip this one: the [next PR](https://github.com/foundry-rs/starknet-foundry/pull/4205/changes) already removes this logic and replaces it with `with_context` uses

crates/sncast/src/helpers/configuration.rs
DelevoXDG: I’m not sure this is the best way to do it, but `ValidatedWaitParams::from` calls `ValidatedWaitParams::new`, which then calls `ValidatedWaitParams::validate`.

So essentially we're validating wait params at partial config level this way

crates/sncast/src/helpers/configuration.rs
DelevoXDG: Discussed offline, not going to implement for now, as it seems there's no clean way to do it.

You'd need to implement custom deserialization. Otherwise, the enum will silently take the first value it finds from either:
```
network = ...
url = ...
```
So, as workaround, you'd have to keep both raw and enum-based network params:
```rs
pub enum NetworkSource {
    Url { url: Url },
    Network { network: Network },
}

struct NetworkParamsRaw {
    #[serde(skip_serializing_if = "Option::is_none")]
    url: Option<Url>,
    #[serde(skip_serializing_if = "Option::is_none")]
    network: Option<Network>,
}
```
and all the logic to convert from one to the other:
```rs
    fn try_from(raw: NetworkParamsRaw) -> Result<Self, Self::Error> {
        match (raw.url, raw.network) {
            (Some(_), Some(_)) => {
                Err("Only one of `url` or `network` may be specified".to_string())
            }
            (Some(url), None) => Ok(Self {
                source: Some(NetworkSource::Url { url }),
            }),
            (None, Some(network)) => Ok(Self {
                source: Some(NetworkSource::Network { network }),
            }),
            (None, None) => Ok(Self::default()),
        }
    }
```
which I'm not sure is beneficial compared to current approach.

crates/sncast/src/helpers/configuration.rs
```diff
@@ -28,45 +74,77 @@ impl NetworksConfig {
...
+            mainnet: other.mainnet.or_else(|| self.mainnet.clone()),
+            sepolia: other.sepolia.or_else(|| self.sepolia.clone()),
+            devnet: other.devnet.or_else(|| self.devnet.clone()),
         }
     }
 }
 
-#[derive(Deserialize, Serialize, Clone, Debug, PartialEq)]
+/// Effective config used at runtime.
+/// Note: Built from [`PartialCastConfig`], not (de)sereliazed.
```
MKowalski8: Typo
```suggestion
/// Note: Built from [`PartialCastConfig`], not (de)serialized.
```

crates/sncast/src/helpers/configuration.rs
```diff
@@ -28,45 +74,77 @@ impl NetworksConfig {
...
+    pub wait_params: ValidatedWaitParams,
+    pub block_explorer: Option<block_explorer::Service>,
+    pub show_explorer_links: bool,
+    pub networks: NetworksConfig,
+}
 
-    #[serde(default)]
-    pub network: Option<Network>,
+impl CastConfig {
+    pub fn validate(&self) -> anyhow::Result<()> {
```
MKowalski8, line 104: Is `validate` for `CastConfig` even used? I deleted it, and it compiled

crates/sncast/src/helpers/configuration.rs
```diff
@@ -76,52 +154,22 @@ pub struct CastConfig {
...
-        default = "show_explorer_links_default",
+        default,
         rename(serialize = "show-explorer-links", deserialize = "show-explorer-links")
     )]
     /// Print links pointing to pages with transaction details in the chosen block explorer
-    pub show_explorer_links: bool,
+    pub show_explorer_links: Option<bool>,
 
     #[serde(default)]
-    /// Configurable urls of predefined networks - mainnet, sepolia, and devnet are supported
```
MKowalski8, line 86: Why this comment was removed? Is it not true anymore?

crates/sncast/src/helpers/configuration.rs
DelevoXDG: It isn't, but it should be.

I've added it to `PartialCastConfig` -> `CastConfig` conversion, i.e.`CastConfig::try_from(PartialCastConfig)` in the next PR: f9ed6f8781b3b24ccb8c137af380f7914a3ff915

Note that it currently didn't cause any errors and it cannot be tested - the current config loading and validation logic ensures that all possible validation errors are caught at earlier steps (e.g. `ValidatedWaitParams::try_from` in `CastConfig::validate` or using `PartialCastConfig::validate` to catch earlier issues).

That is however not a guarantee that config won't be extended in the future and some issue could be detectable only when effective config is built.

Hence why I think we should keep this validation (even if it's redundant for now).

crates/sncast/src/helpers/configuration.rs
DelevoXDG: Not sure, perhaps a mistake 😅 Added it back

crates/sncast/src/lib.rs
DelevoXDG: Applied in the next PR https://github.com/foundry-rs/starknet-foundry/pull/4205/changes/a0bbb780de987afadbca3769437fce386444cb5a

crates/sncast/src/lib.rs
DelevoXDG: Nice catch, I've added the issue for this: https://github.com/foundry-rs/starknet-foundry/issues/4289

crates/sncast/src/starknet_commands/show_config.rs
```diff
@@ -28,8 +28,12 @@ pub async fn show_config(
...
     };
 
-    let rpc_url = show.rpc.url.clone().or(cast_config.url);
-    let network = show.rpc.network.or(cast_config.network);
+    let rpc_url = show
+        .rpc
+        .url
+        .clone()
+        .or(cast_config.network_params.url().cloned());
+    let network = show.rpc.network.or(cast_config.network_params.network());
```
franciszekjob: That can produce an invalid effective config containing both values at once. For example, if the config has `network = "sepolia"` and the user runs `show-config --url http://...`, the command will display both the CLI URL and the config network, even though runtime resolution treats these options as mutually exclusive and CLI input is supposed to replace the config choice as a whole. `show-config` should use the same atomic resolution path as the rest of the RPC config handling.

e.g. I have such profile
```toml
[sncast.default]
network = "sepolia"
account = "argent-testnet-account"
url = "http://127.0.0.1:5050"
```

and running `show-config` fails:
```shell
➜  starknet-foundry git:(zdobnikau/refactor-sncast-config) ✗ target/debug/sncast show-config --url http://127.0.0.1:5050
Error: Failed to load config: Only one of `url` or `network` may be specified
```

but if I remove `url` from profile

```toml
[sncast.default]
network = "sepolia"
account = "argent-testnet-account"
```

and pass it via `--url`, it succeeds

```shell
➜  starknet-foundry git:(zdobnikau/refactor-sncast-config) ✗ target/debug/sncast show-config --url http://127.0.0.1:5050
RPC URL:             http://127.0.0.1:5050/
Network:             sepolia
Account:             argent-testnet-account
Accounts File Path:  /Users/franciszekjob/.starknet_accounts/starknet_open_zeppelin_accounts.json
Wait Timeout:        300s
Wait Retry Interval: 5s
Show Explorer Links: true
Block Explorer:      Voyager
```

crates/sncast/src/lib.rs
```diff
@@ -147,30 +148,52 @@ pub struct WaitForTx {
...
-        );
-
-        Self {
+        let res = Self {
             timeout,
             retry_interval,
-        }
+        };
+        res.validate();
+        res
```
franciszekjob, line 196: `PartialCastConfig::validate()` eagerly converts `self.wait_params` into `ValidatedWaitParams`, which immediately validates the value and can panic for a partial config that would become valid after later overrides are merged.

Repro:
```toml
[sncast.default]
network = "sepolia"
account = "argent_testnet_account"
wait-params = { timeout = 2 }
```

then run:

```shell
➜  starknet-foundry git:(zdobnikau/refactor-sncast-config) ✗ target/debug/sncast --wait-retry-interval 1 show-config


thread 'main' (140636) panicked at crates/sncast/src/lib.rs:220:9:
Invalid values for retry_interval and/or timeout!
```

crates/sncast/src/starknet_commands/show_config.rs
DelevoXDG: Good catch, I've completely missed that this happens!

Fixed network params resolution at `show-config` level.

crates/sncast/src/lib.rs
DelevoXDG: > `PartialCastConfig::validate()` eagerly converts `self.wait_params` into `ValidatedWaitParams`, which immediately validates the value and can panic for a partial config that would become valid after later overrides are merged.

You right, this implicitly validates against default `retry-interval` value, which is incorrect.

I've updated validation logic for `PartialWaitParams` to fix this. For simplicity though - I applied it to the next PR: 7b10f7db

franciszekjob (approved):

crates/sncast/src/helpers/configuration.rs
MKowalski8: > I've added it to PartialCastConfig -> CastConfig conversion, i.e.CastConfig::try_from(PartialCastConfig) in the next PR:

Okay, I am satisfied then 😅

> That is however not a guarantee that config won't be extended in the future and some issue could be detectable only when effective config is built.
> Hence why I think we should keep this validation (even if it's redundant for now).

Philosophically speaking, I'm not a fan of this approach. In the case of this type of code, gathering context isn't a problem either. So, I'm closer to [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) principle

But yeah it's used now, so no problem

MKowalski8 (approved):

PR #4179, opened 2026-02-24
Adjust existing cheatnet tests

ddoktorski:
Related #3837

### DIFF
--- a/crates/cheatnet/tests/common/mod.rs
+++ b/crates/cheatnet/tests/common/mod.rs
@@ -112,7 +112,7 @@ pub fn deploy_contract(
     let mut entry_point_execution_context = build_context(
         &cheatnet_state.block_info,
         None,
-        &TrackedResource::CairoSteps,
+        &TrackedResource::SierraGas,
     );
     let hints = HashMap::new();
 
@@ -143,7 +143,7 @@ pub fn deploy(
     let mut entry_point_execution_context = build_context(
         &cheatnet_state.block_info,
         None,
-        &TrackedResource::CairoSteps,
+        &TrackedResource::SierraGas,
     );
     let hints = HashMap::new();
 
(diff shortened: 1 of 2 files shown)


ddoktorski:

- #4113
    - #4129
        - #4120
            - #4179 ◀
                - #4180

crates/cheatnet/tests/common/mod.rs
```diff
@@ -321,12 +323,13 @@ pub fn call_contract_raw(
...
     );
 
-    non_reverting_execute_call_entry_point(
+    execute_call_entry_point(
         &mut entry_point,
         syscall_hint_processor.base.state,
         cheatnet_state,
         syscall_hint_processor.base.context,
         &mut (i64::MAX as u64),
+        &ExecuteCallEntryPointExtraOptions::default(),
```
franciszekjob, line 332: nit: In some places we use explicit initialization instead of default, e.g. here
https://github.com/foundry-rs/starknet-foundry/blob/60c7d522925524be41d76e67b321404fcc68b8b7/crates/cheatnet/src/runtime_extensions/native/call.rs#L26-L34
or here 
https://github.com/foundry-rs/starknet-foundry/blob/60c7d522925524be41d76e67b321404fcc68b8b7/crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/rpc.rs#L217-L225.

Would it make sense to unify passed arg representing `ExecuteCallEntryPointExtraOptions` to `execute_call_entry_point` across all invocations? Seems we can use default everywhere?

franciszekjob (approved):

DelevoXDG (approved):

crates/cheatnet/tests/common/mod.rs
ddoktorski: That can make sense for the core code in `src` directory, but since this is just a test helper the passed value doesn't really matter here. No need for any changes on this for now

PR #4187, opened 2026-02-26
Add `sncast ledger` command

MKowalski8:
---
**Stack**:
- #4190
- #4189
- #4193
- #4187 ⬅
---

### DIFF
--- a/crates/sncast/Cargo.toml
+++ b/crates/sncast/Cargo.toml
@@ -13,7 +13,7 @@ clap_complete.workspace = true
 console.workspace = true
 serde_json.workspace = true
 serde.workspace = true
-starknet-rust.workspace = true
+starknet-rust = { workspace = true, features = ["ledger"] }
 tokio.workspace = true
 url.workspace = true
 rand.workspace = true
@@ -45,6 +45,7 @@ cairo-vm.workspace = true
 blockifier.workspace = true
 semver.workspace = true
 sha3.workspace = true
+sha2.workspace = true
 base16ct.workspace = true
 starknet-rust-crypto.workspace = true
 async-trait.workspace = true
@@ -64,6 +65,7 @@ tracing-log.workspace = true
 strum.workspace = true
 strum_macros.workspace = true
 primitive-types.workspace = true
+coins-ledger.workspace = true
 
 [dev-dependencies]
 ctor.workspace = true
@@ -80,6 +82,7 @@ packages_validation = { path = "../testing/packages_validation" }
 
 [features]
 default = []
+ledger-emulator = []
 
 [[bin]]
 name = "sncast"
--- a/crates/sncast/src/helpers/ledger/emulator_transport.rs
+++ b/crates/sncast/src/helpers/ledger/emulator_transport.rs
@@ -1,0 +1,90 @@
+//! This module is only used behind the `ledger-emulator` feature flag.
+//! It is used to test the ledger commands locally without having to connect to a real Ledger device.
+use async_trait::async_trait;
+use coins_ledger::{
+    APDUAnswer, APDUCommand, LedgerError,
+    transports::{LedgerAsync, native::NativeTransportError},
+};
+use reqwest::Client;
+use serde::Serialize;
+use starknet_rust::signers::ledger::{LedgerError as StarknetLedgerError, LedgerStarknetApp};
+use std::io::Error as IoError;
+
+#[derive(Debug)]
+pub struct SpeculosHttpTransport {
+    client: Client,
+    base_url: String,
+}
+
+#[derive(Serialize)]
+struct ApduRequest {
+    data: String,
+}
+
+fn io_err(e: impl Into<Box<dyn std::error::Error + Send + Sync>>) -> LedgerError {
+    LedgerError::NativeTransportError(NativeTransportError::Io(IoError::other(e)))
+}
+
+impl SpeculosHttpTransport {
+    pub fn new(url: String) -> Result<Self, LedgerError> {
+        let client = Client::builder().build().map_err(io_err)?;
+        Ok(Self {
+            client,
+            base_url: url,
+        })
+    }
+}
+
+#[async_trait]
+impl LedgerAsync for SpeculosHttpTransport {
+    async fn init() -> Result<Self, LedgerError> {
+        let url = std::env::var("LEDGER_EMULATOR_URL")
+            .unwrap_or_else(|_| "http://127.0.0.1:5001".to_string());
+        Self::new(url)
+    }
+
+    async fn exchange(&self, command: &APDUCommand) -> Result<APDUAnswer, LedgerError> {
+        let request = ApduRequest {
+            data: const_hex::encode(command.serialize()),
+        };
+
+        let resp = self
+            .client
+            .post(format!("{}/apdu", self.base_url))
+            .json(&request)
+            .send()
+            .await
+            .map_err(io_err)?;
+
+        if !resp.status().is_success() {
+            return Err(io_err(format!("HTTP Error: {}", resp.status())));
+        }
+
+        let body: serde_json::Value = resp.json().await.map_err(io_err)?;
+
+        if let Some(err) = body.get("error").and_then(|v| v.as_str()) {
+            return Err(io_err(format!("Speculos Error: {err}")));
+        }
+
+        let data_str =
+            body.get("data")
+                .and_then(|v| v.as_str())
+                .ok_or(LedgerError::NativeTransportError(
+                    NativeTransportError::Comm("Missing data field"),
+                ))?;
+
+        let answer = const_hex::decode(data_str).map_err(io_err)?;
+
+        APDUAnswer::from_answer(answer).map_err(|_| {
+            LedgerError::NativeTransportError(NativeTransportError::Comm("Invalid APDU response"))
+        })
+    }
+
+    fn close(self) {}
+}
+
+pub async fn emulator_ledger_app()
+-> Result<LedgerStarknetApp<SpeculosHttpTransport>, StarknetLedgerError> {
+    let transport = SpeculosHttpTransport::init().await?;
+    Ok(LedgerStarknetApp::from_transport(transport))
+}
--- a/crates/sncast/src/helpers/ledger/hd_path.rs
+++ b/crates/sncast/src/helpers/ledger/hd_path.rs
@@ -1,0 +1,468 @@
+use std::str::FromStr;
+
+use crate::response::ui::UI;
+use anyhow::{Result, anyhow, bail};
+use clap::{Arg, Command, Error, builder::TypedValueParser, error::ErrorKind};
+use foundry_ui::components::warning::WarningMessage;
+use sha2::{Digest, Sha256};
+use starknet_rust::signers::DerivationPath;
+
+const EIP2645_LENGTH: usize = 6;
+
+/// BIP-32 encoding of `2645'`
+const EIP_2645_PURPOSE: u32 = 0x8000_0a55;
+
+/// The BIP-32 hardened derivation bit.
+const HARDENED_BIT: u32 = 1 << 31;
+
+/// Mask to extract the non-hardened index from a BIP-32 path component.
+const NON_HARDENED_MASK: u32 = 0x7fff_ffff;
+
+#[derive(Clone)]
+pub struct DerivationPathParser;
+
+/// A parsed derivation path together with any warnings produced during parsing.
+#[derive(Debug, Clone)]
+pub struct ParsedDerivationPath {
+    pub path: DerivationPath,
+    pub warnings: Vec<String>,
+}
+
+impl ParsedDerivationPath {
+    pub fn print_warnings(&self, ui: &UI) {
+        for warning in &self.warnings {
+            ui.print_warning(WarningMessage::new(warning));
+        }
+    }
+}
+
+/// An EIP-2645 HD path, required by the Starknet Ledger app. This type allows users to write
+/// hash-based segments in text, instead of manually finding out the lowest 31 bits of the hash.
+///
+/// The Ledger app requires that the path:
+/// - starts with `2645'`; and
+/// - has 6 levels
+///
+/// Supports shorthand `m//` to omit `2645'`:
+///   `m//starknet'/sncast'/0'/0'/0`
+#[derive(Debug, Clone)]
+struct Eip2645Path {
+    layer: Eip2645Level,
+    application: Eip2645Level,
+    eth_address_1: Eip2645Level,
+    eth_address_2: Eip2645Level,
+    index: Eip2645Level,
+}
+
+#[derive(Debug, Clone)]
+enum Eip2645Level {
+    Hash(HashLevel),
+    Raw(u32),
+}
+
+#[derive(Debug, Clone)]
+struct HashLevel {
+    text: String,
+    hardened: bool,
+}
+
+impl TypedValueParser for DerivationPathParser {
+    type Value = ParsedDerivationPath;
+
+    fn parse_ref(
+        &self,
+        cmd: &Command,
+        _arg: Option<&Arg>,
+        value: &std::ffi::OsStr,
+    ) -> Result<Self::Value, Error> {
+        if value.is_empty() {
+            return Err(cmd
+                .clone()
+                .error(ErrorKind::InvalidValue, "empty Ledger derivation path"));
+        }
+        match value.to_str() {
+            Some(value) => match Eip2645Path::from_str_with_warnings(value) {
+                Ok((path, warnings)) => Ok(ParsedDerivationPath {
+                    path: path.into(),
+                    warnings,
+                }),
+                Err(err) => Err(cmd.clone().error(
+                    ErrorKind::InvalidValue,
+                    format!("invalid Ledger derivation path: {err}"),
+                )),
+            },
+            None => Err(cmd.clone().error(
+                ErrorKind::InvalidValue,
+                "invalid Ledger derivation path: not UTF-8",
+            )),
+        }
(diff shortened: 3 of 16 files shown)


crates/sncast/src/helpers/ledger/hd_path.rs
MKowalski8, line 1: For reference: https://github.com/xJonathanLEI/starkli/blob/master/src/hd_path.rs

crates/sncast/src/helpers/ledger/hd_path.rs
```diff
@@ -0,0 +1,425 @@
...
+    }
+}
+
+impl Eip2645Path {
+    fn from_str_with_warnings(s: &str) -> Result<(Self, Vec<String>)> {
+        let path = s.parse::<Self>()?;
+        let mut warnings = Vec::new();
+
+        // These are allowed but we should serve a warning.
+        match &path.eth_address_1 {
```
MKowalski8, line 108: Since those are not required, we could skip them, or somehow limit them to display only with account creation. Now it would display every time when someone use "not standard" path

crates/sncast/src/helpers/ledger/emulator_transport.rs
```diff
@@ -0,0 +1,92 @@
+/// This module is only used behind the `ledger-emulator` feature flag.
+/// It is used to test the ledger commands locally without having to connect to a real Ledger device.
```
ddoktorski: nit:
```suggestion
//! This module is only used behind the `ledger-emulator` feature flag.
//! It is used to test the ledger commands locally without having to connect to a real Ledger device.
```

crates/sncast/src/starknet_commands/ledger.rs
```diff
@@ -0,0 +1,162 @@
...
+    no_display: bool,
+}
+
+#[derive(Args, Debug)]
+struct SignHash {
+    /// Ledger derivation path in EIP-2645 format
+    #[arg(long, value_parser = DerivationPathParser)]
+    path: DerivationPath,
+
+    /// The raw hash to be signed (hex string with or without 0x prefix)
```
franciszekjob: Could hash be passed as decimal string too?

crates/sncast/src/starknet_commands/ledger.rs
```diff
@@ -0,0 +1,162 @@
...
+    ledger: LedgerStarknetApp<T>,
+    ui: &UI,
+) -> Result<LedgerResponse> {
+    let hash = Felt::from_hex(&args.hash)
+        .with_context(|| format!("Failed to parse hash: {}", args.hash))?;
+
+    ui.print_warning(WarningMessage::new(
+        "Blind signing a raw hash could be dangerous. Make sure you ONLY sign hashes \
+        from trusted sources. If you're sending transactions, use Ledger as a signer instead \
+        of using this command.\n",
```
franciszekjob: nit: When signing hash we also "use Ledger as a signer". I think this below may be a bit clearer, wdyt?

```suggestion
        from trusted sources. For better security, sign full transactions instead \
        of raw hashes whenever possible.\n",
```

crates/sncast/src/starknet_commands/ledger.rs
franciszekjob, line 1: Let's split subcommands into separate files (similarly as we do for e.g. `sncast account`).

crates/sncast/src/starknet_commands/ledger.rs
```diff
@@ -0,0 +1,162 @@
...
+    SignHash(SignHash),
+    /// Get Starknet app version from Ledger device
+    AppVersion(AppVersion),
+}
+
+#[derive(Args, Debug)]
+struct GetPublicKey {
+    /// Ledger derivation path in EIP-2645 format
+    #[arg(long, value_parser = DerivationPathParser)]
+    path: DerivationPath,
```
franciszekjob: How about adding a possibility to specify account by id (similarly as it's done in starknet.js and starknet.py)?
Refs:
- js: https://starknetjs.com/docs/API/classes/LedgerSigner231/#parameters
- py: https://starknetpy.readthedocs.io/en/latest/api/signer.html#starknet_py.net.signer.ledger_signer.LedgerSigner

crates/sncast/src/helpers/ledger/emulator_transport.rs
ddoktorski, line 1: Given that these are test utils, would it be possible to have them in `test` rather than `src` directory?

crates/sncast/src/helpers/ledger/hd_path.rs
```diff
@@ -0,0 +1,427 @@
+/// Adapted from starkli: <https://github.com/xJonathanLEI/starkli/blob/1c1040ece107e8b371f0342b05c81e3fccf09086/src/hd_path.rs>
```
ddoktorski: Is it useful for us to keep this comment?

crates/sncast/Cargo.toml
```diff
@@ -64,6 +65,7 @@ tracing-log.workspace = true
 strum.workspace = true
 strum_macros.workspace = true
 primitive-types.workspace = true
+coins-ledger = "0.12.0"
```
ddoktorski: is it reliable?

crates/sncast/src/helpers/ledger/hd_path.rs
ddoktorski: It'd be good to have an option to disable it, but I think it's not crucial for now

crates/sncast/Cargo.toml
MKowalski8: [starknet-rust](https://github.com/software-mansion/starknet-rust/blob/d268efd10c11075076820b975183e71590ba794e/Cargo.toml#L61) uses it, so I hope so 

Also it has more than 1,6 million downloads on [crates](https://crates.io/crates/coins-ledger)

crates/sncast/src/helpers/ledger/hd_path.rs
MKowalski8: No it's not. I saw that in `bravos.rs` we also have this, so I kept the convention. Should I remove it?

crates/sncast/src/helpers/ledger/emulator_transport.rs
MKowalski8: I think it would end with file that is in `test` directory, but is linked in `src` via path anyway, because we need to get the `ledger_app` in `create_ledger_app()` and I don't know better way how to do it.

crates/sncast/src/starknet_commands/ledger.rs
MKowalski8: Does it answear your question?

Image

crates/sncast/src/starknet_commands/ledger.rs
MKowalski8: I see, so you propose that we could just do something like 

```
sncast ledger get-public-key --account-id 5 
``` 

That would correspond to: 
```
sncast ledger get-public-key --path "m//starknet'/sncast'/0'/5'/0" 
```


**I think it can be done**

crates/sncast/src/starknet_commands/ledger.rs
franciszekjob: @MKowalski8 I mean that ATM it's a hex string
> /// The raw hash to be signed (hex string with or without 0x prefix)

and I asked whether it can be a decimal string too.

MKowalski8: No, in this context, it can't be decimal. It will always be treated as `hex`

franciszekjob: Shouldn't we allow for decimal too? For instance,`--contract-address` in invoke accepts both decimal/hex:

image

crates/sncast/src/starknet_commands/ledger/sign_hash.rs
```diff
@@ -0,0 +1,47 @@
...
+
+use sncast::response::ledger::{LedgerResponse, SignatureResponse};
+
+#[derive(Args, Debug)]
+pub struct SignHash {
+    #[command(flatten)]
+    pub key_locator: LedgerKeyLocator,
+
+    /// The raw hash to be signed (hex string with or without 0x prefix)
+    pub hash: String,
```
ddoktorski:
```suggestion
    pub hash: Felt,
```
Can it have felt type?

crates/sncast/src/helpers/ledger/emulator_transport.rs
```diff
@@ -0,0 +1,94 @@
...
+    }
+}
+
+#[async_trait]
+impl LedgerAsync for SpeculosHttpTransport {
+    async fn init() -> Result<Self, LedgerError> {
+        let url = std::env::var("LEDGER_EMULATOR_URL")
+            .unwrap_or_else(|_| "http://127.0.0.1:5001".to_string());
+        Self::new(url)
+    }
```
ddoktorski, line 44: Where is it used?

crates/sncast/src/helpers/ledger/hd_path.rs
```diff
@@ -0,0 +1,454 @@
...
+
+/// Constructs the canonical sncast derivation path for a given account ID:
+/// `m//starknet'/sncast'/0'/<account_id>'/0`
+pub(super) fn account_id_to_derivation_path(account_id: u32) -> DerivationPath {
+    Eip2645Path::from_str(&format!("m//starknet'/sncast'/0'/{account_id}'/0"))
+        .expect("account_id_to_derivation_path: generated path must be valid")
+        .into()
+}
+
+pub fn parse_derivation_path(path: &str, ui: &UI) -> Result<DerivationPath> {
```
ddoktorski: Is it used only in tests?

crates/sncast/src/helpers/ledger/key_locator.rs
```diff
@@ -0,0 +1,38 @@
...
+    account_id: Option<u32>,
+) -> Option<DerivationPath> {
+    match (path, account_id) {
+        (Some(path), None) => Some(path.clone()),
+        (None, Some(id)) => Some(account_id_to_derivation_path(id)),
+        (None, None) => None,
+        (Some(_), Some(_)) => unreachable!(
+            "clap group allows at most one of --path/--ledger-path or --account-id/--ledger-account-id"
+        ),
+    }
```
ddoktorski: Why isn't this handled by clap?

crates/sncast/src/helpers/ledger/hd_path.rs
ddoktorski: If not useful I'd remove it

crates/sncast/src/starknet_commands/ledger/sign_hash.rs
```diff
@@ -0,0 +1,47 @@
...
+
+    ui.print_notification("Please confirm the signing operation on your Ledger\n");
+
+    let signature = ledger
+        .sign_hash(path, &hash)
+        .await
+        .context("Failed to sign hash with Ledger")?;
+
+    Ok(LedgerResponse::Signature(SignatureResponse {
+        signature: format!("0x{signature}"),
```
ddoktorski: Why signature is one string instead of two (r, s)? 🤔

crates/sncast/src/helpers/ledger/hd_path.rs
```diff
@@ -0,0 +1,454 @@
...
+
+        let segments: Vec<&str> = s.split('/').collect();
+        if segments.len() != EIP2645_LENGTH + 1 {
+            bail!("EIP-2645 paths must have {EIP2645_LENGTH} levels");
+        }
+        if segments[0] != "m" {
+            bail!("HD wallet paths must start with \"m/\"");
+        }
+
+        if !segments[1].is_empty() {
```
ddoktorski: Given the previous checks and replacement, isn't it effectively a dead check?

crates/sncast/src/helpers/ledger/emulator_transport.rs
MKowalski8: Trait `LedgerAsync` needs to have `init()`. I'll change `emulator_ledger_app` func to use it and not to duplicate the logic

crates/sncast/src/helpers/ledger/hd_path.rs
MKowalski8: After recent changes it's a leftover here. I'll change that

crates/sncast/src/starknet_commands/ledger/sign_hash.rs
MKowalski8: This is the way how `starkli` print's that. I can split it if you prefer this way.

crates/sncast/src/helpers/ledger/hd_path.rs
MKowalski8: Hmm... you're right, this `if` is not necessary

crates/sncast/src/helpers/ledger/key_locator.rs
```diff
@@ -0,0 +1,38 @@
...
+}
```
franciszekjob, line 41: `resolve_key_locator` func is not used anywhere else, why not inline?

crates/sncast/src/starknet_commands/ledger/sign_hash.rs
```diff
@@ -0,0 +1,47 @@
...
+    let hash = Felt::from_hex(&args.hash)
+        .with_context(|| format!("Failed to parse hash: {}", args.hash))?;
+
+    ui.print_warning(WarningMessage::new(
+        "Blind signing a raw hash could be dangerous. Make sure you ONLY sign hashes \
+        from trusted sources. For better security, sign full transactions instead \
+        of raw hashes whenever possible.\n",
+    ));
+
+    ui.print_notification("Please confirm the signing operation on your Ledger\n");
```
franciszekjob, line 36: nit: Maybe extra newline between these two so they don't look like a one block?

crates/sncast/src/helpers/ledger/key_locator.rs
MKowalski8: It will be used later. In second PR. I could add comment for that 😅

crates/sncast/src/helpers/ledger/key_locator.rs
MKowalski8: In some way it's handled by clap. Problem is that I need to parse `account_id` -> `DeriviationPath`. So I will always need some way of `resolve`. However I will change it a little bit, to be more clearer

crates/sncast/src/starknet_commands/ledger/sign_hash.rs
franciszekjob: Let's split it

franciszekjob (approved):
Please address changes from Darek, otherwise LGTM

crates/sncast/src/starknet_commands/ledger.rs
MKowalski8: Okay, so this was a case of aFelt type instead of string, now it works with decimal too.

crates/sncast/src/helpers/ledger/hd_path.rs
```diff
@@ -0,0 +1,459 @@
...
+                    https://foundry-rs.github.io/starknet-foundry/starknet/eip-2645-hd-paths.html"
+                        .to_string(),
+                );
+            }
+            Eip2645Level::Raw(raw) => {
+                if raw & 0x7fff_ffff != 0 {
+                    warnings.push(
+                        "using any value other than `0'` for \"eth_address_1\" might \
+                        make automatic key discovery difficult or impossible. Learn more at \
+                        https://foundry-rs.github.io/starknet-foundry/starknet/eip-2645-hd-paths.html"
```
ddoktorski: nit: Since we control all these warnings through `print_warnings` function we can display the link to the docs only once at the end of the loop instead of appending to each warning

crates/sncast/src/helpers/ledger/hd_path.rs
```diff
@@ -0,0 +1,459 @@
...
+            (s, false)
+        };
+
+        if body.chars().all(|char| char.is_ascii_digit()) {
+            let raw_node = body
+                .parse::<u32>()
+                .map_err(|err| anyhow!("invalid path level \"{body}\": {err}"))?;
+
+            if harden_notation {
+                if raw_node & 0x8000_0000 > 0 {
```
ddoktorski: nit: What about introducing a const like:
```rust
const HARDENED_BIT: u32 = 1 << 31;
```
the same for `0x7fff_ffff`

crates/sncast/src/helpers/ledger/hd_path.rs
```diff
@@ -0,0 +1,459 @@
...
+                let hash = hasher.finalize();
+
+                // Safe to unwrap: SHA256 output is fixed size
+                let node = u32::from_be_bytes(hash.as_slice()[28..].try_into().unwrap());
+
+                if level.hardened {
+                    node | 0x8000_0000
+                } else {
+                    node & 0x7fff_ffff
+                }
```
ddoktorski, line 302: nit: Comment with a short explanation might be helpful

crates/sncast/src/helpers/ledger/hd_path.rs
```diff
@@ -0,0 +1,459 @@
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    fn hash_segment(s: &str) -> u32 {
+        let level = Eip2645Level::Hash(HashLevel {
+            text: s.to_owned(),
+            hardened: true,
+        });
+        u32::from(&level) & 0x7fff_ffff
```
ddoktorski: Why do we call it explicitly instead of setting `hardened: false` above?

ddoktorski (approved):

PR #4189, opened 2026-02-26
Add Ledger tests and CI job

MKowalski8:
---
**Stack**:
- #4190
- #4189 ⬅
- #4193
- #4187
---


crates/sncast/tests/data/ledger-app/nanox#strk#0.25.13.elf
ddoktorski, line 1: What does this file?

.github/workflows/ci.yml
```diff
@@ -363,6 +363,21 @@ jobs:
...
+
+    steps:
+      - uses: actions/checkout@v6
+      - uses: asdf-vm/actions/install@05e0d2ed97b598bfce82fd30daf324ae0c4570e6
+      - uses: ./.github/actions/setup-tools
+      - name: Run tests
+        run: |
+          cargo test --profile ci -p sncast --features ledger-emulator \
+            ledger:: \
+            -- --ignored
```
ddoktorski: nit:
```suggestion
          cargo test --profile ci -p sncast --features ledger-emulator ledger -- --ignored
```

crates/sncast/tests/e2e/ledger/mod.rs
```diff
@@ -0,0 +1,386 @@
...
+    speculos_url: &str,
+    path: &str,
+    salt: Felt,
+    account_type: AccountType,
+) -> Felt {
+    let provider = create_jsonrpc_client();
+    let ui = UI::default();
+    let parsed_path = parse_derivation_path(path, &ui).unwrap();
+
+    // SAFETY: Even with different url, the account will be deployed correctly in devnet.
```
ddoktorski: Could you explain this comment?

crates/sncast/tests/e2e/ledger/account.rs
```diff
@@ -0,0 +1,423 @@
...
+use tempfile::tempdir;
+use test_case::test_case;
+
+#[test_case("oz", "open_zeppelin", OZ_CLASS_HASH.into_hex_string(), 6001, &[automation::APPROVE_PUBLIC_KEY]; "oz_account_type")]
+#[test_case("ready", "ready", READY_CLASS_HASH.into_hex_string(), 6004, &[automation::APPROVE_PUBLIC_KEY]; "ready_account_type")]
+// Braavos calls sign_hash twice during fee estimation (tx_hash + aux_hash) because
+// is_signer_interactive() always returns false — see BraavosAccountFactory::is_signer_interactive.
+// That means we need ENABLE_BLIND_SIGN + two APPROVE_BLIND_SIGN_HASH after the public key approval.
+#[test_case(
+    "braavos", "braavos", BRAAVOS_CLASS_HASH.into_hex_string(), 6007,
```
ddoktorski: Is it fine that we use the same port `6007` here and in the `test_import_ledger_account` test?

crates/sncast/tests/e2e/ledger/mod.rs
MKowalski8: Since we use the same devnet for every test, if a race condition occurs and I deploy an account using a different ledger instance (with differnt URL), the account will still be deployed to the same devnet and remain accessible to the original test.

Simply ledger instance has no effect on account deploy, of course with the emulator.

crates/sncast/tests/data/ledger-app/nanox#strk#0.25.13.elf
MKowalski8: It is our starknet ledger emulator app. We run speculos ledger emulator using this.

MKowalski8: I created it myself. The name follows the accepted convention, but I don't have the link right now, if you're interested, I can find it

crates/sncast/tests/e2e/ledger/account.rs
MKowalski8: It shouldn't affect anything, but it's better to have separate ports. I'll change that

crates/sncast/tests/data/ledger-app/nanox#strk#0.25.13.elf
franciszekjob: Would it be possible to build this app in workflow, instead of storing it? We do that in py
https://github.com/software-mansion/starknet.py/blob/6666719f912cce97f3373c428b71aa00e24d6248/.github/workflows/checks.yml#L187

crates/sncast/tests/e2e/ledger/account.rs
```diff
@@ -0,0 +1,423 @@
...
+    let accounts_content = std::fs::read_to_string(&accounts_file).unwrap();
+    let contents_json: serde_json::Value = serde_json::from_str(&accounts_content).unwrap();
+    let address = contents_json["alpha-sepolia"][LEDGER_ACCOUNT_NAME]["address"]
+        .as_str()
+        .unwrap()
+        .to_string();
+
+    mint_token(&address, u128::MAX).await;
+
+    let args = apply_test_resource_bounds_flags(vec![
```
franciszekjob: nit: Seems that `apply_test_resource_bounds_flags` are not needed anymore and we don't need to use them (at some point in past, devnet had an issue with fee estimation, that's why we added them).  I will remove their entire usage in https://github.com/foundry-rs/starknet-foundry/pull/4212

crates/sncast/tests/e2e/ledger/mod.rs
```diff
@@ -0,0 +1,392 @@
...
+        .is_ok();
+    mint_token(&format!("{address:#066x}"), u128::MAX).await;
+    if !is_deployed {
+        deployment
+            .l1_gas(100_000)
+            .l1_gas_price(10_000_000_000_000)
+            .l2_gas(1_000_000)
+            .l2_gas_price(10_000_000_000_000)
+            .l1_data_gas(100_000)
+            .l1_data_gas_price(10_000_000_000_000)
```
franciszekjob: Do we need this?

crates/sncast/tests/e2e/ledger/network.rs
```diff
@@ -0,0 +1,468 @@
...
+        .clone();
+
+    let tx_hash = get_transaction_hash(&output);
+    let receipt = get_transaction_receipt(tx_hash).await;
+    assert!(matches!(receipt, Invoke(_)));
+}
+
+#[tokio::test]
+#[ignore = "requires Speculos installation"]
+async fn test_ledger_multiple_invokes() {
```
franciszekjob: Is there a specific reason to test multiple invokes if we already test one?

crates/sncast/tests/e2e/ledger/basic.rs
franciszekjob, line 1: Do we need a test for ledger command when emulator is not running, i.e. when Starknet app is not running on the device?

crates/sncast/tests/data/ledger-app/nanox#strk#0.25.13.elf
MKowalski8: It'll take longer, but yeah, I probably can

franciszekjob: We can address this later, let's create issue for it.

MKowalski8: I mean, it'll take longer to run. The change itself should be fast to do 😅. At least I'll check that

crates/sncast/tests/e2e/ledger/basic.rs
MKowalski8: Personally, I wouldn't add this. The connection is not handled by us specifically; the `starknet-rust` and `coin-ledger` libraries handle it. But if you think it can be useful, I can add it - though it would have to run with "standard" cast tests, without the `ledger-emulator` flag.

crates/sncast/tests/e2e/ledger/network.rs
MKowalski8: Tbh I can delete it. There is also test with `multicall` that do two `invokes`, so it's redundant.

crates/sncast/tests/data/ledger-app/nanox#strk#0.25.13.elf
franciszekjob: > I mean, it'll take longer to run.

It could take longer, but:

- We can cache the file
- It will be easier to switch to different version (simple commit change instead of need to rebuild the app)

Thoughts @MKowalski8 ?

crates/sncast/tests/e2e/ledger/basic.rs
franciszekjob: I think that a simple test for calling `sncast ledger` with starknet app not running is reasonable - such situation may surely occur and we want to know what will be displayed to user.

crates/sncast/tests/data/ledger-app/nanox#strk#0.25.13.elf
MKowalski8: I agree, it will definitely be easier to maintain, especially by other people.

crates/sncast/tests/e2e/ledger/mod.rs
MKowalski8: Fixed

crates/sncast/tests/data/ledger-app/nanox#strk#0.25.13.elf
MKowalski8: For now, made an issue: https://github.com/foundry-rs/starknet-foundry/issues/4221

crates/sncast/tests/e2e/ledger/basic.rs
MKowalski8: I encountered an issue. When trying to add a test for the Ledger connection, CI returns an HID error because, from what I've read, it lacks support for it. As a result, it throws a different error than what we see locally.

We could wrap [ledger app creation](https://github.com/foundry-rs/starknet-foundry/blob/65f74da899b09e278ca27ffda261963c60facd36/crates/sncast/src/helpers/ledger/mod.rs#L22) in sth like this:
```rust
// Spawned to catch panics from HID API initialization (e.g. in CI without HID support).
// map_err only maps JoinError (panic); inner errors from LedgerStarknetApp::new propagate as-is.
let app = tokio::spawn(LedgerStarknetApp::new())
    .await
    .map_err(|_| anyhow::anyhow!("Ledger device not found"))??;
```

But I do not if it is worth doing only for this test.

franciszekjob: Ok I see. Testing this behavior manually will be enough then.

franciszekjob (approved):

.github/workflows/ci.yml
ddoktorski, line 1: What about scheduled tests?

crates/sncast/tests/e2e/ledger/basic.rs
ddoktorski, line 1: nit: `simple.rs` -> `basic.rs`

crates/sncast/tests/e2e/ledger/network.rs
```diff
@@ -0,0 +1,409 @@
...
+
+    client
+        .automation(&[
+            automation::ENABLE_BLIND_SIGN,
+            automation::APPROVE_BLIND_SIGN_HASH,
+        ])
+        .await
+        .unwrap();
+
+    let account_address = deploy_ledger_account(&url, TEST_LEDGER_PATH, Felt::from(5003_u32)).await;
```
ddoktorski:
```suggestion
    let account_address = deploy_ledger_account(&url, TEST_LEDGER_PATH, Felt::from(5004_u32)).await;
```
nit: Was it supposed to be 5004?

crates/sncast/tests/e2e/ledger/mod.rs
ddoktorski: Please rephrase the comment in the code (it can be done in the next PR with docs)

ddoktorski (approved):

.github/workflows/ci.yml
MKowalski8: Will do in next PR

PR #4190, opened 2026-02-26
Ledger documentation

MKowalski8:
Closes: #2318 
---
**Stack**:
- #4190 ⬅
- #4189
- #4193
- #4187
---
⚠️ *Part of a stack created by [spr](https://github.com/ejoffe/spr). Do not merge manually using the UI - doing so may have unexpected results.*

Issue #3994, opened 2025-12-05
New account interface in cast


cptartur:
### Current State

Currently, accounts in cast support our account format and additionally keystores though they are treated as second-class citizen interface wise.

### Objective

Rework account interface and support in sncast so it supports different account types better

### Additional Context

We should consider supporting natively: our account format, keystores, ledger.

Additionally, we could consider moving `--account` flag to relevant commands only + changing the interface for ledger (removing the extra argument for path).

PR #4193, opened 2026-02-26
Allow Ledger to be used as transaction signer

MKowalski8:
---
**Stack**:
- #4190
- #4189
- #4193 ⬅
- #4187
---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Cast
 
+#### Added
+
+- Support for Ledger hardware wallet as a transaction signer.
+
 #### Changed
 
 - `sncast balance` and `sncast tx-status` commands have been moved under `sncast get` subcommand (`sncast get balance`, `sncast get tx-status`). The old commands still work, but will be removed in the future.
--- a/crates/sncast/src/helpers/ledger/account.rs
+++ b/crates/sncast/src/helpers/ledger/account.rs
@@ -1,0 +1,69 @@
+use super::{SncastLedgerTransport, create_ledger_app};
+use crate::response::ui::UI;
+use anyhow::{Context, Result, bail};
+use starknet_rust::{
+    accounts::{ExecutionEncoding, SingleOwnerAccount},
+    core::types::{BlockId, BlockTag},
+    providers::jsonrpc::{HttpTransport, JsonRpcClient},
+    signers::{DerivationPath, LedgerSigner},
+};
+use starknet_types_core::felt::Felt;
+
+const LEDGER_SIGNER_ERROR: &str = "Failed to create Ledger signer. Ensure the derivation path is correct and the Ledger app is ready.";
+
+pub async fn create_ledger_signer(
+    ledger_path: &DerivationPath,
+    ui: &UI,
+) -> Result<LedgerSigner<SncastLedgerTransport>> {
+    let ledger_app = create_ledger_app().await?;
+
+    ui.print_notification(
+        "Ledger device will display a confirmation screen — approve it to continue...\n",
+    );
+
+    LedgerSigner::new_with_app(ledger_path.clone(), ledger_app).context(LEDGER_SIGNER_ERROR)
+}
+
+pub fn verify_ledger_public_key(ledger_public_key: Felt, stored_public_key: Felt) -> Result<()> {
+    if ledger_public_key != stored_public_key {
+        bail!(
+            "Public key mismatch!\n\
+            Ledger public key: {ledger_public_key:#x}\n\
+            Stored public key: {stored_public_key:#x}\n\
+            \n\
+            This account was created with a different Ledger derivation path or public key.\n\
+            Make sure you're using the same derivation path that was used during account creation."
+        );
+    }
+    Ok(())
+}
+
+pub async fn ledger_account<'a>(
+    ledger_path: &DerivationPath,
+    address: Felt,
+    chain_id: Felt,
+    encoding: ExecutionEncoding,
+    provider: &'a JsonRpcClient<HttpTransport>,
+    ui: &UI,
+) -> Result<SingleOwnerAccount<&'a JsonRpcClient<HttpTransport>, LedgerSigner<SncastLedgerTransport>>>
+{
+    let signer = create_ledger_signer(ledger_path, ui).await?;
+
+    let mut account = SingleOwnerAccount::new(provider, signer, address, chain_id, encoding);
+    account.set_block_id(BlockId::Tag(BlockTag::PreConfirmed));
+
+    Ok(account)
+}
+
+pub async fn get_ledger_public_key(ledger_path: &DerivationPath, ui: &UI) -> Result<Felt> {
+    let ledger_app = create_ledger_app().await?;
+
+    ui.print_notification("Please confirm the public key on your Ledger device...\n");
+
+    let public_key = ledger_app
+        .get_public_key(ledger_path.clone(), true)
+        .await
+        .context("Failed to get public key from Ledger")?;
+
+    Ok(public_key.scalar())
+}
--- a/crates/sncast/src/helpers/ledger/key_locator.rs
+++ b/crates/sncast/src/helpers/ledger/key_locator.rs
@@ -17,6 +17,18 @@ pub struct LedgerKeyLocator {
     pub account_id: Option<u32>,
 }
 
+#[derive(Args, Debug)]
+#[group(id = "ledger_key_locator_account", multiple = false)]
+pub struct LedgerKeyLocatorAccount {
+    /// Ledger derivation path in EIP-2645 format
+    #[arg(long = "ledger-path", value_parser = DerivationPathParser)]
+    pub path: Option<ParsedDerivationPath>,
+
+    /// Account index, expands to "m//starknet'/sncast'/0'/<account-id>'/0"
+    #[arg(long = "ledger-account-id")]
+    pub account_id: Option<u32>,
+}
+
 impl LedgerKeyLocator {
     #[must_use]
     pub fn resolve(&self, ui: &UI) -> DerivationPath {
@@ -24,6 +36,13 @@ impl LedgerKeyLocator {
         // that at least one of `--path` or `--account-id` is provided
         resolve_key_locator(self.path.as_ref(), self.account_id, ui)
             .expect("clap requires one of --path or --account-id")
+    }
+}
+
+impl LedgerKeyLocatorAccount {
+    #[must_use]
+    pub fn resolve(&self, ui: &UI) -> Option<DerivationPath> {
+        resolve_key_locator(self.path.as_ref(), self.account_id, ui)
     }
 }
 
--- a/crates/sncast/src/helpers/signer.rs
+++ b/crates/sncast/src/helpers/signer.rs
@@ -1,0 +1,107 @@
+use crate::helpers::ledger;
+use anyhow::{Result, bail};
+use camino::Utf8PathBuf;
+use serde::{Deserialize, Serialize};
+use starknet_rust::{
+    accounts::SingleOwnerAccount,
+    providers::jsonrpc::{HttpTransport, JsonRpcClient},
+    signers::{DerivationPath, LedgerSigner, LocalWallet},
+};
+use starknet_types_core::felt::Felt;
+
+/// Represents the type of signer stored in the accounts file
+// Uses `untagged` + `flatten` for backward compatibility with the existing accounts file format.
+// Downside: deserialization errors are less descriptive than with tagged variants,
+// and field name collisions across variants would silently misbehave.
+#[derive(Deserialize, Serialize, Clone, Debug)]
+#[serde(untagged)]
+pub enum SignerType {
+    Local { private_key: Felt },
+    Ledger { ledger_path: DerivationPath },
+}
+
+impl SignerType {
+    #[must_use]
+    pub fn private_key(&self) -> Option<Felt> {
+        match self {
+            SignerType::Local { private_key } => Some(*private_key),
+            SignerType::Ledger { .. } => None,
+        }
+    }
+
+    #[must_use]
+    pub fn ledger_path(&self) -> Option<&DerivationPath> {
+        match self {
+            SignerType::Ledger { ledger_path } => Some(ledger_path),
+            SignerType::Local { .. } => None,
+        }
+    }
+}
+
+#[derive(Debug)]
+/// Represents the `SingleOwnerAccount` variant with either `LocalWallet` or `LedgerSigner` as signer
+pub enum AccountVariant<'a> {
+    LocalWallet(SingleOwnerAccount<&'a JsonRpcClient<HttpTransport>, LocalWallet>),
+    Ledger(
+        SingleOwnerAccount<
+            &'a JsonRpcClient<HttpTransport>,
+            LedgerSigner<ledger::SncastLedgerTransport>,
+        >,
+    ),
+}
+
+impl AccountVariant<'_> {
+    #[must_use]
+    pub fn address(&self) -> Felt {
+        use starknet_rust::accounts::Account;
+        match self {
+            AccountVariant::LocalWallet(account) => account.address(),
+            AccountVariant::Ledger(account) => account.address(),
+        }
+    }
+
+    #[must_use]
+    pub fn chain_id(&self) -> Felt {
+        use starknet_rust::accounts::Account;
+        match self {
+            AccountVariant::LocalWallet(account) => account.chain_id(),
+            AccountVariant::Ledger(account) => account.chain_id(),
+        }
+    }
+}
+
+#[macro_export]
+macro_rules! with_account {
+    ($variant:expr, |$account:ident| $body:expr) => {
+        match $variant {
+            &$crate::AccountVariant::LocalWallet(ref $account) => $body,
+            &$crate::AccountVariant::Ledger(ref $account) => $body,
+        }
+    };
+}
+
+/// Represents the source of the signer for account operations
+#[derive(Debug, Clone, Default)]
+pub enum SignerSource {
+    /// Use a keystore file at the given path
+    Keystore(Utf8PathBuf),
+    /// Use a Ledger device with the given derivation path
+    Ledger(DerivationPath),
+    /// Use the accounts file (default)
+    #[default]
+    AccountsFile,
+}
+
+impl SignerSource {
+    pub fn new(keystore: Option<Utf8PathBuf>, signer_type: Option<&SignerType>) -> Result<Self> {
+        let ledger_path = signer_type.and_then(SignerType::ledger_path);
+        match (keystore, ledger_path) {
+            (Some(path), None) => Ok(SignerSource::Keystore(path)),
+            (None, Some(path)) => Ok(SignerSource::Ledger(path.clone())),
+            (None, None) => Ok(SignerSource::AccountsFile),
+            (Some(_), Some(_)) => {
+                bail!("keystore and ledger cannot be used together")
+            }
+        }
+    }
+}
--- a/crates/sncast/src/lib.rs
+++ b/crates/sncast/src/lib.rs
@@ -32,7 +32,7 @@ use starknet_rust::{
         ProviderError::StarknetError,
         jsonrpc::{HttpTransport, JsonRpcClient},
     },
-    signers::{LocalWallet, SigningKey},
+    signers::{DerivationPath, LocalWallet, SigningKey},
 };
 use starknet_types_core::felt::Felt;
 use std::collections::HashMap;
@@ -48,9 +48,11 @@ pub mod helpers;
 pub mod response;
 pub mod state;
 
+use crate::helpers::ledger;
 use crate::response::ui::UI;
 use conversions::byte_array::ByteArray;
 use foundry_ui::components::warning::WarningMessage;
+pub use helpers::signer::{AccountVariant, SignerSource, SignerType};
 
 pub type NestedMap<T> = HashMap<String, HashMap<String, T>>;
 
@@ -124,7 +126,6 @@ impl TryFrom<Felt> for Network {
 
 #[derive(Deserialize, Serialize, Clone, Debug)]
 pub struct AccountData {
-    pub private_key: Felt,
     pub public_key: Felt,
     pub address: Option<Felt>,
     pub salt: Option<Felt>,
@@ -134,6 +135,9 @@ pub struct AccountData {
 
     #[serde(default, rename(serialize = "type", deserialize = "type"))]
     pub account_type: Option<AccountType>,
+
+    #[serde(flatten)]
+    pub signer_type: SignerType,
 }
 
 #[derive(Clone, Copy)]
@@ -253,13 +257,13 @@ pub async fn get_nonce(
 }
 
 pub async fn get_account<'a>(
-    config: &CastConfig,
+    config: &'a CastConfig,
     provider: &'a JsonRpcClient<HttpTransport>,
     rpc_args: &RpcArgs,
-    keystore: Option<&Utf8PathBuf>,
     ui: &UI,
-) -> Result<SingleOwnerAccount<&'a JsonRpcClient<HttpTransport>, LocalWallet>> {
+) -> Result<AccountVariant<'a>> {
     let chain_id = get_chain_id(provider).await?;
+
     let network_name = chain_id_to_network_name(chain_id);
     let account = &config.account;
     let is_devnet_account = is_devnet_account(account);
@@ -283,19 +287,32 @@ pub async fn get_account<'a>(
                 To use an inbuilt devnet account, please rename your existing account or use an account with a different number."
             )));
             ui.print_blank_line();
-            return get_account_from_accounts_file(account, accounts_file, provider, keystore)
-                .await;
+            get_account_from_accounts_file(
+                account,
+                accounts_file,
+                provider,
+                config.keystore.as_ref(),
+                ui,
+            )
+            .await
         }
         (true, false) => {
             let url = rpc_args
                 .get_url(config)
                 .await
                 .context("Failed to get url")?;
-            return get_account_from_devnet(account, provider, &url).await;
+            let local_account = get_account_from_devnet(account, provider, &url).await?;
+            Ok(AccountVariant::LocalWallet(local_account))
         }
         _ => {
-            return get_account_from_accounts_file(account, accounts_file, provider, keystore)
-                .await;
+            get_account_from_accounts_file(
+                account,
+                accounts_file,
+                provider,
+                config.keystore.as_ref(),
+                ui,
+            )
+            .await
         }
     }
 }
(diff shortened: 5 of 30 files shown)


crates/sncast/src/lib.rs
```diff
@@ -48,12 +48,55 @@ pub mod helpers;
...
+use crate::helpers::ledger;
 use crate::response::ui::UI;
 use conversions::byte_array::ByteArray;
 use foundry_ui::components::warning::WarningMessage;
+use starknet_rust::signers::LedgerSigner;
 
 pub type NestedMap<T> = HashMap<String, HashMap<String, T>>;
 
+#[derive(Debug)]
+pub enum AccountVariant<'a> {
```
MKowalski8: It can be placed in better place while doing this #3994

crates/sncast/src/starknet_commands/account/mod.rs
```diff
@@ -48,29 +48,40 @@ pub enum Commands {
     List(List),
 }
 
+#[allow(clippy::too_many_arguments)]
 pub fn prepare_account_json(
-    private_key: &SigningKey,
+    private_key: Option<&SigningKey>,
+    public_key: Felt,
```
MKowalski8: The ledger account in ledger files is saved like this:

```
"ledger_sepolia": {
      "address": "0x54eee4b7b7c34b2b12b9affb27e54884ecd163bd6feddc5cb571c0ec8fd36ca",
      "class_hash": "0x5b4b537eaa2399e3aa99c4e2e0208ebd6c71bc1467938cd52c798c601e43564",
      "deployed": true,
      "ledger_path": "m//starknet'/sncast'/0'/0'/0'",
      "legacy": false,
      "public_key": "0x9051f863b0cf8337256a88a76a47513ccbcfed6418fdce5ed176fb01b297be",
      "salt": "0xa7f15005f6fc5579",
      "type": "open_zeppelin"
    }
```

There is new field `ledger_path` and `private_key` is no longer saved. 
We could also remove `public_key` from the file, since we can "ask" ledger for it without user interaction. However in order to limit changes, we can leave it in the file despite some redundancy.

Share your thoughts

crates/sncast/src/starknet_commands/account/create.rs
```diff
@@ -160,22 +186,72 @@ async fn generate_account(
...
+    chain_id: Felt,
+    ui: &UI,
 ) -> Result<(serde_json::Value, u128)> {
-    let chain_id = get_chain_id(provider).await?;
-    let private_key = SigningKey::from_random();
-    let signer = LocalWallet::from_signing_key(private_key.clone());
+    if let SignerSource::Ledger(ledger_path) = signer_source {
+        let public_key = ledger::get_ledger_public_key(ledger_path, true, ui).await?;
+        let signer = ledger::create_ledger_signer(ledger_path, ui).await?;
+
```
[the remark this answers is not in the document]
MKowalski8: It could be done, but it's cleaner this way and It doesn't generate extra user interaction with device

crates/sncast/src/starknet_commands/multicall/mod.rs
```diff
@@ -73,11 +73,17 @@ pub async fn multicall(
         starknet_commands::multicall::Commands::Run(run) => {
             let provider = run.rpc.get_provider(&config, ui).await?;
 
-            let account =
-                get_account(&config, &provider, &run.rpc, config.keystore.as_ref(), ui).await?;
-            let result =
-                starknet_commands::multicall::run::run(run.clone(), &account, wait_config, ui)
-                    .await;
+            let signer_source = SignerSource::from_options(config.keystore.clone(), None);
```
MKowalski8: I haven't added ledger support in config file. Do you think it should be added there?

crates/sncast/src/starknet_commands/account/import.rs
```diff
@@ -63,8 +64,13 @@ pub struct Import {
     /// If passed, the command will not trigger an interactive prompt to add an account as a default
     #[arg(long)]
     pub silent: bool,
+
+    /// Ledger derivation path (e.g., "m//starknet'/sncast'/0'/0'/0")
+    #[arg(long, group = "private_key_input")]
+    pub ledger_path: Option<String>,
```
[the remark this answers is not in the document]
MKowalski8: Those are exclusive

crates/sncast/src/lib.rs
```diff
@@ -84,6 +127,33 @@ impl Display for AccountType {
...
 
+/// Represents the source of the signer for account operations
+#[derive(Debug, Clone)]
+pub enum SignerSource {
+    /// Use a keystore file at the given path
+    Keystore(Utf8PathBuf),
+    /// Use a Ledger device with the given derivation path
+    Ledger(String),
+    /// Use the accounts file (default)
+    AccountsFile,
```
franciszekjob:
```suggestion
    /// Use the accounts file (default)
    #[default]
    AccountsFile,
```

crates/sncast/src/helpers/ledger/account.rs
```diff
@@ -0,0 +1,85 @@
...
+    providers::jsonrpc::{HttpTransport, JsonRpcClient},
+    signers::LedgerSigner,
+};
+use starknet_types_core::felt::Felt;
+
+const LEDGER_SIGNER_ERROR: &str = "Failed to create Ledger signer. Ensure the derivation path is correct and the Ledger app is ready.";
+
+pub async fn create_ledger_signer(
+    ledger_path: &str,
+    print_message: bool,
```
franciszekjob: nit:
```suggestion
    verbose: bool,
```

franciszekjob: Is this actually needed? Do we allow to hide messages from cli?

crates/sncast/src/helpers/ledger/account.rs
```diff
@@ -0,0 +1,85 @@
...
+
+    let mut account = SingleOwnerAccount::new(provider, signer, address, chain_id, encoding);
+    account.set_block_id(BlockId::Tag(BlockTag::PreConfirmed));
+
+    Ok(account)
+}
+
+pub async fn get_ledger_public_key(
+    ledger_path: &str,
+    display_on_device: bool,
```
franciszekjob: Same as in https://github.com/foundry-rs/starknet-foundry/pull/4193/changes#r2889309133

crates/sncast/src/starknet_commands/account/create.rs
```diff
@@ -59,15 +60,15 @@ pub struct Create {
...
 pub async fn create(
     account: &str,
     accounts_file: &Utf8PathBuf,
-    keystore: Option<&Utf8PathBuf>,
     provider: &JsonRpcClient<HttpTransport>,
     chain_id: Felt,
     create: &Create,
     config: &CastConfig,
+    signer_source: &SignerSource,
     ui: &UI,
```
franciszekjob, line 76: Considering that `SignerSource::AccountsFile` variant holds a path

```rust
pub enum SignerSource {
    ...
    /// Use the accounts file (default)
    AccountsFile(Utf8PathBuf),
}
```
it seems we should be able to remove `accounts_file` param. WDYT?

crates/sncast/src/starknet_commands/account/create.rs
```diff
@@ -160,22 +186,72 @@ async fn generate_account(
...
+    }
+}
 
-    let (address, fee_estimate) = match account_type {
+#[allow(clippy::too_many_arguments)]
+async fn finalize_account_generation<S>(
+    provider: &JsonRpcClient<HttpTransport>,
+    signer: S,
+    public_key: Felt,
+    private_key: Option<&SigningKey>,
```
franciszekjob: nit: Tbh `KeyPair` abstraction would fit nicely here.

crates/sncast/src/lib.rs
```diff
@@ -134,6 +205,8 @@ pub struct AccountData {
 
     #[serde(default, rename(serialize = "type", deserialize = "type"))]
     pub account_type: Option<AccountType>,
+    #[serde(default)]
+    pub ledger_path: Option<String>,
 }
```
franciszekjob, line 141: We can use an enum with `#[serde(flatten)]` here to enforce mutual exclusivity between `private_key` and `ledger_path`. e.g.

```rust
...
pub enum SignerMethod {
    Local {
        private_key: Felt,
    },
    Ledger {
        ledger_path: String,
    },
}

pub struct AccountData {
    ...

    #[serde(flatten)]
    pub signer_method: SignerMethod,
}
```

WDYT?

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 #### Added
 
 - `sncast ledger` command for simple Ledger operations.
+- Possibility to use Ledger hardware wallet as signer in transactions.
```
franciszekjob: I don't really think this should be split into separate entries in changelog.

crates/sncast/src/lib.rs
franciszekjob, line 1: Could we move introduced code into separate files (or some part at least)? `lib.rs` is already very big 😅

franciszekjob (commented):
Nice job 👍 ! Leaving partial review for now.

Nit: This PR includes both a refactor for `SignerSource` and new ledger-related logic. In the future, please consider splitting these into separate PRs - it makes it easier to review the structural changes first and then focus purely on the new features. Great work overall though!

crates/sncast/src/main.rs
```diff
@@ -101,6 +101,10 @@ struct Cli {
     #[arg(short, long)]
     keystore: Option<Utf8PathBuf>,
 
+    /// Ledger derivation path to use as signer (e.g., "m//starknet'/sncast'/0'/0'/0")
+    #[arg(long)]
+    ledger_path: Option<String>,
```
MKowalski8: On second thought, I can add `ledger_path` and `keystore` to the same group, then it would be exclusive on `clap` level (I will change it with review suggestions)

CHANGELOG.md
MKowalski8: Actually you are right. I'll think about something more general.

crates/sncast/src/starknet_commands/account/create.rs
```diff
@@ -367,3 +448,19 @@ fn generate_deploy_command_with_keystore(
...
+    config: &CastConfig,
+    account: &str,
+    ledger_path: &String,
+) -> String {
+    let accounts_flag = generate_accounts_file_flag(accounts_file);
+    let network_flag = generate_network_flag(rpc_args, config);
+
+    format!(
+        "\n\nAfter prefunding the account, run:\n\
+        sncast{accounts_flag} --ledger-path \"{ledger_path}\" account deploy {network_flag} --name {account}"
```
ddoktorski: Shouldn't the path be already saved at this point? Why is it required to provide it here?

crates/sncast/src/starknet_commands/account/mod.rs
ddoktorski: Both options fine

crates/sncast/src/lib.rs
ddoktorski: Would it be possible to get rid of this enum and instead rely on some common trait `Account`?

crates/sncast/src/starknet_commands/account/create.rs
```diff
@@ -160,22 +186,72 @@ async fn generate_account(
...
+            public_key,
+            Some(&private_key),
+            None,
+            salt,
+            class_hash,
+            account_type,
+            chain_id,
+        )
+        .await
+    }
```
ddoktorski, line 219: Can we move private key - ledger logic to `finalize_account_generation`, so we have only one function call here?

crates/sncast/src/starknet_commands/account/deploy.rs
```diff
@@ -165,21 +184,87 @@ async fn deploy_from_accounts_file(
...
+    )
+    .await?;
+
+    update_account_in_accounts_file(accounts_file, &name, chain_id)?;
+
+    Ok(result)
+}
+
+#[allow(clippy::too_many_arguments)]
+async fn deploy_from_ledger(
```
ddoktorski: Overlap between this function and `deploy_from_accounts_file` feels redundant. For example `get_account_data_from_accounts_file` is called twice and the only variation is the signer, we could likely consolidate these into a single function

crates/sncast/src/starknet_commands/multicall/mod.rs
ddoktorski: What config are you exactly referring to? Only the path?

MKowalski8: Yes, that we can't save `ledger_path` like `network`, `account` etc. 
But since it's connected to account it's maybe not needed anyway.

crates/sncast/src/helpers/ledger/account.rs
MKowalski8: It's something different. It's option that determines if `get-public-key` has to be confirmed on ledger or not.
In `ledger` command it's `no_display` argument.

crates/sncast/src/helpers/ledger/account.rs
MKowalski8: Yeah, I did this way, because it prints message `Connected to Ledger device`, but in some cases before this might be called `get_ledger_public_key` which prints `Please confirm the public key on your Ledger device...`.

And because of needed operations order it would appear as 

```
Please confirm the public key on you Ledger...
Please approve the operation on your Ledger device...
```

So in this one case I skip the second message

crates/sncast/src/lib.rs
MKowalski8: I thought about handling it with this task #3994, as I said here https://github.com/foundry-rs/starknet-foundry/pull/4193/changes/BASE..d088b8d9fc62cd7eaf98c79ae4719b44c3a7527e#r2873272468.

But I'll do something about it, at most it will change later

crates/sncast/src/starknet_commands/account/create.rs
MKowalski8: But `AccountsFile` do not hold path in `SignerSource`, the `Keystore` does. 

It might hold it, but there is one problem, when `Ledger` is a `SignerSource` it still need `accounts_file`, so there always be extra place to extract it no mater what.

crates/sncast/src/starknet_commands/account/create.rs
MKowalski8: Yeah it is, but AFAIK for example with `--network`  it is also shown even if it saved in `config` file. But I can simplify it, it's not necessary

crates/sncast/src/starknet_commands/account/create.rs
MKowalski8: Only thing I can move is getting `public_key` and use function from trait (not mine with extra message)

Since `LocallWallet` and `LedgerSigner` are not the same concreate type it's hard. They share the same trait, but I can't use `Box<dyn Signer>` for it, because it's not object-save and it won't compile.

In order to do that I would need to introduce new `enum`. Which I would prefer not to do only for that.

crates/sncast/src/lib.rs
MKowalski8: It's simillar case to described in here https://github.com/foundry-rs/starknet-foundry/pull/4193#discussion_r2896667627, the trait is not `object-save`, so it won't compile with `dyn`. The only difference is, that it repeats a few times, so it keeps code cleaner.

crates/sncast/src/starknet_commands/account/create.rs
MKowalski8: I agree, but let's think about that during #3994. Right now I don't want to add another layer of abstarction, because this will most likely be changed.

crates/sncast/src/starknet_commands/account/create.rs
```diff
@@ -160,22 +179,75 @@ async fn generate_account(
...
     account_type: AccountType,
+    signer_source: &SignerSource,
+    chain_id: Felt,
+    ui: &UI,
 ) -> Result<(serde_json::Value, u128)> {
-    let chain_id = get_chain_id(provider).await?;
-    let private_key = SigningKey::from_random();
-    let signer = LocalWallet::from_signing_key(private_key.clone());
+    if let SignerSource::Ledger(ledger_path) = signer_source {
+        let public_key = ledger::get_ledger_public_key(ledger_path, true, ui).await?;
```
ddoktorski: Do we need to receive a public key here? Would it be possible later to call `signer.get_public_key()` instead?

crates/sncast/src/starknet_commands/account/deploy.rs
```diff
@@ -270,6 +274,28 @@ async fn get_deployment_result(
...
+
+async fn check_if_already_deployed(
+    provider: &JsonRpcClient<HttpTransport>,
+    address: Felt,
+) -> Result<bool> {
+    Ok(provider
+        .get_class_hash_at(BlockId::Tag(PreConfirmed), address)
+        .await
+        .is_ok())
+}
```
ddoktorski: In case there is a timeout error or something similar we'd treat the account as not deployed

crates/sncast/src/helpers/signer.rs
```diff
@@ -0,0 +1,104 @@
...
+use starknet_rust::{
+    accounts::SingleOwnerAccount,
+    providers::jsonrpc::{HttpTransport, JsonRpcClient},
+    signers::{DerivationPath, LedgerSigner, LocalWallet},
+};
+use starknet_types_core::felt::Felt;
+
+/// Represents the type of signer stored in the accounts file
+#[derive(Deserialize, Serialize, Clone, Debug)]
+#[serde(untagged)]
```
ddoktorski, line 17: LLMs claim that it is extremely error prone to use both `untagged` and `flatten` on the same struct. Could you explore whether it can really cause any problems in our case?

crates/sncast/src/helpers/ledger/account.rs
```diff
@@ -0,0 +1,81 @@
...
+    chain_id: Felt,
+    encoding: ExecutionEncoding,
+    stored_public_key: Felt,
+    provider: &'a JsonRpcClient<HttpTransport>,
+    ui: &UI,
+) -> Result<SingleOwnerAccount<&'a JsonRpcClient<HttpTransport>, LedgerSigner<SncastLedgerTransport>>>
+{
+    let signer = create_ledger_signer(ledger_path, true, ui).await?;
+
+    let ledger_public_key = signer.get_public_key().await?.scalar();
```
ddoktorski: Does it make sense to have an option to omit this if the public key is saved?

crates/sncast/src/helpers/ledger/account.rs
```diff
@@ -0,0 +1,81 @@
...
+
+pub async fn create_ledger_signer(
+    ledger_path: &DerivationPath,
+    verbose: bool,
+    ui: &UI,
+) -> Result<LedgerSigner<SncastLedgerTransport>> {
+    let ledger_app = create_ledger_app().await?;
+
+    if verbose {
+        ui.print_notification("Please approve the operation on your Ledger device...\n");
```
ddoktorski: Is this message accurate? I mean at the moment it is displayed there might be not any transaction to confirm on the device yet

crates/sncast/src/helpers/ledger/account.rs
ddoktorski: Actually I think `print_message` is much clearer with the current usage - this bool doesn't control verbosity, it only controls whether to print a single message or not

crates/sncast/src/starknet_commands/account/create.rs
MKowalski8: We could do that. The reason I included it here is that I wanted to get a specific message to the user about the `public_key`, not just a general operation message, since in this situation, ledger asks for confirmation of the public key. But that's not a big deal; if we want to keep it simple, I can change it.

crates/sncast/src/starknet_commands/account/create.rs
franciszekjob: Ok, we can leave it then.

crates/sncast/src/helpers/ledger/account.rs
MKowalski8: Actually it is used only for verification when the public key is saved, since we stopped allowing `--ledger-path` to be specified in other ways, there should be no differnece between the keys. 

At least if someone uses the same ledger device and haven't changed `accounts` file manually. 

Maybe it's better to remove it completely?

crates/sncast/src/helpers/ledger/account.rs
MKowalski8: There can be one situation with declare. When the message might be printed faster than contract compilation. 

Something like this would be better? 

```
"Ledger device will prompt for approval during this operation..."
```

crates/sncast/src/starknet_commands/account/deploy.rs
MKowalski8: It is how `keystore` deploy has always handled it.
With "normal" deploy this in not handled at all. 

Image

Remove the check from `deploy_from_keystore` then? 
 Btw this is also a leftover, because it is used only in `deploy_from_keystore`

crates/sncast/src/lib.rs
```diff
@@ -366,6 +395,41 @@ async fn build_account(
...
 
+async fn build_ledger_account<'a>(
+    ledger_path: DerivationPath,
+    account_data: AccountData,
+    chain_id: Felt,
+    provider: &'a JsonRpcClient<HttpTransport>,
+    ui: &UI,
+) -> Result<AccountVariant<'a>> {
+    let address = account_data.address.context(
+        "Account address not found. The account must be deployed before using it with Ledger.",
```
franciszekjob: Is this error message correct? Why does absence of `address` in `AccountData` indicate that the account is not deployed?

crates/sncast/src/helpers/ledger/key_locator.rs
```diff
@@ -16,13 +16,32 @@ pub struct LedgerKeyLocator {
     pub account_id: Option<u32>,
 }
 
+#[derive(Args, Debug)]
+#[group(id = "ledger_key_locator_account", multiple = false)]
+pub struct LedgerKeyLocatorAccount {
```
franciszekjob, line 22: nit: Wouldn't `LedgerPathArgs`/`LedgerLocationArgs`  be simpler?

crates/sncast/src/starknet_commands/account/create.rs
ddoktorski: Is there a risk of something like a race condition by sending two requests one after another? If this public key check isn't strictly necessary, I'd lean toward removing it. Otherwise, we could implement it as an optional user setting if it makes sense

crates/sncast/src/starknet_commands/account/deploy.rs
ddoktorski: Could you point to the exact code that you're referring to?

crates/sncast/src/helpers/ledger/account.rs
ddoktorski: Why isn't it the case for other transactions?

crates/sncast/src/helpers/ledger/account.rs
ddoktorski: sgtm

crates/sncast/src/starknet_commands/account/deploy.rs
MKowalski8: This is the [change](https://github.com/foundry-rs/starknet-foundry/pull/4193/changes#diff-e1993fd983a715078555cbacabbec698fecde83c0dcdd93cdde0d577ba00826bL130) in my PR

This is original code I am talking about (in `deploy_from_keystore` function) https://github.com/foundry-rs/starknet-foundry/blob/c510e20fccb327d5dfb2c26f0ee3ce3b3802c54c/crates/sncast/src/starknet_commands/account/deploy.rs#L130

crates/sncast/src/helpers/ledger/account.rs
MKowalski8: I mean, it shouldn't be noticeable in other transactions. 
Of course, with a very weak computer it is possible, but in normal use it is rather fast. 
The declare of a contract is an only transaction where I noticed little delay

crates/sncast/src/starknet_commands/account/create.rs
MKowalski8: There should be no risk of a race condition, at least I don't think so, but if you prefer to remove it, I am okay with that.

crates/sncast/src/lib.rs
MKowalski8: You are right, but then we have wrong error message here https://github.com/foundry-rs/starknet-foundry/blob/c510e20fccb327d5dfb2c26f0ee3ce3b3802c54c/crates/sncast/src/lib.rs#L353

crates/sncast/src/helpers/ledger/key_locator.rs
MKowalski8: Imo, that would be a bit too general for two arguments that are exclusive and have a pretty specific purpose, so I'll stick with the current form

crates/sncast/src/starknet_commands/account/deploy.rs
ddoktorski: I guess we should check the exact error indicating that contract is indeed deployed

crates/sncast/src/helpers/ledger/account.rs
ddoktorski: Thanks for explaining! I'd rather rephrase it, but current message is also fine, leaving this up to you

crates/sncast/src/helpers/ledger/account.rs
franciszekjob: You're right, my bad. @MKowalski8 could you get back to it please?

crates/sncast/src/lib.rs
franciszekjob: Ok, for consistency we can change both errors to just "Failed to get address".

franciszekjob (approved):

crates/sncast/src/helpers/signer.rs
MKowalski8: I see two possible problems:

First, if there were another field with the same name (which most likely won't happen).

Second, and more importantly, it remove some debugging information, and the error messages during deserialization aren't as detailed as those for split arguments.

crates/sncast/src/starknet_commands/account/deploy.rs
MKowalski8: Okay, I will try to add an error handling for that then, to display it better than it is now

crates/sncast/src/helpers/signer.rs
ddoktorski: Can we used tagged with explicit tagging e.g. `#[serde(rename = "local")]` to eliminate the second problem?

crates/sncast/src/starknet_commands/account/deploy.rs
```diff
@@ -165,56 +153,74 @@ async fn deploy_from_accounts_file(
     ui: &UI,
 ) -> Result<InvokeResponse> {
     let account_data = get_account_data_from_accounts_file(&name, chain_id, accounts_file)?;
+    let (account_type, class_hash, salt) = extract_deployment_fields(&account_data)?;
 
-    let private_key = SigningKey::from_secret_scalar(account_data.private_key);
+    let result = match account_data.signer_type {
+        SignerType::Ledger { ref ledger_path } => {
```
ddoktorski: nit:
```suggestion
    let result = match &account_data.signer_type {
        SignerType::Ledger { ledger_path } => {
```
then also `from_secret_scalar(*private_key)` below

crates/sncast/src/starknet_commands/account/deploy.rs
ddoktorski: Was this check for whether the account is deployed completely removed?

ddoktorski (approved):

crates/sncast/src/starknet_commands/account/deploy.rs
MKowalski8: Yeah I removed it, since it was in some redundant with the error we get later. And just returning transaction with `0x0` was not very informativeve anyway.

crates/sncast/src/helpers/signer.rs
MKowalski8: Unfortunately, without breaking the backward compatibility, it's not possible. And that is the reason why I did it this way.

ddoktorski: Ok let's leave it as it is. If that makes sense you can consider adding a comment that it may cause problems due to using flatten and untagged

PR #4205, opened 2026-03-09
Fix various sncast config issues

### LINKED ISSUE #3811: Error handling for `snfoundry.toml` parsing

### LINKED ISSUE #4027: Research and refactor sncast config logic

DelevoXDG:
Closes #3811
Closes #4027

## Stack
- #4175
- #4205 
- #4249 
- #4234 

## Summary

This PR fixes the actual bugs & issues I've been able to find, and changes the sncast handling logic in a more significant way  

## Introduced changes

- 

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/configuration/src/lib.rs
```diff
@@ -271,12 +263,9 @@ mod tests {
     fn empty_config_works() {
```
DelevoXDG, line 263: I'm not sure what is the intention of the original test, should we delete it?

Original PR:
https://github.com/foundry-rs/starknet-foundry/pull/1911/changes

crates/sncast/src/helpers/configuration.rs
```diff
@@ -202,35 +211,46 @@ impl Override for PartialCastConfig {
...
-        });
-
-        let global_config = load_config::<PartialCastConfig>(
-            Some(&global_config_path.clone()),
-            opts.profile.as_deref(),
-        )
-        .or_else(|_| load_config::<PartialCastConfig>(Some(&global_config_path), None))
-        .map_err(|err| anyhow::anyhow!(format!("Failed to load config: {err}")))?;
+impl PartialCastConfig {
+    fn load(path: &Utf8PathBuf, profile: Option<&str>, scope: &str) -> Result<Option<Self>> {
```
franciszekjob: I see that we don't rely on the `scope`'s value anywhere in the logic, but it still should be an enum.

crates/sncast/src/helpers/config.rs
```diff
@@ -4,20 +4,23 @@ use crate::helpers::constants::DEFAULT_ACCOUNTS_FILE;
...
-        if cfg!(target_os = "windows") {
-            dirs::config_dir()
-                .ok_or_else(|| anyhow::anyhow!("Could not determine config directory"))?
-                .join("starknet-foundry")
-        } else {
-            dirs::home_dir()
-                .ok_or_else(|| anyhow::anyhow!("Could not determine home directory"))?
-                .join(".config/starknet-foundry")
+pub fn get_or_create_global_config_path() -> Result<Utf8PathBuf> {
+    let global_config_dir = match env::var("SNFOUNDRY_CONFIG").ok() {
```
franciszekjob, line 12:
```suggestion
    let global_config_dir = match env::var("SNFOUNDRY_GLOBAL_CONFIG_DIR").ok() {
```

crates/sncast/tests/e2e/show_config.rs
franciszekjob, line 1: Could we add test which has wait params split between local and global config (e.g. `retry-interval` is set in global and `timeout` is set in local one)?

franciszekjob: Also, how about adding a test similar to `test_profile_missing_in_local_config`, but when profile exists in global config? This way, we test that once a local config file exists, requested profiles must exist locally, even if they exist globally.

crates/sncast/src/main.rs
```diff
@@ -756,15 +763,68 @@ fn get_cast_config(cli: &Cli, ui: &UI) -> Result<CastConfig> {
...
+        // If local config file exists, profile must be in it.
+        (Some(profile), MaybeConfig::NoProfile, _) => {
+            bail!(
+                "Profile [{profile}] not found in local config at {}",
+                local_path.unwrap_or_default()
+            );
+        }
+        // No local config file; profile must be in global config.
+        (Some(profile), MaybeConfig::NoFile, MaybeConfig::NoProfile) => {
+            // TODO: (#pending) Streamline approach wrt. `--profile` being re-used for foundry and `scarb`.
```
franciszekjob, line 810: Will this be attached to an actual issue?

DelevoXDG: Already resolved in #4249!

crates/sncast/src/helpers/config.rs
DelevoXDG: Scarb uses `SCARB_CONFIG` (also for a directory, not manifest file itself), so I thought it's fair to align the naming within the tools in the ecosystem.

`SNFOUNDRY_GLOBAL_CONFIG_DIR` seems very overcomplicated and verbose to me. We can consider making it `SNFOUNDRY_GLOBAL_CONFIG`, but I'm not even sure about this one 🤔 

WDYT?

crates/sncast/src/helpers/configuration.rs
DelevoXDG: I did the change [24224a4](https://github.com/foundry-rs/starknet-foundry/pull/4205/commits/24224a48950b32b2e272c36381e1010cf2f706b6).

However, I feel like entries like
```rs
    let local_profile = PartialCastConfig::load_maybe(local_path.as_ref(), profile, "local")?;
```
read very neatly, and having enum with dedicated `Display for ConfigScope` just to pass around error message context seems like a bit of an overkill to me. 

Thoughts?

crates/sncast/tests/e2e/show_config.rs
DelevoXDG: > Could we add test which has wait params split between local and global config (e.g. retry-interval is set in global and timeout is set in local one)?

There are already:
- override unit tests: https://github.com/foundry-rs/starknet-foundry/blob/24224a48950b32b2e272c36381e1010cf2f706b6/crates/sncast/src/lib.rs#L1010-L1010
- `test_invalid_effective_config`: https://github.com/foundry-rs/starknet-foundry/blob/24224a48950b32b2e272c36381e1010cf2f706b6/crates/sncast/tests/e2e/show_config.rs#L439-L439 that checks that checks combining wait params across configs that are invalid when combined

I kinda feel like very fairly covered 😅 Or do you have something else specific in mind?

DelevoXDG: > This way, we test that once a local config file exists, requested profiles must exist locally, even if they exist globally.

We're still awaiting final decision on the desired logic, I'd rather wait until then before making tests for it

crates/sncast/src/helpers/config.rs
```diff
@@ -4,20 +4,23 @@ use crate::helpers::constants::DEFAULT_ACCOUNTS_FILE;
...
+        Some(dir) => Utf8PathBuf::from(shellexpand::tilde(&dir).to_string()).into_std_path_buf(),
+        None => {
+            if cfg!(target_os = "windows") {
```
ddoktorski, line 15: Windows logic should be removed

crates/sncast/src/main.rs
```diff
@@ -765,15 +775,72 @@ fn get_cast_config(cli: &Cli, ui: &UI) -> Result<CastConfig> {
...
+    CastConfig::try_from(partial_config).with_context(|| {
+        indoc::formatdoc! {"
+            Unable to combine configs. Fix conflicts between config sources and try again.
+            Sources:
+            - CLI flags
+            - Local config: {local}
+            - Global config: {global}
+        ",
+            global = global_path.as_ref().map_or("missing", |p| p.as_str()),
+            local = local_path.as_ref().map_or("missing", |p| p.as_str()),
```
ddoktorski: nit:
```suggestion
            local = local_path.as_ref().map_or("missing", |p| p.as_str()),
            global = global_path.as_ref().map_or("missing", |p| p.as_str()),
```

crates/sncast/src/main.rs
```diff
@@ -765,15 +775,72 @@ fn get_cast_config(cli: &Cli, ui: &UI) -> Result<CastConfig> {
...
+        (Some(profile), MaybeConfig::NoProfile, _) => {
+            bail!(
+                "Profile [{profile}] not found in local config at {}",
+                local_path.unwrap_or_default()
+            );
+        }
+        // No local config file; profile must be in global config.
+        (Some(profile), MaybeConfig::NoFile, MaybeConfig::NoProfile) => {
+            // TODO: (#pending) Streamline approach wrt. `--profile` being re-used for foundry and `scarb`.
+            ui.print_warning(WarningMessage::new(format!(
```
ddoktorski, line 812: Why is it warning or not hard error?

crates/sncast/src/main.rs
```diff
@@ -765,15 +775,72 @@ fn get_cast_config(cli: &Cli, ui: &UI) -> Result<CastConfig> {
...
+                "Profile [{profile}] not found in global config at {}, and no local config found.",
+                global_path.clone().unwrap_or_default()
+            )));
+        }
+        // Note: this is potentially unreachable: `get_or_create_global_config_path` should always return dir with existing config file.
+        // TODO: (#3436) remove this if missing global config becomes an error
```
ddoktorski, line 818: I'm not sure if the linked issue is accurate here as it seems very general. Also what is the actual action item here, to decide whether missing global config file should be an error?

crates/sncast/src/helpers/config.rs
ddoktorski: I guess it'd be good if this env contained the word "GLOBAL"

crates/configuration/src/lib.rs
ddoktorski: I guess just a sanity check, leaving this up to you whether to keep it or not

crates/sncast/src/helpers/config.rs
franciszekjob: `SNFOUNDRY_GLOBAL_CONFIG` sounds fine.

crates/sncast/src/helpers/config.rs
DelevoXDG: It is, as a part of #4271

crates/sncast/src/main.rs
ddoktorski: I see now it is changed in the next PR, all good then!

crates/sncast/src/main.rs
DelevoXDG: > Also what is the actual action item here, to decide whether missing global config file should be an error?

Previously, function `get_global_config_path` actually didn't only get the config path, but also created it, if missing. That's why I renamed it.

The original logic though allowed us to continue, even if global config is missing for some reason.

The reason for my comment is that described situation ("global config is missing") shouldn't really happen, at least in normal (non-error) circumstances, since we always create the dir instead.

There are some cases where it can:
- Missing `HOME` dir on the os
- Some fs permission issues, disallowing us from creating config file
 
So the question is:
- Should we proceed execution in these cases, as was done before?
- Or should we treat these as hard errors?

---
1. If we treat them as hard errors, then all these `global_maybe_config == MaybeConfig::NoFile` cases become effectively unreachable.
2. If we keep it as it, we: 
   - treat them as errors when unable to resolve requested `--profile`
   - proceed execution with a warning otherwise

I think I attached the wrong issue though, it was planned to be handled under now closed https://github.com/foundry-rs/starknet-foundry/issues/4184

Since this is getting very confusing, I suggest we either create a new issue, or (even better imo) decide on the final approach now 😄

crates/sncast/tests/e2e/show_config.rs
DelevoXDG: Decided offline to allow global-only profile,  added `test_show_config_global_only_profile_with_local_config_present`

https://github.com/foundry-rs/starknet-foundry/pull/4205/changes/75c5c6f1f191b1e2e7d808c4feb644ecac39de7d

crates/sncast/src/main.rs
ddoktorski: As discussed offline I'd lean towards a hard error, but given that so far we proceeded in such cases it also makes perfect sense to only display a warning. Both options work for me, let's choose one of these and address it in scope of this or next PR

ddoktorski (approved):

crates/sncast/src/main.rs
```diff
@@ -765,15 +775,81 @@ fn get_cast_config(cli: &Cli, ui: &UI) -> Result<CastConfig> {
...
+    let profile = opts.profile.as_deref();
+
+    let global_default =
+        PartialCastConfig::load_maybe(global_path.as_ref(), None, ConfigScope::Global)?;
+    let global_profile =
+        PartialCastConfig::load_maybe(global_path.as_ref(), profile, ConfigScope::Global)?;
+    let local_default =
+        PartialCastConfig::load_maybe(local_path.as_ref(), None, ConfigScope::Local)?;
+    let local_profile =
+        PartialCastConfig::load_maybe(local_path.as_ref(), profile, ConfigScope::Local)?;
```
franciszekjob: nit: It seems to me that we can simplify this. When `--profile` is not passed, `profile == None`, so `global_profile` and `local_profile` end up calling `load_maybe(..., None, ...)` again. That means both global and local `snfoundry.toml` files are read/parsed twice in the no-profile case. Behavior is correct, but we could skip the second load when `profile.is_none()` to avoid duplicate work and simplify the flow:

```suggestion
    let global_default =
        PartialCastConfig::load_maybe(global_path.as_ref(), None, ConfigScope::Global)?;
    let global_profile = profile
        .map(|p| PartialCastConfig::load_maybe(global_path.as_ref(), Some(p), ConfigScope::Global))
        .transpose()?
        .unwrap_or(MaybeConfig::NoProfile);
    
    let local_default =
        PartialCastConfig::load_maybe(local_path.as_ref(), None, ConfigScope::Local)?;
    let local_profile = profile
        .map(|p| PartialCastConfig::load_maybe(local_path.as_ref(), Some(p), ConfigScope::Local))
        .transpose()?
        .unwrap_or(MaybeConfig::NoProfile);
```

franciszekjob (approved):

crates/sncast/src/main.rs
DelevoXDG: On the second thought, the proposed approach does seem logical and makes sense, I've just it rewritten to be a bit simpler.

PR #4209, opened 2026-03-12
Bump Scarb to `2.16.1`

ddoktorski:
- Bump Scarb version in `.tool-versions` to `2.16.1`
- Add snapshots for `2.16.1`
- Bump devnet to the latest version


.tool-versions
```diff
@@ -1,2 +1,2 @@
-scarb 2.16.0
+scarb 2.16.1
 starknet-devnet 0.7.1
```
MKowalski8: I don't know if this is the right PR for that, but we could bump `devnet` to `0.7.2` as well

MKowalski8 (approved):

DelevoXDG (approved):

franciszekjob (approved):

.tool-versions
ddoktorski: Done

PR #4212, opened 2026-03-13
Remove `apply_test_resource_bounds_flags`

franciszekjob:
Closes #

## Introduced changes

When bumping RPC some time ago, `apply_test_resource_bounds` was introduced. Because of issues with fee estimation by devnet, we used hard-coded resource bounds. It's not relevant anymore.

## Checklist

- [ ] Linked relevant issue
- [ ] Added relevant tests


crates/sncast/tests/e2e/deploy.rs
```diff
@@ -530,14 +518,9 @@ async fn test_happy_case_with_already_declared() {
         "deploy",
         "--url",
         URL,
-        "--constructor-calldata",
-        "0x1",
-        "0x1",
-        "0x0",
```
franciszekjob, line 536: note: [`Map`](https://github.com/foundry-rs/starknet-foundry/blob/14fab921c90a5cc9a51c3b6d5950bf3e5f1be3e2/crates/sncast/tests/data/contracts/map/src/lib.cairo#L9) contract does not have any constructor params, so passing them was incorrect. (similarly below)

crates/sncast/tests/e2e/deploy.rs
```diff
@@ -679,11 +656,9 @@ async fn test_deploy_with_declare_invalid_nonce() {
...
         output,
-        indoc! {
-            "
-            Command: deploy
-            Error: Invalid transaction nonce
-            "
-        },
+        indoc! {r#"
+        Command: deploy
+        Error: Transaction execution error = TransactionExecutionErrorData { transaction_index: 0, execution_error: Message("Account transaction nonce is invalid.") }
```
DelevoXDG, line 661: Should we update `From<StarknetError> for SNCastStarknetError` for more human-readable error message?

franciszekjob: Hmm, theoretically we can but I guess that there are other similar cases in which we could match the data of tx execution error and map it to a different one. Can apply this, but ideally in separate PR. Is that ok @DelevoXDG ?

ddoktorski (commented):
Do we have any tests that use fee arguments after this change?

franciszekjob:
> Do we have any tests that use fee arguments after this change?

Yes, e.g. [this one](https://github.com/foundry-rs/starknet-foundry/blob/20cb7b8d9cf00ae97bdcb95aabfdf0528c0935cb/crates/sncast/tests/e2e/declare.rs#L214). And similarly for deploy, invoke

ddoktorski (approved):

DelevoXDG (approved):

PR #4215, opened 2026-03-13
Move get commands under `sncast get`

DelevoXDG:
Towards #4144 

## Stack
- #4215 
- #4210

## Introduced changes

- moved `sncast balance`->`sncast get balance`
- moved `sncast tx-status`->`sncast get tx-status`
- add warnings to old commands

## Checklist

- [x] Linked relevant issue
- [x] Updated relevant documentation
- [x] Added relevant tests
- [x] Performed self-review of the code
- [x] Added changes to `CHANGELOG.md`

Issue #4144, opened 2026-02-03
Add support for all read RPC endpoints associated with txn and block


franciszekjob:
### Select Tool

sncast

### Requested Feature

Add support for:
- [ ] `starknet_getBlock`
- [ ] `starknet_getStateUpdate`
- [ ] `starknet_getTransactionReceipt`
- [x] `starknet_getClassHashAt` (#4220)
- [x] `starknet_getTransactionByHash` (#4217)
- [x] `starknet_getNonce` (#4210)
- [x] `selector` (starkli parity)(#4219)
- [ ] `spec_version`
- etc.

noa-starkware:
please add also starknet_getClassHashAt, starknet_getTransactionByHash, starknet_getNonce, selector, thanks!

PR #4217, opened 2026-03-17
Add `sncast get tx`

DelevoXDG:
Towards #4144

## Stack
- #4215 
- #4210
- #4218 
- #4217 
- #4219
- #4220
- #4223
- #4224

## Introduced changes

-

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,7 +25,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 #### Added
 
-- `sncast get nonce` command to get the nonce of a contract.
 - `--nonce` flag to `sncast multicall run`
 
 ## [0.58.0] - 2026-03-18
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Cast
+
+#### Added
+
+- `sncast get tx` command to get transaction details by hash.
+
 ## [0.58.1] - 2026-03-31
 
 ### Forge
@@ -19,6 +25,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 #### Added
 
+- `sncast get nonce` command to get the nonce of a contract.
 - `--nonce` flag to `sncast multicall run`
 
 ## [0.58.0] - 2026-03-18
--- a/crates/sncast/src/response/transaction.rs
+++ b/crates/sncast/src/response/transaction.rs
@@ -272,15 +272,15 @@ fn build_invoke_v3_response(tx: &starknet_rust::core::types::InvokeTransactionV3
         .tx_version("3")
         .tx_hash(transaction_hash)
         .sender_address(sender_address)
-        .calldata(calldata)
-        .signature(signature)
         .nonce(nonce)
+        .calldata(calldata)
+        .account_deployment_data(account_deployment_data)
         .resource_bounds(resource_bounds)
         .tip(*tip)
         .paymaster_data(paymaster_data)
         .nonce_da_mode(*nonce_data_availability_mode)
         .fee_da_mode(*fee_data_availability_mode)
-        .account_deployment_data(account_deployment_data)
+        .signature(signature)
         .build()
 }
 
@@ -374,13 +374,13 @@ fn build_declare_v3_response(tx: &starknet_rust::core::types::DeclareTransaction
         .nonce(nonce)
         .class_hash(class_hash)
         .compiled_class_hash(compiled_class_hash)
-        .signature(signature)
+        .account_deployment_data(account_deployment_data)
         .resource_bounds(resource_bounds)
         .tip(*tip)
         .paymaster_data(paymaster_data)
         .nonce_da_mode(*nonce_data_availability_mode)
         .fee_da_mode(*fee_data_availability_mode)
-        .account_deployment_data(account_deployment_data)
+        .signature(signature)
         .build()
 }
 
@@ -454,12 +454,12 @@ fn build_deploy_account_v3_response(
         .class_hash(class_hash)
         .contract_address_salt(contract_address_salt)
         .constructor_calldata(constructor_calldata)
-        .signature(signature)
         .resource_bounds(resource_bounds)
         .tip(*tip)
         .paymaster_data(paymaster_data)
         .nonce_da_mode(*nonce_data_availability_mode)
         .fee_da_mode(*fee_data_availability_mode)
+        .signature(signature)
         .build()
 }
 
@@ -478,8 +478,8 @@ fn build_l1_handler_response(tx: &starknet_rust::core::types::L1HandlerTransacti
         .tx_version(&version.to_string())
         .tx_hash(transaction_hash)
         .contract_address(contract_address)
-        .entry_point_selector(entry_point_selector)
         .nonce(&Felt::from(*nonce))
+        .entry_point_selector(entry_point_selector)
         .calldata(calldata)
         .build()
 }
--- a/crates/sncast/src/response/transaction.rs
+++ b/crates/sncast/src/response/transaction.rs
@@ -2,6 +2,7 @@ use crate::helpers::block_explorer::LinkProvider;
 use crate::response::cast_message::SncastCommandMessage;
 use crate::response::explorer_link::OutputLink;
 use conversions::padded_felt::PaddedFelt;
+use conversions::string::IntoDecStr;
 use foundry_ui::styling::OutputBuilder;
 use serde::{Serialize, Serializer};
 use starknet_rust::core::types::{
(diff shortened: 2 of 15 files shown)


DelevoXDG:
Some transactions are tricky to test, but you can check these locally:
#### Deploy
```
sncast_build get tx 0x1d08158d139345d562276f0a085d9764e618eba788bed99a238903595b17022 --network mainnet
```
#### L1 handler
```
sncast_build get tx 0x294f5246c1174c240e225c5eaf5d184f8868cb745e28da47dfb80512afbac1c --network mainnet
```

Since we're using starknet-rs logic, I'm not sure there's a point in setting up some workarounds e.g. mock responses to test every case for `sncast` 🤔

crates/sncast/src/response/transaction.rs
```diff
@@ -0,0 +1,492 @@
...
+        entry_point_selector,
+        calldata,
+    } = tx;
+    OutputBuilder::new()
+        .tx_header()
+        .tx_type("L1 HANDLER")
+        .tx_version(&version.to_string())
+        .tx_hash(transaction_hash)
+        .contract_address(contract_address)
+        .nonce(&Felt::from(*nonce))
```
franciszekjob, line 482: `nonce` in `L1HandlerTransaction` is of type `u64` (see [here](https://github.com/xJonathanLEI/starknet-rs/blob/933ac0d61b6d425edb0f8afbb4de44da541b4a26/starknet-core/src/types/codegen.rs#L1336)) - are you sure we want to format it as felt?

crates/sncast/src/response/transaction.rs
```diff
@@ -0,0 +1,492 @@
...
+
+    fn constructor_calldata(self, data: &[Felt]) -> Self {
+        self.felt_list_field("Constructor Calldata", data)
+    }
+
+    fn resource_bounds(self, rb: &ResourceBoundsMapping) -> Self {
+        self.field(
+            "Resource Bounds L1 Gas",
+            &format!(
+                "max_amount={:#x}, max_price_per_unit={:#x}",
```
franciszekjob: Any specific reason to display them in hex format? It seems to me that decimal is more appropriate (considering they're u64/u128).
Even on voyager they're displayed as decimal strings too 😅 

Image

crates/sncast/src/response/transaction.rs
```diff
@@ -0,0 +1,492 @@
...
+        max_fee,
+        signature,
+        contract_address,
+        entry_point_selector,
+        calldata,
+    } = tx;
+    OutputBuilder::new()
+        .tx_header()
+        .tx_type("INVOKE")
+        .tx_version("0")
```
franciszekjob, line 224: nit: In some places we use hardcoded value (like here), while in other we use `&version.to_string()`. Maybe we can unify this?

crates/sncast/src/response/transaction.rs
DelevoXDG: My idea was that nonce formatting should be uniform across all transaction responses, but perhaps we can print it as `u64`. 

Do you maybe know specific reason why it being `u64` is the case, beyond `L1Handler` being an older transaction? 🤔

crates/sncast/src/response/transaction.rs
DelevoXDG: My idea was to destructure the `starknet_rust::core::types::[..]Transaction` response, and use all the fields that it gives me.

In the future, that should allows us to quickly catch `starknet-rs` changes and update our output accordingly.  

`starknet_rust::core::types::InvokeTransactionV3` doesn't specify `version` in the response - but I can infer it from the struct name anyway.

`starknet_rust::core::types::L1HandlerTransaction` does specify `version` in the response - so since I'm using all fields, why not use this one as well? It should be always V0, so we can hardcode it, but the choice between hardcoding and using what RPC returns us - seems stylistic to me.

I prefer using all returned fields, since it seems a bit easier to reason about, and in the future wouldn't require understanding why the value was ignored only to be hardcoded later.

WDYT?

crates/sncast/src/response/transaction.rs
franciszekjob: > Do you maybe know specific reason why it being `u64` is the case, beyond `L1Handler` being an older transaction? 🤔

I assume that's becaise `Felt` is a native L2 type, whereas this specific nonce is generated directly on L1 by the Starknet Core Contract (hence `u64`).

crates/sncast/src/response/transaction.rs
franciszekjob: > `starknet_rust::core::types::InvokeTransactionV3` doesn't specify `version` in the response - but I can infer it from the struct name anyway.

Interesting, spec actually has it 🤔 
https://github.com/starkware-libs/starknet-specs/blob/3b56c139aa53f1fa36d34c977a0508ce6726e59e/api/starknet_api_openrpc.json#L2845

Will take a look, maybe it should be added in `starknet-rust` (?).

> I prefer using all returned fields, since it seems a bit easier to reason about, and in the future wouldn't require understanding why the value was ignored only to be hardcoded later.

Same opinion.

DelevoXDG: > Will take a look, maybe it should be added in starknet-rust (?).

___
Not sure, don't really have an opinion on it 😅 Either way seems ok to me.

crates/sncast/src/response/transaction.rs
DelevoXDG: > My idea was that nonce formatting should be uniform across all transaction responses, but perhaps we can print it as u64.

WDYT?

crates/sncast/src/response/transaction.rs
franciszekjob: Is this going to be changed?

crates/sncast/src/response/transaction.rs
franciszekjob: I think we should print it as `u64`, so according to the type.

franciszekjob (approved):

CHANGELOG.md
```diff
@@ -19,6 +25,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 #### Added
 
+- `sncast get nonce` command to get the nonce of a contract.
```
MKowalski8: This is doubled with `0.58.0` release

Image

crates/sncast/src/response/transaction.rs
```diff
@@ -0,0 +1,492 @@
...
+        .tx_hash(transaction_hash)
+        .contract_address(contract_address)
+        .entry_point_selector(entry_point_selector)
+        .calldata(calldata)
+        .max_fee(max_fee)
+        .signature(signature)
+        .build()
+}
+
+fn build_invoke_v1_response(tx: &starknet_rust::core::types::InvokeTransactionV1) -> String {
```
MKowalski8, line 234: nit: At first glance, most of those methods looks quite repetetive. Mayby there is a way to simplify it? The first thing that came to my mind is some macro. However if this or other solution would only make more noise, then leave it as it is.

MKowalski8 (approved):

crates/sncast/src/response/transaction.rs
DelevoXDG: @MKowalski8 do you perhaps want to contribute a third opinion? 😄

CHANGELOG.md
DelevoXDG: Fixed

crates/sncast/src/response/transaction.rs
DelevoXDG: I already put some thought into avoiding duplication, hence:
- Common helpers `padded_felt_field`, `felt_field`, `felt_list_field` - that we'll probably use in other RPC get endpoints too
- Shared logic for field types, so we get
   `.contract_address(contract_address)` instead of `.field("Contract Address",&self.contract_address.into_padded_hex_str())` logic that is currently prevalent in the codebase
   
I'm not sure there's much to trim: 
- we have to fill the builder with values, even if we simplify how we do it.
- and I wanted to destruct the exact struct so it's enforced that all fields are used - that should be useful e.g. with RPC version bumps:
    ```rs
    let starknet_rust::core::types::InvokeTransactionV1 {
        transaction_hash,
        sender_address,
        calldata,
        max_fee,
        signature,
        nonce,
    } = tx;
    ```

While the number of LOCs is big, we effectively do two lines of logic per method at this point.

However, if there's something specific you can think of, we can consider it, although I'm doubtful of whether it's worth the time effort 🤔

MKowalski8: >However, if there's something specific you can think of, we can consider it,

On second thought I see that those builders need to look like this. And I agree that it isn't worth of the time to change this.

crates/sncast/src/response/transaction.rs
MKowalski8: I agree, that if the type is `u64` we should print it as `u64`

DelevoXDG: changed

franciszekjob (approved):

crates/sncast/src/response/transaction.rs
DelevoXDG: discussed offline, decided to print nonce in decimal across the board:

that's how voyager does it and it should be more convenient to the user

PR #4218, opened 2026-03-17
Fix `sncast get` docs

DelevoXDG:
I thought I moved these sections (#4215), but apparently I accidentally just removed these 😬 

## Stack
- #4215 
- #4210
- #4218 
- #4217 
- #4219

## Introduced changes

- add `sncast get balance, tx-status` docs back 

## Checklist

- [ ] Linked relevant issue
- [x] Updated relevant documentation
- [ ] Added relevant tests
- [ ] Performed self-review of the code
- [ ] Added changes to `CHANGELOG.md`

PR #4219, opened 2026-03-17
Add `sncast utils selector`

DelevoXDG:
Towards #4144

## Stack
- #4215 
- #4210
- #4218 
- #4217 
- #4219
- #4220
- #4223
- #4224

## Introduced changes

- add `sncast utils selector`

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 #### Added
 
 - `sncast get tx` command to get transaction details by hash.
-- `sncast utils selector` command to calculate entrypoint selector (sn_keccak) from function name.
+- `sncast utils selector` command to calculate entrypoint selector (`sn_keccak`) from function name.
 
 ## [0.58.1] - 2026-03-31
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 #### Added
 
 - `sncast get tx` command to get transaction details by hash.
+- `sncast utils selector` command to calculate entrypoint selector (sn_keccak) from function name.
 
 ## [0.58.1] - 2026-03-31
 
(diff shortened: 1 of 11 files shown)


crates/sncast/src/starknet_commands/utils/selector.rs
franciszekjob, line 1: nit: Should we allow passing an empty string here 🤔? Since the command is named `selector` (rather than just `keccak`), it implies the input should be a valid entrypoint name.
Even in docs it's said:
> Calculate entrypoint selector from function name.

Image

However, I checked Cairo's behavior and `selector!("")` doesn't have a guard against this either. Given that, I'm fine with leaving it as is.

franciszekjob: Also, should something like this be allowed (`123_foo` is invalid entrypoint name)?

Image

DelevoXDG: I generally agree with you, but if other tools use `selector` as effectively user-friendly way to describe `keccak`, I'm inclined to keep it as is and not add additional validation 🤔 

WDYT?

franciszekjob: That's ok.

franciszekjob (approved):

CHANGELOG.md
```diff
@@ -11,30 +11,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
-### Cast
-
-#### Added
-
 - `sncast get nonce` command to get the nonce of a contract.
-- `--nonce` flag to `sncast multicall run`
-
-## [0.58.0] - 2026-03-18
+- `sncast get tx` command to get transaction details by hash.
+- `sncast utils selector` command to calculate entrypoint selector (sn_keccak) from function name.
```
MKowalski8: Just reminder, to not forget about moving it to `Unreleased` after branch update. 

Also wdyt about:

```suggestion
- `sncast utils selector` command to calculate entrypoint selector (`sn_keccak`) from function name.
```

MKowalski8 (approved):

CHANGELOG.md
DelevoXDG: Applied

PR #4220, opened 2026-03-17
Add `sncast get class-hash-at`

DelevoXDG:
Towards #4144

## Stack
- #4215 
- #4210
- #4218 
- #4217 
- #4219
- #4220
- #4223
- #4224

## Introduced changes

- Add `sncast get class-hash-at`

## Checklist

- [x] Linked relevant issue
- [x] Updated relevant documentation
- [x] Added relevant tests
- [x] Performed self-review of the code
- [x] Added changes to `CHANGELOG.md`

PR #4223, opened 2026-03-19
Align `sncast` err handling; Add alisases `sncast get transaction`, `sncast get transaction-status` 

DelevoXDG:
Towards #4144
Towards #4184

## Stack
- #4215 
- #4210
- #4218 
- #4217 
- #4219
- #4220
- #4223
- #4224

## Introduced changes

- Align `sncast` err handling
- Add alisases `sncast get transaction`, `sncast get transaction-status` 

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests

### DIFF
--- a/crates/sncast/src/starknet_commands/get/mod.rs
+++ b/crates/sncast/src/starknet_commands/get/mod.rs
@@ -18,9 +18,11 @@ pub struct Get {
 #[derive(Debug, Subcommand)]
 pub enum GetCommands {
     /// Get the status of a transaction
+    #[command(alias = "transaction-status")]
     TxStatus(tx_status::TxStatus),
 
     /// Get the transaction by hash
+    #[command(alias = "transaction")]
     Tx(transaction::Transaction),
 
     /// Fetch balance of the account for specified token
--- a/crates/sncast/src/starknet_commands/get/transaction.rs
+++ b/crates/sncast/src/starknet_commands/get/transaction.rs
@@ -1,4 +1,4 @@
-use anyhow::Context;
+use anyhow::Result;
 use clap::Args;
 use sncast::helpers::command::process_command_result;
 use sncast::helpers::configuration::CastConfig;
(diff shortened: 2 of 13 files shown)


crates/sncast/src/starknet_commands/get/mod.rs
```diff
@@ -18,9 +18,11 @@ pub struct Get {
 #[derive(Debug, Subcommand)]
 pub enum GetCommands {
     /// Get the status of a transaction
+    #[command(alias = "transaction-status")]
```
MKowalski8, line 21: Why do we need aliases for? Usually we do not use them if not needed

crates/sncast/src/starknet_commands/get/transaction.rs
```diff
@@ -45,15 +43,17 @@ async fn get_transaction(
     provider: &JsonRpcClient<HttpTransport>,
     transaction_hash: Felt,
     with_proof_facts: bool,
-) -> Result<TransactionResponse, StarknetCommandError> {
```
MKowalski8, line 48: Why do we delete it here (`StarknetCommandError`), but `get_tx_status` have this? Because if we want to align error handling, how this help?

crates/sncast/src/starknet_commands/get/mod.rs
franciszekjob: I can imagine that some users may like more the full name instead of tx. Imo we can leave it.

franciszekjob (approved):

crates/sncast/src/starknet_commands/get/mod.rs
DelevoXDG: This issue is part of `starkli parity` track, and in starkli full names were used.

Besides, these are very intuitive and are likely something that users will type (and get right) without checking `--help`, unlike our shorthands.

crates/sncast/src/starknet_commands/get/transaction.rs
DelevoXDG: Good catch, I've updated `get_tx_status` as well

crates/sncast/src/starknet_commands/get/mod.rs
MKowalski8: > Besides, these are very intuitive and are likely something that users will type (and get right) without checking --help, unlike our shorthands.

I mean. I agree with that, but usually we don't do that (add aliases for simplicity). And question is, mayby we should just call it like this instead of making an alias then, if it's more user friendly. 

But I know we had, the `tx-status` before, so it's not that simple and the argument about `starkli parity` is enough for me 😅

MKowalski8 (approved):

crates/sncast/src/starknet_commands/get/transaction.rs
DelevoXDG: Note: 

had to eventually revert all `get_` fns in similar context to `Result<[..], StarknetCommandError>` state:

the `anyhow::Result` approach broke when `get_tx_status` was used indirectly (in multicall context).

PR #4226, opened 2026-03-20
Optimize release builds for runtime performance 

maciektr:
- Enable thin LTO
- Set codegen-units to 1
- Use mimalloc 

On OZ:
Before:
```
Benchmark 1: /Users/maciektr/Projects/forks/starknet-foundry/target/release/snforge test --profile=release --fuzzer-seed=11067975336895813681 -w
  Time (mean ± σ):     45.516 s ±  0.186 s    [User: 324.214 s, System: 24.364 s]
  Range (min … max):   45.297 s … 45.776 s    5 runs
```

Thin LTO:
```
Benchmark 1: /Users/maciektr/Projects/forks/starknet-foundry/target/release/snforge test --profile=release --fuzzer-seed=11067975336895813681 -w
  Time (mean ± σ):     43.745 s ±  0.270 s    [User: 306.863 s, System: 24.961 s]
  Range (min … max):   43.436 s … 44.037 s    5 runs
```

Fat LTO:
```
Benchmark 1: /Users/maciektr/Projects/forks/starknet-foundry/target/release/snforge test --profile=release --fuzzer-seed=11067975336895813681 -w
  Time (mean ± σ):     43.460 s ±  0.414 s    [User: 304.455 s, System: 24.738 s]
  Range (min … max):   43.084 s … 44.157 s    5 runs
```

Thin LTO + MiMalloc:
```
Benchmark 1: /Users/maciektr/Projects/forks/starknet-foundry/target/release/snforge test --profile=release --fuzzer-seed=11067975336895813681 -w
  Time (mean ± σ):     38.304 s ±  0.392 s    [User: 271.223 s, System: 7.359 s]
  Range (min … max):   37.938 s … 38.965 s    5 runs
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3483,6 +3483,7 @@ dependencies = [
  "indoc",
  "insta",
  "itertools 0.14.0",
+ "mimalloc",
  "num-bigint",
  "openssl",
  "packages_validation",
@@ -4767,6 +4768,16 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b6d2cec3eae94f9f509c767b45932f1ada8350c4bdb85af2fcab4a3c14807981"
 
 [[package]]
+name = "libmimalloc-sys"
+version = "0.1.44"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "667f4fec20f29dfc6bc7357c582d91796c169ad7e2fce709468aefeb2c099870"
+dependencies = [
+ "cc",
+ "libc",
+]
+
+[[package]]
 name = "libredox"
 version = "0.1.14"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


ddoktorski (approved):
Thanks @maciektr! From what I see size of the binary decreased ~20%

Also credits to @zamazan4ik for the talk at Rustikon

zamazan4ik:
@maciektr thanks for the PR!

By the way, how build time and binaries size are changed between "FatLTO + CU1" vs "ThinLTO + CU1"? Since we don't see a performance difference between them, probably binaries size win is worth enabling more aggressive LTO mode in the prices of build time overhead (not pushing - just curious :)

franciszekjob (approved):

maciektr:
@ddoktorski seems that we can build it with fat lto, if you wish :)  https://github.com/maciektr/starknet-foundry/actions/runs/23436962026

ddoktorski:
Here are rough results from my local runs only for `snforge`. Since there's no performance gain with `fat` setting I guess we can stick with `thin` for now, but both options are fine for me.

| configuration | build time | binary size |
| :--- | :--- | :--- |
| default ([ba73649aa](https://github.com/foundry-rs/starknet-foundry/tree/ba73649aa9c07370767fc93affbf5989590e52ee)) | 2m 50s | 45M |
| ThinLTO + CU1 + mimalloc | 3m 55s | 36M |
| FatLTO + CU1 + mimalloc | 5m 31s | 32M |

zamazan4ik:
> Since there's no performance gain with fat setting I guess we can stick with thin for now, but both options are fine for me.

If both options are fine, I suggesting sticking to FatLTO due to smaller binaries (since other characteristics are equal). Overhead from FatLTO version anyway will/should affect the project only for Releases (which are at most twice per month at most, as far as I see in the Release history).

PR #4227, opened 2026-03-23
Skip publishing crates in nightly dry-run

maciektr:


.github/workflows/nightly.yml
ddoktorski, line 1: What about adding `if: ${{ !(inputs.dry_run) }}` to `publish-std` and `publish-plugin` jobs to skip them instead of adding an input parameter to these workflows?

maciektr: Done!

ddoktorski (approved):

MKowalski8 (approved):

Issue #2316, opened 2024-07-17
Add option to only estimate commands fee without executing them


ksew1:
### Which component is your feature related to?

Cast

### Feature Request

- Commands that require a fee for execution should have an option to estimate the fee without executing the command.
- From the CLI, this should be implemented as a boolean flag.
- Commands that need this option:
    * account deploy
    * deploy
    * invoke
    * declare
    * multicall run
  
- Please check if there are more commands that would benefit from this change.

Arcticae:
@cptartur suggested doing something like `--dry-run` flag with a redirection to an estimation endpoint

PR #4228, opened 2026-03-23
Dry run

franciszekjob:
Towards #2316

## Introduced changes

Introduce dry run functionality in `sncast`:
- `--dry-run` flag
- `--detailed` flag

## Checklist

- [x] Linked relevant issue
- [ ] Added relevant tests


franciszekjob:

- #4228 ◀
    - #4229
        - #4230

crates/sncast/src/helpers/fee.rs
```diff
@@ -41,6 +41,14 @@ pub struct FeeArgs {
...
     #[arg(long)]
     pub estimate_tip: bool,
+
+    /// If passed, the transaction will not be sent to the network and the fee will be estimated instead.
+    #[arg(long)]
+    pub dry_run: bool,
+
+    // If passed, the output will include detailed fee estimation results instead of just overall fee. Only works with `--dry-run` flag.
+    #[arg(long, requires = "dry_run")]
+    pub detailed: bool,
```
ddoktorski: Conceptually both flags should not be part of the `FeeArgs`. Please move them to a separate struct. I'd also reconsider whether we need `detailed` flag at all

crates/sncast/src/response/account/deploy.rs
```diff
@@ -1,43 +1,73 @@
...
 
 #[derive(Serialize, Deserialize, CairoSerialize, Clone, Debug, PartialEq)]
-pub struct AccountDeployResponse {
+pub enum AccountDeployResponse {
+    Transaction(InvokeTransactionResponse),
+    DryRun(DryRunResponse),
+}
+
+#[derive(Serialize, Deserialize, CairoSerialize, Clone, Debug, PartialEq)]
+pub struct AccountDeployTransactionResponse {
```
ddoktorski: Is it used somewhere?

crates/sncast/src/response/multicall/run.rs
```diff
@@ -35,9 +62,12 @@ impl OutputLink for MulticallRunResponse {
...
-        format!(
-            "transaction: {}",
-            provider.transaction(self.transaction_hash)
-        )
+        match self {
+            MulticallRunResponse::Transaction(response) => format!(
+                "transaction: {}",
+                provider.transaction(response.transaction_hash)
+            ),
+            MulticallRunResponse::DryRun(_) => "No links available for fee estimation".to_string(),
```
ddoktorski: Inconsistent with behavior for `AccountDeployResponse` and also other places

crates/sncast/src/starknet_commands/account/deploy.rs
```diff
@@ -196,7 +199,9 @@ async fn deploy_from_accounts_file(
         }
     };
 
-    update_account_in_accounts_file(accounts_file, &name, chain_id)?;
+    if !matches!(&result, InvokeResponse::DryRun(_)) {
```
ddoktorski: as above
```suggestion
    if let InvokeResponse::Transaction(_) = &result {
```

crates/sncast/src/starknet_commands/declare.rs
```diff
@@ -128,7 +131,8 @@ pub async fn declare_with_artifacts<S>(
     ui: &UI,
 ) -> Result<DeclareResponse, StarknetCommandError>
 where
-    S: Signer + Sync + Send,
+    S: Signer + Sync + Send + 'static,
+    S::SignError: 'static,
```
ddoktorski: Why is it needed now?

crates/sncast/src/response/explorer_link.rs
```diff
@@ -58,6 +58,7 @@ pub async fn block_explorer_link_if_allowed<T>(
     result: &anyhow::Result<T>,
     chain_id: Felt,
     config: &CastConfig,
+    dry_run: bool,
```
ddoktorski: Leaky abstraction, what about moving this logic to `OutputLink` so we can call the check below on the `response` variable instead of passing this arg?

crates/sncast/src/starknet_commands/account/deploy.rs
```diff
@@ -333,6 +338,18 @@ where
...
+    if fee_args.dry_run {
+        let fee_estimate = deployment
+            .estimate_fee()
+            .await
+            .map_err(|error| anyhow!("Failed to estimate fee: {error}"))?;
+        return Ok(InvokeResponse::DryRun(DryRunResponse::new(
+            &fee_estimate,
+            fee_args.detailed,
+        )));
+    }
```
ddoktorski: Can we add a new abstraction to replace all blocks like this with one function call?

ddoktorski (changes requested):

crates/sncast/src/helpers/fee.rs
```diff
@@ -41,6 +41,14 @@ pub struct FeeArgs {
     /// If passed, an estimated tip will be added to pay for the transaction.
     #[arg(long)]
     pub estimate_tip: bool,
+
+    /// If passed, the transaction will not be sent to the network and the fee will be estimated instead.
+    #[arg(long)]
+    pub dry_run: bool,
+
+    // If passed, the output will include detailed fee estimation results instead of just overall fee. Only works with `--dry-run` flag.
```
MKowalski8: I suppose:

```suggestion
    /// If passed, the output will include detailed fee estimation results instead of just overall fee. Only works with `--dry-run` flag.
```

crates/sncast/src/main.rs
```diff
@@ -471,10 +479,13 @@ async fn run_async_command(cli: Cli, config: CastConfig, ui: &UI) -> Result<()>
                 )
                 .expect("Failed to build contract");
 
+                let mut declare_fee_args = fee_args.clone();
+                declare_fee_args.dry_run = false;
```
MKowalski8: Should we be able to `declare` with `dry_run` set to `true`?
I mean it's not obvious for user how it works, and I don't know if we should perform operation like this without them knowing.

At least we should document it somehow

crates/sncast/src/starknet_commands/script/run.rs
```diff
@@ -215,20 +225,28 @@ impl<'a> ExtensionLogic for CastScriptExtension<'a> {
...
+                            wait: true,
+                            wait_params: self.config.wait_params,
+                            show_ui_outputs: true,
+                        },
+                        self.ui,
+                    ))
+                    .map(|response| match response {
+                        InvokeResponse::Transaction(tx) => tx,
+                        InvokeResponse::DryRun(_) => {
+                            unreachable!("Dry run is not supported in script context")
```
MKowalski8, line 253: Wouldn't `bail!` be better? If I am right user can reach this? (same above)

crates/sncast/src/response/dry_run.rs
```diff
@@ -0,0 +1,69 @@
...
+            l1_data_gas_price: fee_estimate.l1_data_gas_price,
+            overall_fee: fee_estimate.overall_fee,
+            detailed,
+        }
+    }
+}
+
+impl SncastCommandMessage for DryRunResponse {
+    fn text(&self) -> String {
+        let overall_fee_strk = BigDecimal::new(self.overall_fee.into(), 18.into());
```
MKowalski8: nit:
```suggestion
        let overall_fee_strk = BigDecimal::new(self.overall_fee.into(), 18);
```

crates/sncast/src/starknet_commands/account/deploy.rs
```diff
@@ -333,6 +338,18 @@ where
         l1_data_gas_price => AccountDeploymentV3::l1_data_gas_price,
         tip => AccountDeploymentV3::tip
     );
+
+    if fee_args.dry_run {
+        let fee_estimate = deployment
+            .estimate_fee()
+            .await
```
MKowalski8: Could we move it before `fee_settings`? When `max_fee` is set we'll do `estimate_fee()` twice

crates/sncast/src/helpers/fee.rs
franciszekjob: > Conceptually both flags should not be part of the FeeArgs. Please move them to a separate struct.

Moved to `DryRunArgs`

>  I'd also reconsider whether we need detailed flag at all

This was discussed in design doc and we agreed to have it, I'll leave it.

franciszekjob: >Conceptually both flags should not be part of the FeeArgs. Please move them to a separate struct.

Done

> I'd also reconsider whether we need detailed flag at all

I'd like to keep it. It gives users the option to see the full gas breakdown without making it the default noise in the output. We discussed it in design doc btw.

crates/sncast/src/response/account/deploy.rs
franciszekjob: Removed.

crates/sncast/src/response/multicall/run.rs
franciszekjob: Fixed - changed to `unreachable!()` to match the pattern used everywhere else.

crates/sncast/src/response/explorer_link.rs
franciszekjob: Done - added a default `fn is_dry_run(&self) -> bool { false }` to the OutputLink trait, implemented it on each response enum that has a `DryRun` variant, and removed the `dry_run` parameter from `block_explorer_link_if_allowed` entirely.

crates/sncast/src/starknet_commands/account/deploy.rs
franciszekjob: Yup - added `estimate_if_dry_run` helper on `DryRunArgs` and replaced all such blocks with it.

crates/sncast/src/main.rs
franciszekjob: It's changed now, so the comment is not relevant anymore. Resolving.

crates/sncast/src/starknet_commands/script/run.rs
franciszekjob: Not really. Even if we want to add dry run support in scripts, then it should be possible in Cairo code - not via CLI flags.

crates/sncast/src/starknet_commands/declare.rs
franciszekjob: After changing `context` to `map_err` it's not needed anymore.

crates/sncast/src/main.rs
```diff
@@ -475,10 +483,16 @@ async fn run_async_command(cli: Cli, config: CastConfig, ui: &UI) -> Result<()>
...
+
                 let declare_result = with_account!(&account, |account| {
                     declare(
                         contract_name,
-                        fee_args.clone(),
+                        declare_fee_args,
+                        DryRunArgs {
+                            dry_run: false,
+                            detailed: false,
+                        },
```
MKowalski8: I asked the question before, should we declare a contract, when doing deploy with `dry-run`? I think it's not obvious for the user, that this operation will be done (declare). And if we accept that, we should at least document it.

crates/sncast/src/response/dry_run.rs
```diff
@@ -0,0 +1,68 @@
...
+    fn text(&self) -> String {
+        let overall_fee_strk = BigDecimal::new(self.overall_fee.into(), 18);
+        let mut builder = styling::OutputBuilder::new()
+            .success_message("Dry run completed")
+            .blank_line()
+            .field(
+                "Overall Fee",
+                &format!(
+                    "{} Fri (~{} STRK)",
+                    &overall_fee_strk.to_string(),
```
MKowalski8: Shouldn't be `self.overall_fee` here? We use `overall_fee_strk` twice

crates/sncast/src/starknet_commands/account/mod.rs
MKowalski8, line 315: Should we check for `dry_run` with interactive prompt, now it will appear, right?

crates/sncast/src/starknet_commands/deploy.rs
```diff
@@ -59,6 +60,9 @@ pub struct Deploy {
     #[command(flatten)]
     pub fee_args: FeeArgs,
 
+    #[command(flatten)]
+    pub dry_run_args: DryRunArgs,
```
MKowalski8, line 64: Would it be okay to make it exclusive with `fee_args`? They are not taken into consideration. so there is no need allowing user to add them together

crates/sncast/src/main.rs
franciszekjob: Ah, right, it makes most sense if we prevent `--dry-run` flag to use used with `declare --contract-name`. I agree with you that it's not intuitive if contract is declared. Changed in https://github.com/foundry-rs/starknet-foundry/pull/4228/changes/f8521915988abf55470a22eaa7d42bb76dfef973

crates/sncast/src/starknet_commands/deploy.rs
franciszekjob: Yes, absolutely. https://github.com/foundry-rs/starknet-foundry/pull/4228/commits/f91b9a4a1f82402666783e5413679a3fc0d86693

crates/sncast/src/starknet_commands/account/mod.rs
franciszekjob: https://github.com/foundry-rs/starknet-foundry/pull/4228/commits/0842d34f434db900382f96ce3be56344ad7e95fb

crates/sncast/src/starknet_commands/script/run.rs
```diff
@@ -139,6 +142,10 @@ impl<'a> ExtensionLogic for CastScriptExtension<'a> {
                 let declare_result = self.tokio_runtime.block_on(declare::declare(
                     contract.clone(),
                     fee_args,
+                    DryRunArgs {
```
ddoktorski: nit: default could be added to `DryRunArgs`

crates/sncast/src/response/dry_run.rs
```diff
@@ -0,0 +1,68 @@
+use bigdecimal::BigDecimal;
+use conversions::serde::serialize::CairoSerialize;
+use foundry_ui::styling;
+use serde::{Deserialize, Serialize};
+use starknet_rust::core::types::FeeEstimate;
+
+use crate::response::cast_message::SncastCommandMessage;
+
+#[derive(Serialize, Deserialize, Debug, Clone, CairoSerialize, PartialEq, Eq)]
```
ddoktorski: Can we somehow get rid of `PartialEq` and `Eq`?

crates/sncast/src/helpers/dry_run.rs
```diff
@@ -0,0 +1,36 @@
...
+    {
+        if !self.dry_run {
+            return None;
+        }
+        Some(
+            estimate_fee()
+                .await
+                .map(|e| into_response(DryRunResponse::new(&e, self.detailed))),
+        )
+    }
```
ddoktorski:
```suggestion
      pub async fn estimate<E, Fut>(
          &self,
          estimate_fee: impl FnOnce() -> Fut,
      ) -> Result<DryRunResponse, E>
      where
          Fut: Future<Output = Result<FeeEstimate, E>>,
      {
          estimate_fee()
              .await
              .map(|fee| DryRunResponse::new(&fee, self.detailed))
      }
  }
```
Please let's refactor this function more or less like this. It is easier to follow and does only one specific thing. Converting to command related type can be done on the caller site.

crates/sncast/src/response/dry_run.rs
franciszekjob: Removed `Eq`. PartialEq needs to stay since `DryRunResponse` is a variant in several enums (`DeclareResponse`, `InvokeResponse`, etc.) that all derive `PartialEq`.

crates/sncast/src/helpers/dry_run.rs
franciszekjob: Done

crates/sncast/src/starknet_commands/deploy.rs
MKowalski8: You could also do sth like this:

In `FeeArgs` add group 

```rust
#[group(id = "fee_args", multiple = true)]
pub struct FeeArgs {
```

And in `DryRunArgs` add just conflict with the group

``` rust
pub struct DryRunArgs {
    /// If passed, the transaction will not be sent to the network and the fee will be estimated instead.
    #[arg(long, conflicts_with = "fee_args")]
    pub dry_run: bool,

```

It's little cleaner and we can change the `FeeArgs` frearly in the future.

crates/sncast/src/response/dry_run.rs
franciszekjob: Yes, I've fixed it on an old branch 😅 (in this commit https://github.com/foundry-rs/starknet-foundry/commit/e3029e8c03f857f05b9c658f1629735993b82a81). Changed.

crates/sncast/src/starknet_commands/invoke.rs
```diff
@@ -68,7 +73,22 @@ where
         calldata,
     };
 
-    execute_calls(account, vec![call], fee_args, nonce, wait_config, ui).await
+    if dry_run_args.dry_run {
+        return dry_run_args
+            .estimate(|| async { account.execute_v3(vec![call.clone()]).estimate_fee().await })
```
MKowalski8: nit: We don't need this `clone()` since `call` isn't used later

```suggestion
            .estimate(|| async { account.execute_v3(vec![call]).estimate_fee().await })
```

MKowalski8 (approved):
LGTM now, left one minor comment

crates/sncast/src/starknet_commands/account/deploy.rs
```diff
@@ -303,6 +327,14 @@ where
 {
     let deployment = account_factory.deploy_v3(salt);
 
+    if dry_run_args.dry_run {
+        return dry_run_args
+            .estimate(|| async { deployment.estimate_fee().await })
```
ddoktorski: Is it necessary?
```suggestion
            .estimate(|| deployment.estimate_fee())
```

crates/sncast/src/starknet_commands/account/deploy.rs
```diff
@@ -303,6 +327,14 @@ where
...
+            .await
+            .map(InvokeResponse::DryRun)
+            .map_err(|error| anyhow!("Failed to estimate fee: {error}"));
```
ddoktorski: To be consistent with other errors
```suggestion
            .map_err(|error| anyhow!("Failed to estimate fee for dry run: {error}"));
```

crates/sncast/src/starknet_commands/declare.rs
```diff
@@ -116,11 +122,13 @@ pub fn compile_sierra_to_casm(
     Ok(casm)
 }
 
+#[allow(clippy::too_many_lines)]
```
ddoktorski:
```suggestion
#[expect(clippy::too_many_lines)]
```
also for `too_many_arguments`

crates/sncast/src/main.rs
```diff
@@ -634,7 +657,8 @@ async fn run_async_command(cli: Cli, config: CastConfig, ui: &UI) -> Result<()>
                     contract_address,
                     calldata,
                     nonce,
-                    fee_args,
+                    fee_args.clone(),
+                    dry_run_args.clone(),
```
ddoktorski: Do we need to clone here? Also in a few other places

franciszekjob: Not anymore

crates/sncast/src/starknet_commands/invoke.rs
```diff
@@ -93,10 +112,12 @@ pub async fn execute_calls<S>(
     nonce: Option<Felt>,
     wait_config: WaitForTx,
     ui: &UI,
-) -> Result<InvokeResponse, StarknetCommandError>
+) -> Result<InvokeTransactionResult, StarknetCommandError>
 where
     S: Signer + Sync + Send,
 {
+    // let execution = create_execution(account, calls, fee_args, proof_args, nonce).await;
```
ddoktorski: ?

crates/sncast/src/main.rs
```diff
@@ -647,7 +680,16 @@ async fn run_async_command(cli: Cli, config: CastConfig, ui: &UI) -> Result<()>
...
+                if let Ok(InvokeResponse::Transaction(invoke_response)) = &result {
+                    block_explorer_link_if_allowed(
+                        &Ok(invoke_response.clone()),
+                        provider.chain_id().await?,
+                        &config,
+                    )
+                    .await
+                } else {
+                    None
+                };
```
ddoktorski: Is this change necessary?

crates/sncast/src/starknet_commands/declare.rs
```diff
@@ -144,6 +151,15 @@ where
         casm_class_hash,
     );
 
+    if dry_run_args.dry_run {
+        return dry_run_args
+            .estimate(|| async { declaration.estimate_fee().await })
```
ddoktorski:
```suggestion
            .estimate(|| declaration.estimate_fee())
```
also at least in one more place

crates/sncast/src/main.rs
```diff
@@ -346,14 +348,25 @@ async fn run_async_command(cli: Cli, config: CastConfig, ui: &UI) -> Result<()>
...
                 }
                 DeclareResponse::AlreadyDeclared(_) => {
                     unreachable!("Argument `skip_on_already_declared` is false")
                 }
+                DeclareResponse::DryRun(_) => {
+                    unreachable!(
+                        "Dry run response should have been handled separately, as it does not contain a class hash"
+                    )
+                }
             });
```
ddoktorski: It seems that this entire logic can be simplified to one match:

```rust
let result = match result {
    Ok(DeclareResponse::DryRun(response)) => {
        process_command_result("declare", Ok(response), ui, None);
        return Ok(());
    }
    Ok(DeclareResponse::Success(tx)) => Ok(tx),
    Ok(DeclareResponse::AlreadyDeclared(_)) => {
        unreachable!("Argument `skip_on_already_declared` is false")
    }
    Err(e) => Err(handle_starknet_command_error(e)),
};
```
The same suggestion for `DeclareFrom`

crates/sncast/src/main.rs
```diff
@@ -533,9 +563,10 @@ async fn run_async_command(cli: Cli, config: CastConfig, ui: &UI) -> Result<()>
                 starknet_commands::deploy::deploy(
                     class_hash,
                     &calldata,
-                    salt,
-                    unique,
-                    fee_args,
+                    deploy.common.salt,
+                    deploy.common.unique,
+                    fee_args.clone(),
```
ddoktorski:
```suggestion
                    fee_args,
```
It seems that there are still a few redundant clones. I think we can simply derive `Copy` on both `DryRunArgs` and `FeeArgs`

crates/sncast/src/starknet_commands/declare.rs
```diff
@@ -144,6 +151,15 @@ where
         casm_class_hash,
     );
 
+    if dry_run_args.dry_run {
+        return dry_run_args
+            .estimate(|| declaration.estimate_fee())
+            .await
+            .map(DeclareResponse::DryRun)
+            .map_err(|e| anyhow!("Failed to estimate fee for dry run: {e}"))
+            .map_err(StarknetCommandError::from);
```
ddoktorski: In some places there is this additional mapping, but in others it's omitted (e.g. https://github.com/foundry-rs/starknet-foundry/pull/4228/changes#diff-f19fc5685ef68bf8e9a7674cb9a3b79a919e13bede994bba08d1e650426fd68bR111). What is the reason of this discrepancy?

ddoktorski (approved):

crates/sncast/src/starknet_commands/declare.rs
franciszekjob: The discrepancy comes from different return types of these functions.

`declare`, `deploy`, and `invoke` return `Result<_, StarknetCommandError>`, so after wrapping the estimation error with additional dry-run context using `anyhow!(...)`, we still need to convert it into `StarknetCommandError`.

The other places, like multicall/account deploy, return `anyhow::Result<_>`, so `map_err(|e| anyhow!(...))` is already enough there.

I updated the `StarknetCommandError` cases to make this explicit in a single `map_err`, so the difference should be clearer now (https://github.com/foundry-rs/starknet-foundry/pull/4228/changes/f240413efc7dc11bfed445da8bd48ae35ecd9b28).

PR #4229, opened 2026-03-23
Dry run tests

franciszekjob:
Towards #2316

## Introduced changes

Add tests for `--dry-run`

## Checklist

- [x] Linked relevant issue
- [ ] Updated relevant documentation
- [x] Added relevant tests
- [x] Performed self-review of the code
- [ ] Added changes to `CHANGELOG.md`

PR #4230, opened 2026-03-23
Dry run docs

### LINKED ISSUE #2316: Add option to only estimate commands fee without executing them

franciszekjob:
Closes #2316

## Introduced changes

Add docs for `--dry-run` flag.

## Checklist

- [x] Linked relevant issue
- [ ] Added relevant tests


franciszekjob:

- #4228
    - #4229
        - #4230 ◀

docs/src/starknet/invoke.md
```diff
@@ -52,6 +52,44 @@ transaction: https://sepolia.voyager.online/tx/[..]
...
+### Estimating Fee Without Submitting Transaction (Dry Run)
+
+Use `--dry-run` to estimate the transaction fee without submitting it to the network. Add `--detailed` to see a breakdown by resource type.
+
+<!-- TODO(#2736) -->
+<!-- { "ignored_output": true } -->
+```shell
+$ sncast \
+  --account my_account \
+    invoke \
```
MKowalski8: From what I can see we do not add extra indent for subcommend. [e.g](https://github.com/foundry-rs/starknet-foundry/blob/630df5f22c7b829d692becfac5d3fc55ec7e5bc5/docs/src/starknet/101.md?plain=1#L122)

```shell
$ sncast \
  --account my_account \
  invoke \
  --contract-address 0x00cd8f9ab31324bb93251837e4efb4223ee195454f6304fcfcb277e277653008 \
  --function put \
  --arguments '0x123, 100' \
  --network sepolia
```

So we can stick to the convention:

```suggestion
  --account my_account \
  invoke \
```

docs/src/starknet/invoke.md
```diff
@@ -52,6 +52,44 @@ transaction: https://sepolia.voyager.online/tx/[..]
 > `--max-fee` and the individual resource flags are mutually exclusive.
 > Any individual resource flag that is not provided will be estimated automatically
 
+### Estimating Fee Without Submitting Transaction (Dry Run)
```
MKowalski8: Does this section only belongs to `invoke`? Perhaps it's worth mentioning in other places as well? 
But I know it's always hard to place, and I don't have good solution

MKowalski8 (approved):

docs/src/starknet/invoke.md
franciszekjob: > Does this section only belongs to `invoke`? Perhaps it's worth mentioning in other places as well? But I know it's always hard to place, and I don't have good solution

Dry run is available in other commands, that's right, but I think mentioning it on other places in the same way may be too much? Mentioning it in appendix (for other commands) is imo enough.

CHANGELOG.md
```diff
@@ -19,6 +19,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Cast
 
+#### Added
+
+- `--dry-run` and `--detailed` flags to estimate the fee of a transaction without sending it to the network
```
```suggestion
- `--dry-run` and `--detailed` flags to all transaction commands, allowing to validate transaction logic and estimate fees without submitting to the network
```
Other phrasing can be used, the point is that it's not only about estimating fee. It's more of a functionality to verify correctness without sending to the network (as `dry run` name suggests)

docs/src/starknet/invoke.md
ddoktorski: I think it should be added as a separate site within the `sncast Overview`

ddoktorski (approved):

PR #4232, opened 2026-03-25
Fix using cheats in library calls

### LINKED ISSUE #3880: Cheating with library call from test doesn't work
### Current State

When making a library call from the test code itself, cheats are not reflected in the called contract.

### Objective

Fix cheating for library calls

### Additional Context

_No response_

ddoktorski:
Closes #3880

## Introduced changes

- Use `storage_address` instead of `caller_address` when getting cheated data for library call in tests
- Fix helper test function `library_call_contract` to use default caller address

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Forge
+
+- Bug with using cheats in library calls
+
 ## [0.58.0] - 2026-03-18
 
 ### Forge
--- a/crates/cheatnet/tests/common/mod.rs
+++ b/crates/cheatnet/tests/common/mod.rs
@@ -238,6 +238,9 @@ pub fn call_contract(
     .call_result
 }
 
+// This executes a library call as from a test contract.
+// `entry_point` below should match `library_call_syscall` entry point, except for `selector` and `calldata`:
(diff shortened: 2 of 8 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Forge
+
+- Bug with using cheats in library calls
```
franciszekjob, line 14:
```suggestion
### Forge

#### Fixed

- Bug with using cheats in library calls
```

crates/cheatnet/tests/common/mod.rs
```diff
@@ -253,7 +253,7 @@ pub fn library_call_contract(
         entry_point_selector,
         calldata,
         storage_address: TryFromHexStr::try_from_hex_str(TEST_ADDRESS).unwrap(),
-        caller_address: TryFromHexStr::try_from_hex_str(TEST_ADDRESS).unwrap(),
+        caller_address: ContractAddress::default(),
```
franciszekjob, line 259: nit: ContractAddress::default() as an inline value doesn't communicate why it's zero - for better readability, a separate variable (similarly as we do for `TEST_ADDRESS`) would make it more obvious this is intentional.

franciszekjob (approved):

crates/cheatnet/tests/common/mod.rs
ddoktorski: I want to align it with the actual code from call to blockifier runtime, added a comment instead

franciszekjob (approved):

DelevoXDG (approved):

PR #4240, opened 2026-03-31
Derive fuzzable trait

### LINKED ISSUE #2968: Add derive macro for `Fuzzable` trait

Forge

### Feature Request

Follow up to #2051

piotmag769:
Closes #2968

## Introduced changes

- self-explanatory

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests
- [x] Performed self-review of the code TODO

### DIFF
--- a/crates/forge/tests/data/fuzzing/tests/derive_fuzzable.cairo
+++ b/crates/forge/tests/data/fuzzing/tests/derive_fuzzable.cairo
@@ -1,0 +1,49 @@
+#[derive(Debug, Drop, Fuzzable)]
+struct InnerStruct {
+    value: u64,
+}
+
+#[derive(Debug, Drop, Fuzzable)]
+struct OuterStruct {
+    inner: InnerStruct,
+    extra: u32,
+}
+
+#[derive(Debug, Drop, Fuzzable)]
+struct MultiField<T> {
+    a: u256,
+    b: u128,
+    c: felt252,
+    flag: bool,
+    signed: i64,
+    address: starknet::ContractAddress,
+    t: T,
+}
+
+#[derive(Debug, Drop, Fuzzable)]
+enum InnerEnum {
+    Red: u8,
+    Green,
+    Blue: u32,
+}
+
+#[derive(Debug, Drop, Fuzzable)]
+enum OuterEnum<T> {
+    North: u8,
+    South,
+    East: InnerEnum,
+    West: OuterStruct,
+    Other: MultiField<T>,
+}
+
+#[fuzzer(runs: 10, seed: 1)]
+#[test]
+fn test_derived_enum(d: OuterEnum<u8>) {
+    println!("{d:?}");
+    let is_north = match d {
+        OuterEnum::North(_) => true,
+        _ => false,
+    };
+    // It fails deterministically since seed is set.
+    assert(is_north, 'expected North');
+}
--- a/crates/forge/tests/e2e/fuzzing.rs
+++ b/crates/forge/tests/e2e/fuzzing.rs
@@ -36,7 +36,7 @@ fn fuzzing() {
         [PASS] fuzzing::tests::uint128_arg (runs: 256, [..]
         [PASS] fuzzing::tests::uint256_arg (runs: 256, [..]
         Running 0 test(s) from tests/
-        Tests: 12 passed, 1 failed, 0 ignored, 12 filtered out
+        Tests: 12 passed, 1 failed, 0 ignored, 13 filtered out
         Fuzzer seed: [..]
 
         Failures:
@@ -81,7 +81,7 @@ fn fuzzing_set_runs() {
         [PASS] fuzzing::tests::uint128_arg (runs: 10, [..]
         [PASS] fuzzing::tests::uint256_arg (runs: 10, [..]
         Running 0 test(s) from tests/
-        Tests: 12 passed, 1 failed, 0 ignored, 12 filtered out
+        Tests: 12 passed, 1 failed, 0 ignored, 13 filtered out
         Fuzzer seed: [..]
 
         Failures:
@@ -126,7 +126,7 @@ fn fuzzing_set_seed() {
         [PASS] fuzzing::tests::uint128_arg (runs: 256, [..]
         [PASS] fuzzing::tests::uint256_arg (runs: 256, [..]
         Running 0 test(s) from tests/
-        Tests: 12 passed, 1 failed, 0 ignored, 12 filtered out
+        Tests: 12 passed, 1 failed, 0 ignored, 13 filtered out
         Fuzzer seed: 1234
 
         Failures:
@@ -196,7 +196,7 @@ fn fuzzing_exit_first() {
         Failure data:
             0x32202b2062203d3d2032202b2062 ('2 + b == 2 + b')
 
-        Tests: 0 passed, 1 failed, 0 ignored, 23 filtered out
+        Tests: 0 passed, 1 failed, 0 ignored, 24 filtered out
         Interrupted execution of 1 test(s).
 
         Fuzzer seed: [..]
@@ -232,7 +232,7 @@ fn fuzzing_exit_first_single_fail() {
         Failures:
             fuzzing_integrationtest::exit_first_single_fail::exit_first_fails_test
 
-        Tests: 0 passed, 1 failed, 0 ignored, 23 filtered out
+        Tests: 0 passed, 1 failed, 0 ignored, 24 filtered out
         Interrupted execution of 1 test(s).
         "},
     );
@@ -260,7 +260,7 @@ fn fuzzing_multiple_attributes() {
         [PASS] fuzzing_integrationtest::multiple_attributes::with_should_panic (runs: 256, [..])
         [PASS] fuzzing_integrationtest::multiple_attributes::with_available_gas (runs: 50, [..])
         [PASS] fuzzing_integrationtest::multiple_attributes::with_both (runs: 300, [..])
-        Tests: 3 passed, 0 failed, 1 ignored, 21 filtered out
+        Tests: 3 passed, 0 failed, 1 ignored, 22 filtered out
         "},
     );
 }
@@ -286,7 +286,7 @@ fn generate_arg_cheatcode() {
             "`generate_arg` cheatcode: `min_value` must be <= `max_value`, provided values after deserialization: 101 and 100"
 
         [PASS] fuzzing_integrationtest::generate_arg::use_generate_arg_outside_fuzzer (l1_gas: ~0, l1_data_gas: ~0, l2_gas: ~[..])
-        Tests: 1 passed, 1 failed, 0 ignored, 23 filtered out
+        Tests: 1 passed, 1 failed, 0 ignored, 24 filtered out
         "#},
     );
 }
@@ -337,7 +337,40 @@ fn fuzz_generic_struct() {
         Collected 1 test(s) from fuzzing package
         Running 1 test(s) from tests/
         [PASS] fuzzing_integrationtest::generic_struct::test_generic ([..])
-        Tests: 1 passed, 0 failed, 0 ignored, 24 filtered out
+        Tests: 1 passed, 0 failed, 0 ignored, 25 filtered out
         "},
+    );
+}
+
+#[test]
+fn fuzz_derive_fuzzable() {
+    let temp = setup_package("fuzzing");
+
+    let output = test_runner(&temp)
+        .arg("derive_fuzzable")
+        .env("SNFORGE_DETERMINISTIC_OUTPUT", "1")
+        .assert()
+        .code(1);
+
+    assert_stdout_contains(
+        output,
+        indoc! { r#"
+        [..]Compiling[..]
+        [..]Finished[..]
+
+
+        Collected 1 test(s) from fuzzing package
+        Running 1 test(s) from tests/
+        [FAIL] fuzzing_integrationtest::derive_fuzzable::test_derived_enum (runs: 1, arguments: ["OuterEnum::South(())"])
+
+        Failure data:
+            0x6578706563746564204e6f727468 ('expected North')
+
+        Tests: 0 passed, 1 failed, 0 ignored, 25 filtered out
+        Fuzzer seed: [..]
+
+        Failures:
+            fuzzing_integrationtest::derive_fuzzable::test_derived_enum
+        "#},
     );
 }
--- a/crates/snforge-scarb-plugin/src/derives/fuzzable.rs
+++ b/crates/snforge-scarb-plugin/src/derives/fuzzable.rs
@@ -1,0 +1,252 @@
+use crate::format_ident;
+use crate::utils::{SyntaxNodeUtils, create_single_token};
+use cairo_lang_macro::{Diagnostic, ProcMacroResult, TokenStream, quote};
+use cairo_lang_parser::utils::SimpleParserDatabase;
+use cairo_lang_syntax::node::Terminal;
+use cairo_lang_syntax::node::TypedSyntaxNode;
+use cairo_lang_syntax::node::ast::{
+    GenericParam, ItemEnum, ItemStruct, ModuleItem, OptionTypeClause,
+    OptionWrappedGenericParamList, SyntaxFile, TerminalIdentifier, Variant,
+};
+use cairo_lang_utils::Upcast;
+
+#[must_use]
+pub fn fuzzable_derive(item: &TokenStream) -> ProcMacroResult {
+    let db = SimpleParserDatabase::default();
+    let (parsed, diagnostics) = db.parse_token_stream(item);
+
+    if !diagnostics.is_empty() {
+        return ProcMacroResult::new(TokenStream::empty())
+            .with_diagnostics(Diagnostic::error("Failed because of invalid syntax").into());
+    }
+
+    let syntax_db = db.upcast();
+
+    SyntaxFile::from_syntax_node(syntax_db, parsed)
+        .items(syntax_db)
+        .elements(syntax_db)
+        .find_map(|module_item| match module_item {
+            ModuleItem::Struct(struct_item) => Some(generate_struct_impl(&db, &struct_item)),
+            ModuleItem::Enum(enum_item) => Some(generate_enum_impl(&db, &enum_item)),
+            _ => None,
+        })
+        .unwrap_or_else(|| {
+            ProcMacroResult::new(TokenStream::empty()).with_diagnostics(
+                Diagnostic::error("#[derive(Fuzzable)] can only be used on structs and enums")
+                    .into(),
+            )
+        })
+}
+
+fn generate_struct_impl(db: &SimpleParserDatabase, struct_item: &ItemStruct) -> ProcMacroResult {
+    let struct_name = struct_item.name(db);
+
+    let (impl_name_with_generics, name_with_type_args) =
+        build_generic_parts(db, &struct_name, struct_item.generic_params(db));
+
+    let blank_fields = build_struct_fields(db, struct_item, FuzzableMethod::Blank);
+    let generate_fields = build_struct_fields(db, struct_item, FuzzableMethod::Generate);
+
+    let struct_name = struct_name.to_token_stream(db);
+
+    let impl_code = quote! {
+        impl #impl_name_with_generics of snforge_std::fuzzable::Fuzzable<#name_with_type_args> {
+            fn blank() -> #name_with_type_args {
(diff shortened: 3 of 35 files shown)


crates/snforge-scarb-plugin/src/lib.rs
```diff
@@ -80,3 +84,9 @@ fn should_panic(args: TokenStream, item: TokenStream) -> ProcMacroResult {
 fn disable_predeployed_contracts(args: TokenStream, item: TokenStream) -> ProcMacroResult {
     disable_predeployed_contracts(args, item)
 }
+
+#[derive_macro]
+#[expect(clippy::needless_pass_by_value)]
```
piotmag769, line 89: No better way to do this sadly, I have to use it here or in `fuzzable_derive`

crates/snforge-scarb-plugin/tests/integration/fuzzable.rs
```diff
@@ -0,0 +1,769 @@
...
+                }
+            }
+        ",
+    );
+}
+
+#[test]
+fn enum_with_empty_variant_list() {
+    let item = quote!(
+        enum Empty {}
```
piotmag769: This is a valid Cairo enum

crates/forge/tests/data/fuzzing/tests/derive_fuzzable.cairo
piotmag769, line 1: Just a quite complex example that covers almost every possible case

docs/listings/fuzz_testing/src/derive_fuzzable.cairo
```diff
@@ -0,0 +1,6 @@
+#[cfg(test)]
```
piotmag769, line 1: This is weird, but it has to be like that - `Fuzzable` is available via `snforge_scarb_plugin` which is in dev deps (since `snforge-std` is in dev-deps). Sadly we don't have `cfg_attr` in Cairo as of now, so if a struct is not just for testing, users have to do:

```cairo
#[cfg(test)]
#[derive(Debug, Drop, Fuzzable)]
struct Point {
    x: u64,
    y: u64,
}

#[derive(Debug, Drop)]
struct Point {
    x: u64,
    y: u64,
}
```

Instead of 
```cairo
#[cfg_attr(test, Fuzzable)]
#[derive(Debug, Drop)]
struct Point {
    x: u64,
    y: u64,
}
```

crates/snforge-scarb-plugin/src/derives/fuzzable.rs
```diff
@@ -0,0 +1,307 @@
...
+                let tp_name_str = tp_name_syntax.get_text_without_trivia(db.upcast());
+                let tp_name = SyntaxNodeWithDb::new(&tp_name_syntax, db);
+
+                if !type_args.is_empty() {
+                    type_args.extend(quote!(,));
+                }
+                type_args.extend(quote!(#tp_name));
+
+                let fuzzable_pattern = format!("Fuzzable<{tp_name_str}>");
+                let debug_pattern = format!("Debug<{tp_name_str}>");
```
piotmag769: It has to be added separately since supertraits don't exist in Cairo, we imitate them via generic impl params in trait definitions `trait Fuzzable<+Debug<T>>`. Otherwise we would get a compilation error

crates/snforge-scarb-plugin/tests/integration/snapshots/integration__fuzzable__struct_with_generic_type_param.snap
```diff
@@ -0,0 +1,14 @@
+---
+source: tests/integration/fuzzable.rs
+expression: format_output(&result)
+---
+impl FuzzableContainerImpl<
+    T, +snforge_std::fuzzable::Fuzzable<T>, +core::fmt::Debug<T>, +core::fmt::Debug<T>,
```
ddoktorski: Is it fine to have double `+core::fmt::Debug<T>` bound?

crates/snforge-scarb-plugin/src/derives/fuzzable.rs
```diff
@@ -0,0 +1,259 @@
...
+
+                impl_generic_params.extend(quote!(, +snforge_std::fuzzable::Fuzzable<#tp_name>));
+                impl_generic_params.extend(quote!(, +core::fmt::Debug<#tp_name>));
+            }
+            // Impl params still occupy a generic arg slot in the type,
+            // so we emit `_` to keep arg counts aligned.
+            GenericParam::ImplNamed(_) | GenericParam::ImplAnonymous(_) => {
+                append_comma_separated(&mut type_generic_params, quote!(_));
+            }
+            _ => {
```
ddoktorski: Does it work correctly for `GenericParam::Const` and `GenericParam::NegativeImpl`?

crates/snforge-scarb-plugin/src/derives/fuzzable.rs
```diff
@@ -0,0 +1,259 @@
...
+    let (parsed, diagnostics) = db.parse_token_stream(item);
+
+    if !diagnostics.is_empty() {
+        return ProcMacroResult::new(TokenStream::empty())
+            .with_diagnostics(Diagnostic::error("Failed because of invalid syntax").into());
+    }
+
+    let syntax_db = db.upcast();
+
+    for module_item in SyntaxFile::from_syntax_node(syntax_db, parsed)
```
ddoktorski: nit: `for` loop is a bit misleading as it suggests that there are multiple items to process. Can we add a small comment or replace it with `find_map`? (it'd be consistent with other places then)

crates/snforge-scarb-plugin/tests/integration/snapshots/integration__fuzzable__struct_with_generic_type_param.snap
piotmag769: It is not fine, but user should just not specify this bound manually. It they do - it will be a semantic error. Same thing happens with `Debug` derive btw, but they only add bounds to generic params that are used in members - you can check that this code produces semantic errors:

```cairo
#[derive(Debug)]
struct X<T, +core::fmt::Debug<T>> {
    x: T,
}
```

since the impl it generates looks like this:

```cairo
impl XDebug<
    T, +core::fmt::Debug<T>, impl __MEMBER_IMPL_x_Debug: core::fmt::Debug<T>,
> of core::fmt::Debug<X<T, _>> {
  // ...
}
```

Will change the test to use another trait.

crates/snforge-scarb-plugin/src/derives/fuzzable.rs
piotmag769: Now it does

crates/snforge-scarb-plugin/tests/integration/fuzzable.rs
```diff
@@ -0,0 +1,399 @@
...
+
+    let result = fuzzable_derive(&item);
+
+    assert_diagnostics(&result, &[]);
+
+    insta::assert_snapshot!(format_output(&result));
+}
+
+#[test]
+fn enum_with_empty_variant_list() {
```
ddoktorski: nit: It seems to be identical with test `error_on_empty_enum`

docs/src/appendix/snforge-library/fuzzable.md
```diff
@@ -18,9 +18,37 @@ Any type that is used as a parameter in a test function with the [`#[fuzzer]`](.
...
+#[derive(Debug, Drop, Fuzzable)]
+enum Direction {
+    North,
+    South,
+    East,
+    West,
+}
+```
+
+For generic types, `#[derive(Fuzzable)]` automatically appends a `+snforge_std::fuzzable::Fuzzable<T>` and `+core::fmt::Debug<T>` bound for each type parameter.
```
ddoktorski, line 41: Regarding https://github.com/foundry-rs/starknet-foundry/pull/4240#discussion_r3211315063 maybe let's phrase it more explicitly here so it's clear that `#[derive(Fuzzable)]` will fail if the bound
`+core::fmt::Debug<T>` is already specified

ddoktorski (approved):

docs/src/appendix/snforge-library/fuzzable.md
piotmag769: Is it worth explaining how the language works though? Kinda a weird place to do that imo, especially since `derive(Debug)` will fail as well if this "bound" (generic impl param) is specified. I am all for adding it if someone asks about it on support though

crates/forge/tests/e2e/fuzzing.rs
```diff
@@ -337,7 +337,40 @@ fn fuzz_generic_struct() {
...
+        [FAIL] fuzzing_integrationtest::derive_fuzzable::test_derived_enum (runs: [..], arguments: [[..]])
+
+        Failure data:
+            [..]
+
+        Tests: 0 passed, 1 failed, 0 ignored, 25 filtered out
+        Fuzzer seed: [..]
+
+        Failures:
+            fuzzing_integrationtest::derive_fuzzable::test_derived_enum
```
DelevoXDG, line 373: I think we can assert more precisely
```suggestion
        [..]Compiling[..]
        [..]Finished[..]


        Collected 1 test(s) from fuzzing package
        Running 1 test(s) from tests/
        [FAIL] fuzzing_integrationtest::derive_fuzzable::test_derived_enum (runs: 1, arguments: ["OuterEnum::South(())"])

        Failure data:
            0x6578706563746564204e6f727468 ('expected North')

        Tests: 0 passed, 1 failed, 0 ignored, 25 filtered out
        Fuzzer seed: [..]

        Failures:
            fuzzing_integrationtest::derive_fuzzable::test_derived_enum
```

crates/snforge-scarb-plugin/src/derives/fuzzable.rs
```diff
@@ -0,0 +1,251 @@
...
+
+#[must_use]
+pub fn fuzzable_derive(item: &TokenStream) -> ProcMacroResult {
+    let db = SimpleParserDatabase::default();
+    let (parsed, diagnostics) = db.parse_token_stream(item);
+
+    if !diagnostics.is_empty() {
+        return ProcMacroResult::new(TokenStream::empty())
+            .with_diagnostics(Diagnostic::error("Failed because of invalid syntax").into());
+    }
```
DelevoXDG, line 21: Unrelated to this specific PR, since it's a copypaste from what I see - but isn't this problematic?

I'm not sure derive should fail over warning diagnostics?

crates/snforge-scarb-plugin/tests/integration/snapshots/integration__fuzzable__struct_with_generic_type_param.snap
DelevoXDG: Can't we check whether `+core::fmt::Debug<T>` is already present, and add it conditionally?

crates/snforge-scarb-plugin/src/derives/fuzzable.rs
```diff
@@ -0,0 +1,251 @@
...
+                build_variant_value(db, enum_name, variant, FuzzableMethod::Generate);
+            let idx = int_token(idx);
+            quote!(#idx => { #variant_value },)
+        })
+        .fold(TokenStream::empty(), |mut acc, x| {
+            acc.extend(x);
+            acc
+        });
+
+    let n_minus_one = int_token(variants.len() - 1);
```
DelevoXDG, line 139: Nit: Perhaps worth writing an invariant:
```
Note: `variants` must be non-empty (caller must guard this).
```

crates/snforge-scarb-plugin/tests/integration/snapshots/integration__fuzzable__enum_mixed_variants.snap
```diff
@@ -0,0 +1,17 @@
+---
+source: tests/integration/fuzzable.rs
+expression: format_output(&result)
+---
+impl FuzzableMixedImpl<> of snforge_std::fuzzable::Fuzzable<Mixed<>> {
+    fn blank() -> Mixed<> {
+        Mixed::Unit
+    }
+    fn generate() -> Mixed<> {
```
DelevoXDG, line 9: Can't we / shouldn't we remove empty `<>` params in `build_generic_parts`?

crates/snforge-scarb-plugin/src/derives/fuzzable.rs
piotmag769: These are parser diagnostics - so only errors

crates/snforge-scarb-plugin/tests/integration/snapshots/integration__fuzzable__enum_mixed_variants.snap
piotmag769: It does not matter tbh and the proc macro code is a bit cleaner this way (no need to add early return from the function)

crates/snforge-scarb-plugin/tests/integration/snapshots/integration__fuzzable__struct_with_generic_type_param.snap
piotmag769: >Can't we check whether +core::fmt::Debug<T> is already present, and add it conditionally?

I can create an issue for that, but even `derive(Debug)` which is compiler builtin doesn't do that. It is smarter than us just in this case:

>only add bounds to generic params that are used in members

snforge_std/src/fuzzable.cairo
```diff
@@ -108,3 +108,13 @@ pub impl FuzzableContractAddress of Fuzzable<ContractAddress> {
...
+
+pub impl FuzzableUnitType of Fuzzable<()> {
+    fn blank() -> () {
+        ()
+    }
+
+    fn generate() -> () {
+        ()
+    }
+}
```
piotmag769, line 120: Added it after consulting with @ddoktorski - someone may have a struct field of unit type for some reason, this allows to fuzz such structs

crates/snforge-scarb-plugin/tests/integration/snapshots/integration__fuzzable__struct_with_generic_type_param.snap
DelevoXDG: > I can create an issue for that

Let's do that 

> but even `derive(Debug)` which is compiler builtin doesn't do that.

Yes, but I suppose `#[derive(Debug)]` was meant to be used by itself.

But now we can expect both to be derived (`#[derive(Debug, Fuzzable)]`), which will always produce double `+core::fmt::Debug<T>`.

Although maybe it's a non-issue, same as https://github.com/foundry-rs/starknet-foundry/pull/4240#discussion_r3229314780?

crates/snforge-scarb-plugin/src/derives/fuzzable.rs
DelevoXDG: Shouldn't assert that though?

This is current truth at `cairo-lang-parser` level, but it's not something we guarantee here 🤔 

(once again, this is probably out of scope for this PR, but I generally think this should be changed)

piotmag769: I can make a PR for that, but tbh I have never seen a warning from a parser ;p I guess it can be the case when someone uses a syntax that can be ambiguous from human POV (while still being unambiguous from parser POV)

crates/snforge-scarb-plugin/tests/integration/snapshots/integration__fuzzable__struct_with_generic_type_param.snap
piotmag769: >But now we can expect both to be derived (#[derive(Debug, Fuzzable)]), which will always produce double +core::fmt::Debug<T>.

Nope, not at all. E.g. this

```cairo
#[derive(Debug, Fuzzable, Drop)]
enum U<T> {
    B: T,
}
```

expands to

```cairo
#[derive(Debug, Fuzzable, Drop)]
enum U<T> {
    B: T,
}
impl FuzzableUImpl<
    T, +snforge_std::fuzzable::Fuzzable<T>, +core::fmt::Debug<T>,
> of snforge_std::fuzzable::Fuzzable<U<T>> {
    fn blank() -> U<T> {
        U::B(snforge_std::fuzzable::Fuzzable::<T>::blank())
    }
    fn generate() -> U<T> {
        let variant_idx = snforge_std::fuzzable::generate_arg(0, 0);
        match variant_idx {
            0 => { U::B(snforge_std::fuzzable::Fuzzable::<T>::generate()) },
            _ => core::panics::panic_with_byte_array(@"unexpected value of generate_arg cheatcode"),
        }
    }
}
impl UDebug<T, impl __MEMBER_IMPL_B_Debug: core::fmt::Debug<T>> of core::fmt::Debug<U<T>> {
    fn fmt(self: @U<T>, ref f: core::fmt::Formatter) -> core::result::Result<(), core::fmt::Error> {
        match self {
            U::B(x) => {
                {
                    core::byte_array::ByteArrayTrait::append_word(ref f.buffer, 0x553a3a4228, 5);
                    core::result::Result::<(), core::fmt::Error>::Ok(())
                }?;
                __MEMBER_IMPL_B_Debug::fmt(x, ref f)?;
                {
                    core::byte_array::ByteArrayTrait::append_word(ref f.buffer, 0x29, 1);
                    core::result::Result::<(), core::fmt::Error>::Ok(())
                }
            },
        }
    }
}
impl UDrop<T, impl __MEMBER_IMPL_B_Drop: core::traits::Drop<T>> of core::traits::Drop<U<T>>;
```

and it is perfectly valid

DelevoXDG: I suppose I assumed that this is caused by two derives, based on many snapshots that you added that do have duplicate bound, e.g.:
```rs
impl FuzzableMultiFieldImpl<
    T,
    +snforge_std::fuzzable::Fuzzable<T>,
    +core::fmt::Debug<T>,
    +core::fmt::Debug<T>,
    A,
    +snforge_std::fuzzable::Fuzzable<A>,
    +core::fmt::Debug<A>,
> of snforge_std::fuzzable::Fuzzable<MultiField<T, _, A>> {
    fn blank() -> MultiField<T, _, A> {
        MultiField {
            t: snforge_std::fuzzable::Fuzzable::<T>::blank(),
            a: snforge_std::fuzzable::Fuzzable::<A>::blank(),
        }
    }
    fn generate() -> MultiField<T, _, A> {
        MultiField {
            t: snforge_std::fuzzable::Fuzzable::<T>::generate(),
            a: snforge_std::fuzzable::Fuzzable::<A>::generate(),
        }
    }
}

```

is there some other reason for that?

piotmag769: >that this is caused by two derives

Not at all. It came from this cairo code:

```cairo
#[derive(Fuzzable)]
struct MultiField<T, +core::fmt::Debug<T>, A>  {
    t: T,
    a: A,
}
```

So the double `+core::fmt::Debug<T>` is caused by it being specified in the struct definition, not the presence of another derive. Changed all tests to have different trait bound so it isn't confusing.

You can see that `#[derive(Debug, Fuzzable)]` combination works even in the integration test I added https://github.com/foundry-rs/starknet-foundry/blob/a4bf830647322ba890e23d575c8f620393ca3df3/crates/forge/tests/data/fuzzing/tests/derive_fuzzable.cairo#L1

crates/snforge-scarb-plugin/src/derives/fuzzable.rs
piotmag769: #4361

crates/snforge-scarb-plugin/tests/integration/snapshots/integration__fuzzable__struct_with_generic_type_param.snap
DelevoXDG: Ah, I see it was that `struct_with_impl_bound_between_type_params` that was adding `+core::fmt::Debug<T>`

```rs
#[test]
fn struct_with_impl_bound_between_type_params() {
    let item = quote!(
        struct MultiField<T, +core::fmt::Debug<T>, A> {
            t: T,
            a: A,
        }
    );
    ...

```

> So the double +core::fmt::Debug<T> is caused by it being specified in the struct definition, not the presence of another derive. Changed all tests to have different trait bound so it isn't confusing.

👍🏻

DelevoXDG (approved):
Let's add that issue though 🙏🏻 

https://github.com/foundry-rs/starknet-foundry/pull/4240#discussion_r3233348287

Issue #2676, opened 2024-11-15
Compile workspaces in parallel when running `snforge test -w`


cptartur:
Right now, when running `snforge test -w` to run tests for whole workspace, we compile packages sequentially: The compilation of a next package only starts after previous has been completely executed (all tests run). This introduces unnecessary wait time.

Change the code so that the compilation of a next package in a workspace starts during the execution phase of the previous one.

ddoktorski:
Done in #4246

PR #4246, opened 2026-04-01
Parallelize config passes with test execution across workspace packages

### LINKED ISSUE #2675: Start executing test targets in parallel
Currently, we first prepare (execute config run etc.) test for all targets in package before executing them. This means that forge is spending some time waiting for all targets to compile even though some are already ready to execute.

Change the logic so that after being compiled we already start executing the first target. Prepare other targets in parallel.

Arcticae:
[snforge-profile-2026-04-01T14:55:23.716033+02:00.json](https://github.com/user-attachments/files/26410764/snforge-profile-2026-04-01T14.55.23.716033%2B02.00.json)

image

image

Benchmark (with scarb incremental - on OZ)
image

Spawn config handles for all packages upfront so CASM compilations overlap with test execution rather than blocking it.

Closes #2675 

## Introduced changes

-

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests

### DIFF
--- a/crates/forge/src/run_tests/package.rs
+++ b/crates/forge/src/run_tests/package.rs
@@ -67,7 +67,7 @@ impl PackageTestResult {
 }
 
 pub struct RunForPackageArgs {
-    config_handles: Vec<JoinHandle<Result<TestTargetWithConfig>>>,
+    pub config_handles: Vec<JoinHandle<Result<TestTargetWithConfig>>>,
     pub tests_filter: TestsFilter,
     pub forge_config: Arc<ForgeConfig>,
     pub fork_targets: Vec<ForkTarget>,
--- a/crates/forge/src/run_tests/package.rs
+++ b/crates/forge/src/run_tests/package.rs
@@ -30,7 +30,7 @@ use forge_runner::{
         with_config_resolved::{TestCaseWithResolvedConfig, sanitize_test_case_name},
     },
     partition::PartitionConfig,
-    running::with_config::test_target_with_config,
+    running::target::prepare_test_target,
     scarb::load_test_artifacts,
     test_case_summary::AnyTestCaseSummary,
     test_target_summary::TestTargetSummary,
@@ -67,7 +67,7 @@ impl PackageTestResult {
 }
 
 pub struct RunForPackageArgs {
-    pub config_handles: Vec<JoinHandle<Result<TestTargetWithConfig>>>,
+    pub target_handles: Vec<JoinHandle<Result<TestTargetWithConfig>>>,
     pub tests_filter: TestsFilter,
     pub forge_config: Arc<ForgeConfig>,
     pub fork_targets: Vec<ForkTarget>,
@@ -138,13 +138,13 @@ impl RunForPackageArgs {
 
         let tracked_resource = forge_config.test_runner_config.tracked_resource;
 
-        let config_handles = raw_test_targets
+        let target_handles = raw_test_targets
             .into_iter()
-            .map(|t| spawn_config_pass(t, tracked_resource))
+            .map(|t| spawn_prepare_test_target(t, tracked_resource))
             .collect();
 
         Ok(RunForPackageArgs {
-            config_handles,
+            target_handles,
             forge_config,
             tests_filter,
             fork_targets: forge_config_from_scarb.fork,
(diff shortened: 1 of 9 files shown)


maciektr:
image
I am impressed, nice job Claude!

franciszekjob (approved):
I also triggered maat with nightly (created from this branch) [here](https://docs.swmansion.com/maat/#s=nightly-nightly-2026-04-01-nightly-2026-04-03&d=forge).

LGTM, thanks guys!

crates/forge/src/run_tests/package.rs
```diff
@@ -131,49 +132,33 @@ impl RunForPackageArgs {
...
-
-fn sum_test_cases_from_package(
-    test_targets: &[TestTargetWithResolvedConfig],
-    partitioning_config: &PartitionConfig,
-) -> usize {
-    test_targets
-        .iter()
-        .map(|tt| sum_test_cases_from_test_target(&tt.test_cases, partitioning_config))
-        .sum()
+fn spawn_config_pass(
```
ddoktorski:
```suggestion
fn spawn_prepare_test_target(
```
Given that `test_target_with_config` does both compilation and running config pass I'd rename it to sth more general. As a result `test_target_with_config` should be also renamed to `prepare_test_target` or similar

crates/forge/tests/integration/setup_fork.rs
```diff
@@ -132,54 +134,65 @@ fn fork_aliased_decorator() {
...
-                            .unwrap(),
-                        tracked_resource: ForgeTrackedResource::CairoSteps,
-                        environment_variables: test.env().clone(),
+        .block_on(async {
+            let config_handles = raw_test_targets
+                .into_iter()
+                .map(|t| {
+                    tokio::task::spawn_blocking(move || {
+                        test_target_with_config(t, &ForgeTrackedResource::CairoSteps)
+                    })
```
ddoktorski, line 143: nit: This code is repeated a few times, what about extracting it to a utility function?

crates/forge/tests/integration/setup_fork.rs
```diff
@@ -74,6 +75,7 @@ fn fork_simple_decorator() {
     assert_passed(&result);
 }
 
+#[allow(clippy::too_many_lines)]
```
ddoktorski:
```suggestion
#[expect(clippy::too_many_lines)]
```

ddoktorski (commented):
very nice improvement 💪 

I think it also closes #2676

crates/forge/tests/integration/setup_fork.rs
Arcticae: i don't see how i would do that apart from aliasing, which would only obstruct what is actually happening with the thread spawning

crates/forge/tests/integration/setup_fork.rs
Arcticae: Done

crates/forge/src/run_tests/package.rs
Arcticae: Done

crates/forge/src/run_tests/package.rs
```diff
@@ -200,61 +185,71 @@ fn sum_test_cases_from_test_target(
...
-    let not_filtered =
-        sum_test_cases_from_package(&test_targets, &tests_filter.partitioning_config);
     ui.println(&CollectedTestsCountMessage {
-        tests_num: not_filtered,
+        tests_num: not_filtered_total,
         package_name: package_name.clone(),
     });
 
-    let mut summaries = vec![];
+    warn_if_incompatible_rpc_version(&resolved_targets, ui.clone()).await?;
```
ddoktorski: Actually why it is printed now after the collected tests message and not before? Both are fine for me, just asking if there is any reason

crates/forge/tests/integration/setup_fork.rs
ddoktorski: Ok up to you

ddoktorski (approved):

crates/forge/src/run_tests/package.rs
Arcticae: Ah i think this is no good and accidental here, will revert it not to tamper with the user output

PR #4234, opened 2026-03-30
Docs: sncast config changes

DelevoXDG:
WIP, awaiting feedback and final decisions in #4205 

## Stack
- #4175
- #4205 
- #4249 
- #4234 

## Checklist

- [x] Updated relevant documentation
- [x] Added changes to `CHANGELOG.md`

PR #4249, opened 2026-04-05
Separate Scarb profile used from `sncast --profile`

DelevoXDG:
## Stack
- #4175
- #4205 
- #4249
- #4234 

## Introduced changes

Allow configuring scarb profile separately from sncast config:
- `--scarb-profile` in cli
- `scarb-profile` in `snfoundry.toml` 

## Checklist

- [ ] Linked relevant issue
- [ ] Added relevant tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,10 +30,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Aliases `sncast get transaction` and `sncast get transaction-status` for `get tx` and `get tx-status`.
 - Support for serialization of corelib `Option` and `Result`, values passed via `--arguments`
 - `--dry-run` and `--detailed` flags for all transaction commands, allowing users to validate transaction logic and estimate fees without submitting to the network
+- Config setting for Scarb profile, as `scarb-profile` field in `snfoundry.toml`, and `--scarb-profile` cli flag
+
 
 #### Changed
 
 - In JSON output for `sncast utils` commands, the `"command"` field now includes the `utils` prefix (e.g. `"serialize"` -> `"utils serialize"`).
+- `--profile` now only determines `snfoundry.toml` profile. For Scarb profile, use `--scarb-profile` flag or `scarb-profile` from `snfoundry.toml` instead. Defaults to `release` if unspecified.
 
 #### Fixed
 
--- a/crates/sncast/src/starknet_commands/script/mod.rs
+++ b/crates/sncast/src/starknet_commands/script/mod.rs
@@ -51,7 +51,7 @@ pub fn run_script_command(
                 &BuildConfig {
                     scarb_toml_path: manifest_path.clone(),
                     json: cli.json,
-                    profile: cli.profile.clone().unwrap_or("dev".to_string()),
+                    profile: config.scarb_profile.clone(),
                 },
                 true,
                 // TODO(#3959) Remove `base_ui`
--- a/docs/src/appendix/snfoundry-toml.md
+++ b/docs/src/appendix/snfoundry-toml.md
@@ -104,7 +104,7 @@ block-explorer = "Voyager"
 
 The `scarb-profile` field sets which [Scarb profile](https://docs.swmansion.com/scarb/docs/reference/manifest.html#profiles) `sncast` uses when it runs `scarb` commands (e.g. `scarb build` when declaring contracts or running scripts).
 
-If omitted, the effective default is `release`.
+Defaults to `release`.
 
 ```toml
 [sncast.myprofile]
--- a/docs/src/appendix/snfoundry-toml.md
+++ b/docs/src/appendix/snfoundry-toml.md
@@ -102,7 +102,7 @@ block-explorer = "Voyager"
 
 #### `scarb-profile`
 
-The `scarb-profile` field sets which [Scarb profile](https://docs.swmansion.com/scarb/docs/reference/manifest.html#profiles) `sncast` uses when it runs `scarb` commands (e.g. `scarb build` when declaring contracts or running scripts).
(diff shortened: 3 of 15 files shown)


docs/src/appendix/snfoundry-toml.md
```diff
@@ -100,6 +100,17 @@ The `block-explorer` field specifies the block explorer service used to display
 block-explorer = "Voyager"
 ```
 
+#### `scarb-profile`
+
+The `scarb-profile` field sets which [Scarb profile](https://docs.swmansion.com/scarb/docs/reference/manifest.html#profiles) `sncast` uses when it runs `scarb` commands (e.g. `scarb build` when declaring contracts or running scripts).
+
+If omitted, the effective default is `release`.
```
franciszekjob, line 107:
```suggestion
nit: Defaults to `release`.
```

docs/src/appendix/snfoundry-toml.md
```diff
@@ -100,6 +100,17 @@ The `block-explorer` field specifies the block explorer service used to display
 block-explorer = "Voyager"
 ```
 
+#### `scarb-profile`
+
+The `scarb-profile` field sets which [Scarb profile](https://docs.swmansion.com/scarb/docs/reference/manifest.html#profiles) `sncast` uses when it runs `scarb` commands (e.g. `scarb build` when declaring contracts or running scripts).
```
franciszekjob, line 105: Wouldn't https://docs.swmansion.com/scarb/docs/reference/profiles.html be better to link 🤔 ?

CHANGELOG.md
```diff
@@ -7,6 +7,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 
+### Cast
+
+#### Added
+
+- Config setting for Scarb profile, as `scarb-profile` field in `snfoundry.toml`, and `--scarb-profile` cli flag
+
+#### Changed
+
+- `--profile` now only determines `snfoundry.toml` profile. For Scarb profile, use `--scarb-profile` flag or `scarb-profile` from `snfoundry.toml` instead. Defaults to `release` if unspecified.
```
franciszekjob, line 18: nit: Imo need to mention default value
```suggestion
- `--profile` now only determines `snfoundry.toml` profile. For Scarb profile, use `--scarb-profile` flag or `scarb-profile` from `snfoundry.toml` instead
```

crates/sncast/src/starknet_commands/script/mod.rs
```diff
@@ -50,7 +50,7 @@ pub fn run_script_command(
                 &BuildConfig {
                     scarb_toml_path: manifest_path.clone(),
                     json: cli.json,
-                    profile: cli.profile.clone().unwrap_or("dev".to_string()),
```
ddoktorski, line 53: I'm not sure if we can make this change. Scripts require `dev` profile to run properly (at least that was the case in the past). Could you verify it?

docs/src/appendix/snfoundry-toml.md
DelevoXDG:
```suggestion
Defaults to `release`.
```

docs/src/appendix/snfoundry-toml.md
DelevoXDG: Sure

DelevoXDG:
```suggestion
The `scarb-profile` field sets which [Scarb profile](https://docs.swmansion.com/scarb/docs/reference/profiles.html) `sncast` uses when it runs `scarb` commands (e.g. `scarb build` when declaring contracts or running scripts).
```

CHANGELOG.md
DelevoXDG: Why not?

franciszekjob: It seems mentioning a default value in changelog may be a bit too detailed? On the other hand, it's just one flag added, so it may be helpful. Leaving this up to you.

crates/sncast/src/starknet_commands/script/mod.rs
DelevoXDG: Please note that even prior to my changes:
1. It was overridable via cli `--profile`.
2. In some cases, we defaulted to 'release` profile 
   ```
   [WARNING] Profile profile5 does not exist in scarb, using 'release' profile.
   ```
   Which I assume is intentional: https://github.com/foundry-rs/starknet-foundry/pull/2456
3. There was a test `test_using_release_profile` that verified that `sncast script runs` does work with scarb `release` profile

So I assume scripts *should* work with `release` profile 🤔 

I don't mind defaulting to dev in applicable cases - but do you know of such cases, or what specifically can go wrong?

Importantly, if there are contexts in which `release` profile is breaking to the script behaviour, we should consider disallowing override, or at least displaying a warning

ddoktorski: Thanks for linking the PR, it introduced the logic to always use `dev` for scripts which is also clearly stated in the description there. The actual logic is located here https://github.com/foundry-rs/starknet-foundry/blob/zdobnikau/fix-config-parsing-2/crates/sncast/src/helpers/scarb_utils.rs#L152 and this PR doesn't change it, so my above comment can be ignored as it seems that no matter what profile is used here, eventually script uses `dev` built artifacts

ddoktorski (approved):

franciszekjob (approved):

crates/sncast/src/starknet_commands/script/mod.rs
DelevoXDG: > Thanks for linking the PR, it introduced the logic to always use `dev` for scripts which is also clearly stated in the description there.

Good catch 👍🏻

PR #4251, opened 2026-04-07
Support blake opcode for profiler

### LINKED ISSUE #249: Support new blake2s opcode
It should be treated as a builtin for cost and profiling purposes

Arcticae:
Closes https://github.com/software-mansion/cairo-profiler/issues/249

## Introduced changes

Profiler ran with profile generated with: `--tracked-resource cairo-steps`
image

## Checklist

- [ ] Linked relevant issue
- [ ] Added relevant tests

### DIFF
--- a/crates/forge-runner/src/build_trace_data.rs
+++ b/crates/forge-runner/src/build_trace_data.rs
@@ -146,6 +146,7 @@ pub fn build_profiler_execution_resources(
     // Subtract syscall related resources to get the values expected by the profiler.
     // The profiler operates on resources excluding syscall overhead.
     let versioned_constants = VersionedConstants::latest_constants();
+    let opcodes = execution_resources.opcode_instance_counter.clone();
     let execution_resources = &execution_resources.vm_resources
         - &versioned_constants.get_additional_os_syscall_resources(syscall_usage_vm_resources);
     let gas_consumed =
@@ -171,6 +172,12 @@ pub fn build_profiler_execution_resources(
                 .builtin_instance_counter
                 .into_iter()
                 .map(|(key, value)| (key.to_str_with_suffix().to_owned(), value))
+                .chain(
+                    // Treat opcodes as builtin instances.
+                    opcodes
+                        .into_iter()
+                        .map(|(key, value)| (key.to_str_with_suffix().to_owned(), value)),
+                )
                 .collect(),
         },
         gas_consumed: Some(gas_consumed),


crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
```diff
@@ -646,7 +646,16 @@ pub fn add_resources_to_top_call(
...
-        TrackedResource::CairoSteps => top_call.used_execution_resources += resources,
+        TrackedResource::CairoSteps => {
+            top_call.used_execution_resources += &resources.vm_resources;
+            for (opcode, count) in &resources.opcode_instance_counter {
+                *top_call
+                    .used_execution_resources
+                    .opcode_instance_counter
+                    .entry(*opcode)
+                    .or_insert(0) += count;
+            }
```
ddoktorski: `ExtendedExecutionResources` has `AddAssign` implementation. Isn't it the same code as before, but longer?

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/cairo1_execution.rs
```diff
@@ -102,14 +103,29 @@ pub(crate) fn execute_entry_point_call_cairo1(
         .syscalls_usage
         .clone();
 
-    let call_info = finalize_execution(
+    let mut call_info = finalize_execution(
```
ddoktorski: I don't think we should ever modify the returned call info

crates/cheatnet/src/runtime_extensions/call_to_blockifier_runtime_extension/execution/cairo1_execution.rs
```diff
@@ -102,14 +103,29 @@ pub(crate) fn execute_entry_point_call_cairo1(
...
         runner,
         cheatable_runtime.extended_runtime.hint_handler,
         n_total_args,
         program_extra_data_length,
         tracked_resource,
     )?;
 
+    // Propagate opcode counts (e.g. blake) into resources so they accumulate like builtins do in
+    // CairoSteps mode. In SierraGas mode blockifier zeroes resources.opcode_instance_counter, but
+    // builtin_counters (from prover_cairo_primitives) holds the actual flat per-call counts.
```
ddoktorski: After rough checks it doesn't seem to be the case. Are you sure this is true when tracked resource is sierra gas?

ddoktorski (changes requested):
I'm not sure about the current approach, but if we want to go this direction similar changes should be applied to the cheatnet trace, not call info. This way it should also simplify the logic for gas calculation as we mostly rely on the returned call info, so it should be possible to avoid this double counting.

ddoktorski (approved):

ddoktorski:
Does profiler parsing ignore unknown keys or is it going to fail if someone uses a new trace that contains blake with old profiler?

DelevoXDG (approved):

Arcticae:
@ddoktorski there is no validation of what names of builtins are used in the trace, so no extra parsing logic required - i tested it with the same profiler version as before, fix is fully on forge side

PR #4252, opened 2026-04-07
Support for `Option` and `Result` in `--arguments`

### LINKED ISSUE #3976: Investigate `Option` support in `--arguments`
### Current State

Enums are supported in `--arguments` but user reports a problem with using `Option` in data transformer.

### Objective

Check if `Option` is supported correctly in `--arguments` and fix if not

### Additional Context

_No response_

MKowalski8:
Closes #3976

## Introduced changes

- Add support for `Option` and `Result` in data transformer, to allow them to be used with `--arguments` in `sncast`

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,7 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 #### Added
 
 - Aliases `sncast get transaction` and `sncast get transaction-status` for `get tx` and `get tx-status`.
-- Support serializing corelib `Option` and `Result` values passed via `--arguments`
+- Support for serialization of corelib `Option` and `Result`, values passed via `--arguments`
 
 #### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,6 +53,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `sncast get class-hash-at` command to get the class hash of a contract at a given address.
 - `--with-proof-facts` flag for `sncast get tx` to include proof facts in tx response. Read more [here](https://community.starknet.io/t/snip-36-in-protocol-proof-verification/116123l).
 - `--proof-file` and `--proof-facts-file` flags for `sncast invoke` to attach proof data from files.
+- Support serializing corelib `Option` and `Result` values passed via `--arguments`
 
 ## [0.58.1] - 2026-03-31
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 #### Added
 
 - Aliases `sncast get transaction` and `sncast get transaction-status` for `get tx` and `get tx-status`.
+- Support serializing corelib `Option` and `Result` values passed via `--arguments`
 
 #### Changed
 
@@ -53,7 +54,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `sncast get class-hash-at` command to get the class hash of a contract at a given address.
 - `--with-proof-facts` flag for `sncast get tx` to include proof facts in tx response. Read more [here](https://community.starknet.io/t/snip-36-in-protocol-proof-verification/116123l).
 - `--proof-file` and `--proof-facts-file` flags for `sncast invoke` to attach proof data from files.
-- Support serializing corelib `Option` and `Result` values passed via `--arguments`
 
 ## [0.58.1] - 2026-03-31
 
--- a/crates/data-transformer/src/shared/path.rs
+++ b/crates/data-transformer/src/shared/path.rs
@@ -86,5 +86,5 @@ fn extract_generic_args(
         return Err(PathSplitError::MoreThanOneGenericArg);
     };
 
-    Ok((*generic_arg).to_string())
+    Ok(generic_arg.to_string())
 }
--- a/crates/data-transformer/src/shared/path.rs
+++ b/crates/data-transformer/src/shared/path.rs
@@ -86,5 +86,5 @@ fn extract_generic_args(
         return Err(PathSplitError::MoreThanOneGenericArg);
     };
 
-    Ok(generic_arg.to_string())
+    Ok((*generic_arg).to_string())
 }
--- a/crates/data-transformer/src/transformer/sierra_abi/complex_types.rs
+++ b/crates/data-transformer/src/transformer/sierra_abi/complex_types.rs
@@ -6,7 +6,7 @@ use super::{SupportedCalldataKind, build_representation};
 use crate::shared;
 use crate::shared::parsing::parse_expression;
 use crate::shared::path::SplitResult;
-use anyhow::{Context, Result, bail, ensure};
+use anyhow::{Context, Result, anyhow, bail, ensure};
 use cairo_lang_parser::utils::SimpleParserDatabase;
 use cairo_lang_syntax::node::ast::{
     Expr, ExprFunctionCall, ExprListParenthesized, ExprPath, ExprStructCtorCall,
@@ -50,6 +50,8 @@ impl EnumOrStruct for AbiEnum {
 }
 
 const UNIT_TYPE: &str = "()";
+const OPTION_TYPE: &str = "core::option::Option";
+const RESULT_TYPE: &str = "core::result::Result";
 
 fn strip_generic_suffix(type_str: &str) -> &str {
     if let Some(idx) = type_str.find("::<") {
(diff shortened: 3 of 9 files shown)


CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `sncast get class-hash-at` command to get the class hash of a contract at a given address.
 - `--with-proof-facts` flag for `sncast get tx` to include proof facts in tx response. Read more [here](https://community.starknet.io/t/snip-36-in-protocol-proof-verification/116123l).
 - `--proof-file` and `--proof-facts-file` flags for `sncast invoke` to attach proof data from files.
+- Support serializing corelib `Option` and `Result` values passed via `--arguments`
```
ddoktorski: Move it to `Unreleased` section

crates/data-transformer/src/shared/path.rs
```diff
@@ -86,5 +86,5 @@ fn extract_generic_args(
         return Err(PathSplitError::MoreThanOneGenericArg);
     };
 
-    Ok(generic_arg.to_string())
+    Ok((*generic_arg).to_string())
```
ddoktorski: Is it necessary?

crates/data-transformer/src/transformer/sierra_abi/complex_types.rs
```diff
@@ -41,19 +41,142 @@ impl EnumOrStruct for AbiEnum {
...
+        .last()
+        .expect("path_argument must be non-empty: caller ensures split_last() succeeded");
+    if *last != base_type_name(param_type) && path_argument_joined != param_type {
         bail!(r#"Invalid argument type, expected "{param_type}", got "{path_argument_joined}""#,)
     }
     Ok(())
 }
 
+/// A resolved enum variant with its serialization position and optional inner type.
+struct ResolvedEnumVariant<'a> {
```
ddoktorski: nit: Please place struct definitions above under trait definitions

crates/data-transformer/src/transformer/sierra_abi/complex_types.rs
```diff
@@ -41,19 +41,142 @@ impl EnumOrStruct for AbiEnum {
...
+    }
+
+    // core::result::Result::<T, E>  ->  Ok(T) at 0, Err(E) at 1
+    if let Some(inner) = expected_type
+        .strip_prefix("core::result::Result::<")
+        .and_then(|s| s.strip_suffix('>'))
+    {
+        // A well-formed Result type always contains a top-level comma separating T and E.
+        // If absent the ABI entry is malformed; fall through to ABI lookup.
+        let split_pos = top_level_comma_pos(inner)?;
```
ddoktorski: Does it work correctly if `T` type is a tuple?

crates/data-transformer/tests/integration/transformer.rs
```diff
@@ -565,6 +565,432 @@ async fn test_happy_case_implicit_contract_constructor() {
...
+fn test_option_value_variant_missing_value() {
+    let abi = option_u32_abi();
+    let result = transform(
+        "Option::Some",
+        &abi,
+        &get_selector_from_name("option_fn").unwrap(),
+    );
+
+    result.unwrap_err().assert_contains(
+        r#"Variant "Some" of "core::option::Option::<core::integer::u32>" takes no value"#,
```
ddoktorski: nit: I'm not sure if this is accurate message if actual error seems to be missing parentheses

crates/data-transformer/src/transformer/sierra_abi/complex_types.rs
```diff
@@ -41,19 +41,142 @@ impl EnumOrStruct for AbiEnum {
...
+/// A resolved enum variant with its serialization position and optional inner type.
+struct ResolvedEnumVariant<'a> {
+    /// 0-based position used for serialization
+    position: usize,
+    /// Inner type, or `None` for unit variants
+    inner_type: Option<&'a str>,
+}
+
+fn is_valid_corelib_enum_path(type_name: &str, enum_path: &[String]) -> bool {
+    let module = type_name.to_lowercase();
```
ddoktorski: This logic relies strongly on the current corelib paths and for sure it's not universal (e.g. `IsZeroResult` enum is in `core::zeroable` not `core::iszeroresult`). Please verify if we're sure that it never breaks

crates/sncast/tests/e2e/call.rs
```diff
@@ -293,3 +296,106 @@ fn test_json_output_format() {
...
+            json["contract_address"].as_str().map(str::to_string)
+        })
+        .unwrap();
+
+    (tempdir, contract_address)
+}
+
+#[tokio::test]
+async fn test_happy_case_call_with_option() {
+    let (tempdir, contract_address) = deploy_data_transformer_contract().await;
```
ddoktorski, line 347: nit: Both tests deploy this contract, but given that we only do calls the same contract instance could be used

crates/data-transformer/src/transformer/sierra_abi/complex_types.rs
```diff
@@ -41,19 +41,142 @@ impl EnumOrStruct for AbiEnum {
...
+        _ => None,
+    })
+}
+
+fn split_variant_from_path(path: &[String]) -> Result<(String, Vec<String>)> {
+    let (variant, rest) = path
+        .split_last()
+        .ok_or_else(|| anyhow::anyhow!("Expected an enum variant path, got an empty path"))?;
+    Ok((variant.clone(), rest.to_vec()))
+}
```
ddoktorski: nit: We can avoid extra allocation here, requires also a small change on caller sites
```suggestion
fn split_variant_from_path(path: &[String]) -> Result<(&str, &[String])> {
    let (variant, rest) = path
        .split_last()
        .ok_or_else(|| anyhow::anyhow!("Expected an enum variant path, got an empty path"))?;
    Ok((variant, rest))
}
```

crates/data-transformer/src/shared/path.rs
MKowalski8: Linter gave me warning for this, so I fixed that. It doesn't affect how it works, case here is that the double `&&` is sent to `to_string()` function. So I suppose dereferencing it is some kind of performance improvement

crates/data-transformer/tests/integration/transformer.rs
MKowalski8: My bad, that's true error message is wrong. For `Option:Some` it should return message that the values is missing.

crates/data-transformer/src/transformer/sierra_abi/complex_types.rs
MKowalski8: Now it does 😅
https://github.com/foundry-rs/starknet-foundry/pull/4252/changes#diff-e61d6927ea4f4a1a0e33f18405187e192c6eae1837a3826e1f019dd6574c4f36R733

crates/data-transformer/src/shared/path.rs
ddoktorski: I'm checking locally on Rust version `1.94.1` and there is no linter error 🤔

MKowalski8: Here is the warning:

obraz


But this was on version `1.91`, after update to `1.94` it doesn't appear. I'll revert it then, since it looks like it's no longer needed

crates/data-transformer/src/transformer/sierra_abi/complex_types.rs
```diff
@@ -41,19 +49,134 @@ impl EnumOrStruct for AbiEnum {
...
+        .ok_or_else(|| {
+            anyhow::anyhow!(r#"Invalid variant "{variant_name}" for type "{expected_type}""#)
+        })
+}
+
+fn resolve_corelib_enum_variant<'a>(
+    expected_type: &'a str,
+    variant_name: &str,
+    enum_path: &[String],
+) -> Option<Result<ResolvedEnumVariant<'a>>> {
```
franciszekjob: `resolve_corelib_enum_variant` would be clearer as `Result<Option<ResolvedEnumVariant>>` instead of `Option<Result<ResolvedEnumVariant>>`. `None` should mean “not a supported corelib enum, continue with ABI lookup”, while recognized-but-invalid corelib types should return `Err`.

For example, `core::result::Result::<core::integer::u32>` is clearly intended to be `Result`, but it is malformed because it has no top-level comma separating `Ok` and `Err` types. With the current `Option<Result<_>>` shape, that can fall through to ABI lookup as if it was not a corelib type at all.

WDYT?

crates/data-transformer/src/transformer/sierra_abi/complex_types.rs
```diff
@@ -41,19 +49,134 @@ impl EnumOrStruct for AbiEnum {
...
+        let ok_type = inner[..split_pos].trim();
+        let err_type = inner[split_pos + 1..].trim();
+        return Some(resolve_corelib_enum_variant_with(
+            expected_type,
+            "Result",
+            enum_path,
+            variant_name,
+            &[("Ok", 0, Some(ok_type)), ("Err", 1, Some(err_type))],
+        ));
+    }
```
franciszekjob: How about separating corelib type recognition from variant resolution? Returning a small `CorelibEnumSpec` for `Option` / `Result` would let the code parse the expected type once, then call the shared variant resolver once. This removes the duplicated `resolve_corelib_enum_variant_with` calls and makes the extension point clearer if more corelib enums are supported later.

For example, the spec could expose `type_name` and `variants`:

```rust
struct CorelibEnumSpec<'a> {
    type_name: &'static str,
    variants: [(&'static str, usize, Option<&'a str>); 2],
}
```

and then resolution becomes:


```rust
fn resolve_corelib_enum_variant<'a>(
    expected_type: &'a str,
    variant_name: &str,
    enum_path: &[String],
) -> Result<Option<ResolvedEnumVariant<'a>>> {
    let Some(spec) = corelib_enum_spec(expected_type)? else {
        return Ok(None);
    };

    resolve_corelib_enum_variant_with(
        expected_type,
        spec.type_name,
        enum_path,
        variant_name,
        &spec.variants,
    )
    .map(Some)
}
```

WDYT?

franciszekjob (commented):
[supported expressions](https://foundry-rs.github.io/starknet-foundry/starknet/calldata-transformation.html#supported-expressions) docs section should be updated as well.

Overall 👌

crates/data-transformer/src/transformer/sierra_abi/complex_types.rs
MKowalski8: It was, a good change. I tried to do it, however after some logic change and refactoring whole `resolve_corelib_enum_variant_with` function was deleted, but thanks for it anyway. 
Of course if you still see some places for improvement feel free.

crates/data-transformer/src/transformer/sierra_abi/complex_types.rs
MKowalski8: The validation logic was changed a little bit, because it was unneded double check with `validate_path_argument`. Now it only check if the given type equal the expected type. So any corelib change etc. won't affect us. 

However we still depend on `Option` and `Result` paths, e.g `core::option::Option`, but they are passed explicite and it's not likely for them to change.

crates/data-transformer/src/transformer/sierra_abi/complex_types.rs
```diff
@@ -6,6 +6,7 @@ use super::{SupportedCalldataKind, build_representation};
 use crate::shared;
 use crate::shared::parsing::parse_expression;
 use crate::shared::path::SplitResult;
+use anyhow::anyhow;
 use anyhow::{Context, Result, bail, ensure};
```
ddoktorski: nit: double `anyhow` import, but actually why isn't it fix by `cargo fmt`? 🤔

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 #### Fixed
 
 - `sncast verify` now uses the configured network or infers it from the RPC chain ID when `--network` is omitted.
+- Support serializing corelib `Option` and `Result` values passed via `--arguments`
```
ddoktorski: What about interpreting is as a new functionality rather than a bug fix and move it under `Added`?

crates/data-transformer/src/transformer/sierra_abi/complex_types.rs
```diff
@@ -41,19 +50,110 @@ impl EnumOrStruct for AbiEnum {
...
+        .split("::")
+        .last()
+        .unwrap_or(type_str)
+}
+
+fn validate_path_argument(param_type: &str, path_argument_joined: &str) -> Result<()> {
+    let canonical = strip_generic_suffix(param_type);
+
+    if canonical != path_argument_joined
+        && !canonical.ends_with(&format!("::{path_argument_joined}"))
```
ddoktorski, line 75: nit: In a few places comments might be very helpful for the future (example here), short ones explaining what it does and why 🙏

ddoktorski (approved):

crates/data-transformer/src/transformer/sierra_abi/complex_types.rs
MKowalski8: Right, forgot to fix that. 

> but actually why isn't it fix by cargo fmt?

I was thinking the same

CHANGELOG.md
MKowalski8: Yeah, in some way it was expected to work like this, so I'll move it

crates/data-transformer/src/transformer/sierra_abi/complex_types.rs
MKowalski8: Okay, I'll try to think about some then

CHANGELOG.md
```diff
@@ -19,6 +19,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Cast
 
+### Added 
+
+- Support serializing corelib `Option` and `Result` values passed via `--arguments`
```
franciszekjob: nit:
```suggestion
- Support for serialization of corelib `Option` and `Result`, values passed via `--arguments`
```

franciszekjob (approved):

PR #3248, opened 2025-04-18
Support RPC 0.8.1, use auto estimate

cptartur:
Closes #3182 
Closes #3100

## Introduced changes

- Bumped starknet rs and subsequently supported RPC version to 0.8.1
- Restored auto estimations

## Checklist

- [x] Linked relevant issue
- [x] Updated relevant documentation
- [x] Added relevant tests
- [x] Performed self-review of the code
- [x] Added changes to `CHANGELOG.md`

PR #4255, opened 2026-04-09
Add support for SNIP36: `sncast get tx`, `invoke`

DelevoXDG:
Closes #4239 

## Introduced changes

- add `--with-proof-facts` to `sncast get tx`
- allow to pass proof & proof facts to `sncast invoke`
- bump `starknet-rust` to `0.19.0-rc.2`

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/sncast/src/starknet_commands/get/transaction.rs
```diff
@@ -7,24 +7,30 @@ use sncast::response::errors::StarknetCommandError;
...
 #[derive(Debug, Args)]
 #[command(about = "Get the details of a transaction")]
 pub struct Transaction {
+    #[allow(clippy::struct_field_names)]
     /// Hash of the transaction
     pub transaction_hash: Felt,
 
+    /// Include proof facts in transaction response
+    #[arg(long)]
+    pub with_proof_facts: bool,
```
DelevoXDG, line 24: We previously considered always including `IncludeProofFacts` flag.

However that will fail:
- tests with current devnet version
- nodes serving rpc 0.10.0

Considering that, we could either make `IncludeProofFacts` inclusion version-dependent, meaning that flag is included/excluded automatically based on rpc version.

Or we can bump minimal rpc version (we did so in the past https://github.com/foundry-rs/starknet-foundry/pull/3248), but that will make it hard error, which I'm not sure is desired.

Besides, as of now, that will require disabling plenty of tests.

So I suggest, at least for now, to make it optional via `--with-proof-facts`.

Thoughts?

ddoktorski: sgtm

crates/sncast/src/helpers/proof.rs
```diff
@@ -0,0 +1,171 @@
...
+        }
+    }
+    value
+}
+
+impl Default for ProofArgs {
+    fn default() -> Self {
+        Self::none()
+    }
+}
```
ddoktorski: Why not to derive `Default` on `ProofArgs` instead?

crates/sncast/src/starknet_commands/invoke.rs
```diff
@@ -104,6 +119,13 @@ where
         tip,
     } = fee_settings.expect("Failed to convert to fee settings");
 
+    let proof = proof_args
+        .resolve_proof()
+        .map_err(|e| anyhow!("Failed to resolve proof: {e}"))?;
+    let proof_facts = proof_args
+        .resolve_proof_facts()
+        .map_err(|e| anyhow!("Failed to resolve proof facts: {e}"))?;
```
ddoktorski:
```suggestion
    let proof = proof_args
        .resolve_proof()
        .context("Failed to resolve proof")?;
    let proof_facts = proof_args
        .resolve_proof_facts()
        .context("Failed to resolve proof facts")?;
```

crates/sncast/src/helpers/proof.rs
```diff
@@ -0,0 +1,171 @@
...
+
+#[cfg(test)]
+mod tests {
```
ddoktorski, line 80: Can we also add a test for an empty file?

crates/sncast/src/helpers/proof.rs
```diff
@@ -0,0 +1,171 @@
...
+        match &self.proof_facts_file {
+            Some(path) => {
+                let contents = std::fs::read_to_string(path)
+                    .with_context(|| format!("Failed to read proof facts file at {path}"))?;
+                let felts = contents
+                    .split(',')
+                    .map(|s| {
+                        strip_quotes(s.trim())
+                            .parse::<Felt>()
+                            .with_context(|| format!("Failed to parse felt from '{}'", s.trim()))
```
ddoktorski: nit: Currently error message displays the initial value instead of the stripped one which was used in parsing, this should be a quick fix:
```suggestion
                        let stripped = strip_quotes(s.trim());
                        stripped
                            .parse::<Felt>()
                            .with_context(|| format!("Failed to parse felt from '{stripped}'"))
```

crates/sncast/src/helpers/proof.rs
```diff
@@ -0,0 +1,171 @@
...
+    if value.len() >= 2 {
+        let bytes = value.as_bytes();
+        let first = bytes[0];
+        let last = bytes[value.len() - 1];
+        if (first == b'"' && last == b'"') || (first == b'\'' && last == b'\'') {
+            return &value[1..value.len() - 1];
+        }
+    }
+    value
+}
```
franciszekjob, line 77: nit: We can avoid manual indexing


```suggestion
fn strip_quotes(value: &str) -> &str {
    for quote in ['"', '\''] {
        if let Some(stripped) = value
            .strip_prefix(quote)
            .and_then(|s| s.strip_suffix(quote))
        {
            return stripped;
        }
    }
    value
}
```

docs/src/appendix/sncast/invoke.md
```diff
@@ -98,3 +98,15 @@ If passed, an estimated tip will be added to pay for the transaction. The tip is
 Optional.
 
 Nonce for transaction. If not provided, nonce will be set automatically.
+
+## `--proof-file <PROOF_FILE>`
+Optional.
+Requires: [`--proof-facts-file`](#--proof-facts-file-proof_facts_file)
+
+Path to a file containing a base64-encoded proof string attached to the invoke transaction.
```
franciszekjob: nit: (same below)
```suggestion
Path to a plain text file containing a base64-encoded proof string attached to the invoke transaction.
```

franciszekjob (commented):
Maybe it's worth extending our docs (ofc as todo, not in this PR) with full dedicated example for submitting tx with proof?

crates/sncast/src/helpers/proof.rs
```diff
@@ -0,0 +1,210 @@
...
+    pub proof_facts_file: Option<Utf8PathBuf>,
+}
+
+impl ProofArgs {
+    #[must_use]
+    pub fn none() -> Self {
+        Self {
+            proof_file: None,
+            proof_facts_file: None,
+        }
```
ddoktorski, line 23: nit: It's redundant, we can use `ProofArgs::default()` instead

ddoktorski (approved):

franciszekjob (approved):

crates/sncast/src/helpers/proof.rs
DelevoXDG: We can, but I wanted to be more explicit, `ProofArgs::none()` is searchable at call site, unlike `Default::default()`, and is clear wrt. what defaults are.

Thoughts?

PR #4263, opened 2026-04-10
Remove `argent` type

### LINKED ISSUE #3556: Remove `--argent` option

MKowalski8:
Closes #3556 

## Introduced changes

- Remove the depracated `argent` type

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `--with-proof-facts` flag for `sncast get tx` to include proof facts in tx response. Read more [here](https://community.starknet.io/t/snip-36-in-protocol-proof-verification/116123l).
 - `--proof-file` and `--proof-facts-file` flags for `sncast invoke` to attach proof data from files.
 
+#### Removed
+
+- `argent` option for `--type` flag in `account create` and `account import` commands. Use `ready` instead. Old account files with `"type": "argent"` are still loaded correctly as `ready`.
+
 ## [0.58.1] - 2026-03-31
 
 ### Forge
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - `sncast completions` no longer fails due to invalid local or global `snfoundry.toml` config.
 
+### Cast
+
+#### Removed
+
+- `argent` option for `--type` flag in `account create` and `account import` commands. Use `ready` instead. Old account files with `"type": "argent"` are still loaded correctly as `ready`.
+
 ## [0.60.0] - 2026-04-27
 
 ### Forge
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,11 +24,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - `sncast completions` no longer fails due to invalid local or global `snfoundry.toml` config.
 
-### Cast
-
 #### Removed
 
-- `argent` option for `--type` flag in `account create` and `account import` commands. Use `ready` instead. Old account files with `"type": "argent"` are still loaded correctly as `ready`.
+- `argent` option for `--type` flag in `account create` and `account import` commands. Use `ready` instead. Old account files with `"type": "argent"` are still loaded correctly.
 
 ## [0.60.0] - 2026-04-27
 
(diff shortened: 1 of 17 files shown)


docs/src/starknet/account-import.md
```diff
@@ -13,7 +13,7 @@ This section shows how to export your private key from specific wallets.
 
 ### Examples
 
-#### Ready (formerly Argent)
```
ddoktorski, line 16: I think we can keep this comment

CHANGELOG.md
```diff
@@ -40,6 +40,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `sncast verify` now uses the configured network or infers it from the RPC chain ID when `--network` is omitted.
 - `sncast` now returns non-zero exit code when a command fails.
 
+#### Removed
+
+- `argent` option for `--type` flag in `account create` and `account import` commands. Use `ready` instead. Old account files with `"type": "argent"` are still loaded correctly as `ready`.
+
```
ddoktorski: ⬆️ ⬆️ ⬆️

crates/sncast/src/lib.rs
```diff
@@ -72,7 +72,6 @@ impl FromStr for AccountType {
     fn from_str(s: &str) -> Result<Self, Self::Err> {
         match s {
             "open_zeppelin" | "open-zeppelin" | "oz" => Ok(AccountType::OpenZeppelin),
-            "argent" => Ok(AccountType::Argent),
```
ddoktorski, line 78: Isn't it needed for backwards compatibility in keystore files?

franciszekjob: It seems that the change above

```rust
#[serde(alias = "argent")] // backward compatibility with pre-rebranding account files
```

should guarantee that. However I checked now and it does not 🤔 


Image


Btw, I think we should keep a test for that.

MKowalski8: The alias is for deserialization from `account_file` only. If we keep it with `from_str` we could still do:
`sncast account create --type argent` which we don't want

MKowalski8: > Isn't it needed for backwards compatibility in keystore files?

`from_str` is only used for `--type` argument, so I don't think so. I do not checked it with keystore per se, I can do that, but I don't see what could break there

CHANGELOG.md
MKowalski8: Yop, I haven't rebased yet

crates/sncast/src/lib.rs
franciszekjob: Ah, you're right, thanks for clarification.

MKowalski8: > Btw, I think we should keep a test for that.

There is the test `test_happy_case_with_legacy_argent_account_type` which check this

franciszekjob (approved):

crates/sncast/tests/e2e/invoke.rs
```diff
@@ -94,6 +94,45 @@ async fn test_happy_case(class_hash: Felt, account_type: AccountType) {
...
+        "put",
+        "--calldata",
+        "0x1 0x2",
+    ];
+
+    let snapbox = runner(&args).current_dir(tempdir.path());
+    let output = snapbox.assert().success();
+    let stdout = output.get_output().stdout.clone();
+
+    let hash = get_transaction_hash(&stdout);
```
ddoktorski: nit:
```suggestion
    let stdout = &output.get_output().stdout;

    let hash = get_transaction_hash(stdout);
```

crates/sncast/src/lib.rs
```diff
@@ -536,10 +535,15 @@ pub fn get_account_data_from_keystore(
         .pointer("/variant/legacy")
         .and_then(Value::as_bool);
     let account_type = get_string_value_from_json(&account_info, "/variant/type")
+        // "argent" was renamed to "ready"; map for backward compat with old account files
+        .map(|t| match t.as_str() {
```
MKowalski8, line 539: note: I am not a big fan of this solution. But making extra parser for this is not better imo. 
To avoid removing it by mistake I added a test

CHANGELOG.md
```diff
@@ -13,6 +13,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Bug in forge debugging and `--gas-report` that caused panic in case of a call to non-existent selector
 
+### Cast
+
+#### Removed
+
+- `argent` option for `--type` flag in `account create` and `account import` commands. Use `ready` instead. Old account files with `"type": "argent"` are still loaded correctly as `ready`.
```
ddoktorski: nit:
```suggestion
- `argent` option for `--type` flag in `account create` and `account import` commands. Use `ready` instead. Old account files with `"type": "argent"` are still loaded correctly.
```

ddoktorski (approved):

PR #4270, opened 2026-04-13
Fix sncast verify network selection without --network

ametel01:
## Summary
- remove the hardcoded Sepolia fallback in `sncast verify`
- resolve the verifier network in this order: CLI `--network`, `snfoundry.toml` `network`, then RPC `chain_id`
- update Voyager verify E2E coverage to reflect the new inference path and cover config-driven network selection

## Bugs fixed
- passing `--url <mainnet-rpc>` without `--network` no longer submits verification to Sepolia by default
- setting `network` in `snfoundry.toml` no longer gets ignored when CLI `--network` is omitted

## Implementation details
- `verify` now builds the provider first and only infers the network from `chain_id` when both CLI and config networks are absent
- if the RPC chain ID is not recognized as mainnet or sepolia, `verify` now returns an explicit error instead of silently assuming Sepolia
- added unit tests for network precedence and inference behavior
- updated the existing no-`--network` Voyager E2E case to mock `starknet_chainId`
- added an E2E test that verifies `config.network` is honored without performing `chainId` inference

## Testing
- `cargo fmt --all --check`
- `cargo test -p sncast --bin sncast starknet_commands::verify::tests:: -- --nocapture`
- `cargo test -p sncast --test main verify:: -- --nocapture --test-threads=1`

### DIFF
--- a/crates/sncast/src/starknet_commands/verify/mod.rs
+++ b/crates/sncast/src/starknet_commands/verify/mod.rs
@@ -167,7 +167,7 @@ pub async fn verify(
     };
     let provider = get_provider(&rpc_url)?;
     let inferred_network = if network.is_none() && config.network.is_none() {
-        Some(Network::try_from(get_chain_id(&provider).await?)?)
+        Network::try_from(get_chain_id(&provider).await?).ok()
     } else {
         None
     };
@@ -259,7 +259,7 @@ pub async fn verify(
 #[cfg(test)]
 mod tests {
     use super::resolve_verification_network;
-    use sncast::{MAINNET, Network, SEPOLIA};
+    use sncast::Network;
 
     #[test]
     fn uses_cli_network_when_provided() {
@@ -284,18 +284,14 @@ mod tests {
 
     #[test]
     fn infers_mainnet_from_chain_id_when_no_network_is_configured() {
-        let network =
-            resolve_verification_network(None, None, Some(Network::try_from(MAINNET).unwrap()))
-                .unwrap();
+        let network = resolve_verification_network(None, None, Some(Network::Mainnet)).unwrap();
 
         assert_eq!(network, Network::Mainnet);
     }
 
     #[test]
     fn infers_sepolia_from_chain_id_when_no_network_is_configured() {
-        let network =
-            resolve_verification_network(None, None, Some(Network::try_from(SEPOLIA).unwrap()))
-                .unwrap();
+        let network = resolve_verification_network(None, None, Some(Network::Sepolia)).unwrap();
 
         assert_eq!(network, Network::Sepolia);
     }
--- a/crates/sncast/src/starknet_commands/verify/mod.rs
+++ b/crates/sncast/src/starknet_commands/verify/mod.rs
@@ -7,6 +7,7 @@ use sncast::helpers::rpc::FreeProvider;
 use sncast::response::ui::UI;
 use sncast::{Network, response::verify::VerifyResponse};
 use sncast::{get_chain_id, get_provider};
+use starknet_rust::providers::jsonrpc::{HttpTransport, JsonRpcClient};
 use starknet_types_core::felt::Felt;
 use std::{collections::HashMap, fmt};
 use url::Url;
@@ -82,17 +83,20 @@ impl fmt::Display for Verifier {
     }
 }
 
-fn resolve_verification_network(
+async fn resolve_verification_network(
     cli_network: Option<Network>,
     config_network: Option<Network>,
-    inferred_network: Option<Network>,
+    provider: &JsonRpcClient<HttpTransport>,
 ) -> Result<Network> {
-    cli_network
-        .or(config_network)
-        .or(inferred_network)
-        .ok_or_else(|| {
+    if let Some(network) = cli_network.or(config_network) {
+        return Ok(network);
+    }
+
+    let chain_id = get_chain_id(provider).await?;
+
+    Network::try_from(chain_id).map_err(|_| {
             anyhow!(
-                "Failed to infer verification network from the RPC chain ID; pass `--network mainnet` or `--network sepolia` explicitly"
+                "Failed to infer verification network from the RPC chain ID {chain_id:#x}; pass `--network mainnet` or `--network sepolia` explicitly"
             )
         })
 }
@@ -166,11 +170,6 @@ pub async fn verify(
         }
     };
     let provider = get_provider(&rpc_url)?;
-    let inferred_network = if network.is_none() && config.network.is_none() {
-        Network::try_from(get_chain_id(&provider).await?).ok()
-    } else {
-        None
-    };
 
     // Build JSON Payload for the verification request
     // get the parent dir of the manifest path
@@ -191,7 +190,7 @@ pub async fn verify(
         }
     };
 
-    let network = resolve_verification_network(network, config.network, inferred_network)?;
+    let network = resolve_verification_network(network, config.network, &provider).await?;
 
     // Handle test_files warning for Walnut
     if matches!(verifier, Verifier::Walnut) && test_files {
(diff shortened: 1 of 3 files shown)


crates/sncast/src/starknet_commands/verify/mod.rs
```diff
@@ -152,6 +166,11 @@ pub async fn verify(
         }
     };
     let provider = get_provider(&rpc_url)?;
+    let inferred_network = if network.is_none() && config.network.is_none() {
+        Some(Network::try_from(get_chain_id(&provider).await?)?)
```
ddoktorski: Let `resolve_verification_network` function to handle error scenario
```suggestion
        Network::try_from(get_chain_id(&provider).await?).ok()
```

crates/sncast/src/starknet_commands/verify/mod.rs
```diff
@@ -244,3 +255,59 @@ pub async fn verify(
...
+            resolve_verification_network(None, Some(Network::Mainnet), Some(Network::Sepolia))
+                .unwrap();
+
+        assert_eq!(network, Network::Mainnet);
+    }
+
+    #[test]
+    fn infers_mainnet_from_chain_id_when_no_network_is_configured() {
+        let network =
+            resolve_verification_network(None, None, Some(Network::try_from(MAINNET).unwrap()))
```
ddoktorski:
```suggestion
            resolve_verification_network(None, None, Some(Network::Mainnet))
```

crates/sncast/src/starknet_commands/verify/mod.rs
```diff
@@ -244,3 +255,59 @@ pub async fn verify(
...
+                .unwrap();
+
+        assert_eq!(network, Network::Mainnet);
+    }
+
+    #[test]
+    fn infers_sepolia_from_chain_id_when_no_network_is_configured() {
+        let network =
+            resolve_verification_network(None, None, Some(Network::try_from(SEPOLIA).unwrap()))
```
ddoktorski:
```suggestion
            resolve_verification_network(None, None, Some(Network::Sepolia)))
```

crates/sncast/tests/e2e/verify/voyager.rs
ddoktorski, line 1: Can we also add a test to check flow for unrecognized chain id and missing `--network` flag?

crates/sncast/src/starknet_commands/verify/mod.rs
ametel01: Applied in 9510bcb37b22973504916889a87b9ea32396027a: simplified the test to pass `Some(Network::Mainnet)` directly.

crates/sncast/tests/e2e/verify/voyager.rs
ametel01: Added in 9510bcb37b22973504916889a87b9ea32396027a: an e2e test now covers an unrecognized RPC chain ID with no `--network` and asserts the inference error without calling Voyager.

crates/sncast/src/starknet_commands/verify/mod.rs
ametel01: Applied in 9510bcb37b22973504916889a87b9ea32396027a: simplified the test to pass `Some(Network::Sepolia)` directly.

crates/sncast/src/starknet_commands/verify/mod.rs
ametel01: Applied in 9510bcb37b22973504916889a87b9ea32396027a: unrecognized chain IDs now map to `None`, so `resolve_verification_network` handles the explicit missing-network error.

ametel01:
CI update: the current failures do not appear related to the verify changes in this PR.

- `Test Cast` failed before running tests during `asdf-vm/actions/install` with `Unable to locate executable file: asdf`.
- `Test Ledger` failed the same way, also before running tests.
- `Test Building Docs` failed during mdbook linkcheck because external `https://www.starknet.io/cairo-book/...` links returned `415 Unsupported Media Type`.

The relevant code-path checks I ran locally passed: `cargo fmt --check`, `git diff --check`, and `cargo test -p sncast --bin sncast starknet_commands::verify::tests`. The e2e target compiled, but this local environment could not run it because the suite-level devnet bootstrap timed out before filtered tests executed.

ametel01:
Additional context on why this passed yesterday: the previous successful CI run was on 2026-04-13 at `d6be40d7...`, and the failing jobs were green there (`Test Cast`, `Test Ledger`, `Test Building Docs`). The runner image was the same (`ubuntu-24.04`, image `20260406.80.1`).

For `Test Cast`/`Test Ledger`, the pinned `asdf-vm/actions/install@05e0d2...` action still clones `asdf` from the mutable `master` branch. In yesterday’s log it successfully executed `/home/runner/.asdf/bin/asdf`; in today’s log it fails immediately after cloning with `Unable to locate executable file: asdf`. That points to an upstream/asdf action setup issue, not this PR.

For docs, yesterday mdbook linkcheck completed; today it fails on external `www.starknet.io/cairo-book` links returning `415 Unsupported Media Type`. That also points to an external website/linkcheck response change, not this PR.

ddoktorski (approved):
Please also add an entry about this fix to the changelog 🙏 (under the sections `Unreleased` -> `Cast` -> `Fixed`)

ametel01:
Done, added the changelog entry under `Unreleased` -> `Cast` -> `Fixed` in 9968a38f.

crates/sncast/src/starknet_commands/verify/mod.rs
```diff
@@ -152,6 +166,11 @@ pub async fn verify(
         }
     };
     let provider = get_provider(&rpc_url)?;
+    let inferred_network = if network.is_none() && config.network.is_none() {
+        Network::try_from(get_chain_id(&provider).await?).ok()
+    } else {
+        None
+    };
```
DelevoXDG: We're currently duplicating the logic, and checking cli and config network twice. 

We can pass `provider` to `resolve_verification_network` instead, resole network from chain ID there.

crates/sncast/src/starknet_commands/verify/mod.rs
```diff
@@ -152,6 +166,11 @@ pub async fn verify(
         }
     };
     let provider = get_provider(&rpc_url)?;
+    let inferred_network = if network.is_none() && config.network.is_none() {
+        Network::try_from(get_chain_id(&provider).await?).ok()
```
DelevoXDG: Not sure it's optimal that we're swallowing error here.

If we move it as I suggested in the other comment https://github.com/foundry-rs/starknet-foundry/pull/4270/changes#r3085704752, I think we should be able to properly propagate it and add `context()`

crates/sncast/src/starknet_commands/verify/mod.rs
```diff
@@ -82,6 +82,21 @@ impl fmt::Display for Verifier {
...
+    config_network: Option<Network>,
+    inferred_network: Option<Network>,
+) -> Result<Network> {
+    cli_network
+        .or(config_network)
+        .or(inferred_network)
+        .ok_or_else(|| {
+            anyhow!(
+                "Failed to infer verification network from the RPC chain ID; pass `--network mainnet` or `--network sepolia` explicitly"
+            )
```
DelevoXDG, line 100: nit:
```suggestion
            anyhow!(
                "Failed to infer verification network from the RPC chain ID {chain_id:#x}; pass `--network mainnet` or `--network sepolia` explicitly"
            )
```

crates/sncast/src/starknet_commands/verify/mod.rs
ametel01: Applied in 25057aa3e: `resolve_verification_network` now takes the provider, returns the CLI/config network early, and fetches/maps the chain ID internally only when neither is configured.

crates/sncast/src/starknet_commands/verify/mod.rs
ametel01: Applied in 25057aa3e: the inference error now includes the formatted RPC chain ID, e.g. `0x1234`, before asking users to pass `--network` explicitly.

DelevoXDG (approved):

PR #4272, opened 2026-04-13
Skip config loading for `sncast completions`

DelevoXDG:
Closes #4233 

## Introduced changes

-

## Checklist

- [ ] Linked relevant issue
- [ ] Added relevant tests

### DIFF
--- a/crates/sncast/src/main.rs
+++ b/crates/sncast/src/main.rs
@@ -735,9 +735,9 @@ async fn run_async_command(cli: Cli, config: CastConfig, ui: &UI) -> Result<()> 
         }
 
         Commands::Ledger(ledger) => {
-            let result = ledger::ledger(&ledger, ui).await?;
+            let result = ledger::ledger(&ledger, ui).await;
 
-            process_command_result("ledger", Ok(result), ui, None);
+            process_command_result("ledger", result, ui, None);
 
             Ok(())
         }
--- a/crates/sncast/src/main.rs
+++ b/crates/sncast/src/main.rs
@@ -512,13 +512,12 @@ async fn run_async_command(cli: Cli, config: CastConfig, ui: &UI) -> Result<Exit
                     Err(err) => {
                         // TODO(#3960) This will return json output saying that `deploy` command was run
                         //  even though the invoked command was declare.
-                        let exit_code = process_command_result::<DeclareTransactionResponse>(
+                        return Ok(process_command_result::<DeclareTransactionResponse>(
                             "deploy",
                             Err(err),
                             ui,
                             None,
-                        );
-                        return Ok(exit_code);
+                        ));
                     }
                 }
             } else {
--- a/crates/sncast/src/main.rs
+++ b/crates/sncast/src/main.rs
@@ -433,7 +433,12 @@ async fn run_async_command(cli: Cli, config: CastConfig, ui: &UI) -> Result<Exit
 
             let block_explorer_link =
                 block_explorer_link_if_allowed(&result, provider.chain_id().await?, &config).await;
-            Ok(process_command_result("declare-from", result, ui, block_explorer_link))
+            Ok(process_command_result(
+                "declare-from",
+                result,
+                ui,
+                block_explorer_link,
+            ))
         }
 
         Commands::Deploy(deploy) => {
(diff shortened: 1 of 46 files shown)


DelevoXDG:
Awaiting #4205 merge to add test

crates/sncast/src/main.rs
```diff
@@ -319,14 +319,21 @@ fn main() -> Result<ExitCode> {
...
     } else {
         let config = get_cast_config(&cli, &ui)?;
         runtime.block_on(run_async_command(cli, config, &ui))
     }
 }
 
+fn run_completions_command(completions: &Completions) -> Result<ExitCode> {
+    generate_completions(completions.shell, &mut Cli::command())?;
+    Ok(ExitCode::SUCCESS)
+}
```
MKowalski8: Is it worth making new func for this?

crates/sncast/src/main.rs
```diff
@@ -786,7 +788,7 @@ async fn run_async_command(cli: Cli, config: CastConfig, ui: &UI) -> Result<Exit
             Ok(process_command_result("ledger", result, ui, None))
         }
 
-        Commands::Script(_) => unreachable!(),
+        Commands::Completions(_) | Commands::Script(_) => unreachable!(),
```
MKowalski8: nit: I think we could add some message in `unreachable!()` just out of principle

crates/sncast/tests/e2e/completions.rs
```diff
@@ -38,3 +40,28 @@ fn test_generate_completions_unsupported_shell() {
...
 }
+
+#[test]
+fn test_completions_invalid_local_config() {
+    let t = copy_config_to_tempdir("tests/data/files/snfoundry_malformed.toml", None);
+    let args = vec!["completions", "bash"];
+
+    let snapbox = runner(&args).current_dir(t.path());
+
+    snapbox.assert().success();
```
MKowalski8: nit: We could check if there is actually any output (like `#compdef sncast` or sth like this).  . Since `success()` doesn't always mean everyting was in fact correct 😅. S

franciszekjob (approved):
LGTM (after applying/resolving comments from Maks)

crates/sncast/src/main.rs
DelevoXDG: Why not? 

Makes this very readable and consistent
```rs
    if let Commands::Completions(completions) = &cli.command {
        run_completions_command(completions)
    } else if let Commands::Script(script) = &cli.command {
        run_script_command(&cli, runtime, script, &ui)
    } else {
        let config = get_cast_config(&cli, &ui)?;
        runtime.block_on(run_async_command(cli, config, &ui))
    }
```

Personally, I would rewrite all main.rs to use functions for commands 🤔

crates/sncast/src/main.rs
DelevoXDG: Added

crates/sncast/tests/e2e/completions.rs
DelevoXDG: Makes sense, added

crates/sncast/src/main.rs
MKowalski8: > Why not?

Our `main.rs` is just to big. And making extra functions that could be easly inlined is not a way to go in my opinion. 

> Makes this very readable and consistent

Imo this is not less readable

``` rust 
    generate_completions(completions.shell, &mut Cli::command())?;
    Ok(ExitCode::SUCCESS)
```

The other way to go, could be returing this `ExitCode` from `generate_completions`.

> Personally, I would rewrite all main.rs to use functions for commands 🤔

Mayby it's okay, refactor is much needed here, but we should focus on keeping this file as simple as possible for now.

DelevoXDG: > > Makes this very readable and consistent
> 
> Imo this is not less readable
> 
>     generate_completions(completions.shell, &mut Cli::command())?;
>     Ok(ExitCode::SUCCESS)

But is less consistent 🤔  
We have `run_script_command` and `run_async_command` just below.

> The other way to go, could be returing this `ExitCode` from `generate_completions`.

This variation is also ok to me, updated

> Our `main.rs` is just to big. And making extra functions that could be easly inlined is not a way to go in my opinion.

However, note that we're talking about 4 LOC difference (3 if we don't count `}`), which seems negligible when compared to regular formatted one-liners:
```rs
            Ok(process_command_result(
                "declare-from",
                result,
                ui,
                block_explorer_link,
            ))
```

```rs
            let result = with_account!(&account, |account| {
                starknet_commands::declare::declare(
                    declare.contract_name.clone(),
                    declare.common.fee_args,
                    declare.common.dry_run_args,
                    declare.common.nonce,
                    account,
                    &artifacts,
                    wait_config,
                    false,
                    ui,
                )
                .await
            });
```

> Mayby it's okay, refactor is much needed here, but we should focus on keeping this file as simple as possible for now.

I've added https://github.com/foundry-rs/starknet-foundry/issues/4348 so we can address in this future.

Considering what how small the difference is though, I disagree that our focus should be addressing 4 LOC difference in almost 900 LOC file - IMO this is a bit out of scope and shouldn't be the blocker for this specific PR

MKowalski8 (approved):

PR #4280, opened 2026-04-14
Run most workspace tests in one job on CI

ddoktorski:
- To limit number of used runners, run most workspace tests within a one job instead of spawning a separate runner for each crate
- Run checks: formatting, lint, typos in one job

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -354,11 +354,33 @@ jobs:
       - name: Run tests
         run: cargo test --profile ci -p sncast --features ledger-emulator ledger -- --ignored
 
-  scarbfmt:
+  fmt-lint-typos:
     runs-on: ubuntu-latest
+    env:
+      # Make sure CI fails on all warnings - including Clippy lints.
+      RUSTFLAGS: "-Dwarnings"
     steps:
       - uses: actions/checkout@v6
-      - uses: software-mansion/setup-scarb@v1
+      - uses: ./.github/actions/setup-tools
+        with:
+          install-llvm: 'true'
+          setup-usc: 'false'
+      - run: rustup component add rustfmt
+
+      - name: Check formatting
+        run: cargo fmt --check
+
+      - name: Check formatting snforge-scarb-plugin
+        run: cargo fmt --check
+        working-directory: crates/snforge-scarb-plugin
+
+      - name: Lint
+        run: cargo lint
+
+      - name: Lint snforge-scarb-plugin
+        run: cargo lint
+        working-directory: crates/snforge-scarb-plugin
+
       - name: Check cairo files format
         run: |
           output=$(find . -type f -name "Scarb.toml" -execdir sh -c '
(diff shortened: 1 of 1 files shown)


piotmag769 (approved):
Worth to but scarbfmt, cargo fmt, clippy and typos in the same job as well

ddoktorski:
> Worth to but scarbfmt, cargo fmt, clippy and typos in the same job as well

Added: https://github.com/foundry-rs/starknet-foundry/pull/4280/commits/dce4f227a4be7beda3518c0a920cdd40d64b910f

.github/workflows/ci.yml
```diff
@@ -34,12 +34,16 @@ jobs:
           echo ${scarb_versions[@]}
           echo "versions=[${scarb_versions[@]}]" >> "$GITHUB_OUTPUT"
 
-  test-forge-unit:
-    name: Test Forge / Unit Tests
+  test-workspace:
+    # Runs tests from all crates across the workspace, except:
+    # - `sncast` tests
+    # - `forge` integration and e2e tests
```
DelevoXDG, line 40: Should we also mention `native-api` there? I assume the assumption here is that there are no tests for `native-api` and we're saving on build time?

.github/workflows/ci.yml
```diff
@@ -376,42 +354,33 @@ jobs:
       - name: Run tests
         run: cargo test --profile ci -p sncast --features ledger-emulator ledger -- --ignored
 
-  test-conversions:
-    name: Test Conversions
+  fmt-lint-typos:
```
DelevoXDG, line 358: Nit: 

every other job has a name, perhaps add one for this one as well? 🤔 

Perhaps `Lint and Format`?

.github/workflows/ci.yml
```diff
@@ -376,42 +354,33 @@ jobs:
...
     steps:
       - uses: actions/checkout@v6
       - uses: ./.github/actions/setup-tools
         with:
-          setup-scarb: 'false'
+          install-llvm: 'true'
           setup-usc: 'false'
-      - name: Run tests
-        run: cargo test --profile ci -p conversions
+      - run: rustup component add rustfmt
```
DelevoXDG, line 370: Should we perhaps move this to `./.github/actions/setup-tools`? 

I'm not sure if doing it this way impacts caching? (since with current approach we set up cache *before* adding the component)

.github/workflows/ci.yml
ddoktorski: > I assume the assumption here is that there are no tests

That's correct, but it's not about saving build time - `native-api` requires additional `llvm` setup that we don't install it here. I can extend this comment if you think it might be helpful

.github/workflows/ci.yml
ddoktorski: Not all, `build-test-forge-nextes-*` also use the default one, still we can change it

.github/workflows/ci.yml
ddoktorski: It doesn't modify neither `Cargo.lock` nor artifacts in `target` so ti shouldn't affect caching. Other thing is that caching logic is currently highly inefficient, I'm going to address it in a separate PR

franciszekjob (approved):

.github/workflows/ci.yml
DelevoXDG: Yes please 🙏🏻

.github/workflows/ci.yml
DelevoXDG: Yeah, I'd still add it

DelevoXDG (approved):

.github/workflows/ci.yml
ddoktorski: [869e994](https://github.com/foundry-rs/starknet-foundry/pull/4280/commits/869e994abb8f2ecf00e8be7bede39b55f205e22e)

.github/workflows/ci.yml
ddoktorski: [bb51e04](https://github.com/foundry-rs/starknet-foundry/pull/4280/commits/bb51e0461ff3a2d89f7957cc32e1d57eaa821938)

PR #4284, opened 2026-04-16
Use Voyager verifier library in sncast

ametel01:
## Summary
- replace custom Voyager verification calls with the voyager-verifier client library
- add the new sncast dependency and update the lockfile

## Testing
- cargo test -p sncast --bin sncast starknet_commands::verify::tests:: -- --nocapture
- cargo test -p sncast --test main verify:: -- --nocapture --test-threads=1


crates/sncast/src/starknet_commands/verify/voyager.rs
```diff
@@ -288,108 +79,43 @@ impl<'a> VerificationInterface<'a> for Voyager<'a> {
...
+        let api_base_url = Url::parse(&explorer_url)?;
+        let class_hash = format!("{class_hash:#066x}");
+        let job_id =
+            submit_verification_request(&api_base_url, &class_hash, &prepared.request).await?;
+        let status_url = format!(
+            "{}/{}/{}",
+            explorer_url.trim_end_matches('/'),
+            STATUS_ENDPOINT,
+            job_id,
+        );
```
ddoktorski, line 105: Would it be possible to use [get_job_status_url](https://github.com/NethermindEth/voyager-verifier/blob/4054534e12cf36bdce6eff52ec964b9dad826653/crates/voyager-verifier/src/api/client.rs#L227) instead?

ddoktorski (approved):
Thanks @ametel01

crates/sncast/src/starknet_commands/verify/voyager.rs
ametel01: Updated in 39aa1a1e. The status URL now comes from `voyager_verifier::api::ApiClient::get_job_status_url` instead of manual string formatting, and I added an e2e assertion covering the emitted link.

ddoktorski:
@ametel01 `e2e::verify::voyager` tests are failing now. Could you check this?

franciszekjob (approved):

PR #4290, opened 2026-04-17
Fix compiler config validation

### LINKED ISSUE #109: Verify Scarb.toml compiler settings in snforge when launching debugger

### LINKED ISSUE #4254: Config validation for coverage, backtrace and debugger is faulty
1. Doesn't take into account raw `[cairo]` section - only `[profile.{current_profile}.cairo]`.
2. Doesn't take into account new field aliases (without `unstable-` prefixes) - needs Scarb version check to see if this Scarb version provides such aliases.

### LINKED ISSUE #4292: Add tests for cairo config validation fix
Should also cover scarb versions before 2.15.0 
#4290

piotmag769:
Closes #4254
Closes #4292 
Closes https://github.com/software-mansion/cairo-debugger/issues/109

## Introduced changes

- previously the config validation relied on checking the content of the Scarb.toml, which led to poor UX (e.g. `[cairo]` section was ignored completely). Now we read the config from scarb metadata which is more reliable. Also, previously a wrong Scarb.toml was used - the `runtime_manifest` instead of `workspace.manifest_path`

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Minimal recommended `Scarb` version is now `2.15.2` (updated from `2.14.0`)
 
+#### Fixed
+
+- Validation of Scarb's compiler config now properly takes into account `[cairo]` section and keys without `unstable-` prefix
+
 ### Cast
 
 #### Added
--- a/crates/forge/src/profile_validation/backtrace.rs
+++ b/crates/forge/src/profile_validation/backtrace.rs
@@ -1,20 +1,20 @@
 use crate::TestArgs;
-use crate::profile_validation::{check_cairo_profile_entries, get_manifest};
+use crate::profile_validation::{bool_field, bool_field_or_unstable};
 use anyhow::ensure;
+use camino::Utf8PathBuf;
 use indoc::formatdoc;
-use scarb_metadata::Metadata;
+use serde_json::Value;
 
-/// Checks if backtrace can be generated based on scarb version, profile settings extracted from
-/// the provided [`Metadata`] and if native execution is disabled in the provided [`TestArgs`].
 #[allow(unused_variables)]
 pub fn check_backtrace_compatibility(
     test_args: &TestArgs,
-    scarb_metadata: &Metadata,
+    compiler_config: &Value,
+    profile: &str,
+    workspace_manifest_path: &Utf8PathBuf,
 ) -> anyhow::Result<()> {
     #[cfg(feature = "cairo-native")]
     check_if_native_disabled(test_args)?;
-    check_profile(scarb_metadata)?;
-    Ok(())
+    check_profile(compiler_config, profile, workspace_manifest_path)
 }
 
 /// Checks if native execution is disabled in the provided [`TestArgs`].
@@ -27,23 +27,20 @@ fn check_if_native_disabled(test_args: &TestArgs) -> anyhow::Result<()> {
     Ok(())
 }
 
-/// Checks if the runtime profile settings in the provided from [`Metadata`] contain the required entries for backtrace generation.
-fn check_profile(scarb_metadata: &Metadata) -> anyhow::Result<()> {
-    const BACKTRACE_REQUIRED_ENTRIES: &[(&str, &str)] = &[
-        ("unstable-add-statements-functions-debug-info", "true"),
-        ("unstable-add-statements-code-locations-debug-info", "true"),
-        ("panic-backtrace", "true"),
-    ];
-
-    let manifest = get_manifest(scarb_metadata)?;
-
+fn check_profile(
+    compiler_config: &Value,
+    profile: &str,
+    workspace_manifest_path: &Utf8PathBuf,
+) -> anyhow::Result<()> {
     let has_needed_entries =
-        check_cairo_profile_entries(&manifest, scarb_metadata, BACKTRACE_REQUIRED_ENTRIES);
+        bool_field_or_unstable(compiler_config, "add_statements_code_locations_debug_info")
+            && bool_field_or_unstable(compiler_config, "add_statements_functions_debug_info")
+            && bool_field(compiler_config, "panic_backtrace");
 
     ensure!(
         has_needed_entries,
         formatdoc! {
-            "Scarb.toml must have the following Cairo compiler configuration to run backtrace:
+            "{workspace_manifest_path} must have the Cairo compiler configuration equivalent to the following one to run backtrace:
 
             [profile.{profile}.cairo]
             unstable-add-statements-functions-debug-info = true
@@ -51,7 +48,6 @@ fn check_profile(scarb_metadata: &Metadata) -> anyhow::Result<()> {
             panic-backtrace = true
             ... other entries ...
             ",
-            profile = scarb_metadata.current_profile
         },
     );
 
--- a/crates/forge/src/profile_validation/debugger.rs
+++ b/crates/forge/src/profile_validation/debugger.rs
@@ -1,26 +1,23 @@
-use crate::profile_validation::{check_cairo_profile_entries, get_manifest};
+use crate::profile_validation::{bool_field, bool_field_or_unstable, str_field};
 use anyhow::ensure;
+use camino::Utf8PathBuf;
 use indoc::formatdoc;
-use scarb_metadata::Metadata;
+use serde_json::Value;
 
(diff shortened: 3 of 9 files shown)


crates/forge/src/profile_validation/mod.rs
```diff
@@ -1,69 +1,59 @@
...
+        check_backtrace_compatibility(test_args, compiler_config, profile)?;
     }
 
     Ok(())
 }
 
-/// Gets the runtime manifest from the [`Metadata`] and parses it into a [`DocumentMut`].
-fn get_manifest(scarb_metadata: &Metadata) -> anyhow::Result<DocumentMut> {
-    Ok(fs::read_to_string(&scarb_metadata.runtime_manifest)?.parse::<DocumentMut>()?)
+fn bool_field_or_unstable(value: &Value, key: &str) -> bool {
```
piotmag769: This function exists for backwards compability purposes - the keys in config were changed in Scarb [v2.15.0-rc.0](https://github.com/software-mansion/scarb/releases/tag/v2.15.0-rc.0)

crates/forge/src/profile_validation/mod.rs
```diff
@@ -1,69 +1,59 @@
...
 ) -> anyhow::Result<()> {
+    // Config of the first CU is as good as any as all compilation units have the same compiler
+    // config.
+    let Some(compiler_config) = scarb_metadata
+        .compilation_units
+        .first()
+        .map(|cu| &cu.compiler_config)
+    else {
+        return Ok(());
+    };
```
piotmag769, line 27: @maciektr idk if it makes sense to check all compilations units - I can do that, but setting different debug info for different targets in the future sounds like a weird edge case that we don't need to cover. WDYT?

maciektr: While this is true, this assumptions relies on underlying Scarb implementation details. Compiler configs are not guaranteed to be identical and this behaviour can change in the future. 
On the other hand, validating against all configs is also not something that I would advice - what you should really do is select targets that you care about (like test targets for test runner) and only validate those. 

If you decide to stick to the first one for now, I am rather okay with that, but it's not my decision to make and please mark it with TODO in such case.

crates/forge/src/profile_validation/debugger.rs
```diff
@@ -1,26 +1,18 @@
...
+    let has_needed_entries = bool_field(compiler_config, "add_functions_debug_info")
+        && bool_field_or_unstable(compiler_config, "add_statements_code_locations_debug_info")
+        && bool_field_or_unstable(compiler_config, "add_statements_functions_debug_info")
+        && str_field(compiler_config, "compiler_optimizations") == "Disabled";
 
     ensure!(
         has_needed_entries,
         formatdoc! {
-            "Scarb.toml must have the following Cairo compiler configuration to launch the debugger:
+            "workspace Scarb.toml must have the following Cairo compiler configuration to launch the debugger:
```
maciektr: Nit: showing relative path to the Scarb.toml in users project would be nice here

maciektr (approved):

crates/forge/src/profile_validation/backtrace.rs
```diff
@@ -27,31 +25,23 @@ fn check_if_native_disabled(test_args: &TestArgs) -> anyhow::Result<()> {
...
-        check_cairo_profile_entries(&manifest, scarb_metadata, BACKTRACE_REQUIRED_ENTRIES);
+        bool_field_or_unstable(compiler_config, "add_statements_code_locations_debug_info")
+            && bool_field_or_unstable(compiler_config, "add_statements_functions_debug_info")
+            && bool_field(compiler_config, "panic_backtrace");
 
     ensure!(
         has_needed_entries,
         formatdoc! {
-            "Scarb.toml must have the following Cairo compiler configuration to run backtrace:
+            "workspace Scarb.toml must have the following Cairo compiler configuration to launch the debugger:
```
maciektr:
```suggestion
            "workspace Scarb.toml must have the following Cairo compiler configuration to run the backtrace:
```

piotmag769: Thx, missed that

crates/forge/src/profile_validation/mod.rs
piotmag769: Changed it to iterate over all relevant targets - please re-review

crates/forge/tests/e2e/coverage.rs
```diff
@@ -33,6 +33,70 @@ fn test_coverage_project_and_pass_args() {
...
+        .arg("--output-path")
+        .arg("./my_file.lcov")
+        .assert()
+        .success();
+
+    assert!(temp.join("my_file.lcov").is_file())
+}
+
+// `skip-optimizations` also sets `inlining-strategy` to `avoid` - available from Scarb 2.15.0
+#[cfg(feature = "run_test_for_scarb_since_2_15_1")]
```
MKowalski8: I think this feature flag is not needed. In scheduled tests oldest version we run is 2.15.2, so it will never go below 2.15.0. (apply to other places as well). We are also getting rid of this in the near future

crates/forge/src/profile_validation/mod.rs
```diff
@@ -1,69 +1,97 @@
...
     test_args: &TestArgs,
     scarb_metadata: &Metadata,
 ) -> anyhow::Result<()> {
+    let Some(compiler_config) = scarb_metadata
+        .compilation_units
+        .iter()
+        .find(|cu| &cu.target == target)
+        .map(|cu| &cu.compiler_config)
+    else {
+        return Ok(());
```
MKowalski8, line 53: Should it just return `Ok(())`?

MKowalski8 (approved):

crates/forge/src/profile_validation/mod.rs
piotmag769: This is just a failsafe - it will never fail anyways. If there exists a target for a package, it has to have a compilation unit corresponding to it

crates/forge/tests/e2e/coverage.rs
piotmag769: I will manually test it with 2.14.0 then to make sure I didn't break backwards compatibility without a reason

crates/forge/src/profile_validation/mod.rs
MKowalski8: Could you add this as a comment? It might help someone in the future to understand this. Or even use `unreachable!`here, if you say it will never fail?

crates/forge/src/profile_validation/mod.rs
```diff
@@ -1,69 +1,97 @@
...
-use toml_edit::{DocumentMut, Table};
+/// Checks if the compiler settings provided in [`Metadata`] can be used to run
+/// coverage, backtrace and debugger if applicable.
+pub fn check_compiler_config_compatibility(
+    test_args: &TestArgs,
+    scarb_metadata: &Metadata,
+    packages_to_build: &[PackageMetadata],
+) -> anyhow::Result<()> {
+    for package in packages_to_build {
+        if test_args.no_optimization
```
maciektr, line 22: What does this flag represent? 🤔

piotmag769: https://foundry-rs.github.io/starknet-foundry/appendix/snforge/test.html#--no-optimization

maciektr: Ahhh, right

crates/forge/src/profile_validation/mod.rs
piotmag769: Added a comment

crates/forge/tests/e2e/coverage.rs
piotmag769: Tested - works with 2.13.0 and 2.14.0

maciektr (approved):

CHANGELOG.md
```diff
@@ -17,6 +17,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Minimal recommended `Scarb` version is now `2.15.2` (updated from `2.14.0`)
 
+#### Fixed
+
+- Validation of Scarb's compiler config now properly takes into account `[cairo]` section and keys without `unstable-`
+prefix
```
ddoktorski: nit:
```suggestion
- Validation of Scarb's compiler config now properly takes into account `[cairo]` section and keys without `unstable-` prefix
```

ddoktorski (approved):

PR #4291, opened 2026-04-17
Optimize caching on CI

ddoktorski:
- Optimize rust caching, stop storing a separate cache for each job
- Omit rust caching in jobs that use `nextest` binaries
- Install `mdBook` via `taiki-e/install-action`
- Do not store all debug information in `ci` profile - to decrease cached file sizes


.github/workflows/ci.yml
```diff
@@ -55,8 +57,9 @@ jobs:
         with:
           setup-scarb: 'false'
           setup-usc: 'false'
+          save-cache: 'true'
```
franciszekjob, line 60: This job saves the cache, but `test-workspace` above also saves it without setting `cache-key`. In `setup-tools`, the default `cache-key` is `'default'`, so both jobs end up writing the same shared cache entry. Since GitHub caches are immutable, whichever job finishes first wins, and the other save is skipped. Could we give these jobs different cache keys, or make only one of them save the cache? If this is intentional, can you confirm that sharing the same cache here is ok in practice?

.github/actions/setup-tools/action.yml
```diff
@@ -18,13 +18,29 @@ inputs:
     description: 'Whether to setup universal-sierra-compiler'
     required: false
     default: 'true'
+  setup-rust-cache:
+    description: 'Whether to restore/save the Rust build cache.'
+    required: false
+    default: 'true'
+  cache-key:
+    description: 'Suffix appended to the rust-cache shared-key to isolate cache entries between different feature sets.'
```
franciszekjob: nit: The `cache-key` description says this is a suffix appended to the rust-cache key, but it is passed directly as rust-cache’s `shared-key`. Since `shared-key` replaces the automatic job-based part of the cache key rather than acting like an appended suffix, it seems to me that we should reword this description to make that clearer?

.github/workflows/ci.yml
ddoktorski: This is not a problem, both jobs can use the same cache. Even though the first one to finish will handle the write, that works perfectly fine for our setup

franciszekjob (approved):

ddoktorski:
Time comparison based on execution on fork:
* [cold start](https://github.com/ddoktorski/starknet-foundry/actions/runs/24828718183) (no cache at all) ~24 minutes
* [cache hit](https://github.com/ddoktorski/starknet-foundry/actions/runs/24829778671) (the most optimistic scenario - no changes in `.rs` files) ~12 minutes

.github/workflows/ci.yml
```diff
@@ -401,29 +424,21 @@ jobs:
...
-        run: |
-          cargo install --version ${MDBOOK_VERSION} mdbook
-          cargo install --version ${MDBOOK_VARIABLES_VERSION} mdbook-variables
-      - name: Install mdBook Link-Check
-        run: |
-          cargo install mdbook-linkcheck
+      - uses: taiki-e/install-action@055f5df8c3f65ea01cd41e9dc855becd88953486 # 2.75.18
+        with:
+          # TODO(#3970): Upgrade mdbook-variables after resolving issue
+          tool: mdbook@0.4.52,mdbook-variables@0.3.0,mdbook-linkcheck
```
MKowalski8, line 433: Do we want to pin `mdbook-linkcheck` version as well? Or is it expected to remain as it was before? (with no pinned version)

MKowalski8 (approved):

.github/workflows/ci.yml
ddoktorski: I guess it can stay as is, afaik this tool is also no longer maintained

Issue #4091, opened 2026-01-13
Bump minimal recommended scarb version to 2.15.2


MKowalski8:
### Current State

.

### Objective

.

ddoktorski:
In the scope of this task please remove all outdated features from this list: https://github.com/foundry-rs/starknet-foundry/blob/1660d194ead7c50a372aabfa566f4eaddf64ec01/crates/forge/Cargo.toml#L8-L17

PR #4295, opened 2026-04-20
Remove unneded feature flags in tests

MKowalski8:
Towards #4091 

## Introduced changes

- Delete unneded feature flags and unify oracles tests in normal `e2e`

## Checklist

- [x] Linked relevant issue
- [ ] Added relevant tests

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -296,34 +296,25 @@ jobs:
       - name: Run snap E2E tests
         run: cargo test --profile ci -p forge --test main snap_
 
-  test-plugin-checks:
-    name: Test plugin across different scarb versions
+  test-special-conditions:
+    name: Test special conditions
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v6
       - uses: ./.github/actions/setup-tools
-        # No setup scarb action is used because we want to install multiple versions of scarb through asdf
-        with:
-          setup-scarb: 'false'
-      - uses: asdf-vm/actions/install@b7bcd026f18772e44fe1026d729e1611cc435d47
-      - run: asdf install scarb 2.12.0
-      - run: cargo test --profile ci --package forge --features scarb_2_12_0 --test main e2e::plugin_versions
-
-  test-requirements-check-special-conditions:
-    name: Test requirements check special conditions
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v6
-      - uses: ./.github/actions/setup-tools
+        # Scarb is installed manually below to control the version
         with:
           setup-scarb: 'false'
 
       - run: cargo test --profile ci --package forge --features no_scarb_installed --lib compatibility_check::tests::failing_tool_not_installed
 
+      - uses: asdf-vm/actions/install@b7bcd026f18772e44fe1026d729e1611cc435d47
+
       - uses: software-mansion/setup-scarb@v1
         with:
-          scarb-version: "2.12.0"
+          scarb-version: '2.12.0'
 
+      - run: cargo test --profile ci --package forge --features scarb_2_12_0 --test main e2e::plugin_versions::new_with_minimal_scarb
       - run: cargo test --profile ci --package forge --features scarb_2_12_0 --test main e2e::requirements::test_warning_on_scarb_version_below_recommended
 
   test-forge-scarb-plugin:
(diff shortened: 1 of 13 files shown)


.github/workflows/scheduled.yml
```diff
@@ -97,13 +97,7 @@ jobs:
...
         id: test_features
         run: |
-          SCARB_VERSION="${{ matrix.version }}"
-
-          if [ "$(printf '%s\n' "2.15.1" "$SCARB_VERSION" | sort -V | head -n1)" = "2.15.1" ]; then
-            FEATURES="skip_test_for_only_latest_scarb,run_test_for_scarb_since_2_15_1"
-          else
-            FEATURES="skip_test_for_only_latest_scarb"
-          fi
+          FEATURES="skip_test_for_only_latest_scarb"
```
piotmag769, line 100: Please rename this flag while we are at it - I spent 10 minutes today wondering if it should skip tests while Scarb is latest or if it should skip them for non-latest xD

MKowalski8: Yeah I can do that. I'm cleaning those features a little bit 😅

crates/forge/Cargo.toml
```diff
@@ -10,7 +10,6 @@ edition.workspace = true
 scarb_2_12_0 = []
 no_scarb_installed = []
 non_exact_gas_assertions = []
-skip_test_for_only_latest_scarb = []
 test_for_multiple_scarb_versions = []
```
piotmag769: I was skimming through features before - this flag is useless. You never even execute tests that have this flag since they are all in `crates/forge/tests/e2e/plugin_versions.rs`, but the only tests you execute with this flag set are `e2e::plugin_diagnostics` tests. 

https://github.com/foundry-rs/starknet-foundry/blob/99073c96af7dcd45f98b78605d80ae9acce6c74a/.github/workflows/ci.yml#L297

And even if you want to keep these tests (and make them execute in CI ofc) - they do not require multiple scarb versions installed - one of them needs scarb 2.12.0 and that's all.

piotmag769 (commented):
Also one general note, isn't this pattern  `#[cfg_attr(not(feature = "some_feature"), ignore)]` a bit convoluted?
I would do `#[cfg(feature = "some_feature")]` and get the same result - unless you really need the info about ignored test in `cargo test` output

crates/forge/Cargo.toml
MKowalski8: I'll look into it

piotmag769 (approved):

.github/workflows/ci.yml
```diff
@@ -297,45 +282,24 @@ jobs:
       - name: Run snap E2E tests
         run: cargo test --profile ci -p forge --test main snap_
 
-  test-plugin-checks:
```
MKowalski8, line 314: Since the plugin is already covered by the `e2e` tests and the `test_for_multiple_scarb_versions` flag appeared to be a redundant leftover, I merged the special condition tests into a single job. 
We can think of moving it only to scheduled tests. But for example mayby it's good to know as fast as possible if the backward compatibility with `snforge new` is broken.

crates/forge/tests/e2e/plugin_diagnostics.rs
```diff
@@ -126,10 +126,6 @@ fn syntax() {
 }
 
 #[test]
-#[cfg_attr(
-    feature = "skip_test_for_only_latest_scarb",
```
MKowalski8, line 130: Those tests pass without it. So imo there is no reason to keep it. If something brake in the future we can just simply add it back.

franciszekjob (approved):

.github/workflows/ci.yml
```diff
@@ -97,13 +87,8 @@ jobs:
...
         with:
           tool: nextest@0.9.98
-      - name: Determine test features
-        id: test_features
-        run: |
-          FEATURES="cairo-native,run_test_for_scarb_since_2_15_1"
-          echo "features=$FEATURES" >> "$GITHUB_OUTPUT"
       - name: Build and archive tests
-        run: cargo nextest archive --cargo-profile ci -p forge --features "${{ steps.test_features.outputs.features }}" --archive-file 'nextest-archive-forge-native-${{ runner.os }}.tar.zst'
+        run: cargo nextest archive --cargo-profile ci -p forge --features "cairo-native" --archive-file 'nextest-archive-forge-native-${{ runner.os }}.tar.zst'
```
ddoktorski: nit:
```suggestion
        run: cargo nextest archive --cargo-profile ci -p forge --features cairo-native --archive-file 'nextest-archive-forge-native-${{ runner.os }}.tar.zst'
```

ddoktorski (approved):

Issue #3899, opened 2025-11-03
Find performance bottlenecks


cptartur:
### Current State

Currently, we don't have a good idea of areas in snforge that are possible performance bottlenecks to be improved.

### Objective

Measure performance and investigate possible bottlenecks in snforge

### Additional Context

We should focus on investigating these possible paint points and producing a performance report (in any structure) that covers these areas:

* Processing workspace (metadata, tests contracts compilation)
* Processing package (resolving config, config run, filtering)
* Test case execution (steps in `run_test_case` function)

While finding edge cases where performance takes a hit will be beneficial, focus should be on standard use cases of snforge.

Focus on execution time.

The report should document which methods were used to measure/get performance data so we can replicate these after we improve the eventual bottlenecks.

I recommend using existing performance harnesses (`tracing::instrument`) and expanding them where relevant.

PR #4296, opened 2026-04-21
Optimize config resolution for non-matching targets in `--exact` runs

franciszekjob:
Towards #3899

## Introduced changes

Skip config resolution for non-matching targets when `--exact` flag is used. This change is helpful for test run code lens in LS, since they use `--exact`.

<br>

**With `--workspace` flag**

**Before**
```shell
➜  cairo-contracts git:(main) ✗ hyperfine --runs 5 "snforge test openzeppelin_token::tests::erc721::test_erc721::test_transfer_from_nonexistent --exact --workspace"                 
Benchmark 1: snforge test openzeppelin_token::tests::erc721::test_erc721::test_transfer_from_nonexistent --exact --workspace
  Time (mean ± σ):     38.935 s ±  7.293 s    [User: 97.577 s, System: 19.230 s]
  Range (min … max):   30.666 s … 48.018 s    5 runs
```

**After**
```shell
➜  cairo-contracts git:(main) ✗ hyperfine --runs 5 "/Users/franciszekjob/Projects/SWM/starknet-foundry/target/release/snforge test openzeppelin_token::tests::erc721::test_erc721::test_transfer_from_nonexistent --exact --workspace"                 
Benchmark 1: /Users/franciszekjob/Projects/SWM/starknet-foundry/target/release/snforge test openzeppelin_token::tests::erc721::test_erc721::test_transfer_from_nonexistent --exact --workspace
  Time (mean ± σ):     29.249 s ±  2.614 s    [User: 22.787 s, System: 14.182 s]
  Range (min … max):   25.940 s … 32.706 s    5 runs
```

<br>

**With `--package` flag**

**Before**

```shell
➜  cairo-contracts git:(main) ✗ hyperfine --runs 5 "snforge test openzeppelin_token::tests::erc721::test_erc721::test_transfer_from_nonexistent --exact --package openzeppelin_token" 
Benchmark 1: snforge test openzeppelin_token::tests::erc721::test_erc721::test_transfer_from_nonexistent --exact --package openzeppelin_token
  Time (mean ± σ):     16.125 s ±  1.337 s    [User: 38.437 s, System: 6.164 s]
  Range (min … max):   14.457 s … 17.824 s    5 runs
```

**After**

```shell
➜  cairo-contracts git:(main) ✗ hyperfine --runs 5 "/Users/franciszekjob/Projects/SWM/starknet-foundry/target/release/snforge test openzeppelin_token::tests::erc721::test_erc721::test_transfer_from_nonexistent --exact --package openzeppelin_token"
Benchmark 1: /Users/franciszekjob/Projects/SWM/starknet-foundry/target/release/snforge test openzeppelin_token::tests::erc721::test_erc721::test_transfer_from_nonexistent --exact --package openzeppelin_token
  Time (mean ± σ):     11.330 s ±  1.318 s    [User: 7.184 s, System: 4.180 s]
  Range (min … max):    9.696 s … 13.110 s    5 runs
```

(Benches done in comparison with snforge `nightly-2026-04-22`)

## Checklist

- [x] Linked relevant issue
- [ ] Added relevant tests


crates/forge-runner/src/running/target.rs
```diff
@@ -38,10 +48,6 @@ pub fn prepare_test_target(
     let funcs = by_id!(funcs);
     let type_declarations = by_id!(type_declarations);
 
-    let casm_program = Arc::new(compile_raw_sierra_at_path(
-        test_target_raw.sierra_program_path.as_std_path(),
-    )?);
-
```
franciszekjob, line 44: note: Compilation was moved below so it happens only if there are tests to be run from a test target.

crates/forge-runner/src/running/target.rs
```diff
@@ -12,16 +13,54 @@ use cairo_lang_sierra::{
...
 use rayon::iter::ParallelIterator;
 use std::{collections::HashMap, sync::Arc};
 use universal_sierra_compiler_api::compile_raw_sierra_at_path;
+use universal_sierra_compiler_api::representation::{AssembledCairoProgram, RawCasmProgram};
+
+#[derive(Debug, Clone, Copy)]
+pub enum TestNameSelection<'a> {
+    All,
+    ExactMatch(&'a str),
+}
```
franciszekjob: note: We can extend it later so the optimization will work with test name filter too. However, this will be related with the number of filtered out tests, which is displayed. For `--exact` we simply display `other filtered out` while for matching name filter, we actually count them.

crates/forge-runner/src/running/target.rs
```diff
@@ -42,30 +81,39 @@ pub fn prepare_test_target(
...
-                config: raw_config.into(),
-                name: case.debug_name.clone().unwrap().into(),
-                test_details,
+    let test_cases = if let Some(exact_matches) = selected_test_cases {
+        exact_matches
+            .into_par_iter()
+            .map(|case| {
+                build_test_case_with_config(
+                    funcs[&case.id],
+                    case.debug_name
```
MKowalski8: nit: We could get rid of this second `clone()`

When we retrive `selected_test_cases` we could do sth like this and save (id, name):

``` rust
let selected = match test_name_selection {
        TestNameSelection::All => None,
        TestNameSelection::ExactMatch(exact_match) => Some(
            executables
                .iter()
                .filter_map(|case| {
                    let raw_name: String = case.debug_name.clone()?.into();
                    let name = sanitize_test_case_name(&raw_name);
                    if name == exact_match {
                        Some((&case.id, raw_name)) 
                    } else {
                        None
                    }
                })
                .collect(),
        ),
    };
```

Then here do:
```rust
         exact_matches
            .into_par_iter()
            .map(|(id, name)| {
                build_test_case_with_config(
                    funcs[id],
                    name,
                    &type_declarations,
                    &casm_program,
                    *tracked_resource,
                )
            })
```

But that's just a suggestion, not a big deal

crates/forge-runner/src/running/target.rs
```diff
@@ -12,16 +13,54 @@ use cairo_lang_sierra::{
...
 use rayon::iter::IntoParallelRefIterator;
 use rayon::iter::ParallelIterator;
 use std::{collections::HashMap, sync::Arc};
 use universal_sierra_compiler_api::compile_raw_sierra_at_path;
+use universal_sierra_compiler_api::representation::{AssembledCairoProgram, RawCasmProgram};
+
+#[derive(Debug, Clone, Copy)]
+pub enum TestNameSelection<'a> {
+    All,
+    ExactMatch(&'a str),
```
MKowalski8: nit: We could change this to:

```suggestion
    ExactMatch(String),
```

It will remove lifetime and remove `as_deref()` from  `spawn_prepare_test_target`.
```rust      
        let selection = match maybe_exact_name {
            Some(exact_match) => TestNameSelection::ExactMatch(exact_match),
            None => TestNameSelection::All,
        };
```

However it's only suggestion, because in the end `&str` should be cheaper 😅

MKowalski8 (approved):

crates/forge-runner/src/running/target.rs
franciszekjob: > because in the end &str should be cheaper

That's why I'll leave current code 😅

piotmag769 (commented):
Just a conceptual question - couldn't you apply this optimization to regular (non-exact) filtering as well? I believe you could just use `TestFilter` and [this](https://github.com/foundry-rs/starknet-foundry/blob/af3fbb51135acd648e9abc4f4c0a146e3a6f569d/crates/forge/src/test_filter.rs#L107-L125) part of code since it does not depend on resolved config

crates/forge-runner/src/running/target.rs
DelevoXDG: Is that related to the reverted fa6719fd751274a9cbb7772d273b3aebf7f4f3a4?

I think it's quite reasonable to apply optimization across the board, not only in `ExactMatch` context.

If that's indeed significant time effort, it can be done separately (perhaps worth creating an issue?).

But if it's not, I think it's a logical addition.

Specifically `#num filtered out` - is that something tricky to implement? 

If that would at all simplify things, we can consider unifying the way we display in `--exact` and non-exact context, e.g. cargo always displays a number 🤔 
```
cargo test --package forge --test main -- e2e::build_profile::simple_package_build_profile --exact     
    Finished `test` profile [unoptimized + debuginfo] target(s) in 0.81s
     Running tests/main.rs (target/debug/deps/main-dbbe155b9b48fef5)

running 1 test
test e2e::build_profile::simple_package_build_profile ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 448 filtered out; finished in 8.67s
```

crates/forge-runner/src/running/target.rs
```diff
@@ -12,16 +13,51 @@ use cairo_lang_sierra::{
...
+                    (sanitize_test_case_name(&raw_name) == exact_match)
+                        .then_some((&case.id, raw_name))
+                })
+                .collect::<Vec<_>>(),
+        ),
+    };
+
+    if selected_test_cases.as_ref().is_some_and(Vec::is_empty) {
+        return Ok(empty_test_target(test_target_raw));
+    }
```
DelevoXDG: Instead of creating synthetic `empty_test_target`, wouldn't it be simpler + cleaner to just have `prepare_test_target` return `Result<Option<TestTargetWithConfig>>`, and return and explicitly handle `None` in no-matches context?

crates/forge/src/run_tests/package.rs
```diff
@@ -125,10 +125,14 @@ impl RunForPackageArgs {
         }
 
         let tracked_resource = forge_config.test_runner_config.tracked_resource;
+        let maybe_exact_name = match &tests_filter.name_filter {
+            NameFilter::ExactMatch(name) => Some(name.clone()),
+            _ => None,
+        };
```
DelevoXDG: Nit: 

The `maybe_exact_name` back-and-forth `enum -> string -> enum` seems a bit odd to me.

Wouldn't it be easier to:
1. Convert `TestNameSelection::ExactMatch` to use `String`,
```rs
#[derive(Debug, Clone)]
pub enum TestNameSelection {
    All,
    ExactMatch(String),
}
```
2. Pass `NameFilter` directly to `spawn_prepare_test_target`, match it there:
```rs
        let test_name_selection = match name_filter {
            NameFilter::ExactMatch(name) => TestNameSelection::ExactMatch(name),
            NameFilter::Match(_) | NameFilter::All => TestNameSelection::All,
        };
```
thus :
- matching just once
- and if we decide to go back to the idea of applying this change to `Match` as well (https://github.com/foundry-rs/starknet-foundry/commit/f36729159d88d344454cc474483f1fe39177c7a8), it should be more trivial to update.

crates/forge/tests/data/lazy_config_filtering/src/lib.cairo
```diff
@@ -0,0 +1,16 @@
...
+        assert(true, 'ok');
+    }
+
+    // This config is intentionally invalid. When running with
+    // `--exact lazy_config_filtering::tests::selected_exact`, this test should be
+    // filtered out and thus not fail.
+    #[test]
+    #[fork("missing_fork")]
+    fn filtered_out_broken_config() {
+        assert(true, 'should never run');
```
DelevoXDG: From what I understand that should fail before test running due to missing fork, but I think this should be:

```suggestion
        assert(false, 'should never run');
```

to guard against case somehow runs & succeeds despite missing fork?

Otherwise `'should never run'` doesn't really make sense, since it's never printed, even if test somehow runs

DelevoXDG (commented):
Generally LGTM

Let's decide whether non-exact filter is in scope, or create an issue for it

crates/forge-runner/src/running/target.rs
```diff
@@ -12,16 +14,43 @@ use cairo_lang_sierra::{
...
 ) -> Result<TestTargetWithConfig> {
+    let default_executables = vec![];
+    let executables = test_target_raw
+        .sierra_program
+        .debug_info
+        .as_ref()
+        .and_then(|info| info.executables.get("snforge_internal_test_executable"))
+        .unwrap_or(&default_executables);
+
+    let selected_test_cases = name_filter.exact_match().map(|exact_match| {
```
piotmag769: Why do you need this distinction? Imo just filter regularly via `matches` method and then check if `selected_test_cases` is empty - this code doesn't need to be aware of filter types, no?

crates/forge-runner/src/running/target.rs
franciszekjob: Makes sense, refactored.

crates/forge/tests/data/lazy_config_filtering/src/lib.cairo
franciszekjob: Good point! https://github.com/foundry-rs/starknet-foundry/pull/4296/commits/40c77f8f7359b6ff3b086b9c852608ba960bfc21

crates/forge-runner/src/running/target.rs
franciszekjob: https://github.com/foundry-rs/starknet-foundry/pull/4296/commits/1d1b3409f952783fb1643d90e64f32dcca10e84a

crates/forge/src/run_tests/package.rs
franciszekjob: Closing, as this comment is not relevant anymore.

crates/forge-runner/src/running/target.rs
franciszekjob: Optimized filtering for non-exact match will be added in separate PR.

piotmag769 (approved):

PR #4299, opened 2026-04-22
Display events in the debugging trace

integraledelebesgue:
Closes #

## Introduced changes

-

## Checklist

- [x] Linked relevant issue
- [x] Updated relevant documentation
- [x] Added relevant tests
- [x] Performed self-review of the code
- [x] Added changes to `CHANGELOG.md`

PR #4300, opened 2026-04-22
Implement a reversed data transformer for events

integraledelebesgue:
Closes #

## Introduced changes

-

## Checklist

- [ ] Linked relevant issue
- [ ] Added relevant tests

### DIFF
--- a/crates/data-transformer/src/reverse_transformer/event.rs
+++ b/crates/data-transformer/src/reverse_transformer/event.rs
@@ -1,0 +1,399 @@
+use crate::reverse_transformer::transform::{ReverseTransformer, TransformationError};
+use crate::reverse_transformer::types::{Enum, Struct, StructField, Type};
+use cairo_lang_parser::utils::SimpleParserDatabase;
+use conversions::serde::deserialize::BufferReadError;
+use starknet_rust::core::types::contract::{
+    AbiEntry, AbiEvent, AbiEventEnum, AbiEventStruct, EventField, EventFieldKind, TypedAbiEvent,
+};
+use starknet_rust::core::utils::get_selector_from_name;
+use starknet_types_core::felt::Felt;
+
+#[derive(Debug, thiserror::Error)]
+pub enum ReverseTransformEventError {
+    #[error(r#"Typed event matching emitted selector "{0:#x}" not found in ABI of the contract"#)]
+    EventNotFound(Felt),
+    #[error("untyped event `{0}` is unsupported by reverse transformer")]
+    UnsupportedUntypedEvent(String),
+    #[error("event `{0}` is unsupported by reverse transformer")]
+    UnsupportedEvent(String),
+    #[error("abi is invalid")]
+    InvalidAbi,
+    #[error(transparent)]
+    TransformationError(#[from] TransformationError),
+}
+
+#[derive(Clone, Copy)]
+enum SelectorSource {
+    Consume,
+    Provided(Felt),
+    None,
+}
+
+enum InternalEventError {
+    SelectorMismatch,
+    Other(ReverseTransformEventError),
+}
+
+impl From<ReverseTransformEventError> for InternalEventError {
+    fn from(value: ReverseTransformEventError) -> Self {
+        Self::Other(value)
+    }
+}
+
+/// Transforms a set of event keys and data into a Cairo-like string representation of the event.
+pub fn reverse_transform_event(
+    keys: &[Felt],
+    data: &[Felt],
+    abi: &[AbiEntry],
+) -> Result<String, ReverseTransformEventError> {
+    let selector = keys.first().copied().ok_or({
+        ReverseTransformEventError::TransformationError(TransformationError::BufferReaderError(
+            BufferReadError::EndOfBuffer,
+        ))
+    })?;
+
+    let db = SimpleParserDatabase::default();
+
+    let mut first_error = None;
+
+    for typed_event in top_level_typed_events(abi) {
+        let mut transformer = EventReverseTransformer::new(keys, data, abi);
+
+        match transformer.transform_typed_event(typed_event, SelectorSource::Consume, &db) {
+            Ok(decoded) if transformer.keys.is_empty() && transformer.data.is_empty() => {
+                return Ok(decoded.to_string());
+            }
+            Ok(_) => {
+                if first_error.is_none() {
+                    first_error = Some(ReverseTransformEventError::InvalidAbi);
+                }
+            }
+            Err(InternalEventError::SelectorMismatch) => {}
+            Err(InternalEventError::Other(error)) => {
+                if first_error.is_none() {
+                    first_error = Some(error);
+                }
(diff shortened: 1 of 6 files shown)


integraledelebesgue:

- #4299
    - #4300 ◀
        - #4301

crates/data-transformer/src/reverse_transformer/event.rs
```diff
@@ -0,0 +1,399 @@
...
+    // Select only top-level events, i.e. such that are not referenced by other events.
+    abi.iter().filter_map(move |entry| match entry {
+        AbiEntry::Event(AbiEvent::Typed(typed_event))
+            if !all_types_in_events.contains(&typed_event_name(typed_event)) =>
+        {
+            Some(typed_event)
+        }
+        _ => None,
+    })
+}
```
franciszekjob: nit: Could we use `HashSet` instead of a `Vec` for `all_types_in_events`? Currently, calling `.contains()` inside the `.filter_map` closure gives us `O(N^2)` complexity, while using `HashSet`, lookup will decrease to `O(1)`.

crates/data-transformer/src/reverse_transformer/event.rs
```diff
@@ -0,0 +1,399 @@
...
+                .transform_from_keys(&field.r#type, db)
+                .map_err(Into::into),
+            EventFieldKind::Data => self
+                .transform_from_data(&field.r#type, db)
+                .map_err(Into::into),
+            EventFieldKind::Nested => {
+                self.transform_event_type(&field.r#type, SelectorSource::Consume, db)
+            }
+            EventFieldKind::Flat => {
+                self.transform_event_type(&field.r#type, SelectorSource::None, db)
```
franciszekjob, line 220: Could this still fail for a valid flattened struct field, e.g. `OuterEvent { #[flat] inner: InnerEvent, tail: felt252 }`? In that case `InnerEvent` would not have its own selector in the emitted payload, but here we pass `SelectorSource::None` and the struct path still seems to require one.

crates/data-transformer/src/reverse_transformer/event.rs
```diff
@@ -0,0 +1,399 @@
...
+
+        for variant in &abi_event_enum.variants {
+            match variant.kind {
+                EventFieldKind::Flat => {
+                    if self.event_type_matches_selector(&variant.r#type, selector)? {
+                        return self.transform_event_type(
+                            &variant.r#type,
+                            SelectorSource::Provided(selector),
+                            db,
+                        );
```
franciszekjob: Could this be lossy for `#[flat]` enum variants? It looks like we return only the inner decoded event here, without preserving the outer variant wrapper, which might become ambiguous if multiple flat branches match the same selector.

crates/data-transformer/tests/integration/event_reverse_transformer.rs
franciszekjob, line 1: Could we add a regression test for a directly-emitted typed event that is also referenced as a nested payload in another event? That seems like an edge case behind the top-level filtering logic here.

crates/data-transformer/src/reverse_transformer/event.rs
```diff
@@ -0,0 +1,399 @@
...
+}
+
+impl From<ReverseTransformEventError> for InternalEventError {
+    fn from(value: ReverseTransformEventError) -> Self {
+        Self::Other(value)
+    }
+}
+
+/// Transforms a set of event keys and data into a Cairo-like string representation of the event.
+pub fn reverse_transform_event(
```
franciszekjob, line 55: nit: Could we update the ordering so struct/enums are at the top of file?

crates/data-transformer/src/reverse_transformer/event.rs
```diff
@@ -0,0 +1,399 @@
...
+#[derive(Debug, thiserror::Error)]
+pub enum ReverseTransformEventError {
+    #[error(r#"Typed event matching emitted selector "{0:#x}" not found in ABI of the contract"#)]
+    EventNotFound(Felt),
+    #[error("untyped event `{0}` is unsupported by reverse transformer")]
+    UnsupportedUntypedEvent(String),
+    #[error("event `{0}` is unsupported by reverse transformer")]
+    UnsupportedEvent(String),
+    #[error("abi is invalid")]
+    InvalidAbi,
```
MKowalski8: We could potentailly split it, to sth like:
``` rust
    #[error("abi is invalid: event `{0}` matched selector but left {1} key(s) and {2} data felt(s) unconsumed")]
    InvalidAbi(String, usize, usize),
    #[error("abi is invalid: type `{0}` referenced in event but not found in abi")]
    InvalidAbiMissingType(String),
```

Then in `reverse_transform_event` we will use

```rust
            Ok(decoded) => {
                if first_error.is_none() {
                    first_error = Some(ReverseTransformEventError::InvalidAbi(
                        decoded.to_string(),
                        transformer.keys.len(),
                        transformer.data.len(),
                    ));
                }
            }
```

And in other 3 occurences `InvalidAbiMissingType`

However I'm not sure if this split is really necessary, since with proper ABI from compiler it shouldn't occur. But mayby more detailed error handling won't be so bad after all (and `reverse_data_transformer` don't have the best one). 

Left to consideration

crates/data-transformer/src/reverse_transformer/event.rs
integraledelebesgue: Done

crates/data-transformer/tests/integration/event_reverse_transformer.rs
integraledelebesgue: Done, I added a test to verify that the event referenced in an other event can be decoded when emitted directly - https://github.com/foundry-rs/starknet-foundry/blob/fbf382a38edd94ee97e4726390f9695ce7b60f3c/crates/data-transformer/tests/integration/event_reverse_transformer.rs#L152

crates/data-transformer/src/reverse_transformer/event.rs
integraledelebesgue: Done

crates/data-transformer/src/reverse_transformer/event.rs
integraledelebesgue: The struct path interprets the `SelectorSource::None` exactly as an information that the flattened content is being decoded. I added a test: https://github.com/foundry-rs/starknet-foundry/blob/fbf382a38edd94ee97e4726390f9695ce7b60f3c/crates/data-transformer/tests/integration/event_reverse_transformer.rs#L98

crates/data-transformer/src/reverse_transformer/event.rs
integraledelebesgue: That's a good idea - it's better to have explicit, descriptive errors. I divided the existing error according to your suggestion

franciszekjob (approved):

MKowalski8 (approved):

PR #4312, opened 2026-04-23
Apply debugger feedback to docs

piotmag769:


MKowalski8 (approved):

crates/forge/src/lib.rs
```diff
@@ -249,6 +249,7 @@ pub struct TestArgs {
     ///
     /// It makes snforge act as a debug adapter, enabling communication with an editor/IDE
     /// such as `VSCode` over DAP protocol.
+    /// It is meant for communication with a DAP client, not command line live debugging.
```
ddoktorski, line 252: As discussed offline it'd be better to hide this flag completely using https://docs.rs/clap/latest/clap/struct.Arg.html#method.hide. For now it's not possible because VS code extension relies on its presence in the output, but let's consider changing it in the future

docs/src/snforge-advanced-features/debugging.md
```diff
@@ -10,26 +10,27 @@ of the issue. To aid in debugging, `snforge` offers the following features:
...
-enabling an editor/IDE to connect and control the execution of the test.
-
-> 📝 **Note**
+> 💡 **Tip**
 >
-> `--launch-debugger` requires `--exact` - only a single test can be debugged at a time.
+> Make sure to check [Cairo Debugger website](https://docs.swmansion.com/cairo-debugger/) for more information!
+
+
+From version `0.60.0` onwards, the `snforge` is integrated with [Cairo Debugger](https://github.com/software-mansion/cairo-debugger),
```
ddoktorski, line 21: Should we mention that contract calls are not supported yet?

ddoktorski (approved):

crates/forge/src/lib.rs
piotmag769: LS*, not vscode extension. Changing this would make Scarb 2.18.0 not produce debug lens if future snforge version is in PATH  :/

docs/src/snforge-advanced-features/debugging.md
piotmag769: https://docs.swmansion.com/cairo-debugger/docs/troubleshooting.html#contract-calls-are-skipped
I can duplicate this info if you want

piotmag769: Done

franciszekjob (approved):

ddoktorski (approved):

PR #4336, opened 2026-05-04
update docs: when incremental compilation is used, integration tests are preferred

### LINKED ISSUE #3567: Show in docs that integration tests are preffered with incremental compilation
### Current State

We should include that information in the docs and link relevant scarb documentation once/if it exists.

### Objective

Show in docs that integration tests are preffered with incremental compilation

### Additional Context

_No response_

DelevoXDG:
Closes #3567 

## Introduced changes

-

## Checklist

- [x] Linked relevant issue
- [ ] Added relevant tests

### DIFF
--- a/docs/src/testing/running-tests.md
+++ b/docs/src/testing/running-tests.md
@@ -73,9 +73,9 @@ Tests: 1 passed, 0 failed, 0 ignored, other filtered out
 
 ## Speeding Up Test Compilation
 
-For best cache reuse across repeat runs, prefer the **integration test layout** (`tests/`) over unit tests in `src/`. 
+For best cache reuse across repeat runs, prefer the **integration test layout** (`tests/`) over unit tests in `src/`.
 
-Thanks to Scarb's [incremental compilation](https://docs.swmansion.com/scarb/docs/procedural-macros/incremental.html) (enabled by default), each `tests/*.cairo` target gets its own cache slot, so editing one test does not invalidate the others. 
+Thanks to Scarb's [incremental compilation](https://docs.swmansion.com/scarb/docs/procedural-macros/incremental.html) (enabled by default), the main package and the test code are cached separately, so editing tests does not invalidate the main package's build cache.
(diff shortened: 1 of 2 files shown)


docs/src/testing/running-tests.md
```diff
@@ -71,6 +71,13 @@ Tests: 1 passed, 0 failed, 0 ignored, other filtered out
 </details>
 <br>
 
+## Speeding Up Test Compilation
+
+For best cache reuse across repeat runs, prefer the **integration test layout** (`tests/`) over unit tests in `src/`. 
+
+Thanks to Scarb's [incremental compilation](https://docs.swmansion.com/scarb/docs/procedural-macros/incremental.html) (enabled by default), each `tests/*.cairo` target gets its own cache slot, so editing one test does not invalidate the others.
```
maciektr: The suggestion is correct, the explanation is not. While it plays nicer with caches to use integration tests layout, all tests from `tests/*.cairo` will be compiled as one unit in the compilation. So the gain is in the ability to skip recompiling the main package you are testing, if only tests change. 
Note this may be subject to changes in the future.

DelevoXDG: Thanks for checking this! I've updated the explanation

> Note this may be subject to changes in the future.

Having separate compilation unit per tests file sounds like good improvement though, right?

I see there's already https://github.com/software-mansion/scarb/issues/588, might be something worthy of another look 🤔

maciektr (approved):

docs/src/testing/running-tests.md
maciektr: Generally yes, but would require some benchmarking prior to know if it makes a difference (and what kind of a difference). Some inefficiencies in the cache layers may cause the difference to drop (for instance, when you are building contracts from dependencies as part of test compilation, etc).

MKowalski8 (approved):

franciszekjob (approved):

PR #4338, opened 2026-05-05
Proper sncast error handling

### LINKED ISSUE #3957: Remove `anyhow::Result` from cast top level functions
### Current State

Currently, if the error is bubbled anywhere in cast, it never gets handled, even at top level (main function). Instead it is thrown by the program causing it to be printed to stdout by default, bypassing foundry UI and it's configuration (e.g. `--json` output)

### Objective

Remove `anhow::Result` from cast top level functions and add handling of errors printing through foundry UI

### Additional Context

_No response_

DelevoXDG:
Closes #3957 

## Introduced changes

- Non-panic errors no longer bypass foundry UI, meaning they will be correctly displayed in line with human-readable / json format.
  So instead of: 
  ```
  Error: Failed to load global config at /Users/runner/.config/starknet-foundry/snfoundry.toml

  Caused by:
      Failed to parse field `block-explorer`: unknown variant `Starkscan`, expected one of `Voyager`, `StarkScan`, `ViewBlock`, `OkLink`
  ```
  we have:
  - in human readable format:
  ```
  Command: show-config
  Error: Failed to load global config at /Users/runner/.config/starknet-foundry/snfoundry.toml
  
  Caused by:
      Failed to parse field `block-explorer`: unknown variant `Starkscan`, expected one of `Voyager`, `StarkScan`, `ViewBlock`, `OkLink`
  ```
  - in json format:
  ```
  {"command":"show-config","error":"Failed to load global config at /Users/runner/.config/starknet-foundry/snfoundry.toml: Failed to parse field `block-explorer`: unknown variant `Starkscan`, expected one of `Voyager`, `StarkScan`, `ViewBlock`, `OkLink`","type":"error"}
  ```
- Build failures now return command errors instead of panicking.
- In command errors, `command` tag now universally displays a full command path (`get tx-status`, `account import`) 
  - previously in some cases, only top-level command name was shown, e.g. `get`, `account`

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - `#[derive(Fuzzable)]` macro that automatically generates `Fuzzable` trait implementations for structs and enums
 
+### Cast
+
+#### Fixed
+
+- Errors no longer bypass foundry UI; `--json` now works for top-level errors where plain text was printed before.
+- Build failures now return command errors instead of panicking.
+- In command errors, `command` tag now universally displays full command path (`get tx-status`, `account import`) (previously in some cases, only top-level command name was shown, e.g. `get`, `account`).
+
 ## [0.60.0] - 2026-04-27
 
 ### Forge
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,7 +37,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 #### Fixed
 
-- Non-panic errors no longer bypass foundry UI; `--json` now works for user-facing errors where plain text was printed before (excluding clap arg-parsing errors).
+- Non-panic errors no longer bypass foundry UI. `--json` now works for user-facing errors where plain text was printed before (excluding clap arg-parsing errors).
 - Build failures now return command errors instead of panicking.
 - In command errors, `command` field now universally displays a full command path (`get tx-status`, `account import`) (previously in some cases, only top-level command name was shown, e.g. `get`, `account`).
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,7 +39,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Non-panic errors no longer bypass foundry UI; `--json` now works for user-facing errors where plain text was printed before (excluding clap arg-parsing errors).
 - Build failures now return command errors instead of panicking.
-- In command errors, `command` tag now universally displays a full command path (`get tx-status`, `account import`) (previously in some cases, only top-level command name was shown, e.g. `get`, `account`).
+- In command errors, `command` field now universally displays a full command path (`get tx-status`, `account import`) (previously in some cases, only top-level command name was shown, e.g. `get`, `account`).
 
 ## [0.60.0] - 2026-04-27
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,9 +37,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 #### Fixed
 
-- Errors no longer bypass foundry UI; `--json` now works for top-level errors where plain text was printed before.
+- Non-panic errors no longer bypass foundry UI; `--json` now works for user-facing errors where plain text was printed before (excluding clap arg-parsing errors).
 - Build failures now return command errors instead of panicking.
-- In command errors, `command` tag now universally displays full command path (`get tx-status`, `account import`) (previously in some cases, only top-level command name was shown, e.g. `get`, `account`).
+- In command errors, `command` tag now universally displays a full command path (`get tx-status`, `account import`) (previously in some cases, only top-level command name was shown, e.g. `get`, `account`).
 
 ## [0.60.0] - 2026-04-27
 
--- a/crates/sncast/src/main.rs
+++ b/crates/sncast/src/main.rs
@@ -859,8 +859,7 @@ fn get_cast_config(cli: &Cli, ui: &UI) -> Result<CastConfig> {
             Sources:
             - CLI flags
             - Local config: {local}
-            - Global config: {global}
-        ",
+            - Global config: {global}",
             local = local_path.as_ref().map_or("missing", |p| p.as_str()),
             global = global_path.as_ref().map_or("missing", |p| p.as_str()),
         }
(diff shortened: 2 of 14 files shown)


CHANGELOG.md
```diff
@@ -13,6 +13,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Bug in forge debugging and `--gas-report` that caused panic in case of a call to non-existent selector
 
+### Cast
+
+#### Fixed
+
+- Non-panic errors no longer bypass foundry UI; `--json` now works for user-facing errors where plain text was printed before (excluding clap arg-parsing errors).
```
DelevoXDG: > Non-panic errors

I assume that is desired since panics are panics, but correct me if I'm wrong

> (excluding clap arg-parsing errors)

That is true because we don't have `--json` arg parsed at the moment of an error. 
We can add logic that will separately determine if json flag is passed, but that will be a bit tricky, plus it seems like that would go beyond the scope of this PR

Thoughts?

crates/sncast/tests/e2e/show_config.rs
```diff
@@ -533,12 +552,14 @@ async fn test_invalid_effective_config() {
...
         output,
         indoc! { r"
+            Command: show-config
             Error: Unable to combine configs. Fix conflicts between config sources and try again.
             Sources:
             - CLI flags
             - Local config: [..]snfoundry.toml
             - Global config: [..]snfoundry.toml
 
+
```
franciszekjob: Why double enter here?

CHANGELOG.md
```diff
@@ -13,6 +13,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- Build failures now return command errors instead of panicking.
+- In command errors, `command` tag now universally displays a full command path (`get tx-status`, `account import`) (previously in some cases, only top-level command name was shown, e.g. `get`, `account`).
```
franciszekjob: nit:
```suggestion
- In command errors, `command` field now universally displays a full command path (`get tx-status`, `account import`) (previously in some cases, only top-level command name was shown, e.g. `get`, `account`).
```

franciszekjob (commented):
It seems to me that there are still panic-based paths that bypass the new human-readable / `--json` error flow. In particular, `script run` still uses `expect(...)` for Scarb build failures in https://github.com/foundry-rs/starknet-foundry/blob/master/crates/sncast/src/starknet_commands/script/mod.rs#L49-L71, and manifest-based commands can still abort in https://github.com/foundry-rs/starknet-foundry/blob/master/crates/sncast/src/helpers/scarb_utils.rs#L65-L67 
because `assert_manifest_path_exists()` calls `get_scarb_manifest().expect(...)`. As a result, some build/setup failures still won’t be reported through the intended structured command error path.

crates/sncast/tests/e2e/show_config.rs
DelevoXDG: It was there before (see `sncast --wait-timeout=1 show-config`), caused by error message new line + `\n\nCaused by:` newlines.

It's just the way `assert_stderr_contains` works that allowed this to stay under radar.

Now, I may:
1. keep it as it
2. revert the test to be as before
3. consider removing extra new line in
```rs
        indoc::formatdoc! {"
            Unable to combine configs. Fix conflicts between config sources and try again.
            Sources:
            - CLI flags
            - Local config: {local}
            - Global config: {global}
        ",
```

Thoughts?

Theoretically, might also be worth looking into `assert_stderr_contains` or using assert_eq so it catches such things - but I'm not sure an extra whitespace is sufficiently important 🤔

DelevoXDG:
> It seems to me that there are still panic-based paths that bypass the new human-readable / `--json` error flow. In particular, `script run` still uses `expect(...)` for Scarb build failures in https://github.com/foundry-rs/starknet-foundry/blob/master/crates/sncast/src/starknet_commands/script/mod.rs#L49-L71,

Good catch, fixed

> and manifest-based commands can still abort in https://github.com/foundry-rs/starknet-foundry/blob/master/crates/sncast/src/helpers/scarb_utils.rs#L65-L67
because assert_manifest_path_exists() calls get_scarb_manifest().expect(...). 

I'm not sure about this one - might be reasonable to keep as panic.

Unlike invalid contract code, `scarb manifest-path` really shouldn't fail and I'm not sure that is something a user can recover from manually 🤔

franciszekjob:
> I'm not sure about this one - might be reasonable to keep as panic.
> 
> Unlike invalid contract code, `scarb manifest-path` really shouldn't fail and I'm not sure that is something a user can recover from manually 🤔

Fair point, let's leave it.

crates/sncast/tests/e2e/show_config.rs
franciszekjob: I see, thx for clarification 👍 . I’d remove the extra newline in the message itself. It makes the output cleaner, and tbh I prefer not to preserve accidental spacing in the test.

> Theoretically, might also be worth looking into assert_stderr_contains or using assert_eq so it catches such things - but I'm not sure an extra whitespace is sufficiently important 🤔

I don’t think this alone justifies changing `assert_stderr_contains` for now.

CHANGELOG.md
```suggestion
- Non-panic errors no longer bypass foundry UI. `--json` now works for user-facing errors where plain text was printed before (excluding clap arg-parsing errors).
```

franciszekjob: > That is true because we don't have --json arg parsed at the moment of an error.
We can add logic that will separately determine if json flag is passed, but that will be a bit tricky, plus it seems like that would go beyond the scope of this PR

Separate PR is fine, though I'm not sure it's super needed ATM (considering it's tricky so may take some time).

franciszekjob (approved):

crates/sncast/src/main.rs
```diff
@@ -308,22 +286,53 @@ fn init_logging() {
         .expect("could not set up global logger");
 }
 
-fn main() -> Result<ExitCode> {
+fn main() -> ExitCode {
     init_logging();
 
-    let cli = Cli::parse();
+    let matches = Cli::command().get_matches();
+    // Avoids duplicate arguments parsing
```
ddoktorski: nit: This comment makes sense taking into consideration the current change (from `Cli::parse()`), but it doesn't seem very useful out of this context

ddoktorski (approved):
LGTM

CHANGELOG.md
DelevoXDG: https://github.com/foundry-rs/starknet-foundry/issues/4363

PR #4341, opened 2026-05-06
Implement custom cache path

### LINKED ISSUE #3425: Allow setting cache path
### Select Tool

snforge

### Requested Feature

Allow specifying caches path by env variable.

franciszekjob:
Closes #3425

## Introduced changes

Allow setting custom cache path via `SNFOUNDRY_CACHE` env.

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests

### DIFF
--- a/crates/forge-runner/src/lib.rs
+++ b/crates/forge-runner/src/lib.rs
@@ -5,7 +5,7 @@ use crate::test_case_summary::TestCaseSummary;
 use anyhow::{Result, bail};
 use build_trace_data::save_trace_data;
 use cairo_lang_sierra::program::{ConcreteTypeLongId, Function, TypeDeclaration};
-use camino::Utf8PathBuf;
+use camino::{Utf8Path, Utf8PathBuf};
 use cheatnet::runtime_extensions::forge_config_extension::config::RawFuzzerConfig;
 use foundry_ui::UI;
 use foundry_ui::components::warning::WarningMessage;
@@ -45,7 +45,7 @@ pub mod tests_summary;
 
 pub const DEFAULT_CACHE_DIR: &str = ".snfoundry_cache";
 
-pub fn resolve_cache_dir(workspace_root: &Utf8PathBuf) -> Utf8PathBuf {
+pub fn resolve_cache_dir(workspace_root: &Utf8Path) -> Utf8PathBuf {
     let cache_dir = env::var("SNFOUNDRY_CACHE").unwrap_or_else(|_| DEFAULT_CACHE_DIR.to_string());
     workspace_root.join(cache_dir)
 }
--- a/crates/forge-runner/src/lib.rs
+++ b/crates/forge-runner/src/lib.rs
@@ -309,8 +309,8 @@ mod tests {
 
     #[test]
     fn rejects_empty_string() {
-        let err = resolve_cache_dir_impl(Utf8Path::new("/tmp/workspace"), Ok(String::new()))
-            .unwrap_err();
+        let err =
+            resolve_cache_dir_impl(Utf8Path::new("/tmp/workspace"), Ok(String::new())).unwrap_err();
         assert_eq!(err.to_string(), "SNFOUNDRY_CACHE must be an absolute path");
     }
 }
--- a/crates/forge-runner/src/lib.rs
+++ b/crates/forge-runner/src/lib.rs
@@ -264,3 +264,46 @@ pub fn function_args(
         })
         .collect()
 }
+
+#[cfg(test)]
+mod tests {
+    use super::{DEFAULT_CACHE_DIR, resolve_cache_dir};
+    use camino::Utf8Path;
+    use std::env;
+
+    #[test]
+    fn defaults_to_workspace_subdir_when_var_unset() {
+        // SAFETY: variable only modified within this test and cleaned up afterwards
+        unsafe { env::remove_var("SNFOUNDRY_CACHE") };
+
+        let workspace = Utf8Path::new("/tmp/workspace");
+        assert_eq!(
+            resolve_cache_dir(workspace).unwrap(),
+            workspace.join(DEFAULT_CACHE_DIR)
+        );
+    }
+
+    #[test]
+    fn accepts_absolute_custom_path() {
+        // SAFETY: variable only modified within this test and cleaned up afterwards
+        unsafe { env::set_var("SNFOUNDRY_CACHE", "/var/cache/snfoundry") };
+
+        let resolved = resolve_cache_dir(Utf8Path::new("/tmp/workspace")).unwrap();
+
+        unsafe { env::remove_var("SNFOUNDRY_CACHE") };
+
+        assert_eq!(resolved, Utf8Path::new("/var/cache/snfoundry"));
+    }
+
+    #[test]
+    fn rejects_relative_custom_path() {
+        // SAFETY: variable only modified within this test and cleaned up afterwards
+        unsafe { env::set_var("SNFOUNDRY_CACHE", "relative/cache") };
+
+        let err = resolve_cache_dir(Utf8Path::new("/tmp/workspace")).unwrap_err();
+
+        unsafe { env::remove_var("SNFOUNDRY_CACHE") };
+
+        assert!(err.to_string().contains("absolute path"));
+    }
+}
--- a/crates/forge-runner/src/lib.rs
+++ b/crates/forge-runner/src/lib.rs
@@ -47,7 +47,14 @@ pub mod tests_summary;
 pub const DEFAULT_CACHE_DIR: &str = ".snfoundry_cache";
 
 pub fn resolve_cache_dir(workspace_root: &Utf8Path) -> Result<Utf8PathBuf> {
-    match env::var("SNFOUNDRY_CACHE") {
+    resolve_cache_dir_impl(workspace_root, env::var("SNFOUNDRY_CACHE"))
+}
+
+fn resolve_cache_dir_impl(
+    workspace_root: &Utf8Path,
+    cache_var: Result<String, VarError>,
+) -> Result<Utf8PathBuf> {
+    match cache_var {
         Ok(cache_dir) => {
             let cache_dir = Utf8PathBuf::from(cache_dir);
             ensure!(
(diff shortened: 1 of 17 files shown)


franciszekjob:

- #4341 ◀
    - #4342

franciszekjob:
note: I didn't mention the possibility of setting env in `clean-cache` docs, because the command removed in the next PR (due to deprecation).

crates/forge/src/clean.rs
```diff
@@ -40,28 +40,25 @@ pub fn clean(args: CleanArgs, ui: &UI) -> Result<()> {
...
-        clean_dir(root_dir, dir_name, ui)?;
-    }
-    Ok(())
-}
-fn clean_dir(dir: &Utf8PathBuf, dir_name: &str, ui: &UI) -> Result<()> {
-    let dir = dir.join(dir_name);
-    if dir.exists() {
-        fs::remove_dir_all(&dir).with_context(|| format!("Failed to remove directory: {dir}"))?;
-        ui.println(&format!("Removed directory: {dir}"));
+fn clean_dir(path: &Utf8PathBuf, ui: &UI) -> Result<()> {
```
MKowalski8: nit: It's better to use `Utf8Path` if possible
```suggestion
fn clean_dir(path: &Utf8Path, ui: &UI) -> Result<()> {
```

crates/forge-runner/src/lib.rs
```diff
@@ -42,7 +43,13 @@ pub mod test_case_summary;
 pub mod test_target_summary;
 pub mod tests_summary;
 
-pub const CACHE_DIR: &str = ".snfoundry_cache";
+pub const DEFAULT_CACHE_DIR: &str = ".snfoundry_cache";
+
+#[must_use]
+pub fn resolve_cache_dir(workspace_root: &Utf8Path) -> Utf8PathBuf {
+    let cache_dir = env::var("SNFOUNDRY_CACHE").unwrap_or_else(|_| DEFAULT_CACHE_DIR.to_string());
+    workspace_root.join(cache_dir)
```
MKowalski8: Could we add a test that check absoulte path?

Or at lest add a comment here, sth like `// If SNFOUNDRY_CACHE is absolute, join() discards workspace_root`

For me it was not obvious that it work like this, so it might be helpful.

docs/src/appendix/snforge/clean.md
```diff
@@ -8,7 +8,12 @@ Clean all generated directories
 
 ## `cache`
 
-Clean the `.snfoundry_cache` directory.
+Clean the cache directory.
+
+> ℹ️ **Info**
+>
+> The default cache directory is `.snfoundry_cache`.
+> You can override it with the `SNFOUNDRY_CACHE` environment variable.
```
MKowalski8: We could add also info that it is relative to workspace root. 

For example some env in cargo is relative to CWD, so it's not obvious. 

`CARGO_TARGET_DIR — Location of where to place all generated artifacts, relative to the current working directory. See [build.target-dir](https://doc.rust-lang.org/cargo/reference/config.html#buildtarget-dir) to set via config.`

https://doc.rust-lang.org/cargo/reference/environment-variables.html

BTW. We could add `Environment Variables` page in appendix as well, it will make things easier. Added issue for that #4347

crates/forge-runner/src/lib.rs
ddoktorski: I believe `SNFOUNDRY_CACHE` should always be an absolute path, otherwise we introduce a lot of ambiguity

crates/forge/tests/e2e/clean.rs
```diff
@@ -156,6 +194,47 @@ fn test_clean_all() {
     );
 }
 
+#[test]
+#[cfg_attr(
+    feature = "cairo-native",
+    ignore = "Native doesn't support trace, coverage and profiler yet"
```
ddoktorski: nit:
```suggestion
    ignore = "Native doesn't support trace, coverage and profiler"
```

crates/forge-runner/src/lib.rs
franciszekjob: I think comment should be enough. Added.

crates/forge/src/clean.rs
```diff
@@ -40,28 +40,25 @@ pub fn clean(args: CleanArgs, ui: &UI) -> Result<()> {
...
-    Ok(())
-}
-fn clean_dir(dir: &Utf8PathBuf, dir_name: &str, ui: &UI) -> Result<()> {
-    let dir = dir.join(dir_name);
-    if dir.exists() {
-        fs::remove_dir_all(&dir).with_context(|| format!("Failed to remove directory: {dir}"))?;
-        ui.println(&format!("Removed directory: {dir}"));
+fn clean_dir(path: &Utf8Path, ui: &UI) -> Result<()> {
+    if path.exists() {
+        fs::remove_dir_all(path).with_context(|| format!("Failed to remove directory: {path}"))?;
```
ddoktorski, line 70: I think we should be more cautious when removing files in a custom path. What about first removing files that match the layout and then removing the dir if it's empty?

docs/src/appendix/snforge/clean.md
franciszekjob: > BTW. We could add Environment Variables page in appendix as well, it will make things easier. Added issue for that https://github.com/foundry-rs/starknet-foundry/issues/4347

Good idea

crates/forge/tests/e2e/forking.rs
```diff
@@ -102,11 +111,64 @@ fn with_clean_cache() {
...
+
+    let default_cache_dir = temp.path().join(DEFAULT_CACHE_DIR);
+    let custom_cache_dir = temp.path().join("custom_cache");
+    fs::create_dir(&custom_cache_dir).unwrap();
+
+    for entry in fs::read_dir(&default_cache_dir).unwrap() {
+        let entry = entry.unwrap();
+        fs::copy(entry.path(), custom_cache_dir.join(entry.file_name())).unwrap();
+    }
+    fs::remove_dir_all(&default_cache_dir).unwrap();
```
ddoktorski:
```suggestion
    fs::rename(&default_cache_dir, &custom_cache_dir).unwrap();
```

MKowalski8 (approved):

ddoktorski (changes requested):

crates/forge/src/clean.rs
MKowalski8: I agree, I just noticed that too. Someone could unknowingly set their cache path to the `home directory` or so.

franciszekjob: Good catch. Changed and added a test which checks such scenario.

crates/forge/src/clean.rs
```diff
@@ -31,6 +33,7 @@ pub fn clean(args: CleanArgs, ui: &UI) -> Result<()> {
         ..MetadataOpts::default()
     })?;
     let workspace_root = scarb_metadata.workspace.root;
+    let has_custom_cache_dir = env::var("SNFOUNDRY_CACHE").is_ok();
```
ddoktorski: It can be checked in `clean_cache_dir` function without passing it as an argument

crates/forge/src/clean.rs
```diff
@@ -40,29 +43,96 @@ pub fn clean(args: CleanArgs, ui: &UI) -> Result<()> {
...
+        if is_snfoundry_cache_file(entry_path) {
+            fs::remove_file(entry_path)
+                .with_context(|| format!("Failed to remove cache file: {entry_path}"))?;
+            ui.println(&format!("Removed file: {entry_path}"));
+        }
+    }
+
+    let is_empty = path
+        .read_dir_utf8()
+        .with_context(|| format!("Failed to read cache directory: {path}"))?
```
ddoktorski: nit: We can avoid reading the same directory twice, one idea is to add a bool variable that will be set to true in the loop if there is any non snfoundry cache file:
```rust
  if is_snfoundry_cache_file(entry_path) {
      fs::remove_file(entry_path)
          .with_context(|| format!("Failed to remove cache file:
{entry_path}"))?;
      ui.println(&format!("Removed file: {entry_path}"));
  } else {
      any_skipped = true;
  }
```

crates/forge/src/clean.rs
```diff
@@ -40,29 +43,96 @@ pub fn clean(args: CleanArgs, ui: &UI) -> Result<()> {
...
+
+    if is_empty {
+        fs::remove_dir(path).with_context(|| format!("Failed to remove directory: {path}"))?;
+        ui.println(&format!("Removed directory: {path}"));
     }
 
     Ok(())
 }
+
+fn is_snfoundry_cache_file(path: &Utf8Path) -> bool {
```
ddoktorski, line 107: Let's add some unit tests for this function

crates/forge-runner/src/lib.rs
```diff
@@ -42,7 +43,21 @@ pub mod test_case_summary;
...
+    match env::var("SNFOUNDRY_CACHE") {
+        Ok(cache_dir) => {
+            let cache_dir = Utf8PathBuf::from(cache_dir);
+            ensure!(
+                cache_dir.is_absolute(),
+                "SNFOUNDRY_CACHE must be an absolute path"
+            );
+            Ok(cache_dir)
+        }
+        Err(_) => Ok(workspace_root.join(DEFAULT_CACHE_DIR)),
```
ddoktorski: Shouldn't we distinguish between different errors `NotPresent` and `NotUnicode`? In the second case I think it should be a hard error

crates/forge/src/clean.rs
```diff
@@ -40,29 +43,96 @@ pub fn clean(args: CleanArgs, ui: &UI) -> Result<()> {
...
+        return false;
+    };
+    let Some((sanitized_url, block_number)) = prefix.rsplit_once('_') else {
+        return false;
+    };
+
+    !sanitized_url.is_empty()
+        && !block_number.is_empty()
+        && block_number.chars().all(|char| char.is_ascii_digit())
+        && version.split('_').count() >= 3
```
ddoktorski: Why `>=` not `==`?

ddoktorski: Why not to use regex matching for this logic?

crates/forge/src/clean.rs
```diff
@@ -1,13 +1,15 @@
...
-use camino::Utf8PathBuf;
+use camino::{Utf8Path, Utf8PathBuf};
+use forge_runner::resolve_cache_dir;
 use foundry_ui::UI;
 use scarb_api::metadata::{MetadataOpts, metadata_with_opts};
+use std::env;
 use std::fs;
 
 const COVERAGE_DIR: &str = "coverage";
+const FILE_WITH_PREV_TESTS_FAILED: &str = ".prev_tests_failed";
```
ddoktorski: This const is duplicated now within the same crate

crates/forge/src/clean.rs
franciszekjob: c2ffdcb7cf8091e9b5ff05e07bd63f96fd1a235f

franciszekjob: Not relevant, cause we use regex now.

crates/forge-runner/src/lib.rs
```diff
@@ -245,3 +264,48 @@ pub fn function_args(
...
+        assert_eq!(
+            resolve_cache_dir(workspace).unwrap(),
+            workspace.join(DEFAULT_CACHE_DIR)
+        );
+    }
+
+    #[test]
+    fn accepts_absolute_custom_path() {
+        // SAFETY: variable only modified within this test and cleaned up afterwards
+        unsafe { env::set_var("SNFOUNDRY_CACHE", "/var/cache/snfoundry") };
```
ddoktorski: With this approach it seems that there can be race conditions as different tests require different cache dirs. What about splitting `resolve_cache_dir`  to public and private function, for example:
```rust
  pub fn resolve_cache_dir(workspace_root: &Utf8Path) -> Result<Utf8PathBuf> {
      resolve_cache_dir_impl(workspace_root, env::var("SNFOUNDRY_CACHE"))
  }
```
and then test here only `resolve_cache_dir_impl`?

crates/forge/src/clean.rs
```diff
@@ -40,29 +51,122 @@ pub fn clean(args: CleanArgs, ui: &UI) -> Result<()> {
...
-    if dir.exists() {
-        fs::remove_dir_all(&dir).with_context(|| format!("Failed to remove directory: {dir}"))?;
-        ui.println(&format!("Removed directory: {dir}"));
+
+pub fn clean_cache_dir(path: &Utf8Path, ui: &UI) -> Result<()> {
+    if !path.exists() {
+        return Ok(());
+    }
+
+    let has_custom_cache_dir = env::var("SNFOUNDRY_CACHE").is_ok();
```
ddoktorski: To simplify this logic we can always check content as if it were a custom path. Alternatively `resolve_cache_dir` could return a tuple containing a boolean to indicate if it's a custom path

franciszekjob: Done, went with always checking content.
https://github.com/foundry-rs/starknet-foundry/pull/4341/commits/7257b5854d0bc2a0e749cb217e6034d51aa84684

crates/forge-runner/src/lib.rs
franciszekjob: https://github.com/foundry-rs/starknet-foundry/pull/4341/commits/26dd4fe69dc33afef99237183f22d33fcd709223

docs/src/appendix/snforge/clean.md
```diff
@@ -8,7 +8,12 @@ Clean all generated directories
 
 ## `cache`
 
-Clean the `.snfoundry_cache` directory.
+Clean the cache directory.
+
+> ℹ️ **Info**
+>
+> The default cache directory is `.snfoundry_cache`, relative to the workspace root.
+> You can override it with the `SNFOUNDRY_CACHE` environment variable, which must be an absolute path.
```
ddoktorski: nit:
```suggestion
> You can override it with the `SNFOUNDRY_CACHE` environment variable, which must be an absolute path to a dedicated directory.
```

crates/forge-runner/src/lib.rs
```diff
@@ -245,3 +271,39 @@ pub fn function_args(
         })
         .collect()
 }
+
+#[cfg(test)]
+mod tests {
```
ddoktorski, line 276: Let's add a test to make sure that it fails if `SNFOUNDRY_CACHE` is set to an empty string

crates/forge-runner/src/lib.rs
```diff
@@ -245,3 +271,39 @@ pub fn function_args(
...
+    }
+
+    #[test]
+    fn rejects_relative_custom_path() {
+        let err = resolve_cache_dir_impl(
+            Utf8Path::new("/tmp/workspace"),
+            Ok("relative/cache".to_string()),
+        )
+        .unwrap_err();
+        assert!(err.to_string().contains("absolute path"));
```
ddoktorski: nit: Assert exact full string

ddoktorski (approved):

PR #4353, opened 2026-05-11
Fix incorrect address being written to keystore account file

### LINKED ISSUE #4345: Keystore address mismatch and bugs
1. Keystore address displayed for user and the one saved to file are different. 

2. As well, this message is wrong, `--keystore` argument is missing
```
To deploy a contract of this class, run:
sncast --account file.json deploy --class-hash 0x5ca3731e8eb5491a840e2db274b27240db243759f954ecf0016ca56d3588299 --network devnet
```

MKowalski8:
Closes #4345

## Introduced changes

- Now address that is saved to file is taken from `factory`. 
- Also wrong helper after `declare` was fixed, the `keystore` argument was missing

## Checklist

- [x] Linked relevant issue
- [ ] Added relevant tests

### DIFF
--- a/crates/sncast/src/response/declare.rs
+++ b/crates/sncast/src/response/declare.rs
@@ -77,6 +77,7 @@ impl OutputLink for DeclareTransactionResponse {
 #[derive(Clone, Serialize, Deserialize, Debug, PartialEq)]
 pub struct DeployCommandMessage {
     accounts_file: Option<String>,
+    keystore: Option<String>,
     account: String,
     class_hash: PaddedFelt,
     arguments_flag: Option<String>,
@@ -89,6 +90,7 @@ impl DeployCommandMessage {
         response: &DeclareTransactionResponse,
         account: &str,
         accounts_file: &Utf8PathBuf,
+        keystore: Option<&Utf8PathBuf>,
         network_flag: String,
     ) -> Result<Self, Error> {
         let arguments_flag: Option<String> = generate_arguments_flag(abi);
@@ -100,6 +102,7 @@ impl DeployCommandMessage {
         Ok(Self {
             account: account.to_string(),
             accounts_file,
+            keystore: keystore.map(std::string::ToString::to_string),
             class_hash: response.class_hash,
             arguments_flag,
             network_flag,
@@ -114,6 +117,10 @@ impl Message for DeployCommandMessage {
         let accounts_file_flag = generate_accounts_file_flag(self.accounts_file.as_ref());
         if let Some(flag) = accounts_file_flag {
             write!(command, " {flag}").unwrap();
+        }
+
+        if let Some(keystore) = &self.keystore {
+            write!(command, " --keystore {keystore}").unwrap();
         }
 
         let account_flag = format!("--account {}", self.account);
--- a/crates/sncast/src/response/declare.rs
+++ b/crates/sncast/src/response/declare.rs
@@ -114,13 +114,11 @@ impl Message for DeployCommandMessage {
     fn text(&self) -> String {
         let mut command = String::from("sncast");
 
-        let accounts_file_flag = generate_accounts_file_flag(self.accounts_file.as_ref());
-        if let Some(flag) = accounts_file_flag {
-            write!(command, " {flag}").unwrap();
-        }
-
+        // TODO(#4367): --accounts-file and --keystore should be mutually exclusive at the CLI level
         if let Some(keystore) = &self.keystore {
             write!(command, " --keystore {keystore}").unwrap();
+        } else if let Some(flag) = generate_accounts_file_flag(self.accounts_file.as_ref()) {
+            write!(command, " {flag}").unwrap();
         }
 
         let account_flag = format!("--account {}", self.account);
--- a/crates/sncast/src/starknet_commands/account/import.rs
+++ b/crates/sncast/src/starknet_commands/account/import.rs
@@ -159,7 +159,7 @@ pub async fn import(
         .await?;
         ensure!(
             computed_address == import.address,
-            "Computed address {:#x} does not match the provided address {:#x}. Please ensure that the provided salt, class hash, and account type are correct.",
+            "Computed address {:#x} does not match the provided address {:#x}. Please ensure that the provided salt, class hash, and account type are correct. Note: for Ready accounts, address validation only supports accounts deployed with guardian set to 0.",
             computed_address,
             import.address
         );
--- a/crates/sncast/src/starknet_commands/account/import.rs
+++ b/crates/sncast/src/starknet_commands/account/import.rs
@@ -159,7 +159,7 @@ pub async fn import(
         .await?;
         ensure!(
             computed_address == import.address,
-            "Computed address {:#x} does not match the provided address {:#x}. Please ensure that the provided salt, class hash, and account type are correct. Note: for Ready accounts, address validation only supports accounts deployed with guardian set to 0.",
+            "Computed address {:#x} does not match the provided address {:#x}. Please ensure that the provided salt, class hash, and account type are correct.",
             computed_address,
             import.address
         );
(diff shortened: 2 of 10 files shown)


franciszekjob (commented):
Changes look good. Two comments regarding test coverage:

1. We have [such test](https://github.com/foundry-rs/starknet-foundry/blob/7086c7e0a750892550047da96a3aad7b5a4efb61/crates/sncast/tests/e2e/main_tests.rs#L360) and it doesn't check for the hint deploy command which is printed out as a follow up after declaration. Could you please add assertion for it too?
2. In [`test_happy_case_keystore`](https://github.com/foundry-rs/starknet-foundry/blob/7086c7e0a750892550047da96a3aad7b5a4efb61/crates/sncast/tests/e2e/account/deploy.rs#L320) we check whether the address is not null. This should be changed too, so instead of 

```rust
assert!(!items["deployment"]["address"].is_null());
```

we do

```rust
assert_eq!(items["deployment"]["address"], address.into_hex_string());
```

crates/sncast/src/starknet_commands/account/import.rs
```diff
@@ -204,6 +205,34 @@ fn get_private_key_from_input() -> Result<Felt> {
     parse_input_to_felt(&input)
 }
 
+fn compute_account_address(
```
ddoktorski: This function is not very reliable as it requires to be aligned with factories logic. As long as we use it only for validation in `account import` it should be fine, but please add some comments with explanation. Other thing is that there is a significant limitation in terms of allowing to import only ready accounts that use default Starknet signer. Could you check if we have it clearly highlighted in the docs and whether it makes sense to add it in some error message?

ddoktorski (commented):
Please re-request once @franciszekjob suggestion is applied

crates/sncast/src/starknet_commands/account/import.rs
DelevoXDG: > This function is not very reliable as it requires to be aligned with factories logic. 

To avoid having future bugs and  misalignments with factories logic, should we consider just re-using it here instead of doing manual calculations? 🤔 

something like:
```rs
    let signer = match &signer_type {
        SignerType::Ledger { ledger_path } => { ... }
        SignerType::Local { private_key } => { ... }
    };

    ...
    match account_type {
        AccountType::Ready => {
            let factory = ArgentAccountFactory::new(class_hash, chain_id, None, signer, provider).await?;
            factory.deploy_v3(salt).address()
        }
        AccountType::OpenZeppelin => {
            let factory = OpenZeppelinAccountFactory::new(class_hash, chain_id, signer, provider).await?;
            factory.deploy_v3(salt).address()
        }
        AccountType::Braavos => {
            let factory = BraavosAccountFactory::new(class_hash, BRAAVOS_BASE_ACCOUNT_CLASS_HASH, chain_id, signer, provider).await?;
            factory.deploy_v3(salt).address()
        }
    }
    ...
```

crates/sncast/src/response/declare.rs
```diff
@@ -116,6 +119,10 @@ impl Message for DeployCommandMessage {
             write!(command, " {flag}").unwrap();
         }
 
+        if let Some(keystore) = &self.keystore {
+            write!(command, " --keystore {keystore}").unwrap();
+        }
```
DelevoXDG, line 122: Perhaps worth a mention in a changelog as well?

MKowalski8: It's a very minor fix, so I personally wouldn't add it, but if anyone else also feels that it shoud be there, I can mention it

crates/sncast/src/starknet_commands/account/import.rs
MKowalski8: > To avoid having future bugs and misalignments with factories logic, should we consider just re-using it here instead of doing manual calculations? 🤔

I haven't done it in the first place, because I thought it will take more time to do, but wasn't so hard after all. Done

MKowalski8: > Other thing is that there is a significant limitation in terms of allowing to import only ready accounts that use default Starknet signer. Could you check if we have it clearly highlighted in the docs and whether it makes sense to add it in some error message?

With change above it now supports ledger signer as well. I added info that we support `ready` accounts with `guardian` set to 0. (The same info is in `account create`). 
However I will look into this during #3994, becuase I don't know why this check is only with `--salt`. I once created an account with `--salt` imported it without and it worked, so it might be unnecessary.

crates/sncast/src/response/declare.rs
```diff
@@ -111,8 +114,10 @@ impl Message for DeployCommandMessage {
     fn text(&self) -> String {
         let mut command = String::from("sncast");
 
-        let accounts_file_flag = generate_accounts_file_flag(self.accounts_file.as_ref());
-        if let Some(flag) = accounts_file_flag {
+        // TODO(#4367): --accounts-file and --keystore should be mutually exclusive at the CLI level
```
MKowalski8, line 117: Note: I added task, because this can confuse users, that we can pass `--accounts-file` and it does nothing (when `--keystore` is present).

crates/sncast/src/starknet_commands/account/import.rs
```diff
@@ -147,11 +147,19 @@ pub async fn import(
...
+            chain_id,
+            &signer_type,
+            provider,
+            ui,
+        )
+        .await?;
         ensure!(
             computed_address == import.address,
-            "Computed address {:#x} does not match the provided address {:#x}. Please ensure that the provided salt, class hash, and account type are correct.",
+            "Computed address {:#x} does not match the provided address {:#x}. Please ensure that the provided salt, class hash, and account type are correct. Note: for Ready accounts, address validation only supports accounts deployed with guardian set to 0.",
```
franciszekjob:
```suggestion
            "Computed address {:#x} does not match the provided address {:#x}. Please ensure that the provided salt, class hash, and account type are correct. Note: for Ready accounts, address validation only supports accounts deployed with Guardian Key set to 0.",
```

crates/sncast/tests/e2e/account/import.rs
```diff
@@ -706,10 +706,10 @@ pub async fn test_invalid_address_computation() {
     ];
 
     let snapbox = runner(&args).current_dir(tempdir.path());
-    let computed_address = "0xaf550326d32c8106ef08d25cbc0dba06e5cd10a2201c2e9bc5ad4cbbce45e6";
+    let computed_address = "0x7298d9fc4bde13623bd53f4adb0110bd77ab5f3f3675402b5dfb418e149e56a";
     snapbox.assert().stderr_eq(formatdoc! {r"
         Command: account import
-        Error: Computed address {computed_address} does not match the provided address 0x123. Please ensure that the provided salt, class hash, and account type are correct.
+        Error: Computed address {computed_address} does not match the provided address 0x123. Please ensure that the provided salt, class hash, and account type are correct. Note: for Ready accounts, address validation only supports accounts deployed with guardian set to 0.
```
franciszekjob: I think it's not relevant to display the note about Ready account for other account types (oz, braavos).

docs/src/appendix/sncast/account/import.md
```diff
@@ -75,6 +75,8 @@ Optional.
 
 Salt for the account address.
 
+> **Limitation:** For `ready` (Argent) accounts, address validation only supports accounts deployed with guardian set to 0.
```
franciszekjob:
```suggestion
> **Limitation:** For `ready` (formerly `argent`) accounts, address validation only supports accounts deployed with guardian key set to 0.
```

crates/sncast/tests/e2e/account/import.rs
MKowalski8: Okay, I will delete it then. Optionally we can do `if` on `type`

franciszekjob (approved):

crates/sncast/src/response/declare.rs
ddoktorski: We can mention it under `Fixed`

ddoktorski (approved):

PR #4359, opened 2026-05-14
Add workflow to notify about pending reviews

### LINKED ISSUE #4356: Create a slack bot to notify about waiting reviews
### Current State

.

### Objective

.

### Additional Context

_No response_

ddoktorski:
Closes #4356


scripts/notify_pending_reviews.sh
```diff
@@ -0,0 +1,129 @@
...
+    BODY=${RESP%$'\n'*}
+    if [ "$HTTP_CODE" = "200" ]; then
+        echo "✅ Sent to Slack (HTTP $HTTP_CODE)" >&2
+    else
+        echo "❌ Slack webhook failed (HTTP $HTTP_CODE): $BODY" >&2
+        exit 1
+    fi
+else
+    printf '%s\n' "$PAYLOAD"
+    echo "✅ Block Kit payload built (set SLACK_WEBHOOK_URL to post to Slack)" >&2
```
DelevoXDG, line 130: I think we should `exit 1` here, since this should fail on CI if webhook is not set

scripts/notify_pending_reviews_payload.jq
```diff
@@ -0,0 +1,101 @@
...
+                        type: "rich_text_section",
+                        elements: ([
+                          {type: "emoji", name: (.waiting_s | age_name)},
+                          {type: "text",  text: " "},
+                          {type: "text",  text: rpad($w; $wmax), style: {bold: true, code: true}},
+                          {type: "text",  text: "  "},
+                          {type: "link",  url: .url, text: rpad($num; $nmax)},
+                          {type: "text",  text: "  "}
+                        ] + (.title | title_elements) + [
+                          {type: "text",  text: " (\($sz))\n"}
```
DelevoXDG: I think this would be a bit more readable
```suggestion
                          {type: "text", text: " "},
                          {type: "text", text: "(\($sz))", style: {bold: true}}
```

.github/workflows/notify-pending-reviews.yml
```diff
@@ -0,0 +1,21 @@
+name: 'Notify about pending review requests'
+
+on:
+  schedule:
+    # Monday, Wednesday, Friday at 06:00 UTC.
+    - cron: '0 6 * * 1,3,5'
```
DelevoXDG: As previously discussed, I think we should post it daily

scripts/notify_pending_reviews_payload.jq
DelevoXDG, line 1: Could be also great to display review progress (approves, comments, etc.).

What I have in mind, more or less:

Image

scripts/notify_pending_reviews_payload.jq
```diff
@@ -0,0 +1,101 @@
...
+  elif . < 86400 then "\(. / 3600 | floor)h"
+  else "\(. / 86400 | floor)d"
+  end;
+
+# Age dot: green < 2d, yellow 2-4d, red > 4d.
+def age_name:
+  if . == null then "grey_question"
+  elif . < 172800 then "large_green_circle"
+  elif . < 345600 then "large_yellow_circle"
+  else "red_circle"
```
DelevoXDG, line 26: `🔴 5d` and `🔴 33d` seem a bit non-equivalent.

I think there's a place for 🟠 (e.g. 4-8 days).

scripts/notify_pending_reviews.sh
```diff
@@ -0,0 +1,129 @@
...
+done
+
+echo "Fetching pending review requests from: $REPO..." >&2
+
+# Pull current pending reviewers + ReviewRequestedEvent timestamps via GraphQL.
+# shellcheck disable=SC2016  # $owner/$name are GraphQL variables, not shell.
+QUERY='
+query($owner: String!, $name: String!) {
+  repository(owner: $owner, name: $name) {
+    pullRequests(first: 50, states: OPEN, orderBy: {field: UPDATED_AT, direction: DESC}) {
```
DelevoXDG: I suppose this is fine and we realistically wouldn't cross this limit, but perhaps we can add a const at the top of the file so we're aware of this

scripts/notify_pending_reviews_payload.jq
DelevoXDG: Can skip though if it's too big of an effort 😅

DelevoXDG (commented):
Looks great!

scripts/notify_pending_reviews.sh
```diff
@@ -0,0 +1,129 @@
...
+      | sort_by(.reviewer, -(.waiting_s // -1))
+    ')
+
+# Build the Block Kit payload.
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+PAYLOAD=$(jq -f "$SCRIPT_DIR/notify_pending_reviews_payload.jq" --arg repo "$REPO" <<<"$ROWS_JSON")
+
+# Send to Slack if a webhook is configured, otherwise print the payload.
+if [ -n "$SLACK_WEBHOOK_URL" ]; then
+    RESP=$(curl -sS -w $'\n%{http_code}' -X POST \
```
franciszekjob: nit: Just in case, if slack webhook is unresponsive:
```suggestion
    RESP=$(curl -sS --max-time 30 -w $'\n%{http_code}' -X POST \
```

scripts/notify_pending_reviews_payload.jq
ddoktorski: Let's skip it for now to have the script already running "in production". We can modify it any moment

scripts/notify_pending_reviews_payload.jq
franciszekjob: Honestly, I don't think we need to keep so many levels of distinction 😅 . I personally prefer a 🔴 for everything  > 4 days, since it's already long.

scripts/notify_pending_reviews.sh
```diff
@@ -0,0 +1,129 @@
...
+    }
+  }
+}
+'
+
+# Emit a JSON array, one object per (reviewer, PR) pair.
+# Sorted by reviewer asc, then waiting time desc.
+ROWS_JSON=$(gh api graphql -f query="$QUERY" \
+        -F owner="${REPO%/*}" -F name="${REPO#*/}" \
+    | jq --arg ignore_team "$IGNORE_TEAM" '
```
franciszekjob, line 80: nit: Let's fail fast in case when API returns an error:

```suggestion
    | jq --arg ignore_team "$IGNORE_TEAM" '
      if .errors then error("GitHub API error: \(.errors | tostring)") else . end
      | [
```

.github/workflows/notify-pending-reviews.yml
```diff
@@ -0,0 +1,21 @@
+name: 'Notify about pending review requests'
+
+on:
+  schedule:
+    # Monday to Friday at 06:00 UTC.
+    - cron: '0 6 * * 1-5'
+  workflow_dispatch:
```
franciszekjob, line 7: nit: Doesn't make much difference, but should it stay?

scripts/notify_pending_reviews.sh
```diff
@@ -0,0 +1,132 @@
...
+# Otherwise the JSON payload is printed to stdout (useful for previewing
+# in Slack's Block Kit Builder: https://app.slack.com/block-kit-builder).
+#
+# Requires: gh (authenticated), jq, and curl (only when posting).
+# Block Kit construction lives in notify_pending_reviews_payload.jq.
+
+set -euo pipefail
+
+REPO="foundry-rs/starknet-foundry"
+IGNORE_TEAM="starknet-foundry"
```
franciszekjob, line 19: very nit: `IGNORE_TEAM` seems more like a name for boolean variable:

```suggestion
IGNORED_TEAM="starknet-foundry"
```

franciszekjob (approved):
👌

scripts/notify_pending_reviews_payload.jq
DelevoXDG: Sure, added issue so we can consider this in the future: https://github.com/foundry-rs/starknet-foundry/issues/4370

.github/workflows/notify-pending-reviews.yml
DelevoXDG: why not?

DelevoXDG (approved):
🚀

PR #4361, opened 2026-05-15
Don't return unnecessary diagnostics from macros

piotmag769:
## Introduced changes

- `check_error_free` instead of `is_empty` - https://github.com/foundry-rs/starknet-foundry/pull/4240#discussion_r3242641239
- stopped returning additional diags when parser error happen - this makes the UX better while writing the code. Now:

image

vs before: 

image

## Checklist

- [ ] Linked relevant issue
- [ ] Added relevant tests

### DIFF
--- a/crates/snforge-scarb-plugin/src/derives/fuzzable.rs
+++ b/crates/snforge-scarb-plugin/src/derives/fuzzable.rs
@@ -15,9 +15,8 @@ pub fn fuzzable_derive(item: &TokenStream) -> ProcMacroResult {
     let db = SimpleParserDatabase::default();
     let (parsed, diagnostics) = db.parse_token_stream(item);
 
-    if !diagnostics.is_empty() {
-        return ProcMacroResult::new(TokenStream::empty())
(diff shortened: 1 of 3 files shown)


crates/snforge-scarb-plugin/src/derives/fuzzable.rs
DelevoXDG, line 1: https://github.com/foundry-rs/starknet-foundry/blob/fb41b9979e92b9d3af2b87de372192961352546f/crates/forge/tests/e2e/plugin_diagnostics.rs#L81

I think these tests should be updated

Are you sure there are no contexts in which we do want this diagnostic to be emitted? 🤔

piotmag769: >I think these tests should be updated

True, realised after CI xD

>Are you sure there are no contexts in which we do want this diagnostic to be emitted? 

I am sure. It just clutters the output, we return the same item anyways and the parser diagnostic will be present since the regular compilation flow will report it itself

DelevoXDG (approved):

ddoktorski (approved):

Issue #2240, opened 2024-06-20
Add a possibility to save (alias) deployed contracts and declared hashes


Arcticae:
### Which components does the task require to be changed? (think hard pls)

sncast

### Description

It would be easier to refer to contracts in a manner that we can distinguish between calls in a human-readable way. 

i.e.
`sncast call --contract-alias my-swap-contract --function "swap" --calldata 123 123 123`
or 
`sncast declare --class-alias the-custom-account  --constructor-calldata ....`

PR #4365, opened 2026-05-15
Initial alias logic in sncast (`call`, `invoke`)

DelevoXDG:
Towards #2240

## Stack
- #4365 
- #4366
- #4376 
- #4382
- #4383

## Introduced changes

- Initial alias logic setup; `[sncast.<profile>.aliases]` in `snfoundry.toml`
- Support alias in `--contract-address` in `call`, `invoke`
- **tests:** `call` and `invoke` (alias resolves, unknown alias fails)

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/sncast/src/helpers/felt.rs
franciszekjob, line 1: Could we re-use existing conversion, e.g. [`Felt::try_from_hex_str`](https://github.com/foundry-rs/starknet-foundry/blob/05df5734c03748c97e924551da67733cf3a20b1d/crates/conversions/src/felt.rs#L69) or [`Felt::from_short_string`](https://github.com/foundry-rs/starknet-foundry/blob/05df5734c03748c97e924551da67733cf3a20b1d/crates/conversions/src/felt.rs#L85), instead of adding a new helper?

crates/sncast/src/helpers/configuration.rs
```diff
@@ -86,6 +89,39 @@ impl Override for NetworksConfig {
...
+        D: serde::Deserializer<'de>,
+    {
+        BTreeMap::<String, String>::deserialize(deserializer)?
+            .into_iter()
+            .map(|(name, value)| felt_from_string(&value).map(|felt| (name, felt)))
+            .collect::<Result<BTreeMap<_, _>, _>>()
+            .map_err(serde::de::Error::custom)
+            .map(AliasesConfig)
+    }
+}
```
franciszekjob, line 107: Should we also allow for numeric (non-quoted) alias values?

crates/sncast/src/main.rs
```diff
@@ -658,6 +655,10 @@ async fn run_async_command(cli: Cli, config: CastConfig, ui: &UI) -> Result<Exit
         }) => {
             let provider = rpc.get_provider(&config, ui).await?;
 
+            let contract_address = contract_address
+                .resolve_alias_or_felt(&config)
+                .context("Invalid contract address")?;
```
franciszekjob, line 660: It should be validated before resolving provider.

docs/src/starknet/aliases.md
```diff
@@ -0,0 +1,75 @@
+# Aliases
+
+When you use commands like `sncast call`, `sncast invoke`, etc. you typically need to specify some felt values, for example `--contract-address`:
+```shell
+$ sncast call --contract-address 0xcd8f9ab31324bb93251837e4efb4223ee195454f6304fcfcb277e277653008 --function get --calldata 0x0 --block-id latest
+```
+
+Instead, you can create aliases for felt values (contract addresses, class hashes, etc.) in `snfoundry.toml`.
```
franciszekjob: nit: Consistent with naming of below section:
```suggestion
Instead, you can define aliases for felt values (contract addresses, class hashes, etc.) in `snfoundry.toml`.
```

CHANGELOG.md
```diff
@@ -37,6 +37,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `sncast get spec-version` command that returns the version of the Starknet JSON-RPC specification used by the node
 - `--no-abi` flag for `declare`, `declare-from`, and `deploy`, erasing the ABI before class declaration.
 - `sncast get tx-receipt` (alias: `get transaction-receipt`) command that returns the receipt of a transaction
+- Aliases in `snfoundry.toml`. Read more [here](https://foundry-rs.github.io/starknet-foundry/appendix/snfoundry-toml.html#sncastprofile-namealiases).
+- `--contract-address @alias` syntax in `sncast call` and `sncast invoke`.
+
```
```suggestion
- Aliases in `snfoundry.toml`. Read more [here](https://foundry-rs.github.io/starknet-foundry/appendix/snfoundry-toml.html#sncastprofile-namealiases).
- `--contract-address @alias` syntax in `sncast call` and `sncast invoke`.
```

crates/sncast/src/helpers/felt.rs
DelevoXDG: - `try_from_hex_str` expects a `0x..` hex.
- `from_short_string` expects a shortstring.

In commands, clap args allow both hex and decimal values.

So aliases should support decimal values too, and that's why I'm using `Felt::from_str` here:

```rs
    /// Converts a hex (0x-prefixed) or decimal string to a [Felt].
    /// e.g., '0x123abc' or '1337'.
        if s.starts_with("0x") {
            Felt::from_hex(s)
        } else {
            Felt::from_dec_str(s)
        }
    }
```

Besides, `From::str(..).with_context("..")` is already widely used pattern in the codebase.

As for the helper, it just dedupes this calls and allows to unify error message

crates/sncast/src/helpers/configuration.rs
DelevoXDG: Maybe we should, but I'm not sure that even possible considering parser's 64 bit integer limit, which is far lower than a typical felt value.

Considering it's tricky at best, I think it's beyond the scope for initial aliases support 🤔 

Thoughts?

crates/sncast/src/main.rs
DelevoXDG: It does make sense to me to validate before the provider, in a sense that arguments should be validated before we make any calls to the provider.

But I note that all of these functions do:

```rs
let provider = ...
```
before doing anything else, including the validations. 

If we do this change, we should also be moving all of these:

```rs
```

```rs                
let calldata = arguments
    .try_into_calldata(&abi_from_contract_class(&contract_class)?, &selector)?;
```

before resolving provider.

And if that's the case, I believe this should be handled as a separate issue, since it affects all sncast commands and it'd be reasonable to handle it consistently across the board.

But, perhaps it was intentional to always resolve provider first?

CHANGELOG.md
DelevoXDG: What's the difference? 👀

DelevoXDG: Note: I've update the changelog entry to point to the "Aliases" guide rather than `snfoundry.toml#aliases`

franciszekjob: > What's the difference? 👀

No duplicated newline 😅

crates/sncast/src/main.rs
franciszekjob: > And if that's the case, I believe this should be handled as a separate issue, since it affects all sncast commands and it'd be reasonable to handle it consistently across the board.

Sure, makes sense.

> But, perhaps it was intentional to always resolve provider first?

Tbh I don't see a good reason for it.

crates/sncast/src/helpers/configuration.rs
franciszekjob: Fine, let's drop it then.

crates/sncast/src/helpers/felt.rs
```diff
@@ -0,0 +1,7 @@
+use anyhow::{Context, Result};
+use starknet_types_core::felt::Felt;
+use std::str::FromStr;
+
+pub fn felt_from_string(data: &str) -> Result<Felt> {
+    Felt::from_str(data).with_context(|| format!("Failed to parse {data} to felt"))
```
franciszekjob, line 6: Wrapping with single quotes makes the error more readable:

```suggestion
    Felt::from_str(data).with_context(|| format!("Failed to parse '{data}' to felt"))
```

docs/src/starknet/aliases.md
```diff
@@ -0,0 +1,75 @@
...
+TBD
+
+## Supported commands
+
+Currently, aliases are supported for:
+
+- `sncast call --contract-address`
+- `sncast invoke --contract-address`
+
+
```
franciszekjob, line 75: nit:
```suggestion
- `sncast invoke --contract-address`

```

franciszekjob (approved):

crates/sncast/src/helpers/felt.rs
DelevoXDG: Hmmm, I've already wrapped it in the backticks in the next PR 👀 

24334a72aa58bb465cd13d487a615920aba87c1d

is that ok to you?

crates/sncast/src/main.rs
DelevoXDG: https://github.com/foundry-rs/starknet-foundry/issues/4389

docs/src/starknet/aliases.md
DelevoXDG: why add another one - perhaps you meant I should remove one? 

in any case the formatting here should be normalized by the next PR s

franciszekjob: > perhaps you meant I should remove one?

yes, lol 😅

crates/sncast/src/main.rs
```diff
@@ -658,6 +655,10 @@ async fn run_async_command(cli: Cli, config: CastConfig, ui: &UI) -> Result<Exit
...
+
```
MKowalski8, line 661: nit: Do we need this `context` here? Later in [Invoke](https://github.com/foundry-rs/starknet-foundry/blob/00799a6364d2ae7f2422be950a27004bb5da0104/crates/sncast/src/main.rs#L719) we don't have it.  We can unify it either way 

```suggestion
            let contract_address = contract_address.resolve_alias_or_felt(&config)?;
```

docs/src/appendix/sncast/invoke.md
```diff
@@ -8,7 +8,9 @@ Send an invoke transaction to Starknet.
 ## `--contract-address, -d <CONTRACT_ADDRESS>`
 Required.
 
-The address of the contract being called in hex (prefixed with '0x') or decimal representation.
+The address of the contract being called. It can be either:
+- Felt in hex (prefixed with `0x`) or decimal representation.
+- `@alias` defined in `[sncast.<profile>.aliases]` in `snfoundry.toml`. See [aliases](../../starknet/aliases.md)..
```
MKowalski8, line 13: Double dot

```suggestion
- `@alias` defined in `[sncast.<profile>.aliases]` in `snfoundry.toml`. See [aliases](../../starknet/aliases.md).
```

docs/src/starknet/aliases.md
```diff
@@ -0,0 +1,75 @@
...
+> Aliases follow the same rules as other `snfoundry.toml` settings.
+> That means you can define aliases in the global and local config files, and create the per-profile aliases, for example:
+> ```toml
+> [sncast.myprofile]
+> network = "devnet"
+> 
+> [sncast.myprofile.aliases]
+> predeployed-account = "0x691a61b12a7105b1372cc377f135213c11e8400a546f6b0e7ea0296046690ce"
+> ```
+> Aliases follow the same [precedence rules](../projects/configuration.md#interaction-between-local-and-global-profiles) as other `snfoundry.toml` settings: aliases defined in higher-precedence config override lower-precedence ones.
```
MKowalski8: Can we avoid this repetition here? 
`Aliases follow the same rules as other snfoundry.toml settings.`
`Aliases follow the same precedence rules as other snfoundry.toml settings`

Mayby we don't need this split into two paragraphs, but include this two informations in one or so.

MKowalski8 (approved):

docs/src/appendix/sncast/invoke.md
DelevoXDG: Fixed in next PR, to avoid extra rebases I'd prefer to just merge the fix then, if that's ok with you

https://github.com/foundry-rs/starknet-foundry/commit/8c98b613cfc11481186572dd8b743bec672d6256

crates/sncast/src/main.rs
DelevoXDG: I've unified all the `resolve_alias_or_felt(..).context(..)` logic in #4376

docs/src/starknet/aliases.md
DelevoXDG: I've tried to simplify this line, please TAL.

> Mayby we don't need this split into two paragraphs, but include this two informations in one or so.

I've considered that too, but since this is a very minor sub-tip, I feel like after the example is a correct place for it.

MKowalski8: LGTM

PR #4373, opened 2026-05-20
Support strings in tuples for `#[should_panic]` attribute

### LINKED ISSUE #4354: Support regular strings inside `#[should_panic(expected: (...))]` tuples
### Current State

**Context**
While investigating the possibility of testing Cairo corelib with `snforge`, we encountered an issue with tests that use the `#[should_panic]` attribute containing mixed types (including regular strings) in the expected tuple.

Example from corelib:

```
#[test]
#[should_panic(expected: ("error", 11, "hello", 5, 'short_string'))]
fn test_panic_with_stacked_errors() {
    // ...
}
```

**The Problem**
Currently, snforge plugin states that the argument must be a string, short string, number, or a list of short strings/numbers. It does not support regular strings inside a mixed tuple. As a workaround, tests have to be migrated to assert tuple of felts.

**Proposed Solution**
1. Enhance snforge plugin to natively support regular strings inside `expected: (...)` tuples.

or

2. Adjust or simplify some `#[should_panic(expected: ...)]` cases to match the current plugin limitations

### Objective

Support regular strings inside `#[should_panic(expected: (...))]` tuples

### Additional Context

_No response_

franciszekjob:
Closes #4354

## Introduced changes

Add support for regular strings inside mixed tuples (alongside short strings and numbers) in `#[should_panic(expected: (...))]`  attribute.

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/forge/tests/data/should_panic_test/tests/should_panic_test.cairo
```diff
@@ -22,6 +22,21 @@ fn should_panic_multiple_messages() {
     panic(arr);
 }
 
+#[test]
+#[should_panic(expected: ("error", 11, "hello", 5, 'short_string'))]
```
ddoktorski: nit: Let's make one of the strings longer than a felt to test that it's really treated as a bytearray and not as a felt

crates/snforge-scarb-plugin/src/attributes/should_panic/expected.rs
```diff
@@ -77,3 +104,23 @@ impl ParseFromExpr<Expr> for Expected {
...
+        expr: &Expr,
+        arg_name: &str,
+    ) -> Result<Self, Diagnostic> {
+        match expr {
+            Expr::ShortString(_) | Expr::Literal(_) => {
+                Ok(Self::Felt(Felt::parse_from_expr::<T>(db, expr, arg_name)?))
+            }
+            Expr::String(string) => Ok(Self::ByteArray(
+                string.text(db).trim_matches('"').to_string(),
+            )),
```
ddoktorski, line 118: nit: Does it make sense to use this instead?
```suggestion
            Expr::String(string) => {
                let string = string.string_value(db).ok_or_else(|| {
                    T::error(format!("<{arg_name}> contains invalid string literal"))
                })?;
                Ok(Self::ByteArray(string))
            }
```

ddoktorski (approved):

crates/snforge-scarb-plugin/src/attributes/should_panic/expected.rs
```diff
@@ -47,7 +70,7 @@ impl ParseFromExpr<Expr> for Expected {
         arg_name: &str,
     ) -> Result<Self, Diagnostic> {
         let error_msg = format!(
-            "<{arg_name}> argument must be string, short string, number or list of short strings or numbers in regular brackets ()"
+            "<{arg_name}> argument must be string, short string, number or tuple of strings, short strings or numbers in regular brackets ()"
```
MKowalski8: nit: First I've read it as if tuples could only contain strings. Imo we can rewrite it a little bit. I give two suggestions, but do what you like. 

```suggestion
            "<{arg_name}> argument must be a string, short string, number, or tuple of those in parentheses ()"
```


```suggestion
            "<{arg_name}> argument must be a string, short string, number, or tuple containing any mix of strings, short strings, and numbers in parentheses ()"
```

crates/snforge-scarb-plugin/src/attributes/should_panic/expected.rs
MKowalski8: nit: I think we could do the same change https://github.com/foundry-rs/starknet-foundry/pull/4373#discussion_r3304418759 here for consistency
```rust
Expr::String(string) => {
    let string = string.string_value(db).ok_or_else(|| {
        T::error(format!("<{arg_name}> contains invalid string literal"))
    })?;
    Ok(Self::ByteArray(string))
}
```

MKowalski8 (approved):

crates/snforge-scarb-plugin/src/attributes/should_panic/expected.rs
franciszekjob: Unfortunately it's not 100% correct - `string_value(db)` returns the interpreted value (e.g. `\"` becomes a literal `"`), which then gets re-emitted without escaping, producing invalid output. The current `text(db).trim_matches('"')` is correct since it preserves raw escape sequences as-is, which is why 
`work_with_expected_string_escaped fails` failed with this change.

crates/snforge-scarb-plugin/src/attributes/should_panic/expected.rs
franciszekjob: Same reason as in https://github.com/foundry-rs/starknet-foundry/pull/4373#discussion_r3309925351

PR #4376, opened 2026-05-21
Add more aliases

DelevoXDG:
Towards #2240

## Stack
- #4365 
- #4366
- #4376 
- #4382
- #4383

## Introduced changes

- Support `@alias` on felt args in `call`, `invoke`, `deploy`, `declare-from`, `account create`/`import`, `get balance`/`nonce`/`class-hash-at`, `verify`, `utils serialize`, `utils contract-address`
- **tests:** ok + fail per command above

## Checklist

- [x] Linked relevant issue
- [x] Updated relevant documentation
- [x] Added relevant tests
- [ ] Performed self-review of the code
- [x] Added changes to `CHANGELOG.md`

PR #4366, opened 2026-05-15
Add `alias list`; Update `show-config` to display aliases

DelevoXDG:
Towards #2240

## Stack
- #4365 
- #4366
- #4376 
- #4382
- #4383

## Introduced changes

- add `alias list` - show aliases from effective config
- update `show-config` - display number of configured aliases

## Checklist

- [x] Linked relevant issue
- [x] Updated relevant documentation
- [x] Added relevant tests
- [ ] Performed self-review of the code
- [x] Added changes to `CHANGELOG.md`

PR #4382, opened 2026-05-25
Add aliases in multicalls

DelevoXDG:
Towards #2240

## Stack
- #4365 
- #4366
- #4376 
- #4382
- #4383

## Introduced changes

- `@alias` in multicall (`execute`, `run`)
- `id` overrides alias when both set

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests

### DIFF
--- a/crates/sncast/src/starknet_commands/multicall/mod.rs
+++ b/crates/sncast/src/starknet_commands/multicall/mod.rs
@@ -1,4 +1,4 @@
-use anyhow::{Context, Result};
+use anyhow::Result;
 use clap::{Args, Subcommand};
 use serde::Serialize;
 use serde_json::{Value, json};
@@ -89,6 +89,7 @@ pub async fn multicall(
                     run.clone(),
                     account,
                     &provider,
+                    &config,
                     wait_config,
                     ui,
                 )
@@ -113,6 +114,7 @@ pub async fn multicall(
                     *execute.clone(),
                     account,
                     &provider,
+                    &config,
                     wait_config,
                     ui,
                 )
@@ -130,11 +132,17 @@ pub async fn multicall(
     }
 }
 
-/// Replaces arguments that reference user-defined ids with their corresponding values from the contract registry.
-// TODO: add full support for multicall id <> alias id resolution (will be done in next PR)
+/// Replaces arguments that reference `@ID`s with their corresponding felt values.
+///
+/// If `@ID`, attempt to resolve felt value from, in the following order:
+/// 1. [`ContractRegistry`]
+/// 2. Aliases from config
+///
+/// Non-`@` values are left unchanged.
 pub fn replaced_arguments(
     arguments: &Arguments,
     contracts: &ContractRegistry,
+    config: &CastConfig,
 ) -> Result<Arguments> {
     Ok(match (&arguments.calldata, &arguments.arguments) {
         (Some(calldata), None) => {
@@ -142,12 +150,9 @@ pub fn replaced_arguments(
                 .iter()
                 .map(|raw_input| {
                     let input = FeltOrId::new(raw_input.clone());
-                    if let Some(id) = input.as_id() {
-                        contracts
-                            .get_address_by_id(id)
-                            .with_context(|| {
-                                format!("Failed to find contract address for id: {id}")
-                            })
+                    if input.as_id().is_some() {
+                        input
+                            .resolve_for_multicall(contracts, config)
                             .map(|f| f.to_string())
                     } else {
                         Ok(raw_input.clone())
--- a/crates/sncast/src/starknet_commands/multicall/run.rs
+++ b/crates/sncast/src/starknet_commands/multicall/run.rs
@@ -3,7 +3,7 @@ use crate::starknet_commands::invoke::execute_calls;
 use crate::starknet_commands::multicall::contract_registry::ContractRegistry;
 use crate::starknet_commands::multicall::deploy::MulticallDeploy;
 use crate::starknet_commands::multicall::invoke::MulticallInvoke;
-use crate::starknet_commands::utils::felt_or_id::FeltOrId;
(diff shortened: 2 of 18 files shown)


crates/sncast/src/starknet_commands/multicall/run.rs
```diff
@@ -151,15 +152,29 @@ where
...
+                if s.starts_with(ID_PREFIX) {
+                    FeltOrId::new(s.clone()).resolve_for_multicall(contract_registry, config)?
+                // Note: Unlike other id cases, we allow bare-name `id` here (rather than `@id`)
+                // TODO: Consider requiring `@` consistently here too and deprecating the bare-name path so calldata, contract addresses, and class hashes share same syntax.
+                } else if let Some(addr) = contract_registry.get_address_by_id(s) {
+                    addr
+                } else {
+                    felt_from_string(s)?
+                }
+            }
```
DelevoXDG, line 170: What are your thoughts on this?

I feel like we should make a breaking change here and proceed with clarity ("@ is required for aliases / ids").

franciszekjob: Could you please give an example of before vs after this breaking change?

crates/sncast/src/starknet_commands/utils/felt_or_id.rs
```diff
@@ -38,6 +39,33 @@ impl FeltOrId {
...
+                return Ok(addr);
+            }
+            config.aliases.get(name).copied().with_context(|| {
+                format!(
+                    "`@{name}`: not found as multicall step id or in [sncast.<profile>.aliases]"
+                )
+            })
+        } else {
+            self.try_into_felt()
+        }
```
franciszekjob: nit: Isn't this easier to read?
```suggestion
        let Some(name) = self.as_id() else {
            return self.try_into_felt();
        };

        if name.is_empty() {
            bail!("Alias name cannot be empty");
        }

        registry
            .get_address_by_id(name)
            .or_else(|| config.aliases.get(name).copied())
            .with_context(|| {
                format!(
                    "`@{name}`: not found as multicall step id or in [sncast.<profile>.aliases]"
                )
            })
```

crates/sncast/src/starknet_commands/multicall/mod.rs
```diff
@@ -130,24 +132,27 @@ pub async fn multicall(
...
-                    if let Some(id) = input.as_id() {
-                        contracts
-                            .get_address_by_id(id)
-                            .with_context(|| {
-                                format!("Failed to find contract address for id: {id}")
-                            })
+                    if input.as_id().is_some() {
+                        input
+                            .resolve_for_multicall(contracts, config)
                             .map(|f| f.to_string())
```
MKowalski8, line 156: nit: Can't be this just 👇, since `as_id().is_some()` is checked in `resolve_for_multicall`? 

```suggestion
                    FeltOrId::new(raw_input.clone())
                        .resolve_for_multicall(contracts, config)
                        .map(|f| f.to_string())
```

MKowalski8 (commented):
Please re-request review again when there is more context in https://github.com/foundry-rs/starknet-foundry/pull/4382#discussion_r3325106712

crates/sncast/src/starknet_commands/multicall/run.rs
DelevoXDG: For example, in https://github.com/foundry-rs/starknet-foundry/blob/85151b6e9874e8bc6a15cb556cc66a45074ae415/crates/sncast/tests/data/multicall_configs/deploy_invoke_calldata_ids.toml#L1-L20, we would no longer support non-`@` references to multicall step ids:

```diff
[[call]]
call_type = "deploy"
class_hash = "0x02a09379665a749e609b4a8459c86fe954566a6beeaddd0950e43f6c700ed321"
inputs = []
id = "map_contract"
unique = false

[[call]]
call_type = "invoke"
contract_address = "0xcd8f9ab31324bb93251837e4efb4223ee195454f6304fcfcb277e277653008"
function = "put"
- inputs = ["0x123", "map_contract"]
+ inputs = ["0x123", "@map_contract"] 

[[call]]
call_type = "deploy"
class_hash = "0x059426c817fb8103edebdbf1712fa084c6744b2829db9c62d1ea4dce14ee6ded"
- inputs = ["map_contract", "0x1", "0x1"]
+ inputs = ["@map_contract", "0x1", "0x1"]
id = "constructor-params"
unique = false
```

crates/sncast/src/starknet_commands/multicall/mod.rs
DelevoXDG: They're not equivalent, as current approach leaves strings as is:
```
                        Ok(raw_input.clone())
```
and `resolve_for_multicall` emits felts.

Maybe the whole chain:

```
        let constructor_arguments = replaced_arguments(
            &Arguments::from(self.common.arguments.clone()),
            contract_registry,
            config,
        )?;
       ...
               let constructor_calldata = if let Some(raw_calldata) = &constructor_arguments.calldata {
            calldata_to_felts(raw_calldata)?
        } else {
            let contract_class = contract_registry
                .get_contract_class_by_class_hash(&class_hash)
                .await?;
            constructor_arguments.try_into_calldata(
                &abi_from_contract_class(contract_class)?,
                &constructor_selector,
            )?
        };
```
deserves a refactor to avoid intermediate strings, but I feel like this is out of scope for this PR, wdyt? 🤔

MKowalski8: Okay, never mind then

crates/sncast/src/starknet_commands/multicall/run.rs
franciszekjob: Agree, breaking changes is reasonable here.

DelevoXDG: done

franciszekjob (approved):

MKowalski8 (approved):

PR #4386, opened 2026-05-26
Warn on unknown keys in config (instead of error)

### LINKED ISSUE #4378: Aggressive global config unknown-key rejection
## Current state

For example, using a feature from a newer version of sncast in config will be breaking the older sncast versions: 
```
Error: Failed to load global config at /Users/delevo/.config/starknet-foundry/snfoundry.toml

Caused by:
    unknown field(s) ["aliases"]
```

## Proposal
- Emit warning instead of hard error on unknown keys in global config
- This would allow backwards-compatibility in configs

## TDB 
- In local config, it make sense to enforce no extra keys
- But, from consistency perspective, it could make sense to just emit warnings instead of errors in both local and global configs

DelevoXDG:
Closes #4378 

## Introduced changes

Now, instead of unknown config keys being hard error, a warning will be emmited:
- List of unknown keys
- Affected config files

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,8 +54,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 #### Changed
 
-- Unknown keys in `snfoundry.toml` (`[sncast.<profile>]`, `[sncast.<profile>.networks]`) now emit a warning and are ignored instead of causing a hard error, so configs can be shared across `sncast` versions. Read more in [configuration](https://foundry-rs.github.io/starknet-foundry/projects/configuration.html).
-- Unknown keys inside `wait-params` in `snfoundry.toml` now emit a warning.
+- Unknown keys in `snfoundry.toml` (`[sncast.<profile>]`, `[sncast.<profile>.networks]`, `wait-params`) now emit a warning and are ignored instead of causing a hard error, so configs can be shared across `sncast` versions. Read more in [configuration](https://foundry-rs.github.io/starknet-foundry/projects/configuration.html).
 
 #### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,7 +54,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 #### Changed
 
-- Unknown keys in `snfoundry.toml` now, instead of causing hard error, emit a warning and are ignored, so configs can be used across `sncast` versions.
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -51,6 +51,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `sncast get spec-version` command that returns the version of the Starknet JSON-RPC specification used by the node
 - `sncast get tx-receipt` (alias: `get transaction-receipt`) command that returns the receipt of a transaction
 
+#### Changed
+
+- Unknown keys in `snfoundry.toml` (`[sncast.<profile>]`, `[sncast.<profile>.networks]`, `wait-params`) now emit a warning and are ignored instead of causing a hard error, so configs can be shared across `sncast` versions. Read more in [configuration](https://foundry-rs.github.io/starknet-foundry/projects/configuration.html).
```
ddoktorski: Please move it under the section `## [Unreleased]`

docs/src/appendix/snfoundry-toml.md
```diff
@@ -18,6 +18,15 @@ If it is not found, default values will be used.
...
 Profiles allow you to define different sets of configurations for various environments or use cases. For more details, see the [profiles explanation](../projects/configuration.md).
 
+> 📝 **Tip**
+> 
+> If this `sncast` version does not recognize a key, this key will be ignored with a warning:
+> ```toml
+> [sncast.myprofile]
+> unknown-key = "some value"
+> ```
+> To avoid these warnings, use latest `sncast` version and update your config to the latest format.
```
ddoktorski: nit:
```suggestion
> To avoid these warnings, use the latest `sncast` version and update your config to the latest format.
```

ddoktorski (approved):

docs/src/appendix/snfoundry-toml.md
```diff
@@ -18,6 +18,15 @@ If it is not found, default values will be used.
 All fields are optional and do not have to be provided. In case a field is not defined in a manifest file, it must be provided in CLI when executing a relevant `sncast` command.
 Profiles allow you to define different sets of configurations for various environments or use cases. For more details, see the [profiles explanation](../projects/configuration.md).
 
+> 📝 **Tip**
+> 
+> If this `sncast` version does not recognize a key, this key will be ignored with a warning:
```
MKowalski8: nit: what does "`this sncast`" refer to in this context? Can we rewrite it? 

Mayby something like this?
```suggestion
> Keys unrecognized by the current sncast version are ignored with a warning:
```

Or even shorter without `current`: 
```suggestion
 If sncast does not recognize a key, it will be ignored with a warning:
```

MKowalski8 (approved):

docs/src/appendix/snfoundry-toml.md
DelevoXDG: rewritten, i kept the "currently used" to keep it clear

PR #4401, opened 2026-06-02
Add `sncast config-path`

### LINKED ISSUE #4372: Display config path in `sncast`
## Current state

There's no way to view the current sources of the configuration.

This is problematic from UX standpoint - the user may have problems with config or may want to change something in config, and not know what they're supposed to edit.

And even if they do, it would be much more convenient for CLI to print path to the files, allowing quickly opening them in the editor. 

## Objective

Either:
1. Update `sncast show-config` to display path to whatever config files contributed to the effective config, e.g.:
   ```
    $ sncast_build show-config
    Local Config:        missing
    Global Config:       /Users/user/.config/starknet-foundry/snfoundry.toml
    Chain ID:            alpha-sepolia 
    Network:             sepolia
    ...
    ```
2. Or add separate command for this, e.g. `sncast config-path`:
    ```
    $ sncast_build config-path
    Local Config:        missing
    Global Config:       /Users/user/.config/starknet-foundry/snfoundry.toml
    ```

DelevoXDG:
Closes #4372 

### Stack
- this PR
- #4403  
 

## Introduced changes

- display local and global config paths

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests

### DIFF
--- a/crates/sncast/src/main.rs
+++ b/crates/sncast/src/main.rs
@@ -25,7 +25,7 @@ use foundry_ui::components::warning::WarningMessage;
 use mimalloc::MiMalloc;
 use shared::auto_completions::{Completions, generate_completions};
 use sncast::helpers::command::process_command_result;
-use sncast::helpers::config::{get_or_create_global_config_path, resolve_config_file_paths};
+use sncast::helpers::config::get_or_create_global_config_path;
 use sncast::helpers::configuration::{
     CastConfig, CliConfigOpts, ConfigScope, MaybeConfig, PartialCastConfig, warn_unknown_keys,
 };
@@ -337,8 +337,7 @@ fn run(cli: Cli, ui: &UI) -> Result<ExitCode> {
     } else if let Commands::Script(script) = &cli.command {
         run_script_command(&cli, runtime, script, ui)
     } else if let Commands::ConfigPath(_) = &cli.command {
-        let paths = resolve_config_file_paths(ui);
-        let result = Ok(starknet_commands::config_path::config_path(paths));
+        let result = starknet_commands::config_path::config_path(ui);
         Ok(process_command_result("config-path", result, ui, None))
     } else {
         let config = get_cast_config(&cli, ui)?;
--- a/crates/sncast/src/main.rs
+++ b/crates/sncast/src/main.rs
@@ -9,9 +9,9 @@ use crate::starknet_commands::invoke::InvokeCommonArgs;
 use crate::starknet_commands::script::run_script_command;
 use crate::starknet_commands::utils::{self, Utils};
 use crate::starknet_commands::{
-    account, account::Account as AccountCommand, alias::Alias, call::Call, declare::Declare,
-    deploy::Deploy, get::tx_status::TxStatus, invoke::Invoke, multicall::Multicall, script::Script,
-    show_config::ShowConfig,
+    account, account::Account as AccountCommand, alias::Alias, call::Call, config_path::ConfigPath,
+    declare::Declare, deploy::Deploy, get::tx_status::TxStatus, invoke::Invoke,
+    multicall::Multicall, script::Script, show_config::ShowConfig,
 };
 use crate::starknet_commands::{get, multicall};
 use anyhow::{Context, Result, bail};
@@ -25,7 +25,7 @@ use foundry_ui::components::warning::WarningMessage;
 use mimalloc::MiMalloc;
 use shared::auto_completions::{Completions, generate_completions};
 use sncast::helpers::command::process_command_result;
-use sncast::helpers::config::get_or_create_global_config_path;
+use sncast::helpers::config::{get_or_create_global_config_path, resolve_config_file_paths};
 use sncast::helpers::configuration::{
     CastConfig, CliConfigOpts, ConfigScope, MaybeConfig, PartialCastConfig, warn_unknown_keys,
 };
@@ -191,6 +191,9 @@ enum Commands {
     /// Show current configuration being used
     ShowConfig(ShowConfig),
 
+    /// Show paths to the config files contributing to the effective configuration
+    ConfigPath(ConfigPath),
+
     /// Run or initialize a deployment script
     Script(Script),
 
@@ -333,6 +336,10 @@ fn run(cli: Cli, ui: &UI) -> Result<ExitCode> {
         generate_completions(completions.shell, &mut Cli::command())
     } else if let Commands::Script(script) = &cli.command {
         run_script_command(&cli, runtime, script, ui)
+    } else if let Commands::ConfigPath(_) = &cli.command {
+        let paths = resolve_config_file_paths(ui);
+        let result = Ok(starknet_commands::config_path::config_path(paths));
+        Ok(process_command_result("config-path", result, ui, None))
     } else {
         let config = get_cast_config(&cli, ui)?;
         runtime.block_on(run_async_command(cli, config, ui))
@@ -814,7 +821,7 @@ async fn run_async_command(cli: Cli, config: CastConfig, ui: &UI) -> Result<Exit
             Ok(process_command_result("ledger", result, ui, None))
         }
 
-        Commands::Completions(_) | Commands::Script(_) => {
+        Commands::Completions(_) | Commands::Script(_) | Commands::ConfigPath(_) => {
             unreachable!("should be handled before this function is called")
         }
     }
--- a/crates/sncast/src/starknet_commands/config_path.rs
+++ b/crates/sncast/src/starknet_commands/config_path.rs
@@ -9,7 +9,6 @@ use sncast::response::ui::UI;
 #[derive(Args)]
 #[command(
     about = "Show paths to the config files contributing to the effective configuration",
-    long_about = None
 )]
 pub struct ConfigPath {}
 
--- a/crates/sncast/src/starknet_commands/config_path.rs
+++ b/crates/sncast/src/starknet_commands/config_path.rs
@@ -7,9 +7,7 @@ use sncast::response::config_path::ConfigPathResponse;
 use sncast::response::ui::UI;
 
 #[derive(Args)]
-#[command(
-    about = "Show paths to the config files contributing to the effective configuration",
-)]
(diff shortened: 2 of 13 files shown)


crates/sncast/src/starknet_commands/config_path.rs
```diff
@@ -0,0 +1,37 @@
...
+)]
+pub struct ConfigPath {}
+
+#[allow(clippy::unnecessary_wraps)]
+pub fn config_path(ui: &UI) -> Result<ConfigPathResponse> {
+    let local = find_config_file()
+        .ok()
+        .map(|path| path.canonicalize_utf8().unwrap_or(path));
+
+    // TODO: consider adding a wrapper-helper for this logic
```
DelevoXDG, line 19: #4403

crates/sncast/tests/e2e/config_path.rs
franciszekjob, line 1: Let's add a test case for json output when a path is missing.

crates/sncast/src/starknet_commands/config_path.rs
```diff
@@ -0,0 +1,37 @@
...
+use configuration::find_config_file;
+use foundry_ui::components::warning::WarningMessage;
+use sncast::helpers::config::get_or_create_global_config_path;
+use sncast::response::config_path::ConfigPathResponse;
+use sncast::response::ui::UI;
+
+#[derive(Args)]
+#[command(
+    about = "Show paths to the config files contributing to the effective configuration",
+    long_about = None
```
franciszekjob: nit: Seems we can remove it?
```suggestion
```

crates/sncast/src/starknet_commands/config_path.rs
```diff
@@ -0,0 +1,37 @@
...
+)]
+pub struct ConfigPath {}
+
+#[allow(clippy::unnecessary_wraps)]
+pub fn config_path(ui: &UI) -> Result<ConfigPathResponse> {
+    let local = find_config_file()
+        .ok()
```
franciszekjob, line 16: This converts all errors (e.g. for permission denied) to `None`. Don't we want to match on the specific error type and perhaps log a warning 🤔 ?

franciszekjob (commented):
In https://github.com/foundry-rs/starknet-foundry/issues/4372 it's mentioned that `show-config` will also display config paths. Don't we want to include it as well?

crates/sncast/src/starknet_commands/config_path.rs
DelevoXDG: I think you raise potentially valid point, but I think it's a bit out of scope for this issue + might be more time-taking for this simple problem than it seems.

1. Our current `find_config_file` was never intended it seems to propagate any permission errors.
In it's current form, it converts all errors to "not found". :

```
pub fn search_config_upwards_relative_to(current_dir: &Utf8PathBuf) -> Result<Utf8PathBuf> {
    current_dir
        .ancestors()
        .find(|path| fs::metadata(path.join(CONFIG_FILENAME)).is_ok())
        .map(|path| path.join(CONFIG_FILENAME))
        .ok_or_else(|| {
            anyhow!(
                "Failed to find snfoundry.toml - not found in current nor any parent directories"
            )
        })
}

pub fn find_config_file() -> Result<Utf8PathBuf> {
    search_config_upwards_relative_to(&Utf8PathBuf::try_from(
        env::current_dir().expect("Failed to get current directory"),
    )?)
}
```


So we wouldn't have to just change `ok()` here, but rather do a refactor across the board

2. The refactor would have to:
- Figure out whether it's feasible
- Update plenty of callers
- Make tricky permission-based tests
- Check if other file access cases are affected by the same problem, we have for example:
   ```rs
    pub fn check_account_file_exists(accounts_file_path: &Utf8PathBuf) -> Result<()> {
        if !accounts_file_path.exists() {
            bail! {"Accounts file = {accounts_file_path} does not exist! If you do not have an account create one with `account create` command \
            or if you're using a custom accounts file, make sure to supply correct path to it with `--accounts-file` argument." }
        }
        Ok(())
    }
   ```
   which is essentially the same case because `exists` treats no perm as `false`:
   ```
    if you cannot access the directory containing the file, e.g., because of a permission error, this will return false.
   ```

DelevoXDG:
> In #4372 it's mentioned that `show-config` will also display config paths. Don't we want to include it as well?

The issue had "Either:" 😅 

But to clarify the reason for the current state, I decided on `show-config` because:
- `config-path` skips config parsing and validation. Thus, it allows us to display config paths even if the configs themselves are invalid
- `show-config` in that scenario would produce an error:
   ```
   Command: show-config
   Error: Failed to load local config at [..]snfoundry.toml

   Caused by:
       Only one of `url` or `network` may be specified
   ```
   The error message is kinda ok, but not what user requested (both paths)

We can of course add these paths to `show-config` output too if you think it would be useful, but I think it may just unnecessarily make output longer.

 
WDYT?

crates/sncast/src/starknet_commands/config_path.rs
franciszekjob: Ok, thx for clarification. If this is not a local-only matter, then let's drop it, at least for now. The benefit/effort ration from such refactor seems low.

franciszekjob:
> > In #4372 it's mentioned that `show-config` will also display config paths. Don't we want to include it as well?
> 
> The issue had "Either:" 😅
> 
> But to clarify the reason for the current state, I decided on `show-config` because:
> 
> * `config-path` skips config parsing and validation. Thus, it allows us to display config paths even if the configs themselves are invalid
> * `show-config` in that scenario would produce an error:
>   ```
>   Command: show-config
>   Error: Failed to load local config at [..]snfoundry.toml
>   
>   Caused by:
>       Only one of `url` or `network` may be specified
>   ```
>   
>   
>       
>         
>       
>   
>         
>       
>   
>       
>     
>   The error message is kinda ok, but not what user requested (both paths)
> 
> We can of course add these paths to `show-config` output too if you think it would be useful, but I think it may just unnecessarily make output longer.
> 
> WDYT?

Just wanted some clarification on this one, let's stay with current form.

docs/src/appendix/sncast/config_path.md
```diff
@@ -0,0 +1,5 @@
+# `config_path`
```
franciszekjob, line 1:
```suggestion
# `config-path`
```

franciszekjob (approved):

docs/src/starknet/show_config.md
```diff
@@ -3,18 +3,22 @@
 ## Overview
 
 Sometimes, before executing any other `sncast` command, one may want to make sure that the right
-configuration settings are being used (eg proper network or account is used).
+configuration settings are being used (e.g., proper network or account is used).
 
-To see just that, a `show-config` subcommand can be used. You can just
-replace any subcommand (and its parameters) with `show-config` and it will show currently used configuration.
+Top see what `sncast` will use before running other commands, you can use:
```
MKowalski8: Typo:
```suggestion
To see what `sncast` will use before running other commands, you can use:
```

crates/sncast/src/main.rs
MKowalski8, line 899: What would you say about moving warning into `PartialCastConfig`? 

My concern is that there might be places where we will load config, but forget to do `warn_unknown_keys`. 

Mayby we can add some func in `PartialCastConfig` that would do this 👇, there is only `global_path` and `local_path` that we need to pass. Then we could bundle the logic of warning together there. 
```rust
    let global_default =
        PartialCastConfig::load_maybe(global_path.as_ref(), None, ConfigScope::Global)?;
    let global_profile = if profile.is_some() {
        PartialCastConfig::load_maybe(global_path.as_ref(), profile, ConfigScope::Global)?
    } else {
        MaybeConfig::NoProfile
    };
    let local_default =
        PartialCastConfig::load_maybe(local_path.as_ref(), None, ConfigScope::Local)?;
    let local_profile = if profile.is_some() {
        PartialCastConfig::load_maybe(local_path.as_ref(), profile, ConfigScope::Local)?
    } else {
        MaybeConfig::NoProfile
    };
```

I'am not saying it's better. Mayby this is only refactor for refactor, but please take a look, mayby it makes sense to do this way.

crates/sncast/tests/e2e/show_config.rs
MKowalski8, line 676: Would it be better to give full paths, e.g. `sncast.default.networks.custom` or at least `networks.custom`? 

Mayby the problem with overlapping names in warning is not something important, but we would solve this too.

MKowalski8 (approved):

crates/sncast/src/main.rs
DelevoXDG: Note that this PR only addresses addition of `scarb show-config`, it this branch was just based on #4386 that was since then merged.

However, to address your suggestion itself:

> My concern is that there might be places where we will load config, but forget to do `warn_unknown_keys`.

I'm fairly confident in us promptly catching such a blunder on our part 😄 

> Mayby we can add some func in PartialCastConfig that would do this 👇

I've considered that, but that would mean generating warnings for each `PartialCastConfig::load_maybe` call, which would lead to duplicate warnings for each config & profile combination that has invalid syntax.

Current approach allows us to dedup these warnings, and I think good UX is priority here.

crates/sncast/tests/e2e/show_config.rs
DelevoXDG: As I mentioned before, I wanted to deduplicate the warnigns.

There are trade-off of course - your approach produces more clear diagnostics.

But for tool with a relatively small config size - I think scannability and lack of duplicates could be more important.

WDYT?

MKowalski8: I am okay with that, just checking it's something we are aware of

crates/sncast/src/main.rs
MKowalski8: > I've considered that, but that would mean generating warnings for each PartialCastConfig::load_maybe call, 

I thought more of a new function that would do all of the `load_maybe` and`warn`. Not moving warn into `load_maybe`

> I'm fairly confident in us promptly catching such a blunder on our part 😄

In that case I agree, there aren't many changes with `configs` so this shouldn't be a problem after all.

MKowalski8 (approved):

Issue #2464, opened 2024-09-11
Research invariant & differential testing


Arcticae:
### Which components does the task require to be changed? (think hard pls)

snforge

### Description

Foundry supports those 2 modes:
https://book.getfoundry.sh/forge/advanced-testing

michael-moffett:
Hi, author of [`caliperforge/cf-invariants`](https://github.com/caliperforge/cf-invariants) here. This issue maps to the ROADMAP entry *Research Variant and Differential Testing, Better Fuzzing Algorithms* (size 5). Rather than open a parallel issue I wanted to put a concrete reference in front of you.

cf-invariants is a Rust CLI sidecar for stateful invariant testing on Cairo 2.x. It generates harnesses that fuzz call sequences against a deployed contract and runs `snforge test` as a subprocess. Output is a `scorecard.json` plus a readable summary reporting `N/M invariants violated` and the failing call sequence. snforge stays the engine. cf-invariants is a sequence-search and reporting layer around it. Apache-2.0, no exclusivity, no claim on the upstream namespace.

Three reference Cairo 2.x contracts are deployed to Starknet Sepolia and source-verified on Voyager, each with a planted bug on a distinct surface (token conservation, access-control monotonicity, AMM solvency). Manifest and repro recipe live at [`STARKNET_SEPOLIA_DEPLOYMENTS.md`](https://github.com/caliperforge/cf-invariants/blob/main/STARKNET_SEPOLIA_DEPLOYMENTS.md). All three findings reproduce from a clean clone against snforge 0.61.0, scarb 2.18.0, Cairo 2.18.0.

One disclosure since it's relevant. There's an opt-in AI suggestion module that calls Claude Sonnet 4.6 against a versioned prompt. AI-suggested invariants are tagged at the type level via `InvariantSource::AiSuggested` carrying model and prompt version, and the developer reviews each one before it runs. Two of the three findings above came that way; one is hand-written.

Open question on my end is integration shape. cf-invariants currently shells out to `snforge test` and parses output. If you'd prefer a structured runner API, a library entry point, or a particular scorecard schema, I'd rather align with your direction than guess. Same for whether a documented sidecar pattern or a focused upstream PR is the right landing spot.

— Michael Moffett
michael@caliperforge.com

michael-moffett:
[...]

PR #4409, opened 2026-06-04
docs: add stateful invariant testing pattern recipe

michael-moffett:
Adds a small user-land recipe for stateful invariant testing under
docs/src/snforge-advanced-features/, with a worked example listing under
docs/listings/invariant_testing_pattern/. The pattern uses a single u256
fuzzer seed sliced into per-step action bytes, walks a call sequence
against a minimal Vault contract, and asserts a ledger invariant between
every transition.

I left a comment on #2464 a few days ago about a sidecar tool we built for
this same problem on Cairo 2.x. A worked recipe in the book felt more
useful than another comment, since the pattern works against existing
snforge primitives today (#[fuzzer] + dispatcher calls, optionally
start_cheat_caller_address for caller rotation). No new attributes, no
runner changes. The doc page is explicit that this is a user-land pattern
and that native support is tracked in #2464.

Changes:
- docs/listings/invariant_testing_pattern/ contains the Scarb package
  with the Vault contract and the invariant test.
- docs/src/snforge-advanced-features/invariant-testing-pattern.md is the
  recipe walkthrough.
- docs/src/SUMMARY.md gets one new entry under "snforge Advanced Features".

Happy to adjust the framing, the file layout, or the example contract to
fit your conventions. If you'd rather the recipe live somewhere else, or
not in the book until #2464 lands, I can also pull this.

Michael Moffett
michael@caliperforge.com
caliperforge.com/ai-disclosure

### DIFF
--- a/docs/listings/invariant_testing_pattern/Scarb.toml
+++ b/docs/listings/invariant_testing_pattern/Scarb.toml
@@ -1,0 +1,17 @@
+[package]
+name = "invariant_testing_pattern"
+version = "0.1.0"
+edition = "2024_07"
+
+[dependencies]
+starknet = "2.8.5"
+assert_macros = "2.8.5"
+
+[dev-dependencies]
+snforge_std = { path = "../../../snforge_std" }
+
+[[target.starknet-contract]]
+sierra = true
+
+[scripts]
+test = "snforge test"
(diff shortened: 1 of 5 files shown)


ddoktorski:
Hi @michael-moffett , thank you for taking the time to open this pull request. While we appreciate the effort, we try to keep the official documentation focused on core, general-purpose setups. Because of that, we've decided not to merge this. Thanks again for contributing!

PR #4413, opened 2026-06-09
Handle prerelease RPC versions in compatibility checks

franciszekjob:
Closes #

## Introduced changes

This PR makes RPC version compatibility checks ignore prerelease/build and compare only the core `major.minor.patch` version.

## Checklist

- [ ] Linked relevant issue
- [x] Added relevant tests

### DIFF
--- a/crates/shared/src/rpc.rs
+++ b/crates/shared/src/rpc.rs
@@ -52,7 +52,7 @@ mod tests {
     }
 
     #[test]
-    fn matches_expected_build_metadata_version() {
+    fn matches_expected_build_version() {
         assert!(is_expected_version(
             &Version::parse("0.10.3+build.1").unwrap()
         ));
--- a/crates/shared/src/rpc.rs
+++ b/crates/shared/src/rpc.rs
@@ -14,7 +14,7 @@ pub fn create_rpc_client(url: &Url) -> Result<JsonRpcClient<HttpTransport>> {
 
 #[must_use]
 pub fn is_expected_version(version: &Version) -> bool {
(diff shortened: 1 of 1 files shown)


crates/shared/src/rpc.rs
```diff
@@ -14,9 +14,13 @@ pub fn create_rpc_client(url: &Url) -> Result<JsonRpcClient<HttpTransport>> {
...
 #[must_use]
 pub fn is_expected_version(version: &Version) -> bool {
+    // TODO(#4412): Remove the fallback check for "0.10.3-rc.0".
     VersionReq::from_str(EXPECTED_RPC_VERSION)
         .expect("Failed to parse the expected RPC version")
         .matches(version)
+        || VersionReq::from_str("0.10.3-rc.0")
+            .expect("Failed to parse the RPC version")
+            .matches(version)
 }
```
ddoktorski, line 21: What about modifying this function to only assert version with stripped prerelease/build metadata? This way we have a long term solution instead of a quick fix only for the current version.

```rust
#[must_use]
pub fn is_expected_version(version: &Version) -> bool {
    let core_version = Version::new(version.major, version.minor, version.patch);
    VersionReq::from_str(EXPECTED_RPC_VERSION)
        .expect("Failed to parse the expected RPC version")
        .matches(&core_version)
}
```

ddoktorski: Let's also add a few unit tests for it

MKowalski8 (approved):

ddoktorski (approved):

PR #4421, opened 2026-06-12
Add `get block` command

ddoktorski:
Towards #4144

### DIFF
--- a/crates/sncast/src/response/block.rs
+++ b/crates/sncast/src/response/block.rs
@@ -51,7 +51,7 @@ fn append_block(builder: OutputBuilder, block: &BlockWithTxHashes) -> OutputBuil
         starknet_version,
         transactions,
         transaction_count,
-        // Omit Merkle-Patricia tries and redundant information form the formatted output.
+        // Omit Merkle-Patricia tries and redundant information from the formatted output.
         // They are still returned when using the `--json` flag.
         event_commitment: _,
         transaction_commitment: _,
--- a/crates/sncast/src/response/block.rs
+++ b/crates/sncast/src/response/block.rs
@@ -36,45 +36,72 @@ impl SncastCommandMessage for BlockResponse {
 }
 
 fn append_block(builder: OutputBuilder, block: &BlockWithTxHashes) -> OutputBuilder {
+    let BlockWithTxHashes {
+        status,
+        block_hash,
+        parent_hash,
+        block_number,
+        new_root,
+        timestamp,
+        sequencer_address,
+        l1_gas_price,
+        l2_gas_price,
+        l1_data_gas_price,
+        l1_da_mode,
+        starknet_version,
+        event_commitment: _,
+        transaction_commitment: _,
+        receipt_commitment: _,
+        state_diff_commitment: _,
+        event_count: _,
+        transaction_count,
+        state_diff_length: _,
+        transactions,
+    } = block;
     builder
-        .field("Status", fmt_status(block.status))
-        .padded_felt_field("Block Hash", &block.block_hash)
-        .field("Block Number", &block.block_number.to_string())
-        .padded_felt_field("Parent Hash", &block.parent_hash)
-        .padded_felt_field("New Root", &block.new_root)
-        .field("Timestamp", &block.timestamp.to_string())
-        .padded_felt_field("Sequencer Address", &block.sequencer_address)
-        .field("L1 Gas Price", &fmt_resource_price(&block.l1_gas_price))
-        .field("L2 Gas Price", &fmt_resource_price(&block.l2_gas_price))
-        .field(
-            "L1 Data Gas Price",
-            &fmt_resource_price(&block.l1_data_gas_price),
-        )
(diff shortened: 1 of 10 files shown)


ddoktorski:

- #4421 ◀
    - #4422
        - #4423
            - #4429

crates/sncast/src/response/block.rs
```diff
@@ -0,0 +1,100 @@
...
+        .field("L2 Gas Price", &fmt_resource_price(&block.l2_gas_price))
+        .field(
+            "L1 Data Gas Price",
+            &fmt_resource_price(&block.l1_data_gas_price),
+        )
+        .field("L1 DA Mode", fmt_da_mode(block.l1_da_mode))
+        .field("Starknet Version", &block.starknet_version)
+        .field("Transaction Count", &block.transactions.len().to_string())
+        .felt_list_field("Transactions", &block.transactions)
+}
```
DelevoXDG, line 107: I suggest taking two things from `get transaction` handling:

1. [Destructure](https://github.com/foundry-rs/starknet-foundry/blob/243f6578cc24bd9297f25050f4b5f33b0b78d617/crates/sncast/src/response/transaction.rs#L222]) `BlockWithTxHashes` / `PreConfirmedBlockWithTxHashes`, so we ensure all fields from schema are used, and would have to update the response if new fields are added:
    
   I notice there's currently mismatch between human-readable and json output: `event_commitment`, `receipt_commitment`, `state_diff_commitment`, `transaction_count` are present in json variant but absent in human-readable. If that's intentional, it's better to explicitly ignore fields after destructuring, and document why are we omitting these.
2. Nit: Use [common helpers](https://github.com/foundry-rs/starknet-foundry/blob/master/crates/sncast/src/response/transaction.rs#L84) - should ensure the same kind of format is used for same type of fields, and reduce some duplication

crates/sncast/src/starknet_commands/get/block.rs
```diff
@@ -0,0 +1,48 @@
...
+use std::process::ExitCode;
+
+#[derive(Debug, Args)]
+#[command(about = "Get a block with transaction hashes")]
+pub struct Block {
+    /// Block identifier on which the block should be fetched.
+    /// Possible values: `pre_confirmed`, `latest`, block hash (0x prefixed string)
+    /// and block number (u64)
+    #[arg(default_value = "latest")]
+    pub block_id: String,
```
DelevoXDG: Why are we defaulting to `latest`? 

It's pre-confirmed for other `get` commands 🤔 

https://github.com/foundry-rs/starknet-foundry/blob/5d710c72ecd9ee24bb6111185ce9fa825d78b019/crates/sncast/src/starknet_commands/get/nonce.rs#L25-L25

ddoktorski: Changed to `latest` for consistency

crates/sncast/src/response/block.rs
ddoktorski: > I notice there's currently mismatch between human-readable and json output: event_commitment, receipt_commitment, state_diff_commitment, transaction_count are present in json variant but absent in human-readable. 

I think it's not useful to have these information in the output

> If that's intentional, it's better to explicitly ignore fields after destructuring, and document why are we omitting these.

Done

> Nit: Use [common helpers](https://github.com/foundry-rs/starknet-foundry/blob/master/crates/sncast/src/response/transaction.rs#L84) - should ensure the same kind of format is used for same type of fields, and reduce some duplication

Do you have any specific places in mind? I made some changes related to this in next PRs

DelevoXDG: > Do you have any specific places in mind? I made some changes related to this in next PRs

What I had in mind is adding a helper for common fields, e.g.
```diff
- .padded_felt_field("Sequencer Address", sequencer_address)
+ .sequencer_address(sequencer_address)
```
Similarly to `TransactionOutputBuilder`

crates/sncast/src/starknet_commands/get/block.rs
DelevoXDG: It was `latest` though 😅

docs/src/appendix/sncast/get/block.md
```diff
@@ -0,0 +1,26 @@
+# `get block`
+
+Get a block with transaction hashes.
+
+## `[BLOCK_ID]`
```
franciszekjob: nit: Same as we do for other positional args:
```suggestion
## `<BLOCK_ID>`
```

franciszekjob (approved):

crates/sncast/src/starknet_commands/get/block.rs
ddoktorski: typo 🤦, in code it's correctly set to `pre_confirmed`

crates/sncast/src/response/block.rs
ddoktorski: Added in #4423: https://github.com/foundry-rs/starknet-foundry/pull/4423/changes/00b32a5cede7f56f3649bd85c231100d01188031

DelevoXDG (approved):

PR #4422, opened 2026-06-12
Add `--full` flag to `get block` command

ddoktorski:
Towards #4144


ddoktorski:

- #4421
    - #4422 ◀
        - #4423
            - #4429

crates/sncast/src/response/transaction.rs
```diff
@@ -218,7 +249,10 @@ impl TransactionOutputBuilder for OutputBuilder {
     }
 }
 
-fn build_invoke_v0_response(tx: &starknet_rust::core::types::InvokeTransactionV0) -> String {
+fn build_invoke_v0_response(
+    builder: OutputBuilder,
+    tx: &starknet_rust::core::types::InvokeTransactionV0,
+) -> OutputBuilder {
```
DelevoXDG: Nit: since these functions no longer do the building, I think it would be appropriate to rename them like so:

`build_invoke_v0_response` -> `append_invoke_v0_response` / `append_invoke_v0_tx` / `append_invoke_v0`

crates/sncast/tests/e2e/get/block.rs
DelevoXDG, line 1: Should we add a test for zero transactions?

crates/sncast/tests/e2e/get/block.rs
```diff
@@ -63,6 +63,36 @@ async fn test_happy_case_json() {
...
+    // instead of being listed as a flat list of hashes.
+    assert_stdout_contains(
+        output,
+        indoc! {r"
+            Transaction Count:[..]
+
+            Transaction #1
+            Type:[..]
+            [..]Transaction Hash:[..]0x[..]
+        "},
```
DelevoXDG, line 83: I think it would be better to assert full output, including all block details and tx details

crates/sncast/tests/e2e/get/block.rs
```diff
@@ -63,6 +63,36 @@ async fn test_happy_case_json() {
...
+    // With `--full`, transactions are rendered with their full details
+    // instead of being listed as a flat list of hashes.
+    assert_stdout_contains(
+        output,
+        indoc! {r"
+            Transaction Count:[..]
+
+            Transaction #1
+            Type:[..]
+            [..]Transaction Hash:[..]0x[..]
```
DelevoXDG: Nit: If not complicated, I'd consider adding something like `OutputBuilder::indent()` so the tx outputs are more scannable
```
Transaction #1
  Type: ...
  Version: ...

Transaction #2
  Type: ...
  Version: ...
```

crates/sncast/tests/e2e/get/block.rs
ddoktorski: I'm not sure this test is necessary given our existing coverage

crates/sncast/tests/e2e/get/block.rs
ddoktorski: https://github.com/foundry-rs/starknet-foundry/pull/4422/changes/8a47d1e77e1d73fd344553981566d377b6454ac4

crates/sncast/tests/e2e/get/block.rs
ddoktorski: https://github.com/foundry-rs/starknet-foundry/pull/4422/changes/2cc1f91870c8765d8aaf89954b160fc759af6d29

crates/sncast/tests/e2e/get/block.rs
DelevoXDG: The output is a bit different for example in tx mode (which shows `Transactions: []`), and full / receipt mode (which show nothing after `Transaction Count: 0`)

Just to confirm we're aware of it and that's the desired state, but I suppose having coverage is not essential here

crates/sncast/src/response/block.rs
```diff
@@ -1,109 +1,157 @@
...
+            transactions,
+            // Omit Merkle-Patricia tries and redundant information from the formatted output.
+            // They are still returned when using the `--json` flag.
+            event_commitment: _,
+            transaction_commitment: _,
+            receipt_commitment: _,
+            state_diff_commitment: _,
+            event_count: _,
+            transaction_count: _,
+            state_diff_length: _,
```
DelevoXDG: 1. I generally agree with your reasoning on not including these, but the user might expect that `--full` flag returns the "full" data 🤔 
2. Perhaps some of these should always be included - like `event_count` or `transaction_count`?

ddoktorski: Added `event_count`, `transaction_count` and `state_diff_length`, so only Merkle-Patricia tries are omitted

crates/sncast/tests/e2e/get/block.rs
```diff
@@ -60,6 +62,97 @@ async fn test_happy_case_json() {
...
+            Transaction Count:             4
+            Event Count:                   4
+            State Diff Length:             6
+
+            Transaction #1
+              Type:                        INVOKE
+              Version:                     3
+              Transaction Hash:            0x044eee159a2983e017adfe9dc84908b9f6ba0e938c9e30c0ed613fe72cf5dea3
+              Sender Address:              0x05677b3b308df6821be9a3c10423db8d9f7ff4d5bef09e32d973c8c89bf5683b
+              Nonce:                       4425792
```
franciszekjob, line 129: For `get nonce` command, we display nonce as hex. Maybe we should unify that to always use hex or decimal (depending on the chosen format)?

franciszekjob (commented):
Let's update the changelog.

crates/sncast/tests/e2e/get/block.rs
DelevoXDG: For nonce, I suppose decimal is better 🤔

DelevoXDG (approved):

ddoktorski:
> Let's update the changelog.

General entry added in https://github.com/foundry-rs/starknet-foundry/pull/4421. I guess no need to provide more details there

crates/sncast/tests/e2e/get/block.rs
ddoktorski: Agree that decimal is better, it's also displayed as decimals in block explorers. We can change it in `get nonce`

ddoktorski: https://github.com/foundry-rs/starknet-foundry/pull/4432

franciszekjob (approved):

PR #4423, opened 2026-06-12
Add `--receipts` flag to `get block` command

ddoktorski:
Towards #4144

### DIFF
--- a/crates/sncast/src/response/block.rs
+++ b/crates/sncast/src/response/block.rs
@@ -9,6 +9,7 @@ use starknet_rust::core::types::{
     MaybePreConfirmedBlockWithTxs, PreConfirmedBlockWithReceipts, PreConfirmedBlockWithTxHashes,
     PreConfirmedBlockWithTxs, ResourcePrice, Transaction, TransactionWithReceipt,
 };
+use starknet_types_core::felt::Felt;
 
 #[derive(Clone)]
 pub enum BlockResponse {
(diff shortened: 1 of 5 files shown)


ddoktorski:

- #4421
    - #4422
        - #4423 ◀
            - #4429

crates/sncast/src/response/block.rs
```diff
@@ -95,6 +99,18 @@ impl SncastCommandMessage for BlockResponse {
...
+                    append_confirmed_header!(builder, block),
+                    &block.transactions,
+                )
+            }
+            BlockResponse::WithReceipts(MaybePreConfirmedBlockWithReceipts::PreConfirmedBlock(
+                block,
+            )) => append_receipts(
+                append_pre_confirmed_header!(builder, block),
+                &block.transactions,
+            ),
```
DelevoXDG, line 241: I think there's a missing "Transaction Receipts: " header
 
```suggestion
            BlockResponse::WithReceipts(MaybePreConfirmedBlockWithReceipts::Block(block)) => {
                append_receipts(
                    append_confirmed_header!(builder, block)
                        .felt_list_field("Transaction Receipts", &block.transactions),
                    &block.transactions,
                )
            }
            BlockResponse::WithReceipts(MaybePreConfirmedBlockWithReceipts::PreConfirmedBlock(
                block,
            )) => append_receipts(
                append_pre_confirmed_header!(builder, block)
                    .felt_list_field("Transaction Receipts", &block.transactions),
                &block.transactions,
            ),
```

crates/sncast/src/response/block.rs
```diff
@@ -116,6 +132,21 @@ fn append_full_transactions(
...
+    mut builder: OutputBuilder,
+    transactions: &[TransactionWithReceipt],
+) -> OutputBuilder {
+    for (index, transaction) in transactions.iter().enumerate() {
+        builder = append_receipt(
+            builder
+                .blank_line()
+                .text_field(&format!("Transaction #{}", index + 1)),
+            &transaction.receipt,
+        );
```
DelevoXDG: JSON-format entry displays both tx + receipt, in human-readable we display only receipt - is that intentional?

Since tx is part of this schema, the user might expect both the tx and receipt to be present.

I understand that this makes the output too long though 🤔 

In that case, perhaps we should either:
1. Change the docs, as current wording implies both are included:
   ```diff
   - Retrieve full transactions along with their receipts. Cannot be used together with `--full`.
   + Retrieve transactions receipts instead of only their hashes. Use `--full` to retrieve transaction details instead.
   ``` 

2. Alternatively, perhaps we should allow passing `--full` and `--receipts` together. In that case, if both are passed, we include both the transaction and the receipt in the response, so `--full` becomes a tx inclusion modifier, and `--receipts` becomes receipt inclusion modifier.

WDYT?

crates/sncast/src/response/block.rs
ddoktorski: `Transactions` are added there to have a list of felts, for receipts we don't want to do the same

crates/sncast/src/response/block.rs
ddoktorski: Docs updated, no need to do 2. in my opinion

DelevoXDG (approved):

franciszekjob (approved):
LGTM

PR #4431, opened 2026-06-17
Allow devnet accounts without default accounts file

### LINKED ISSUE #4311: Using devnet inbuilt accounts require `accounts-file` to exist
## Current state

Relates to the inbuilt `devnet-i` accounts introduced in #3761 

`sncast` treats `--account devnet-<i>` as dependent on accounts file existence.  

If default accounts file is missing, it fails early:
```
    Error: Accounts file = /home/runner/.starknet_accounts/starknet_open_zeppelin_accounts.json does not exist! If you do not have an account create one with `account create` command or if you're using a custom accounts file, make sure to supply correct path to it with `--accounts-file` argument.
```

However, you don't have to need an accounts-file to be able to use devnet account.
Current logic asserts that you only can use devnet account if `devnet-i` is not present in accounts-file:
https://github.com/foundry-rs/starknet-foundry/blob/899f17ae20cda3d46ee135cda3245ea889942e12/crates/sncast/src/lib.rs#L299
That is also implied by the tests:
https://github.com/foundry-rs/starknet-foundry/blob/899f17ae20cda3d46ee135cda3245ea889942e12/crates/sncast/tests/e2e/devnet_accounts.rs#L12

## Objetive
`devnet-<i>` should not require accounts file to exist.  
If the file is missing, continue with inbuilt Devnet account resolution.

## Additional context
Recent test hotfixes in #4298 pass `--accounts-file` explicitly to avoid relying on ambient default file state.
That should be reverted once the fix is implemented

### LINKED ISSUE #4402: Account file existence is required when using predeployed devnet accounts
### Select Tool

sncast

### Foundry Version

0.61.0

Linux

x86

### Issue Description

Running `sncast --account devnet-1 deploy --network devnet --contract-name MyContract` fails if the default accounts file doesn't exist.
Since the devnet account names are special identifiers though, such a file existence check shouldn't even happen.

### Command Line Output

Error: Accounts file = ~/.starknet_accounts/starknet_open_zeppelin_accounts.json does not exist! If you do not have an account create one with `account create` command or if you're using a custom accounts file, make sure to supply correct path to it with `--accounts-file` argument.

ddoktorski:
Fixes #4402 
Closes #4311 

Allow using devnet accounts if default accounts file doesn't exist. If accounts file is set explicitly by `--accounts-file` flag or in `snfoundry.toml` it will still throw an error.


crates/sncast/src/helpers/account.rs
```diff
@@ -57,22 +57,31 @@ pub fn load_accounts(accounts_file: &Utf8PathBuf) -> Result<Value> {
     Ok(accounts)
 }
 
+/// When `required` is `true`, a missing accounts file or network entry is an error.
+/// It is set to `false` only for devnet accounts, which don't require the default accounts file to exist.
 pub fn check_account_exists(
     account_name: &str,
     network_name: &str,
     accounts_file: &Utf8PathBuf,
+    required: bool,
```
franciszekjob: nit:
```suggestion
    required_accounts_file: bool,
```

crates/sncast/src/lib.rs
```diff
@@ -328,7 +330,18 @@ pub async fn get_account<'a>(
     }
 
     let accounts_file = &config.accounts_file;
-    let exists_in_accounts_file = check_account_exists(account, &network_name, accounts_file)?;
+    // Devnet accounts don't require an accounts file.
+    // When the default accounts file is used, we don't enforce its existence.
+    // When accounts file is set explicitly, it is still required to exist then.
```
franciszekjob, line 335: Is there a specific reason for this? How about making `--accounts-file` mutually exclusive with a devnet account? Personally, such exclusion seems clearer from user-perspective.

ddoktorski: > How about making --accounts-file mutually exclusive with a devnet account?

Devnet accounts are determined only by names. We'd need to disallow using `devnet-i` names in account files, which feels like an artificial limitation

franciszekjob: > Devnet accounts are determined only by names. We'd need to disallow using devnet-i names in account files, which feels like an artificial limitation

Right, disregard. I was under assumption that custom accounts cannot be named `devnet-<i>` (in fact they can 😅 ).

franciszekjob (commented):
I think I noticed a small inconsistency that we have (not introduced by this PR). Account resolution behaves differently if a user has a custom account named `devnet-<i>`, depending on the accounts file they use. With the default accounts file, their custom devnet account will be used. However, with a custom accounts file, the inbuilt devnet account will be used instead. I know it's an edge case, but we should handle it consistently. Could you please take a look at this and address/create an issue?

Example:
image

(see in the first command we get a warning)

ddoktorski:
@franciszekjob 
I'm unable to reproduce this behavior on my end. Could you double-check the steps or share the exact content of your account files?

franciszekjob:
> @franciszekjob I'm unable to reproduce this behavior on my end. Could you double-check the steps or share the exact content of your account files?

Nevermind, my custom.json was indeed flawed, sry.

franciszekjob (approved):
LGTM

crates/sncast/src/helpers/account.rs
DelevoXDG: My take 😅 
```suggestion
    accounts_file_should_exist: bool,
```

crates/sncast/tests/e2e/devnet_accounts.rs
```diff
@@ -50,22 +44,44 @@ pub async fn happy_case(account_number: u8) {
     );
 }
 
+#[tokio::test]
+pub async fn explicit_nonexistent_accounts_file_errors() {
+    // When `--accounts-file` is passed explicitly, its existence is still enforced even for devnet accounts.
+    let temp_dir = tempdir().expect("Unable to create a temporary directory");
```
DelevoXDG, line 54: The logic that requires explicitly specified via `--accounts-file` file to be present - doesn't work when we explicitly pass the default accounts file (that does not exist):

```rs
#[tokio::test]
pub async fn explicit_default_accounts_file_existence_not_enforced() {
    let home_dir = tempdir().expect("Unable to create a temporary HOME directory");
    let temp_dir = tempdir().expect("Unable to create a temporary directory");

    let args = vec![
        "--accounts-file",
        "~/.starknet_accounts/starknet_open_zeppelin_accounts.json",
        "--account",
        "devnet-3",
        "invoke",
        "--url",
        URL,
        "--contract-address",
        MAP_CONTRACT_ADDRESS_SEPOLIA,
        "--function",
        "put",
        "--calldata",
        "0x1 0x2",
    ];
`,
    let snapbox = runner(&args)
        .env("HOME", home_dir.path())
        .current_dir(temp_dir.path());
    let output = snapbox.assert().failure();

    assert_stderr_contains(
        output,
        "Error: Accounts file = [..]starknet_open_zeppelin_accounts.json does not exist! [..]",
    );
}
```
````
Expected failure, was success
stdout:
```
Success: Invoke completed

Transaction Hash: 0x01bb53f951666f05ad40604596351c45ecfd034596ccfac86587b9261dfe54dd

```
````

After some considerations though, I since this is ok from practical standpoint, but maybe worth noting this "intentional inconsistency" e.g. as test comment.

DelevoXDG (approved):

crates/sncast/tests/e2e/devnet_accounts.rs
ddoktorski: Added a comment: https://github.com/foundry-rs/starknet-foundry/pull/4431/changes/0931e80db164eb6a69531d3f91fccdd454b64701

PR #4428, opened 2026-06-15
Merge contract backtrace and test backtraces handling

DelevoXDG:
Closes #3610 

## Stack
- #4400
- #4428
- #4442
- #4444 

## Introduced changes

Now, both contract and test-level backtrace is displayed at once 

## Checklist

- [x] Linked relevant issue
- [x] Updated relevant documentation
- [x] Added relevant tests
- [x] Performed self-review of the code
- [x] Added changes to `CHANGELOG.md`

PR #4439, opened 2026-06-18
Refactor backtrace logic 

DelevoXDG:
## Stack
- #4400
- #4428 
- #4439

## Introduced changes

- refactor backtrace logic to avoid "too many lines" errors


DelevoXDG:
On the second thought, perhaps `too_many_lines` is not worthy of review time 🤔

PR #4441, opened 2026-06-19
Remove `unstable` from `debug-info` features in docs and codebase

### LINKED ISSUE #4454: Remove unstable from debug-info features in docs and codebase
As in the title (check linked closed PR)

MKowalski8:
Closes #4454 

## Introduced changes

- The `debug-info` features were stabilized in scarb `2.15`. So it's now older than the latest version we support, so because of that I deleted every occurence instead of just fixing a few places in docs. If we don't want to get rid of this at the moment, I can limit the scope

## Checklist

- [ ] Linked relevant issue
- [ ] Added relevant tests

### DIFF
--- a/crates/forge/src/profile_validation/backtrace.rs
+++ b/crates/forge/src/profile_validation/backtrace.rs
@@ -43,8 +43,8 @@ fn check_profile(
             "{workspace_manifest_path} must have the Cairo compiler configuration equivalent to the following one to run backtrace:
 
             [profile.{profile}.cairo]
-            unstable-add-statements-functions-debug-info = true
-            unstable-add-statements-code-locations-debug-info = true
+            add-statements-functions-debug-info = true
+            add-statements-code-locations-debug-info = true
             panic-backtrace = true
             ... other entries ...
             ",
--- a/crates/forge/src/profile_validation/coverage.rs
+++ b/crates/forge/src/profile_validation/coverage.rs
@@ -29,8 +29,8 @@ pub fn check_coverage_compatibility(
             "{workspace_manifest_path} must have the Cairo compiler configuration equivalent to the following one to run coverage:
 
             [profile.{profile}.cairo]
-            unstable-add-statements-functions-debug-info = true
-            unstable-add-statements-code-locations-debug-info = true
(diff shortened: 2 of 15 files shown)


MKowalski8:
Just to make sure 
obraz

crates/forge/src/profile_validation/backtrace.rs
DelevoXDG, line 1: Our minimal supported Scarb version is still 2.13.1.

https://github.com/foundry-rs/starknet-foundry/blob/e93d7f70dc6b66e3bbb708de40c279cd2828830e/crates/forge/src/lib.rs#L41

I think we should wait until we bump it to 2.15 before we make these changes, otherwise *technically supported* 2.13-2.14 versions will fail without unstable prefix.

On the other hand, the aliases should still work, so there would be no consequence to keeping `unstable-` prefixes in the docs for now:

https://github.com/software-mansion/scarb/releases/tag/v2.15.0-rc.0

> The unstable_* prefix is removed from keys in cairo section of Scarb manifest file, the keys with prefixes are kept as aliases.

DelevoXDG (changes requested):

crates/forge/src/profile_validation/backtrace.rs
ddoktorski: I guess we can bump it with the next Scarb update (to `2.19.0`) so maybe wait till then?

MKowalski8: > I think we should wait until we bump it to 2.15 before we make these changes, otherwise technically supported 2.13-2.14 versions will fail without unstable prefix.

Since it's one feature and we explicite have `prerequisites` I would say it's enough and bumping `MINIMAL_SCARB_VERSION` is not needed. 

But `2.19.0` should be soon available we can wait until then and bump it

ddoktorski: > On the other hand, the aliases should still work, so there would be no consequence to keeping unstable- prefixes in the docs for now

Ok maybe let's just keep the `unstable-` in the docs for now and update it at some point in the future

Issue #3610, opened 2025-08-04
Add support for backtrace on test level


ksew1:
### Select Tool

snforge

### Current State
If a panic occurs at the test level, there won't be an option to display a backtrace, as it is currently only supported at the contract level.

### Objective
Add backtrace support for test-level functions.

### Additional Context
Since there is separate logic for running test functions and for running contracts, implementation might be challenging.

PR #4442, opened 2026-06-23
Capture test-level backtrace on demand only

DelevoXDG:
Towards #3610 

## Stack
- #4400
- #4428
- #4442
- #4444 

## Introduced changes

Addresses https://github.com/foundry-rs/starknet-foundry/pull/4400#discussion_r3451073907

Refactor test-level backtrace capture. Now, instead of always collecting casm debug offsets and PCs, we only do so if:
- test panics
- backtrace is enabled

The hint (``note: run with `{BACKTRACE_ENV}=1` environment variable to display a backtrace``) is included even if backtrace capture was skipped

## Checklist

- [x] Linked relevant issue
- [ ] Added relevant tests

### DIFF
--- a/crates/forge-runner/src/backtrace/mod.rs
+++ b/crates/forge-runner/src/backtrace/mod.rs
@@ -20,11 +20,11 @@ pub fn add_test_backtrace_footer(
     contracts_data: &ContractsData,
     encountered_errors: &EncounteredErrors,
     test_backtrace: Option<&TestBacktraceContext>,
+    test_panicked: bool,
     versioned_program_path: &Utf8Path,
     test_name: &str,
 ) -> String {
-    let has_backtrace =
-        !encountered_errors.is_empty() || test_backtrace.is_some_and(|bt| !bt.pcs.is_empty());
+    let has_backtrace = !encountered_errors.is_empty() || test_panicked;
 
     if !has_backtrace {
         return message;
--- a/crates/forge-runner/src/backtrace/mod.rs
+++ b/crates/forge-runner/src/backtrace/mod.rs
@@ -14,17 +14,43 @@ pub struct TestBacktraceContext {
     pub casm_start_offsets: Vec<usize>,
 }
 
+/// Backtraces-scoped outcome of a test run.
+///
+/// This mirrors the VM-run outcome, not the test verdict.
+/// E.g. `#[should_panic]` test case that panics as expected is still `Panic` here.
+pub enum TestBacktraceOutcome {
+    /// No panic occurred, so there's no backtrace.
+    Success,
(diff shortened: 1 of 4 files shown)


crates/forge-runner/src/backtrace/mod.rs
```diff
@@ -14,17 +14,43 @@ pub struct TestBacktraceContext {
...
     contracts_data: &ContractsData,
     encountered_errors: &EncounteredErrors,
-    test_backtrace: Option<&TestBacktraceContext>,
+    test_backtrace: &TestBacktraceOutcome,
     versioned_program_path: &Utf8Path,
     test_name: &str,
 ) -> String {
-    let has_backtrace =
-        !encountered_errors.is_empty() || test_backtrace.is_some_and(|bt| !bt.pcs.is_empty());
+    // Include hint even if backtrace capture was skipped (due to backtrace being disabled).
```
ddoktorski, line 52: Not ideal if enabling backtrace will show no results. I'm not sure if this is possible, but previously we were sure that there're some pcs saved to show. Is it a lot effort to keep the previous approach without losing the improvement from this PR?

DelevoXDG: > Is it a lot effort to keep the previous approach without losing the improvement from this PR?

`casm_start_offsets` optimization will remain, but the `pc` optimization will not.

> Not ideal if enabling backtrace will show no results. 

However, the **panic + zero backtrace data** outcome doesn't seem possible at the moment.

As of now, `get_reversed_pc_traceback` always yields >= 1 pc:

```rs
impl VirtualMachineExt for VirtualMachine {
    fn get_reversed_pc_traceback(&self) -> Vec<usize> {
        self.get_traceback_entries()
            .into_iter()
            // The cairo-vm implementation doesn't include the start location, so we add it manually.
            .chain(once((self.get_fp(), self.get_pc())))
            .rev()
            .map(|(_, pc)| pc)
            .filter(|pc| pc.segment_index == 0)
            .map(|pc| pc.offset + 1)
            .collect()
    }
}
```
because of `.chain(once((self.get_fp(), self.get_pc())))` that always injects the current PC.

So unless `get_reversed_pc_traceback` changes, we can assume that `is_panic() == true` is equivalent to there being >= 1 pc.

We can add a note so we're aware of that.

But even if that'd change for some reason, there's also a question of trade-off:  is it worth calculating PCs only to determine whether we display the hint or not in a very rare case that is **panic + no backtrace**?

ddoktorski: Thanks for checking. In this case let's only add a note and leave the current solution

ddoktorski (approved):

franciszekjob (approved):

Issue #1531, opened 2024-01-15
Typesafe contract `declare`


Arcticae:

Arcticae:
Artifacts with module info was merged
https://github.com/software-mansion/scarb/pull/1082

franciszekjob:
Closed in #4447

PR #4445, opened 2026-06-23
Support partial module paths in `snforge_std::declare`

franciszekjob:
Closes #

## Introduced changes

Support partial module paths in `snforge_std::declare`

## Checklist

- [ ] Linked relevant issue
- [ ] Updated relevant documentation
- [x] Added relevant tests
- [x] Performed self-review of the code
- [ ] Added changes to `CHANGELOG.md`

PR #4426, opened 2026-06-12
Use contract module paths instead of names in tests

franciszekjob:
Closes #

## Introduced changes

Refactor forge integration tests to use contract module paths instead of names.

## Checklist

- [ ] Linked relevant issue
- [ ] Updated relevant documentation
- [x] Added relevant tests
- [x] Performed self-review of the code
- [ ] Added changes to `CHANGELOG.md`

PR #4420, opened 2026-06-11
Support module path in `--contract-name`

franciszekjob:
Closes #

## Introduced changes

Allow to pass absolute/partial module tree path in `--contract-name` flag.

## Checklist

- [ ] Linked relevant issue
- [x] Updated relevant documentation
- [x] Added relevant tests
- [x] Performed self-review of the code
- [x] Added changes to `CHANGELOG.md`

PR #4447, opened 2026-06-24
Add `declare!` in `snforge_std`

franciszekjob:
Towards #1531

## Introduced changes

Introduce type-safe `declare!` macro in `snforge_std`. It accepts a contract name, full module path, or partial module path, validates that the argument is a plain Cairo path, and emits a compile-time type check against the contract `ContractState`. At runtime it still calls `snforge_std::declare` with the resolved path string, so existing declare behavior and artifact resolution are preserved.

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


franciszekjob:

- #4411
    - #4426
        - #4416
            - #4420
                - #4445
                    - #4447 ◀
                        - #4468
                            - #4469

franciszekjob:
note: We intentionally do not support paths with a leading `::` in `declare!`. The macro uses the provided path to generate a Cairo type path for the compile-time `ContractState` check, and Cairo does not accept leading `::` in that position. This remains supported only by runtime `declare("::...")`.

franciszekjob:
note: Do we want to prefer`declare!` in our docs now? If so, I will provide a separate PR for that.

docs/src/appendix/snforge-library/declare.md
```diff
@@ -29,3 +29,58 @@ Returns the `DeclareResult` that encapsulated possible outcomes in the enum:
...
+
+#[test]
+fn declare_by_alias() {
+    let contract = declare!(Alias).unwrap().contract_class();
+}
+```
+
+For contract names, the compile-time check uses the name visible in scope, but
+runtime artifact resolution still uses the name string. If multiple contracts
+share the same name, `declare!(MyContract)` can still fail due to am ambiguity. Use a
```
MKowalski8: typo?

```suggestion
share the same name, `declare!(MyContract)` can still fail due to an ambiguity. Use a
```

MKowalski8:
> note: Do we want to prefer `declare!` in our docs now? If so, I will provide a separate PR for that.

I think we can prefer `declare!`. But imo it's best to "vote" and, if necessary, discuss it on the channel with the rest of the team members.

docs/src/appendix/snforge-library/declare.md
```diff
@@ -29,3 +29,58 @@ Returns the `DeclareResult` that encapsulated possible outcomes in the enum:
...
+Accepted paths are:
+- full module tree paths (e.g. `my_package::module::MyContract`)
+- partial module tree paths (e.g. `module::MyContract`)
+- contract names (e.g. `MyContract`)
+
+Contract names require the contract to be in scope, either by being defined in the same module or imported.
+
+```rust
+use my_package::module::MyContract;
+use snforge_std::{ContractClassTrait, DeclareResultTrait, declare};
```
MKowalski8: Do we need to import `declare` in those test cases?

crates/snforge-scarb-plugin/src/inline_macros/declare.rs
```diff
@@ -0,0 +1,141 @@
...
+            ProcMacroResult::new(TokenStream::empty()).with_diagnostics(vec![diagnostic].into())
+        }
+    }
+}
+
+fn expand(args: &TokenStream) -> Result<TokenStream, Diagnostic> {
+    let raw_path = args.to_string();
+    let Some(contract_path) = normalize_path(&raw_path) else {
+        return Err(Diagnostic::span_error(
+            TextSpan::call_site(),
```
MKowalski8: nit: We could point exactly to wrong argument not `declare!` itself. At the moment it works like this (if I am not wrong):
```
error: `declare!` expects either a contract name ...
 --> tests/contract.cairo:5:5
  |
5 |     declare!("HelloStarknet");
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^
```
Not like this 
```
error: `declare!` expects either a contract name ...
 --> tests/contract.cairo:5:13
  |
5 |     declare!("HelloStarknet");
  |              ^^^^^^^^^^^^^^^
```

So we could check for real span e.g:

```rust
fn args_span(args: &TokenStream) -> TextSpan {
    let mut spans = args.tokens.iter().map(|token| match token {
        TokenTree::Ident(token) => &token.span,
    });
    let Some(first) = spans.next() else {
        return TextSpan::call_site();
    };
    let (mut start, mut end) = (first.start, first.end);
    for span in spans {
        start = start.min(span.start);
        end = end.max(span.end);
    }
    TextSpan::new(start, end)
}
```

And then use it here
```suggestion
            args_span(args),
```

But it's a UX thing, choose what you prefer

MKowalski8 (approved):
LGTM, left some comments

docs/src/appendix/snforge-library/declare.md
franciszekjob: No, good catch!

crates/snforge-scarb-plugin/src/inline_macros/declare.rs
```diff
@@ -0,0 +1,158 @@
...
+/// - a contract name (e.g. `MyContract`)
+/// - an absolute module tree path (e.g. `my_package::module::MyContract`)
+/// - a partial module tree path (e.g. `module::MyContract`).
+///
+/// Expects whitespace around `::` separators to be normalized away before validation.
+/// Rejects empty paths, empty segments, segments starting with digits, non-identifier
+/// characters and whitespace inside path segments.
+fn is_valid_contract_path(path: &str) -> bool {
+    CONTRACT_PATH_REGEX.is_match(path)
+}
```
DelevoXDG, line 113: If I understand correctly, current `CONTRACT_PATH_REGEX` will reject leading `::` in the contract identifier.

That creates a mismatch with current logic used by non-macro declare:

I wonder if we should allow it here as well as a part of the normalization flow. 


https://github.com/foundry-rs/starknet-foundry/blob/570f8a4f0a71c8c1d731013ee997f82474281b31/crates/sncast/src/helpers/artifacts.rs#L25-L27

WDYT?

docs/src/appendix/snforge-library/declare_macro.md
DelevoXDG, line 1: As discussed offline, we should document that it's type-safe only from the contract path existence perspective, not from contract contents perspective.

crates/forge/tests/e2e/declare_macro.rs
DelevoXDG, line 1: I'd also add assert that clearly shows that invalid contract contents is runtime error, differentiating it from `compile_from_file!` where it's compile-time error

crates/snforge-scarb-plugin/src/inline_macros/declare.rs
```diff
@@ -0,0 +1,158 @@
...
+    };
+
+    let (mut start, mut end) = (first.start, first.end);
+    for span in spans {
+        start = start.min(span.start);
+        end = end.max(span.end);
+    }
+
+    TextSpan::new(start, end)
+}
```
DelevoXDG, line 68: Can we provide some docstring for this helper?

crates/snforge-scarb-plugin/src/inline_macros/declare.rs
```diff
@@ -0,0 +1,158 @@
...
+    #[test_case("alias::HelloStarknet"; "partial module path")]
+    fn valid_contract_path(path: &str) {
+        assert!(is_valid_contract_path(path));
+    }
+
+    #[test_case("\"HelloStarknet\""; "non-path argument")]
+    #[test_case("my-package::HelloStarknet"; "invalid module path")]
+    #[test_case("1_Contract"; "identifier starting with digit")]
+    #[test_case("hello_starknet::1_Contract"; "path segment starting with digit")]
+    #[test_case("::HelloStarknet"; "leading empty segment")]
```
DelevoXDG: https://github.com/foundry-rs/starknet-foundry/pull/4447/changes#r3529910899

docs/listings/declare_examples/tests/test_declare_macro.cairo
```diff
@@ -0,0 +1,9 @@
+use declare_examples::hello_starknet::HelloStarknet;
+use snforge_std::DeclareResultTrait;
+
+#[test]
+fn test_declare() {
+    let contract = declare!(HelloStarknet).unwrap().contract_class();
+    // ...
```
DelevoXDG, line 7: Nit: Why? 😅

crates/snforge-scarb-plugin/src/inline_macros/declare.rs
franciszekjob: Good point. Addressed in https://github.com/foundry-rs/starknet-foundry/pull/4447/commits/5cbf7956e77802f7fde673c37a112092ae48c40b

docs/listings/declare_examples/tests/test_declare_macro.cairo
franciszekjob: It's here to hint the further code 😆

crates/snforge-scarb-plugin/src/inline_macros/declare.rs
franciszekjob: https://github.com/foundry-rs/starknet-foundry/pull/4447/commits/1795254ae143a38eb38b5799aad03cd5908175f2

crates/forge/tests/e2e/declare_macro.rs
franciszekjob: ccb43aeabe7e9c74cf69e70e58bf855aeb4bced3

docs/src/appendix/snforge-library/declare_macro.md
franciszekjob: https://github.com/foundry-rs/starknet-foundry/pull/4447/commits/1bb91ef3457be12b8166653650aaf08d4096e95e

docs/src/appendix/snforge-library/declare_macro.md
```diff
@@ -0,0 +1,49 @@
...
+Partial module tree paths require their first segment to be in scope:
+
+```rust
+{{#include ../../../listings/declare_examples/tests/test_declare_macro_partial_path.cairo}}
+```
+
+### Limitations
+
+`declare!` is type-safe only from the contract path existence perspective. It should not be treated as a
+compile-time guarantee about the contract contents.
```
DelevoXDG: nit:
```suggestion
`declare!` is type-safe only from the contract path existence perspective. IIt should not be treated as a compile-time guarantee that the contract is valid, or that compilation/declaration will succeed.
```

DelevoXDG (approved):

PR #4451, opened 2026-06-26
Remove error handling dead code

### LINKED ISSUE #3830: Investigate errors forge returns during execution
### Current State

It seems that there might be inconsistencies between how forge handles certain errors and blockifier and within forge itself as well.
For example, for some errors we return encoded byte arrays, for some single felts, maybe other formats as well.
It might be beneficial to check this and choose a consistent approach.

### Objective

Investigate errors forge returns during execution

### Additional Context

_No response_

ddoktorski:
Closes #3830

- Remove outdated error handling of `call_entry_point` result - removed arms are effectively a dead code except the `StateError` which can be handled by the catch-all arm
- Remove custom error parser


ddoktorski:

- #4451 ◀
    - #4452

franciszekjob (approved):

crates/cheatnet/src/runtime_extensions/outer_call_runtime_extension/rpc.rs
```diff
@@ -136,21 +82,9 @@ impl CallFailure {
...
-            error => {
-                let error_string = error.to_string();
-                if let Some(panic_data) = try_extract_panic_data(&error_string) {
-                    CallFailure::Recoverable { panic_data }
-                } else {
-                    CallFailure::Unrecoverable {
-                        msg: ByteArray::from(error_string.as_str()),
-                    }
-                }
-            }
```
DelevoXDG, line 160: Are we deliberately letting `StateError` to be routed through `error => CallFailure::Unrecoverable`?

If yes, the error message seems fine by me, and perhaps even better now:
```diff
- Unable to reach the node. Check your internet connection and node url
+ Failed to read from state: Unable to reach the node. Check your internet connection and node url.
```

However, the PR description only mentions removing dead code, and `StateReadError` arm is not unreachable.

ddoktorski: I probably should have mentioned this in the description, but yes it doesn't require a dedicated arm and as you mentioned, it is handled well by the catch-all arm

crates/cheatnet/src/runtime_extensions/outer_call_runtime_extension/rpc.rs
```diff
@@ -73,57 +65,11 @@ pub enum CallFailure {
...
 impl CallFailure {
     /// Maps blockifier-type error, to one that can be put into memory as panic-data (or re-raised)
     #[must_use]
-    pub fn from_execution_error(
-        err: &EntryPointExecutionError,
-        starknet_identifier: &AddressOrClassHash,
-    ) -> Self {
+    pub fn from_execution_error(err: &EntryPointExecutionError) -> Self {
         match err {
-            EntryPointExecutionError::ExecutionFailed { error_trace } => {
```
DelevoXDG, line 96: `EntryPointExecutionError::ExecutionFailed` is *potentially* reachable as well:

https://github.com/foundry-rs/starknet-foundry/blob/master/crates/cheatnet/src/runtime_extensions/outer_call_runtime_extension/execution/entry_point.rs#L210-L215 

Or are we assuming that `!enable_reverts` implies that error is `Unrecoverable`?

ddoktorski: `EntryPointExecutionError::ExecutionFailed` is raised only as a result of `non_reverting_execute_call_entry_point` function (`enable_reverts` is always `true`, so linked code is effectively a dead code). Therefore, it seems to be correct to classify it as `Unrecoverable`

DelevoXDG: > `enable_reverts` is always `true`, so linked code is effectively a dead code

That is an assumption, however, I don't think it's guaranteed `enable_reverts` is always true.  
- it's `true` in `0.14.2`
- and it's `false` in `0.13.2`: https://github.com/starkware-libs/sequencer/blob/8b090b27bca8a8366c63ae864507f59c78c8bdc5/crates/blockifier/resources/blockifier_versioned_constants_0_13_2.json#L52

In any case, it's a variable controlled by blockifier, not us.

If for some reason it changes again, `EntryPointExecutionError::ExecutionFailed` will become a possibility, and I think it would be difficult to notice/remember to re-add it.

If something is not a guaranteed dead code, I'd rather keep it, or at least assert early that we expect `enable_reverts` to be `true` (probably the better option if we assume it won't change).

DelevoXDG (approved):

crates/cheatnet/src/runtime_extensions/outer_call_runtime_extension/rpc.rs
ddoktorski: Since we always use the latest Starknet version, the behavior in previous versions won't impact our flow here. Beyond that, it's very unlikely to revert to false in the future since that would be a major breaking change for the network.

> or at least assert early that we expect enable_reverts to be true

Do you have anything specific in mind?

DelevoXDG: `assert!` / `debug_assert!` somewhere, e.g. at `build_block_context`.

Or, we can add a single test like:
```rs
#[test]
fn snforge_assumes_enable_reverts() {
    assert!(
        VersionedConstants::latest_constants().enable_reverts,
        "..."
    );
}
```

ddoktorski: Added test https://github.com/foundry-rs/starknet-foundry/pull/4451/changes/00f5c16a962506e8c2389b96241b1ae72ff619fc

PR #4455, opened 2026-06-29
Allow specifying private key in account creation

ddoktorski:
Allow speciyfing private key, if not provided generate a random pair as it worked so far.

PR #4456, opened 2026-06-29
Share common logic bettween `create` and `import` commands

ddoktorski:
[Refactor only]

### DIFF
--- a/crates/sncast/src/starknet_commands/account/mod.rs
+++ b/crates/sncast/src/starknet_commands/account/mod.rs
@@ -9,6 +9,7 @@ use camino::Utf8PathBuf;
 use clap::{Args, Subcommand};
 use configuration::resolve_config_file;
 use configuration::{load_config, search_config_upwards_relative_to};
+use conversions::string::{TryFromDecStr, TryFromHexStr};
 use serde_json::json;
 use sncast::helpers::account::{generate_account_name, load_accounts};
 use sncast::helpers::braavos::BraavosAccountFactory;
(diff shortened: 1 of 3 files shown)


ddoktorski:

- #4455
    - #4509
        - #4456 ◀

crates/sncast/src/starknet_commands/account/mod.rs
```diff
@@ -56,6 +57,42 @@ pub enum Commands {
...
+impl PrivateKeyArgs {
+    fn resolve_optional(&self) -> Result<Option<Felt>> {
+        match (&self.private_key, &self.private_key_file_path) {
+            (Some(key), _) => Ok(Some(*key)),
+            (None, Some(path)) => get_private_key_from_file(path)
+                .with_context(|| format!("Failed to obtain private key from the file {path}"))
+                .map(Some),
+            (None, None) => Ok(None),
+        }
+    }
```
franciszekjob, line 91: If `PrivateKeyArgs` is instantiated manually with both fields set, `private_key` always takes precedence over `private_key_file_path`. I know in practice we use this structure for clap only, but let's have a guard.

ddoktorski: There is now unreachable arm if both fields are set, as suggested in https://github.com/foundry-rs/starknet-foundry/pull/4455#discussion_r3641518765

franciszekjob (approved):

integraledelebesgue (approved):

PR #4399, opened 2026-06-01
Add predeployed contracts' ABI

franciszekjob:
Closes #4398

## Introduced changes

Add predeployed contracts' ABI which makes them work with debugging features.

## Checklist

- [x] Linked relevant issue
- [x] Updated relevant documentation
- [x] Added relevant tests
- [x] Performed self-review of the code
- [x] Added changes to `CHANGELOG.md`

PR #4461, opened 2026-06-30
Stop creating `ContractsDataStore` in each test run

ddoktorski:
This PR fixes a regression introduced in https://github.com/foundry-rs/starknet-foundry/pull/4399. Creating `ContractsDataStore` is expensive and it significantly affects total execution time, especially in packages containing many contracts.

### DIFF
--- a/crates/forge-runner/src/test_case_summary.rs
+++ b/crates/forge-runner/src/test_case_summary.rs
@@ -303,26 +303,39 @@ impl TestCaseSummary<Single> {
         gas_report_enabled: bool,
     ) -> Self {
         let name = test_case.name.clone();
-        let contracts_data_store = build_contracts_data_store(
-            contracts_data,
-            fork_data.as_ref(),
-            test_case.config.disable_predeployed_contracts,
-        );
+        let trace_components = trace_args.to_components();
+        // `ContractsDataStore` is expensive to build and is needed only by gas report and debugging trace.
+        // Keep this lazy so normal test runs do not parse and extract contract artifacts for every test case.
+        let contracts_data_store = (gas_report_enabled || trace_components.is_some()).then(|| {
+            build_contracts_data_store(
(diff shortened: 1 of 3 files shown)


franciszekjob (approved):

MKowalski8 (approved):

crates/forge-runner/src/test_case_summary.rs
```diff
@@ -303,26 +303,39 @@ impl TestCaseSummary<Single> {
...
+        let trace_components = trace_args.to_components();
+        // `ContractsDataStore` is expensive to build and is needed only by gas report and debugging trace.
+        // Keep this lazy so normal test runs do not parse and extract contract artifacts for every test case.
+        let contracts_data_store = (gas_report_enabled || trace_components.is_some()).then(|| {
+            build_contracts_data_store(
+                contracts_data,
+                fork_data.as_ref(),
+                test_case.config.disable_predeployed_contracts,
+            )
+        });
```
DelevoXDG, line 315: The problem still remains for `--gas-report`, `--trace-*` cases.

I wonder if we should try to make this solution universal, e.g. introducing `Arc<ContractsDataStore` / `OnceLock` and sharing it?

If I understand correctly, for non-fork tests `ContractsDataStore ` is effectively the same.

Can be done as separate issue ofc if you prefer not to block this, but I think it would better to solve this for all cases.

Thoughts?

piotmag769 (approved):

crates/forge-runner/src/test_case_summary.rs
ddoktorski: Good point, but this is rather uncommon to run multiple tests with any of these features enables, so not crucial for now. Added issue https://github.com/foundry-rs/starknet-foundry/issues/4467

DelevoXDG (approved):

PR #4416, opened 2026-06-11
(sncast) Fix ambiguous contract name resolution

franciszekjob:
Closes #2626

## Introduced changes

Fix non-deterministic contract resolution if multiple contracts share the same name.

## Checklist

- [x] Linked relevant issue
- [x] Updated relevant documentation
- [x] Added relevant tests
- [x] Performed self-review of the code
- [x] Added changes to `CHANGELOG.md`

PR #4468, opened 2026-07-01
Add `snforge_std::declare_from_file`

### LINKED ISSUE #3266: Cheatcode for declaring already compiled contract
### Current State

Currently it is only possible to declare contract through a cheatcode by it's name. Ale the contracts for this declaration have to be first compiled by Scarb and USC (this takes place before test execution).

There are use cases (e.g. testing upgrading a large contract), where this compilation before tests execution slows things down.

### Objective

Cheatcode for declaring already compiled contract

### Additional Context

It should accept the path to the compiled contract and declare that.

There are possible problems with this solution:

- Relative paths (and paths in general) could resolve to different things depending on the feature implementation if `snforge test` is run from different directories. A possible logical solution would be to treat workspace root (or package? depending on how Cargo treats that) as a root.
- We currently compile all contracts from the project (and external that are defineded in Scarb.toml). For the contract declared by path to not be compiled, it would have to be located outside of the tests project or some other mechanism would have to be implemented.

Details should be clarified in a short design document first.

franciszekjob:
Closes #3266

## Introduced changes

Add `snforge_std::declare_from_file` function which allows to declare a contract by passing a path to Sierra contract class JSON

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests

### DIFF
--- a/crates/cheatnet/src/runtime_extensions/forge_runtime_extension/cheatcodes/declare.rs
+++ b/crates/cheatnet/src/runtime_extensions/forge_runtime_extension/cheatcodes/declare.rs
@@ -61,7 +61,16 @@ pub fn declare(
 pub fn declare_from_file(
     state: &mut dyn State,
     sierra_path: &Path,
+    contracts_data: &ContractsData,
 ) -> Result<DeclareResult, CheatcodeError> {
+    if let Some(contract) = contracts_data.find_by_sierra_path(sierra_path) {
+        return declare_contract_class(
+            state,
+            contract.class_hash,
+            get_contract_class(&contract.artifacts),
+        );
+    }
+
     let sierra = std::fs::read_to_string(sierra_path).map_err(|error| {
         CheatcodeError::Unrecoverable(EnhancedHintError::from(anyhow!(
             "Failed to read Sierra file at {}: {error}",
--- a/crates/cheatnet/src/runtime_extensions/forge_runtime_extension/cheatcodes/declare.rs
+++ b/crates/cheatnet/src/runtime_extensions/forge_runtime_extension/cheatcodes/declare.rs
@@ -9,7 +9,7 @@ use blockifier::execution::contract_class::{CompiledClassV1, RunnableCompiledCla
 use blockifier::execution::native::contract_class::NativeCompiledClassV1;
 use blockifier::state::{errors::StateError, state_api::State};
 #[cfg(feature = "cairo-native")]
-use cairo_lang_starknet_classes::contract_class::ContractClass as CairoContractClass;
+use cairo_lang_starknet_classes::contract_class::ContractClass;
 use conversions::IntoConv;
 use conversions::serde::serialize::CairoSerialize;
 use indoc::formatdoc;
@@ -121,7 +121,7 @@ pub fn declare_from_file(
         })?;
     #[cfg(feature = "cairo-native")]
     let contract_class = if contracts_data.run_native {
-        let sierra_class: CairoContractClass = serde_json::from_str(&sierra).map_err(|error| {
+        let sierra_class: ContractClass = serde_json::from_str(&sierra).map_err(|error| {
             CheatcodeError::Unrecoverable(EnhancedHintError::from(anyhow!(
                 "Failed to parse Sierra contract class JSON for native execution at {}: {error}",
                 sierra_path.display()
--- a/crates/cheatnet/src/runtime_extensions/forge_runtime_extension/cheatcodes/declare.rs
+++ b/crates/cheatnet/src/runtime_extensions/forge_runtime_extension/cheatcodes/declare.rs
@@ -83,13 +83,12 @@ pub fn declare_from_file(
             sierra_path.display()
         )))
     })?;
-    let sierra_class: SierraClass =
-        serde_json::from_value(sierra_json.clone()).map_err(|error| {
-            CheatcodeError::Unrecoverable(EnhancedHintError::from(anyhow!(
-                "Failed to parse Sierra contract class JSON at {}: {error}",
-                sierra_path.display()
-            )))
-        })?;
+    let sierra_class: SierraClass = serde_json::from_str(&sierra).map_err(|error| {
+        CheatcodeError::Unrecoverable(EnhancedHintError::from(anyhow!(
+            "Failed to parse Sierra contract class JSON at {}: {error}",
+            sierra_path.display()
+        )))
+    })?;
     let class_hash = get_class_hash(&sierra_class).map_err(|error| {
         CheatcodeError::Unrecoverable(EnhancedHintError::from(anyhow!(
             "Failed to calculate class hash for Sierra file at {}: {error}",
--- a/crates/cheatnet/src/runtime_extensions/forge_runtime_extension/cheatcodes/declare.rs
+++ b/crates/cheatnet/src/runtime_extensions/forge_runtime_extension/cheatcodes/declare.rs
@@ -77,12 +77,6 @@ pub fn declare_from_file(
             sierra_path.display()
         )))
     })?;
-    let sierra_json: serde_json::Value = serde_json::from_str(&sierra).map_err(|error| {
-        CheatcodeError::Unrecoverable(EnhancedHintError::from(anyhow!(
-            "Failed to parse Sierra contract class JSON at {}: {error}",
-            sierra_path.display()
-        )))
-    })?;
     let sierra_class: SierraClass = serde_json::from_str(&sierra).map_err(|error| {
         CheatcodeError::Unrecoverable(EnhancedHintError::from(anyhow!(
             "Failed to parse Sierra contract class JSON at {}: {error}",
@@ -92,6 +86,21 @@ pub fn declare_from_file(
     let class_hash = get_class_hash(&sierra_class).map_err(|error| {
         CheatcodeError::Unrecoverable(EnhancedHintError::from(anyhow!(
             "Failed to calculate class hash for Sierra file at {}: {error}",
+            sierra_path.display()
+        )))
+    })?;
+
+    if let Some(contract) = contracts_data.get_contract_by_class_hash(&class_hash) {
+        return declare_contract_class(
+            state,
+            contract.class_hash,
+            get_contract_class(&contract.artifacts),
+        );
+    }
+
+    let sierra_json: serde_json::Value = serde_json::from_str(&sierra).map_err(|error| {
+        CheatcodeError::Unrecoverable(EnhancedHintError::from(anyhow!(
+            "Failed to parse Sierra contract class JSON at {}: {error}",
             sierra_path.display()
         )))
     })?;
--- a/crates/cheatnet/src/runtime_extensions/forge_runtime_extension/cheatcodes/declare.rs
+++ b/crates/cheatnet/src/runtime_extensions/forge_runtime_extension/cheatcodes/declare.rs
@@ -8,6 +8,8 @@ use blockifier::execution::contract_class::{CompiledClassV1, RunnableCompiledCla
 #[cfg(feature = "cairo-native")]
 use blockifier::execution::native::contract_class::NativeCompiledClassV1;
 use blockifier::state::{errors::StateError, state_api::State};
+#[cfg(feature = "cairo-native")]
+use cairo_lang_starknet_classes::contract_class::ContractClass as CairoContractClass;
 use conversions::IntoConv;
 use conversions::serde::serialize::CairoSerialize;
 use indoc::formatdoc;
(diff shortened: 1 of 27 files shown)


franciszekjob:

- #4411
    - #4426
        - #4416
            - #4420
                - #4445
                    - #4447
                        - #4468 ◀
                            - #4499
                                - #4469

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/cheatcodes/declare.rs
```diff
@@ -49,10 +51,65 @@ pub fn declare(
...
+            "Failed to calculate class hash for Sierra file at {}: {error}",
+            sierra_path.display()
+        )))
+    })?;
+    let casm = compile_contract_sierra(&sierra_json).map_err(|error| {
+        CheatcodeError::Unrecoverable(EnhancedHintError::from(anyhow!(
+            "Failed to compile Sierra file at {}: {error}",
+            sierra_path.display()
+        )))
+    })?;
```
DelevoXDG, line 114: I wonder whether we should separately handle the case where the `sierra_path` points to an in-package contract.

Currently, in this case we basically compile the same Sierra twice:

1. At load-time
   https://github.com/foundry-rs/starknet-foundry/blob/570f8a4f0a71c8c1d731013ee997f82474281b31/crates/scarb-api/src/artifacts.rs#L93-L113

2. Here in `declare_from_file`, again

I'm not sure if this has significant perf impact, but theoretically it could, since it's still duplication:

I would consider adding a check that would prevent a repeat compile
```diff
// mod.rs - pass `contracts_data`
- handle_declare_result(declare_from_file(*state, Path::new(&sierra_path)))
+ handle_declare_result(declare_from_file(*state, Path::new(&sierra_path), self.contracts_data))

// declare.rs - early match
+ if let Some(contract) = contracts_data.find_by_sierra_path(sierra_path) {
+     return declare_contract_class(state, contract.class_hash, get_contract_class(&contract.artifacts));
+ }
  // existing read + USC path for external JSON
```

Thoughts?

docs/src/appendix/snforge-library/declare_from_file.md
DelevoXDG, line 1: We should probably document the fact that paths are relative to the package root

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/cheatcodes/declare.rs
franciszekjob: Good point 👍  
https://github.com/foundry-rs/starknet-foundry/pull/4468/changes/6df096851f550c9d9695798f74c9555c8c9197a0

DelevoXDG (commented):
Please rebase 🙏🏻

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/cheatcodes/declare.rs
```diff
@@ -49,10 +51,74 @@ pub fn declare(
...
+    let contract_class = RunnableCompiledClass::V1(
```
MKowalski8: What about the `RunnableCompiledClass:V1Native`, shouldn't we add support for it as well? 

At the moment it will be handled only in the first case in the function:
https://github.com/foundry-rs/starknet-foundry/blob/57136f8282e7d70b4ad77b8ba52405a090c577b3/crates/cheatnet/src/runtime_extensions/forge_runtime_extension/cheatcodes/declare.rs#L66

It calls [get_contract_class](https://github.com/foundry-rs/starknet-foundry/blob/57136f8282e7d70b4ad77b8ba52405a090c577b3/crates/cheatnet/src/runtime_extensions/forge_runtime_extension/cheatcodes/declare.rs#L153) which handles `V1Native`

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/cheatcodes/declare.rs
```diff
@@ -49,10 +51,74 @@ pub fn declare(
...
+        )))
+    })?;
+    let sierra_json: serde_json::Value = serde_json::from_str(&sierra).map_err(|error| {
+        CheatcodeError::Unrecoverable(EnhancedHintError::from(anyhow!(
+            "Failed to parse Sierra contract class JSON at {}: {error}",
+            sierra_path.display()
+        )))
+    })?;
+    let sierra_class: SierraClass =
+        serde_json::from_value(sierra_json.clone()).map_err(|error| {
```
MKowalski8: We could drop this `clone()` and just do `from_str` here. 

```suggestion
    let sierra_class: SierraClass = serde_json::from_str(&sierra).map_err(|error| {
```

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/cheatcodes/declare.rs
franciszekjob: Yes, however a raw Sierra file does not contain the native executor, so we can’t always build `V1Native` in the fallback path.

I updated it to compute the class hash first and check `ContractsData` by class hash. If the contract is loaded
there, we reuse `get_contract_class(&contract.artifacts)`, so `V1Native` is preserved. Truly external Sierra files still fall back to V1.

I also added a regression test for the class-hash lookup path.

crates/cheatnet/tests/cheatcodes/declare.rs
```diff
@@ -330,3 +338,96 @@ fn declare_ambiguous_partial_module_path() {
...
+        .unwrap()
+        .unwrap_success();
+
+    assert_eq!(class_hash, expected_class_hash);
+
+    #[cfg(feature = "cairo-native")]
+    assert!(matches!(
+        cached_state.get_compiled_class(class_hash).unwrap(),
+        RunnableCompiledClass::V1Native(_)
+    ));
```
DelevoXDG, line 430: The test actually fails locally.

This is hidden by the fact that it runs as part of our "workspace tests": https://github.com/foundry-rs/starknet-foundry/actions/runs/29267318485/job/86876053774?pr=4468#step:4:230

These tests actually do not enable `cairo-native` feature:
https://github.com/foundry-rs/starknet-foundry/blob/979a70adf76c25ca5932732ec30910b37e6db469/.github/workflows/ci.yml#L49-L50

I've also encountered these type of errors while attempting to build this:

```
error[E0609]: no field `executor` on type `&StarknetContractArtifacts`
   --> crates/cheatnet/src/runtime_extensions/forge_runtime_extension/cheatcodes/declare.rs:158:37
    |
158 |     return match &contract_artifact.executor {
    |                                     ^^^^^^^^ unknown field
    |
    = note: available fields are: `sierra`, `casm`
```

It seems we need to make sure to propagate the feature flag to `scarb-api`:

```diff
# crates/cheatnet/Cargo.toml
- cairo-native = ["dep:cairo-native", "blockifier/cairo_native"]
+ cairo-native = ["dep:cairo-native", "blockifier/cairo_native", "scarb-api/cairo-native"]
```

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/contracts_data.rs
```diff
@@ -138,6 +139,23 @@ impl ContractsData {
...
+        let sierra_path = sierra_path.canonicalize().ok()?;
+
+        self.contracts.values().find(|contract| {
+            contract
+                .source_sierra_path
+                .as_std_path()
+                .canonicalize()
+                .is_ok_and(|path| path == sierra_path)
+        })
+    }
```
DelevoXDG, line 162: What is our intention here:
1. allow re-use by correct sierra path only?
2. allow re-use by class hash, regardless of the file path?

This logic seems to hint at 1., but the test `declare_from_file_uses_loaded_artifact_when_class_hash_matches` seems to imply 2 (from name and the fact that it copies the file)

Which one do you think is more desirable? 

Personally, in my original [comment](https://github.com/foundry-rs/starknet-foundry/pull/4468#discussion_r3530241428) I think I've probably implied 1 😅, since this is pure deduplication. 

But if you think option 2. is viable and requires no significant extra effort / has no additional associated problems, it's probably something we can consider as well 🤔

DelevoXDG: Note: Option. 1 seems straightforward, but 2. would probably require a mention in a docs as it might be unexpected

crates/cheatnet/tests/cheatcodes/declare.rs
franciszekjob: 07ba5b25fe12270aadebc601655f83edc0acc3e5

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/contracts_data.rs
franciszekjob: I think supporting both is reasonable here. Path match is the fast path for the original dedup case. If the path differs, we still parse the Sierra and reuse a loaded artifact when the class hash matches. Mentioned this class hash re-use behavior in docs too.

Addressed in https://github.com/foundry-rs/starknet-foundry/pull/4468/commits/463e7ccf54d0a0d68abaa2eaada2cc5e3714134d

crates/cheatnet/tests/cheatcodes/declare.rs
franciszekjob: Also, added job with cheatnet tests with `cairo-native` flag in https://github.com/foundry-rs/starknet-foundry/pull/4499

DelevoXDG (approved):

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/cheatcodes/declare.rs
MKowalski8: Can we add a short comment exaplaining that? Why this is only `V1` wthiout `V1Natvie` here? Imo it could be useful

crates/forge/tests/data/declare_from_file/Scarb.toml
```diff
@@ -0,0 +1,17 @@
...
+starknet = "2.16.0"
+
+[dev-dependencies]
+snforge_std = { path = "../../../../../snforge_std" }
+
+[[target.starknet-contract]]
+
+[profile.dev.cairo]
+unstable-add-statements-code-locations-debug-info = true
+unstable-add-statements-functions-debug-info = true
```
MKowalski8: nit:
```suggestion
add-statements-code-locations-debug-info = true
add-statements-functions-debug-info = true
```

MKowalski8 (approved):

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/cheatcodes/declare.rs
franciszekjob: Actually, after looking into it more closely we can add an explicit support for native. TAL https://github.com/foundry-rs/starknet-foundry/pull/4468/commits/378448c3bc393c044f8017c39d734e31bb376a4b

franciszekjob: Tests are run in https://github.com/foundry-rs/starknet-foundry/pull/4499

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/contracts_data.rs
```diff
@@ -43,6 +45,11 @@ pub enum ContractResolutionError {
 
 impl ContractsData {
     pub fn try_from(contracts: ScarbContractsData) -> Result<Self> {
+        #[cfg(feature = "cairo-native")]
+        let run_native = contracts
```
MKowalski8: This is not used. Only place when you run `try_from_with_run_native` is [here:](https://github.com/foundry-rs/starknet-foundry/pull/4468/changes/378448c3bc393c044f8017c39d734e31bb376a4b#diff-3122a4101cf70b9e1b0731e654e157613dd784f03c46890e2eb459ac3bcd7418R110)
```rust
let contracts_data = ContractsData::try_from_with_run_native(contracts, run_native)?;
```

So it drops this auto computation anyway. 

Also function overload is not needed as well (for `try_from_with_run_native`). Or is there a reason for that?

franciszekjob: Simplified in https://github.com/foundry-rs/starknet-foundry/pull/4468/commits/226a5d1ab212db5cf654fc821819690056ca9957

MKowalski8 (approved):

PR #4408, opened 2026-06-03
Fix `#[should_panic]` matching for byte array panic data

franciszekjob:
Closes #

## Introduced changes

Fix `#[should_panic(expected: (...))]` matching for panic data containing `ByteArray` values inside tuples.

Tuple items now include byte array magic encoding as standalone byte arrays. This was unfortunately introduced while adding support for strings inside `#[should_panic(expected: (...))]` in https://github.com/foundry-rs/starknet-foundry/commit/1cc1d571ab1875650c2aa6698096aa4194216dad.

Additionally, this fixes an edge case where a `ByteArray` payload equal to `"ENTRYPOINT_FAILED"` was incorrectly treated as an explicit generic-error matcher. The matcher now only applies exact generic-error comparison to non `ByteArray` expected data.

## Checklist

- [ ] Linked relevant issue
- [ ] Updated relevant documentation
- [x] Added relevant tests
- [x] Performed self-review of the code
- [x] Added changes to `CHANGELOG.md`

PR #4472, opened 2026-07-02
Improve `#[should_panic]` mismatch message formatting

franciszekjob:
Closes #

## Introduced changes

Follow-up to https://github.com/foundry-rs/starknet-foundry/pull/4408#pullrequestreview-4563879714

Format `Actual` and `Expected` panic data in the incorrect panic data message using `format_panic_data`, the same formatter used for regular failure data. Byte arrays are now displayed as quoted strings and felts as hex values with a short string representation.

Additionally, validate byte array data in `ByteArray::deserialize_with_magic` (pending word length and word sizes).

## Checklist

- [ ] Linked relevant issue
- [x] Added relevant tests


crates/conversions/src/byte_array.rs
```diff
@@ -102,6 +113,54 @@ mod tests {
...
+            ByteArray::deserialize_with_magic(&serialized).unwrap(),
+            array
+        );
+    }
+
+    #[test]
+    fn test_deserialize_with_magic_pending_word_len_too_large() {
+        // [magic, num_full_words, pending_word, pending_word_len]
+        let mut data = ByteArray::from("x").serialize_with_magic();
+        *data.last_mut().unwrap() = Felt::from(100_u8);
```
DelevoXDG: I think it's better to address the very edge of the edge case
```suggestion
        *data.last_mut().unwrap() = Felt::from(32_u8);
```

crates/forge-runner/src/test_case_summary.rs
```diff
@@ -661,6 +648,59 @@ mod tests {
...
+            "\n    Incorrect panic data\
+             \n    Actual:    (\"this_string_is_longer_than_31_bytes\", 0xb, \"hello\", 0x5, 0x73686f72745f737472696e67 ('short_string'))\
+             \n    Expected:  (\"this_string_is_longer_than_31_bytes\", 0xb, \"helloo\", 0x5, 0x73686f72745f737472696e67 ('short_string'))\n"
+        );
+    }
+
+    #[test]
+    fn test_incorrect_panic_data_message_malformed_byte_array() {
+        // Magic followed by data that is not a valid `ByteArray` serialization
+        // must not crash and should fall back to felt-by-felt formatting.
```
DelevoXDG, line 770: This makes me wonder whether we should actually clearly hint somewhere that the actual output is not even proper `ByteArray`?

```
"\n    Incorrect panic data\
 \n    Actual:    (0x46a6158a16a947e5916b2a2ca68501a45e93d7110e81aa2d6438b1c57c879a3, 0x0 (''), 0x78 ('x'), 0x64 ('d'))\
 \n    Expected:  0x70616e6963206d657373616765 ('panic message')\n"
```

From this, it's difficult to infer that `ByteArray` was even intended, but is malformed 🤔 

I wonder if there's a place here in the error message for information that conveys that, wdyt?

crates/forge/tests/e2e/running.rs
```diff
@@ -1061,25 +1061,32 @@ fn should_panic() {
 
         Failure data:
             Incorrect panic data
-            Actual:    [0x546869732077696c6c2070616e6963] (This will panic)
-            Expected:  [0x46a6158a16a947e5916b2a2ca68501a45e93d7110e81aa2d6438b1c57c879a3, 0x0, 0x546869732077696c6c2070616e6963, 0xf] (This will panic)
+            Actual:    0x546869732077696c6c2070616e6963 ('This will panic')
+            Expected:  "This will panic"
```
DelevoXDG: This seems a bit problematic to me.

Original version allowed you to clearly see the felt-based difference:

```
            Actual:    [0x546869732077696c6c2070616e6963] (This will panic)
            Expected:  [0x46a6158a16a947e5916b2a2ca68501a45e93d7110e81aa2d6438b1c57c879a3, 0x0, 0x546869732077696c6c2070616e6963, 0xf] (This will panic)
```

Now, there's seemingly no clear indication that the error reason is that one is `shortstring`, and other is a `ByteArray`.

One might mistakenly assume that `should_panic` is just buggy.

Should we consider how to signal it to the user more clearly?

crates/forge-runner/src/test_case_summary.rs
franciszekjob: Good point. The fallback is technically correct, but the message hides why we stopped rendering it as a `ByteArray`. How about adding a note for the case where the data starts with `ByteArray` magic but fails `ByteArray` deserialization, so it’s clear that this looked like an intended `ByteArray` but was malformed?

e.g.

```
Incorrect panic data
Actual:    (0x46a..., 0x0 (''), 0x78 ('x'), 0x64 ('d'))
Expected:  0x70616e6963206d657373616765 ('panic message')
Note: actual panic data starts with ByteArray magic, but is not a valid ByteArray serialization; showing raw felts.
```

crates/forge/tests/e2e/running.rs
franciszekjob: Good point. I see a few possible approaches:

1. Restore old raw felt rendering for mismatches. 
This is the clearest about the exact data layout, but it regresses readability for normal ByteArray mismatches.

e.g.:
```
Incorrect panic data
Actual:    [0x546869732077696c6c2070616e6963] (This will panic)
Expected:  [0x46a6158a16a947e5916b2a2ca68501a45e93d7110e81aa2d6438b1c57c879a3, 0x0, 0x546869732077696c6c2070616e6963,
0xf] (This will panic)
```

2. Show readable values plus raw felts.
This preserves all information from master, but makes every mismatch more verbose.

e.g.:

```
Incorrect panic data
Actual:    0x546869732077696c6c2070616e6963 ('This will panic')
           raw: [0x546869732077696c6c2070616e6963]
Expected:  "This will panic"
           raw: [0x46a6158a16a947e5916b2a2ca68501a45e93d7110e81aa2d6438b1c57c879a3, 0x0,
           0x546869732077696c6c2070616e6963, 0xf]
```

and for tuples

```
Incorrect panic data
Actual:    ("this_string_is_longer_than_31_bytes", 0xb, "hello", 0x5, 0x73686f72745f737472696e67 ('short_string'))
           raw: [0x46a6158a16a947e5916b2a2ca68501a45e93d7110e81aa2d6438b1c57c879a3, 0x1,
           0x746869735f737472696e675f69735f6c6f6e6765725f7468616e5f33315f62, 0x79746573, 0x4, 0xb,
           0x46a6158a16a947e5916b2a2ca68501a45e93d7110e81aa2d6438b1c57c879a3, 0x0, 0x68656c6c6f, 0x5, 0x5,
           0x73686f72745f737472696e67]
Expected:  ("this_string_is_longer_than_31_bytes", 0xb, "world", 0x5, 0x73686f72745f737472696e67 ('short_string'))
           raw: [0x46a6158a16a947e5916b2a2ca68501a45e93d7110e81aa2d6438b1c57c879a3, 0x1,
           0x746869735f737472696e675f69735f6c6f6e6765725f7468616e5f33315f62, 0x79746573, 0x4, 0xb,
           0x46a6158a16a947e5916b2a2ca68501a45e93d7110e81aa2d6438b1c57c879a3, 0x0, 0x776f726c64, 0x5, 0x5,
           0x73686f72745f737472696e67]
```

3. Add a hint only when the encodings differ.
This is minimally invasive, but the key distinction is separated from the actual/expected lines.

  For example:
```
  Incorrect panic data
  Actual:    0x546869732077696c6c2070616e6963 ('This will panic')
  Expected:  "This will panic"

  Hint: actual panic data is a felt252 short string, while expected panic data is a ByteArray string.
```
  Or the opposite case:

```
  Incorrect panic data
  Actual:    "This will panic"
  Expected:  0x546869732077696c6c2070616e6963 ('This will panic')

  Hint: actual panic data is a ByteArray string, while expected panic data is a felt252 short string.
```

For option 3, there probably wouldn’t be an additional hint for the mixed tuple, because the mismatch is simply "hello" vs "world" with the same type/encoding.

4. Label the rendered values inline.

e.g.:

```
Incorrect panic data
Actual:    "This will panic" (ByteArray)
Expected:  'This will panic' (felt252)
```

and for tuples

```
Actual:    (ByteArray("this_string_is_longer_than_31_bytes"), felt252 0xb, ByteArray("hello"), felt252 0x5, felt252 0x73686f72745f737472696e67 ('short_string'))
Expected:  (ByteArray("this_string_is_longer_than_31_bytes"), felt252 0xb, ByteArray("world"), felt252 0x5, felt252 0x73686f72745f737472696e67 ('short_string'))
```

I think option 4 is the best balance. It keeps the readable string output, but makes the reason for the mismatch explicit at the point where the user compares actual and expected values.

WDYT?

MKowalski8: IMO, the 4 seems resonable. It can be because there is an example, and it's easier to imagine 😅.
Alternatively 2 would be good enough

DelevoXDG: Option 4 seems best

crates/forge-runner/src/test_case_summary.rs
franciszekjob: https://github.com/foundry-rs/starknet-foundry/pull/4472/changes/cf9489dfe6463b884abfe4f587027250ffd93282

DelevoXDG (approved):

crates/forge-runner/src/test_case_summary.rs
```diff
@@ -11,17 +11,25 @@ use crate::package_tests::with_config_resolved::TestCaseWithResolvedConfig;
...
 use starknet_api::execution_resources::GasVector;
+use starknet_api::execution_utils::format_panic_data;
 use starknet_types_core::felt::Felt;
 use std::fmt;
 use std::option::Option;
+use std::sync::LazyLock;
+
+static BYTE_ARRAY_MAGIC_FELT: LazyLock<Felt> =
+    LazyLock::new(|| TryFromHexStr::try_from_hex_str(&format!("0x{BYTE_ARRAY_MAGIC}")).unwrap());
+const BYTE_ARRAY_FIXED_PART_LEN: usize = 4;
```
MKowalski8, line 36: nit: We could add a comment informing what is part of this `fixed_part_len`, meaning why it's 4

CHANGELOG.md
```diff
@@ -14,6 +14,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `declare!` macro for declaring contracts using Cairo paths. Read more [here](https://foundry-rs.github.io/starknet-foundry/appendix/snforge-library/declare_macro.html).
 - Debugging traces now display the class hash for forked contracts.
 
+#### Changed
+
+- `Actual` and `Expected` panic data in the `#[should_panic]` mismatch message now labels rendered values inline, making it clear whether each value is a `ByteArray` or a `felt252`.
+
+#### Fixed
+
+- Malformed byte array panic data (e.g. an invalid `pending_word_len`) no longer causes a crash when matching `#[should_panic]` expected data.
```
MKowalski8: nit: From what I've checked `pending_word_len` didn't cause a crash it only generated wrong output. The `Vec::with_capacity(length)` could cause a crash. (but please correct me if I'am wrong)

We can rephrase it. For instance: 
```suggestion
- Malformed byte array panic data (e.g. an untrusted words length) no longer causes a crash when matching `#[should_panic]` expected data. Additionally, invalid byte array serializations (e.g. bad `pending_word_len`) are now rejected instead of rendered as a corrupted string.
```

MKowalski8 (approved):
LGTM, left two nits

PR #4481, opened 2026-07-06
Bump scarb 2.19.1; Re-gen snaps

DelevoXDG:


MKowalski8 (approved):
Please only run the scheduled tests from this branch to double check if we don't miss something. Also I would delete the old `2.16.1` snaps, since we don't need them anymore.

DelevoXDG:
> Please only run the scheduled tests from this branch to double check if we don't miss something. 

https://github.com/foundry-rs/starknet-foundry/actions/runs/28793080218

> Also I would delete the old `2.16.1` snaps, since we don't need them anymore.

👍

franciszekjob (approved):

Issue #4484, opened 2026-07-06
[tests] `assert_gas` should show `expected` vs `actual` on failure


DelevoXDG:
### Current state

[`assert_gas`](https://github.com/foundry-rs/starknet-foundry/blob/0f1ce2a1e5531b32a73601b4d98f7237bc1d015f/crates/forge/tests/utils/runner.rs#L290-L312) panics with the full `iter().any(...)` predicate and does not print expected or actual gas, so a mismatch is hard to diagnose from the test failure alone:

```
    assertion failed: result.test_case_summaries.iter().any(|any_case|
            {
                match any_case {
                    AnyTestCaseSummary::Fuzzing(_) => {
                        panic!("Cannot use assert_gas! for fuzzing tests")
                    }
                    AnyTestCaseSummary::Single(case) =>
                        match case {
                            TestCaseSummary::Passed { gas_info: gas, .. } => {
                                assert_gas_with_margin(gas.gas_used, asserted_gas) &&
                                    any_case.name().unwrap().ends_with(test_name_suffix.as_str())
                            }
                            _ => false,
                        },
                }
            })
    note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

### Objective

On failure, print:
- expected vs actual gas values
- test case name

PR #4494, opened 2026-07-16
Show gas difference on assertion failure

franciszekjob:
Towards #4484

## Introduced changes

Show gas diff on [`assert_gas`](https://github.com/foundry-rs/starknet-foundry/blob/0f1ce2a1e5531b32a73601b4d98f7237bc1d015f/crates/forge/tests/utils/runner.rs#L290-L312) failure.

### Before

image

### After

image

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests

### DIFF
--- a/crates/forge/tests/integration/gas.rs
+++ b/crates/forge/tests/integration/gas.rs
@@ -1,12 +1,9 @@
-use crate::utils::runner::{Contract, assert_gas, assert_passed, capture_assertion_panic};
+use crate::utils::runner::{Contract, assert_gas, assert_passed};
 use crate::utils::running_tests::run_test_case;
 use crate::utils::test_case;
 use forge_runner::forge_config::ForgeTrackedResource;
-use forge_runner::test_case_summary::{AnyTestCaseSummary, TestCaseSummary};
-use forge_runner::test_target_summary::TestTargetSummary;
 use indoc::{formatdoc, indoc};
 use shared::test_utils::node_url::node_rpc_url;
-use shared::test_utils::output_assert::assert_stdout_contains;
 use starknet_api::execution_resources::{GasAmount, GasVector};
 use std::path::Path;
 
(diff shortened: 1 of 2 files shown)


franciszekjob:

- #4494 ◀
    - #4501
        - #4500
            - #4513

crates/forge/tests/utils/runner.rs
```diff
@@ -266,6 +271,25 @@ pub fn assert_failed(result: &[TestTargetSummary]) {
...
+
+    if let Some(message) = payload.downcast_ref::<String>() {
+        message.clone()
+    } else if let Some(message) = payload.downcast_ref::<&str>() {
+        (*message).to_string()
+    } else {
+        panic!("Unexpected panic payload type. Expected `String` or `&str`.");
+    }
+}
+
```
franciszekjob, line 292: note: It will be used in #4501 too.

crates/forge/tests/integration/gas.rs
```diff
@@ -21,6 +24,155 @@ use std::path::Path;
 //
 // Sierra syscall l2 gas: `n_steps * step_gas_cost + sum(builtin_count * builtin_gas_cost)`
 
+#[test]
```
ddoktorski: Not sure if we need tests for test utils, but if we want to keep them let's move them to a different file

franciszekjob: I would keep them. Moved to `crates/forge/tests/utils/runner.rs`

ddoktorski (approved):

crates/forge/tests/utils/runner.rs
```diff
@@ -30,6 +29,11 @@ use std::{
     str::FromStr,
 };
 
+// Allowed absolute gas differences when `non_exact_gas_assertions` is enabled.
+const MARGIN_L1_GAS: u64 = 10;
+const MARGIN_L1_DATA_GAS: u64 = 10;
+const MARGIN_L2_GAS: u64 = 200_000;
```
DelevoXDG, line 36: This seems a bit large  - from what I understand the margin is carry-over from the past (#3049), but perhaps worth adding an issue and investigation why is it so large?

Seems like a potential path to hidden regressions 🤔 

Perhaps we should eventually (not necessarily in this PR) make it %-relative?

Or, once the skill exists, perhaps we could rely more on the skill to keep the asserted value closer to truth, while having the ability to more clearly review what causes the change?

DelevoXDG (approved):

crates/forge/tests/utils/runner.rs
franciszekjob: Sure thing, added https://github.com/foundry-rs/starknet-foundry/issues/4516 and linked.

MKowalski8 (approved):

PR #4501, opened 2026-07-17
Show syscall/builtin difference on assertion failure

franciszekjob:
Closes #4484

## Before

image

## After

image

## Checklist

- [x] Linked relevant issue
- [ ] Updated relevant documentation
- [x] Added relevant tests
- [x] Performed self-review of the code
- [ ] Added changes to `CHANGELOG.md`

PR #4500, opened 2026-07-17
Gas expectations update skill

franciszekjob:
Closes #4485 

## Introduced changes

Add `update-gas-expectations` skill for updating gas values in forge integration tests. Skill is recommended to use when bumping Scarb/blockifier.

## How to test it

- Change scarb version in order to cause gas tests to fail (due to incorrect gas value expectations)
  
image

- Run gas/resources/available gas tests - some of them should fail
image

- Run `/update-gas-expectations` in claude session and wait once it's finished 

image

- See updated gas values

image

## Checklist

- [x] Linked relevant issue
- [x] Updated relevant documentation
- [ ] Added relevant tests
- [x] Performed self-review of the code
- [ ] Added changes to `CHANGELOG.md`

PR #4513, opened 2026-07-24
Unblock scarb pre-release versions for tests

### LINKED ISSUE #4483: Unable to use Scarb pre-release versions for tests
### Current state

Bumping `.tool-versions` to a Scarb pre-release breaks builds/tests.

For example, that happens in packages with `cairo-version` requirements:

```diff
error: the required Cairo version of package openzeppelin_utils is not compatible with current version
! Cairo version required: ^2.13.1
! Cairo version of Scarb: 2.19.0-rc.3

error: the required Cairo version of each package must match the current Cairo version
help: pass `--ignore-cairo-version` to ignore Cairo version mismatch
```

This blocks Scarb bumps that require the latest available toolchain, including pre-releases. 

(e.g. blockifier bumps **ahead** of a stable Scarb release).

### Objective

Unblock Scarb RC bumps in this repo. 

Determine the right fix and implement it so CI/e2e and local bumps work.

### Additional context

Likely areas worth considering:

- [`matches_version_with_special_rules`](https://github.com/foundry-rs/starknet-foundry/blob/0f1ce2a1e5531b32a73601b4d98f7237bc1d015f/crates/scarb-api/src/lib.rs#L168-L186) drops prerelease for `snforge_std`, we can consider doing the same for `starknet` and other packages 
- `ScarbCommand` can be updated to automatically pass `--ignore-cairo-version` in rc context
- Perhaps Scarb should treat rc-versions as satisfying non-prerelease `cairo-version` requirements?

franciszekjob:
Closes #4483

## Introduced changes

This PR unblocks validation against scarb pre-releases by setting `SCARB_IGNORE_CAIRO_VERSION` in `snforge` / `sncast` test runners and in `scarb` command helper.

## Checklist

- [x] Linked relevant issue
- [ ] Added relevant tests


franciszekjob:

- #4494
    - #4501
        - #4500
            - #4513 ◀

franciszekjob:
note: At https://github.com/foundry-rs/starknet-foundry/pull/4513/commits/cf6053ee8462250c5eec14f04c1c7a113d43a99f scarb version is `2.20.0-rc.2` and some jobs are failing:
- Workspace Tests - `package_matches_version_requirement_test` fails, obviously because scarb `2.20.0-rc.2` != `2.20.0` expected, so this failure is desired
- Forge E2E Tests - snapshot tests fail because they generate versioned snapshots for the active scarb version. With `2.20.0-rc.2`, insta looks for `@2.20.0-rc.2` snapshots, which are not present/accepted. This is not a `cairo-version`  compatibility failure. The command runs successfully and the failure happens at the snapshot assertion stage.

ddoktorski (commented):
Could you give examples of tests that fail without this change?

franciszekjob:
> Could you give examples of tests that fail without this change?

Sure:
- `sncast::main e2e::call::test_happy_case_call_with_option_and_result`
- `e2e::new::create_new_project_dir_not_exist::erc20_contract`
- `e2e::templates::validate_templates::erc20_contract`
- `e2e::new::create_new_project_dir_not_exist::erc20_contract`
- `e2e::oracles::wasm`

You can see them failing in [this CI run](https://github.com/foundry-rs/starknet-foundry/pull/4513/checks), which includes only the changed scarb version to rc (and two new files, but they are not used).

After adding changes back (in https://github.com/foundry-rs/starknet-foundry/pull/4513/commits/cf6053ee8462250c5eec14f04c1c7a113d43a99f), there are no tests which fail with `cairo-version`-related error. Reason of jobs failure is described in comment above.

ddoktorski (commented):
Then it looks like the issue is isolated to the erc20 tests + oracles. I'd prefer not to change the global test setup, as that would make our test environment diverge from what standard users experience. What about modifying the setup only for those specific tests (and adding a code comment to explain why)?

franciszekjob:
> Then it looks like the issue is isolated to the erc20 tests + oracles. I'd prefer not to change the global test setup, as that would make our test environment diverge from what standard users experience. What about modifying the setup only for those specific tests (and adding a code comment to explain why)?

Changed, in https://github.com/foundry-rs/starknet-foundry/pull/4513/commits/996a47db40e4b3157343da2798ddf88f20827688 it's tested with rc scarb. No tests fail because of cairo-version error, they only fail because of gas mismatch.

crates/forge/tests/e2e/templates.rs
```diff
@@ -35,7 +35,16 @@ fn validate_templates(template: &Template) {
     let package_path = Utf8PathBuf::from_path_buf(package_path.to_path_buf())
         .expect("Failed to convert to Utf8PathBuf");
 
-    let scarb_add = ScarbCommand::new()
+    // Cairo version is ignored on purpose. Without it, the test would fail with pre-release scarb.
+    let envs = if matches!(template, Template::Erc20Contract) {
+        &[("SCARB_IGNORE_CAIRO_VERSION", "true")][..]
+    } else {
+        &[]
+    };
```
ddoktorski, line 43: nit: For consistency we can use `vec` as in `validate_init` or change here to align with the `validate_init` code

crates/forge/tests/e2e/templates.rs
```diff
@@ -35,7 +35,16 @@ fn validate_templates(template: &Template) {
...
+
+    let mut scarb_add = ScarbCommand::new();
+    scarb_add.envs(envs.iter().copied());
+    let scarb_add = scarb_add
```
ddoktorski:
```suggestion
    let scarb_add = ScarbCommand::new()
        .envs(envs.iter().copied())
```
Why not like this? The same question to changes in function `check_and_lint_with_envs`

franciszekjob: https://github.com/foundry-rs/starknet-foundry/pull/4513/commits/d03d42c2c5a6b35ad8f7b5aad0992fdd70593839

crates/forge/tests/e2e/templates.rs
franciszekjob: https://github.com/foundry-rs/starknet-foundry/pull/4513/commits/342bed1657481b9b8ab4ff77b8fdce35f7ffd833

ddoktorski (approved):

MKowalski8 (approved):

PR #4520, opened 2026-07-28
Allow `test_detect_devnet_url` to pass with running instances

### LINKED ISSUE #4519: Investigate failing `test_detect_devnet_url`
### Current State

`helpers::devnet::detection::tests::test_detect_devnet_url` is often failing on CI, e.g. here: https://github.com/foundry-rs/starknet-foundry/actions/runs/30361030301/job/90281151691.

Investigate and fix/redesign the test.

### Objective

Investigate failing `test_detect_devnet_url`

### Additional Context

_No response_

franciszekjob:
Closes #4519 

## Introduced changes

- Add deterministic unit coverage for devnet URL detection by separating the core logic from process discovery and reachability checks.
- Widen acceptable outcomes in `test_detect_devnet_url` by allowing the test to pass when `starknet-devnet` is already running on the runner. The test now accepts no instance, a single detected instance, or multiple detected instances as valid outcomes.

## Checklist

- [x] Linked relevant issue
- [ ] Added relevant tests

### DIFF
--- a/crates/sncast/src/helpers/devnet/detection.rs
+++ b/crates/sncast/src/helpers/devnet/detection.rs
@@ -247,13 +247,8 @@ runner 2686 0.0 0.1 100000 20000 ? Sl 13:00 0:00 starknet-devnet --host 127.0.0.
 
     #[tokio::test]
     async fn test_detect_devnet_url() {
-        // Uses the real runner state, so valid outcomes can vary.
-        // Operational failures should still fail the test.
-        let result = detect_devnet_url().await;
-        println!("result: {result:?}");
-        assert!(matches!(
-            result,
-            Ok(_) | Err(DevnetDetectionError::NoInstance | DevnetDetectionError::MultipleInstances)
-        ));
+        // Do not check the result, just ensure that the function runs without panicking
+        // as sanity check.
+        detect_devnet_url().await
     }
 }
--- a/crates/sncast/src/helpers/devnet/detection.rs
+++ b/crates/sncast/src/helpers/devnet/detection.rs
@@ -36,19 +36,17 @@ pub enum DevnetDetectionError {
 }
 
 pub async fn detect_devnet_url() -> Result<Url, DevnetDetectionError> {
-    detect_devnet_from_processes().await
+    detect_devnet_from_processes(find_devnet_process_info(), is_devnet_url_reachable).await
 }
 
 #[must_use]
 pub async fn is_devnet_running() -> bool {
-    detect_devnet_from_processes().await.is_ok()
+    detect_devnet_from_processes(find_devnet_process_info(), is_devnet_url_reachable)
+        .await
+        .is_ok()
 }
 
-async fn detect_devnet_from_processes() -> Result<Url, DevnetDetectionError> {
-    detect_devnet_url_from_process_info(find_devnet_process_info(), is_devnet_url_reachable).await
-}
-
-async fn detect_devnet_url_from_process_info<F, Fut>(
+async fn detect_devnet_from_processes<F, Fut>(
     process_info: Result<ProcessInfo, DevnetDetectionError>,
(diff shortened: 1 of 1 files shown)


crates/sncast/src/helpers/devnet/detection.rs
```diff
@@ -106,12 +121,138 @@ mod tests {
...
+    }
+
+    #[tokio::test]
+    async fn test_detect_devnet_url() {
+        let result = detect_devnet_url().await;
+        assert!(matches!(
+            result,
+            Ok(_) | Err(DevnetDetectionError::NoInstance | DevnetDetectionError::MultipleInstances)
+        ));
+    }
```
franciszekjob, line 253: note: This test is now not ideal since it accepts multiple states of runner, but IMO it's better to keep it so we at least know that it correctly detects number of running instances (0/1/multiple). Proper logic testing is placed in unit tests.

crates/sncast/src/helpers/devnet/detection.rs
```diff
@@ -45,17 +45,28 @@ pub async fn is_devnet_running() -> bool {
 }
 
 async fn detect_devnet_from_processes() -> Result<Url, DevnetDetectionError> {
-    match find_devnet_process_info() {
+    detect_devnet_url_from_process_info(find_devnet_process_info(), is_devnet_url_reachable).await
+}
+
+async fn detect_devnet_url_from_process_info<F, Fut>(
```
MKowalski8: nit: Personally I would drop the wrapper here. `detect_devnet_from_processes` isn't public, so we could update it.

crates/sncast/src/helpers/devnet/detection.rs
ddoktorski: I'm not sure if it makes sense to keep this test as it doesn't check anything concrete now + the logic is covered by newly added unit tests. If we want to keep it maybe it'd be better to treat it as a sanity check that only asserts that some result has been returned but without asserting variants?

ddoktorski (approved):

MKowalski8 (approved):

crates/sncast/src/helpers/devnet/detection.rs
franciszekjob: > it doesn't check anything concrete now

It still checks that the real detection path doesn't fail with an unexpected operational error. In practice, it allows the environment-dependent valid outcomes, but still rejects cases like `ProcessNotReachable` or `CommandFailed`.

> If we want to keep it maybe it'd be better to treat it as a sanity check that only asserts that some result has been returned but without asserting variants?

That would make the test accept those error variants, which doesn't make much sense to me. Keeping the  assertion more narrow gives the test at least some sort of meaningfulness.

ddoktorski: > It still checks that the real detection path doesn't fail with an unexpected operational error.

It's a little difficult to tell that this is the main goal of the test just from looking at the code.

> Keeping the assertion more narrow gives the test at least some sort of meaningfulness.

At the very least, adding a comment to explain this would be helpful.

franciszekjob: Added comment in https://github.com/foundry-rs/starknet-foundry/pull/4520/commits/bd647c19637e26de9708dadfa890d82ac718d147

ddoktorski (approved):

crates/sncast/src/helpers/devnet/detection.rs
franciszekjob: It turns out that this test can fail with more than currently allowed variants. For example in [this run](https://github.com/foundry-rs/starknet-foundry/actions/runs/30451819375/job/90575935835) it failed with `ProcessNotReachable`. I will leave it as a sanity check then, without asserting specific variant.

PR #4534, opened 2026-08-05
Move sncast arg validation before provider resolution

### LINKED ISSUE #4389: Call provider only after validating args
https://github.com/foundry-rs/starknet-foundry/pull/4365#discussion_r3302042130

Currently, in sncast, we attempt to resolve provider at the start of a given command:
```rs
let provider = ...
``` 
before doing anything else, including the validations. 

We should move all of the user-provided args validations:

```rs
```

```rs                
let calldata = arguments
    .try_into_calldata(&abi_from_contract_class(&contract_class)?, &selector)?;
```

before provider resolution - to avoid sending unnecessary calls.

franciszekjob:
Closes #4389

## Introduced changes

self-descriptive

## Checklist

- [x] Linked relevant issue
- [ ] Added relevant tests

### DIFF
--- a/crates/sncast/src/main.rs
+++ b/crates/sncast/src/main.rs
@@ -510,6 +510,7 @@ async fn run_async_command(cli: Cli, config: CastConfig, ui: &UI) -> Result<Exit
             } = deploy;
 
             let provider = rpc.get_provider(&config, ui).await?;
+
             let account = get_account(&config, &provider, &rpc, ui).await?;
 
             let (class_hash, declare_response, local_abi) = if let Some(class_hash) =
@@ -720,6 +721,7 @@ async fn run_async_command(cli: Cli, config: CastConfig, ui: &UI) -> Result<Exit
 
             let selector = get_selector_from_name(&function)
                 .context("Failed to convert entry point selector to FieldElement")?;
+
             let contract_address = contract_address.resolve(&config)?;
             let Arguments {
                 calldata,
--- a/crates/sncast/src/main.rs
+++ b/crates/sncast/src/main.rs
@@ -732,10 +732,10 @@ async fn run_async_command(cli: Cli, config: CastConfig, ui: &UI) -> Result<Exit
 
             let provider = rpc.get_provider(&config, ui).await?;
 
+            let class_hash = get_class_hash_by_address(&provider, contract_address).await?;
             let calldata = if let Some(calldata) = calldata {
                 calldata
             } else {
-                let class_hash = get_class_hash_by_address(&provider, contract_address).await?;
                 let contract_class = get_contract_class(class_hash, &provider).await?;
                 transform(
                     &arguments.unwrap_or_default(),
--- a/crates/sncast/src/main.rs
+++ b/crates/sncast/src/main.rs
@@ -732,6 +732,8 @@ async fn run_async_command(cli: Cli, config: CastConfig, ui: &UI) -> Result<Exit
 
             let provider = rpc.get_provider(&config, ui).await?;
 
+            let account = get_account(&config, &provider, &rpc, ui).await?;
+
             let class_hash = get_class_hash_by_address(&provider, contract_address).await?;
             let calldata = if let Some(calldata) = calldata {
                 calldata
@@ -743,8 +745,6 @@ async fn run_async_command(cli: Cli, config: CastConfig, ui: &UI) -> Result<Exit
                     &selector,
                 )?
             };
-
-            let account = get_account(&config, &provider, &rpc, ui).await?;
 
             let result = with_account!(&account, |account| {
                 starknet_commands::invoke::invoke(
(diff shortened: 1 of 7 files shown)


crates/sncast/tests/e2e/main_tests.rs
franciszekjob, line 1: note: Some tests were adjusted because Scarb.toml was missing in their tempdirs, hence they're we're failing with different error than expected.

crates/sncast/src/main.rs
```diff
@@ -359,6 +362,9 @@ async fn run_async_command(cli: Cli, config: CastConfig, ui: &UI) -> Result<Exit
             )
             .context("Failed to build contract")?;
 
+            let provider = declare.common.rpc.get_provider(&config, ui).await?;
+            let account = get_account(&config, &provider, &declare.common.rpc, ui).await?;
```
ddoktorski: For `declare` command it seems reasonable to keep the previous order due to the scarb build. Currently if there is an invalid account the command will first compile everything only to report at the end that account is incorrect

crates/sncast/src/main.rs
```diff
@@ -703,22 +719,33 @@ async fn run_async_command(cli: Cli, config: CastConfig, ui: &UI) -> Result<Exit
...
+                .map(|raw_calldata| resolve_calldata_to_felts(&raw_calldata, &config))
+                .transpose()?;
+
+            let provider = rpc.get_provider(&config, ui).await?;
+
+            let calldata = if let Some(calldata) = calldata {
+                calldata
+            } else {
+                let class_hash = get_class_hash_by_address(&provider, contract_address).await?;
+                let contract_class = get_contract_class(class_hash, &provider).await?;
```
ddoktorski: Previously it was executed unconditionally so if contract doesn't exist it fails immediately with descriptive error. What is the error now for nonexistent contract and `--calldata`?

crates/sncast/src/main.rs
ddoktorski, line 511: Why the order is not changed for the `deploy` command?

crates/sncast/src/main.rs
```diff
@@ -399,6 +405,7 @@ async fn run_async_command(cli: Cli, config: CastConfig, ui: &UI) -> Result<Exit
                 let contract_definition: SierraClass =
                     serde_json::from_str(&contract_artifacts.sierra)
                         .context("Failed to parse sierra artifact")?;
+                let rpc = declare.common.rpc.clone();
```
ddoktorski: `rpc` is borrowed by `generate_network_flag` function, so does it need to be cloned?

crates/sncast/src/main.rs
franciszekjob: Valid

crates/sncast/src/main.rs
franciszekjob: For `deploy --contract-name`, moving validation before provider/account would also move the scarb build before account validation (same as issue as the one you pointed for declare). For `deploy --class-hash`, we could move some local-only validation earlier, but handling only that subpath would make the deploy flow more branched and complicated.

crates/sncast/src/main.rs
franciszekjob: I see. For nonexistent contract address it gives
```
Error: An error occurred in the called contract = ContractErrorData { revert_error: Message("Requested contract address 0x1 is not deployed") }
```

while with `--calldata` it gives
```
Error: Transaction execution error = TransactionExecutionErrorData { transaction_index: 0, execution_error: Nested(InnerContractExecutionError { contract_address: 0xf6ecd22832b7c3713cfa7826ee309ce96a2769833f093795fafa1b8f20c48b, class_hash: 0x5b4b537eaa2399e3aa99c4e2e0208ebd6c71bc1467938cd52c798c601e43564, selector: 0x15d40a3d6ca2ac30f4031e42be28da9b056fef9bb7357ac5e85627ee876e5ad, error: Nested(InnerContractExecutionError { contract_address: 0x1, class_hash: 0x0, selector: 0x1d7377b4b2053672e38039a02d909f73c4e538c9fddbb7e97aadf700cb9a01a, error: Message("Requested contract address 0x0000000000000000000000000000000000000000000000000000000000000001 is not deployed.\n") }) }) }
```

I moved the line with class hash resolution, it's unconditional as it was before. Also added `test_contract_does_not_exist_with_calldata` for such case.

crates/sncast/tests/e2e/main_tests.rs
ddoktorski: Could you explain why do we need these changes? It seems that running these tests on the previous version makes no difference

franciszekjob: Not relevant anymore, as I changed the order of account validation after https://github.com/foundry-rs/starknet-foundry/pull/4534/commits/ce5cadb3aa1b14fc7fd2786d05d655f60266b999. Restored as they were before.

ddoktorski (approved):

integraledelebesgue (approved):

PR #4539, opened 2026-08-06
Use raw Sierra function gas cost as initial budget in `forge_runner`

### LINKED ISSUE #4493: "Unexpected remaining gas" error in `snforge test` (`skip-optimizations = true`)
### Current State
`snforge test` accepts `[cairo] skip-optimizations = true`. 

However, a simple `Felt252Dict` test fails at runtime with a non-actionable blockifier error:

```diff
+ [PASS] dicttest_integrationtest::test_dict::test_dict_variables (l2_gas: ~34800)
- [FAIL] dicttest_integrationtest::test_dict::test_dict_variables
-
- Failure data:Malformed return data : Unexpected remaining gas: 9223372036854786057..
```

Test in question:
```cairo
#[test]
fn test_dict_variables() {
    let mut balances: Felt252Dict<felt252> = Default::default();

    balances.insert(0, 0);
    balances.insert(0, 0);
    balances.insert(0, 0);
    balances.insert(0, 0);
    balances.insert(0, 0);
    balances.insert(0, 0);
}
```

#### Check that gets hit

The number in the error appears to be `9223372036854775807 + 10250`, i.e. `i64::MAX + 10250`. Which probably means reported gas consumption ends up negative 🤔 

Snforge sets `initial_gas = i64::MAX` in [`build_test_entry_point`](https://github.com/foundry-rs/starknet-foundry/blob/16a109adaca9412ec1927f91e03d10837f735b3b/crates/cheatnet/src/constants.rs#L95). So, `remaining_gas > initial_gas` check gets tripped in blockifier:
https://github.com/software-mansion-labs/sequencer/blob/c2e2dff39f7105ef3adb622e7e38cf9a1ebbd996/crates/blockifier/src/execution/entry_point_execution.rs#L519-L521

#### Root cause

1. `skip-optimizations = true` maps to cairo-lang `Optimizations::Disabled`. 
2. That switches `final_optimization_strategy`: 
    ```diff
    - [GasRedeposit, EarlyUnsafePanic, TrimUnreachable, LowerImplicits, ReorganizeBlocks]
    + [LowerImplicits, ReorganizeBlocks]
    ```
    Thus, `GasRedeposit` gets dropped. 
3. The emitted Sierra then contains no `redeposit_gas` libfunc

### Action points

- Confirm `redeposit_gas` is indeed behind the error (it most likely is)

   | Config | `redeposit_gas` in Sierra | Result |
   |--------|---------------------------|--------|
   | `skip-optimizations = true` | **0** | FAIL (`Unexpected remaining gas: i64::MAX + 10250`) |
   | default / `skip-optimizations = false` | **10** | PASS (`l2_gas: ~34800`) |
   | `inlining-strategy = "avoid"` only (no `skip-optimizations`) | **≥1** | PASS (`l2_gas: ~59870`) |

- ~~Investigate **why** the gas consumption ends up negative~~ 
   `felt252_dict_squash` **adds** ("refunds") gas at runtime for repeated keys:
   https://github.com/starkware-libs/cairo/blob/9de65e878c4e24b9cfee1efe283c6d62485cd30a/crates/cairo-lang-sierra-to-casm/src/invocations/felt252_dict.rs#L99-L100

   ```rs
   refund_per_repeated_access = UNIQUE.cost() - REPEATED.cost()
     UNIQUE   = {steps:46, range_checks:6} → 5020
     REPEATED = {steps:9,  range_checks:1} → 970
     refund   = 4050
   ```

- Determine: Why is above true? Why is refund higher than the gas use?

#### skip-optimizations

Decide how `snforge test` should handle `skip-optimizations = true`:

1. Fix it on `scarb` / `cairo-lang` side
    Arguably, `GasRedeposit` is a gas-correctness pass, not an optimization. So perhaps gating it behind `skip-optimizations` is something we can consider as an actual bug?
~~2. Override `skip-optimizations` when building test artifacts~~
~~3. Hard-reject `skip-optimizations` (mirror the `enable-gas = false` check from #4335).~~
    
We should really consider 2. and 3. only if we deem `skip-optimizations` to be incompatible with `snforge test` for some reason.

#### Negative gas

Research what else contributes to resulting gas being negative.
Perhaps addressing this is the key to solving this bug.

integraledelebesgue:
Closes: #4493 
Requires: https://github.com/software-mansion/universal-sierra-compiler/pull/133

Blockifier charges a fixed amount of gas as an "entry-point cost", equal to 10 000 units and sets the gas budget to `u64::MAX - 10_000`. That initial entry-point cost limits the amount of gas that can be refunded after execution. The cap of 10 000 units is correct for contracts **but not for raw Sierra functions**, which we execute as entry-points in Forge and causes the final gas balance to overflow.

Overriding the entry-point cost with a function gas cost inferred by the Sierra → CASM compiler ensures the final balance will not overflow.

> [!important]
> Changes in snapshots stem from the change in gas accounting logic and were verified for correctness.


crates/forge-runner/src/running/setup.rs
```diff
@@ -85,13 +89,55 @@ fn builtins_from_program(program: &Program) -> Vec<BuiltinName> {
...
-        .base
-        .context
-        .gas_costs()
-        .base
-        .entry_point_initial_budget
+/// Returns the compiler-inferred initial gas budget for a raw Sierra function.
+pub fn raw_program_initial_budget(
+    test_details: &TestDetails,
+    casm_program: &RawCasmProgram,
+    syscall_hint_processor: &SyscallHintProcessor,
```
ddoktorski: nit: Given that we need only builtin costs, can we pass `BuiltinGasCosts` instead of `SyscallHintProcessor`?

crates/forge-runner/src/running/setup.rs
```diff
@@ -85,13 +89,55 @@ fn builtins_from_program(program: &Program) -> Vec<BuiltinName> {
...
+    builtin_gas_costs: &BuiltinGasCosts,
+) -> anyhow::Result<u64> {
+    function_cost
+        .costs
+        .iter()
+        .try_fold(0_u64, |total, (token_type, count)| {
+            let count = u64::try_from(*count)
+                .with_context(|| format!("Negative raw Sierra function cost for {token_type:?}"))?;
+            let token_gas_cost = match token_type {
+                RawCostTokenType::Const => 1,
```
ddoktorski: Why is it 1?

crates/universal-sierra-compiler-api/src/representation.rs
```diff
@@ -16,6 +17,28 @@ pub struct RawCasmProgram {
     ///
     /// Those 2 values are usually not equal since the instruction sizes in CASM may vary
     pub debug_info: Vec<(CasmCodeOffset, CasmInstructionIdx)>,
+    /// Compiler-inferred costs of raw Sierra functions, keyed by their Sierra entry-point
+    /// statement index.
+    pub function_costs: Vec<RawFunctionCost>,
```
ddoktorski: Actually why is it a vector and not `HashMap<usize, BTreeMap<RawCostTokenType, i64>>`?

crates/forge-runner/src/running/setup.rs
```diff
@@ -85,13 +89,55 @@ fn builtins_from_program(program: &Program) -> Vec<BuiltinName> {
...
+                RawCostTokenType::MulMod => builtin_gas_costs.mul_mod,
+                RawCostTokenType::Blake => builtin_gas_costs.blake,
+            };
+            let token_cost = count
+                .checked_mul(token_gas_cost)
+                .ok_or_else(|| anyhow!("Raw Sierra function cost overflow for {token_type:?}"))?;
+
+            total
+                .checked_add(token_cost)
+                .ok_or_else(|| anyhow!("Raw Sierra function cost overflow"))
```
ddoktorski:
```suggestion
                .ok_or_else(|| anyhow!("Raw Sierra function cost overflow for {token_type:?}"))
```
As above

crates/forge-runner/src/running/setup.rs
```diff
@@ -85,13 +89,55 @@ fn builtins_from_program(program: &Program) -> Vec<BuiltinName> {
...
 
-pub fn entry_point_initial_budget(syscall_hint_processor: &SyscallHintProcessor) -> u64 {
-    syscall_hint_processor
-        .base
-        .context
-        .gas_costs()
-        .base
-        .entry_point_initial_budget
+/// Returns the compiler-inferred initial gas budget for a raw Sierra function.
+pub fn raw_program_initial_budget(
```
ddoktorski: nit:
```suggestion
pub fn raw_function_initial_budget(
```

crates/forge/tests/e2e/snapshots/gas_report/main__e2e__gas_report__snap_recursive_calls@2.18.0.snap
```diff
@@ -2,7 +2,10 @@
 source: crates/forge/tests/e2e/gas_report.rs
 expression: stdout
 ---
-[PASS] debugging_integrationtest::test_trace::test_debugging_trace_success (l1_gas: ~0, l1_data_gas: ~288, l2_gas: ~1516050)
+
+
+
```
ddoktorski: What is the reason of the newlines?

ddoktorski (commented):
* We can also mention this in the Changelog under `Fixed`
* There is [MINIMAL_USC_VERSION](https://github.com/foundry-rs/starknet-foundry/blob/fix/use-raw-sierra-function-costs/crates/forge/src/lib.rs#L44) used when checking requirements, it should be updated

crates/forge/tests/e2e/snapshots/gas_report/main__e2e__gas_report__snap_recursive_calls@2.18.0.snap
integraledelebesgue: They were a noise introduced when updating the gas values

crates/forge-runner/src/running/setup.rs
integraledelebesgue: Done

crates/forge-runner/src/running/setup.rs
integraledelebesgue: Done

crates/universal-sierra-compiler-api/src/representation.rs
integraledelebesgue: Good call. I replaced the vector with a hashmap in both places

crates/forge-runner/src/running/setup.rs
integraledelebesgue: Done

crates/forge-runner/src/running/setup.rs
integraledelebesgue: The `count` value received from the compiler is already expressed **in Sierra gas units** for operations that count as `Const`, so we do not need to multiply it by the price

integraledelebesgue: I replaced the raw `1` with a documented const

integraledelebesgue:
> * We can also mention this in the Changelog under `Fixed`
> * There is [MINIMAL_USC_VERSION](https://github.com/foundry-rs/starknet-foundry/blob/fix/use-raw-sierra-function-costs/crates/forge/src/lib.rs#L44) used when checking requirements, it should be updated

Both things are done ✅

CHANGELOG.md
```diff
@@ -32,14 +24,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 #### Fixed
 
 - Malformed byte array panic data (e.g. an untrusted words length) no longer causes a crash when matching `#[should_panic]` expected data. Additionally, invalid byte array serializations (e.g. bad `pending_word_len`) are now rejected instead of rendered as a corrupted string.
+- Raw Sierra tests now use compiler-inferred function costs as their initial gas budget, allowing correct gas refunds beyond the contract entry-point precharge.
```
ddoktorski: Wrong place 👀
Please fix after rebasing

ddoktorski (commented):
LGTM 👌 

Please re-request once the relevant change is released on the USC side and this PR is updated

franciszekjob (approved):

ddoktorski (approved):

Issue #3105, opened 2025-03-17
Allow to set all flags in `snfoundry.toml`


franciszekjob:
### Which component is your feature related to?

Cast

### Feature Request

In the scope of this task we should implement possibility of setting resource bounds values and other missing flags in `snfoundry.toml`. This doesn't apply to `--max-fee` flag, it should be only available to set from command level.

Example usage:

```
[sncast.default]
l1_gas = ...
l1_gas_unit_price = ...
l2_gas = ...
l2_gas_unit_price = ...
l1_data_gas = ...
l1_data_gas_unit_price = ...
```

PR #4544, opened 2026-08-17
Allow setting fee params in `snfoundry.toml`

ddoktorski:
Towards #3105

### DIFF
--- a/crates/sncast/src/helpers/fee.rs
+++ b/crates/sncast/src/helpers/fee.rs
@@ -1,10 +1,15 @@
 use super::felt::felt_from_string;
 use anyhow::{Result, ensure};
 use clap::Args;
+use configuration::Override;
+use serde::{Deserialize, Serialize};
+use serde_json::Value;
+use serde_with::skip_serializing_none;
 use starknet_rust::core::types::FeeEstimate;
 use starknet_types_core::felt::{Felt, NonZeroFelt};
+use std::collections::HashMap;
 
-#[derive(Args, Debug, Clone, Copy)]
+#[derive(Args, Debug, Clone, Copy, Default)]
 #[group(id = "fee_args", multiple = true)]
 pub struct FeeArgs {
     /// Max fee for the transaction. If not provided, will be automatically estimated.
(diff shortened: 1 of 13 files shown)


ddoktorski:

- #4544 ◀
    - #4545

crates/sncast/src/main.rs
MKowalski8, line 780: Doesn't multicall need `fee-params`? Like `Deploy`, `Invoke` etc.? Or is it used somehow later?

MKowalski8, line 783: Same as above. Should we pass `fee-params` to account?

crates/sncast/src/helpers/fee.rs
```diff
@@ -45,6 +50,32 @@ pub struct FeeArgs {
...
+            resolved.l1_gas_price = self.l1_gas_price.or(config.l1_gas_price);
+            resolved.l2_gas = self.l2_gas.or(config.l2_gas);
+            resolved.l2_gas_price = self.l2_gas_price.or(config.l2_gas_price);
+            resolved.l1_data_gas = self.l1_data_gas.or(config.l1_data_gas);
+            resolved.l1_data_gas_price = self.l1_data_gas_price.or(config.l1_data_gas_price);
+        }
+
+        if !self.defines_tip() {
+            resolved.tip = config.tip;
+            resolved.estimate_tip = config.estimate_tip;
```
MKowalski8, line 68: Can we set bigger `tip` than `max_fee`? My concern is, should we somehow guard it so that we don't exceed `max_fee`?

crates/sncast/src/main.rs
ddoktorski: It's fixed in https://github.com/foundry-rs/starknet-foundry/pull/4545

crates/sncast/src/helpers/fee.rs
ddoktorski: I'll check this, but if such a problem exists it's not introduced by this change

crates/sncast/src/helpers/fee.rs
```diff
@@ -72,6 +103,63 @@ impl FeeArgs {
...
+
+    /// Whether this layer defines the tip group at all.
+    fn defines_tip(&self) -> bool {
+        self.tip.is_some() || self.estimate_tip
+    }
+}
+
+impl Override for FeeParams {
+    fn override_with(&self, other: FeeParams) -> FeeParams {
+        let tip = if other.defines_tip() { &other } else { self };
```
integraledelebesgue, line 147: This way, we override the tip only if `other` has it enabled and cannot **disable** it via override. Is it expected?

ddoktorski: We can disable it by explicit `tip = 0`, but it's probably not very intuitive. Actually I'd leave an option to set only `tip` in the `snfoundry.toml` to simplify it. Then it's clear what and how is overridden, `estimate-fee` will be only cli option similarly as `max-fee`. Wdyt?

integraledelebesgue (approved):

MKowalski8 (approved):

Issue #3557, opened 2025-07-17
Add utils to get transaction trace


ddoktorski:
### Select Tool

sncast

### Requested Feature

We can add a command to call `starknet_traceTransaction`, which would return the output in a human readable format by decoding entry points and leveraging data transformer. This is inspired by our debugging crate that prints test trace in a format that’s easy to read and debug.

PR #4546, opened 2026-08-18
Add `sncast get tx-trace`

franciszekjob:
Towards #3557

## Introduced changes

Add `sncast get tx-trace` command

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests


crates/foundry-ui/src/styling.rs
```diff
@@ -163,6 +165,20 @@ impl OutputBuilder {
...
+                        } else {
+                            writeln!(
+                                content,
+                                "{}{}",
+                                " ".repeat(field_width + 1),
+                                style(line).yellow()
+                            )
+                            .unwrap();
+                        }
+                    }
```
franciszekjob, line 181: note: This change is needed to align continuation lines with the field value, keeping multiline values (such as revert reason) readable.

franciszekjob:
note: Currently, both human-readable and JSON outputs decode selected trace fields. I think it would be useful to introduce a `--raw flag` that preserves the original, non-decoded values. WDYT?

Cargo.toml
```diff
@@ -76,7 +76,8 @@ rayon = "1.10"
 regex = "1.11.3"
 serde = { version = "1.0.219", features = ["derive"] }
 serde_json = "1.0.149"
-starknet-rust = "0.19.0"
+# TODO: Restore version once starknet-rust releases new version
+starknet-rust = { git = "https://github.com/software-mansion/starknet-rust", rev = "56efca00feffac6beb39f464eca71afa42c1813b" }
```
franciszekjob: note: Temporary, due to change in the content type in `NoTraceAvailable` error.

docs/src/starknet/multicall.md
```diff
@@ -154,18 +154,17 @@ Success: Multicall template created successfully
...
+         class_hash = ""
+         inputs = []
+         id = ""
+         unique = false
+
+         [[call]]
+         call_type = "invoke"
+         contract_address = ""
+         function = ""
+         inputs = []
```
franciszekjob, line 167: note: Change related to https://github.com/foundry-rs/starknet-foundry/pull/4546/changes#r3820707168

crates/sncast/tests/e2e/get/tx_receipt.rs
```diff
@@ -54,18 +54,17 @@ async fn test_get_tx_receipt_reverted() {
...
+                              Got an exception while executing a hint: Execution failed. Failure reason: 0x7374616c65207265706f7274 ('stale report').
+                              Cairo traceback (most recent call last):
+                              Unknown location (pc=0:65)
+                              Unknown location (pc=0:1787)
+                              Unknown location (pc=0:2271)
+                              Unknown location (pc=0:2984)
+                              Unknown location (pc=0:3611)
+
+                              Error in the called contract (0x00132303a40ae2f271f4e1b707596a63f6f2921c4d400b38822548ed1bb0cbe0):
+                              Execution failed. Failure reason: 0x7374616c65207265706f7274 ('stale report').
```
franciszekjob, line 67: note: Changed related to https://github.com/foundry-rs/starknet-foundry/pull/4546/changes#r3820707168

crates/sncast/src/starknet_commands/get/tx_trace.rs
```diff
@@ -0,0 +1,129 @@
...
+};
+use sncast::response::ui::UI;
+use starknet_rust::core::types::{BlockId, BlockTag};
+use starknet_rust::providers::jsonrpc::{HttpTransport, JsonRpcClient};
+use starknet_rust::providers::{Provider, ProviderError};
+use starknet_types_core::felt::Felt;
+use std::collections::{HashMap, HashSet};
+use std::process::ExitCode;
+
+const MAX_CONCURRENT_CLASS_REQUESTS: usize = 4;
```
franciszekjob, line 23: note: 4 is an arbitrary value for concurrent requests, but it seems it should be fairly safe, even for public nodes. For example, quicknode documents a limit of 15 requests per second.
https://www.quicknode.com/pricing

Image

crates/sncast/src/response/get/tx_trace.rs
```diff
@@ -0,0 +1,878 @@
...
+    if full {
+        let builder = append_constructor(builder);
+        let builder = append_execution_resources(builder, &r.execution_resources, 0);
+        let builder = append_fee_transfer(builder);
+        let builder = append_optional_state_diff(builder, r.state_diff.as_ref(), 0);
+        append_validate(builder)
+    } else {
+        let builder = append_validate(builder);
+        let builder = append_constructor(builder);
+        append_fee_transfer(builder)
```
integraledelebesgue, line 506: The short version has a different order of components than the full version - `append_validate` is first in the former and last in the latter.

crates/sncast/src/response/get/tx_trace.rs
```diff
@@ -0,0 +1,878 @@
...
+        append_validate(builder)
+    } else {
+        let builder = append_validate(builder);
+        append_fee_transfer(builder)
+    }
+}
+
+fn append_deploy_account(
+    builder: OutputBuilder,
+    r: &DeployAccountTransactionTrace,
```
integraledelebesgue:
```suggestion
    trace: &DeployAccountTransactionTrace,
```
Here and in other places as well

crates/sncast/src/starknet_commands/get/tx_trace.rs
```diff
@@ -0,0 +1,129 @@
...
+
+#[derive(Debug, Args)]
+#[command(about = "Get the execution trace of a transaction")]
+pub struct TxTrace {
+    /// Hash of the transaction
+    pub transaction_hash: Felt,
+
+    /// Display all transaction trace fields
+    #[arg(long)]
+    pub full: bool,
```
integraledelebesgue, line 43: `--full` has no effect when used with `--json`. Maybe they should be mutually-exclusive?

crates/sncast/src/starknet_commands/get/tx_trace.rs
```diff
@@ -0,0 +1,129 @@
...
+            let decoder = TraceDecoder::new(classes);
+            Ok(TransactionTraceResponse::new(trace, decoder, tx_trace.full))
+        }
+        Err(error) => Err(error),
+    };
+
+    Ok(process_command_result("get tx-trace", result, ui, None))
+}
+
+async fn fetch_contract_classes(
```
integraledelebesgue, line 89: (off-top) Fetching classes can be expensive. Maybe we should introduce a cache for them in a separate task?

crates/sncast/src/response/errors.rs
```diff
@@ -162,6 +164,8 @@ pub enum SNCastStarknetError {
     UnsupportedTxVersion,
     #[error("Unsupported contract class version")]
     UnsupportedContractClassVersion,
+    #[error("No trace is available for the transaction (status: {:?})", .0.status)]
```
integraledelebesgue, line 167: Is debug display intentional here?

crates/sncast/src/response/get/tx_trace.rs
```diff
@@ -0,0 +1,878 @@
...
+    indent: usize,
+) -> OutputBuilder {
+    if let Some(diff) = state_diff {
+        append_state_diff(builder, diff, indent)
+    } else {
+        builder
+    }
+}
+
+fn append_state_diff(
```
integraledelebesgue, line 728: Tests cover only State Diff Length. The diff itself is not tested

crates/sncast/src/response/get/tx_trace.rs
```diff
@@ -0,0 +1,878 @@
...
+            decoder,
+            full,
+        } = self;
+        {
+            let human_text = append_trace(OutputBuilder::new(), trace, decoder, *full).build();
+            let builder = if decoder.abi_decoding_incomplete() {
+                OutputBuilder::new()
+                    .text_field(
+                        &WarningMessage::new(
+                            "Some trace data could not be decoded with the fetched ABIs. Raw felts are shown instead.",
```
integraledelebesgue: This warning is very vague. It does not explain *why* the data was not decoded (missing or malformed ABI, Cairo 0 contract, etc.). Maybe we should introduce dedicated errors that clearly indicate the reasons?

Also, if I understand correctly, if this warning is `.text()`-ed, it does not appear in JSON mode? We get a non-decoded output without a warning then.

crates/data-transformer/src/shared/extraction.rs
```diff
@@ -28,7 +29,7 @@ fn default_constructor() -> AbiFunction {
 
 fn search_for_function(abi: &[AbiEntry], searched_selector: Felt) -> Option<AbiFunction> {
     abi.iter().find_map(|entry| match entry {
-        AbiEntry::Function(func) => {
+        AbiEntry::Function(func) | AbiEntry::L1Handler(func) => {
```
integraledelebesgue, line 32: Including `L1Handler` variant here leads to ambiguity: if a contract exposes a function and an L1 handler **with the same name** (it's possible), one can be decoded with the signature of the other.
Also, this way a user can try to invoke the L1 handler using `sncast invoke`.
Idk if ABI's standard defines the order in which "normal" functions and L1 handlers appear. Nevertheless, maybe we should make the L1 handler lookup opt-in or even implemented in a separate method?

crates/sncast/src/response/get/tx_trace.rs
```diff
@@ -0,0 +1,878 @@
...
+            } else {
+                OutputBuilder::new()
+            };
+
+            builder
+                .success_message("Transaction trace retrieved")
+                .blank_line()
+                .text_field(&human_text)
+                .build()
+        }
```
integraledelebesgue: nit: this code doesn't need to be wrapped with a `{ ... }` block

crates/sncast/src/starknet_commands/get/tx_trace.rs
franciszekjob: https://github.com/foundry-rs/starknet-foundry/pull/4546/commits/47130c89dcba6e2a9cf9e312b58edd4116a27be4

crates/sncast/src/starknet_commands/get/tx_trace.rs
franciszekjob: Good idea. Added https://github.com/foundry-rs/starknet-foundry/issues/4571 for it.

crates/sncast/src/response/errors.rs
franciszekjob: Yes, it's fine as debug is just `Received` / `Rejected`.

crates/data-transformer/src/shared/extraction.rs
franciszekjob: Good catch! Introduced #4572 to address this.

crates/sncast/src/response/get/tx_trace.rs
```diff
@@ -0,0 +1,876 @@
...
+    decoder: &TraceDecoder,
+) {
+    if let (ExecuteInvocation::Success(invocation), Some(json)) = (invocation, json) {
+        decode_invocation_json(invocation, json, decoder);
+    }
+}
+
+fn decode_invocation_json(
+    invocation: &FunctionInvocation,
+    json: &mut Value,
```
ddoktorski, line 149: Why `decode_execute_invocation_json` takes `Option` and `decode_invocation_json` not?

crates/sncast/src/response/get/tx_trace.rs
```diff
@@ -0,0 +1,876 @@
...
+                decoder,
+            );
+        }
+        TransactionTrace::DeployAccount(trace) => {
+            decode_optional_invocation_json(
+                trace.validate_invocation.as_ref(),
+                json.get_mut("validate_invocation"),
+                decoder,
+            );
+            if let Some(json) = json.get_mut("constructor_invocation") {
```
ddoktorski: Shouldn't we rather throw an error if required field is missing? 🤔

crates/sncast/src/response/get/tx_trace.rs
```diff
@@ -0,0 +1,876 @@
...
+        }
+    }
+}
+
+impl Serialize for TransactionTraceResponse {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        let mut json = serde_json::to_value(&self.trace).map_err(S::Error::custom)?;
```
ddoktorski, line 47: What is the rationale that it is first parsed to json instead of direct serialization of the struct like for other commands e.g. `get tx-receipt`?

crates/sncast/src/response/get/tx_trace.rs
```diff
@@ -0,0 +1,876 @@
...
+) -> OutputBuilder {
+    let builder = builder
+        .with_indent(indent)
+        .field("Call Type", format_call_type(invocation.call_type))
+        .field("Calldata", &decoder.calldata(invocation))
+        .felt_field("Caller Address", &invocation.caller_address);
+    let builder = append_calls(builder, &invocation.calls, decoder, indent);
+    let builder = builder
+        .with_indent(indent)
+        .felt_field("Class Hash", &invocation.class_hash)
```
ddoktorski: Let's use `padded_felt_field` for class hashes and contract addresses

crates/sncast/src/response/get/tx_trace.rs
```diff
@@ -0,0 +1,876 @@
...
+
+    if let Some(json_calls) = json.get_mut("calls").and_then(Value::as_array_mut) {
+        for (call, json_call) in invocation.calls.iter().zip(json_calls) {
+            decode_invocation_json(call, json_call, decoder);
+        }
+    }
+}
+
+#[must_use]
+pub fn contract_addresses_by_class_hash(
```
ddoktorski: It seems that this and related code should be placed in `starknet_commands/get/tx_trace.rs` not in `reponse` dir

crates/sncast/src/response/get/tx_trace.rs
```diff
@@ -0,0 +1,876 @@
...
+    transaction_trace: &TransactionTrace,
+) -> HashMap<Felt, HashSet<Felt>> {
+    let mut contract_addresses_by_class_hash = HashMap::new();
+    for invocation in root_invocations(transaction_trace) {
+        collect_contract_addresses(invocation, &mut contract_addresses_by_class_hash);
```
ddoktorski: Do we need to collect contract addresses at all? Is it used for anything other than error messages?

crates/sncast/src/starknet_commands/get/tx_trace.rs
```diff
@@ -0,0 +1,129 @@
...
+        .map_err(|error| StarknetCommandError::ProviderError(error.into()))
+        .map_err(handle_starknet_command_error);
+
+    let result = match result {
+        Ok(trace) => {
+            let contract_classes_fetch_response =
+                fetch_contract_classes(&provider, contract_addresses_by_class_hash(&trace)).await;
+
+            let ContractClassesFetchResponse { classes, failures } =
+                contract_classes_fetch_response;
```
ddoktorski:
```suggestion
            let ContractClassesFetchResponse { classes, failures } =
                fetch_contract_classes(&provider, contract_addresses_by_class_hash(&trace)).await;
```

crates/sncast/src/response/get/tx_trace.rs
```diff
@@ -0,0 +1,876 @@
...
+    L1HandlerTransactionTrace, LegacyContractAbiEntry, MigratedCompiledClassItem, NonceUpdate,
+    OrderedEvent, OrderedMessage, ReplacedClassItem, StateDiff, TransactionTrace,
+};
+use starknet_rust::core::utils::get_selector_from_name;
+use starknet_rust::providers::ProviderError;
+use starknet_types_core::felt::Felt;
+use std::cell::Cell;
+use std::collections::{HashMap, HashSet};
+
+pub struct ContractClassesFetchResponse {
```
ddoktorski:
```suggestion
pub struct FetchedContractClasses {
```

crates/sncast/src/starknet_commands/get/tx_trace.rs
```diff
@@ -0,0 +1,129 @@
...
+        "Could not fetch contract classes needed to decode the trace:\n{details}\nAffected calls are displayed as raw felts."
+    )
+}
+
+fn provider_error_message(error: &ProviderError) -> String {
+    match error {
+        ProviderError::StarknetError(error) => error.message().to_string(),
+        error => error.to_string(),
+    }
+}
```
ddoktorski: Does it make sense to fix it on the `starknet-rust` side?

crates/sncast/src/response/get/tx_trace.rs
```diff
@@ -0,0 +1,876 @@
...
+        builder = append_section(builder, "Calls", indent);
+        for nested_call in &invocation.calls {
+            builder = append_compact_invocation(builder, nested_call, decoder, indent + 2);
+        }
+    }
+
+    builder
+}
+
+fn append_full_invocation(
```
ddoktorski, line 605: The current display seems to be in this order:
  1. root Call Type, Calldata, Caller Address
  2. root Calls
  3. nested call details
  4. root Class Hash, Contract Address, Entry Point Selector etc.

which is very hard to read, please fix it so all root data go first

docs/src/appendix/sncast/get/tx-trace.md
```diff
@@ -0,0 +1,37 @@
...
+
+Use predefined network with public provider
+
+Possible values: `mainnet`, `sepolia`, `devnet`.
+
+Overrides network from `snfoundry.toml`.
+
+## `--full`
+Optional.
+Conflicts with: [`--json`](../common.md#--json--j)
```
ddoktorski, line 31: If `--json` is passed directly after `sncast` there is no conflict and `--full` is ignored

ddoktorski (changes requested):

crates/sncast/src/response/get/tx_trace.rs
franciszekjob: After some though, I simplified it, so warning shows only class hash now. https://github.com/foundry-rs/starknet-foundry/pull/4546/commits/8c9663e6e935daaf0ff1ea941ce0e81cf4f70d52

crates/sncast/src/response/get/tx_trace.rs
franciszekjob: Changed, https://github.com/foundry-rs/starknet-foundry/pull/4546/commits/54735d203cc0b66a4e8bf6ea4cc7084b418c131a

crates/sncast/src/response/get/tx_trace.rs
franciszekjob: https://github.com/foundry-rs/starknet-foundry/pull/4546/commits/9d7c6e8462d06afaa648b984410cc232626db986

crates/sncast/src/response/get/tx_trace.rs
franciszekjob: Good catch, https://github.com/foundry-rs/starknet-foundry/pull/4546/commits/e82b05b6303450750f0db292a7426b34b4ee2e39

crates/sncast/src/response/get/tx_trace.rs
franciszekjob: https://github.com/foundry-rs/starknet-foundry/pull/4546/commits/85e9fee4c77a2bee97ac5a28903c7f824af694c5

crates/sncast/src/response/get/tx_trace.rs
franciszekjob: https://github.com/foundry-rs/starknet-foundry/pull/4546/commits/a5d6db7b9827e604404ddf11976cc198a5e986f6

crates/sncast/src/response/get/tx_trace.rs
franciszekjob: Direct serialization would return raw felts. We convert to JSON first to preserve the original trace schema while replacing selectors, calldata etc. with decoded values. Mentioned `tx-receipt` (and others) do not require such transformations.

docs/src/appendix/sncast/get/tx-trace.md
franciszekjob: Yeah, right. In https://github.com/foundry-rs/starknet-foundry/pull/4567 I made `--json` flag global, so the conflict could be defined at clap level. However, clap does not detect it when `--json` appears before the subcommand. I think the cleanest solution is to validate the `--json` / `--full` conflict at runtime remove the now-unnecessary change from mentioned PR.

Runtime validation -> https://github.com/foundry-rs/starknet-foundry/pull/4546/commits/7758ae07250688a47668aa6c1ed75ab08f005c38

crates/sncast/src/response/get/tx_trace.rs
franciszekjob: Good point. I replaced the flag with structured decoding warnings that include the failure reason and class hash. In JSON mode, they're exposed in `decoding_warnings` field. https://github.com/foundry-rs/starknet-foundry/pull/4546/commits/1986bbff4ca20df8f418407b4e67c459950bbee4

crates/sncast/src/response/get/tx_trace.rs
franciszekjob: The `state_diff` field is optional in the [RPC specification](https://github.com/starkware-libs/starknet-specs/blob/1e79af61071a77e0031d397e1fbe81e9a0637072/api/starknet_trace_api_openrpc.json#L268) and is not returned by the node for the transactions used in these tests, hence it's omitted from the output in tests. I added a separate deterministic test with a mocked state diff that verifies all of its formatted sections.

crates/sncast/src/starknet_commands/get/tx_trace.rs
franciszekjob: Yeah it does, changed in https://github.com/software-mansion/starknet-rust/pull/160

integraledelebesgue (approved):

docs/src/starknet/tx-status.md
```diff
@@ -30,3 +30,141 @@ Execution Status: Succeeded
...
+<details>
+<summary>Output:</summary>
+
+```shell
+Success: Transaction trace retrieved
+
+Type:                     INVOKE
+Validate Invocation
+  Entry Point Selector:   __validate__
+  Contract Address:       0x[..]
```
ddoktorski, line 54: Why not to show the exact values here?

crates/sncast/src/starknet_commands/get/tx_trace.rs
```diff
@@ -0,0 +1,174 @@
...
+        .map_err(|error| StarknetCommandError::ProviderError(error.into()))
+        .map_err(handle_starknet_command_error);
+
+    let result = match result {
+        Ok(trace) => {
+            let FetchedContractClasses { classes, failures } =
+                fetch_contract_classes(&provider, class_hashes(&trace)).await;
+
+            if !failures.is_empty() {
+                ui.print_warning(WarningMessage::new(format_class_fetch_warning(&failures)));
```
ddoktorski, line 76: Would it be possible to make it part of `decoding_warnings` so all decoding errors are gathered in one place? It should work better for json output then

crates/sncast/src/response/get/tx_trace.rs
```diff
@@ -0,0 +1,903 @@
...
+                }
+            }
+        }
+
+        decoder
+    }
+
+    fn selector(&self, invocation: &FunctionInvocation) -> String {
+        if let Some(abi) = self.sierra_abis.get(&invocation.class_hash.into_())
+            && let Some(function) =
```
ddoktorski, line 276: `invocation.class_hash.into_()` is evaluated a few times in this function, let's just add
```rust
let class_hash: ClassHash = invocation.class_hash.into_();
```
also in the functions below

crates/sncast/src/response/get/tx_trace.rs
```diff
@@ -0,0 +1,903 @@
...
+
+    let details = warnings
+        .iter()
+        .map(|issue| format!("- {}", issue.description()))
+        .join("\n");
+
+    format!("Some trace data is shown as raw felts:\n{details}")
+}
+
+#[must_use]
```
ddoktorski, line 360:
```suggestion
```

crates/sncast/src/response/get/tx_trace.rs
```diff
@@ -0,0 +1,903 @@
...
+        let TransactionTraceResponse {
+            trace,
+            decoder,
+            full,
+        } = self;
+        let human_text = append_trace(OutputBuilder::new(), trace, decoder, *full).build();
+        let builder = if decoder.decoding_warnings().is_empty() {
+            OutputBuilder::new()
+        } else {
+            let warning_message = format_decoding_warning(&decoder.decoding_warnings());
```
ddoktorski, line 174: nit: `decoder.decoding_warnings()` is called twice

crates/sncast/src/response/get/tx_trace.rs
```diff
@@ -0,0 +1,903 @@
...
+    label: &str,
+    invocation: Option<&FunctionInvocation>,
+    decoder: &TraceDecoder,
+    full: bool,
+) -> OutputBuilder {
+    if let Some(invocation) = invocation {
+        append_invocation_section(builder, label, invocation, decoder, full)
+    } else {
+        builder
+    }
```
ddoktorski, line 547: Can we use `OutputBuilder::if_some` instead? Also in `append_optional_state_diff`

crates/sncast/src/response/get/tx_trace.rs
```diff
@@ -0,0 +1,903 @@
...
+        }
+    }
+}
+
+#[derive(Default)]
+pub struct TraceDecoder {
+    sierra_abis: HashMap<ClassHash, Vec<AbiEntry>>,
+    legacy_class_hashes: HashSet<ClassHash>,
+    legacy_selectors: HashMap<(ClassHash, Felt), String>,
+    decoding_warnings: RefCell<BTreeSet<TraceDecodingWarning>>,
```
ddoktorski, line 233: Can we omit using interior mutability here?

docs/src/appendix/sncast/get/tx-trace.md
ddoktorski: Could you remind me why actually we can't use both these flags together? 🙏

crates/sncast/src/starknet_commands/get/tx_trace.rs
```diff
@@ -0,0 +1,174 @@
...
+}
+
+async fn fetch_contract_classes(
+    provider: &JsonRpcClient<HttpTransport>,
+    class_hashes: HashSet<ClassHash>,
+) -> FetchedContractClasses {
+    let results = stream::iter(class_hashes)
+        .map(|class_hash| async move {
+            match provider
+                .get_class(BlockId::Tag(BlockTag::PreConfirmed), *class_hash)
```
ddoktorski, line 96: Why pre confirmed and not latest?

crates/sncast/src/response/get/tx_trace.rs
```diff
@@ -0,0 +1,903 @@
...
+    decoder: &TraceDecoder,
+    indent: usize,
+) -> OutputBuilder {
+    let builder = builder
+        .with_indent(indent)
+        .field("Call Type", format_call_type(invocation.call_type))
+        .field("Calldata", &decoder.calldata(invocation))
+        .padded_felt_field("Caller Address", &invocation.caller_address)
+        .padded_felt_field("Class Hash", &invocation.class_hash)
+        .padded_felt_field("Contract Address", &invocation.contract_address)
```
ddoktorski, line 617:
```suggestion
        .contract_address(&invocation.contract_address)
```

docs/src/starknet/tx-status.md
ddoktorski, line 5: It can be updated

crates/sncast/src/response/get/tx_trace.rs
franciszekjob: Yup, addressed in https://github.com/foundry-rs/starknet-foundry/pull/4546/commits/691604e8c7aec1f3762dc93453cc73fef97f43d1

PR #4549, opened 2026-08-19
refactor(sncast) Introduce accounts repository

integraledelebesgue:
## Introduced changes
This PR introduces the `AccountRepository` - a layer between the physical storage of accounts and their corresponding schema-agnostic representations which abstracts filesystem operations from code that reads/modifies the accounts.

It is used later in the stack.

## Checklist

- [ ] Linked relevant issue
- [x] Added relevant tests


crates/sncast/src/accounts/repository.rs
```diff
@@ -0,0 +1,610 @@
...
+            source,
+        })?;
+        temporary
+            .write_all(contents)
+            .map_err(|source| AccountsError::Storage {
+                operation: "write temporary accounts file",
+                path: self.path.clone(),
+                source,
+            })?;
+        temporary.flush().map_err(|source| AccountsError::Storage {
```
franciszekjob: TIL: `flush` for `File` is noop.
https://github.com/rust-lang/rust/blob/16a623ad672a92409b5c04beb303583c6cf72a7e/library/std/src/fs.rs#L1439
But docs say it may change in future so let's leave it.

crates/sncast/src/accounts/repository.rs
```diff
@@ -0,0 +1,610 @@
...
+                return Err(AccountsError::DuplicateAccount {
+                    network: network_display,
+                    account: name_display,
+                });
+            }
+            Ok(())
+        })
+    }
+
+    pub fn remove(
```
franciszekjob, line 131: Same as above.

crates/sncast/src/accounts/repository.rs
```diff
@@ -0,0 +1,610 @@
...
+
+    #[test]
+    fn reports_missing_file() {
+        let (_directory, path) = path();
+
+        assert!(!AccountRepository::new(path).exists().unwrap());
+    }
+
+    #[test]
+    fn generates_first_account_name_when_file_is_missing() {
```
franciszekjob, line 456: Let's add a test which will test the logic of account name generation, when file is not empty, e.g. includes account-1 and account-2.

crates/sncast/src/accounts/repository.rs
```diff
@@ -0,0 +1,610 @@
...
+        if !self.exists()? {
+            return Err(AccountsError::FileNotFound {
+                path: self.path.clone(),
+            });
+        }
+        DecodedAccountRegistry::decode(&self.read()?)
+            .map_err(|error| attach_file_path(error, &self.path))
+    }
+
+    pub fn find(&self, network: &str, account: &str) -> Result<AccountRecord, AccountsError> {
```
franciszekjob, line 96: Could we add test for this method?

crates/sncast/src/accounts/repository.rs
```diff
@@ -0,0 +1,610 @@
...
+use tempfile::Builder;
+
+use crate::accounts::schema::{DecodedAccountRegistry, SourceVersion};
+use crate::accounts::{AccountName, AccountRecord, AccountRegistry, AccountsError, NetworkName};
+use crate::helpers::filesystem::{reject_symlink, set_secret_permissions};
+
+#[derive(Debug)]
+pub struct MutationResult<T> {
+    pub value: T,
+    pub migrated_from_v1: bool,
```
franciszekjob: `migrated_from_v1` is not really extensible for potential future versions. I think we can improve it, e.g. by introducing `MigrationOutcome` (or smth similar):


```rust
#[derive(Debug)]
pub enum MigrationOutcome {
    NotRequired {
        version: SourceVersion,
    },
    Performed {
        from: SourceVersion,
        to: SourceVersion,
        backup_path: Utf8PathBuf,
    },
}

#[derive(Debug)]
pub struct MutationResult<T> {
    pub value: T,
    pub migration_outcome: MigrationOutcome,
}
```

WDYT?

franciszekjob (commented):
One remark about migration: could we model it more generically (for potential future schema migrations) instead of coupling it specifically to v1 -> latest?

crates/sncast/src/accounts/error.rs
```diff
@@ -24,4 +32,21 @@ pub enum AccountsError {
...
+    },
+
+    #[error("accounts file `{path}` does not exist")]
+    FileNotFound { path: Utf8PathBuf },
+
+    #[error("account `{account}` already exists on network `{network}`")]
+    DuplicateAccount { network: String, account: String },
+
+    #[error("account `{account}` was not found on network `{network}`")]
+    AccountNotFound { network: String, account: String },
```
ddoktorski: nit: Can we keep error variants in alphabetical order?

docs/src/starknet/account.md
```diff
@@ -100,6 +100,30 @@ If you created an account with `sncast account create` it by default it will be
 `~/.starknet_accounts/starknet_open_zeppelin_accounts.json` file which we call `default accounts file` in the following
 sections.
 
+Native accounts files use a versioned schema. Version 2 stores networks below the `accounts` field and gives every account exactly one explicitly tagged signer:
+
+```json
+{
+  "version": 2,
+  "accounts": {
+    "alpha-sepolia": {
```
ddoktorski, line 109: What about replacing network names in V2 with just chain ids instead of strange `alpha-sepolia` etc.?

crates/sncast/src/accounts/repository.rs
```diff
@@ -0,0 +1,610 @@
...
+    })?;
+    Ok(parent)
+}
+
+fn lock_path(path: &Utf8Path) -> Utf8PathBuf {
+    let file_name = path.file_name().unwrap_or("accounts.json");
+    path.with_file_name(format!("{file_name}.lock"))
+}
+
+#[cfg(unix)]
```
ddoktorski: We don't support Windows anyway so we can rather drop all unix conditional compilations

crates/sncast/src/accounts/repository.rs
```diff
@@ -0,0 +1,610 @@
...
+            for name in accounts.keys() {
+                if let Some(id) = name
+                    .as_str()
+                    .strip_prefix("account-")
+                    .and_then(|id| id.parse::<u32>().ok())
+                {
+                    used_ids.insert(id);
+                }
+            }
+        }
```
ddoktorski: nit:
```suggestion
        let used_ids: HashSet<u32> = self.load()?
            .registry
            .networks()
            .values()
            .flat_map(|accounts| accounts.keys())
            .filter_map(|name| {
                name.as_str()
                    .strip_prefix("account-")
                    .and_then(|id| id.parse::<u32>().ok())
            })
            .collect();
```

crates/sncast/src/starknet_commands/account/migrate.rs
```diff
@@ -0,0 +1,55 @@
...
+#[derive(Args, Debug)]
+#[command(about = "Migrate an accounts file to the latest schema version")]
+pub struct Migrate;
+
+pub fn migrate(accounts_file: &Utf8PathBuf) -> Result<AccountMigrateResponse> {
+    let repository = AccountRepository::new(accounts_file.clone());
+    repository.load().map_err(|error| anyhow::anyhow!(error))?;
+    let result = repository
+        .mutate(|_| Ok(()))
+        .map_err(|error| anyhow::anyhow!(error))?;
```
ddoktorski: It seems redundant
```suggestion
        .mutate(|_| Ok(()))?;
```

crates/sncast/src/accounts/repository.rs
```diff
@@ -0,0 +1,610 @@
...
+
+    #[must_use]
+    pub fn v1_backup_path(&self) -> Utf8PathBuf {
+        v1_backup_path(&self.path)
+    }
+
+    pub fn generate_account_name(&self) -> anyhow::Result<String> {
+        let mut id = 1;
+
+        if !self.path.exists() {
```
ddoktorski: Should it be `!self.exists()`?

crates/sncast/src/accounts/repository.rs
```diff
@@ -0,0 +1,610 @@
...
+
+            Ok(MutationResult {
+                value,
+                migrated_from_v1,
+            })
+        })
+    }
+
+    pub fn exists(&self) -> Result<bool, AccountsError> {
+        reject_symlink(&self.path).map_err(|source| AccountsError::Storage {
```
ddoktorski: `map_err(|source| AccountsError::Storage` is repeated multiple times throughout the file, maybe we can somehow extract it to a helper or improve readability in some other way?

crates/sncast/src/accounts/repository.rs
```diff
@@ -0,0 +1,610 @@
...
+    let parent = path.parent().unwrap_or(Utf8Path::new("."));
+    fs::create_dir_all(parent).map_err(|source| AccountsError::Storage {
+        operation: "create accounts directory",
+        path: parent.to_owned(),
+        source,
+    })?;
+    Ok(parent)
+}
+
+fn lock_path(path: &Utf8Path) -> Utf8PathBuf {
```
ddoktorski: It's almost identical as `v1_backup_path`, can we merge them into 1 function with an extra `suffix` arg?

crates/sncast/src/accounts/repository.rs
```diff
@@ -0,0 +1,610 @@
...
+        &self,
+        operation: impl FnOnce(&mut AccountRegistry) -> Result<T, AccountsError>,
+    ) -> Result<MutationResult<T>, AccountsError> {
+        self.with_exclusive_lock(|| {
+            let existed = self.exists()?;
+            let original = if existed { self.read()? } else { Vec::new() };
+            let mut decoded = DecodedAccountRegistry::decode(&original)
+                .map_err(|error| attach_file_path(error, &self.path))?;
+            let value = operation(&mut decoded.registry)?;
+            let encoded = DecodedAccountRegistry::encode_v2(&decoded.registry)?;
```
ddoktorski: Why `decode` function maps error with `attach_file_path`, but `encode_v2` not?

ddoktorski (changes requested):

crates/sncast/src/accounts/repository.rs
integraledelebesgue: Added coverage for a non-empty accounts file containing `account-1` and `account-2`; the repository now generates `account-3`.

crates/sncast/src/accounts/repository.rs
integraledelebesgue: Added repository coverage for both finding an existing account and returning `AccountNotFound` for a missing one.

crates/sncast/src/accounts/repository.rs
integraledelebesgue: Replaced the boolean with `MigrationOutcome`, including source/target versions and the actual backup path for performed migrations.

crates/sncast/src/accounts/error.rs
integraledelebesgue: Alphabetized the error variants and kept the later `InvalidAccount` variant in order while replaying the stack.

crates/sncast/src/accounts/repository.rs
integraledelebesgue: Removed the Unix/non-Unix conditionals here and in the filesystem helper; permissions, directory syncing, symlink checks, and tests are now unconditionally Unix-based.

crates/sncast/src/accounts/repository.rs
integraledelebesgue: Applied the suggested iterator-based `HashSet<u32>` construction.

crates/sncast/src/starknet_commands/account/migrate.rs
integraledelebesgue: Removed the redundant `anyhow` mappings and now use `?` directly while retaining the initial `load()`.

crates/sncast/src/accounts/repository.rs
integraledelebesgue: Updated account-name generation to use `self.exists()?`.

crates/sncast/src/accounts/repository.rs
integraledelebesgue: Extracted the repeated storage-error construction into a shared mapping helper and applied it throughout the repository.

crates/sncast/src/accounts/repository.rs
integraledelebesgue: Merged the duplicated backup/lock path logic into a single `path_with_suffix` helper.

crates/sncast/src/accounts/repository.rs
integraledelebesgue: Encoding errors now receive the accounts-file path through the same mapping used for decoding errors.

crates/sncast/src/accounts/repository.rs
integraledelebesgue: Thanks for info! Let's keep it in case it gains a real behaviour in the future

docs/src/starknet/account.md
integraledelebesgue: I'm not sure if that's a good idea because network names are more readable than chain IDs

crates/sncast/src/accounts/repository.rs
```diff
@@ -0,0 +1,668 @@
...
+    pub fn path(&self) -> &Utf8Path {
+        &self.path
+    }
+
+    #[must_use]
+    pub fn v1_backup_path(&self) -> Utf8PathBuf {
+        path_with_suffix(&self.path, ".v1.bak")
+    }
+
+    pub fn generate_account_name(&self) -> anyhow::Result<String> {
```
franciszekjob: nit: Could it return `Result<_, AccountsError>` instead of `anyhow::Result<String>`?

crates/sncast/src/starknet_commands/account/migrate.rs
```diff
@@ -0,0 +1,69 @@
...
+        let response = migrate(&path).unwrap();
+
+        assert!(response.migrated);
+        assert_eq!(response.backup, Some(format!("{path}.v1.bak")));
+        assert_eq!(
+            serde_json::from_str::<serde_json::Value>(&fs::read_to_string(&path).unwrap()).unwrap()
+                ["version"],
+            2
+        );
+    }
```
franciszekjob: Would it make sense to add e2e test for `sncast account migrate` as well?

crates/sncast/src/response/account/migrate.rs
```diff
@@ -0,0 +1,25 @@
...
+            "Accounts file migrated to version 2"
+        } else {
+            "Accounts file is already version 2"
+        };
+        styling::OutputBuilder::new()
+            .success_message(message)
+            .if_some(self.backup.as_ref(), |builder, backup| {
+                builder.field("V1 Backup", backup)
+            })
+            .build()
```
franciszekjob: We made the migration result extensible for future schema versions, so this should be updated as well.

crates/sncast/src/response/account/migrate.rs
```diff
@@ -0,0 +1,25 @@
+use crate::response::cast_message::SncastCommandMessage;
+use foundry_ui::styling;
+use serde::Serialize;
+
+#[derive(Clone, Debug, Serialize)]
+pub struct AccountMigrateResponse {
+    pub migrated: bool,
+    pub backup: Option<String>,
```
franciszekjob: Let's change `String` to path type.

integraledelebesgue: Done. I also changed the structure a little bit - `AccountMigrateResponse` now has a `MigrationOutcome`

crates/sncast/src/response/account/migrate.rs
integraledelebesgue: Good catch, forgot about it. Done

crates/sncast/src/starknet_commands/account/migrate.rs
integraledelebesgue: Yes, a lot more sense than unit tests

crates/sncast/src/accounts/repository.rs
integraledelebesgue: Good idea, done. `anyhow::Error` leaked from the previous implementation of this function

crates/sncast/src/starknet_commands/account/migrate.rs
integraledelebesgue: I added some e2e tests: https://github.com/foundry-rs/starknet-foundry/blob/66a75c305a8038d1882201041f611b5711c15514/crates/sncast/tests/e2e/account/migrate.rs

crates/sncast/src/accounts/error.rs
```diff
@@ -1,11 +1,23 @@
+use std::fmt::Display;
+
+use camino::Utf8PathBuf;
 use thiserror::Error;
 
 /// Errors raised while validating and resolving native account data.
 #[derive(Debug, Error)]
 pub enum AccountsError {
+    #[error("account `{account}` was not found on network `{network}`")]
```
ddoktorski:
```suggestion
    #[error("account `{account}` was not found for network `{network}`")]
```
nit: `on` suggests on-chain state rather than local config state. I'd also rephrase `DuplicateAccount` message

crates/sncast/tests/e2e/account/migrate.rs
```diff
@@ -0,0 +1,120 @@
+use crate::helpers::{fixtures::get_accounts_path, runner::runner};
+use camino::Utf8Path;
+use indoc::{formatdoc, indoc};
+use std::fs::{self};
```
ddoktorski: nit:
```suggestion
use std::fs;
```

crates/sncast/src/accounts/error.rs
```diff
@@ -24,19 +36,96 @@ pub enum AccountsError {
...
+        )
+    }
+}
+
+#[derive(Clone, Copy, Debug)]
+pub enum FileType {
+    AccountsFile,
+    TemporaryAccountsFile,
+    Backup,
+    ParentDirectory,
```
ddoktorski, line 118: Why is it a variant of `FileType`?

crates/sncast/src/accounts/repository.rs
```diff
@@ -0,0 +1,733 @@
...
+                .get_mut(network)
+                .and_then(|accounts| accounts.remove(name))
+                .ok_or_else(|| AccountsError::AccountNotFound {
+                    network: network.to_owned(),
+                    account: name.to_owned(),
+                })
+        })
+    }
+
+    pub fn mutate<T>(
```
ddoktorski: `mutate` seems a bit vague name, what about `update`?

crates/sncast/src/accounts/repository.rs
```diff
@@ -0,0 +1,733 @@
...
+    fs::create_dir_all(parent).map_err(map_storage_error(
+        StorageOperation::Create,
+        FileType::ParentDirectory,
+        parent,
+    ))?;
+
+    Ok(parent)
+}
+
+fn set_secret_open_options(options: &mut OpenOptions) {
```
ddoktorski:
```suggestion
fn set_owner_only_permissions(options: &mut OpenOptions) {
```

crates/sncast/src/accounts/error.rs
```diff
@@ -1,11 +1,23 @@
+use std::fmt::Display;
```
ddoktorski: nit: Import also `Formatter` and `Result` for consistency

crates/sncast/tests/e2e/account/migrate.rs
```diff
@@ -0,0 +1,120 @@
...
+
+    let snapbox = runner(&args).current_dir(tempdir.path());
+
+    snapbox.assert().stderr_eq(indoc!(
+        r"
+        Command: account migrate
+        Error: invalid schema of field alpha-sepolia.? in the accounts file: expected `,` or `}` at line 8 column 9
+
+        Caused by:
+            expected `,` or `}` at line 8 column 9
```
ddoktorski: Should we keep `source: serde_json::Error` in error variants and display them? It seems to only duplicate the error

crates/sncast/src/accounts/repository.rs
```diff
@@ -0,0 +1,733 @@
...
+        file.sync_all().map_err(map_storage_error(
+            StorageOperation::Sync,
+            FileType::Backup,
+            path,
+        ))?;
+
+        sync_parent(parent, path)
+    }
+
+    fn with_exclusive_lock<T>(
```
ddoktorski, line 298: Can we simplify the code so this function only locks and returns `File`? (can be renamed to `lock_exclusive`)

crates/sncast/src/accounts/schema/mod.rs
```diff
@@ -14,6 +16,19 @@ pub enum SourceVersion {
     V2,
 }
 
+impl Display for SourceVersion {
```
ddoktorski: Let's replace all `Display` implementations with `strum_macros::Display` where possible

crates/sncast/src/accounts/repository.rs
```diff
@@ -0,0 +1,733 @@
...
+        ))?;
+
+        sync_parent(parent, &self.path)?;
+
+        result
+    }
+}
+
+fn path_with_suffix(path: &Utf8Path, suffix: &str) -> Utf8PathBuf {
+    let file_name = path.file_name().unwrap_or("accounts.json");
```
ddoktorski: What if someone passes `..` or `/` as `--accounts-file`?

crates/sncast/src/starknet_commands/account/migrate.rs
```diff
@@ -0,0 +1,19 @@
...
+
+#[derive(Args, Debug)]
+#[command(about = "Migrate an accounts file to the latest schema version")]
+pub struct Migrate;
+
+pub fn migrate(accounts_file: &Utf8PathBuf) -> Result<AccountMigrateResponse> {
+    let repository = AccountRepository::new(accounts_file.clone());
+    repository.load()?;
+
+    let result = repository.mutate(|_| Ok(()))?;
```
ddoktorski: Maybe add `migrate` function on `AccountRepository` to simplify code here?

ddoktorski (changes requested):
PR title seems inaccurate, it is not only refactor as it introduces `account migrate` command

crates/sncast/src/accounts/error.rs
integraledelebesgue: It simplifies the construction of errors related to files. We don't need to repeat a nearly identical message for every type of file that may produce a given kind of error

crates/sncast/src/accounts/repository.rs
integraledelebesgue: I think the current implementation is optimal. Returning a plain `File` is not possible because it needs to be unlocked after performing the operation anyway. We could wrap it with a guard with custom `Drop`, but it will be a more complex solution than this.

crates/sncast/tests/e2e/account/migrate.rs
integraledelebesgue: Not relevant anymore since I stick to the convention of `#[source]` error that we discussed offline

crates/sncast/src/accounts/error.rs
integraledelebesgue: Outdated - I use `strum_macros`

crates/sncast/src/accounts/repository.rs
integraledelebesgue: Done

crates/sncast/tests/e2e/account/migrate.rs
integraledelebesgue: Done

crates/sncast/src/accounts/error.rs
integraledelebesgue: Done

crates/sncast/src/accounts/repository.rs
integraledelebesgue: Done

crates/sncast/src/accounts/schema/mod.rs
integraledelebesgue: Done

crates/sncast/src/accounts/repository.rs
integraledelebesgue: It's verified as a regular file in the beginning

crates/sncast/src/starknet_commands/account/migrate.rs
integraledelebesgue: Done

crates/sncast/src/accounts/error.rs
```diff
@@ -1,11 +1,21 @@
+use camino::Utf8PathBuf;
 use thiserror::Error;
 
 /// Errors raised while validating and resolving native account data.
 #[derive(Debug, Error)]
 pub enum AccountsError {
+    #[error("account `{account}` was not found for network `{network}`")]
+    AccountNotFound { network: String, account: String },
+
+    #[error("account `{account}` already exists on network `{network}`")]
```
ddoktorski, line 10:
```suggestion
    #[error("account `{account}` already exists for network `{network}`")]
```
The same suggestion as for `AccountNotFound`

docs/src/starknet/account.md
ddoktorski: bump

crates/sncast/src/accounts/repository.rs
```diff
@@ -0,0 +1,725 @@
...
+        ensure_regular_file(&path.to_path_buf(), false).map_err(map_storage_error(
+            StorageOperation::Inspect,
+            FileType::AccountsFile,
+            &path,
+        ))?;
+
+        let parent = ensure_parent(&path)?;
+
+        let mut options = OpenOptions::new();
+        options.write(true).create(true).mode(0o600);
```
ddoktorski, line 273: It's rather an edge case if possible at all, but we can add `truncate(true)` so it has correct content at the end if file already existed

crates/sncast/src/starknet_commands/account/migrate.rs
```diff
@@ -0,0 +1,17 @@
...
+use clap::Args;
+
+use sncast::accounts::AccountRepository;
+use sncast::response::account::migrate::AccountMigrateResponse;
+
+#[derive(Args, Debug)]
+#[command(about = "Migrate an accounts file to the latest schema version")]
+pub struct Migrate;
+
+pub fn migrate(accounts_file: &Utf8PathBuf) -> Result<AccountMigrateResponse> {
```
ddoktorski, line 12: Currently running `sncast --accounts-file missing.json account migrate` creates files `missing.json` and `missing.json.lock` and reports success

crates/sncast/src/accounts/repository.rs
```diff
@@ -0,0 +1,725 @@
...
+        ))?;
+
+        sync_parent(parent, &self.path)?;
+
+        result
+    }
+}
+
+fn path_with_suffix(path: &Utf8Path, suffix: &str) -> Utf8PathBuf {
+    let file_name = path.file_name().expect("regular file should have a name");
```
ddoktorski, line 341: This panic is reachable by user input, try to run `sncast --accounts-file sub/.. account migrate` (`sub` doesn't exist)

crates/sncast/src/helpers/filesystem.rs
```diff
@@ -0,0 +1,25 @@
+use std::fs;
+use std::io::{self, ErrorKind};
+use std::os::unix::fs::PermissionsExt;
+use std::path::Path;
+
+pub(crate) fn ensure_regular_file(path: &impl AsRef<Path>, must_exist: bool) -> io::Result<()> {
```
ddoktorski, line 6: This function only rejects dirs, but still accept non-regular files (e.g. special FIFO file). Should we use `metadata.is_file()` here?

ddoktorski, line 6: Currently this function is alway called with `must_fix == false`, is it going to change in next PRs?

crates/sncast/tests/e2e/account/migrate.rs
```diff
@@ -0,0 +1,120 @@
+use crate::helpers::{fixtures::get_accounts_path, runner::runner};
+use camino::Utf8Path;
+use indoc::{formatdoc, indoc};
+use std::fs;
+use tempfile::tempdir;
+
+#[tokio::test]
```
ddoktorski, line 7: Do we need to use `#[tokio::test]` and `async` for these tests?

crates/sncast/src/accounts/error.rs
```diff
@@ -50,3 +69,54 @@ pub enum AccountsFileError {
...
+
+#[derive(Clone, Copy, Debug, strum_macros::Display)]
+pub enum FileType {
+    #[strum(to_string = "accounts file")]
+    AccountsFile,
+
+    #[strum(to_string = "temporary accounts file")]
+    TemporaryAccountsFile,
+
+    #[strum(to_string = "v1 backup of the accounts file")]
```
ddoktorski, line 114: It shouldn't have `v1` hardcoded

crates/sncast/src/helpers/filesystem.rs
ddoktorski, line 6:
```suggestion
pub(crate) fn ensure_regular_file(path: &Utf8PathBuf, must_exist: bool) -> io::Result<()> {
```

crates/sncast/src/accounts/repository.rs
```diff
@@ -0,0 +1,725 @@
...
+    pub migration_outcome: MigrationOutcome,
+}
+
+#[derive(Clone, Debug)]
+pub struct AccountRepository {
+    path: Utf8PathBuf,
+}
+
+impl AccountRepository {
+    pub fn new(path: Utf8PathBuf) -> Result<Self, AccountsError> {
```
ddoktorski, line 40: I think `new` shouldn't return `Result`, it should be just a wrapper with validation run on a different methods

crates/sncast/src/accounts/repository.rs
```diff
@@ -0,0 +1,725 @@
...
+
+        let result = operation();
+
+        FileExt::unlock(&lock).map_err(map_storage_error(
+            StorageOperation::Unlock,
+            FileType::AccountsFile,
+            &self.path,
+        ))?;
+
+        sync_parent(parent, &self.path)?;
```
ddoktorski, line 334: If any of these fails the actual error returned by `operation()` can be dropped and only some I/O error is displayed

crates/sncast/src/accounts/repository.rs
ddoktorski: It is hard to follow that first `update` takes closure as an argument which is called in a closure that is passed to `with_exclusive_lock`. Maybe we can somehow simplify it? 🙏

ddoktorski (changes requested):

PR #4568, opened 2026-08-25
Display decoded event and contract's full path in `assert_emitted`/`assert_not_emitted`

### LINKED ISSUE #2989: Report Event name when `assert_emmited!` fails

Forge

### Feature Request

The error when an event is not emitted when using `assert_emitted` is not very informative.

There could be several events tested in a single contract in a single test.
```
    spy
        .assert_emitted(
            @array![
                (
...
```

Currently, the error reported is 
```
ailure data:
    "Event with matching data and keys was not emitted from 1685819267771795056935952165641627265001175894081276381033716133281735439506"
```

Would be more useful to get the missing event name and contract name

integraledelebesgue:
Closes #2989

## Introduced changes
- New cheatcode `format_event`, wrapped by `format_event` in `snforge_std`, allows reverse-transformation of the event at runtime
- Now, the error message from `assert_emitted` has a form:
```
Event <event> was not emitted from <contract-full-path> at <contract-address>
```
- `assert_not_emitted` has analogical behaviour

## Checklist

- [x] Linked relevant issue
- [x] Added relevant tests

### DIFF
--- a/crates/cheatnet/src/runtime_extensions/forge_runtime_extension/cheatcodes/event_formatter.rs
+++ b/crates/cheatnet/src/runtime_extensions/forge_runtime_extension/cheatcodes/event_formatter.rs
@@ -1,0 +1,64 @@
+use std::str::FromStr;
+
+use crate::runtime_extensions::forge_runtime_extension::contracts_data::ContractsData;
+use conversions::FromConv;
+use data_transformer::reverse_transform_event;
+use starknet_api::core::{ClassHash, ContractAddress};
+use starknet_rust::core::types::contract::AbiEntry;
+use starknet_types_core::felt::Felt;
+
+pub struct FormattedEvent {
+    pub event: String,
+    pub contract_address: String,
+    pub contract_full_module_path: Option<String>,
+}
+
+impl FormattedEvent {
+    pub fn new(
+        contracts_data: &ContractsData,
+        contract_address: ContractAddress,
+        class_hash: Option<ClassHash>,
+        keys: &[Felt],
+        data: &[Felt],
+    ) -> Self {
+        let contract_address = format!("{:#x}", Felt::from_(contract_address));
+
+        let contract_full_module_path = class_hash
+            .and_then(|class_hash| contracts_data.get_contract_module_path(&class_hash))
+            .map(String::from);
+
+        let event = class_hash
+            .and_then(|class_hash| {
+                let contract = contracts_data.get_contract_by_class_hash(&class_hash)?;
+                let sierra = serde_json::Value::from_str(&contract.artifacts.sierra).ok()?;
+                let abi = sierra.get("abi").cloned()?;
+                let abi = serde_json::from_value::<Vec<AbiEntry>>(abi).ok()?;
+                reverse_transform_event(keys, data, &abi).ok()
+            })
+            .unwrap_or_else(|| format_raw_event(keys, data));
+
+        Self {
+            event,
+            contract_address,
+            contract_full_module_path,
+        }
+    }
+}
+
+fn format_raw_event(keys: &[Felt], data: &[Felt]) -> String {
+    format!(
+        "{{ keys: {}, data: {} }}",
+        format_felt_slice(keys),
+        format_felt_slice(data)
+    )
+}
+
+fn format_felt_slice(felts: &[Felt]) -> String {
+    let felts = felts
+        .iter()
+        .map(|felt| format!("{felt:#x}"))
+        .collect::<Vec<_>>()
+        .join(", ");
+
+    format!("[{felts}]")
+}
--- a/crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
+++ b/crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
@@ -1,5 +1,6 @@
 use self::contracts_data::ContractsData;
 use crate::runtime_extensions::common::sum_syscall_usage;
+use crate::runtime_extensions::forge_runtime_extension::cheatcodes::event_formatter::FormattedEvent;
 use crate::runtime_extensions::forge_runtime_extension::cheatcodes::replace_bytecode::ReplaceBytecodeError;
 use crate::runtime_extensions::outer_call_runtime_extension::rpc::UsedResources;
 use crate::runtime_extensions::{
@@ -327,6 +328,46 @@ impl<'a> ExtensionLogic for ForgeExtension<'a> {
                     .get_events(events_offset);
 
                 Ok(CheatcodeHandlingResult::from_serializable(events))
+            }
+            "format_event" => {
+                let contract_address = input_reader.read()?;
+                let keys: Vec<Felt> = input_reader.read()?;
+                let data: Vec<Felt> = input_reader.read()?;
+
+                let state = &extended_runtime
+                    .extended_runtime
+                    .extended_runtime
+                    .hint_handler
+                    .base
+                    .state;
+
+                let replacement_class_hash = extended_runtime
+                    .extended_runtime
+                    .extension
+                    .cheatnet_state
+                    .replaced_bytecode_contracts
+                    .get(&contract_address)
+                    .copied();
+
+                let class_hash = replacement_class_hash
+                    .or_else(|| state.get_class_hash_at(contract_address).ok());
+
+                let formatted_event = FormattedEvent::new(
+                    self.contracts_data,
+                    contract_address,
+                    class_hash,
+                    &keys,
+                    &data,
+                );
+
+                Ok(CheatcodeHandlingResult::from_serializable((
+                    ByteArray::from(formatted_event.event.as_str()),
+                    ByteArray::from(formatted_event.contract_address.as_str()),
+                    formatted_event
+                        .contract_full_module_path
+                        .as_deref()
+                        .map(ByteArray::from),
+                )))
             }
             "spy_messages_to_l1" => {
                 let messages_offset = extended_runtime
--- a/crates/forge/tests/integration/spy_events.rs
+++ b/crates/forge/tests/integration/spy_events.rs
@@ -78,6 +78,145 @@ fn spy_events_simple() {
 }
 
 #[test]
+fn spy_events_emitted_by_library_call() {
+    let test = test_case!(
+        indoc!(
+            r#"
+            use array::ArrayTrait;
+            use result::ResultTrait;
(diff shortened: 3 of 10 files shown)


crates/cheatnet/src/runtime_extensions/forge_runtime_extension/cheatcodes/event_formatter.rs
```diff
@@ -0,0 +1,40 @@
...
+/// Formats an event assertion target using the ABI of the contract deployed at `contract_address`.
+#[must_use]
+pub fn format_event(
+    state: &mut dyn State,
+    contracts_data: &ContractsData,
+    contract_address: ContractAddress,
+    keys: &[Felt],
+    data: &[Felt],
+) -> (String, String) {
+    let raw_event = "Raw event with matching keys and data".to_string();
```
ddoktorski:
```suggestion
    let raw_event = "Event with matching keys and data".to_string();
```
I'd rather omit the word `Raw`, it seems more confusing that helpful

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/cheatcodes/event_formatter.rs
```diff
@@ -0,0 +1,40 @@
...
+    let address = Felt::from_(contract_address);
+    let unknown_contract = format!("unknown contract at {address:#x}");
+
+    let Ok(class_hash) = state.get_class_hash_at(contract_address) else {
```
ddoktorski: Should we cover an edge case when a given contract address has its class replaced with `replace_bytecode`? 🤔

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/cheatcodes/event_formatter.rs
ddoktorski: Also maybe in this case include `keys` and `data`?

crates/forge/tests/integration/spy_events.rs
```diff
@@ -147,9 +147,8 @@ fn assert_emitted_fails() {
     assert_case_output_contains(
         &result,
         "assert_emitted_fails",
-        "Event with matching data and",
+        "Event FirstEvent { some_data: 0x7b } was not emitted from contract::SpyEventsChecker at",
```
franciszekjob, line 289: Unrelated to this PR, but we could improve `assert_case_output_contains` to support the `[..]` wildcard (same as `assert_stdout_contains` does). This would let us explicitly verify that the message includes a contract address:

```
Event FirstEvent { some_data: 0x7b } was not emitted from contract::SpyEventsChecker at 0x[..]
```

ATM, this is technically equivalent to ending the expected message with `at 0x`, but using `[..]` better communicates that there should be some dynamic value there.

crates/forge/tests/integration/spy_events.rs
franciszekjob, line 1: Could we add a test for formatting an event emitted via library call?

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/cheatcodes/event_formatter.rs
franciszekjob, line 1: Could we separate resolving the emitter location from selecting the ABI used to decode the event? They are the same for direct calls, but can differ for `library_call` and `replace_bytecode`.

crates/forge/tests/integration/spy_events.rs
integraledelebesgue: Done: https://github.com/foundry-rs/starknet-foundry/blob/d27dcb56152761914aa2a586481a59b25220130d/crates/forge/tests/integration/spy_events.rs#L81-L217

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/cheatcodes/event_formatter.rs
integraledelebesgue: Done: https://github.com/foundry-rs/starknet-foundry/blob/a1507fa4e87208b3f6f849e111e1e6b5745c4a22/crates/forge/tests/integration/spy_events.rs#L1002-L1302

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/cheatcodes/event_formatter.rs
integraledelebesgue: Done

crates/forge/tests/integration/spy_events.rs
integraledelebesgue: I added a task: https://github.com/foundry-rs/starknet-foundry/issues/4583

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/cheatcodes/event_formatter.rs
integraledelebesgue: Events emitted using `library_call` cannot be decoded in the current implementation of `cheatnet` because the class hash of the library contract is not available anywhere, making the appropriate ABI unavailable, too. We would need to somehow save the class hash of the library contract at the very moment the call emits the event.

`replace_bytecode` is handled properly since the runtime class hash of the emitter is exactly the one that is linked to the ABI used for decoding.

Summing up: I think it's not necessary now to change the construction of the decoder, but we can add a task to extend the cheatnet to allow handling library calls properly.

snforge_std/src/cheatcodes/events.cairo
```diff
@@ -68,8 +68,14 @@ impl EventSpyAssertionsTraitImpl<T, +starknet::Event<T>, +Drop<T>> of EventSpyAs
 
         for (from, event) in events.span() {
             if !received_events.is_emitted(*from, event) {
-                let from: felt252 = (*from).into();
-                panic!("Event with matching data and keys was not emitted from {}", from);
+                let (event, contract_address, contract_full_path) = format_event(*from, event);
+                let contract_full_path = contract_full_path.unwrap_or("contract");
```
franciszekjob, line 72: nit: WDYT?
```suggestion
                let contract_full_path = contract_full_path.unwrap_or("unknown contract");
```

franciszekjob (approved):

snforge_std/src/cheatcodes/events.cairo
integraledelebesgue: I'm not sure if this is necessary. "unknown" may suggest some kind of additional error while it's a side-effect of our implementation that shouldn't concern users

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
```diff
@@ -328,6 +329,46 @@ impl<'a> ExtensionLogic for ForgeExtension<'a> {
 
                 Ok(CheatcodeHandlingResult::from_serializable(events))
             }
+            "format_event" => {
+                let contract_address = input_reader.read()?;
+                let keys: Vec<Felt> = input_reader.read()?;
+                let data: Vec<Felt> = input_reader.read()?;
+
+                let state = &mut extended_runtime
```
ddoktorski: nit:
```suggestion
                let state = &extended_runtime
```

crates/forge/tests/integration/spy_events.rs
```diff
@@ -488,6 +621,61 @@ fn emit_unnamed_event() {
...
+    assert_case_output_contains(
+        &result,
+        "emit_unnamed_event_fails",
+        "Event { keys: [0x7b], data: [0x1c9] } was not emitted from contract::SpyEventsChecker at",
+    );
+    assert_case_output_contains(
+        &result,
+        "emit_unnamed_event_fails",
+        "contract::SpyEventsChecker",
+    );
```
ddoktorski: nit: Seems redundant as these are substrings of checked data in the `assert_case_output_contains` above

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/cheatcodes/event_formatter.rs
```diff
@@ -0,0 +1,60 @@
...
+
+        Self {
+            event,
+            contract_address,
+            contract_full_module_path,
+        }
+    }
+}
+
+fn raw_event(keys: &[Felt], data: &[Felt]) -> String {
```
ddoktorski: nit:
```suggestion
fn format_raw_event(keys: &[Felt], data: &[Felt]) -> String {
```

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/cheatcodes/event_formatter.rs
```diff
@@ -0,0 +1,60 @@
...
+        let contract_address = format!("{:#x}", Felt::from_(contract_address));
+
+        let contract_full_module_path = class_hash
+            .and_then(|class_hash| contracts_data.get_contract_module_path(&class_hash))
+            .map(String::from);
+
+        let event = class_hash
+            .and_then(|class_hash| contracts_data.get_contract_by_class_hash(&class_hash))
+            .and_then(|contract| {
+                serde_json::from_str::<SierraClass>(&contract.artifacts.sierra).ok()
```
ddoktorski: nit: Would it be possible to parse only ABI instead of entire Sierra?

ddoktorski (approved):

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/mod.rs
integraledelebesgue: Done

crates/forge/tests/integration/spy_events.rs
integraledelebesgue: Indeed, didn't notice that

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/cheatcodes/event_formatter.rs
integraledelebesgue: Done

crates/cheatnet/src/runtime_extensions/forge_runtime_extension/cheatcodes/event_formatter.rs
integraledelebesgue: I now parse the Sierra string to a `Value`, extract the `abi` and deserialize it

PR #4577, opened 2026-09-01
Trigger maat after release

hakiers:
## Summary

Adds an automated step to the `release.yml` workflow that triggers the `experiment.yml` workflow in `software-mansion/maat` whenever a new Foundry release is published.

The job extracts the clean version string from the tag (`vX.Y.Z` → `X.Y.Z`) and invokes the remote `maat` workflow via the GitHub CLI.

---

> ⚠️ **Prerequisite Setup Required**
>
> This workflow requires cross-repository Actions permissions to trigger `software-mansion/maat`.
>
> 1. **Generate PAT:** Create a Fine-Grained Personal Access Token (or Classic PAT) with access to `software-mansion/maat` [Link](https://github.com/settings/tokens/new).
>    * **Permissions:** 
image

> 2. **Add Secret:** Save the token in repository secrets:
>    * **Secret Name:** `MAAT_EXPERIMENTS_DISPATCH_TOKEN`

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -190,3 +190,30 @@ jobs:
     with:
       plugin_dep_version: ${{ needs.verify-version.outputs.version }}
       prod_registry: true
+
+  trigger-maat-experiment:
+    name: Trigger Maat experiment
+    runs-on: ubuntu-latest
+    needs: [ verify-version, publish-to-registry ]
+    steps:
+      - name: Extract version from tag
+        env:
+          RAW_REF: ${{ github.ref_name }}
+        run: |
+          VERSION=${RAW_REF}
+
+          # Strip 'v' prefix from the version if it exists
+          VERSION=${VERSION#v}
+
+          echo "VERSION=${VERSION}" >> $GITHUB_ENV
+          
+      - name: Trigger Maat experiment
+        env:
+          GH_TOKEN: ${{ secrets.MAAT_EXPERIMENTS_DISPATCH_TOKEN }}
+        run: |
+          gh workflow run experiment.yml \
+            --repo software-mansion/maat \
+            --ref main \
+            -f workspace="release" \
+            -f foundry="$VERSION" \
+            -f notify_team="true"


.github/workflows/release.yml
```diff
@@ -190,3 +190,30 @@ jobs:
...
       prod_registry: true
+
+  trigger-maat-experiment:
+    name: Trigger Maat experiment
+    runs-on: ubuntu-latest
+    needs: [ verify-version, publish-to-registry ]
+    steps:
+      - name: Extract version from tag
+        env:
+          RAW_REF: ${{ github.ref_name }}
```
franciszekjob, line 201: This workflow also runs on pushes to master, where `github.ref_name` is `master`. Could we use `${{ needs.verify-version.outputs.version }}` directly here?

.github/workflows/release.yml
```diff
@@ -190,3 +190,30 @@ jobs:
...
+          # Strip 'v' prefix from the version if it exists
+          VERSION=${VERSION#v}
+
+          echo "VERSION=${VERSION}" >> $GITHUB_ENV
+          
+      - name: Trigger Maat experiment
+        env:
+          GH_TOKEN: ${{ secrets.MAAT_EXPERIMENTS_DISPATCH_TOKEN }}
+        run: |
+          gh workflow run experiment.yml \
```
franciszekjob, line 214: At this point, the GitHub release is still a draft, so its binaries are not publicly available yet. Maat installs foundry from those release binaries, so the experiment will fail. It seems we should trigger it only after the release is published.

.github/workflows/release.yml
```diff
@@ -190,3 +190,30 @@ jobs:
     with:
       plugin_dep_version: ${{ needs.verify-version.outputs.version }}
       prod_registry: true
+
+  trigger-maat-experiment:
+    name: Trigger Maat experiment
+    runs-on: ubuntu-latest
+    needs: [ verify-version, publish-to-registry ]
+    steps:
+      - name: Extract version from tag
```
ddoktorski, line 199: Do we need to check the version given that we always want to run it with latest?

.github/workflows/release.yml
```diff
@@ -190,3 +190,30 @@ jobs:
...
+          VERSION=${RAW_REF}
+
+          # Strip 'v' prefix from the version if it exists
+          VERSION=${VERSION#v}
+
+          echo "VERSION=${VERSION}" >> $GITHUB_ENV
+          
+      - name: Trigger Maat experiment
+        env:
+          GH_TOKEN: ${{ secrets.MAAT_EXPERIMENTS_DISPATCH_TOKEN }}
```
ddoktorski, line 212:
```suggestion
          GH_TOKEN: ${{ secrets.MAAT_CONTENTS_READ_ACTIONS_WRITE }}
```
Let's use this name so it aligns with the name used in the nightly repo: https://github.com/software-mansion-labs/starknet-foundry-nightlies/blob/main/.github/workflows/maat-experiment.yml#L31 (we can also reuse the same token)

ddoktorski (commented):
Can we make it a separate workflow that triggers on a new release?

PR #4594, opened 2026-09-16
Release 0.64.0

ddoktorski:


franciszekjob (approved):

Cargo.lock
piotmag769, line 1: Why are there all of a sudden changes in locks? Did you run `cargo update` by accident?

ddoktorski: It is part of the release script: https://github.com/foundry-rs/starknet-foundry/blob/master/scripts/release.sh#L42

integraledelebesgue (approved):

